Issue #52, opened 2017-02-12
Windows event notification


njsmith:
# Problem

Windows has 3 incompatible families of event notifications APIs: IOCP, `select`/`WSAPoll`, and `WaitForMultipleEvents`-and-variants. They each have unique capabilities. This means: if you want to be able to react to all the different possible events that Windows can signal, then you *must use all 3 of these*. Needless to say, this creates a challenge for event loop design. There are a number of potentially viable ways to arrange these pieces; the question is which one we should use.

(Actually, all 3 together still isn't sufficient, b/c there are some things that still require threads – like console IO – and I'm ignoring GUI events entirely because Trio isn't a GUI library. But never mind. Just remember that when someone tells you that Windows' I/O subsystem is great, that their statement isn't *wrong* but does require taking a certain narrow perspective...)

# Considerations

## The WaitFor*Event family

The `Event`-related APIs are necessary to, for example, wait for a notification that a child process has exited. (The job object API provides a way to request IOCP notifications about process death, but the docs warn that the notifications are lossy and therefore useless...) Otherwise though they're very limited – in particular they have both O(n) behavior and max 64 objects in an interest set – so you definitely don't want to use these as your primary blocking call. We're going to be calling these in a background thread of some kind. The two natural architectures are to use `WaitForSingleObject(Ex)` and allocate one-thread-per-event, or else use `WaitForMultipleObjects(Ex)` and try and coalesce up to 64 events into each thread (substantially more complicated to implement but with 64x less memory overhead for thread stacks, if it matters). This is orthogonal to the rest of this issue, so it gets its own thread: #233

## IOCP

IOCP is the crown jewel of Windows the I/O subsystem, and what you generally hear recommended. It follows a natively asynchronous model where you just go ahead and issue a read or write or whatever, and it runs in the background until eventually the kernel tells you it's done. It provides an O(1) notification mechanism. It's pretty slick. But... it's not as obvious a choice as everyone makes it sound. (Did you know the Chrome team has mostly given up on trying to make it work?)

Issues:

[...]

Issue #26, opened 2017-01-22
Design: kqueue "power user" interface


njsmith:
Hasn't really been thought through, needs a good look over

njsmith:
One definite change is that instead of returning a new `UnboundedQueue`, these should have you pass in an existing `UnboundedQueue`. This allows multiple event streams to be multiplexed onto a single queue. (In the Windows case this also means our completion event objects should include the completion key as a field.)

auvipy:
is there anything trio could be benefitted from this thread?

njsmith:
@auvipy Sorry, I don't quite understand. Are you asking whether there's anything you can do to help with this, or...?

auvipy:
yes, and if there is anything important we get from the problem described here? I am considering trio as an option beside asyncio to be used as the foundational event loop library for celery 5 :)

auvipy:
so knowing the internals of trio, and its potential API breakage is very important to me for considering trio very seriously :)

njsmith:
@auvipy ah, I see!

This particular issue is about some very low-level APIs, that are used to get direct access to exotic features of the underlying OS – like if you need to access Windows's IOCP interfaces directly for some reason, or the [raw BSD `kqueue` APIs](https://www.freebsd.org/cgi/man.cgi?kqueue). They're useful for some kinds of projects, but most people will never touch them. (Asyncio doesn't give you any way to use these APIs at all ;-).) For some more details on what this might look like for kqueue, see #578 and #579. For more background on Windows, see #52. If this still sounds like something you're interested in, I can say more :-). I doubt this will be very relevant to celery though.

About the more general questions about Trio's internals, potential API breakage, and whether Trio would be a good fit for celery: I would love to talk about this! (And the short answer is: it depends on your time frame.) But, I would rather not have that discussion here on this issue, because it will be confusing later :-). I have two ideas, take your pick:

* We just (like, yesterday) created a new discourse forum to talk about Trio, that's available at: https://trio.discourse.group. It's very empty right now, but the plan is to move this kind of discussion over there. Would you be up for posting in the [Help and advice area](https://trio.discourse.group/c/help-and-advice)? (You can use your github login to post. I think.)

* Or if you'd rather stay on Github, that'd be fine too... but can you open a new issue?

Thanks!

auvipy:
thanks a lot for your detailed response :) I will do what seems more appropriate and right place to do :)

njsmith:
[...]

PR #81, opened 2017-03-11
Fix some typos/missing words

brettcannon:


dimaqq:
Wow codecov is so confused!

brettcannon:
@dimaqq actually it's accurate since I did add lines to the project that didn't increase the coverage. I believe there's a way to say "ignore these directories" which would solve this.

brettcannon:
@njsmith FYI Codecov has a YAML file you can check in to configure it if you want help tweaking settings.

njsmith:
No, codecov doesn't care about README -- the problem is that there's a branch in `_io_windows.py` that gets taken non-deterministically, so there are 3 lines that are flopping between 2 covered + 1 uncovered versus 1 covered + 2 uncovered :-(.

brettcannon:
@njsmith have you thought about using `# pragma: no cover` for those non-deterministic lines?

njsmith:
It's a legitimate complaint though :-) We should have coverage for both branches, just I got lazy about writing tests for a few bits of code that aren't actually used yet, and are just placeholders for future expansion. (Specifically the non-socket APIs for Windows and Kevent, which I figure will get sorted out properly once we add subprocess support - see also #4, #26, #52.) I have a patch here that I think makes the test coverage deterministic at least though...

PR #82, opened 2017-03-11
Use `python3 -m pip` over `pip`

brettcannon:


njsmith:
ugh what do you mean there are still people who aren't running py3 as default

Is this correct on Windows too? (the whole `py` thing confuses me.) And the README probably needs to be fixed as well.

brettcannon:
😄 I honestly promote it more for e.g. 3.5 vs 3.6.

And on Windows it depends on how they install it. Everyone gets `py`, so it can be `py -3 -m pip`, but you can choose to put `python3` on PATH.

brettcannon:
Updated for Windows and updated the README.

I should start that conversation again about getting a `py` equivalent on UNIX again.

njsmith:
Thanks!

PR #83, opened 2017-03-11
Mention iterable.__aiter__()

brettcannon:


njsmith:
Actually `__aiter__` is sync, not async (since 3.5.2, motivation being to make it work if you define `__aiter__` as an async generator function).

I think maybe it's clearer as originally written? This isn't meant to be a detailed reference on iterator internals, but just to get across the basic idea that we're swapping a sync operation for an async operation.

PR #86, opened 2017-03-11
Recognize ERROR_INVALID_HANDLE as indicating the IOCP was closed

njsmith:
There's a race condition when we signal the IOCP thread to exit by
closing the IOCP: the thread might be blocked in
GetQueuedCompletionStatusEx, or it might (rarely) be in the rest of
the loop. The effect of this is that there are two possible error
messages we might get that both indicate that the IOCP was
closed. Before we were only detecting one of them; now we detect
both.

This has the side-effect of making it so that during the test suite we
deterministically only take the regular-exit path, rather than
sometimes taking the blow-up-on-unrecognized-error path. Which is
good because codecov was getting cranky about that non-deterministic
coverage and yelling at people for no reason (see gh-81).


njsmith:
@brettcannon: hopefully this should fix the codecov thing - any interest in reviewing?

trio/_core/_io_windows.py
```diff
@@ -268,6 +268,18 @@ def do_select():
                     queue.put_nowait(info)
 
     def _iocp_thread_fn(self):
+        # This thread sits calling GetQueuedCompletionStatusEx forever. To
+        # signal that it should shut down, the main thread just closes the
+        # IOCP, which causes GetQueuedCompletionStatusEx to return with an
+        # error:
+        IOCP_CLOSED_ERRORS = {
```
brettcannon: Any reason this constant is a local and not a global?

njsmith: Eh, could go either way, but I feel like the category being defined here is more specific to the semantics of this particular function than something that's generally meaningful.

njsmith:
> I have zero IOCP experience, but LGTM. 😄

Then you're in the same boat as me :-)

njsmith:
@brettcannon: btw, want to be added as a collaborator? (Can you tell I'm eager to get more folks involved 😄)

brettcannon:
Sure, but I make no promises I will ever use such powers. 😁

njsmith:
That's fair :-). You should have an invite.

PR #93, opened 2017-03-12
Make the summary more obvious

brettcannon:
Else it isn't much of a "tl;dr" as people will overlook it.


njsmith:
Does ReST recognize `_underscored_` emphasis? It's not in [this table](http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup)... maybe it should be `*italics*`?

brettcannon:
Probably a Markdown habit of mine; swapped to using *.

PR #95, opened 2017-03-13
More warning!

brettcannon:
Any more warning and you would need the blink tag.


njsmith:
Weird how sphinx forgot to include a `:blink:` role

brettcannon:
Maybe there is a blink; I was actually joking about it. 😄

PR #96, opened 2017-03-13
Drop the need for a comma

brettcannon:


njsmith:
I think I like this one the way it is :-)

PR #102, opened 2017-03-13
Typo fix

idlesign:
Hi,

Besides a small typo there's a problem in `There are only 4 lines of code that really do anything here. On line 17, we use [...] ` paragraph. Even though in `tasks-intro.py` relevant code chunk starts from line 15, rtd.io starts it with line 16, making mentioning lines 17, 19 and 22 in the paragraph bogus.

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -794,7 +794,7 @@ And the second task's job is to process the data the server sends back:
    :pyobject: receiver
 
 It repeatedly calls ``await client_sock.recv(...)`` to get more data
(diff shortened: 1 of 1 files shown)


njsmith:
Thanks!

> Even though in tasks-intro.py relevant code chunk starts from line 15, rtd.io starts it with line 16, making mentioning lines 17, 19 and 22 in the paragraph bogus.

Ugh, right, this is a sphinx bug – I think I just fixed it by forcing rtd to use the latest sphinx dev version.

PR #104, opened 2017-03-14
Fix custom supervisor example in documentation

welchbj:
I believe this PR fixes the broken example in the Custom Supervisor example in the docs, where ``finished_task`` is an undeclared variable and ``reap_and_unwrap`` should be called on the ``nursery``.

While this is a solid example and this PR should fix the immediate error, I was wondering if you would be open to making this example a little more expressive. For me, this type of monitoring demonstrates some of the most elegant parts of the trio API. What I had in mind would be expanding the example to something like

```python
import functools
import trio


async def sleep_and_return_num(num):
    await trio.sleep(num)
    return num


async def race(*async_fns):
    if not async_fns:
        raise ValueError("must pass at least one argument")

        for async_fn in async_fns:
            nursery.spawn(async_fn)

        task_batch = await nursery.monitor.get_batch()
        nursery.cancel_scope.cancel()

        finished_task = task_batch[0]
        return nursery.reap_and_unwrap(finished_task)


async def main():
    winner = await race(
        functools.partial(sleep_and_return_num, 1),
        functools.partial(sleep_and_return_num, 2),
        functools.partial(sleep_and_return_num, 3)
    )

    print('The winner is', winner)


if __name__ == '__main__':
    trio.run(main)
```

This sample, when run, prints out ``The winner is 1``. While this may be a bit long for inlining in the main documentation, would you be open to having a collection of snippets somewhere and link to examples like this?

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -745,7 +745,8 @@ first::
                nursery.spawn(async_fn)
            task_batch = await nursery.monitor.get_batch()
            nursery.cancel_scope.cancel()
-           return task_batch[0].reap_and_unwrap(finished_task)
+           finished_task = task_batch[0]
+           return nursery.reap_and_unwrap(finished_task)
 
 This works by waiting until at least one task has finished, then
 cancelling all remaining tasks and returning the result from the first


njsmith:
Wow, no idea what I was thinking there. Thanks for the catch :-)

Regarding the larger example: I think having a collection of worked examples would be great! And I agree that we could really do more with the supervisor concept. I was tempted to write a whole article on it, even, but realized that (a) I needed to get something released, and (b) I'm not sure I know enough to write a useful article on supervisors yet – it needs some focused attention and experimentation :-).

Maybe an `examples/` directory, plus some inclusion in the docs so that they're findable? One possible model is how I did this for h11: [examples/](https://github.com/njsmith/h11/tree/master/examples), [docs](https://h11.readthedocs.io/en/latest/examples.html), [doc source](https://raw.githubusercontent.com/njsmith/h11/master/docs/source/examples.rst)

njsmith:
Also, I think for trio I'd like to try following the hyper project's model of inviting everyone whose PR is merged to join as a collaborator [1]. [Details here.](https://python-hyper.org/en/latest/one-of-the-team.html). Any interest?

[1] technically I think this model was invented by rubinius, but I like python-hyper's document :-)

welchbj:
I agree that the examples directory plus linking in the docs would be a good way to go. This is something I could look into as I use trio more and get a feel for common patterns.

I would also definitely be interested in joining as a collaborator! I am really excited about this project and hope to get more involved in its development.

njsmith:
@welchbj: cool, you should have an invitation to join the organization.

Issue #135, opened 2017-04-23
Support "auto reaping"


merrellb:
The tutorial and examples don't go into much detail on reaping best practices.  As clients connect and disconnect from long-running nurseries, should we be reaping to prevent the monitor queue and zombie frozenset from growing forever?  Perhaps an autoreap option when creating the nursery (disabling the monitor and zombie attributes) or taking an async function argument for minimal cleanup.

```
async def autoreap(nursery, reap_fn=None):
    while True:
        batch = await nursery.monitor.get_batch()
        for task in batch:
            if reap_fn:
                await reap_fn(task)
            nursery.reap(task)
```

njsmith:
Ah, I think I see the problem... The docs are not as clear as they could be about an important issue :-)

The intended way to use nurseries (unless you're actually implementing a custom supervisor) is that you spawn some tasks and then sit parked in the nursery `__aexit__`. The task that holds the nursery shouldn't be doing other work itself; it should all be delegated to children. ("Parenting is a full time job.")

I think this needs to be emphasized more in the docs, and in particular the tutorial... I'm seeing now that it's really not obvious (esp. since the call to `__aexit__` is literally invisible!). But even though from the user point of view it looks like the parent task isn't doing anything, in fact it's performing the critical work of watching for child tasks to exit, reaping them, and responding to any unexpected exceptions that bubble up from the children. (I think this is also why you were having trouble with `KeyboardExit`not working.)

Basically the nursery `__aexit__` **is** your autoreap function and more :-). (In particular, the snippet you show silently discards any unexpected exceptions; if you look at the code that implements nurseries then like 90% of the complications are from dealing with exceptions. Exceptions are super annoying.)

Does that make sense?

merrellb:
Yeah, the invisible ```__aexit__``` is both quite elegant and invisible enough to be potentially confusing :-) .  I sometimes find myself wishing for something a bit more explicit like requiring an ```await nursery.done()``` at the end of a nursery scope (which would also give something to wrap in a try/except block for KeyboardInterrupt).

In this case, I am thinking about long running servers where a nursery could be going for days or months.  Each connection will have multiple tasks starting and ending which will need to be managed within the nursery scope.   Some simple tests I've tried suggests that ```__aexit__``` doesn't perform any reaping until the nursery scope actually ends.  With a long running server, how do we prevent zombies from growing indefinitely without some explicit reaping task?

PR #139, opened 2017-04-24
[doc] Emphasize that parenting is a full-time job in the tutorial

njsmith:
Some early user feedback suggests that this is possible to
miss (gh-135), and it's important and kinda subtle (it involves an
invisible function call!). So let's make sure to point it out
explicitly the very first time we see a call to open_nursery, and
reinforce it as we go.


njsmith:
cc @merrellb

PR #145, opened 2017-05-01
Implement happy eyeballs

njsmith:
Not tested, not integrated, just want to get it off my laptop.


njsmith:
This is now working, and has comprehensive tests!

The docs are a mess; I just shoved the new functions in somewhere that isn't *too* terrible, but really that whole section needs to be reorganized soon. Maybe better to wait until we have the server stream APIs working though? I dunno.

PR #165, opened 2017-05-19
Correctly preserve exception __context__ in MultiError.catch

njsmith:
Python's implicit exception chaining logic insists on corrupting
__context__ when we re-raise an unpacked exception in
MultiError.catch. This commit introduces a counter-measure.


njsmith:
That's weird, appveyor seems to have just... not noticed this PR at all. It hasn't even attempted to build it. Maybe closing and opening again will make it notice?

Issue #140, opened 2017-04-24
coverage flapping in acontextmanager


njsmith:
Kind of annoying, e.g. this is a doc-only PR: https://codecov.io/gh/python-trio/trio/pull/139/changes

The solution is probably to write proper tests for acontextmanager.

imrn:
Are you going to maintain acontextmanager and friends in your
project or are there any plans to have in python stdlib?

njsmith:
In the long run I expect there'll be something like this in the stdlib, but not before 3.7, and it'll be a while before we can drop support for 3.5 and 3.6. What might make sense until then is to move `acontextmanager` into some other library, like maybe my `async_generator`. But another consideration is that trio might actually need to keep this so that it can have full control over the keyboard interrupt handling (see my blog post about control-C).

In any case I should probably bite the bullet and write some tests :-)

njsmith:
@kyrias: oh, if you're looking for a very straightforward place to get your feet wet, this is also an option :-). Basically the issue is just that we need some tests written for `trio.util.acontextmanager` to systematically check all the different paths in `__aexit__`, both because this is a good idea in general and because it's causing annoying issues where codecov keeps complaining about random lines here becoming covered/uncovered. (Apparently there's something in the test suite that randomly covers a few lines or not, so codecov keeps thing random unrelated changes are causing coverage regressions.)

(And note for anyone else stumbling across this later, if it's no longer mid-June 2017 and the issue is still open then you can probably assume that @kyrias didn't take this up and you're welcome to it!)

njsmith:
Actually, I'm baffled about this. Here it is happening again now: https://codecov.io/gh/python-trio/trio/pull/269/changes
But that code clearly has a test that should cover it in our current testsuite: https://github.com/python-trio/trio/blob/8229f06728831bb682b9df8c3f6fa6a71cd2e955/trio/tests/test_util.py#L95-L97

njsmith:
(whoops, fat-fingered the close button I think)

Oh, I just had a silly misunderstanding -- to cover those lines we need a test where the context manager allows a `StopAsyncIteration` to *pass through*, which we didn't have -- the test I pointed at raises `StopAsyncIteration` but then the context manager catches it. #270 has a test like this.

njsmith:
This just came back: https://codecov.io/gh/python-trio/trio/pull/292/changes

WTF

njsmith:
[...]

PR #166, opened 2017-05-19
socket: ignore OSError during IPV6_V6ONLY-normalization

### LINKED ISSUE #164: Invalid argument exception during socket.accept on AF_INET6
minimal-ish testcase and exception: https://gist.github.com/8d647f4c8368bc318e73833474601526

If I remove the `setsockopt` call on line 192, the testcase works as expected.

Is this a bug, or am I not creating an IPv6 socket correctly?

buhman:
~This doesn't appear to work with all types of AF_INET6 sockets--it is likely better to have the caller handle this instead.~ I've now changed this to ignore OSError instead based on your feedback in #164.

Fixes #164.

### DIFF
--- a/trio/socket.py
+++ b/trio/socket.py
@@ -188,7 +188,10 @@ class SocketType:
 
         # Defaults:
         if self._sock.family == AF_INET6:
-            self.setsockopt(IPPROTO_IPV6, IPV6_V6ONLY, False)
+            try:
+                self.setsockopt(IPPROTO_IPV6, IPV6_V6ONLY, False)
+            except OSError:
+                pass
 
         # See https://github.com/python-trio/trio/issues/39
         if _sys.platform == "win32":
--- a/trio/tests/test_socket.py
+++ b/trio/tests/test_socket.py
@@ -316,12 +316,13 @@ async def test_SocketType_shutdown():
         assert await a.recv(3) == b"yyy"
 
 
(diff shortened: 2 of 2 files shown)


njsmith:
Thanks for the patch! Codecov failure is annoyingly caused by #140, nothing to do with you. And having pondered https://github.com/python-trio/trio/issues/164#issuecomment-302622899 and slept on it, I think I'm going to go ahead and merge this, and worry about fixing it up again at the next level up :-)

njsmith:
Oh wait, of course I had the tab open from yesterday and as soon as I hit post it updated to show I was replying to the old different version.

Oh well, this one is fine too :-). The immediate goal is to get this unbroken, and the details can be sorted out when the higher-level code arrives.

njsmith:
Oh, right, now the codecov failure is real though – there's no test where the `setsockopt` fails. Is that something you can easily add? (Probably a good idea no matter which approach we take...)

buhman:
> there's no test where the setsockopt fails. Is that something you can easily add? 

Sure thing.

buhman:
@njsmith :+1:

njsmith:
Awesome, thanks!

njsmith:
Want a commit bit? You can join in as much or little as you want, but my policy is to offer one to everyone who contributes a patch :-)

Issue #20, opened 2017-01-22
async file I/O


njsmith:
Maybe a wrapper around `pathlib.Path` that wraps everything in `run_in_worker_thread`, and where `open` returns a file-like object whose methods have again been wrapped in `run_in_worker_thread`? (#10 and #6 are relevant.)

On Windows it'd be nice to go the extra step and use the IOCP methods for actual file I/O.

buhman:
Because I've used and am very familiar with aiofiles, I thought I'd work on this.  Instead of duplicating aiofiles though, I thought I'd try to be fancy and wrap methods instead of each file object. I came up with [this monstrosity](https://gist.github.com/buhman/94214ed5d3bc9efaffda87009f12e862), but it doesn't work because:

 * I include wrappers for non-heap classes, and the `__class__` hackery on line 117 won't work for those.
 * some functions, like `close()` will call `flush()` internally, which causes problems if `flush()` is a coroutine function

I think I learned enough from this experiment that I can come up with something even better and actually-working.

buhman:
Should this be a new `io` package within trio, or completely separated code inside `trio_io`? (maybe the latter would convince me to help with #158).

njsmith:
Hey, sorry for the slow response! I'm still getting caught up again after PyCon.

In general, my preference is to avoid magic introspection, dispatch, subclassing, and similar things... I like delegation and explicit lists :-). In theory, it's extra typing and they can get out-of-sync with Python, but in practice I think it's worth it because even if it does go wrong sometimes, it's much easier to fix than complicated magic. Example: in trio's socket wrapper, here are the attributes that [get delegated directly](https://github.com/python-trio/trio/blob/5d046ba7613c2a3063c3fc93fc4b9a9d94cdd8b0/trio/socket.py#L235-L246) and [here's one that gets wrapped](https://github.com/python-trio/trio/blob/5d046ba7613c2a3063c3fc93fc4b9a9d94cdd8b0/trio/socket.py#L527-L528).

I guess in this case the problem is that the `io` module is a messy tangle of subclasses? But even so, it seems like it should be pretty straightforward to keep a list of which classes we want to wrap (basically `Path`, `TextIO`, `BufferedReader`, `BufferedWriter`, `BufferedRandom`?), and which methods are sync and which need to be wrapped in `run_in_worker_thread`. What do you think?

[...]

Issue #105, opened 2017-03-15
Best approach to building trio ecosystem


merrellb:
My async experience has largely been with the aio-lib ecosystem  (particularly http, websockets, elasticsearch and kafka).  I would love to have this functionality with trio and would happily contribute to help make it happen.  A few thoughts/questions:
1) If I wanted to for example add http functionality (which I believe trio currently lacks) would it be best to focus on porting aiohttp or perhaps focus on h11 or hyper-h2 (which already seem to have some basic curio integration).
2) Are there any useful patterns to port synchronous libraries (eg the official Elasticsearch library) to trio?
3) Would it be worthwhile to add subprojects under python-trio for various popular APIs (similar to aio-lib)?

njsmith:
Hi Brian, great to hear from you!

There's definitely lots to work on. Regarding http: the main immediate prerequisite inside trio itself is TLS support (#9), which is very doable, though we need to sort out what our idea of a "stream" looks like (part of #73).

We also have some todo items to split out our pytest and sphinx infrastructure into reusable packages (#27, #71), which should be relatively straightforward and is something that any downstream projects are going to want.

Regarding aiohttp vs. h11/h2: I may be biased, but I would definitely try to build something on top of h11 rather than trying to do a direct port of aiohttp. My impression is that aiohttp's internals are... unnecessarily complex, and would need significant reworking to match the structure that trio requires. There definitely is a fair amount of stuff that a full fledged http client needs (I assume you're thinking mostly about clients?) that isn't in h11, though.

Another possibility is to look at urllib3; @Lukasa is working on porting it to use h11: https://github.com/shazow/urllib3/pull/1067
In the long run the ideal situation would probably be for urllib3 itself to gain the ability to run in "trio mode", because there is a ton of complicated stuff that a really high-quality http client has to do and there's less wasted work across the ecosystem if we can all share one implementation of most of it. It's possible that this will take long enough to become viable that we want something else as a stop-gap in the mean time; that could be a minimalist library written from scratch, or a hacky fork of @Lukasa's branch with a bunch of extra `await`s sprinkled about... I dunno. Just throwing it out there as something to think about.

[...]

PR #180, opened 2017-05-29
add asynchronous file io and path wrappers

buhman:
This is an attempt to implement #20.

Todo:

`_file_io`:
* [x] make duck-file definition more restrictive https://github.com/python-trio/trio/pull/180#discussion_r121320844

`_path`:
* [x] improve test_path_wraps_path https://github.com/python-trio/trio/pull/180#discussion_r121321347
* [x] add tests for non-rewrapping forwards https://github.com/python-trio/trio/pull/180#discussion_r121323894
* [x] add `__div__` 
* [x] py35 https://github.com/python-trio/trio/pull/180#discussion_r121320712
* [x] ~improve Path docstring~ I was going to write some glossary entry on `asynchronous path object`, then I realized there is no upstream concept of `path object`, only `PathLike`
* [x] properties can return new `pathlib.Path` instances too https://github.com/python-trio/trio/pull/180#discussion_r121582161

---

Currently supported usages:

```python
# files

async with await trio.open_file(filename) as f:
    await f.read()

async_string = trio.wrap_file(StringIO('test'))
assert await async_string.read() == 'test'

# paths

path = trio.Path('foo')
await path.resolve()
```


trio/io/types.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        line = await self.readline()
+        if line:
+            return line
+        else:
+            return StopAsyncIteration
```
buhman: This is copy-paste from aiofiles, but it might be more consistent to wrap `__next__` and await that instead.

trio/io/io.py
```diff
@@ -0,0 +1,97 @@
...
+
+        assert f.closed
+
+    """
+    _file = wrap(await trio.run_in_worker_thread(io.open, file, mode,
+                                                 buffering, encoding, errors, newline, closefd, opener))
+    return _file
+
+
+@singledispatch
```
njsmith: I'd prefer an explicit `if` chain to the use of `singledispatch`, exactly because allowing random modules to change the behavior of a trio function is a great example of spooky action at a distance.

I get that you want to support `pyfakefs.fake_filesystem.FakeFileWrapper`, but if we're OK with mutating global state like `singledispatch` does then surely we can accomplish this just as easily by doing `io.RawIOBase.register(pyfakefs.fake_filesystem.FakeFileWrapper)` or whatever (and if `pyfakefs` doesn't already do this then that's surely a bug on their end?).

...actually looking more closely, I guess `FakeFileWrapper` actually changes which interface it implements depending on the specific object it's wrapping, which is a bit of a mess but also not one that `@singledispatch` can help with.

njsmith:
Hey, sorry for being slow to get to this. Even though you can't see it here, I've been thinking about it (and the annoying problem of how to best wrap that IO class hierarchy) all week...

# High level comments

I'm not a huge fan of the metaclass thing, but there aren't any really *nice* solutions here, and it's local and well-contained so meh, we can switch it out for something else later if we want. (I suspect in the long run we may want to define some delegation helpers that work as class decorators and inject methods at class definition time. But there's no reason to try and solve that design problem here; async file IO already raises enough tricky design problems :-)

Regarding the overall public API: So far the rule has been that if there's a standard library module `X` and a module `trio.X`, then the trio version's public API is a mirror of the stdlib version. That's not the case here. Nor should it be: it doesn't make sense for trio to be re-exporting all the stuff in `io`. (Also `trio.io` would be a super confusing name given that like, everything in trio is about io in one way or another. I don't know why Python's designers thought that "io" was a good short name for "file io". But this is just me being grumpy.) So, my suggestion is that we move the stuff in your `io.py` and `io/types.py` to a `_file_io.py` file, and have it define `open_file`, `wrap_file`, and `Path` objects to re-export from the top-level `trio` namespace.

I have mixed feelings about whether to make the `AsyncIO*` classes public, except via `wrap_file`; it's just asking for someone to subclass them or something. We definitely shouldn't make `AsyncIOBase` public, because as a concrete class it's useless – it doesn't even have `read` or `write`! I guess for the other ones I'll probably come around eventually...

Regarding the "`async with await open(...)` nonsense": I agree that it's nonsense, but it's nonsense that accurately reflects the standard Python semantics, so I actually think we should keep it that way :-). And the trio style is quite strict about enforcing the idea that async functions are just functions that you call with `await` – I don't want to have functions that support both `await foo(...)` and `async with foo(..)` as calling conventions. Obviously the way coroutines work under the hood makes it *possible* to get up to all sorts of shenanigans with them, but that's not unusual... generally in Python you *can* do *anything*, but most of it's in bad taste, so we don't :-).

Also, it should be totally legal to write code like:
```python
file_obj = await open(...)
async with file_obj:
    ...
```
so we need `__aenter__` and `__aexit__` methods on `AsyncIOBase` in any case.

Regarding custom handling for `Path.open`: I think this is no big deal? We can use the generated wrappers for the easy methods, and for the methods that need special handling we can write the wrapper methods manually. There's no rule that you can't use both techniques within a single class definition :-)

Speaking of which, `AsyncIOBase.close` also need to be written manually, because `close` methods have special semantics: they guarantee that even if they're cancelled, they still close the object before returning. This is a bit awkward given the way we're building on top of `run_in_worker_thread`, so I think the best solution is to write it like:
```python
    async def close(self):
        # run_in_worker_thread will check for cancellation; make sure it can't be cancelled
        with _core.open_cancel_scope(shield=True):
            await run_in_worker_thread(self._wrapped.close)
        # but if there was a cancellation, we *should* raise `Cancelled`, now that the file is safely closed
        await _core.yield_if_cancelled()
```
As an optimization it might also make sense to check `self._wrapped.closed` before spawning a thread to call `close`. This whole async `close` thing is also pretty awkward given that e.g. socket objects in trio have *synchronous* `close` methods, but I don't know that there's much to be done about it. I guess we could call the file `close` methods `aclose`, like async generators have.

Regarding the threadpool thing: not sure what you're thinking of. Probably just some issue where I was mumbling incomprehensibly to myself :-). `run_in_worker_thread` is the right thing to use here.

njsmith:
Hmm, looking again at [the docs](https://docs.python.org/3/library/io.html), they actually say that "Even though IOBase does not declare read(), readinto(), or write() because their signatures will vary, implementations and clients should consider those methods part of the interface". So maybe we should move those onto `AsyncIOBase`, and then `wrap_file` can use that as the fallback for wrapping unrecognized but allegedly file-like objects.

njsmith:
...and, more annoyingly, basically *all* the concrete classes in `io` add random extra attributes. `FileIO` objects have `mode` and `name`. (And empirically, `BufferedReader` and `TextIOWrapper` objects do too when wrapping a `FileIO`, even though this isn't documented.) `BytesIO` has `getbuffer` and `getvalue`; `StringIO` has `getvalue`. `BufferedReader` adds `peek`. `TextIOWrapper` adds `line_buffering`.

I *really* don't want to define our own file IO APIs but man these ones are a mess :-(.

imrn:
On 6/4/17, Nathaniel J. Smith <notifications@github.com> wrote:
> Regarding the "`async with await open(...)` nonsense": I agree that it's nonsense.

I don't think it is nonsense.
- Opening the file is an async operation.
- Entering its context is an other async thing.

So what is the problem? (Other than the possibility forgetting "await".)
User should be aware of the chain of operations and their conceptual
meaning.

Providing a shortcut like "async with open(...)" will cause
moving the async lines in open(...) to __aenter__(). This
will distort ortogonality of both methods as well as their possible
interactions with others.

njsmith:
@imrn
> I don't think it is nonsense. Opening the file is an async operation. Entering its context is an other async thing.

Yes, we're agreeing. It's nonsense in the sense that it's an awkward thing to deal with and "looks wrong". (Before this I would have said that `async with await` was always a bug because the blocking operation could be folded into the `__aenter__`, and I think it might even be useful to have a version of `open` that only works with `async with`). But it's not nonsense because, well, what you said.

In fact it might might sense to support something like
```python
async with path_obj.mode("rb") as file_obj:
    ...
```
Or maybe that's too much a violation of There's Only One Way To Do It, not sure.

trio/io/types.py
```diff
@@ -0,0 +1,72 @@
...
+def _method_factory(cls, meth_name):
+    async def wrapper(self, *args, **kwargs):
+        meth = getattr(self._file, meth_name)
+        func = partial(meth, *args, **kwargs)
+        return await trio.run_in_worker_thread(func)
+
+    wrapper.__name__ = meth_name
+    wrapper.__qualname__ = '.'.join((__name__,
+                                     cls.__name__,
+                                     meth_name))
```
buhman: I don't think so; `__name__` is `trio.io.types` (the module name), so it becomes something like `trio.io.types.*IOBase.close`.

buhman:
> I really don't want to define our own file IO APIs but man these ones are a mess :-(.

Do you think it would be useful to add more wrappers for the non-base io classes?

buhman:
> Even though IOBase does not declare read(), readinto(), or write() because their signatures will vary, implementations and clients should consider those methods part of the interface

TextIOBase doesn't define `readinto`.

Do you think it makes sense to do this anyway?

njsmith:
> Do you think it would be useful to add more wrappers for the non-base io classes?

I'd really rather not. Probably the simplest option would be to make the attribute that holds the wrapped object public, so in the rare cases where people need to access quirky attributes they can do `async_file_obj.wrapped.mode` or whatever?

> TextIOBase doesn't define `readinto`. Do you think it makes sense to do this anyway?

Ugh, good point. It might, since `readinto` is commonly available on random user-defined "file like" objects.

Well, here's an idea to consider: we could write the wrapper classes so that they only expose attributes that actually exist on the wrapped object. In Python, the only ways to tell whether a class has an attribute are (a) whether it shows up in `__dir__`, and (b) whether trying to access it raises `AttributeError`. (E.g. `hasattr` internally just tries to access the attribute, and returns `False` if it gets an `AttributeError`.) And notice that for the synchronous attributes, this PR's current `__getattr__`-based implementation will already raise `AttributeError` if the corresponding attribute is missing on the underlying object, so we're already half-way there. If we also used `__getattr__` for the async methods (i.e. generating the wrapper on each call to `__getattr__` instead of doing it at class initialization), and added a line like `attrs = [a for a in attrs if hasattr(self, a)]` to our `__dir__` method, then I think that would do it.

If going this way, it's even tempting to just have a single wrapper class for all "vaguely file-like objects" in Python, since as we've learned, the "file-like" interface is pretty vague to start with. That would also solve the problem with `pyfakefs`.

njsmith:
Oh, another point I just realized: `detach` returns a file-like, so our version needs a special case to wrap the return value, like:
```python
async def detach(self):
    return wrap(run_in_worker_thread(self._wrapped.detach))
```

buhman:
> i.e. generating the wrapper on each call to __getattr__ instead of doing it at class initialization

Doesn't this mean each call to `__getattr__` also returns a different instance of that method? Calling the method factory for each getattr seems a bit wasteful.

buhman:
> so our version needs a special case to wrap the return value, like:

I noticed this while I was playing with `AsyncPath`, which has things like `resolve` and `with_name`. Instead of special-casing, I was a bit more heavy-handed: https://github.com/python-trio/trio/pull/180/commits/24dc536ccb5de3ae4136d335a03cc649effbdaa4#diff-c4df1d85a0d10c3c2ebc8ae9f39dd617R31

njsmith:
> Doesn't this mean each call to `__getattr__` also returns a different instance of that method? Calling the method factory for each getattr seems a bit wasteful.

I'm not really worried about the overhead, given that each method call is also creating a thread :-). I guess it's a bit weird maybe if `foo.meth is foo.meth` returns `False`. If it matters then the simple solution is to have `__getattr__` save the computed attribute the first time it's called, so the next time it gets pulled from the instance dict instead of calling `__getattr__`. (Maybe that's phrased confusingly. I'm talking about having `__getattr__` do `value = ...; setattr(self, attr, value); return value`.) An LRU cache would be overkill here I think.

> I also thought it was a bit too tedious to sort out the methods manually, so after reading pathlib#L606-L607, I decided to automatically compute `_forward` and `_wrap` based on whether this method is defined in `PurePath` or `Path`.

Hmm. This makes me twitch a bit because it's [extremely close to the sororicide pattern](https://glyph.twistedmatrix.com/2017/05/the-sororicide-antipattern.html) but I think maybe in this specific case it's OK?

From a glance at your work-in-progress though it seems like you might be overcomplicating it a bit though – we only need to wrap `pathlib.Path`, nothing else. (I don't know why python documents `WindowsPath` and `PosixPath` as always being available; in any given install only one of them is usable, and it's the one aliased to `Path`.) And we can compute the `_forwards` and `_wraps` sets like:

```python
_forwards = [attr for attr in dir(pathlib.PurePath) if not attr.startswith("_")]
_wraps = [attr for attr in dir(pathlib.Path) if not attrs.startswith("_") and attr not in _forwards]
```

and then the actual forwarding code can be identical to the code for files, except that we need the `if isinstance(result, pathlib.Path): result = Path(result)` check.

njsmith:
And btw, thanks for your patience here – I'm picky about these things to start with, and then this is an extraordinarily tricky API design challenge on top of that. But I think we're getting there!

buhman:
> thanks for your patience here

I'm just here to learn; I'm in no rush!

buhman:
I thought about this all day, and came up with 93ccad2, which seems to make the most sense to me—if a wrapped file object doesn't support a particular method/property, `AsyncIO` will just end up throwing AttributeError, which is the same thing that would happen if you tried to do that on the underlying file object anyway.

If that looks ok, I'll sort out the other forwarded or wrapped attributes accordingly, and clean up the docs.

njsmith:
It seems like this is maybe more complicated than it needs to be? Like I haven't tested this, but I think it has everything we need for the file-like wrapper, and I can almost fit the whole class on my screen at once:

```python
# This class is "clever". These two globals define the attributes that it
# *can* delegate to the wrapped object, but any particular instance only
# *does* delegate the attributes that are actually found on the object its
# wrapping. So e.g. if you wrap an object that has no "readinto", then
# hasattr(wrapper, "readinto") returns False, tab completion doesn't offer
# readinto, etc.
_FILE_LIKE_SYNC_ATTRS = {...}
_FILE_LIKE_ASYNC_METHODS = {...}
class AsyncFileLikeWrapper:
    def __init__(self, wrapped):
        self.wrapped = wrapped

    def __getattr__(self, name):
        if name in _FILE_LIKE_SYNC_ATTRS:
            # May raise AttributeError which we let escape
            return getattr(self.wrapped, name)
        elif name in _FILE_LIKE_ASYNC_METHODS:
            # May raise AttributeError which we let escape
            meth = getattr(self.wrapped, name)
            async def async_wrapper(*args, **kwargs):
                return await run_in_worker_thread(partial(meth, *args, **kwargs))
            async_wrapper.__name__ = name
            async_wrapper.__qualname__ = self.__class__.__qualname__ + "." + name)
            # cache the generated method so that self.foo is self.foo == True
            setattr(self, name, async_wrapper)
            return async_wrapper
        else:
            raise AttributeError(name)

    def __dir__(self):
        attrs = set(dir(self))
        attrs.update(a for a in _FILE_LIKE_SYNC_ATTRS if hasattr(self, a))
        attrs.update(a for a in _FILE_LIKE_ASYNC_METHODS if hasattr(self, a))
        return attrs

    def __repr__(self):
        return "trio.wrap_file({!r})".format(self.wrapped)

    async def detach(self):
        ret = await run_in_worker_thread(self.wrapped.detach)
        return wrap_file(ret)

    async def close(self):
        with _core.open_cancel_scope(shield=True):
            await run_in_worker_thread(self.wrapped.close)
        # run_in_worker_thread did a schedule point, but we carefully
        # prevented it from doing a cancel point, so do that now.
        await _core.yield_if_cancelled()

    async def __aenter__(self):
        return self

    async def __aexit__(self, *_):
        await self.close()

    @_util.aiter_compat
    def __aiter__(self):
        return self

    async def __anext__(self):
        try:
            return await run_in_worker_thread(next, self.wrapped)
        except StopIteration:
            raise AsyncStopIteration

def wrap_file(file_like):
    if not hasattr(file_like, "read") and not hasattr(file_like, "write"):
        raise TypeError(
            "{!r} object doesn't quack like a file"
            .format(file_like.__class__))
    return AsyncFileLikeWrapper(file_like)

async def open_file(*args, **kwargs):
    do_open = partial(open, *args, **kwargs)
    return wrap_file(await run_in_worker_thread(do_open))
```

The `Path` wrapper is a little different, because it doesn't need the cleverness where the available attributes change each time, and it needs the special handling for methods that return `Path`s. I'm not sure it's worth trying to share code between them, because it's really only a few lines in common and they end up needing tweaking anyway because of these differences.

buhman:
> maybe more complicated than it needs to be

Probably.

```python
attrs.update(a for a in _FILE_LIKE_SYNC_ATTRS if hasattr(self, a))
attrs.update(a for a in _FILE_LIKE_ASYNC_METHODS if hasattr(self, a))
```

I thought about doing this too before I wrote 93ccad2; if we call hasattr for every supported attribute, I think we might as well just do it once during `__init__`:

```python
def __init__(self, wrapped):
    …
    self._available_sync_methods = [attr for attr in _FILE_LIKE_SYNC_ATTRS if hasattr(self.wrapped, attr)]
    self._available_async_methods = [attr for attr in _FILE_LIKE_ASYNC_METHODS if hasattr(self.wrapped, attr)]
```

I'll back up a few commits and try it this way.

buhman:
Also, is `Closing(Async)ContextManager` undesirable?

njsmith:
> if we call hasattr for every supported attribute, I think we might as well just do it once during `__init__`

Does freezing/caching the attribute lists like this simplify the code, though? `__dir__` is extremely *not* performance sensitive, and usually more state = bad.

> Also, is `Closing(Async)ContextManager` undesirable (making `async with await` the only supported context manager convention)?

Yeah, this is what I was trying to say somewhere up above... the `await` should be mandatory.

Rationale: I don't think `async with await open_file(...)` is the prettiest, but I like it better than having the same function support being called like `await open_file(...)` and like `async with open_file(...)` at the same time. Trio functions should pick one of those and stick to it; anything else is too magic and breaks the simplifying illusion that the difference between async functions and regular functions is just that async functions require `await` when called (see e.g. the tutorial). And we kind of have to support `await open_file(...)`. So `async with open_file(...)` loses. "Special cases aren't special enough to break the rules", as `import this` says. Plus since we have to support both `await open_file(...)` and `async with file_obj`, the `async with await open_file(...)` case falls out as supported automatically, whereas the `async with open_file(...)` case would be an extra thing people would have to memorize ("there should by one – and preferably only one – obvious way to do it").

If it makes you feel better, people will also be writing things like `async with await open_tcp_stream(hostname, port)` as soon as I finish implementing the high level stream code :-)

buhman:
> Does freezing/caching the attribute lists like this simplify the code, though?

Not really, and `__init__` is probably the more performance-sensitive thing anyway; I'll swap this around.

buhman:
@njsmith I think this is ready to merge!

buhman:
> Would it be worthwhile to add subprojects under python-trio for various popular APIs

Not to suggest refactoring this all over again, but I've been thinking about #105 and #171. I think, at least the `_file_io` stuff is actually pretty generically usable, even for asyncio.

What do you think about ripping all of this out, then making a `python-trio/async-file-io` or similar that supports creating both asyncio and trio threads?

Akuli:
I would expect `type(asyncfile).method(asyncfile)` to do the same thing as `asyncfile.method()`. It doesn't work if the methods are generated in the `__getattr__` of `type(asyncfile)`.

It's OK if `instance.method is instance.method` is False, but at least `instace.method == instance.method` should be True IMO. This is actually how normal methods work:

```python
>>> class Thing:
...   def method(self): pass
... 
>>> t = Thing()
>>> t.method is t.method
False
>>> t.method == t.method
True
>>> 
```

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
...
+    assert not any(attr in dir(async_file) for attr in attrs if attr not in dir(wrapped))
+
+
+def test_unsupported_not_forwarded(async_file):
+    async_file._wrapped = sentinel
+
+    assert hasattr(async_file.wrapped, 'unsupported_attr')
+
+    with pytest.raises(AttributeError):
+        getattr(async_file, 'unsupported_attr')
```
njsmith: I don't understand how this test works either. What's the advantage of using a 1-line fixture for `wrap_file(mock_object)` if you're then going to spend a line violating the abstraction boundary to convert that into `wrap_file(sentinel)`? And what magic properties are we relying on here -- `sentinel` objects have all attributes? I feel like this would be clearer if written like:

```python
async def test_unsupported_not_forwarded():
    class FakeFile:
        def write(self, data):
            pass
        def some_other_method(self):
           pass
    wrapper = wrap_file(FakeFile())
    assert hasattr(wrapper.wrapped, "write")
    assert hasattr(wrapper.wrapped, "some_other_method")
    assert hasattr(wrapper, "write")
    assert not hasattr(wrapper, "some_other_method")
    assert not hasattr(wrapper, "read")
```
...actually I guess I just rewrote part of the previous test in there too :-)

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
...
+async def test_async_methods_match_wrapper(async_file, wrapped):
+    for meth_name in _FILE_ASYNC_METHODS:
+        if meth_name in dir(async_file):
+            continue
+
+        with pytest.raises(AttributeError):
+            getattr(async_file, meth_name)
+
+        with pytest.raises(AttributeError):
+            getattr(wrapped, meth_name)
```
njsmith: Isn't this a different way of writing that test above with the `any` and the `all`?

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
...
+async def test_open_context_manager(path):
+    async with await trio.open_file(path, 'w') as f:
+        assert isinstance(f, trio.AsyncIOWrapper)
+        assert not f.closed
+
+    assert f.closed
+
+
+async def test_async_iter(async_file):
+    async_file._wrapped = io.StringIO('test\nfoo\nbar')
```
njsmith: Again, this would be clearer and more future-proof if you dropped the fixture and wrote `async_file = wrap_file(io.StringIO(...))'

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
...
+    expected = iter(list(async_file.wrapped))
+    async_file.wrapped.seek(0)
+
+    async for actual in async_file:
+        assert actual == next(expected)
+
+    with pytest.raises(StopIteration):
+        next(expected)
```
njsmith: A less error-prone way to write this might be:
```python
got = []
async for line in async_file:
    got.append(line)
assert got == expected
```

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
...
+
+async def test_close_cancelled(path):
+    with _core.open_cancel_scope() as cscope:
+        f = await trio.open_file(path, 'w')
+        cscope.cancel()
+
+        with pytest.raises(_core.Cancelled):
+            await f.write('a')
+
+        await f.close()
```
njsmith: This line should also raise `Cancelled`, so we might as well add a `with pytest.raises(...)` to assert that.

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
...
+async def test_detach_rewraps_asynciobase():
+    raw = io.BytesIO()
+    buffered = io.BufferedReader(raw)
+
+    async_file = trio.wrap_file(buffered)
+
+    detached = await async_file.detach()
+
+    assert isinstance(detached, trio.AsyncIOWrapper)
+    assert detached.wrapped == raw
```
njsmith: Use `is`, not `==` for testing object identity.

njsmith:
@Akuli:
> I would expect type(asyncfile).method(asyncfile) to do the same thing as asyncfile.method(). It doesn't work if the methods are generated in the `__getattr__` of type(asyncfile). (Edit: curio just defines the methods one by one.)

The curio way also makes sense to me. I can certainly see arguments either way. The trade-off is that with the way in this PR, tab-completion works correctly (only offers options that actually make sense for the given object) but `type(asyncfile).method(asyncfile)` breaks (why would you ever do this?) and it's a bit more complicated to set up. The curio approach is the other way around. I think people use tab completion more often then they try to pull class functions off of an object, but it's a pretty small-scale problem in either case.

@buhman 
> What do you think about ripping all of this out, then making a python-trio/async-file-io or similar that supports creating both asyncio and trio threads?

Personally, I would prefer not to. Fundamentally for me trio is an experiment, and the question it's trying to answer is: "can we make an API that's so much nicer than asyncio-and-friends that it can overcome the massive gap in maturity/ecosystem/mindshare?" I'm OK with the answer being "no", but I'm not OK with answer being "yes we could have made such an API, but we failed to do so because we got distracted with other things and now no-one will ever know" :-/. So I want to try hard to avoid distractions that might compromise trio's progress, like writing a new file io library for asyncio and then having to support both versions. Plus I think there might be unexpected complexities here – in particular cancellation works pretty differently between them, so `close` would need special casing, and potentially some of the side-effect-free methods on `Path` could be improved by passing `cancellable=True` to `run_in_worker_thread`, and after #181 gets fixed I can imagine we might want to allow users to specify the `limiter=` argument on a per-file basis, etc. I'm dubious about the value of supporting OS-native AIO primitives, but if someone wants to experiment with that it's much easier if they don't have to worry about asyncio in the process. And so forth. It's generally much easier to make these kinds of changes if everything lives inside the same tree and only has to worry about trio's peculiarities. Of course, if you want to make such a library, you totally should :-). I just suspect it'll end up being easier to make it two libraries instead of trying to wedge them both together into a single library under the `python-trio` org.

Akuli:
Maybe the `__getattr__` way isn't too bad after all. The `converters` argument of `configparser.ConfigParser` doesn't support `type(thing).method(thing)` either, it just adds a bunch of partials to the parser and its sections:

```python
>>> p = configparser.ConfigParser(converters={'wololo': int})
>>> p.read_dict({'a': {'b': '123'}})
>>> p.getwololo('a', 'b')
123
>>> p['a'].getwololo('b')
123
>>> p.getwololo
functools.partial(<bound method RawConfigParser._get_conv of <configparser.ConfigParser object at 0x7f37f424c860>>, conv=<class 'int'>)
>>> p['a'].getwololo
functools.partial(<bound method SectionProxy.get of <Section: a>>, _impl=functools.partial(<bound method RawConfigParser._get_conv of <configparser.ConfigParser object at 0x7f37f424c860>>, conv=<class 'int'>))
>>> 
```

trio/_file_io/_path.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def generate_magic(cls, attrs):
+        # generate wrappers for magic
+        for attr_name in cls._forward_magic:
+            attr = getattr(cls._forwards, attr_name)
+            wrapper = _forward_factory(cls, attr_name, attr)
+            setattr(cls, attr_name, wrapper)
+
+
+class AsyncPath(metaclass=AsyncAutoWrapperType):
```
Akuli: Doesn't this just blindly add a thread wrapper around all methods? I don't think it makes sense to run things like `with_name` in threads.

trio/_file_io/_file_io.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Example::
+
+        async_file = trio.wrap_file(StringIO('asdf'))
+
+        assert await async_file.read() == 'asdf'
+
+    """
+
+    if isinstance(file, io.IOBase):
```
buhman: I think all that a file really needs is `close`, because that's the only assumption `AsyncIOWrapper` makes.

trio/_file_io/_path.py
```diff
@@ -0,0 +1,138 @@
...
+    """:class:`trio.AsyncPath` is a :class:`~pathlib.Path` wrapper executes concrete
+    non-computational Path methods in :meth:`trio.run_in_worker_thread`.
+
+    """
+
+    _wraps = Path
+    _forwards = PurePath
+    _forward_magic = [
+        '__str__', '__bytes__',
+        '__eq__', '__lt__', '__le__', '__gt__', '__ge__'
```
buhman: > actually call them the way we've done this? 

Yeah; I played with it for awhile, and `__str__` at least definitely works.

I'll add tests for the equalities too.

> tp == sp returns True? What about sp == tp?

Should be `True` for both.

trio/_file_io/_path.py
```diff
@@ -0,0 +1,138 @@
...
+    ]
+
+    def __new__(cls, *args, **kwargs):
+        path = Path(*args, **kwargs)
+
+        self = cls._from_wrapped(path)
+        return self
```
buhman: Uhh, me not refactoring this? :smile:

trio/_file_io/_path.py
```diff
@@ -0,0 +1,138 @@
...
+        return super().__dir__() + self._forward
+
+    def __repr__(self):
+        return 'AsyncPath({})'.format(str(self))
+
+    def __fspath__(self):
+        try:
+            return self._wrapped.__fspath__()
+        except AttributeError:
+            return str(self)
```
buhman: > might reasonably expect to be accepted everywhere paths are accepted

If `PathLike` is supposed to mean "literally implements all of pathlib.Path", then we should invent `__afspath__` probably. If we can make it mean "implements all of pathlib.PurePath" then this is fine.

This is also related to https://github.com/python-trio/trio/pull/180#discussion_r121260360

trio/_file_io/_path.py
```diff
@@ -0,0 +1,138 @@
...
+        return self
+
+    async def open(self, *args, **kwargs):
+        func = partial(self._wrapped.open, *args, **kwargs)
+        value = await trio.run_in_worker_thread(func)
+        return trio.wrap_file(value)
+
+
+# not documented upstream
+delattr(AsyncPath.absolute, '__doc__')
```
buhman: The value of `trio.AsyncPath.absolute.__doc__` makes a reference to `:meth:~pathlib.Path.absolute`, which does not exist. If this line were removed, building docs would fail.

trio/_file_io/_path.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+# python3.5 compat
+if hasattr(os, 'PathLike'):  # pragma: no cover
```
buhman: Do we combine coverage between 3.5/3.6+ properly?

docs/source/glossary.rst
```diff
@@ -0,0 +1,17 @@
+:orphan:
+
+.. _glossary:
+
+********
+Glossary
+********
+
+.. glossary::
```
buhman: :+1:

trio/_file_io/_helpers.py
```diff
@@ -0,0 +1,19 @@
+import trio
+
+from functools import wraps, partial
```
buhman: Hmm, should I add `flake8` or something to the build process in another PR?

trio/_file_io/__init__.py
```diff
@@ -0,0 +1,7 @@
+__all__ = []
+
+from ._file_io import *
+__all__ += _file_io.__all__
+
+from ._path import *
+__all__ += _path.__all__
```
buhman: I tend to prefer making modules as tiny as possible, but if you want them combined, I'll do it.

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
+import io
+import tempfile
+
+import pytest
+from unittest import mock
+from unittest.mock import patch, sentinel
```
buhman: > APIs defined like that should be rewritten so that their guarantees are defined in terms of real things

I am no testing expert, but my thinking is that a unit test should test only the functionality of that unit; so the mocking and patching removes functionality that depends on other units.

trio/tests/test_file_io.py
```diff
@@ -0,0 +1,169 @@
...
+    assert async_file.wrapped == wrapped
+
+
+def test_dir_matches_wrapped(async_file, wrapped):
+    attrs = _FILE_SYNC_ATTRS + _FILE_ASYNC_METHODS
+
+    # all supported attrs in wrapped should be available in async_file
+    assert all(attr in dir(async_file) for attr in attrs if attr in dir(wrapped))
+    # all supported attrs not in wrapped should not be available in async_file
+    assert not any(attr in dir(async_file) for attr in attrs if attr not in dir(wrapped))
```
buhman: > So like if StringIO actually defined all the IO attributes, then the second line would fail to test anything, right?

Well yes. If we wanted to be more rigorous I could test all three cases (having none of the supported attributes implemented, having some of them, and having all of them).

trio/tests/test_file_io.py
buhman: > sentinel objects have all attributes?

Yes.

> spend a line violating the abstraction boundary to convert that into wrap_file(sentinel)?

You are correct; this is bad.

trio/tests/test_file_io.py
buhman: No; because this test is testing `gettattr` while the any/all test is testing `__dir__`.

trio/_file_io/__init__.py
buhman: Unless you mean dfff7ffa25cb2b30aa6bb8f9334049892ff451eb

trio/_file_io/_path.py
buhman: Actually, after looking at this again, the idea was to not duplicate the `_wrapper` assignment in `_from_wrapped`. Even though it is slightly non-dry, it does look nicer just to convert this into an `__init__`, so 5ad8b2c8e23e3b6877d33162bdc827959fe69aa1.

trio/_file_io/_path.py
```diff
@@ -0,0 +1,138 @@
...
+        return 'AsyncPath({})'.format(str(self))
+
+    def __fspath__(self):
+        try:
+            return self._wrapped.__fspath__()
+        except AttributeError:
+            return str(self)
+
+    @classmethod
+    def _from_wrapped(cls, wrapped):
```
buhman: It's supposed to make it convenient to directly wrap an existing `pathlib.Path`, which is used for path methods that return new paths.

buhman:
@njsmith I fixed everything except https://github.com/python-trio/trio/pull/180#discussion_r121260360 which depends on https://github.com/python-trio/trio/pull/180#discussion_r121260207

trio/_file_io/_path.py
njsmith: Yep!

trio/_file_io/_path.py
njsmith: Why not just use `trio.Path(pathlib.Path(...))`? Notice that `pathlib.Path(pathlib.Path("foo")) == pathlib.Path("foo")`.

buhman: Oh, I didn't realize that `pathlib.Path` could do that; I'll duplicate this behavior.

trio/_file_io/_path.py
njsmith: `PathLike` is supposed to mean "provides an `__fspath__` method", nothing more or less. It means the object can be passed to functions like `open` or `os.stat` that have traditionally only taken strings, and will be automatically converted into a string. So we totally *can* implement it; you already did :-). The question here is more subtle: given that all the extant functions that do this coercion are blocking, you shouldn't be calling them from trio *anyway* (at least without taking some special care), so making it easy to call them has little value, maybe negative value.

buhman:
Of course 93f4500 would fail on python 3.5…

trio/_file_io/_path.py
njsmith: `PathLike` just means "has an `__fspath__` method", meaning it can be accepted in places that used to only accept string-paths.

Anyway, we talked about this on gitter and decided that it's probably not really worth trying to nudge people here – the friction just isn't directed in a very useful direction, and in fact `__fspath__` is super useful even just for us in allowing `trio.Path` to interoperate with itself. (Or it will be once we can drop support for 3.5, anyway.)

trio/_file_io/_helpers.py
njsmith: I have mixed feelings about `flake8` because it tends to produce lots of annoying false positives... pyflakes or pylint might be ok, but I suspect that [they'd take some work to configure](https://codewithoutrules.com/2016/10/19/pylint/) to work well, since trio does a bit more metaprogramming than static analysis tools tend to like.

I guess a simple incremental approach would be to start running one of them exclusively in find-unused-imports mode, and then maybe grow from there :-). I tried running them on the code and they do indeed spew tons of false positives, but also caught a whole bunch of unused imports...

trio/tests/test_file_io.py
njsmith: I dunno. If that's unit testing, then I don't like unit testing. I want to write APIs where the relationship between the inputs and outputs is well-defined, and I want tests that check that contract is fulfilled. If other units get used internally then that's an internal implementation detail, and writing tests to make sure that internal implementation details haven't changed sounds like the opposite of helpful; I want my tests to make it *easier* to change internal implementation details!

trio/_file_io.py
```diff
@@ -154,7 +158,7 @@ def wrap_file(file):
 
     """
 
-    if isinstance(file, io.IOBase):
-        return AsyncIOWrapper(file)
+    if not hasattr(file, 'close') or not callable(file.close):
+        raise TypeError('{} does not implement required method close()'.format(file))
```
njsmith: So... the reason I suggested checking for `read` or `write` here is that it seems useful to catch mistakes where someone accidentally passes in some non-file-like object. The problem with checking for `close` is that there are lots of objects in Python that have `close` methods but aren't file-like (e.g. sockets, generators, ...), so it will tend to produce false negatives. OTOH I can't think of any case where an object can useful be called "a file" but it doesn't have `read` or `write`. (Technically I guess we don't even *require* `close` since if the user never calls our `close` method then all will be well. Though it is a bit weird that `close` shows up in `dir()` even if the wrapped object doesn't have it.) So I think checking for `read` or `write` (and possibly also `close`, I dunno) would be better?

trio/_file_io/_path.py
njsmith: Sounds like this `# pragma: no cover` still needs to go?

trio/tests/test_path.py
```diff
@@ -3,30 +3,44 @@
...
+async def test_cmp_magic():
+    a, b = trio.Path(''), trio.Path('')
+    assert a == b
+    assert not a != b
+
+    a, b = trio.Path('a'), trio.Path('b')
+    assert a < b
+    assert b > a
+
+    assert not a == None
```
njsmith: This would probably be a good test to use a loop (either explicitly or via pytest parametrization) to check all combinations of `pathlib.Path` and `trio.Path` objects.

We probably also want similar tests for passing different types to `trio.Path.__init__`, `trio.Path.__div__`, `trio.Path.joinpath`.

trio/tests/test_path.py
```diff
@@ -119,3 +133,14 @@ class MockWrapper:
     Type.generate_wraps(MockWrapper, {'unsupported': None})
 
     assert not hasattr(MockWrapper, '_private')
+
+
+async def test_path_wraps_path(path):
+    other = trio.Path(path)
+
+    assert path == other
```
njsmith: Testing that this succeeds is good, but we should also have a test that `trio.Path(1)` or `trio.Path(object())` fails :-)

trio/_path.py
```diff
@@ -109,30 +121,25 @@ def __dir__(self):
...
-
     async def open(self, *args, **kwargs):
         func = partial(self._wrapped.open, *args, **kwargs)
         value = await trio.run_in_worker_thread(func)
         return trio.wrap_file(value)
 
 
 # not documented upstream
-delattr(AsyncPath.absolute, '__doc__')
+delattr(Path.absolute, '__doc__')
```
njsmith: Can you expand the comment here? Also, this should probably be written `del Path.absolute.__doc__`, since the attribute is a constant.

trio/_path.py
```diff
@@ -0,0 +1,145 @@
...
+
+__all__ = ['Path']
+
+
+def _forward_factory(cls, attr_name, attr):
+    @wraps(attr)
+    def wrapper(self, *args, **kwargs):
+        attr = getattr(self._wrapped, attr_name)
+        value = attr(*args, **kwargs)
+        if isinstance(value, cls._forwards):
```
njsmith: Oh also: [Codecov says](https://codecov.io/gh/python-trio/trio/pull/180/diff#D4-18) that we never have a test where this branch fails. Probably we should test some method that returns a non-`PurePath`? (Are there any?)

...and for that matter, I guess wrapping `PurePath` objects isn't even necessary, it's only `Path` objects that we need to worry about. (This is an example of a place where I'm finding the generic-ness of this code to add unnecessary confusion; if this was hardcoded to `if isinstance(value, PurePath)` it would be easier to notice :-).)

trio/_path.py
```diff
@@ -77,28 +89,28 @@ def generate_magic(cls, attrs):
...
     _forward_magic = [
         '__str__', '__bytes__',
         '__eq__', '__lt__', '__le__', '__gt__', '__ge__'
     ]
 
-    def __new__(cls, *args, **kwargs):
-        path = Path(*args, **kwargs)
+    def __init__(self, *args):
+        # python3.5 compat
+        args = [str(a) for a in args]
```
buhman: > And we also need to call this on the inputs to all Path methods

Welp, this is going to be fun.

trio/_file_io.py
buhman: What about "at least one of `read` or `write`"?

njsmith:
Sounds good to me.

trio/_path.py
buhman: > Probably we should test some method that returns a non-PurePath? 

This is necessary, but the comparison with PurePath isn't strictly correct. I simplified this logic and added a test to trigger this case (a forwarded method that does *not* return a pathlib.Path) in  557dbdb0d6749eba6d79498b64f250c15419ff26 (`::test_forward_methods_without_rewrap`).

By pure accident, the way I wrote that test revealed a new bug:

```python
======================================== FAILURES ========================================
__________________________ test_forward_methods_without_rewrap ___________________________

path = trio.Path(/tmp/pytest-of-zack/pytest-30/test_forward_methods_without_r0/test)
tmpdir = local('/tmp/pytest-of-zack/pytest-30/test_forward_methods_without_r0')

    async def test_forward_methods_without_rewrap(path, tmpdir):
>       path = await path.parent.resolve()
E       TypeError: object PosixPath can't be used in 'await' expression

trio/tests/test_path.py:110: TypeError
========================== 1 failed, 22 passed in 0.07 seconds ===========================
```

buhman:
:100:

PR #181, opened 2017-05-29
Implement capacity limitation for run_in_worker_thread

njsmith:
You can pass a Semaphore here to avoid overwhelming the system with
too many simultaneous threads.

### DIFF
--- a/trio/_sync.py
+++ b/trio/_sync.py
@@ -207,6 +207,8 @@ class CapacityLimiter:
 
         Raises:
           WouldBlock: if no tokens are available.
+          RuntimeError: if the current task already holds one of this sack's
+              tokens.
 
         """
         self.acquire_on_behalf_of_nowait(_core.current_task())
@@ -226,6 +228,8 @@ class CapacityLimiter:
 
         Raises:
           WouldBlock: if no tokens are available.
+          RuntimeError: if ``borrower`` already holds one of this sack's
+              tokens.
 
         """
         if borrower in self._borrowers:
@@ -241,6 +245,10 @@ class CapacityLimiter:
     async def acquire(self):
         """Borrow a token from the sack, blocking if necessary.
 
+        Raises:
+          RuntimeError: if the current task already holds a one of this sack's
+              tokens.
+
         """
         await self.acquire_on_behalf_of(_core.current_task())
 
@@ -253,6 +261,10 @@ class CapacityLimiter:
           borrower: A :class:`Task` or arbitrary opaque object used to record
              who is borrowing this token; see
              :meth:`acquire_on_behalf_of_nowait` for details.
+
+        Raises:
+          RuntimeError: if ``borrower`` task already holds one of this sack's
+             tokens.
 
         """
         await _core.yield_if_cancelled()
--- a/trio/_sync.py
+++ b/trio/_sync.py
@@ -7,7 +7,9 @@ from . import _core
 from ._util import aiter_compat
 
 __all__ = [
-    "Event", "Semaphore", "Lock", "StrictFIFOLock", "Condition", "Queue"]
+    "Event", "CapacityLimiter", "Semaphore", "Lock", "StrictFIFOLock",
+    "Condition", "Queue",
+]
 
 @attr.s(slots=True, repr=False, cmp=False, hash=False)
 class Event:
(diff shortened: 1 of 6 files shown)


njsmith:
I just pushed a new, much-expanded version of this that I think finishes solving the basic don't-run-too-many-threads-at-once problem, and I think it's ready for review if anyone wants to take a look.

trio/_sync.py
```diff
@@ -81,6 +83,245 @@ def async_cm(cls):
...
+             with the intention in the future of using it to `allow deadlock
+             detection and other useful things
+             <https://github.com/python-trio/trio/issues/182>`__
+
+        Raises:
+          WouldBlock: if no tokens are available.
+
+        """
+        if borrower in self._borrowers:
+            raise RuntimeError(
```
kyrias: Shouldn't this also be in the docstring?

trio/_threads.py
```diff
@@ -206,22 +255,56 @@ def current_await_in_trio_thread():
...
-    def trio_thread_fn(result):
+    name = "trio-worker-{}".format(next(_worker_thread_counter))
+    placeholder = ThreadPlaceholder(name)
+
+    # This function gets scheduled into the trio run loop to deliver the
+    # thread's result.
+    def report_back_in_trio_thread_fn(result):
+        print("in trio thread", result)
+        def do_release_then_return_result():
+            print("asdF")
```
kyrias: ;)

njsmith:
@kyrias: Addressed those issues, and also added some docs and a few more tests.

Let's see what Travis makes of this. The [last run where everything froze on PyPy](https://travis-ci.org/python-trio/trio/builds/244172010?utm_source=github_status&utm_medium=notification) makes me very nervous. I can't reproduce it locally, but it seems too much too be a coincidence that both PyPy runs froze inside the thread tests on a PR dealing with the thread code...

njsmith:
Welp, this appears to be a PyPy bug: https://bitbucket.org/pypy/pypy/issues/2591/closed-over-variable-updates-made-in-one

I think we might need to drop PyPy support in our CI until this gets fixed. But for now I'm going to bed, and we'll see if @arigo has magically fixed it by the time I wake up :-)

Carreau:
You can set it in an allow_faillure section in travis.ci. it still tests
but does not mark the GitHub status a red.

njsmith:
I just pushed a commit that skips the test for now on PyPy. I'm not super happy about this, because the test is valid and important. But this is an important feature that I don't want to get blocked by PyPy, and this has a few advantages over `allow-failures`:
* With allow-failures, no-one will notice if other stuff breaks on pypy
* In fact, probably no-one will ever look at the pypy results at all, because when a PR is green then who clicks on the detailed results?
* And the pypy bug was causing freezes and then timing out after 10 minutes, which substantially increases travis's total turnaround time.
* I version-gated the skip logic, so even if we forget about it then eventually a new version of pypy will come along and start running the test again

njsmith:
Here's an exciting plot twist: the pypy bug where local variable assignments can just disappear, also affects all versions of CPython, probably back to 2.1 when closures were introduced! [bpo-30744](https://bugs.python.org/issue30744)

It just happens that we don't trigger it because one of the required conditions is the presence of a Python-level tracing function, which on PyPy is provided by coverage.py, but on CPython coverage.py has a C accelerator module.

trio/_threads.py
```diff
@@ -2,23 +2,31 @@
...
     _core.spawn_system_task(await_in_trio_thread_task, name=afn)
 
 def _run_in_trio_thread_cb(q, fn, args):
-    fn = _core.disable_ki_protection(fn)
-    q.put_nowait(_core.Result.capture(fn, *args))
+    @_core.disable_ki_protection
+    def unprotected_fn():
+        return fn(*args)
+    res = _core.Result.capture(unprotected_fn)
+    print(res)
```
kyrias: I wonder if it would make sense to add some off-by-default logging to various parts of trio, hooking into the logging module or somesuch.  Might help with debugging things in general.

kyrias:
Wowza, that's pretty nasty behavior indeed. AFAICT the code does look sane at least, though I haven't actually played with it yet.

njsmith:
I'm going to call that a review and merge this. This has been way more of a roller coaster than I expected and I want to get it landed before another surprise pops up :-).

(Note for posterity: since it turns out that the test that broke PyPy is also fragile on all versions of CPython, I rewrote it to avoid using `nonlocal` and then reenabled it everywhere.)

PR #186, opened 2017-06-07
Fixing "already say" -> "already saw" typo.

dhermes:
Great tutorial! Thanks for taking the time to write it.


njsmith:
You're welcome, and thanks for helping make it better :-)

Issue #169, opened 2017-05-23
Catching important warnings in test harness


njsmith:
It would be nice if we could make the tests error out on:

* Warning: coroutine '...' was never awaited
* ResourceWarning

What these have in common is that they're both issued by the garbage collector, so the normal way of converting them into errors doesn't work; the only way to detect these is to monkeypatch the `warnings` module, perhaps by adding an `always` filter for the relevant warnings and then installing a custom `warnings.showwarning` hook that records the information somewhere that the test harness can find it.

Some care might be needed to avoid colliding with pytest's warning functionality: https://docs.pytest.org/en/features/warnings.html

kalvdans:
Here is some py.test code that I use to catch never awaited warnings. First it test that the warning actually occur, and then it tests my 'in_parallel_do' function. The basic idea is `with warnings.catch_warnings(record=True) as w` in setup, and `gc.collect()` + ` assert w == []`in the test harness teardown phase.

<pre>
def test_unawaited_tasks():
    import gc
    import pytest
    import warnings

    async def nop(name):
        return name

    gc.collect()
    with pytest.warns(RuntimeWarning):
        nop(3)
        gc.collect()   # To raise RuntimeWarning if we have un-awaited tasks.

    class MyException(Exception):
        pass
    async def raiser():
        raise MyException

    with warnings.catch_warnings(record=True) as w:
        with pytest.raises(MyException):
            in_parallel_do(nop(1), raiser(), nop(3))
        gc.collect()   # To raise RuntimeWarning if we have un-awaited tasks.

    assert w == []
</pre>

njsmith:
Cool, thanks!

Just remembered that another option for detecting `coroutine '...' was never awaited` is to use `sys.set_coroutine_wrapper`, which interacts with #152. And from discussion with @1st1 this week it sounds like 3.7 might have an even nicer way to detect it. But neither of those help with detecting ResourceWarning.

And as an additional wrinkle, the features discussed in this issue are – logically speaking – orthogonal to whether the test case itself is async (which is what `@trio_test` is for). For trio itself I'd like to get errors for these even for synchronous tests.

njsmith:
I guess one option would be to make `@trio_test` work on non-async tests too, and just enable this stuff plus #152.

PR #187, opened 2017-06-07
Turn warnigns into errors in the test suite

Carreau:


Carreau:
Hum apparently it's in core pytest since 2 weeks ago. Is that ok to require py.test >= 3.1 ?

njsmith:
Requiring a new pytest is totally fine.

You might want to modify appveyor.yml as well.

Carreau:
Wooow. The travis job from my fork [got a coredump](https://travis-ci.org/p/trio/jobs/240213854), hopefully the PR is fine, but it's concerning.

njsmith:
> got a coredump

Umm. Wow. That is definitely a bit worrisome. Your link doesn't work for me, though... does it say anything useful?

(I have a suspicion it might be because `ResourceWarning` and `Warning: coroutine '...' was never awaited` run in a weird `__del__` context and python might not be expecting exceptions here.)

njsmith:
(I should say: specifically what I'm nervous of is that this might be a real bug in Python that we might hit again on random test runs depending on the phase of the moon etc. If so we need to figure it out and do whatever we have to to make sure it never happens again. Flaky tests are a constant issue for an I/O library like trio, so I'm super super vigilant about trying to keep them out. Anything else and you end up with like N different bugs that each causes a random test failure in 1/N times and it's unbearable.)

Carreau:
Link updated.
No nothing useful, as it's on travis I can't get it. Python 3.5.2:
```
collected 213 items 
[0m
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_epoll.py .
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_io.py ..................................
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_ki.py .........
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_local.py ......
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_multierror.py ...........
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_parking_lot.py ....
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_result.py ......
../../../../virtualenv/python3.5.2/lib/python3.5/site-packages/trio/_core/tests/test_run.py .....................................ci/travis.sh: line 55:  1366 Aborted                 (core dumped) pytest -W error -ra --run-slow ${INSTALLDIR} --cov="$INSTALLDIR" --cov-config=../.coveragerc

travis_time:end:003d71f4:start=1496806863236690128,finish=1496806912040002488,duration=48803312360
[0K
[31;1mThe command "ci/travis.sh" exited with 134.[0m
```

Carreau:
And it does trigger on travis about every-other time, but apparently AFAICT only 3.5 and 3.5.2 are affected. Relaunch most of the jobs of [this batch](https://travis-ci.org/Carreau/trio/builds/240457205) a  couple of time and only 3.5. and 3.5.2 got failures. So I'm tempted to say it's "fixed".

njsmith:
If it's only 3.5 then that's great for the python devs, but not so helpful for us, since we do want to keep testing on 3.5 :-).

I left `pytest -W error trio` running in a loop for a while last night on my laptop using debian's version of 3.5, and couldn't reproduce.

I guess next step will be to figure out how to get a stack trace, or otherwise figure out where exactly it's crashing...

njsmith:
Just getting a failing run with `-v -s` added might be informative. Especially if we can then compare against the same run but with `-W always`.

Carreau:
> Just getting a failing run with -v -s added might be informative. Especially if we can then compare against the same run but with -W always.

I'll try to do that, and see fi I can bisect which test(s) create that. 
I may attempt to run that using their docker image to reproduce this.

njsmith:
I just built a copy of 3.5.0 locally, and the first time running `~/src/cpython/b/install/bin/pytest -W error -v trio` got a segfault.

It seems to be in `trio/_core/tests/test_run.py::test_error_in_run_loop`, which is extremely suspicious because that's the one test we have that intentionally triggers a `Warning: coroutine '...' was never awaited`

njsmith:
Misc/NEWS for 3.5.3 says:

- Issue [27811](https://bugs.python.org/issue27811): Fix a crash when a coroutine that has not been awaited is finalized with warnings-as-errors enabled.

njsmith:
See gh-189

Carreau:
Ha, good catch thanks !

njsmith:
We still need to come up with something for resource warnings and unawaited coroutine warnings (#169) – ironically these are the two types of warnings that this PR doesn't catch (because they're both generated from `__del__` methods, so can't raise errors), and also the two types of warnings that we actually have in our test suite currently :-). But this is definitely helpful to make sure we don't accidentally add more warnings we weren't expecting. Thanks!

Carreau:
In my other PR, I do attach the un awaited coroutines to NonAwaitedCoroutine error; which does allow to catch the exception and potentially consume them. I can or maybe (should ?) extend it with a context manager that keep a ref to all non-awaited coroutines within it ?

njsmith:
Whoops, meant to merge this :-)

Carreau:
Thanks.

njsmith:
...wtf, somehow the segfault still happened after merging into master: https://travis-ci.org/python-trio/trio/builds/240581553

...ugh, yeah, I can reproduce locally too if I re-run it enough times. Apparently #189 didn't actually fix anything.

njsmith:
#190

njsmith:
Oh right, with all this I almost forgot. My rule is anyone whose PR gets merged gets offered a commit bit :-). Want one? You can participate as much or as little as you like.

Carreau:
> Oh right, with all this I almost forgot. My rule is anyone whose PR gets merged gets offered a commit bit :-). Want one? You can participate as much or as little as you like.

Happily, that will force me to try to review a bit more code :-) 
Side question not to open another issue, are you avoiding type hints on purpose, or by habits ?

njsmith:
> Happily, that will force me to try to review a bit more code :-)

Well, if you put it *that* way... invitation sent ;-)

> Side question not to open another issue, are you avoiding type hints on purpose, or by habits ?

I just haven't had a chance to look at them seriously yet and form an opinion. I'm certainly open to providing them if it would be useful.

A minor issue is that sphinxcontrib-trio might need to be taught to handle them in a way that keeps the docs looking good (https://github.com/python-trio/sphinxcontrib-trio/issues/4), but I'm sure that's solvable and probably needs to happen anyway.

PR #188, opened 2017-06-07
Mark timeout test (~20s) as slow.

Carreau:
Make it skipped by default which highly reduce the runtime of the all test suite, at least locally when "just" running py.test.

### DIFF
--- a/trio/tests/test_socket.py
+++ b/trio/tests/test_socket.py
@@ -4,6 +4,8 @@ import os
 import socket as stdlib_socket
 import inspect
 
+from .._core.tests.tutil import slow
+
 from .. import _core
 from .. import socket as tsocket
 from ..testing import assert_yields, wait_all_tasks_blocked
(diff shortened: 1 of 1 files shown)


njsmith:
Doh, sorry you hit this. This test is actually supposed to run quickly – I run the tests locally all the time, and get grumpy when they take more than ~2.5 seconds total. But I do this on Linux, and it turns out that this test is hitting a [MacOS bug](https://twitter.com/vorpalsmith/status/865825003928928257) :-(. I happened to notice a little while ago while tracking down some other MacOS thing, and fixed it here: 8d404107ed8aa7f7cabd0c2dec3964b4da633360
but unfortunately I was sloppy and so that fix is caught up in the giant branch of #107

Carreau:
No worries, I can easily fix that locally. And yes sometime MacOs is a mess. I'm worried about APFS that will roll as a standard and unlike HFS+ stops normalizing unicode filenames.

njsmith:
#107 is merged, so I think this should be fixed. (Please open an issue if not!)

PR #190, opened 2017-06-08
Fix warnings filter

njsmith:
This is a fixup for 3b831cce397b13b7835bf9384cee75241ee71a73. It turns
out that warning message filters do an *anchored* match, so you have
to write a regexp that matches the beginning of the message. It only
seemed to work because the issue I was trying to fix was intermittent
to start with. Learn something every day...

Before this fix, on 3.5.0 I got a segfault once every ~2-4 runs. After
this fix, I ran 100 times and they all passed.

~/src/cpython/b/install/bin/pytest -W error -v trio -k test_error_in_run_loop --count=100

### DIFF
--- a/trio/_core/tests/test_run.py
+++ b/trio/_core/tests/test_run.py
@@ -880,7 +880,8 @@ def test_error_in_run_loop():
     # turned into errors. See:
     #   https://bugs.python.org/issue27811
     with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", message="was never awaited")
+        warnings.filterwarnings(
+            "ignore", message="coroutine '.*' was never awaited")
 
         with pytest.raises(_core.TrioInternalError):
             _core.run(main)


njsmith:
On further thought, maybe a more general and rigorous solution would be to register this filter globally for 3.5.x < 3.5.3. Either at the top of the test suite, or just when importing trio...

The latter might even help other downstream libraries. Not sure if any of them also want to intentionally set off this warning, but it's possible.

Carreau:
Seeing that we are discussing preventing non-awaited coroutines, enabling that by default may not be the right choice – I also honestly doubt that someone using trio would be on 3.5 . Though a segfault is kinda annoying.

njsmith:
I experimented briefly with moving the filter to `trio/__init__.py` or `conftest.py` and didn't get it to work; maybe pytest only applies the `error` filter after it's finished importing things? Anyway, not interested in fiddling with it more right now, so going to go ahead and merge.

PR #191, opened 2017-06-10
Give better error messages if users mix us up with asyncio

njsmith:
...or other similar libraries.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -654,10 +654,21 @@ class Runner:
         # errors for common mistakes.
         ######
 
-        def _return_value_looks_asyncioish(value):
+        def _return_value_looks_like_wrong_library(value):
+            # Returned by legacy @asyncio.coroutine functions, which includes
+            # a surprising proportion of asyncio builtins.
             if inspect.isgenerator(value):
                 return True
-            if getattr(async_fn, "_asyncio_future_blocking", None) is not None:
+            # The protocol for detecting an asyncio Future-like object
+            if getattr(value, "_asyncio_future_blocking", None) is not None:
+                return True
+            # asyncio.Future doesn't have _asyncio_future_blocking until
+            # 3.5.3. We don't want to import asyncio, but this janky check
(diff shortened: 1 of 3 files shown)


njsmith:
E.g. https://github.com/dabeaz/curio/issues/218

I realized there is one obscure flaw here: because it abandons the task, then it could escape catch-all exception handlers. For example:
```python
async def start_in_trio():
    await trio.run_in_worker_thread(then_in_worker_thread, trio.current_await_in_trio_thread())

def then_in_worker_thread(await_in_trio_thread):
    await_in_trio_thread(back_in_trio)

async def back_in_trio():
    await asyncio.sleep(1)

trio.run(start_in_trio)
```
What this ought to do is propagate an error back through the worker thread to the starting function and then out of `trio.run`. What will actually happen with the current implementation is that the secret system task that `await_in_trio_thread` uses to call `back_in_trio_thread` will be unceremoniously abandoned, and raise an error inside the system task nursery, which will then cancel everything and raise a `TrioInternalError`.

This is tricky to avoid though, because in general we don't *know* how to resume a task that just yielded some foreign object. We could try doing a `coro.throw` but (a) that's awkward to arrange because we never do that otherwise, and (b) it doesn't necessarily work – twisted `Deferred`s don't have a `throw` method (!), and in principle the receiver could do anything with the error (though in practice it looks like asyncio, Tornado, Curio all let the error propagate like we'd want).

Some options:

Make sure that when we're running arbitrary user code and trying to capture the result, we always isolate the code into a standalone `Task`. Seems pretty overkill for this weird corner case? It looks like `await_in_trio_thread` is currently the only place like this in trio though (checked by grepping for `acapture`).

Add the ability to force a call to `throw`: not *too* bad, but if there's no `throw` method then we're back where we started.

Come up with some way to let the `TypeError` propagate properly instead of being converted to `TrioInternalError`?

njsmith:
In any case, this is definitely better than the current behavior (which is to immediately crash the whole run with a `TrioInternalError`), so I'm going to put it in and worry about improving it more later.

Issue #175, opened 2017-05-27
trio.socket should be more careful when re-exporting global functions


njsmith:
It looks like `getprotobyname`, `getservbyname`, `getservbyport` all do disk access and therefore I guess should be made async via `run_in_worker_thread`. (They're also [buggy](https://bugs.python.org/issue30482), but never mind...)

`gethostname` and `sethostname` appear to be direct syscalls (at least on linux). `ntohs` and friends and `inet_*` are pure computation, but `inet_ntoa`/`inet_aton` are considered deprecated in favor of `inet_ntop`/`inet_pton`. Maybe we shouldn't re-export them given this?

The `if_*` functions appear to involve some chatting with the kernel over netlink, but it's all on-CPU and finishes in well under a millisecond, so I don't think we need to treat it as blocking. At least on Linux, haven't checked other OSes.

njsmith:
Well, `getservbyname` and `getservbyport` can generally be replaced by `getaddrinfo` and `getnameinfo` (e.g. `getaddrinfo(None, "http")` gives you port 80, and `getnameinfo(("127.0.0.1", 80), NI_NAMEREQD)` tells you that this is `("localhost", "http")`. I guess in theory someone might want to explicitly only convert the service without touching the hostname (maybe `127.0.0.1` they have a broken system where 127.0.0.1 has no reverse lookup or something), but given that the `socket` module is still available and that putting these into `run_in_worker_thread` is actually buggy (they're not thread-safe), I'm inclined to un-export them at least for now.

njsmith:
Actually, I misunderstood: `getnameinfo(("127.0.0.1", 80), NI_NUMERICHOST)` gives `("127.0.0.1", "http")`, i.e. it's exactly `getservbyport`. So yeah, let's kill 'em.

Another point: `getfqdn` should probably go too: it uses the obsolete `gethostbyaddr`, and AFAICT you can do the same thing better with `getaddrinfo(..., AI_CANONNAME)`. (And otherwise we need to copy the implementation into our code and switch it to use `getaddrinfo`, because IDNA.)

And `getnameinfo` does an implicit `getaddrinfo` if you pass in a string name. So we should probably do the force-numeric-address check on it, like we do for other functions taking an address.

Issue #11, opened 2017-01-22
International domain (IDNA) handling in trio.socket?


njsmith:
Help I have no idea what I'm doing here.

Probably also touches on TLS support, #9.

njsmith:
I have learned a little bit:

For hostname lookup, the versions of python we care about *will* automatically idna-encode unicode strings passed to `socket.getaddrinfo`, so that's nice. However, they use an old version of idna that doesn't always work.

The `idna` package on pypi supports a newer version of idna. I don't know if this always works, but requests seems to use it, so that's a vote of confidence. For hostname lookups it looks like we can just `import idna; getaddrinfo(idna.encode(...), ...)` and it should work.

For SSL hostname verification...... we certainly can't assume that stdlib ssl will do the right thing. (I guess at least it fails closed?) And it's not terribly flexible. ...maybe @lukasa's new PEP will save us? Like everything involving SSL, this appears to be not at all trivial: https://github.com/kennethreitz/requests/blob/362da46e9a46da6e86e1907f03014384ab210151/requests/packages/urllib3/contrib/pyopenssl.py#L141

Lukasa:
> For hostname lookups it looks like we can just `import idna; getaddrinfo(idna.encode(...), ...)` and it should work.

You'll need to make sure that the domain name actually *is* an IDNA, otherwise that'll blow up. The `idna` module is pretty strict about its input. ;)

njsmith:
Are there things that `idna` blows up on where we don't want `getaddrinfo` to blow up? From a quick check it happily accepts `"google.com"` and `"127.0.0.1"`... I guess we do need to check that it's a string, because bytestrings and None are also valid hostnames to `getaddrinfo`... and maybe enable this `uts46=True` thing...

njsmith:
Oh, but `idna.encode("::1")` gives `InvalidCodepoint: Codepoint U+0027 at position 2 of "b'::1'" not allowed`, which just gets more worrisome the more I look at it. It... encoded my text to a bytestring, then took the repr of that bytestring, then got cranky at the apostrophe? do you even python 3.

This is making step 2, "figure out how to make idna and x.509 play together" look *extra* exciting let me tell you. how does anyone do anything.

Lukasa:
It's not cranky at the apostrophe, it's the colon it doesn't like.

Lukasa:
The error message is confusing, but the ultimate issue is that the colon isn't allowed.

njsmith:
[...]

PR #206, opened 2017-06-12
Make trio.socket properly async-correct and IDNA-correct

njsmith:
This fixes part of gh-11.


njsmith:
This grew so that it now also fixes #175.

PR #209, opened 2017-06-12
test_socket: replace imaginary `ip` protocol with `udp`

### LINKED ISSUE #208: test_getprotobyname raises `OSError: protocol not found`
```python
―――――――――――――――――――――――――――――――――――――――――――――――――― test_getprotobyname ――――――――――――――――――――――――――――――――――――――――――――――――――

    async def test_getprotobyname():
        # These are the constants used in IP header fields, so the numeric values
        # had *better* be stable across systems...
>       assert await tsocket.getprotobyname("ip") == 0

trio/tests/test_socket.py:719: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
trio/socket.py:164: in getprotobyname
    _stdlib_socket.getprotobyname, name, cancellable=True)
trio/_threads.py:231: in run_in_worker_thread
    return await _core.yield_indefinitely(abort)
trio/_core/_traps.py:21: in wrapper
    return await fn(*args, **kwargs)
trio/_core/_traps.py:155: in yield_indefinitely
    return (yield YieldIndefinitely(abort_func)).unwrap()
trio/_core/_result.py:118: in unwrap
    raise self.error
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

sync_fn = <built-in function getprotobyname>, args = ('ip',)

    @staticmethod
    def capture(sync_fn, *args):
        """Run ``sync_fn(*args)`` and capture the result.
    
            Returns:
              Either a :class:`Value` or :class:`Error` as appropriate.
    
            """
        try:
>           return Value(sync_fn(*args))
E           OSError: protocol not found

trio/_core/_result.py:32: OSError
```

I guess this ends up reading `/etc/protocols`. It looks like in ubuntu, this starts out like:

```
# Internet (IP) protocols
#
# Updated from http://www.iana.org/assignments/protocol-numbers and other
# sources.
# New protocols will be added on request if they have been officially
# assigned by IANA and are not historical.
# If you need a huge list of used numbers please install the nmap package.

ip	0	IP		# internet protocol, pseudo protocol number
hopopt	0	HOPOPT		# IPv6 Hop-by-Hop Option [RFC1883]
```

Interestingly, the `ip` protocol isn't defined in https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xml, and on my distribution, the content of `/etc/protocols` is generated by parsing that file directly.

buhman:
Fixes #208.

### DIFF
--- a/trio/tests/test_socket.py
+++ b/trio/tests/test_socket.py
@@ -716,5 +716,5 @@ async def test_idna(monkeygai):
 async def test_getprotobyname():
     # These are the constants used in IP header fields, so the numeric values
     # had *better* be stable across systems...
-    assert await tsocket.getprotobyname("ip") == 0
+    assert await tsocket.getprotobyname("ipv4") == 4
     assert await tsocket.getprotobyname("tcp") == 6
--- a/trio/tests/test_socket.py
+++ b/trio/tests/test_socket.py
@@ -716,5 +716,5 @@ async def test_idna(monkeygai):
 async def test_getprotobyname():
     # These are the constants used in IP header fields, so the numeric values
     # had *better* be stable across systems...
-    assert await tsocket.getprotobyname("ipv4") == 4
+    assert await tsocket.getprotobyname("udp") == 17
     assert await tsocket.getprotobyname("tcp") == 6


buhman:
I guess `IPv4 encapsulation` doesn't exist on windows? Maybe we replace this with something just as ancient as TCP, like UDP or ICMP?

njsmith:
Sure, lets go with this... Thanks!

PR #212, opened 2017-06-13
Clearer error message if async context manager used synchronously

Carreau:
Otherwise it says it does not have `__enter__` which is obvious for
advanced pythonista. Still a tiny bit clearer is likely better.

I guess this _might_ fool static type checkers, but I'm unaware of any
that would flag that.

### DIFF
--- a/trio/_core/tests/test_run.py
+++ b/trio/_core/tests/test_run.py
@@ -93,7 +93,7 @@ async def test_nursery_warn_use_async_with():
     with pytest.raises(RuntimeError) as excinfo:
         on = _core.open_nursery()
         with on as nursery:
-            pass # pragma: no-cover
+            pass  # pragma: no cover
     excinfo.match(r"use 'async with open_nursery\(...\)', not 'with open_nursery\(...\)'")
                   
     # avoid unawaited coro.
--- a/trio/_core/tests/test_run.py
+++ b/trio/_core/tests/test_run.py
@@ -89,6 +89,16 @@ async def test_basic_spawn_wait():
         await task.wait()
         assert task.result.unwrap() == 20
 
+async def test_nursery_warn_use_async_with():
+    with pytest.raises(RuntimeError) as excinfo:
+        on = _core.open_nursery()
+        with on as nursery:
+            pass # pragma: no-cover
+    excinfo.match(r"use 'async with open_nursery\(...\)', not 'with open_nursery\(...\)'")
+                  
+    # avoid unawaited coro.
+    async with on:
+        pass
 
 async def test_child_crash_basic():
     exc = ValueError("uh oh")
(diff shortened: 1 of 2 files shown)


trio/_util.py
```diff
@@ -135,6 +136,13 @@ def __init__(self, func, args, kwds):
                 if sys.exc_info()[1] is not value:
                     raise
 
+    def __enter__(self):
+        raise RuntimeError("The function `{}` create an async context manager, use with `async with` and not `with`.".format(self._func_name))
```
njsmith:
```python
"use 'async with {func_name}(...)', not 'with {func_name}(...)'".format(name=self._func_name)"
```
(and 80 column limit!)

trio/_util.py
```diff
@@ -135,6 +136,13 @@ def __init__(self, func, args, kwds):
...
+
+    def __exit__(self):
```
njsmith: `# pragma: no-cover`

Carreau:
Hum, also unawaited coroutines. I need to fix the test.

njsmith:
If it does create a problem for static type checkers then it should be fixable by adding an explicit annotation to `acontextmanager`.

Something else we might want to think about is coming up with a more standard/comprehensive way of doing this kind of checking, like a decorator we can put on all our context managers and iterators. But that doesn't need to block this; this case is a bit special because we can give a nice error message.

trio/_util.py
```diff
@@ -135,6 +136,13 @@ def __init__(self, func, args, kwds):
                 if sys.exc_info()[1] is not value:
                     raise
 
+    def __enter__(self):
+        raise RuntimeError("use 'async with {func_name}(...)', not 'with {func_name}(...)'".format(func_name=self._func_name))
+
+    def __exit__(self):
+        assert False, """Never called, but should be defined""" # pragma: no-cover
```
njsmith: Should be `no cover`, not `no-cover`; it's not actually being recognized right now :-). Also better to put two spaces before the `#`, and to move it up one line, so:
```python
def __exit__(self):  # pragma: no cover
```

trio/_core/tests/test_run.py
```diff
@@ -89,6 +89,16 @@ def test_run_nesting():
         await task.wait()
         assert task.result.unwrap() == 20
 
+async def test_nursery_warn_use_async_with():
+    with pytest.raises(RuntimeError) as excinfo:
+        on = _core.open_nursery()
+        with on as nursery:
+            pass # pragma: no-cover
```
njsmith: Also needs an extra space before `#` and removal of the `-`

njsmith:
Oo, I figured out how to fix the nitpicks through github. Let's see what the CI thinks...

njsmith:
Thanks!

Carreau:
Thanks !

Issue #217, opened 2017-06-13
path operator magic dispatch


buhman:
I was in the middle writing about `Path.__{,r}truediv__` on bpo when I realized we can't have intuitive behavior unless we don't participate in `__fspath__` (undesireable I think), or we justify some really awkward duck-checking in `pathlib.Path.__truediv__` for 3.5+

```python
>>> import trio
>>> import pathlib
>>> p1 = trio.Path('asdf')
>>> p2 = pathlib.Path('qwer')
>>> p1 / p2
trio.Path(asdf/qwer)
>>> p2 / p1 # TypeError on 3.5
PosixPath('qwer/asdf')
```

oremanj:
@buhman What would the fixed behavior here look like?

buhman:
2017-me would have insisted that this issue is thoroughly valid, but I don't remember the context here.

Feel free to close if desired.

PR #218, opened 2017-06-13
path: repr path value in repr

buhman:
I realized in #217 that our repr isn't consistent.


njsmith:
Nice catch.

Possibly useful tip for the future: `"trio.Path({!r})".format(str(self))` would also work – `!r` is format language magic for "format this item by calling repr on it".

PR #222, opened 2017-06-15
test_multierror: Use IPython's code_to_run instead of exec_lines

kyrias:
The reason the tests "freeze" on AppVeyor should have been that
`exec_lines` runs the lines at IPython startup and then leaves the
python shell open, meaning that the process would never exit.  This
should no longer be the case with `code_to_run`, so remove the skip
marker as well.

### DIFF
--- a/trio/_core/tests/test_multierror.py
+++ b/trio/_core/tests/test_multierror.py
@@ -419,7 +419,7 @@ def run_script(name, use_ipython=False):
         cmd = [sys.executable, "-u", "-m", "IPython",
                # no startup files
                "--quick",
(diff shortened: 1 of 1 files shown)


njsmith:
Whoa, good catch! I admit that those tests were written based on a bunch of wild flailing at the IPython command line. That's great to see that that's working. (Now I'm curious why it works for me on Linux and in my windows vm but not on appveyor...  but oh well, not curious enough to actually track it down :-))

Anyway, excellent job, thanks!

njsmith:
Meh, it's not letting me send you an invitation to join the organization from my phone. I'll do it from my computer in a few hours.

PR #225, opened 2017-06-16
Correct the example in wait_all_tasks_blocked

### LINKED ISSUE #224: wait_all_tasks_blocked example correct?
https://github.com/python-trio/trio/blob/ec4cdee7de14ede9fdd8b863af54510c714be99d/trio/_core/_run.py#L1089

When I try to run this example, I get an AssertionError, I believe on the linked line.

Reading the code, I can't tell why we expect the lock to be unlocked at that point when we acquired the lock before entering the "async with" block, but I'm unsure if I'm misunderstanding part of the trio API.

remleduff:
The previous example had a small thinko and asserted the lock wasn't locked when in fact we had just acquired it.

Fix the example, and add a couple of assertions about lock ownership to attempt to clarify things a little.

Fixes #224 but doesn't attempt to find a simpler example to use instead.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1084,17 +1084,21 @@ class Runner:
                  lock = trio.Lock()
                  await lock.acquire()
                  async with trio.open_nursery() as nursery:
-                     nursery.spawn(lock_taker, lock)
-                     # child hasn't run yet
-                     assert not lock.locked()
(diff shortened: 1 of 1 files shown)


kyrias:
:+1: LGTM.

Carreau:
Thanks !

PR #228, opened 2017-06-17
Port fixes for unchaining exceptions in contextmanager from cpython

### LINKED ISSUE #203: Need to fix bpo-29692
Someone recently found some bugs in some subtle `contextlib.contextmanager` corner cases. The fixes are here: https://github.com/python/cpython/pull/949/files ([discussion](https://bugs.python.org/issue29692))

`trio._util.acontextmanager` is a fork of `contextlib.contextmanager`, and has the same bugs. We should import those fixes.

remleduff:
Cherry-pick 647438fa5a7b9a5fb3fd1af57541b52646c7a013,
bpo-29692: contextlib.contextmanager may incorrectly unchain RuntimeError

Fixes #203

### DIFF
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -99,7 +99,7 @@ class _AsyncGeneratorContextManager:
             try:
                 await self._agen.asend(None)
             except StopAsyncIteration:
-                return
+                return False
             else:
                 raise RuntimeError("async generator didn't stop")
         else:
@@ -122,7 +122,7 @@ class _AsyncGeneratorContextManager:
                 # Likewise, avoid suppressing if a StopIteration exception
                 # was passed to throw() and later wrapped into a RuntimeError
                 # (see PEP 479).
-                if exc.__cause__ is value:
+                if isinstance(value, (StopIteration, StopAsyncIteration)) and exc.__cause__ is value:
                     return False
                 raise
             except:
@@ -133,8 +133,9 @@ class _AsyncGeneratorContextManager:
                 # fixes the impedance mismatch between the throw() protocol
                 # and the __exit__() protocol.
                 #
-                if sys.exc_info()[1] is not value:
-                    raise
+                if sys.exc_info()[1] is value:
+                    return False
+                raise
 
     def __enter__(self):
         raise RuntimeError("use 'async with {func_name}(...)', not 'with {func_name}(...)'".format(func_name=self._func_name))
--- a/trio/tests/test_util.py
+++ b/trio/tests/test_util.py
@@ -1,6 +1,10 @@
 import pytest
 
 import signal
+import sys
+import textwrap
+
+from async_generator import async_generator, yield_
 
 from .._util import *
 from .. import _core
@@ -58,3 +62,68 @@ async def test_UnLock():
                 async with ul1:
                     pass  # pragma: no cover
     assert "ul1" in str(excinfo.value)
+
+
+async def test_contextmanager_do_not_unchain_non_stopiteration_exceptions():
+    @acontextmanager
+    @async_generator
+    async def manager_issue29692():
+        try:
+            await yield_()
+        except Exception as exc:
+            raise RuntimeError('issue29692:Chained') from exc
+
+    with pytest.raises(RuntimeError) as excinfo:
+        async with manager_issue29692():
+            raise ZeroDivisionError
+    assert excinfo.value.args[0] == 'issue29692:Chained'
+    assert isinstance(excinfo.value.__cause__, ZeroDivisionError)
+
+    # This is a little funky because of implementation details in async_generator
+    # It can all go away once we stop supporting Python3.5
+    with pytest.raises(RuntimeError) as excinfo:
+        async with manager_issue29692():
+            exc = StopIteration('issue29692:Unchained')
+            raise exc
+    assert excinfo.value.args[0] == 'issue29692:Chained'
+    cause = excinfo.value.__cause__
+    assert cause.args[0] == 'generator raised StopIteration'
+    assert cause.__cause__ is exc
+
+    with pytest.raises(StopAsyncIteration) as excinfo:
+        async with manager_issue29692():
+            raise StopAsyncIteration('issue29692:Unchained')
+    assert excinfo.value.args[0] == 'issue29692:Unchained'
+    assert excinfo.value.__cause__ is None
+
+
+# Native async generators are only available from Python 3.6 and onwards
+nativeasyncgenerators = True
+try:
+    exec("""
+@acontextmanager
+async def manager_issue29692_2():
+    try:
+        yield
+    except Exception as exc:
+        raise RuntimeError('issue29692:Chained') from exc
+""")
+except SyntaxError:
+    nativeasyncgenerators = False
+
+
+@pytest.mark.skipif(not nativeasyncgenerators, reason="Python < 3.6 doesn't have native async generators")
+async def test_native_contextmanager_do_not_unchain_non_stopiteration_exceptions():
+
+    with pytest.raises(RuntimeError) as excinfo:
+        async with manager_issue29692_2():
+            raise ZeroDivisionError
+    assert excinfo.value.args[0] == 'issue29692:Chained'
+    assert isinstance(excinfo.value.__cause__, ZeroDivisionError)
+
+    for cls in [StopIteration, StopAsyncIteration]:
+        with pytest.raises(cls) as excinfo:
+            async with manager_issue29692_2():
+                raise cls('issue29692:Unchained')
+        assert excinfo.value.args[0] == 'issue29692:Unchained'
+        assert excinfo.value.__cause__ is None


trio/tests/test_util.py
```diff
@@ -58,3 +58,27 @@ def handler(signum, _):
...
+            yield
+        except Exception as exc:
+            raise RuntimeError('issue29692:Chained') from exc
+    try:
+        async with test_issue29692():
+            raise ZeroDivisionError
+    except Exception as ex:
+        assert type(ex) is RuntimeError
+        assert ex.args[0] == 'issue29692:Chained'
+        assert isinstance(ex.__cause__, ZeroDivisionError)
```
njsmith: The problem with writing tests this way is that if no exception is raised, then everything looks good... so it's generally better practice to do:
```python
with pytest.raises(RuntimeError) as excinfo:
    ...
assert excinfo.value.args[0] == "..."
...
```
which will actually fail if it doesn't see a `RuntimeError.` OTOH if this is copied directly from the upstream fix and you'd rather keep it close to upstream there's a reasonable argument for that.

trio/tests/test_util.py
```diff
@@ -58,3 +58,27 @@ def handler(signum, _):
                 async with ul1:
                     pass  # pragma: no cover
     assert "ul1" in str(excinfo.value)
+
+
+async def test_contextmanager_do_not_unchain_non_stopiteration_exceptions():
+    @acontextmanager
+    async def test_issue29692():
+        try:
+            yield
```
njsmith: Unfortunately we still support py35, so we can't use native generators like this in the test suite (this is why appveyor and travis are failing) – it needs to be like
```python
@async_generator
async def test_issue29692():
    try:
        await yield_()
    except ...
```

remleduff: I was just digging into this. Unfortunately, it seems that when I try it this way, the actual fix doesn't work anymore.

njsmith: Uh oh. Should I look more closely? `async_generator` is a pretty solid emulation of the built-in async generators, but it's had subtle bugs before....

One thing that just popped out at me looking again though: we surely want to replace all the `StopIteration`s in this patch with `StopAsyncIteration`s

remleduff: That was it, thanks!

trio/tests/test_util.py
remleduff: Went ahead and fixed it up as you suggested. I was just copy-pasting from upstream, but I think it's nicer this way.

trio/tests/test_util.py
njsmith: .....okay, I just re-read PEP 479 and this is more complicated than I realized. I'm actually not sure that what I said above about replacing `StopIteration` with `StopAsyncIteration` is correct, and actually it's quite likely that `async_generator` doesn't handle this correctly. Need to do some experiments to see exactly what the python 3.6 native async generator behavior actually is...

remleduff: OK, current version has passing tests for what it's worth, but I'll sign off for now

njsmith:
Here's my current understanding of this.

The problem being fixed is: suppose we have some code like:
```python
try:
    with my_context_manager():
        raise StopIteration
except StopIteration:
    print("caught it!")
```
this is totally fine and normal, and should print `caught it!` (unless the context manager specifically catches this exception, but let's assume it doesn't). Now let's assume `my_context_manager()` is implemented as a `contextlib.contextmanager` wrapper around a generator. Then the way this code should work is that `StopIteration` will be thrown into the generator, which lets it escape... and then we have a problem! PEP 479 made it so that if a generator raises `StopIteration`, this gets automatically converted into a `RuntimeError`. In general this is a good thing because it lets us tell the difference between a generator that raises `StopIteration`, and one that exits normally (which causes the interpreter to automatically raise `StopIteration`). But it makes life complicated for `contextlib.contextmanager`, because if it lets this `RuntimeError` escape, then it'll give the wrong answer on our code above.

So the solution is that the `contextmanager` code needs to detect when we've got a `RuntimeError` that was wrapped around a `StopIteration`, and unwrap it again. And that's what the version I originally copied this code from tried to do. However, it was a bit over-eager in how it did this: it unwrapped *any* `RuntimeError` that it saw that had a `__cause__` attribute, not just `RuntimeError`s that were automatically wrapped around a `StopIteration`. So the idea of the patch we're trying to adapt, is that it makes this check more careful. The two tests are: (a) now if we get a `RuntimeError` that's wrapped around something that's not `StopIteration`, then the `RuntimeError` should be preserved, and (b) if we *do* get a `RuntimeError` that's wrapped around a `StopIteration`, then it should continue to be correctly unwrapped.

Phew, ok, now we understand what's going on upstream with regular synchronous `contextmanager`. How does this translate to our async `acontextmanager`? This is complicated because internally async generators use both `StopIteration` (to signal the end of each call to `anext`/`asend`/`athrow`) and also `StopAsyncIteration` (to signal the end of the overall iteration).

I don't think this is documented anywhere, but based on some experimentation, it looks like the way PEP 479 was applied to the native async generators in 3.6 is: if they raise *either* `StopIteration` *or* `StopAsyncIteration`, then that gets wrapped into a `RuntimeError`.

So this suggests that `acontextmanager`'s unwrapping code needs to check for both of these special exception types, and indeed, [that's exactly what the upstream `asynccontextmanager` (only available in 3.7-dev builds) does](https://github.com/python/cpython/blob/6a98a04e21fc3209b2170c3930ce4488e89680e3/Lib/contextlib.py#L175).

Then there's an additional wrinkle: it looks like currently, `async_generator` async generators correctly wrap `StopIteration` in `RuntimeError`, but they *don't* correctly wrap `StopAsyncIteration` in `RuntimeError` – they just let `StopAsyncIteration` pass through unchanged.

So:

We should make the unwrapping logic check for both `StopIteration` and `StopAsyncIteration`, like the upstream `asynccontextmanager` does.

We should probably test throwing both `StopIteration` and `StopAsyncIteration` through an async context manager (like the last test in here does currently).

From comments above, it sounds like making that last test use `StopIteration` doesn't actually work? If so then I'm confused – it seems like it should work to me!

And currently that last test isn't actually testing the unwrapping logic, because it's using `StopAsyncIteration` + an `async_generator`-based generator, so the exception is never getting wrapped in the first place. I'm looking at making `async_generator` properly implement PEP 479, but I'm currently hitting some weird error that I don't understand.

Arguably this is not a big deal though because `acontextmanager` is not a public API, and all of our internal users use it with `async_generator`-based generators, so as long as those two working together give the right result it doesn't really matter why.

njsmith:
Also, sorry about this -- I really underestimated how complicated this would be when I put the good-for-new-contributors tag on #203!

njsmith:
> I'm looking at making `async_generator` properly implement PEP 479, but I'm currently hitting some weird error that I don't understand.

I just sorted this out and released async_generator 1.8, which now properly implements PEP 479. Going to try hitting the rebuild buttons for this PR now and see if it changes anything...

trio/_util.py
```diff
@@ -122,7 +122,7 @@ def __init__(self, func, args, kwds):
                 # Likewise, avoid suppressing if a StopIteration exception
                 # was passed to throw() and later wrapped into a RuntimeError
                 # (see PEP 479).
-                if exc.__cause__ is value:
+                if type is StopAsyncIteration and exc.__cause__ is value:
```
njsmith: Should check for `StopIteration` as well. I guess `if issubclass(type, (StopIteration, StopAsyncIteration)) and exc.__cause__ is value:` might be the most idiomatic way to write it

trio/_util.py
```diff
@@ -133,8 +133,10 @@ def __init__(self, func, args, kwds):
                 # fixes the impedance mismatch between the throw() protocol
                 # and the __exit__() protocol.
                 #
-                if sys.exc_info()[1] is not value:
-                    raise
+                if sys.exc_info()[1] is value:
+                    return False
+                raise
+            raise RuntimeError("generator didn't stop after throw()")
```
njsmith: This last line (`raise RuntimeError(...)`) is redundant -- we there's another check for the same thing up above. Should be deleted.

trio/tests/test_util.py
```diff
@@ -58,3 +60,25 @@ def handler(signum, _):
...
+        async with test_issue29692():
+            raise ZeroDivisionError
+    assert excinfo.value.args[0] == 'issue29692:Chained'
+    assert isinstance(excinfo.value.__cause__, ZeroDivisionError)
+
+    with pytest.raises(StopAsyncIteration) as excinfo:
+        async with test_issue29692():
+            raise StopAsyncIteration('issue29692:Unchained')
+    assert excinfo.value.args[0] == 'issue29692:Unchained'
+    assert excinfo.value.__cause__ is None
```
njsmith: ...And we need to a copy of this test, except with `raise StopIteration(...)` replacing `raise StopAsyncIteration`

remleduff: Thanks for the intense commentary, very interesting. I think I follow. ;)

I've updated everything with your comments and grabbed the newest async_generator, and most things look better, but still running into one issue.

I added some tests gated on version that use native async generators and they all pass, but using async_generator passes for StopAsyncIteration and fails for StopIteration.

I've dug in a bit, but hope you see something obvious that I'm missing.

Pushed my updates for now.

trio/_util.py
remleduff: I copied the code from the 3.7 version you linked and used "isinstance(value, (StopIteration, StopAsyncIteration)" which appeared to be equivalent to your suggestion.

trio/tests/test_util.py
```diff
@@ -58,3 +61,54 @@ def handler(signum, _):
...
+pytestmark = pytest.mark.skipif(not nativeasyncgenerators, reason="Python with native async generators")
+
+
+async def test_native_contextmanager_do_not_unchain_non_stopiteration_exceptions():
+    @acontextmanager
+    async def test_issue29692_2():
+        try:
+            yield
+        except Exception as exc:
+            raise RuntimeError('issue29692:Chained') from exc
```
njsmith: Unfortunately 3.5 blows up if you even have syntax like this in your *file*, it doesn't matter whether you try executing it or not. So the only way I know to write a test like this is to use `exec`:
```python
code = textwrap.dedent("""
    @acontextmanager
    async def test_issue29692_2():
        ...
""")
ns = {"acontextmanager": acontextmanager}
exec(code, ns)
test_issue29692_2 = ns["test_issue29692_2"]
```
(and then you still need a version check to make sure you only do this on 3.6+, or alternatively try it always and then catch the `SyntaxError` if it doesn't work)

It might make sense to consolidate with the previous test... like make a version dependent list of test context managers, and then do a `for` loop over it? Or not, sometimes it's simpler just to have a bit of copy-paste in test code :-)

trio/tests/test_util.py
```diff
@@ -58,3 +61,54 @@ def handler(signum, _):
...
+        with pytest.raises(cls) as excinfo:
+            async with test_issue29692():
+                raise cls('issue29692:Unchained')
+        assert excinfo.value.args[0] == 'issue29692:Unchained'
+        assert excinfo.value.__cause__ is None
+
+
+# Native async generators are only available from Python 3.6 and onwards
+nativeasyncgenerators = sys.version_info >= (3,6)
+pytestmark = pytest.mark.skipif(not nativeasyncgenerators, reason="Python with native async generators")
```
njsmith: I think this marks the whole file as skipped on 3.5? That doesn't seem like what we want. IIRC the syntax is
```python
# skip a single test
@pytest.mark.skipif(check, reason="whatever")
def test_something():
   ...

# skip multiple tests with the same reason
need_native_async_generators = pytest.mark.skipif(...)

@need_native_async_generators
def test_something():
   ...

@need_native_async_generators
def test_something_else():
   ...
```

trio/tests/test_util.py
```diff
@@ -58,3 +61,54 @@ def handler(signum, _):
...
+
+
+async def test_contextmanager_do_not_unchain_non_stopiteration_exceptions():
+    @acontextmanager
+    @async_generator
+    async def test_issue29692():
+        try:
+            await yield_()
+        except Exception as exc:
+            raise RuntimeError('issue29692:Chained') from exc
```
njsmith: OKAY, I figured out the problem with `StopIteration` and `async_generator`!

Here it is: when a `StopIteration` gets thrown into an `async_generator`, it actually ends up inside the magic `await yield_()` function. But Python doesn't know that it's supposed to be a magic function that acts like `yield`; it just thinks its some random async function. And then it tries to raise `StopIteration`, and Python's like uh, no, haven't you heard of PEP 479? Async functions can't raise `StopIteration`. So it immediately wraps it into a `RuntimeError`. Which ... is a little weird and annoying, but there's nothing we can do about it, and it doesn't stop us from unwrapped it later. Except... the way this particular test is written, where it wraps everything in a `RuntimeError`, means that we end up wrapping our `StopIteration` in a `RuntimeError` *twice*. We get `RuntimeError` → `RuntimeError` → `StopIteration`.

So there is something of a bug here, but it's intrinsic to how `async_generator` works, and we can't fix that part.

But! We can still test the main thing we care about, which is that a `StopIteration` is able to pass through a context manager like
```python
@acontextmanager
@async_generator
async def simpler_context_manager():
    await yield_()
```

And this should work, though slightly differently than if it were a native generator. With a native generator, the `yield` would raise `StopIteration`, which would then get wrapped in a `RuntimeError` when it tried to leave the function. With `async_generator`, the `await yield_()` directly wraps the `StopIteration` into a `RuntimeError`, and then that propagates through the function and out. But either way we can catch it and unwrap it in `@acontextmanager`.

njsmith: See #229 for an issue to record the weird edge case that this creates in trio proper...

trio/tests/test_util.py
remleduff: Yeah, was just discovering that myself. I hadn't actually installed Python3.5 until about 15 minutes ago and didn't realize what I was doing didn't work until Travis let me know.

trio/tests/test_util.py
njsmith: > Thanks for the intense commentary, very interesting. I think I follow. ;)

Feel free to ask for clarifications if I get too telegraphic! Often when I write a big chunk of text like that I'm partly writing it as a way to think things through for myself, so I sometimes do a bad job of making it interpretable to other people – but part of the goal is to explain to others and to leave a record for when we come back to an issue later. And both of these work much better if someone pokes me when I'm being unclear :-)

trio/tests/test_util.py
njsmith: > I hadn't actually installed Python3.5 until about 15 minutes ago

That's very sensible of you really... I'm surprised at how quickly the py35 support is becoming an annoying limitation :-(. I'm really reluctant to drop pypy support though, and they don't even have an ETA for starting to do 3.6 yet :-(.

remleduff: Got the test stuff fixed up on my computer for both versions of python.

If I'm understanding you correctly, we're ok with some differences in async_generator for StopIteration, so I just fleshed that out enough to know what the result is at the moment.

Probably not a huge deal until Python3.5 support isn't needed anymore, hopefully.

njsmith: > If I'm understanding you correctly, we're ok with some differences in async_generator for StopIteration

Yep!

kyrias:
Canceled and restarted the job since [Travis was having some maintenance](https://www.traviscistatus.com/incidents/m2mq4r4smxp4) hopefully it gets scheduled now?

kyrias:
Seems they have some problems with non container builds actually: https://www.traviscistatus.com

Backlogs keep building without being processed. Build slaves fell over maybe.

njsmith:
Thanks!

It looks like I missed doing this on your last PR – our policy is that anyone who submits at least 1 merged PR automatically gets an invitation to join the github organization if they want it. So, check your inbox! No pressure – it's totally up to you whether you want to accept, and if you do then you can participate as much or as little as you like. Basically this is our way of saying hey, welcome, we'd love it if you want to stick around :-).

PR #232, opened 2017-06-21
Two small followups to gh-228

njsmith:


kyrias:
LGTM.  Though looking at this diff in GitHub reminds me, I wonder if they use git to create the diff, or something of their own.  Would be nice if it could detect async functions as a context header marker thing, rather than just using the last non-async function...

njsmith:
@kyrias: oh, huh, that's a good point. Probably worth emailing support@github about – do you want to do that? They've been pretty responsive in my experience.

PR #244, opened 2017-07-24
Switch to using trustme for testing TLS certs

njsmith:
Don't merge this yet - it needs
  https://github.com/python-trio/trustme/pull/4
to be merged and released first.

I guess this won't be a problem b/c all the tests are going to fail
mightily until that happens. But I figured I'd get put the branch up
in any case so it doesn't get misplaced.


njsmith:
New trustme is released; let's see how this goes.

njsmith:
Ok, did *another* release of trustme, now actually testing on Windows. Maybe this will work better.

Issue #170, opened 2017-05-24
Mock network for testing


njsmith:
@glyph gave a [great talk at PyCon this year](https://www.youtube.com/watch?v=0By5yfhkiRs) that involved using a virtual (= in memory, in python) networking layer to build a virtual server to test a real client.

As far as the virtual networking part goes, we have some of this, e.g. #107 has some pretty solid in-memory implementations of the stream abstraction. But it would be neat to virtualize more of networking, e.g. so in a test I can have tell my real server code to listen on some-server.example.org:12345 and tell my real client code to connect to that and they magically get an in-memory connection between them.

Fixing #159 would reduce the amount of monkeypatching needed to do this, but OTOH I guess monkeypatching the whole `trio.socket` module is probably the simplest and most direct way to do this anyway... or we could hook in at the socket layer (have it check a special flag before allocating a new socket) or at the high-level networking layer (`open_tcp_stream` checks a special flag and then returns a `FakeSocketStream` etc.). Fundamentally there's going to be some global state because no-one will put up with passing around the whole library interface as an argument everywhere, literally every async library has some kind of contextual/global state they use to solve this problem, and I can't think why it would matter a huge amount whether that's `from twisted.internet import reactor` vs `asyncio._get_running_loop()` vs `trio.socket.socket()`. So I'm leaning towards not worrying about monkeypatching. (The one practical issue I can think of is if someone is trying to use trio in two threads simultaneously, then this will cause some problems because the monkeypatch would be global, not thread-local. Maybe we can make it thread-local somehow? Or maybe we just don't care, because there really isn't any good reason to run your test suite multi-*threaded* in Python.)

[...]

PR #253, opened 2017-07-27
Make it possible to override hostname resolution and socket behavior

njsmith:
See gh-170.

Still needs tests.


njsmith:
https://github.com/python-trio/trio/issues/170#issuecomment-318314243 in particular has some discussion of the design decisions here.

PR #263, opened 2017-08-04
Add a checkpoint to fix #119

### LINKED ISSUE #119: control-C sometimes getting missed on Windows
We're getting occasional failures on appveyor, that look like:

```
\trio\_core\tests\test_ki.py::test_ki_is_good_neighbor PASSED
..\trio\_core\tests\test_ki.py::test_ki_wakes_us_up 
[... coverage info ...]
!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
to show a full traceback on KeyboardInterrupt use --fulltrace
C:\projects\trio\trio\_core\_run.py:1136: KeyboardInterrupt
==================== 14 passed, 1 skipped in 23.20 seconds ====================
```

([source](https://ci.appveyor.com/project/njsmith/trio/build/1.0.271/job/fucboe5llmwcn0eu))

This is a test that delivers a synthetic SIGINT while the trio thread is sleeping, to make sure that the `KeyboardInterrupt` gets delivered promptly.

`_run.py:1136` is the `raise KeyboardInterrupt` at the end of `run()` – the one that checks `ki.pending` one last time as `run` is exiting.

Given the time report at the end, it looks like the `await sleep(20)` in `test_ki_wakes_us_up` may be expiring? And that's the only checkpoint in the test, so if we somehow aren't being woken up by the signal arriving, but the signal is in fact arriving, then this would make sense.

...how can this be happening though? I'm really not sure :-(. May need to push some instrumented builds in a PR and keep rebuilding them on appveyor until I get failures...

njsmith:


njsmith:
I think this should actually finally fix #119. We'll see! But: https://github.com/python-trio/trio/issues/119#issuecomment-320347013

PR #282, opened 2017-08-11
Add high-level Listener interface

njsmith:
Todo:
- [x] `open_ssl_over_tcp_listeners`
- [ ] `start_listening_server` (or maybe `spawn_listening_server`?)
- [ ] `start_{tcp,ssl_over_tcp}_server`
- [ ] rearrange docs
- [ ] fix tutorial
- [ ] ~~switch to preferring port=None instead of port=0 for "pick a
  random port" (apparently python's getaddrinfo treats them the same)~~ never mind, see [bpo-31198](https://bugs.python.org/issue31198)
- [ ] ~~maybe make `open_tcp_listeners` accept a list of host names? not
  sure if this is really useful. But asyncio supports it.~~ OTOH twisted and tornado don't and it makes #283 more complicated, so never mind. We can always add it later; it's a backwards compatible change.


njsmith:
Actually, it looks like we need to resolve #284 before adding server helpers, so I'm going to defer that part to another PR and get this in.

Issue #272, opened 2017-08-08
ParkingLot and Queue are embarrassingly slow


njsmith:
@sorcio sent a small benchmark that I don't think I can share publically, but that implements a simple message broker: clients can connect and say "send this blob of data to channel X" or "please send me the data from channel X", and it brokers between them. His initial version used a `trio.Queue` for message passing and a `trio.hazmat.ParkingLot` to implement a little `Event`-like object. It turned out that some ridiculous proportion of runtime was going into these, like when I dropped in a some simple non-fair versions using a `set` to track sleeping tasks, the overall execution got 2x faster.

Looking at `ParkingLot`, I think at least for now we should be able to make it much faster without losing any features, by replacing the `SortedDict`-of-tickets with an `OrderedDict`-of-tasks. The trick to getting rid of the tickets is that when we requeue, we can go modify `task._abort_fn` directly. (Hey, we're in `_core`, might as well make use of it!) This means that requeuing tasks from one lot to another would put them behind the tasks that were waiting there rather than preserving the global order, but I think that's fine, maybe even better than the current system. This also means if we *wanted* to add tags to sleeping tasks to implement task-fair RWLocks, we could do that for "free", by storing it in the value entry in the `OrderedDict` (otherwise the value would just be `None`).

I'm not sure if switching to a shared global ParkingLot API would be better or not. It would allow us to skip allocating a `ParkingLot` entirely for uncontended objects – like a `Queue(1)` that just gets used once wouldn't need a put lot at all. OTOH with a global object we'd have to jump through several dict lookups each time to find it. It likely doesn't make a big difference either way, but it's probably worth benchmarking at some point.

I'm not sure if the Queue issue is the same or different – Queue obviously depends on ParkingLot, but has a somewhat overcomplicated implementation on top of that using a bunch of semaphores and stuff.

We could redo Queue as:
* `put`: if there's a `get` waiting, hand it the object; otherwise, immediately append to deque, and if over capacity, go to sleep.
* `get`: if there's data in the queue, take it and wake a `put`; otherwise, go to sleep and wait to be handed some data.

[...]

PR #287, opened 2017-08-15
Faster ParkingLot

njsmith:
See gh-272 for context.


njsmith:
I wanted to get a queue rewrite in here as well, but I want to get 0.2.0 out even more, so I'm going to put in the `ParkingLot` changes alone and defer the `Queue` part until later.

Issue #291, opened 2017-08-17
Deprecate socket sendall method in favor of SocketStream.send_all


parity3:
I looked at the high level async socket.sendall() trio 0.1.0 code and realized nothing would stop a user from calling sendall() from 2 tasks on the same socket, and having the individual lowlevel send()'s being intertwined, which is probably not what the user wants (ie fragmented/mashed sendall() data ordering).

Some options are:

- write some concurrency protection, raise an error
- write some documentation saying sendall() is not atomic
- make sendall atomic (via locking/parking lots, etc because I don't think we can safely disable checkpoints)

njsmith:
The high-level `SocketStream` wrapper [does do this check](https://github.com/python-trio/trio/blob/ec2b5a83ab21c8a532ae9761359f42f78c1ccda8/trio/_network.py#L106). (Though [that code is a bit confusing](https://github.com/python-trio/trio/issues/197) -- it may say `lock`, but it actually does raise an error if multiple tasks try to call it concurrently.) So does `trio.hazmat.wait_socket_writable`, which is the low-level primitive that `send` calls when it needs to block. So the situation where this could go wrong is when someone is using the socket `sendall` method directly (hopefully rare once the high-level API is released), and a somewhat tricky race happens like:

* Task A's `sendall` calls `send`, which does a partial `send` (meaning the output buffer is full)
* Task B's `sendall` gets scheduled immediately afterwards, before Task A can run again, and its `send` *succeeds*, because those extra few milliseconds were enough for the kernel to drain the send buffer
* Task A gets rescheduled and its `sendall` calls `send` again.

This is probably rare but I bet it *can* happen in real life.

[...]

Issue #68, opened 2017-03-01
Use consistent terminology to distinguish functions that run sync callables vs. async callables


njsmith:
We have a bunch of functions that take a callable and run it:

* `trio.run`
* `trio.run_in_worker_thread`
* `run_in_trio_thread`
* `await_in_trio_thread`

The first and last take an async function, and the middle two take a sync function. Our terminology here is really confusing. And unfortunately the obvious solution of renaming it to `trio.await` is impossible because `await` will soon be a keyword. So not sure what the best solution is.

(There's also `call_soon_thread_and_signal_safe`, but it's kind of an oddball because it schedules the call for later rather than running it right there.)

Brainstorming some options (format: run-an-async-fn / run-a-sync-fn / call soon):

* arun / run / call_soon
* run / call / call_soon
* run / run_sync / (call_soon or run_sync_soon)
* run_async / run_sync / ...

njsmith:
I'm leaning towards `run` / `run_sync`, but this has the downside that `run_in_trio_thread` is used both before and after, with different meanings, so a normal deprecation doesn't work. We could just rip off the band-aid, but maybe we should take this opportunity to make another change. I've noticed it's kind of weird and annoying to have to call both `current_run_in_trio_thread` and `current_run_sync_in_trio_thread` and keep track of both returned values; if you want to call trio APIs from a thread you'll probably need both. So maybe we should define a concept of a ReentryVehicle: an object that provides `run` and `run_sync` methods that are appropriate to where-ever you're launching from. So it'd be like:

```python
in_trio = trio.current_blocking_reentry_vehicle()
# then in your thread
in_trio.run(...)
```

And you can imagine providing similar vehicles to reenter from asyncio or twisted or whatever.

...for asyncio / twisted / etc., it also makes sense to have trio→them reentry vehicles. Might want to keep that in mind when naming things: an asyncio→trio vehicle is no the same as a trio→asyncio vehicle.

(If an API design allows for an awesomely punny name then it's probably a good idea. I'm pretty sure that's in the zen of python somewhere.)

njsmith:
Oh, googling "reentry vehicles" doesn't give me spaceships like I was thinking of, it's all nukes. Maybe that's not as cool as I thought.

njsmith:
We could call them "portals" maybe, retreating from the grim science fictional present into a pastoral fantasy past. Or, you know, Aperture Science 🍰.

[...]

PR #296, opened 2017-08-19
Add deprecation framework, and do some deprecations

njsmith:
`trio.socket`'s `sendall` method is deprecated in favor of the new
high-level interface (see #291)

`trio.run_in_worker_thread` is renamed to
`trio.run_sync_in_worker_thread` (see #68), with the old name retained
for now as a deprecated alias.

### DIFF
--- a/docs/source/_static/hackrtd.css
+++ b/docs/source/_static/hackrtd.css
@@ -5,6 +5,13 @@ pre {
     line-height: normal !important;
 }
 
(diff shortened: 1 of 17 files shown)


njsmith:
The codecov results here don't make any sense at all. The coverage clearly isn't even aligned with the underlying source code ­– there are blank lines marked as "covered", and lines with code marked as "no code here".

Issue #284, opened 2017-08-11
How to start up a complex task tree in an orderly way, given spawn's non-determinism?


njsmith:
(Text below was posted by @matham as https://github.com/python-trio/trio/issues/32#issuecomment-321672889; I'm splitting it off into a new issue to keep the threads tidy)

-------

So I kinda just ran into this. I have a parent and a bunch of children (ordered).

Naively I tried:

```py
def init_start(self):
    init_things  # get things started

async def wait_to_finish_work(self):
    await do_things

async def start(self):
    self.init_start()
    async with trio.open_nursery() as nursery:
        for child in self.children:
            nursery.spawn(self.start)
        nursery.spawn(self.wait_to_finish_work)

>>> await root.start()
```

I was hoping that the children would spawn deterministically in order and at least run until the first await in `wait_to_finish_work`. This would start the tree in **depth first order**. However, as I noticed in the result and in the design above, the initial spawning is also not in order.

Specifically, you mention that newly spawned tasks may inherit their parent's `vtime`. I wonder if it may not be better that the nursery initially *birth* its children and start their life until the first suspension in the deterministic order. Spawn *looks* deterministic and the docs don't really say much either way.

The alternative for a case like mine where you want to start off each spawnee in order in current trio would be to re-write my example as:

```py
def init_start(self):
    init_things  # get things started

async def wait_to_finish_work(self):
    await do_things

async def start(self):
    for child in self.children:
        child.init_start()
    async with trio.open_nursery() as nursery:
        for child in self.children:
            nursery.spawn(self.start)
        nursery.spawn(self.wait_to_finish_work)

>>> root.init_start()
>>> await root.start()
```
The second version is not that different, but it does make it more complex, especially for cleanup as the parent starts it in two methods. But also, this is now **breadth first order**, where at each level a node  is picked randomly to start off its children.

Edit: I'm thinking that maybe my model of spawn is wrong. E.g.

```py
async with trio.open_nursery() as nursery:
    nursery.spawn(a)
    nursery.spawn(b)
    nursery.spawn(c)
```

Looks to me like 

```py
async with trio.open_nursery() as nursery:
    magic await a()
    magic await b()
    magic await c()
```
Where the magic collects the tasks at the `with` exit and runs them side by side. Perhaps I'm wrong and the better analogy is:

```py
t1 = Thread(target=a)
t2 = Thread(target=b)
t3 = Thread(target=c)
for t in (t1, t2, t3):
    t.start()
[...]

Issue #136, opened 2017-04-23
Potential API changes to help users who don't realize parenting is a full-time job


merrellb:
I've managed to block in a nursery scope several times (eg "while True:" loops) while learning trio, not realizing I had not hit ```__aexit__```.  For long-running processes, this will appear to succeed as we don't expect to exit until KeyboardInterrupt.  However, this will have disabled all of the reaping machinery in ```__aexit__``` and will be silently accumulating zombie tasks indefinitely.

The documentation is fairly clear on this so I am hesitant to suggest the need to add idiot-proofing, but if there is an easy way to detect the nursery is running but not in ```__aexit__``` it might prevent some grief.

njsmith:
The challenge here is that we do want to allow for things like custom supervisors, where code inside the nursery *takes over* for `__aexit__`. For example, see `trio._core._run.Runner.init`, which is the actual root of trio's task tree. It's nursery supervises. the user's main task + system tasks, and it has a special bit of semantics: in addition to the normal rule that it cancels everything when any task crashes, it also cancels everything when the user's main task exits successfully.

But, it just occurred to me that because implementing a custom supervisor is a pretty unusual thing to be doing, we might be able to define some heuristic that catches most accidental misuses, and then if you intentionally want to block inside the nursery then there's a special flag you can enable. Like maybe it would make sense to have `open_nursery` *by default* raise an error if the body tries to execute a checkpoint, but you can do `open_nursery(allow_checkpoints=True)` to allow it.

I guess the question then is whether "contains a checkpoint" is a good heuristic for "something is wrong"... thinking out loud here, the cases we want to catch are when someone starts doing arbitrary amounts of work directly inside the nursery. That's almost never the right thing to do because "parenting is a full time job". In trio, if you're doing arbitrary amounts of work then you pretty much have to execute some checkpoints in the process. It's possible to avoid this (e.g. `time.sleep(10000)`), but there's not much trio can do about this in general (it literally has no control while the `sleep` is running), it's a much more general problem than just nurseries, and people are relatively unlikely to do this by accident. So "no checkpoints" does probably rule out the cases we want to rule out.

[...]

PR #298, opened 2017-08-21
Add nursery.start and nursery.start_soon

njsmith:
start_soon is just a new name for spawn, except it doesn't return the
new task (in preparation for gh-136, where we're going to stop
emphasizing task objects in the main api)

start is a major new feature: it provides a very simple way to start
up a long running task, while blocking until it's finished whatever
initialization it wants to do. At least... it's simple from the user's
point of view. Internally it's quite tricky indeed. The whole _run.py
file probably needs some refactoring and splitting up, but this is one
of those cases where I think it's best to first get the new
functionality working and nailed down, and then we can see what shape
the new abstractions should be.

Fixes gh-284.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,6 @@ dist: trusty
 matrix:
   include:
     - os: linux
-      language: python
-      python: 3.6
-      env: DOC_BUILD=1
-    - os: linux
-      language: python
-      python: 3.6
-      env: FORMATTING=1
-    - os: linux
       language: generic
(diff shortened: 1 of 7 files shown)


matham:
I think there's a typo at https://github.com/python-trio/trio/pull/298/files#diff-e4df893b734a53b4ac0bb2ec532fccbfR926. It should say Like :meth:`start_soon`.

njsmith:
Good catch :-)

Issue #266, opened 2017-08-04
[discussion] Mechanism for "sharing" tasks between different branches of the task tree


njsmith:
Use case 1: there's an expensive idempotent operation that multiple tasks might want to call. Examples within trio itself would include `getaddrinfo` for a particular host, or `WaitForSingleObject` on a single object (as needed for e.g. the windows version of `Popen.wait`). If two tasks make the same call at the same time, then we'd like to coalesce those into a single underlying call and then report the result back to both callers.

Use case 2: there are multiple user-level objects that underneath are multiplexed onto a single object. (For example: multiple HTTP/2 channels on top of a single TCP connection.) Now you need the invariant: whenever at least one task is blocked in `await channel.receive_some()`, there needs to be a task reading from the underlying TCP connection -- *but* there should never be more than one task receiving from the underlying TCP connection. In some cases the right solution to this is probably to create one background task to handle receiving from the TCP connection. For example, in HTTP/2 it's expected that you can respond to pings at any time, even if all the logical channels are paused, so HTTP/2 isn't really a good example here -- the TCP receiver needs to be running constantly for as long as the connection is open. But in other protocols you might want to *only* receive on the TCP channel when there are tasks waiting for data. (IIRC something like this shows up in ZeroMQ's API, or the SSH protocol has logical channels but I don't think it has any pings.) In this case it's possible in principle to have the first task that arrives in `channel.receive_some` initiate a `tcp_transport.receive_some`, and later tasks block waiting for it, and then if the first task gets cancelled then it hands off the job of calling `tcp_transport.receive_some` to someone else.... but this is super complicated. It would be nice if there were some standard way for all the tasks in `channel.receive_some` to simply "share" a single call to `tcp_transport.receive_some` (or more realistically: `protocol_object.pump_underlying_transport`).

The interesting thing about these is that they don't fit well into trio's nursery system, BUT they're still possible to do without violating the invariants that the nursery system was created to enforce -- in particular, if the shared call crashes, then we have somewhere to report that. (Though there might be a minor issue with throwing the same exception into multiple receivers and still getting a sensible traceback -- maybe `Error.unwrap` should `copy.copy` the exception object before raising it?)

[...]

PR #303, opened 2017-08-21
[wip] Sketch of how shared tasks might work

njsmith:
See gh-266

This is surprisingly interesting and tricky.

### DIFF
--- a/trio/_shared_task.py
+++ b/trio/_shared_task.py
@@ -1,0 +1,118 @@
+__all__ = ["SharedTaskRegistry"]
+
+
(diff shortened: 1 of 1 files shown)


njsmith:
One issue with this as currently written is that it uses `outcome.unwrap()` multiple times, which of course [doesn't work](https://github.com/python-trio/outcome/issues/7).

We could mess around somehow trying to copy exceptions, but maybe it's better to do something simpler: if there's an error in a shared task, do `raise SharedTaskError from <actual exception>`. It's fine for multiple exceptions to have `__cause__` pointing to the same object, because `__cause__` doesn't get mutated.

PR #309, opened 2017-08-22
Deprecations for #136 and #284

njsmith:
Deprecates nursery.spawn and most of the nursery and task APIs.

Todo:

* [x] do a pass through the docs to get rid of now-irrelevant prose
* [x] Move now-less-central stuff into hazmat (Task, UnboundedQueue, Result?)

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -417,9 +417,9 @@ Specific style guidelines
   and the ``nowait`` version raises :exc:`trio.WouldBlock` if it would block.
 
 * The word ``monitor`` is used for APIs that involve an
-  :class:`UnboundedQueue` receiving some kind of events. (Examples:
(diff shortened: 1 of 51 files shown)


matham:
Should this be mentioned in the first issue?

njsmith:
I'm planning to make a consolidated post to the first issue once I have all these deprecations sorted out. It looks like the actual breakage in 0.2.0 will be very minimal in any case; almost all the changes are easy to put in backcompat shims with warnings.

njsmith:
Sorry, just unearthed this tab again and realized that last sentence was pretty incomprehensible. What I mean is, I realized that almost all the changes are ones where it's easy to do a proper deprecation cycle, so I'm less worried about the disruptiveness of 0.2.0 than I was when I set up issue number 1. (But of course I will still raise a flag there ahead of the release.)

Issue #314, opened 2017-09-05
PyCharm autodetection of trio.* not working


parity3:
It would be nice if PyCharm could resolve things like trio.run, and pretty much everything else.

In the below code snippet, PyCharm is unable to find run:

```
import trio

run = trio.run

```

I don't know how to get around this without either a pre-compile step in the trio package release process (which messes up a ton of other things like line number mappings to the master files for example), or doing a bunch of stubs in trio/`__init__.py` and having to maintain them.  PyCharm MAY have a way to get around this (they have done some magic with Django's ORM in detecting Model.objects.* methods but they have 'built-in' support considering Django is so popular) but I have no idea where to begin.

njsmith:
Yeah, the question here is how PyCharm is actually getting its list of module attributes for autocompletion. It'd be nice to fix this, and I can't blame PyCharm for being confused by our current extremely complex export logic, but without knowing what they need it's hard to fix :-). We're obviously not going to actually literally define all of our exports directly in `trio/__init__.py`, but almost no packages do that so surely they don't require it...

@parity3 to confirm -- do you have this problem for everything in `trio` and also everything in `trio.hazmat`?

parity3:
A number of names ARE detected; just not all of them. Here is a screenshot from a brand new project:

![trio_pycharm](https://user-images.githubusercontent.com/9721234/30051053-7cfeddba-91d5-11e7-9ac9-48ebce312dc9.png)

I hope that info can crack the case, but to me, from the outside, it seems rather arbitrary what PyCharm is picking up and what it's not.

As far as trio.hazmat, it doesn't pickup any sub-names at all.

njsmith:
Ok, so it looks like PyCharm understands:

```python
from ._version import __version__
```

and it understands

```python
from ._timeouts import *
__all__ += _timeouts.__all__
```

(I guess probably it's just picking up on the first line there and ignoring the `__all__` part.)

What it doesn't understand is:
```python
for _symbol in _core.__all__:
    _value = getattr(_core, _symbol)
    if getattr(_value, "_hazmat", False):
        setattr(hazmat, _symbol, _value)
        hazmat.__all__.append(_symbol)
    else:
        globals()[_symbol] = _value
        __all__.append(_symbol)
del _symbol, _value
```

which... you know... fair enough, I guess. But this explains the whole pattern you see for `trio` and `trio.hazmat`.

[...]

PR #316, opened 2017-09-05
Make our exports more visible to static analysis

njsmith:
In particular, this should make it so PyCharm's completion starts
working.

Fixes gh-314.

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -437,10 +437,9 @@ implementation of the core scheduling/cancellation/IO handling logic,
 and then the other ``trio.*`` modules are implemented in terms of the
 API it exposes. (If you want to see what this API looks like, then
 ``import trio; print(trio._core.__all__)``). Everything exported from
-``trio._core`` is *also* exported as part of either the ``trio`` or
-``trio.hazmat`` namespaces. (This is managed through the use of a
-``@_hazmat`` decorator that marks which items in
-``trio._core.__all__`` should go into ``trio.hazmat``.)
+``trio._core`` is *also* exported as part of the ``trio``,
+``trio.hazmat``, or ``trio.testing`` namespaces. (See their respective
+``__init__.py`` files for details; there's a test to enforce this.)
 
 Rationale: currently, trio is a new project in a novel part of the
 design space, so we don't make any stability guarantees. But the goal
--- a/docs/source/history.rst
+++ b/docs/source/history.rst
@@ -99,6 +99,8 @@ Fix ``sock.accept()`` for IPv6 sockets (https://github.com/python-trio/trio/issu
 
 deprecate most of the task and nursery APIs
 
+make our exports visible to PyCharm (314)
+
 Renames from https://github.com/python-trio/trio/issues/157
 
 Note that pypy needs 5.9+ to support deprecations properly
--- a/docs/source/history.rst
+++ b/docs/source/history.rst
@@ -81,6 +81,10 @@ task.parent_nursery
 
 Fix ``sock.accept()`` for IPv6 sockets (https://github.com/python-trio/trio/issues/164)
 
+PyCharm (and hopefully other IDEs) offer better completions for
+the :mod:`trio` and :mod:`trio.hazmat` modules
+https://github.com/python-trio/trio/issues/314
+
 
 * ``trio.socket.SocketType`` will no longer be exposed publically in
   0.3.0. Since it had no public constructor, the only thing you could
--- a/trio/__init__.py
+++ b/trio/__init__.py
@@ -12,35 +12,10 @@
 
 from ._version import __version__
 
-# PyCharm tries to statically infer the set of attributes of this module, so
-# that it can offer completions. (Other IDEs probably do similar things.)
-#
-# Specifically, it seems to need to see:
-#   __all__ = ["some_name"]
-# or
-#   __all__.append("some_name")
-# or
-#   from ... import some_name
-# or an actual definition of 'some_name'. (See
-# https://github.com/python-trio/trio/issues/314 for details.)
-#
-# _core's exports use all kinds of wacky runtime tricks to set up their
-# exports, and then they get divided between trio, trio.hazmat, and
-# trio.testing. In an attempt to make this easier to understand for static
-# analysis, we now list the re-exports directly here and in trio.hazmat and
-# trio.testing, and then we have a test to make sure that every _core export
-# does get re-exported in one of these places or another.
-__all__ = [
-    "TrioInternalError", "RunFinishedError", "WouldBlock", "Cancelled",
-    "ResourceBusyError", "MultiError", "format_exception", "run",
-    "open_nursery", "open_cancel_scope", "current_effective_deadline",
-    "STATUS_IGNORED", "current_time", "current_instruments", "current_clock",
-    "remove_instrument", "add_instrument", "current_statistics", "TaskLocal"
-]
+__all__ = []
 
-from . import _core
-
-globals().update({sym: getattr(_core, sym) for sym in __all__})
+from ._toplevel_core_reexports import *
+__all__ += _toplevel_core_reexports.__all__
 
 from ._timeouts import *
 __all__ += _timeouts.__all__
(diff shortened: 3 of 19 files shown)


njsmith:
@parity3 Can you check if this fixes your problem?

njsmith:
For testing, it should be possible to do: `pip install https://github.com/njsmith/trio/archive/issue-314.zip` (do this in a venv of course!)

parity3:
I checked out your commit. PyCharm still has an issue with resolving trio.run.
I modified `trio/__init__.py`  and added this right before the `__all__` definition, and it works:

```
TrioInternalError = _core.TrioInternalError
RunFinishedError = _core.RunFinishedError
WouldBlock = _core.WouldBlock
Cancelled = _core.Cancelled
ResourceBusyError = _core.ResourceBusyError
MultiError = _core.MultiError
format_exception = _core.format_exception
run = _core.run
open_nursery = _core.open_nursery
open_cancel_scope = _core.open_cancel_scope
current_effective_deadline = _core.current_effective_deadline
STATUS_IGNORED = _core.STATUS_IGNORED
current_time = getattr(_core, 'current_time')
current_instruments = getattr(_core, 'current_instruments')
current_clock = getattr(_core, 'current_clock')
remove_instrument = getattr(_core, 'remove_instrument')
add_instrument = getattr(_core, 'add_instrument')
current_statistics = getattr(_core, 'current_statistics')
TaskLocal = _core.TaskLocal

__all__ = [
    "TrioInternalError", "RunFinishedError", "WouldBlock", "Cancelled",
    "ResourceBusyError", "MultiError", "format_exception", "run",
    "open_nursery", "open_cancel_scope", "current_effective_deadline",
    "STATUS_IGNORED", "current_time", "current_instruments", "current_clock",
    "remove_instrument", "add_instrument", "current_statistics", "TaskLocal"
]
```

Notice the getattr calls (ie it's ugly). PyCharm won't add things to the trio root module that it thinks do not exist unfortunately.

Hope this helps.

njsmith:
Ugh. I really don't want to do that. Any idea why this seems to contradict what you [observed here](https://github.com/python-trio/trio/issues/314#issuecomment-327106705)?

parity3:
It does seem to contradict the simple experiments I did above; the `__all__` name strings were all a fiercely-colored red and did not resolve. No idea why one worked and not the other; perhaps something to do with the other activity going on in `__init__.py`.

Maybe just auto-generate a .pyi file and check it into [typeshed](https://github.com/python/typeshed)? I hear pycharm and others use that for stub generation for similar hard-to-introspect packages to trio.

njsmith:
Typeshed is for declaring the types of variables. I don't actually know if you can use it to declare the attributes of a module?

parity3:
Oh well, I don't have any more knowledge here (regarding typeshed). I have yet another question that may provide insight.. how is [numpy](https://github.com/numpy/numpy/blob/master/numpy/__init__.py) using `__all__.extend` without issue? Or is PyCharm just not resolving those as well?

njsmith:
@parity3 Those numpy imports are similar to the parts of trio that pycharm does understand ([ref](https://github.com/python-trio/trio/issues/314#issuecomment-327104611)). I'm guessing that it's actually the `from .core import *` that it detects, rather than the `__all__.extend(core.__all__)`. Or maybe it's not resolving them, I dunno; you're the one with PyCharm installed ;-)

parity3:
Maybe numpy's less notable names in `_mat.__all__` would be unresolved in PyCharm, who knows. I'm out of time researching this right now unfortunately. I really hope someone can swoop in and save the day here. At the moment I see some bad options like:

- have __init__.py reference all public names with knowledge of where they lie in their sub-packages (not DRY)
- move out all the imports that should not be exposed, directly into hazmat and just use simple import *'s to dictate what's public (remove the _public / _hazmat fn attributes that currently dictate exposure)
- move the hazmat methods into a sub-sub-module underneath trio._core.hazmat.... (ugh)

All of those are frightening workarounds; why does this have to be so challenging!?

parity3:
Here is a SLIGHTLY better workaround for `trio/__init__.py` PyCharm friendliness:
```
from . import _core
from ._core import *
to_keep = [
    "TrioInternalError", "RunFinishedError", "WouldBlock", "Cancelled",
    "ResourceBusyError", "MultiError", "format_exception", "run",
    "open_nursery", "open_cancel_scope", "current_effective_deadline",
    "STATUS_IGNORED", "current_time", "current_instruments", "current_clock",
    "remove_instrument", "add_instrument", "current_statistics", "TaskLocal"
]
any(globals().__delitem__(_) for _ in _core.__dict__ if _ not in to_keep)
del _core
del to_keep
```

... And I think we can remove all references/modifications to `__all__` in the file as well?

Also, note that hazmat.py, defining its `__all__` had no red in PyCharm. So it may be specific to an `__init__.py` package module, and less strict for non-package modules.

njsmith:
@parity3 (please don't keep poking at this if you have other things you need to be doing, it is not urgent :-))

If you do that, then does pycharm actually show the correct suggestions for `trio.*`, or does it also show all the hazmat stuff that gets deleted (but pycharm doesn't know that)?

> So it may be specific to an `__init__.py` package module, and less strict for non-package modules.

If that's true then I guess a potential workaround would be to move the trickiness into `trio/_reexport.py` and then have `__init__.py` do `from ._rexport import *`...

parity3:
You're right, PyCharm gives no error when resolving trio.checkpoint_if_cancelled when it should (runtime error happens).

Ya I really need to stop now ;)

njsmith:
@parity3 Just pushed a new commit based on the discussion in #314 -- I think it should work for you, but can you confirm?

(Or same question for anyone else with PyCharm)

parity3:
I can confirm that trio.run and many more names are able to be resolved by PyCharm.

However, `trio.hazmat.checkpoint_if_cancelled` still did not resolve. I went into hazmat.py and sure enough, the `__all__` definition had a bunch of red even though they are simple constants. I narrowed it down to the presence of __all__.remove().

I outsmarted PyCharm by replacing the code in hazmay.py starting at line 41 (`for _sym ...`):
```python
# Some hazmat symbols are platform specific
import itertools
contained_in_core = set(filter(_core.__all__.__contains__, __all__))
globals().update(zip(contained_in_core, map(getattr, itertools.repeat(_core), contained_in_core)))
any(map(__all__.remove, set(__all__).difference(contained_in_core)))
```

This achieves the same thing but PyCharm somehow does not parse through the functional programming tricks. It also worked with an eval() approach but that's even more clunky.

njsmith:
@parity3: ugh, of course. Hmm.

parity3:
Actually, even simpler.. above the for-loop, just assign a variable like so: `remove = __all__.remove`
and switch `__all__.remove(_sym)` with `remove(_sym)` and everything works fine as-is.

parity3:
For anyone else who wants to get involved here, the test code I'm using to verify proper resolution in PyCharm is:
```python
import trio, sys

# make sure ".run" isn't a bad color
run = trio.run

# make sure ".checkpoint_if_cancelled" isn't a bad color
checkpoint_if_cancelled = trio.hazmat.checkpoint_if_cancelled
ON_POSIX = 'posix' in sys.builtin_module_names
if ON_POSIX: # haven't tested this but should work
    assert 'wait_kevent' in trio.hazmat.__all__ and 'current_iocp' not in trio.hazmat.__all__
else: # yes, tested this
    assert 'wait_kevent' not in trio.hazmat.__all__ and 'current_iocp' in trio.hazmat.__all__
```

njsmith:
> `remove = __all__.remove`

ughhhh this solution is so terrible and so beautiful in different ways.

I guess if that's what we have to do then that's what we have to do. I'm really curious if this fiddling helps with other IDEs. And wish we had a better way to test this. It's seems very likely that PyCharm will someday get cleverer about these, but will it be because it becomes clever enough not to need this tricks, or clever enough to see through our tricks and break everything again? It is a mystery 👻 

I wish we could write a test for this, but I suppose that's asking. Oh well.

parity3:
Yes, confirmed the stupid hack is working now. As for the future, unknown as always.

njsmith:
Okay, let's cross our fingers and see how it goes...

Issue #321, opened 2017-09-08
Drop task management from trio.Queue


njsmith:
Specifically this means deprecating and removing `task_done` and `join`.

I've been tempted to do this for a while b/c they're complicated, very rarely useful (as in: I'm not sure they have any use at all in Trio), and deeply inelegant (why is there an independent synchronization mechanism bolted onto `Queue`?). The only reason they're there in the first place is to match stdlib `queue.Queue`. Final straw: https://bugs.python.org/issue14976#msg301173

PR #322, opened 2017-09-09
Deprecate Queue.join and Queue.task_done

njsmith:
See gh-321.


njsmith:
Codecov is probably confused because one of the MacOS builds had to be re-run hours after the rest, and in the mean time there was a merge to master that shifted line numbers around. Before that build came in codecov found no coverage changes.

Issue #317, opened 2017-09-06
Move introspection APIs into hazmat


njsmith:
Now that we have a little more experience with how trio is used, I think I want to move the API that's purely for introspection into `hazmat`; this keeps the core namespace and core reference manual smaller and focused on regular implementation code.

Affected symbols:

* `current_clock`
* `add_instrument`, `remove_instrument` (no deprecation needed b/c these are new in 0.2.0 anyway)
* `current_statistics`

PR #325, opened 2017-09-15
Move introspection APIs into trio.hazmat

njsmith:
See #317


njsmith:
Coverage changes appear to be spurious, due to #326 being merged in between the Linux/Windows builds and the MacOS builds.

Issue #274, opened 2017-08-09
Fix our test layout


njsmith:
Right now our test layout is a mess. Tests are included in the package so that users can run `pytest --pyargs trio` to make sure it works on their system, but this prevents the use of `pytest.ini` or configuring extra arguments like `--run-slow` from `conftest.py`. We have tests split across `trio/tests` and `trio/_core/tests`, and since they're not subdirectories of each other then there's configury that has to be copy-pasted into their appropriate `conftest.py` files. (And we're not even consistent about this layout, b/c the tests for `trio/testing/` are in `trio/tests/test_testing.py`.) Also these look like public modules, but they aren't really.

The major decision to make is, should we continue distributing tests inside trio? The advantage is that users can (perhaps) test the actual real installed version of trio, and they don't even have to set up some weird dev environment or anything. (Though they do need to install the test requirements, which aren't documented anywhere outside a source tree.) The main disadvantages are (a) we're swimming upstream a bit here, as noted by the issues with pytest.ini and friends, (b) it ~doubles the on-disk size of our installed package for a feature that only a tiny tiny fraction of deployments will use.

(Another commonly cited argument for separating tests out of your package is so that you can run the tests from version A against package version B. This doesn't make any sense to me, since our tests are unit tests that are tightly tied to a specific version of trio. It's not like we're implementing some generic 3rd party spec, and if we were then the compliance tests should be a standalone package anyway...)

How much extra disk space are we talking about? It's small, but less trivial than I was expecting. (I guess that 4 KiB minimum block size adds up.) On my laptop on 2017-08-09, building a wheel from trio master and then installing it in a venv, `du` reports that the `.py` and `.pyc` files take up 1.8 MB total, of which 779 KB is tests. This is not huge in the grand scheme of things, but it's not nothing, either. And in particular in these days of docker deployments, it seems likely that some of our users will be obsessed with making their docker containers smaller and get frustrated and angry about a megabyte of "waste". Maybe this makes sense, maybe not, but it doesn't really matter: making users frustrated and angry is a good thing to avoid no matter why it happens :-).

OTOH twisted does ship their tests inside their source tree, and twisted is massive: 27 MB installed, of which 15 MB is tests. And tornado and asyncio seem to as well. (Though distributors do often take special measures to delete CPython's test suite, which would include asyncio's test suite.)

If we do keep the tests inside `trio/`, then they should probably all move and be consolidated under `trio/_tests/`.

[...]

PR #330, opened 2017-09-26
Fix --run-slow test option

smurfix:
Symptom:
```======================================================================
ERROR: trio._core.tests.test_ki (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: trio._core.tests.test_ki
  File "/usr/lib/python3.6/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/smurf/src/trio/.pybuild/pythonX.Y_3.6/build/trio/_core/tests/test_ki.py", line 15, in <module>
    from .tutil import slow
  File "/home/smurf/src/trio/.pybuild/pythonX.Y_3.6/build/trio/_core/tests/tutil.py", line 9, in <module>
    not pytest.config.getoption("--run-slow", True),
AttributeError: module 'pytest' has no attribute 'config'
```
Fix:
Use the pattern described in
https://docs.pytest.org/en/latest/example/simple.html

### DIFF
--- a/trio/_core/tests/test_ki.py
+++ b/trio/_core/tests/test_ki.py
@@ -12,7 +12,6 @@ from ... import _core
 from ...testing import wait_all_tasks_blocked
 from ..._util import acontextmanager, signal_raise
 from ..._timeouts import sleep
-from .tutil import slow
 
 
 def ki_self():
@@ -446,7 +445,7 @@ def test_ki_is_good_neighbor():
 #   pthread_kill(pthread_self, SIGINT)
 # in the child thread, to make sure signals in non-main threads also wake up
 # the main loop... but currently that test would fail (see gh-109 again).
-@slow
+@pytest.mark.slow
 def test_ki_wakes_us_up():
     assert threading.current_thread() == threading.main_thread()
 
--- a/trio/_core/tests/test_multierror.py
+++ b/trio/_core/tests/test_multierror.py
@@ -7,8 +7,6 @@ import re
 from pathlib import Path
 import subprocess
 
-from .tutil import slow
-
 from .._multierror import (
     MultiError,
     format_exception,
@@ -509,21 +507,21 @@ else:
 need_ipython = pytest.mark.skipif(not have_ipython, reason="need IPython")
 
 
-@slow
+@pytest.mark.slow
 @need_ipython
 def test_ipython_exc_handler():
     completed = run_script("simple_excepthook.py", use_ipython=True)
     check_simple_excepthook(completed)
 
 
-@slow
+@pytest.mark.slow
 @need_ipython
 def test_ipython_imported_but_unused():
     completed = run_script("simple_excepthook_IPython.py")
     check_simple_excepthook(completed)
 
 
-@slow
+@pytest.mark.slow
 @need_ipython
 def test_ipython_custom_exc_handler():
     # Check we get a nice warning (but only one!) if the user is using IPython
--- a/trio/_core/tests/tutil.py
+++ b/trio/_core/tests/tutil.py
@@ -4,13 +4,6 @@ import pytest
 
 import gc
 
-# See trio/tests/conftest.py for the other half of this
-slow = pytest.mark.skipif(
-    not pytest.config.getoption("--run-slow", True),
-    reason="use --run-slow to run slow tests",
-)
-
-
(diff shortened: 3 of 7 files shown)


njsmith:
Do you know why you're getting this error? I'm confused by the traceback
showing unittest/loader.py given that we don't use unittest...

njsmith:
The travis failures are #332 and #334, nothing to do with this PR (though we do need to get them sorted out before we can merge anything). In any case this PR is waiting for me to understand better what the underlying problem is :-)

smurfix:
The trigger was that I was playing naïve and started "setup.py test".

That doesn't work even after applying this patch (there's a way to teach setup.py to use pytest instead …) for some interesting reasons; however, I've seen variations of this sort of problem with pytest in other projects – which could easily be fixed by switching to the canonical way of handling this usecase. Thus I'd suggest applying this change.

njsmith:
Okay, finally had a little more time to look at this. Sorry for the wait, and thanks for taking the time to put this together!

The details are, like everything involving testing and especially pytest, bizarrely complicated...

First, the patch currently causes a regression: a hook in `trio/tests/conftest.py` does *not* apply to tests in `trio/_core/tests/`. This means that slow tests in `trio/_core/tests` will be executed unconditionally, regardless of whether `--run-slow` is passed. This thing with the two `conftest.py` files is a preexisting mess that we should fix anyway (see #274), but it's still a regression here and now that would need to be sorted out somehow.

I'm also uncertain whether this works if you use `pytest --pyargs trio` to test an installed trio. The problem is that in this mode, pytest doesn't find `conftest.py` until after it's parsed the command line, so it doesn't run `pytest_addoption` hooks (I think?), so `config.getoption("--run-slow")` might crash due to not being a recognized option. Even if it doesn't crash, it probably doesn't return `True` to cause slow tests to be run, which is what we do now and maybe superior when running in this degraded mode where it's impossible to choose whether to run them or not?

Before we get too far into these details though there's the question of whether this is useful.

I think we could fix that specific error message by just adding an `import pytest.config` somewhere?

Also I pretty much 100% don't care about supporting `setup.py test`. It's a "nice idea" that never took off, it will never become standard in practice because `setup.py` itself is starting the long-term process of being deprecated (see [PEP 517](https://www.python.org/dev/peps/pep-0517/)), and fundamentally the only reason it's useful is because it potentially gives users a standard one-button way to run the tests without having to check the docs... but it just can't accomplish this goal, because you still need to check the docs to find out what packages you need to install into the venv that you use to run the tests. I think having better docs about how to do things like run the tests is useful (I have a local branch here with contributing docs that I'm going to try to polish up and post soon), and I think it might be useful to have a custom script that actually solves the problem of running the tests with one button press (so the docs become "run `python tools/test-trio.py`" and that automatically sets up a venv if necessary etc.), but `setup.py test` is irrelevant to both of these.

Well, I guess making `setup.py test` print "run `tools/test-trio.py`" might be useful in case anyone else decides to play naive :-)

OTOH, this change is pretty harmless (if the specific issues mentioned above are sorted out) and maybe using the mark mechanism is independently nice in some rare corner cases like doing `-m slow`? So in summary, I'm not necessarily opposed, but it's not feeling like an obvious win either. What do you think? Is it still valuable if we take `setup.py test` off the table? And in particular is it sufficiently valuable to justify figuring out the annoying issues I described?

njsmith:
Closing/re-opening to redo CI now that #336 is merged.

smurfix:
OK, alternate solution: [Writing Plug-ins](https://docs.pytest.org/en/latest/writing_plugins.html) says:

> The config object is passed around on many internal objects via the .config attribute or can be retrieved as the pytestconfig fixture or accessed via (deprecated) pytest.config.

smurfix:
… and getting setup.py to run pytest when calling `setup.py test` is possible too, I'll fix up a patch to do that.

One of the reasons I want that is that some build systems (like Debian's) auto-call `setup.py test` when building and installing a Python module. Another is sheer laziness on my part. ;-)

njsmith:
This has stalled out and it's not clear to me whether there's anything here, so I'm going to close and we can re-open later if appropriate.

Issue #229, opened 2017-06-18
In some obscure cases, StopIteration exceptions can be turned into RuntimeErrors when passing through nurseries


njsmith:
There are some very obscure edge cases where using async generators does the wrong thing. For example, this code

```python
import trio

async def fail():
    raise ValueError

async def main():
    async with trio.open_nursery() as nursery:
        nursery.spawn(fail)
        raise StopIteration

trio.run(main)
```

Ought to raise a `MultiError([StopIteration(), ValueError()])`, but instead it raises a `MultiError([RuntimeError(), ValueError()])`. This is because you can't throw a `StopIteration` into an `async_generator`-based async generator (for details [see here](https://github.com/python-trio/trio/pull/228#pullrequestreview-44714634)) – it gets wrapped into a `RuntimeError`. If we let that `RuntimeError` propagate, then `acontextmanager` could turn it back, but here it gets wrapped into a `MultiError` instead so the magic doesn't work.

I guess we could make `acontextmanager` smart enough to look inside `MultiError`s? Or stop using `acontextmanager` and write this context manager directly? Or just not worry about it until we can drop 3.5 support and switch to using native async generators, which don't have this problem? It's *very* unusual to have free-floating `StopIteration` exceptions wandering around like this, and that's the only case that's affected. But I figured we should record the problem in any case.

PR #340, opened 2017-10-19
Let the Nursery context-manage itself

smurfix:
I was curious whether the commented-out NurseryManager class would actually work, so I turned it on ;-)
I had to insert one conditional because of an impedance mismatch between context managing and cancel handling. All tests pass.

On second thought (second patch), a Nursery can be its own context manager. This can be improved further (third patch) by deferring the initialization to `__aenter__()`, noting that a Nursery is never used outside of a context manager.

This change speeds up the full test run by a bit more than 1% on my system (tested with 3.5 and 3.6).

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -288,38 +288,33 @@ class _TaskStatus:
         self._old_nursery.cancel_scope.cancel()
 
 
-def open_nursery():
-    return Nursery()
-
-
 class Nursery:
-    _scope_manager = None
+    def __init__(self):
+        self._scope_manager = None
 
-    def __init__(self, parent_task = None, cancel_scope = None):
-        # the parent task -- only used for introspection, to implement
-        # task.parent_task
-        if parent_task is None:
-            parent_task = current_task()
-        if cancel_scope is None:
-            self._scope_manager = open_cancel_scope()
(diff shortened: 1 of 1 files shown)


njsmith:
I should note here: the nursery implementation is going to get majorly revisited in 0.3.0 (#136), so I've been putting this off to look at as part of that.

njsmith:
First pass of simplified nurseries is in #375.

Do you remember why you had to add the `if isinstance(exc, Cancelled): return True` bit to your `Nursery.__aexit__` method? I don't see how it can be correct (doesn't it make nurseries swallow all `Cancelled` exceptions?), but I assume you ran into trouble without it...

smurfix:
I'll test my code with the current master branch and report back.

njsmith:
Thanks!

kyrias:
Any updates on this? :)

smurfix:
I'll have a look soon(ish).

njsmith:
We should probably stop using `@asynccontextmanager` in the nursery implementation (#229, #393). But this branch is way out of date, discussion has stalled, and I think it goes a bit to far (I like that the actual `Nursery` object is only returned from `__aenter__`, not from `open_nursery`). I'm going to close this for now, and we can always re-open it or steal bits from it later if we want to.

Issue #306, opened 2017-08-22
When an instrument raises an error, report it via logging module instead of directly to stderr


njsmith:
The stdlib logging module is not so great, but it's better than dumping to stderr directly. By default it prints to stderr, and any serious logging system has a way to capture stuff from the stdlib loggers. So it'd be a minor improvement, and match what `trio.serve_listeners` does.

PR #343, opened 2017-11-05
Report instrument errors via logging

nmalaguti:
See gh-306

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -10,6 +10,7 @@ import select
 import sys
 from math import inf
 import functools
+import logging
 
 import attr
 from sortedcontainers import SortedDict
@@ -60,6 +61,9 @@ else:  # pragma: no cover
     raise NotImplementedError("unsupported platform")
 
 _r = random.Random()
+
+# Used to log exceptions in instruments
+INSTRUMENT_LOGGER = logging.getLogger("trio.abc.Instrument")
 
 
 @attr.s(frozen=True)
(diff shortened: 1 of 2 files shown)


trio/_core/_run.py
```diff
@@ -61,6 +62,9 @@
 
 _r = random.Random()
 
+# Used to log exceptions in instruments
+LOGGER = logging.getLogger("trio.Runner.instrument")
```
njsmith: Minor nitpick: can we make this:
```
INSTRUMENT_LOGGER = logging.getLogger("trio.abc.Instrument")
```
Rationale: 
* More specific name since there's a ton of stuff in this file, some of which might potentially want to use logging some day, and most of it isn't related to instruments.
* `Runner` is a private class, so I'd rather not expose it as part of our logging names -- the string we pass here is effectively public API because people can filter on it etc. `trio.abc.Instrument` is a little odd looking -- let me know if you have a better idea -- but it's a public name and relevant so seems reasonable enough?

njsmith:
(Also, I think codecov is super-confused because I merged #345 and then re-ran the travis tests, so there's some line mismatches between the Travis run and the Appveyor run. It should sort itself out if you push any more changes.)

trio/_core/_run.py
nmalaguti: I wasn't sure about giving the `LOGGER` a more specific name, that works great.

I also wasn't familiar enough with the API surface to pick a good logger name. `trio.abc.Instrument` works great.

njsmith:
Excellent, thanks!

PR #345, opened 2017-11-05
Skip tests that need IPv6 if it is not available

nmalaguti:
IPv6 is not available on all travis images. This will skip IPv6 tests when IPv6 is not available.

### DIFF
--- a/trio/_core/tests/tutil.py
+++ b/trio/_core/tests/tutil.py
@@ -1,4 +1,5 @@
 # Utilities for testing
+import socket as stdlib_socket
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


njsmith:
The codecov complaints here are baffling -- it looks like it didn't get coverage for pypy 5.8? But the travis report shows it running. Anyway, it's clearly not caused by this PR (and even if it were, we should be switching to pypy 5.9 and deleting the 5.8 hacks anyway), so gonna merge.

For future spelunkers: this is required because Travis just broke IPv6 without warning: https://github.com/travis-ci/travis-ci/issues/8361

PR #347, opened 2017-11-06
Monkey patch traceback.TracebackException to support MultiError

nmalaguti:
This monkey patches `traceback.TracebackException` on import. It overrides `TracebackException.__init__()` to extract any inner exceptions from `MultiError`s and overrides `TracebackException.format()` to include the details of the embedded exceptions.

This deprecates `trio.format_exception` as it is no longer needed and is now an alias for `traceback.format_exception`.

Documentation is included detailing some examples where this would be useful e.g. logging.

See gh-305.

### DIFF
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -7,6 +7,8 @@ from contextlib import contextmanager
 
 import attr
 
+from .._deprecate import deprecated
+
 __all__ = ["MultiError", "format_exception"]
 
 ################################################################
@@ -341,12 +343,14 @@ def concat_tb(head, tail):
 
 ################################################################
 # MultiError traceback formatting
+#
+# What follows is terrible, terrible monkey patching of
+# traceback.TracebackException to add support for handling
+# MultiErrors
 ################################################################
 
 
-# format_exception's semantics for limit= are odd: they apply separately to
-# each traceback. I'm not sure how much sense this makes, but we copy it
-# anyway.
+@deprecated("0.2.0", issue=347, instead="traceback.format_exception")
 def format_exception(etype, value, tb, *, limit=None, chain=True):
     """Like :func:`traceback.format_exception`, but with special support for
     printing :exc:`MultiError` objects.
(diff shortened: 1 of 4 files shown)


trio/_core/_multierror.py
```diff
@@ -6,6 +6,9 @@
 from contextlib import contextmanager
 
 import attr
+import logging
```
nmalaguti: I don't think this import is used at all anymore. I should remove it.

trio/_core/_multierror.py
```diff
@@ -341,12 +344,14 @@ def concat_tb(head, tail):
...
+# What follows is terrible, terrible monkey patching of
+# traceback.TracebackException to add support for handling
+# MultiErrors
 ################################################################
 
 
-# format_exception's semantics for limit= are odd: they apply separately to
-# each traceback. I'm not sure how much sense this makes, but we copy it
-# anyway.
+@deprecated("0.2.0", issue=305, instead="traceback.format_exception")
```
nmalaguti: I wasn't sure what issue number to use for this. I used #305 but it doesn't really address the deprecation. Should I make an issue specifically for this?

trio/_core/_multierror.py
```diff
@@ -355,60 +360,72 @@ def format_exception(etype, value, tb, *, limit=None, chain=True):
...
-            chunks += [
-                "\nThe above exception was the direct cause of the "
-                "following exception:\n\n",
-            ]
-        elif value.__context__ is not None and not value.__suppress_context__:
-            v = value.__context__
-            chunks += _format_exception_multi(
-                seen, type(v), v, v.__traceback__, limit=limit, chain=True
+    if isinstance(exc_value, MultiError):
+        embedded = []
```
nmalaguti: There is no explicit `seen` behavior here anymore to log duplicate exceptions. The behavior of `TracebackException` is just to omit it if it has been seen.

Is that okay?

trio/_core/_multierror.py
njsmith: You can also use the PR number if it makes more sense - which in this case it probably does, I agree.

trio/_core/_multierror.py
njsmith: 👍

trio/_core/_multierror.py
njsmith: (This works because github doesn't fully distinguish between PRs and issues – they use the same number space, and the URL for "issue 347" redirects to the PR page if 347 is a PR.)

nmalaguti:
> I guess one approach would be to have the capture code only skip loops, but not throw away "horizontal" duplicates. (Something like, when we recurse we pass through a seen set, but we make a separate copy of it for each entry in a MultiError.)

Just putting down some thoughts to make sure I understand what each part of the multierror actually represents.

**MultiError traceback**: This is the stack from where the MultiError was created to where it was handled.

**MultiError context or cause**: This is any exception that was being handled when the MultiError was created (like if there was a nursery in an except block).

By using the stdlib traceback formatting code, this will work as desired and there should be no duplicates that need to be explicitly printed.

**MultiError sub-exceptions**: Each want to be printed separately. If there are duplicates, especially in different parts of the "tree", we probably want to mention that instead of silently omitting them.

This matches with what you said about "horizontal" duplicates and using a copy of our `_seen` at that point for each. I say a copy rather than an empty seen on the off chance that somehow the context of the MultiError we are handling appears as a cause or context of the sub exception, in which case it would be omitted.

In this way, all sub-exceptions will be printed with a reasonable stack of causes, and there may be duplicates in those stacks. In order to re-use the existing `TracebackException::format` code, we won't be able to mention that the duplicates are duplicates, but as long as the stacks are there it should be okay.

nmalaguti:
@njsmith The only open question left is whether it is important to explicitly mention that we have printed a duplicate exception or if simply printing it is sufficient.

If we want to be explicit that it is a duplicate, we'll need to copy the code from `TracebackException::__init__` and print "duplicate" when an exception is in `_seen` instead of omitting it. I'm not a huge fan of copying the code, but it isn't a lot. I'm also not sure how important it is to **know** that the exception is the same one, rather than just looks the same.

njsmith:
Sorry for leaving you hanging on this! I think this is looking good, and if we decide we want to tweak it more, we can always do that later :-). Thanks!

PR #349, opened 2017-11-13
Make measure_backlog more robust against weird errors

njsmith:
@nmalaguti is reporting that on a particular VPS they have access to,
sometimes, the test_open_tcp_listeners_backlog test is blowing up with
EMFILE and breaking other tests. It seems to involve the kernel
switching to using syn cookies, and then suddenly infinitely many
connection attempts can appear to succeed.

A real fix would be to make the test robust against syncookies, but
I'm not yet sure how to do that. In the mean time this at least should
make the test more robustly clean up after itself, so that this causes
a well-defined failure without shrapnel flying out to hit other tests.

### DIFF
--- a/trio/tests/test_highlevel_open_tcp_listeners.py
+++ b/trio/tests/test_highlevel_open_tcp_listeners.py
@@ -55,22 +55,34 @@ async def test_open_tcp_listeners_specific_port_specific_host():
 
 # Warning: this sleeps, and needs to use a real sleep -- MockClock won't
 # work.
-async def measure_backlog(listener):
+#
+# Also, this measurement technique often works, but not always: sometimes SYN
+# cookies get triggered, and then the backlog measured this way effectively
+# becomes infinite. (In particular, this has been observed happening on
+# Travis-CI.) To avoid this blowing up and eating all FDs / ephemeral ports,
+# we put an upper limit on the number of connections we attempt, and if we hit
+# it then we return the magic string "lots". Then
+# test_open_tcp_listeners_backlog uses a special path to handle this, treating
+# it as a success -- but at least we'll see in coverage if none of our test
+# runs are actually running the test properly.
+async def measure_backlog(listener, limit):
     client_streams = []
-    while True:
-        # Generally the response to the listen buffer being full is that the
-        # SYN gets dropped, and the client retries after 1 second. So we
-        # assume that any connect() call to localhost that takes >0.5 seconds
-        # indicates a dropped SYN.
-        with trio.move_on_after(0.5) as cancel_scope:
-            client_stream = await open_stream_to_socket_listener(listener)
-        if cancel_scope.cancelled_caught:
+    try:
+        while True:
+            # Generally the response to the listen buffer being full is that
+            # the SYN gets dropped, and the client retries after 1 second. So
+            # we assume that any connect() call to localhost that takes >0.5
+            # seconds indicates a dropped SYN.
+            with trio.move_on_after(0.5) as cancel_scope:
+                client_stream = await open_stream_to_socket_listener(listener)
+                client_streams.append(client_stream)
+            if cancel_scope.cancelled_caught:
+                break
+            if len(client_streams) >= limit:  # pragma: no cover
+                return "lots"
+    finally:
+        for client_stream in client_streams:
             await client_stream.aclose()
-            break
-        client_streams.append(client_stream)
-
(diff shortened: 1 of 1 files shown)


nmalaguti:
Just like on travis, this now causes my machine to only fail this test (but the test still fails).

```
================================================================= FAILURES ================================================================$
_____________________________________________________ test_open_tcp_listeners_backlog _____________________________________________________$

    @slow
    async def test_open_tcp_listeners_backlog():
        # Operating systems don't necessarily use the exact backlog you pass
        async def check_backlog(nominal, required_min, required_max):
            listeners = await open_tcp_listeners(0, backlog=nominal)
            actual = await measure_backlog(listeners[0])
            for listener in listeners:
                await listener.aclose()
            print("nominal ", nominal, "actual", actual)
            assert required_min <= actual <= required_max

>       await check_backlog(nominal=1, required_min=1, required_max=10)

trio/tests/test_highlevel_open_tcp_listeners.py:89:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
trio/tests/test_highlevel_open_tcp_listeners.py:83: in check_backlog
    actual = await measure_backlog(listeners[0])
trio/tests/test_highlevel_open_tcp_listeners.py:67: in measure_backlog
    client_stream = await open_stream_to_socket_listener(listener)
trio/testing/_network.py:34: in open_stream_to_socket_listener
    sock = tsocket.socket(family=family)
trio/_socket.py:342: in socket
    stdlib_socket = _stdlib_socket.socket(family, type, proto, fileno)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <socket.socket fd=-1, family=AddressFamily.AF_UNSPEC, type=0, proto=0>, family = <AddressFamily.AF_INET: 2>
type = <SocketKind.SOCK_STREAM: 1>, proto = 0, fileno = None

    def __init__(self, family=AF_INET, type=SOCK_STREAM, proto=0, fileno=None):
        # For user code address family and type values are IntEnum members, but
        # for the underlying _socket.socket they're just integers. The
        # constructor of _socket.socket converts the given argument to an
        # integer automatically.
>       _socket.socket.__init__(self, family, type, proto, fileno)
E       OSError: [Errno 24] Too many open files

/usr/lib/python3.6/socket.py:144: OSError
============================================= 1 failed, 397 passed, 2 skipped in 11.24 seconds =============================================
sys:1: ResourceWarning: unclosed <socket.socket fd=17, family=AddressFamily.AF_INET, type=2049, proto=6, laddr=('0.0.0.0', 43479)>
sys:1: ResourceWarning: unclosed <socket.socket fd=18, family=AddressFamily.AF_INET6, type=2049, proto=6, laddr=('::', 40433, 0, 0)>
```

From the kernel logs:

```
$ sudo dmesg | tail -1
[ 5071.268199] TCP: request_sock_TCP: Possible SYN flooding on port 43479. Sending cookies.  Check SNMP counters.
```

njsmith:
I should probably leave a link here to the chat where we debugged a bunch of this, for any future archaeologists: https://gitter.im/python-trio/general?at=5a08e2b1505b630c05d53b89

njsmith:
Okay, since I can't think of any way to run this test once syn cookies come into play, I've updated it further to put an upper bound on how many connection attempts are made, and if we hit that we consider the test to have passed. (But, sneakily, we don't cover all the lines this way, so hopefully we'll get some warning if we ever end up in a situation where *all* our test runners are giving up and effectively skipping the test.)

@nmalaguti What do you think?

nmalaguti:
It would be nice to someday have a way of detecting that the backlog was triggered rather than using a timeout (to avoid syn cookie weirdness), but this now passes on my linux box. This seems like a fair hack to avoid file descriptor/ephemeral port exhaustion and "infinite" backlog.

LGTM 👍.

njsmith:
👏

PR #353, opened 2017-11-28
Rename STATUS_IGNORED to TASK_STATUS_IGNORED

njsmith:
The asymmetry in task_status=trio.STATUS_IGNORED is bugging me too much. I
gotta change it.

What to change it to? task_status=trio.TASK_STATUS_IGNORED is kind of
long -- 36 characters -- and likely to make 80-character lines wrap.
But the old one was 31 characters, which is already likely to make
80-character lines wrap, and I haven't had any better ideas.

I don't want to make it status=STATUS_IGNORED, because then we're
claiming the kwarg "status" in every Trio server function, which seems
just a little too generic. task_status is more unique and I think
better for attaching this special convention to.

I thought about task_status=NO_TASK_STATUS, but that feels weird. It's
not that the task has no status, it's that the caller is ignoring that
status. I guess we could switch to something like
status_reporter=trio.NO_STATUS_REPORTER but that's even *longer* (39
characters), and not really any clearer.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -924,13 +924,13 @@ Nursery objects provide the following interface:
 
       The conventional way to define ``async_fn`` is like::
 
-         async def async_fn(arg1, arg2, *, task_status=trio.STATUS_IGNORED):
+         async def async_fn(arg1, arg2, *, task_status=trio.TASK_STATUS_IGNORED):
              ...
              task_status.started()
              ...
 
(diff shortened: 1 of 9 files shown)


njsmith:
I guess I'll leave this open for a little while in case anyone comes up with a better suggestion.

njsmith:
There's some more discussion in chat starting here: https://gitter.im/python-trio/general?at=5a1ceed4e5ed52e26c98480f

PR #359, opened 2017-11-30
Give cancel scopes a nicer opaque repr

njsmith:
Previously the cancel scope repr looked like:

```
CancelScope(_tasks={<Task '__main__.main' at 0x7f7f6af99a90>}, _effective_deadline=228508.69968805768, _deadline=228508.69968805768, _shield=False, cancel_called=False, cancelled_caught=False)
```

This reveals a bunch of internal details. Let's hide those.

Maybe an even nicer version would print some details about the public
attributes like deadline, shield, cancel_called? I hit this while
preparing a talk for Friday though so I want to get this in now :-)

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -88,7 +88,7 @@ class SystemClock:
 ################################################################
 
 
-@attr.s(cmp=False, hash=False)
+@attr.s(cmp=False, hash=False, repr=False)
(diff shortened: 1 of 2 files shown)


njsmith:
Codecov is confused, see https://github.com/python-trio/trio/pull/361#issuecomment-348673769

PR #361, opened 2017-12-01
Attempt to fix a flaky test on Windows

njsmith:
This failure:
  https://ci.appveyor.com/project/njsmith/trio/build/1.0.768/job/3lbdyxl63q3h9s21
appears to be due to some floating point wackiness causing a >=
assertion to not *quite* succeed. See comment for more details.


njsmith:
Codecov complaints here are due to a bug in codecov failing to combine the windows and non-windows runs. I emailed them about it and they say they've added a workaround. We'll see, but in the mean time I think this is OK to go in :-)

PR #362, opened 2017-12-05
Add contributing doc, and switch to WeAllJS code of conduct

njsmith:
We obviously need a contributing doc, and the WeAllJS code of conduct
is excellent. These are together because part of the idea here is to
move these into the docs, and have that act as a single standard
location that all the python-trio projects can point to, instead of
copying them into every repo.

I'd particularly like to get feedback on the contributing doc. It's
just a first version and we can improve it as we go, but it's always
good to get more eyes on this kind of thing.

CC: @brettcannon, @smurfix, @kyrias


njsmith:
Here's the formatted version of the contributing doc: https://github.com/njsmith/trio/blob/contributing-and-coc/docs/source/contributing.rst

docs/source/code-of-conduct.rst
```diff
@@ -0,0 +1,274 @@
...
+
+If you see a Code of Conduct violation, follow these steps:
+
+1. Let the person know that what they did is not appropriate and ask
+   them to stop and/or edit their message(s) or commits.
+2. That person should immediately stop the behavior and correct the
+   issue.
+3. If this doesn’t happen, or if you're uncomfortable speaking up,
+   :ref:`contact the maintainers <coc-contacting-maintainers>`.
+4. As soon as available, a maintainer will look into the issue, and take
```
brettcannon: "as soon as possible"?

docs/source/code-of-conduct.rst
```diff
@@ -0,0 +1,274 @@
...
+take for further enforcement, as needed:
+
+1. Repeat the request to stop.
+2. If the person doubles down, they will have offending messages removed
+   or edited by a maintainers given an official warning. The PR or Issue
+   may be locked.
+3. If the behavior continues or is repeated later, the person will be
+   blocked from participating for 24 hours.
+4. If the behavior continues or is repeated after the temporary block, a
+   long-term (6-12mo) ban will be used.
```
brettcannon: I notice there's no contingency for a permanent ban. Is that on purpose?

docs/source/contributing.rst
```diff
@@ -0,0 +1,482 @@
...
+  :mod:`trio.testing` provides powerful tools to help control
+  ordering, like :func:`trio.testing.wait_all_tasks_blocked`,
+  :class:`trio.testing.Sequencer`, and :class:`trio.testing.MockClock`
+  (usually used as a fixture: ``async def
+  test_whatever(autojump_clock): ...``). And if you need more tools
+  than this then we should add them.
+
+* (Trio package only) Slow tests – anything that takes more than about
+  0.25 seconds – should be marked with ``@slow``. This makes it so they
+  only run if you do ``pytest trio --run-slow``. Our CI scripts do pass
```
brettcannon: "do pass run" -> "do run"

docs/source/code-of-conduct.rst
njsmith: Oh, hmm, interesting. 99% of the text here is straight from the WeAllJS code of conduct (via [weallbehave](https://github.com/WeAllJS/weallbehave)); I hadn't noticed this particular detail. Looking at it in context though, I think I'm OK with it – one of the things I like about this CoC is that it has clear and straightforward guidance for "regular" grumpiness and jerkitude, while also being clear that if someone's trying to game the system or acting totally beyond the pale then That Will Be Dealt With. This part is in the former section. Good catch though, and if you have more thoughts I'd be interested to hear, I'm not super confident about all this.

brettcannon: At the very least I would have a step 5 say that if the behaviour continues then the ban will be permanent. Considering at that point they have misbehaved 3 separate times over the span of at least six months that's enough time to learn how to behave online appropriately. And later on there is a mention that steps may be skipped as deemed necessary, so that covers egregious behaviour that warrants skipping the long-term ban and going straight to a permanent one.

njsmith: Fair enough! Done.

njsmith:
Going to try closing/reopening this in hopes that it will reset the CI to pick up #367

njsmith:
On further thought, I think I'm going to put this in now so it can ship with 0.2.0, without delaying 0.2.0 further. But I'm still very interested in any feedback, and these are living documents so we can update them whenever if people have thoughts.

PR #373, opened 2017-12-14
Copy note about cancellation from sendall to send_all

bluetech:
This note is important also for the high-level API. Assuming people will
skip the low-level documentation, it seems appropriate to repeat the
note.


njsmith:
Excellent point!

PR #375, opened 2017-12-19
Simplified nurseries

njsmith:
- Parenting is no longer a full time job
- This commit also removes a bunch of deprecated functions

Fixes: gh-136

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -826,9 +826,9 @@ finishes first::
        async with trio.open_nursery() as nursery:
            for async_fn in async_fns:
                nursery.start_soon(jockey, async_fn)
-               winner = await q.get()
-               nursery.cancel_scope.cancel()
-               return winner
+           winner = await q.get()
+           nursery.cancel_scope.cancel()
+           return winner
 
 This works by starting a set of tasks which each try to run their
 function, and then report back the value it returns. The main task
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -660,28 +660,29 @@ This means that tasks form a tree: when you call :func:`run`, then
 this creates an initial task, and all your other tasks will be
 children, grandchildren, etc. of the initial task.
 
-The crucial thing about this setup is that when execution reaches the
-end of the ``async with`` block, then the nursery cleanup code
-runs. The nursery cleanup code does the following things:
+Essentially, the body of the ``async with`` block acts like an initial
+task that's running inside the nursery, and then each call to
+``nursery.start_soon`` adds another task that runs in parallel. Two
+crucial things to keep in mind:
 
-* If the body of the ``async with`` block raised an exception, then it
-  cancels all remaining child tasks and saves the exception.
+* If any task inside the nursery finishes with an unhandled exception,
+  then the nursery immediately cancels all the tasks inside the
+  nursery.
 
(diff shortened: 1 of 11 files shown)


docs/source/reference-core.rst
```diff
@@ -831,34 +814,33 @@ For example, here's a function that takes a list of functions, runs
...
-their function, report back, and then cancel all the rest. Eventually
-one suceeds, all the tasks are cancelled and exit, and then our
-nursery exits and we return the winning value. And if one or more of
-them raises an unhandled exception then Trio's normal handling kicks
-in: it cancels the others and then propagates the exception. If you
-wanted different behavior, you could do that by adding a ``try`` block
-to the ``racecar`` function to catch exceptions and handle them
-however you like.
+               nursery.start_soon(jockey, async_fn)
+               winner = await q.get()
```
bluetech: IIUC this should run after the loop, not inside it.

njsmith: Nice catch! Fixed.

njsmith:
Closing/reopening to restart Travis build affected by #379

bluetech:
The rationale for the previous behavior (parenting is full-time task) was:

>  The difference between these is that in the first example, if ``walk`` crashes, the parent is off distracted chewing gum, and won't notice. In the second example, the parent is watching both children, and will notice and respond appropriately if anything happens.

This made sense to me.

Now if we look at the example with the new approach:

```py
async def server(handler):
        while True:
            conn = await get_new_connection()
            nursery.start_soon(handler, conn)
```

If `handler` raises, what happens? IIUC, the content of the `async with` is now a "pseudo-task", and it will get cancelled, but then the flow continues like normal after the `async with`.

If I want to perform cancellation cleanup for the "pseudo-task", will the following work?

```py
async def server(handler):
        while True:
            try:
                conn = await get_new_connection()
            except trio.Cancelled:
                # some cleanup...
                raise
            nursery.start_soon(handler, conn)
     print('This will be reached')
```

bluetech:
(I've made some edits to the question above, if you're reading from mail).

njsmith:
> If I want to perform cancellation cleanup for the "pseudo-task", will the following work?

Yeah, that looks correct. Of course in many cases a `finally` block might be simpler :-). In particular, one thing to watch out for: if the code inside being cancelled was more complicated than `sleep(inf)` and happened to use nurseries internally (something you normally wouldn't need to care about), then when it's cancelled you might get a `MultiError` containing multiple `Cancelled`, and `except Cancelled:` won't catch a `MultiError`, even though semantically they'd be the same in this case.

I'm not sure if there's something better we can be doing here. `MultiError` is the place where we're really fighting the language, and everything has some trade-offs. For example, we could collapse together equivalent `Cancelled` exceptions, but then you'd lose proper tracebacks for them.

njsmith:
Guess this has marinated long enough :-)

mentalisttraceur:
Is there some essay on what "parenting is a full-time task" and "parenting is not a full-time task" *mean*, including rationale for each approach?

I haven't recursed through the linked issues or looked at the code changes themselves, but from just this issue discussion and my beginner-level familiarity with *using* trio, I have no idea what exactly is being expressed here (besides the partial but incomplete narrowing down of possible meanings that I can infer from the above example).

mentalisttraceur:
Oh okay the diff contains documentation changes that seem to clarify it a bit. You can just ignore my last comment.

Issue #379, opened 2017-12-21
Intermittent CI failures in test_local.py on PyPy


njsmith:
As originally tracked in #200, we've been seeing an intermittent freeze in `test_local.py` on PyPy:

* 2017-06-16: happened on PyPy nightly but didn't save logs; [failed to reproduce locally](https://github.com/python-trio/trio/issues/200#issuecomment-309187983)

* 2017-09-07 on PyPy 5.8 in `test_run_local_simultaneous_runs`: https://travis-ci.org/python-trio/trio/jobs/272763511

* 2017-12-20 on PyPy nightly in `test_run_local_simultaneous_runs`: https://travis-ci.org/python-trio/trio/jobs/319497598

The test uses threads + blocking queue operations to control sequencing, so it's the sort of thing where you're not shocked to see some weird little freeze, but after staring at it for a while I can't see any bug in the test itself.

njsmith:
Here's another one, on PyPy 5.8 in `test_run_local_simultaneous_runs`: https://travis-ci.org/python-trio/trio/jobs/341367598

I'll paste the key bits below in case something happens to the Travis log:

```
../testenv/site-packages/trio/_core/tests/test_local.py::test_run_local_simultaneous_runs Timeout (0:01:00)!
Thread 0x7FF52834D700 (most recent call first, approximate line numbers):
  File "/home/travis/build/python-trio/trio/pypy3.5-5.8-beta-linux_x86_64-portable/lib-python/3/threading.py", line 261 in wait
  File "/home/travis/build/python-trio/trio/pypy3.5-5.8-beta-linux_x86_64-portable/lib-python/3/queue.py", line 147 in get
  File "/home/travis/build/python-trio/trio/testenv/site-packages/trio/_core/tests/test_local.py", line 100 in main
  File "/home/travis/build/python-trio/trio/testenv/site-packages/trio/_core/_run.py", line 1241 in run_impl
  File "/home/travis/build/python-trio/trio/testenv/site-packages/trio/_core/_run.py", line 1111 in run
  File "/home/travis/build/python-trio/trio/pypy3.5-5.8-beta-linux_x86_64-portable/lib-python/3/threading.py", line 851 in run
  File "/home/travis/build/python-trio/trio/pypy3.5-5.8-beta-linux_x86_64-portable/lib-python/3/threading.py", line 899 in _bootstrap_inner
[...]

PR #384, opened 2017-12-21
Enable pytest-faulthandler

njsmith:
Hopefully this will help us diagnose weird freezes like gh-379


njsmith:
codecov is confused because I merged #383 while the tests were running.

Issue #279, opened 2017-08-11
High-level helpers for working with unix domain sockets


njsmith:
Now that the high level networking API is coming together, we should also add unix domain helpers. Should be pretty straightforward.

Some notes here: https://github.com/python-trio/trio/issues/73#issuecomment-285752193

I guess something like:
```python
async def open_unix_stream(path):
    ...

async def open_unix_listeners(path, *, mode=0o666, backlog=None):
    ...
```
`mode=0o666` matches what twisted does; tornado does `0o600`. Should research which is better as a default.

The biggest issue is to figure out what to do about unlink-on-close. It's nice to keep thing tidy, but it introduces a race condition if you're trying to do no-downtime upgrades...

njsmith:
If we do do unlink on close, then it's important to note that `getsockname` is *not* reliable on AF_UNIX sockets. In particular, if you bind a socket and then rename the inode, `getsockname` will return the name you originally bound to, not the name the socket has now.

So one option would be `SocketListener(..., unlink_on_close=None)` / `SocketListener(..., unlink_on_close="/path/to/unlink")`. Another option would be to add a `sockaddr=...` option, which I guess we'll need anyway for #280, and then `unlink_on_close=True` would trust this instead of calling `getsockname`.

njsmith:
I think the locking algorithm in https://github.com/python-trio/trio/issues/73#issuecomment-285752193 will work, but it's super annoying. So I'm leaning towards: start by implementing a version where we never clean up stale socket files, and then if/when someone complains, implement the full locking version, so we can keep the same API.

Other libraries, in their "try to blow away an existing socket file first" code, check to see if it's a socket and only unlink it if so. We aren't going to unlink the file, but we might still want to do a just-in-case check where if the target file exists and is not a socket, then error out instead of renaming over it.

SunDwarf:
Hi, has this had any work at all? If not, I'd like to take an attempt at making it. I'm working on converting one of my libraries to support curio and trio, and this is a missing part for a bit of it for the trio side.

njsmith:
Are you looking to implement helpers for clients, servers, or both? The
client part should be pretty straightforward; the server part is more
complicated and requires some subtle decisions that I've been
procrastinating on :-). Either way, there's no one currently working on
this, and if you want to go for it then that'd be awesome.

SunDwarf:
I was looking at implementing both, but if you're waiting on design decisions I only need the client part so I can just restrict it to that for now.

[...]

PR #401, opened 2018-01-14
Unix client socket support. See: #279

SunDwarf:
A month later, I finally found the motivation to implement this.

Implementation was relatively simple; I reused SocketStream and wrote a very basic connector function (for client connections, I don't think there's any more that could be done).

Note:

 - Some of the changes are apparently due to yapf reformatting, and were unintentional.
 - The new test uses raw `socket.socket` to emulate a UDS server - not sure if that's the right thing to do.

### DIFF
--- a/trio/_highlevel_open_unix_stream.py
+++ b/trio/_highlevel_open_unix_stream.py
@@ -1,0 +1,32 @@
+import trio
+from trio._highlevel_open_tcp_stream import close_on_error
+from trio.socket import socket, SOCK_STREAM, AF_UNIX
+
+__all__ = ["open_unix_socket"]
+
+
+async def open_unix_socket(filename,):
+    """Opens a connection to the specified
+    `Unix domain socket <https://en.wikipedia.org/wiki/Unix_domain_socket>`_.
+
+    You must have read/write permission on the specified file to connect.
+
+    Args:
+      filename (str or bytes): The filename to open the connection to.
+
+    Returns:
+      SocketStream: a :class:`~trio.abc.Stream` connected to the given file.
+
+    Raises:
+      OSError: If the socket file could not be connected to.
+    """
+    if filename is None:
+        raise ValueError("Filename cannot be None")
+
+    # much more simplified logic vs tcp sockets - one socket type and only one
+    # possible location to connect to
+    sock = socket(AF_UNIX, SOCK_STREAM)
+    with close_on_error(sock):
+        await sock.connect(filename)
+
+    return trio.SocketStream(sock)
--- a/trio/_highlevel_open_unix_stream.py
+++ b/trio/_highlevel_open_unix_stream.py
@@ -1,6 +1,12 @@
 import trio
 from trio._highlevel_open_tcp_stream import close_on_error
-from trio.socket import socket, SOCK_STREAM, AF_UNIX
+from trio.socket import socket, SOCK_STREAM
+
+try:
+    from trio.socket import AF_UNIX
+    has_unix = True
+except ImportError:
+    has_unix = False
 
 __all__ = ["open_unix_socket"]
 
@@ -19,7 +25,11 @@ async def open_unix_socket(filename,):
 
     Raises:
       OSError: If the socket file could not be connected to.
+      RuntimeError: If AF_UNIX sockets are not supported.
     """
+    if not has_unix:
+        raise RuntimeError("Unix sockets are not supported on this platform")
+
     if filename is None:
         raise ValueError("Filename cannot be None")
 
--- a/trio/_highlevel_open_unix_stream.py
+++ b/trio/_highlevel_open_unix_stream.py
@@ -13,7 +13,7 @@ __all__ = ["open_unix_socket"]
 
 async def open_unix_socket(filename,):
     """Opens a connection to the specified
-    `Unix domain socket <https://en.wikipedia.org/wiki/Unix_domain_socket>`_.
+    `Unix domain socket <https://en.wikipedia.org/wiki/Unix_domain_socket>`__.
 
     You must have read/write permission on the specified file to connect.
 
--- a/trio/_socket.py
+++ b/trio/_socket.py
@@ -1,15 +1,14 @@
-from functools import wraps as _wraps, partial as _partial
+import os as _os
 import socket as _stdlib_socket
 import sys as _sys
-import os as _os
-from contextlib import contextmanager as _contextmanager
-import errno as _errno
+from functools import wraps as _wraps
 
 import idna as _idna
 
 from . import _core
 from ._deprecate import deprecated
 from ._threads import run_sync_in_worker_thread
+from ._util import fspath
 
 __all__ = []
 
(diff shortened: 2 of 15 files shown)


pquentin:
Hello, and thank you for the pull request! I'm not part of the team, but just trying to help here.

You have formatting changes because trio currently uses yapf 0.19.0 but you followed the documentation and used the latest version, 0.20.1. It's not easy to notice: it turns out yapf 0.19.0 is happy with the changes made by yapf 0.20.1. Anyway, since https://github.com/google/yapf/issues/484 is now fixed, it would make sense to upgrade to yapf 0.20.1, but maybe not as part of this pull request? Hope that helps!

njsmith:
@pquentin 
> I'm not part of the team, but just trying to help here.

Pff, if you're going to go reviewing pull requests then you're part of the team :-). Check your inbox for an invite.

@SunDwarf 
> Some of the changes are apparently due to yapf reformatting, and were unintentional.

Yeah, sorry about this – as @pquentin mentioned, we've had some trouble with yapf bugs recently and the contributor docs are a little out of sync there. I think we can leave it be in this case, given the weird details. (Specifically it looks like 0.20.1 has fixed the misformatting bugs we were seeing, and also fixed a bug where older versions of yapf were simply skipping over some `if` statements entirely. So normally I'd suggesting running yapf 0.19.0 over your PR to revert back to the old formatting, but this is the one case where that doesn't actually work, and asking you to go through and revert them by hand seems like unnecessary busywork.)

I'll do a proper review in a second.

trio/_socket.py
```diff
@@ -504,6 +506,13 @@ def shutdown(self, flag):
                     "address should be a (host, port, [flowinfo, [scopeid]]) "
                     "tuple"
                 )
+        elif self._sock.family == AF_UNIX:
+            await _core.checkpoint()
+            # unwrap path-likes
+            if hasattr(address, "__fspath__"):
+                return address.__fspath__()
+            return address
```
njsmith: Oh, excellent catch.

At first I was going to suggest using [`os.fspath`](https://docs.python.org/3/library/os.html#os.fspath), but then I remembered that that's only added in 3.6. We could skip this check on 3.5, but... we probably want to support `trio.Path` objects even on 3.5. And then I looked again at trio's async file I/O code, and we actually have a second copy of this `os.fspath` emulation there.

Would you be up to refactoring this into a `trio._util.fspath` helper? Ideally [based on the sample code here](https://www.python.org/dev/peps/pep-0519/#os)? (It turns out that calling `address.__fspath__()` directly is subtly wrong, because python's magic method lookup is weird.)

trio/tests/test_highlevel_open_unix_stream.py
```diff
@@ -0,0 +1,35 @@
...
+    except OSError:
+        pass
+
+    # server socket
+    serv_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    serv_sock.bind(name.__fspath__())
+    serv_sock.listen(1)
+
+    # for some reason the server socket can accept the connection AFTER it has
+    # been opened, so use that to test here
```
njsmith: Yep, that's the point of the `listen(1)` :-). It means "allow at least 1 connection to accepted, and hold onto it in the kernel, before I call `accept`".

trio/tests/test_highlevel_open_unix_stream.py
```diff
@@ -0,0 +1,35 @@
...
+    unix_socket = await open_unix_socket(name.__fspath__())
```
njsmith: Can you also test with passing the `Path` object directly to `open_unix_socket`? (On 3.5 you can use a `trio.Path` to make sure it has an `__fspath__` method.)

trio/_socket.py
njsmith: I also filed [bpo-32562](https://bugs.python.org/issue32562).

trio/_highlevel_open_unix_stream.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Returns:
+      SocketStream: a :class:`~trio.abc.Stream` connected to the given file.
+
+    Raises:
+      OSError: If the socket file could not be connected to.
+      RuntimeError: If AF_UNIX sockets are not supported.
+    """
+    if not has_unix:
+        raise RuntimeError("Unix sockets are not supported on this platform")
```
njsmith: Huh, how to handle this is an interesting question.

One tradition says, unsupported operations should just be missing, so you can do `hasattr(trio, "open_unix_stream")` to check for whether it's there. The other approach is to do this. On the one hand, the missing approach is probably more traditional in Python, and it's what we do for OS-specific operations in `trio.hazmat`. OTOH, it might make it harder to get `trio.__all__` set up correctly and in a way that doesn't confuse static analysis tools, and we might need to switch how we handle OS-specific operations anyway to support pluggable backends. Hmm.

I'm going to hit post on this review so you can see the other comments, while I think about this a little more :-)

trio/tests/test_highlevel_open_unix_stream.py
SunDwarf: Ah, I didn't know that. Been a *long* time since I've used raw sockets.

trio/_socket.py
SunDwarf: Sure, I'll update it when I get home today.

trio/tests/test_highlevel_open_unix_stream.py
SunDwarf: Will do later today.

SunDwarf:
Changes requested have been made. 

There was a few places where the fspath util could be added, but I figured they were out of scope for the PR so I didn't touch them.

njsmith:
The build is failing because yapf is cranky -- maybe we found a place where 0.19 and 0.20.1 disagree after all. If you could do a quick

```
pip install yapf==0.19.0
yapf -rpi setup.py trio
git commit -a -m "Run yapf"
git push
```

Then it should get the test suite running again.

SunDwarf:
No clue why the build failed there. The tests just seemed to not run?

pquentin:
I think this is because of https://www.traviscistatus.com/incidents/p1g70p3qrq4z. Hopefully this will get fixed at some point!

trio/_highlevel_open_unix_stream.py
njsmith: I haven't come to any real conclusions here, but given the problems that autocompleters have when some functions are only conditionally defined (#314), and that this way callers get a better error message than `AttributeError: open_unix_stream`, I'm going to say let's leave this as is for now for now and we can revisit it later if we change our mind.

njsmith:
Meh, I started that review before you did the reformatting and it seems to have confused github a bit. Github is hiding some of my comments; please click "Show outdated" to see them, because they aren't actually outdated :-)

SunDwarf:
Okay, that should be everything done.

pquentin:
Thanks! Unfortunately you have a new yapf issue: https://travis-ci.org/python-trio/trio/jobs/330320293

njsmith:
Looking good! I *think* the `pragma: no cover` on the `if hasattr(os, "fspath")` check was unnecessary (we run tests on both 3.5 and 3.6, so it should exercise both variants of the branch). So I removed it and now the tests are running again, and I'll merge once we confirm I didn't screw anything up with the extremely complicated edit :-)

njsmith:
Done! I'll send [your Github invite](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team) in a moment.

Issue #402, opened 2018-01-15
using a nursery without a context manager?


lilydjwg:
I'm writing a class that will spawn tasks during its lifetime. I'd like to have a `self._nursery` so that I can spawn tasks and let it run (e.g. a heartbeat task). But it seems that the nursery can only be used in a context manager, so it closes in the same scope it gets created. I don't want to pass in a nursery from outside because it's an implementation detail.

How should I write such a thing, which has background tasks during its lifetime?

njsmith:
Excellent question!

So, one of Trio's weirdest, most controversial decisions is that it takes the position that the existence of a background task is *not* an implementation detail, and *should* be exposed as part of your API. On balance I think this is the right decision, but it's definitely a bit experimental and has some trade-offs.

Why does Trio do this? In my experience, other systems make it seem like you can abstract away the presence of a background task or thread, but in reality it leaks in all kinds of ways: they end up breaking control-C handling, or they cause problems when you're trying to exit the program cleanly, or they leak when you try to cancel the main operation, or you have sequencing problems because the function you called completed but the work it promised to do is still going on in the background, or the background task crashes with an unexpected exception and then the exception gets lost and all kinds of weird problems ensue... so while it might make your API feel a little messier in the short term, in the long term everything goes easier if you make this explicit.

Also, keep in mind that everyone else writing and using Trio libraries has the same issue, so your API is not going to feel too weird :-).

I don't know what you're trying to do exactly. Maybe it's something like a websocket connection, where you want to constantly be reading from the socket to respond to heartbeat ("ping") requests. One pattern would be to do something like:

```python
@asynccontextmanager
def open_websocket(url):
    ws = WebSocket()
    await ws._connect(url)
    try:
        async with trio.open_nursery() as nursery:
            nursery.start_soon(ws._heartbeat_task)
            yield ws
            # Cancel the heartbeat task, since we're about to close the connection
            nursery.cancel_scope.cancel()
    finally:
        await ws.aclose()
```

And then your users can use it like:

```python
async with open_websocket("https://...") as ws:
    await ws.send("hello")
    ...
```

If you want to get fancier, another option would be to provide one version where your users pass in their own nursery, for experts:

```python
class WebSocket(trio.abc.AsyncResource):
    def __init__(self, nursery, url):
        self._nursery = nursery
        self.url = url

    async def connect(self):
        # set up the connection
        ...
        # start the heartbeat task
        self._nursery.start_soon(self._heartbeat_task)

[...]

PR #403, opened 2018-01-17
Our StackOverflow tag got renamed to python-trio I guess

njsmith:

PR #404, opened 2018-01-17
Update tag name

mjpieters:
See #402; `trio` is too generic a name to be a suitable tag on Stack Overflow.


pquentin:
Thanks @mjpieters for taking the time to do this! This pull request is more complete than #403. I would have merged it, but I think [the travis-ci.org build got hit by an incident](https://www.traviscistatus.com/incidents/c1s4dlyxd4lf) so I can't push the button. :)

njsmith:
Yes, this is definitely above-and-beyond, thank you :-)

PR #406, opened 2018-01-19
Bump yapf version to 0.20.1

njsmith:


pquentin:
Okay so this was not based on master, so I merged master into this branch and ran yapf again because there were other changes. I'd be happy if someone could double check that I did not make a mistake here.

(I first made a mistake so I had to force push, and I also deleted a codecov comment because for some reason codecov commented twice before my force push.)

(Oh, and the way we use yapf is not really that satisfactory, because we're going to have friction like in #401 for every yapf upgrade. But I don't know what else could be done.)

njsmith:
Codecov likes to randomly double comment. I don't know why. But it does it a lot.

> Oh, and the way we use yapf is not really that satisfactory, because we're going to have friction like in #401 for every yapf upgrade

Maybe... going forward yapf will keep working, and we'll be able to track the latest release without any problems, and everything will be sunshine and rainbows?

...yeah, idk. At least for now we *are* up to date with the latest release, so we've kicked the problem down the road a bit. Maybe there's some clever way to use pyup.io or requires.io?

pquentin:
> ...yeah, idk. At least for now we are up to date with the latest release, so we've kicked the problem down the road a bit. Maybe there's some clever way to use pyup.io or requires.io?

I agree this can wait. Thanks for the merge!

PR #427, opened 2018-02-06
Subprocessing

smurfix:
having fixed the issues raised in chat

### DIFF
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -402,6 +402,17 @@ create and access :class:`Result` objects from any thread you like.
 .. autoclass:: Error
    :members:
 
+Child processes
+===============
+
+Trio supports threads natively.
+See :ref:`Threads <threads>` for details.
+
+Subprocesses are not yet supported.
+If you did acquire a child process by some other means, use
+:func:`wait_for_child` to wait until it terminates.
+
+.. autofunction:: wait_for_child
 
 Sleeping and waking
 ===================
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -152,6 +152,14 @@ All environments provide the following functions:
        if another task is already waiting for the given socket to
        become writable.
 
+In addition, :mod:`trio.hazmat` contains low-level stream adapters for
+OS file handles. You can use them to trio-ize standard input or output. Be
+aware that there is no provision for text encoding yet, which is why
+they're in `trio.hazmat`.
+
+.. autoclass:: ReadFDStream
+
+.. autoclass:: WriteFDStream
 
 Unix-specific API
 -----------------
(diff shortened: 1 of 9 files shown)


smurfix:
NB, somebody needs to implement ProcessWaiter for Windows. Should be straightforward but I can't test it.

njsmith:
There's a lot of important new public API here, and I want to take some time to think it over. Some particular concerns are that it doesn't work on Windows (appveyor is failing due to missing `fcntl`, and more fundamentally there's the problem that most ways of connecting to a child stdin/out/err on Windows don't support non-blocking operation), and I'm not at all convinced `wait_for_child` and `{Read,Write}FdStream` should be public APIs at all. (As opposed to telling people to call `Process.wait`, and that `Process.std{in,out,err}` are streams of some unspecified concrete type.)

I know you want to get some kind of subprocess support into trio-asyncio ASAP. To check I'm understanding right: for the way you're doing this in trio-asyncio currently, you're delegating most of the work to asyncio's existing support for Unix subprocesses, and the two integration points with Trio are (a) the new `wait_for_child` call, (b) you assume that Trio's `wait_{read,write}able` functions work on subprocess pipes. Is that correct?

In the long run I suspect that we'll want to rework how subprocesses are handled in trio-asyncio so that we have our own implementation of `asyncio.SubprocessTransport` that directly wraps `trio.Process` (or whatever we end up calling it). The reason I think this is that there are a *lot* of nasty details on getting subprocesses working reliably cross-platform, and asyncio doesn't do a particularly good job here (in particular, their `SelectorEventLoop` doesn't support subprocesses at all on Windows). And these are problems we're going to eventually need to solve in trio anyway. So for trio-asyncio, instead of mostly using asyncio's subprocess code and forwarding a few bits and pieces to trio, I think in the end we'll be better off replacing the whole thing.

But this is also part of why I'm reluctant to add `wait_for_child` to Trio itself – the one use case seems to be temporary scaffolding to get something working now, but that won't be part of a long-term solution.

Maybe it would make sense to move the `wait_for_child` code into trio-asyncio for now, if you're in a hurry to get that working?

(And BTW, `wait_for_child`/`ProcessWaiter` look like they could be substantially simplified if they're only intended to handle this one case. There seems to be some stray code left over from previous versions, and I don't know why there's Windows-specific code in there, given that (a) this PR doesn't support Windows, (b) `os.waitpid` works on Windows anyway, or at least is documented to? I could be missing things though, I didn't look very carefully.)

smurfix:
OK. Fair enough. How about moving the whole thing to a trio-subprocess repository where we are a bit more free WRT experimenting with what does or does not work / leaving code in with the intention that somebody else should fix it up later (which I did with the Windows stuff)?

thejohnfreeman:
What is the state of this work? In Gitter, you pointed to the newer [`subprocess` branch](https://github.com/smurfix/trio/tree/subprocess), but this PR is for the older `waitpid2` branch. Please, can you summarize the distinction for me?

njsmith:
This PR seems to have stalled, and we're taking subprocess support in a slightly different direction, so I'm going to close it. Some of the code here might well end up being useful, but fortunately closing just takes it off the list of PRs waiting for attention, without actually deleting anything :-)

PR #428, opened 2018-02-07
trio_test accept Instrument as parameter

touilleMan:
Pretty cool to have this with pytest-trio fixtures ;-)

### DIFF
--- a/trio/testing/_trio_test.py
+++ b/trio/testing/_trio_test.py
@@ -26,6 +26,8 @@ def trio_test(fn):
         else:
             raise ValueError("too many clocks spoil the broth!")
         instruments = [i for i in kwargs.values() if isinstance(i, Instrument)]
-        return _core.run(partial(fn, **kwargs), clock=clock, instruments=instruments)
+        return _core.run(
+            partial(fn, **kwargs), clock=clock, instruments=instruments
+        )
 
     return wrapper
--- a/trio/testing/_trio_test.py
+++ b/trio/testing/_trio_test.py
@@ -1,7 +1,7 @@
 from functools import wraps, partial
 
 from .. import _core
-from ..abc import Clock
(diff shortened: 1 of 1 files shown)


trio/testing/_trio_test.py
```diff
@@ -25,6 +25,7 @@ def wrapper(**kwargs):
             clock = clocks[0]
         else:
             raise ValueError("too many clocks spoil the broth!")
-        return _core.run(partial(fn, **kwargs), clock=clock)
+        instruments = [i for i in kwargs.values() if isinstance(i, Instrument)]
```
nicoddemus: Shouldn't `instruments` be removed from `kwargs` as well?

```python
instruments = [kwargs.pop(k) for (k, v) in kwargs.items() if isinstance(v, Instrument)]
```

Otherwise `fn` will also receive the instruments still in `**kwargs`.

touilleMan: This mechanism is typically for pytest fixtures, where you declare the fixture you want by having arguments in your function. Hence if you modify kwargs, the called function will no longer have the right number of arguments.

nicoddemus: Oh I see so you want to pass the instruments to `_core.run` *in addition* to `fn`, got it.

Sorry for the noise! 👍

touilleMan:
I fixed the style, but not sure why travis with python3.7-dev is failing (https://travis-ci.org/python-trio/trio/jobs/338422042)...

njsmith:
The 3.7-dev failures aren't your fault; they're some genuine minor
incompatibilities between 3.7 and the trio test suite, that didn't show up
before because Travis only just started updating their nightly builds after
several months of not doing so.

touilleMan:
@njsmith rebased and ready to be merged ;-)

PR #431, opened 2018-02-13
Replace "path.__fspath__()" with "fspath(path)"

### LINKED ISSUE #419: In the async file code, use trio._util.fspath instead of calling __fspath__ directly

pmav99:
Fixes #419

### DIFF
--- a/trio/_path.py
+++ b/trio/_path.py
@@ -4,7 +4,7 @@ import types
 import pathlib
 
 import trio
-from trio._util import async_wraps
+from trio._util import async_wraps, fspath
 
 __all__ = ['Path']
 
@@ -141,11 +141,7 @@ class Path(metaclass=AsyncAutoWrapperType):
         return 'trio.Path({})'.format(repr(str(self)))
 
     def __fspath__(self):
-        try:
-            return self._wrapped.__fspath__()
-        # python3.5 compat
-        except AttributeError:  # pragma: no cover
-            return str(self)
+        return fspath(self._wrapped)
 
     @wraps(pathlib.Path.open)
     async def open(self, *args, **kwargs):
--- a/trio/_file_io.py
+++ b/trio/_file_io.py
@@ -4,7 +4,7 @@ import io
 import trio
 from . import _core
 from .abc import AsyncResource
-from ._util import aiter_compat, async_wraps
+from ._util import aiter_compat, async_wraps, fspath
 
 __all__ = ['open_file', 'wrap_file']
 
@@ -162,7 +162,7 @@ async def open_file(
     """
     # python3.5 compat
     if isinstance(file, trio.Path):
-        file = file.__fspath__()
+        file = fspath(file)
 
     _file = wrap_file(
         await trio.run_sync_in_worker_thread(
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -2,6 +2,7 @@
 
 import os
 import sys
+import pathlib
 from functools import wraps
 import typing as t
 
@@ -257,18 +258,20 @@ def fixup_module_metadata(module_name, namespace):
         fix_one(obj)
 
 
-# This is copied from PEP 519 as the implementation of os.fspath for
-# Python 3.5. See: https://www.python.org/dev/peps/pep-0519/#os
+# This is based on the PEP 519 fspath implementation.
+# The function has been adapted to work with pathlib objects on python 3.5
 # The input typehint is removed as there is no os.PathLike on 3.5.
+# See: https://www.python.org/dev/peps/pep-0519/#os
 
 
 def fspath(path) -> t.Union[str, bytes]:
     """Return the string representation of the path.
 
-    If str or bytes is passed in, it is returned unchanged. If __fspath__()
-    returns something other than str or bytes then TypeError is raised. If
-    this function is given something that is not str, bytes, or os.PathLike
-    then TypeError is raised.
+    If str or bytes is passed in, it is returned unchanged. If a pre-python 3.6
+    pathlib object is passed, its string representation is returned. If
+    __fspath__() returns something other than str or bytes then TypeError is
+    raised. If this function is given something that is not str, bytes,
+    pathlib.PurePath or os.PathLike then TypeError is raised.
     """
     if isinstance(path, (str, bytes)):
         return path
@@ -281,6 +284,10 @@ def fspath(path) -> t.Union[str, bytes]:
     except AttributeError:
         if hasattr(path_type, '__fspath__'):
             raise
+        # On python 3.5 pathlib objects don't have an __fspath__ method.
+        # but we still want to get their string representation.
+        if isinstance(path, pathlib.PurePath):
+            return str(path)
     else:
         if isinstance(path, (str, bytes)):
             return path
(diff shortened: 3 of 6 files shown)


trio/_path.py
```diff
@@ -142,7 +142,7 @@ def __repr__(self):
 
     def __fspath__(self):
         try:
-            return self._wrapped.__fspath__()
+            return fspath(self._wrapped)
         # python3.5 compat
         except AttributeError:  # pragma: no cover
             return str(self)
```
njsmith: Tests are failing, and I think it's because this fallback isn't being hit anymore.

pmav99: You are right. I think the issue though is that the PEP implementation does not work with pre Python 3.6 pathlib objects.

I just pushed a new commit where I added that check on `fspath()` itself. If it works, then this workaround can probably be removed.

njsmith:
Tests are still failing.

pmav99:
@njsmith 
The tests are still failing on python 3.7, but I think that the failures are not related to the commit.

So the PR updates the `fspath()` implementation so that it works with Python 3.5 pathlib and removes the workaround from `trio.Path.__fspath__()`.

njsmith:
Yep, looks good to me now. We'll have to wait for #434 to fix the CI stuff before merging though (since we have branch protection turned on).

njsmith:
Cycling to trigger a new test run

njsmith:
The complaint from codecov seems to be because `trio._util.fspath` doesn't have full coverage yet, and this PR touched it, so it's a bit cranky about the PR. I'm not bothered by this, so in it goes :-). (It would be good to get proper test coverage for our `fspath` implementation at some point, in case that's something you're interested in – should be pretty easy.)

Thanks @pmav99! And as per [our contributing docs](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team), watch out for a github invite.

pmav99:
@njsmith 
The relevant tests from stdlib + one more test case for pathlib objects:
``` python
# -*- coding: utf-8 -*-

""" Backport of os.PathLike/os.fspath for Python < 3.6. """

# Upstream implementations:
# tests: https://github.com/python/cpython/blob/da6c3da6c33c6bf794f741e348b9c6d86cc43ec5/Lib/test/test_os.py#L3527-L3571
# fspath: https://github.com/python/cpython/blob/da6c3da6c33c6bf794f741e348b9c6d86cc43ec5/Lib/os.py#L1029-L1058
# PathLike: https://github.com/python/cpython/blob/da6c3da6c33c6bf794f741e348b9c6d86cc43ec5/Lib/os.py#L1067-L1078

import os
import abc
import sys
import typing as t
import pathlib
import unittest


class PathLike(abc.ABC):
    """Abstract base class for implementing the file system path protocol."""

    @abc.abstractmethod
    def __fspath__(self):
        """Return the file system path representation of the object."""
        raise NotImplementedError

    @classmethod
    def __subclasshook__(cls, subclass):
        return hasattr(subclass, '__fspath__')


# We monkey patch os here. We could probably avoid this, but this way we can
# copy paste the PathLike implementation + tests from StdLib.
if not hasattr(os, "PathLike"):
    os.PathLike = PathLike


class _PathLike(os.PathLike):
    """ Class implementing the file system path protocol."""

    def __init__(self, path=""):
        self.path = path

    def __str__(self):
        return str(self.path)

    def __fspath__(self):
        if isinstance(self.path, BaseException):
            raise self.path
        else:
            return self.path


def fspath(path) -> t.Union[str, bytes]:
    """Return the path representation of a path-like object.

    Returns
    -------
    - If str or bytes is passed in, it is returned unchanged.
    - If the os.PathLike interface is implemented it is used to get the path
      representation.
    - If the python version is 3.5 or earlier and a pathlib object is passed,
      the object's string representation is returned.

    Raises
    ------
    - Regardless of the input, if the path representation (e.g. the value
      returned from __fspath__) is not str or bytes, TypeError is raised.
    - If the provided path is not str, bytes, pathlib.PurePath or os.PathLike,
      TypeError is raised.
    """
    if isinstance(path, (str, bytes)):
        return path
    # Work from the object's type to match method resolution of other magic
    # methods.
    path_type = type(path)
    try:
        path_repr = path_type.__fspath__(path)
    except AttributeError:
        if hasattr(path_type, '__fspath__'):
            raise
        elif isinstance(path, pathlib.PurePath):
            # On python 3.5 pathlib objects don't have an __fspath__ method.
            # but we still want to get their string representation.
            return str(path)
        else:
            raise TypeError(
                "expected str, bytes or os.PathLike object, "
                "not " + path_type.__name__
            )
    if isinstance(path_repr, (str, bytes)):
        return path_repr
    else:
        raise TypeError(
            "expected {}.__fspath__() to return str or bytes, "
            "not {}".format(path_type.__name__,
                            type(path_repr).__name__)
        )


class TestPEP519(unittest.TestCase):

    fspath = staticmethod(fspath)

    def test_return_bytes(self):
        for b in b'hello', b'goodbye', b'some/path/and/file':
            self.assertEqual(b, self.fspath(b))

    def test_return_string(self):
        for s in 'hello', 'goodbye', 'some/path/and/file':
            self.assertEqual(s, self.fspath(s))

    @unittest.skipIf(
        sys.version_info < (3, 6),
        "fs{en,de}code() functions don't support os.PathLike on Python < 3.6"
    )
    def test_fsencode_fsdecode(self):
        for p in "path/like/object", b"path/like/object":
            pathlike = _PathLike(p)

            self.assertEqual(p, self.fspath(pathlike))
            self.assertEqual(b"path/like/object", os.fsencode(pathlike))
            self.assertEqual("path/like/object", os.fsdecode(pathlike))

    def test_pathlike(self):
        self.assertEqual('#feelthegil', self.fspath(_PathLike('#feelthegil')))
        self.assertTrue(issubclass(_PathLike, os.PathLike))
        self.assertTrue(isinstance(_PathLike(), os.PathLike))

    def test_pathlib_35(self):
        # pathlib on python 3.5 does not implement the PathLike protocol
        # so we need explicitly tests for pathlib.
        self.assertEqual(
            "/home/user", self.fspath(pathlib.PurePath("/home/user"))
        )
        self.assertEqual(
            "C:\\windows\\", self.fspath(pathlib.PurePath("C:\\windows\\"))
        )

    def test_garbage_in_exception_out(self):
        vapor = type('blah', (), {})
        for o in int, type, os, vapor():
            self.assertRaises(TypeError, self.fspath, o)

    def test_argument_required(self):
        self.assertRaises(TypeError, self.fspath)

    def test_bad_pathlike(self):
        # __fspath__ returns a value other than str or bytes.
        self.assertRaises(TypeError, self.fspath, _PathLike(42))
        # __fspath__ attribute that is not callable.
        c = type('foo', (), {})
        c.__fspath__ = 1
        self.assertRaises(TypeError, self.fspath, c())
        # __fspath__ raises an exception.
        self.assertRaises(
            ZeroDivisionError, self.fspath, _PathLike(ZeroDivisionError())
        )
```

njsmith:
I was thinking it would be simplest to just write our own test that exercises the different paths here, especially the red ones: https://codecov.io/gh/python-trio/trio/src/master/trio/_util.py

I guess a full-fledged backport package has some advantages in theory, but it seems like a lot more work?

PR #440, opened 2018-02-14
[testing] Add tests for fspath.

### LINKED ISSUE #439: Add tests for fspath

pmav99:
Fixes #439

### DIFF
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -258,49 +258,63 @@ def fixup_module_metadata(module_name, namespace):
         fix_one(obj)
 
 
-# This is based on the PEP 519 fspath implementation.
-# The function has been adapted to work with pathlib objects on python 3.5
+# os.fspath is defined on Python 3.6+ but we need to support Python 3.5 too
+# This is why we provide our own implementation. On Python 3.6+ we use the
+# StdLib's version and on Python 3.5 our own version.
+# Our own implementation implementation is based on PEP 519 while it has also
+# been adapted to work with pathlib objects on python 3.5
 # The input typehint is removed as there is no os.PathLike on 3.5.
 # See: https://www.python.org/dev/peps/pep-0519/#os
 
 
-def fspath(path) -> t.Union[str, bytes]:
-    """Return the string representation of the path.
+def _fspath(path) -> t.Union[str, bytes]:
+    """Return the path representation of a path-like object.
 
-    If str or bytes is passed in, it is returned unchanged. If a pre-python 3.6
-    pathlib object is passed, its string representation is returned. If
-    __fspath__() returns something other than str or bytes then TypeError is
-    raised. If this function is given something that is not str, bytes,
-    pathlib.PurePath or os.PathLike then TypeError is raised.
+    Returns
+    -------
+    - If str or bytes is passed in, it is returned unchanged.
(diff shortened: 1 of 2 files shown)


pmav99:
BTW, the fspath implementation on both the PEP and on stdlib uses:

``` python
    try:
        path_repr = path_type.__fspath__(path)
    except AttributeError:
        if hasattr(path_type, '__fspath__'):
            raise
```
The `raise` line is the only line that is not being covered by the tests. 
To be honest, I can't understand how the raise branch can ever be reached...!

njsmith:
It can be reached by having an `__fspath__` method that exists, but raises `AttributeError` when called. The goal is to avoid swallowing an error when the implementation of `__fspath__` has a bug in it, like accessing a non-existent attribute.

trio/_util.py
```diff
@@ -258,49 +258,63 @@ def fix_one(obj):
...
+            "not {}".format(path_type.__name__,
+                            type(path_repr).__name__)
+        )
 
 
 if hasattr(os, "fspath"):
-    fspath = os.fspath
+    fspath = os.fspath  # pragma: no cover
+else:
+    fspath = _fspath  # pragma: no cover
```
njsmith: You don't need the `pragma: no cover`s here, we combine coverage across different python versions so both lines should be covered. (And if one of them eventually becomes uncovered, because we dropped 3.5 support, then we want to see that in the coverage report so we know we can delete that code :-).)

trio/tests/test_util.py
```diff
@@ -221,3 +223,106 @@ def test_module_metadata_is_fixed_up():
...
+    def fspath(self):
+        return _fspath
+
+
+if sys.version_info > (3, 6):
+
+    class TestStdLibFspath(BaseTestFspath):
+        @pytest.fixture
+        def fspath(self):
+            return os.fspath
```
njsmith: This is probably overkill too, for the same reason... just testing `trio._util.fspath` will automatically test both versions, and make sure they're consistent, once you take into account the full CI matrix.

trio/tests/test_util.py
```diff
@@ -221,3 +223,106 @@ def test_module_metadata_is_fixed_up():
     # Also check methods
     assert trio.ssl.SSLStream.__init__.__module__ == "trio.ssl"
     assert trio.abc.Stream.send_all.__module__ == "trio.abc"
+
+
+# define a concrete class implementing the PathLike protocol
+# Since we want to have compatibility with Python 3.5 we need
+# to define the base class on runtime.
+BaseKlass = os.PathLike if hasattr(os, "PathLike") else object
```
njsmith: This is harmless, but I think it would work just as well if we left off the base class entirely and dropped the `issubclass` checks below.

njsmith:
Codecov is complaining because your test actually introduces a few lines of new uncovered code: https://codecov.io/gh/python-trio/trio/compare/a72e471e7e00837a5f01a3963c230f8dd1fd6155...cd667905a8ddeffe63557033371258f183fd4766/diff

Specifically it looks like `ConcretePathLike.__str__` and `ConcretePathLike.__subclasshook__` are defined but never executed. I guess they can just be deleted?

Otherwise looks good to me 👍, feel free to merge after fixing that!

njsmith:
Perfect, thanks!

PR #441, opened 2018-02-15
Rework test_run_local_simultaneous_runs to be more reliable

njsmith:
We used to use a single queue to send messages to and from the
threads, which of course is unreliable because the main thread could
end up reading back its own message. In particular, on PyPy this
happened regularly, and occasionally it meant that the test
deadlocked. So this fixed gh-379.

This patch also updates the test harness to actually pull errors back
from the child threads, so that if the test does fail then we can
detect it.


pquentin:
This makes a lot of sense, thanks!

Issue #437, opened 2018-02-14
The echo example in the tutorial is broken in 0.3.0


njsmith:
It uses `sendall` on sockets: https://stackoverflow.com/questions/48777728/python-trio-attributeerror-sendall

We should remove the socket example entirely and replace it with the high-level API... no beginner needs to be exposed to the raw socket API. Except, I like that on the server side it shows you what `serve_tcp` is actually doing with the nursery :-(. Need to figure out some way to explain that without dragging in all the low-level details... or maybe it's not that important. Certainly having a totally broken example is not better.

PR #444, opened 2018-02-19
Rewrite echo client example in tutorial to use high-level interface

njsmith:
Fixes gh-437


njsmith:
CC @sscherfke

njsmith:
Let's get this in, so no-one else is reading the broken example...

PR #446, opened 2018-02-21
The combined limiter example stores and returns undefined 'limiter'

albertogomcas:
I assume the intended object was the freshly created 'combined_limiter'


njsmith:
Okay, that was a bit annoying, but all sorted now, and #448 should fix that problem for good.

Thanks! As [described in our contributor guide](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team), you should get a github invitation shortly.

Issue #447, opened 2018-02-21
Bogus test failure in test_open_unix_socket due to use of an insufficiently unique name


njsmith:
Seen here: https://ci.cryptography.io/blue/rest/organizations/jenkins/pipelines/python-trio/pipelines/trio/branches/PR-446/runs/1/nodes/5/log/?start=0

```
____________________________ test_open_unix_socket _____________________________

    async def test_open_unix_socket():
>       name, serv_sock = await get_server_socket()

../.venv/lib/python3.5/site-packages/trio/tests/test_highlevel_open_unix_stream.py:50: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    async def get_server_socket():
        name = Path() / tempfile.gettempdir() / "test.sock"
        try:
            await name.unlink()
        except OSError:
            pass
    
        serv_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
>       serv_sock.bind(fspath(name))  # bpo-32562
E       OSError: [Errno 48] Address already in use

../.venv/lib/python3.5/site-packages/trio/tests/test_highlevel_open_unix_stream.py:24: OSError
```

PR #448, opened 2018-02-21
Refactor open_unix_stream tests, and use more unique filename

njsmith:
The main substantive change here is switching to using mktemp() to
pick a server socket name, instead of using the fixed name
"/tmp/test.sock". This should fix gh-447.

While I was at it, I also refactored the main test, and made it clean
up after itself more carefully.

Finally, I replaced

   os.path.join(tempfile.gettempdir(), tempfile.mktemp())

with

   tempfile.mktemp()

Because it turns out they're equivalent: mktemp() already returns an
absolute filename inside an appropriate directory.

### DIFF
--- a/trio/tests/test_highlevel_open_unix_stream.py
+++ b/trio/tests/test_highlevel_open_unix_stream.py
@@ -5,7 +5,6 @@ import tempfile
 import pytest
 
 from trio import open_unix_socket, Path
-from trio._util import fspath
 
 try:
     from socket import AF_UNIX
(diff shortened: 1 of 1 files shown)


SunDwarf:
I must've missed this when I wrote the initial tests, I wonder why it never threw for me.

njsmith:
It was fine as long as only one test was running at a time, which is the normal case. The Jenkins build runs two copies of the test suite at the same time on the same host, and even then it mostly worked... To see the bug we had to get super unlucky and have two copies of those exact (very short) tests running at the same time.

No big deal though, these things happen :-)

PR #451, opened 2018-02-21
Add different implementation of a task monitor. [WIP]

SunDwarf:
This is WIP, don't merge yet.

This is a basic implementation of a task monitor as in #413. It is inspired by curio and the other PR open for this (#429) but takes a different approach, requiring the user to spawn a handler for the Monitor server.

Current plans:

 - [x] Use the instrument API to provide a command ala `redis-cli monitor` showing the live status of your application
 - [ ] Add a task tree displayer
 - [ ] Add the ability to execute code in a task's frame (maybe?)

### DIFF
--- a/trio/monitor.py
+++ b/trio/monitor.py
@@ -5,10 +5,13 @@ import sys
 import signal
 
 import inspect
+from types import FunctionType
 
+from . import Queue, WouldBlock, BrokenStreamError
+from ._highlevel_serve_listeners import _run_handler
 from ._version import __version__
 from .abc import Instrument
-from .hazmat import current_task, Task
+from .hazmat import current_task, Task, UnboundedQueue
 
 # inspiration: https://github.com/python-trio/trio/blob/master/notes-to-self/print-task-tree.py
 
@@ -37,6 +40,11 @@ class Monitor(Instrument):
         self.auth_pin = ''.join(
             rand.choice(string.ascii_letters) for x in range(0, 8)
         )
+
+        self._is_monitoring = False
+        # semi-arbitrary size, because otherwise we'll be dropping events
+        # no clue how to make this better, alas.
+        self._monitoring_queue = Queue(capacity=100)
 
     @staticmethod
     def get_root_task() -> Task:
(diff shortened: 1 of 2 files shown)


njsmith:
What do you think of splitting this off into a standalone project in the python-trio organization?

(I'll take a look at the code in a moment.)

trio/monitor.py
```diff
@@ -0,0 +1,380 @@
...
+            if task.parent_nursery is None:
+                parent = "N/A"
+            else:
+                parent = str(id(task.parent_nursery.parent_task))
+
+            result = [str(id(task)).ljust(widths[0]), name.ljust(49), parent]
+
+            lines.append(' '.join(result))
+
+        return lines
```
njsmith: I wonder if we should be showing a hierarchical task tree view as much as possible, as like... a matter of principle almost. People coming to trio are used to unstructured global task lists in other systems (Unix processes, threads, asyncio, ...), and it's just not the right image to have in your head when you're trying to reason about a trio program, so reinforcing that seems like it might be important?

Also, as a usability thing, it'd probably be helpful to have a single command that dumps a combined task tree + stack traces of all tasks, for the use case where you're trying to do remote debugging via a confused user ("can you run the monitor, type <XX>, and paste the output you get into this issue?"). Sort of like how `thread apply all bt` is a magic incantation you ask people to type into gdb.

trio/monitor.py
```diff
@@ -0,0 +1,380 @@
...
+        elif isinstance(coro, ANextIter):
+            # black hole
+            return
+        else:
+            # A generator decorated with @types.coroutine
+            yield coro.gi_frame, coro.gi_frame.f_lineno
+            coro = coro.gi_yieldfrom
+
+
+class Monitor(Instrument):
```
njsmith: Thought: another way to structure this would be to move the main loop into its own class or top-level function, and have the instrument as a separate class. When you ran the `monitor` command, it would create a new instrument object like `instrument = MonitorTrace(myqueue)` or whatever, make it an active instrument (`trio.hazmat.add_instrument`), read its output until done, then deactivate it again (`trio.hazmat.remove_instrument`).

trio/monitor.py
```diff
@@ -0,0 +1,380 @@
...
+        # send the banner
+        version = __version__
+        await stream.send_all(
+            b"Connected to the Trio monitor, using "
+            b"trio " + version.encode(encoding="ascii") + b"\n"
+        )
+
+        while True:
+            await stream.send_all(b"trio> ")
+            command = await stream.receive_some(2048)
```
njsmith: As you're probably aware, this is a bit of a kluge -- there's no guarantee that a call to `receive_some` will get exactly 1 command. It could get less or more. Not really an issue for a proof-of-concept, but I guess sooner or later it'd be good to replace this with a proper streaming line splitter. Could also be useful as a push for us to figure out the right way to do that :-). I started poking at this a while ago in https://github.com/njsmith/sansio_toolbelt, which is supposed to be a nice incremental parse buffer (extracted from h11) plus some simple tools like line parsers and length-prefixed-buffer parsers, but I was being indecisive about the APIs and it's only half-written.

trio/monitor.py
```diff
@@ -0,0 +1,380 @@
...
+
+            else:
+                message = "Unknown event: {}".format(key)
+
+            message = prefix + message
+
+            try:
+                await stream.send_all(message.encode("ascii") + b'\n')
+            except BrokenStreamError:  # client disconnected on us
+                return
```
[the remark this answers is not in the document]
SunDwarf: Oh, yeah, that's probably a good idea.

trio/monitor.py
SunDwarf: Yeah, I was aware; I figured once the core was done I could move onto getting the more fiddly parts of network I/O done.

SunDwarf:
Also, splitting this off into a `trio-monitor` package would probably be a better idea, given some of the "magic" I did to make the feed command behave.

SunDwarf:
Closing in favour of https://github.com/python-trio/trio-monitor

PR #454, opened 2018-02-22
Add `root_task` to get the root task without having to walk the tree.

### LINKED ISSUE #452: Add trio.hazmat.current_root_task() function
Because right now debuggers have to walk up the tree from the current task, and it's silly.

CC: @touilleMan @SunDwarf

SunDwarf:
Closes #452.


SunDwarf:
Oops, seems like PyCharm cleaned up imports on test_run.py automatically by accident.

njsmith:
* yapf is unhappy with the extra lines that I guess pycharm inserted :-). Just running yapf should be enough to fix it.

* Can we call it `current_root_task` to match all the other ambient-state-fetching calls? (compare `current_clock`, `current_statistics`, etc.)

* Instead of writing a global function, it'd be more consistent with other similar functions to make a method on `Runner` and decorate it with `@_public`. Unfortunately `current_task` is a misleading example here -- `current_clock` is better :-). This should also let you drop it from `_run.__all__`. (The magic is in `_generate_method_wrappers` at the bottom of `_run.py`.)

SunDwarf:
Oh no I broke my fork. Gonna re-open this.

SunDwarf:
That said; the newlines were yapf misbehaving, and now yapf crashes with an error when I try and reformat my code, so.

njsmith:
oh, that's not git misbehaving, it's an intentional bug in github's ui! If they were using git normally, it would just work, but instead they've carefully coded it so that that PRs don't show you the new commits in your branch that aren't in master, they show you the commits that weren't in master back when you opened the PR. I complained to support@github about this at some point and they were very insistent that they didn't want the PR UI to show you what commits would actually be merged if you clicked "merge", because apparently it's more usable this way.

So don't blame yourself, is what I'm saying.

I don't know why yapf woudl be crashing :-/. Are you using 0.20.1?

SunDwarf:
It was crashing on all versions I tried from 0.19.0 through 0.20.1.

Issue #36, opened 2017-01-24
Graceful handling of sockets (or whatever) getting closed while in use


njsmith:
Suppose we have one task happily doing its thing:
```python
async def task1(sock):
    await sock.sendall(b"...")
```

and simultaneously, another task is a jerk:
```python
async def task2(sock):
    sock.close()
```

It would be nice to handle this gracefully.

How graceful can we get? There is a limit here, which is that (a) it's Python so we can't actually stop people from closing things if they insist, and (b) the OS APIs we depend on don't necessarily handle this in a helpful way. Specifically I believe that for epoll and kqueue, if a file descriptor that's they're watching gets closed, they just silently stop watching it, which in the situation above would mean `task1` blocks forever or until cancelled. (Windows `select` -- or at least the `select.select` wrapper on Windows -- seems to return immediately with the given socket object marked as readable.)

As an extra complication, there are really two cases here: the one where the object gets closed just before we hand it to the IO layer, and the one where it gets closed while in possession of the IO layer.

And for sockets, one more wrinkle: when a stdlib `socket.socket` object is closed, then its `fileno()` starts returning -1. This is actually kinda convenient, because at least we can't accidentally pass in a valid fd/handle that has since been assigned to a different object.

Some things we could do:

* In our `close` methods, first check with the `IOManager` whether the object is in use, and if so cancel those uses first. (On Windows we can't necessarily cancel immediately, but I guess that's OK b/c on Windows it looks like closing the handle will essentially trigger a cancellation already; it's the other platforms where we have to emulate this.)

* In `IOManager` methods that take an object-with-fileno()-or-fd-or-handle, make sure to validate the fd/handle while still in the caller's context. I think on epoll/kqueue we're OK right now because the `wait_*` methods immediately register the fd, and on Windows the `register_for_iocp` method is similar. But for Windows `select`, the socket could be invalid and we won't notice until it gets `select`ed on in the `select` thread. Or it could *become* invalid on its way to the select thread, or in between calls to `select`... right now I think this will just cause the select loop to blow up.

njsmith:
I fixed some of these issues while working on other things -- not sure if there's any more to do here or not.

njsmith:
[...]

Issue #459, opened 2018-02-25
stream.aclose() not working properly?


sscherfke:
When I modify the tutorial echo-client so that the sender closes the stream after one iteration, I expected the connection to close and the receiver to stop:
```python
# echo-client.py

import sys
import trio

# arbitrary, but:
# - must be in between 1024 and 65535
# - can't be in use by some other program on your computer
# - must match what we set in our echo server
PORT = 12345
# How much memory to spend (at most) on each call to recv. Pretty arbitrary,
# but shouldn't be too big or too small.
BUFSIZE = 16384

async def sender(client_stream):
    print("sender: started!")
    while True:
        data = b"async can sometimes be confusing, but I believe in you!"
        print("sender: sending {!r}".format(data))
        await client_stream.send_all(data)
        await trio.sleep(1)
        break  # stop after one run
    await client_stream.aclose()  # Close stream

async def receiver(client_stream):
    print("receiver: started!")
    while True:
        data = await client_stream.receive_some(BUFSIZE)
        print("receiver: got data {!r}".format(data))
        if not data:
            print("receiver: connection closed")
            sys.exit()

async def parent():
    print("parent: connecting to 127.0.0.1:{}".format(PORT))
    client_stream = await trio.open_tcp_stream("127.0.0.1", PORT)
    async with client_stream:
        async with trio.open_nursery() as nursery:
            print("parent: spawning sender...")
            nursery.start_soon(sender, client_stream)

            print("parent: spawning receiver...")
            nursery.start_soon(receiver, client_stream)

trio.run(parent)
```
The server prints a *connection closed* but the client’s `receiver()` does not detect the stream being closed and does not finish.

I also found no other way to explicitly cancel the `receiver()` task when the `sender()` task is done.

njsmith:
Yeah, the receive call should definitely wake up and complain, but it turns out this has to be explicitly implemented and we haven't done it yet (#36). Sorry this bit you.

BTW, you can explicitly cancel anything in trio by wrapping it in a cancel scope and then calling the scope's `cancel` method. It's not tied to tasks. That you need to in to use case is a bug, but fyi. If you're not sure what I'm talking about here then please ask in chat, I or someone be happy to explain :-)

sscherfke:
Is there anything I can do to help you with the PR?

njsmith:
It's merged :-) Try out master and see if it works for you?

sscherfke:
Works for me :)

PR #460, opened 2018-02-26
When a socket/fd is closed, wake up outstanding waiters

njsmith:
Still todo:

- full test coverage
- updating the stream layer to match
- is InterruptedByCloseError the best name? Should it inherit OSError?
- Which layers should use which exception?

Fixes gh-36, gh-459


njsmith:
CC @sscherke @smurfix, since I know this bit both of you before.

njsmith:
Sorry, I mean @sscherfke

sscherfke:
I tried your branch and my examples are working now. :-)

njsmith:
I think the way to do this is to rename `ClosedStreamError` and `ClosedListenerError` to `ClosedResourceError` (or just `ClosedError`?), and then use that same exception here too, to avoid cross-layer skew.

So todo list:

* renames + deprecation + release notes for same
* update code to use it instead of `InterruptedByCloseError`
* add tests
* fix the testing streams to implement the same behavior
* update the generic stream tests to check for this

smurfix:
Hmm. ClosedResourceError, as a subclass of ClosedError, as a subclass of RuntimeError?

sscherfke:
Why not `ConnectionReset`?  It’s built in and people are used to it.

njsmith:
@smurfix We already use `Resource` to mean ["anything that can be closed"](https://trio.readthedocs.io/en/latest/reference-io.html#trio.abc.AsyncResource), so I'm not sure how you could have a `ClosedError` that wasn't also a `ClosedResourceError` :-).

@sscherfke `ConnectionResetError` is part of Python's `OSError` hierarchy. They don't document this as well as they could, but it means that `ConnectionResetError` has very specific semantics: "I did some syscall, and it returned with errno = `ECONNRESET`." That's not what happened here, and in fact it's a different situation: at the abstract stream level we don't expose errnos and stuff, but just distinguish between "some external event broke this stream" and "you broke this stream by calling `close`". (Because this is the only distinction that, makes sense for a generic stream that might have some arbitrary underlying implementation, and also because mostly users don't care about any details beyond this anyway.) `ECONNRESET` is in the former category (it means a TCP peer sent us a RST packet), while this PR is about what exception to use for the latter category :-).

njsmith:
Just pushed a much more complete patch here.

Still todo:

* Make sure cross-platform tests and coverage look good.
* Make a final decision on exception names. I'm leaning towards renaming `ResourceBusyError` to `BusyResourceError`, so that our exceptions consistently put the most distinctive information first. I'm wondering whether we should have a `BrokenResourceError`, that `BrokenStreamError` could be replaced by (or subclass I guess); right now it really is specific to streams, but we've had to move two different exceptions out of the stream API now (`BusyResourceError` and `ClosedResourceError`) in order to avoid skew between levels of abstraction, so maybe we should be taking a hint...

This is in a good state to review though.

trio/_core/_exceptions.py
```diff
@@ -103,3 +104,16 @@ class ResourceBusyError(Exception):
     the data get scrambled.
 
     """
+
+
+class ClosedResourceError(Exception):
+    """Raised when attempting to use a resource after it has been closed.
+
+    Note that "closed" here means that *your* code closed the resource,
+    generally by calling a method with a name like ``close`` or ``aclose``. If
```
oremanj: Do you think it's worth also mentioning the "exited a context manager" possibility for why the thing was closed?

njsmith:
It looks like the new uncovered lines are in the logic for handling `monitor_kevent`, which is an untested stub anyway, so not an issue here.

trio/_core/_exceptions.py
njsmith: Yeah, that's a good idea!

njsmith:
@Fuyukai raised an interesting question in chat: you might expect `__del__` methods to call `notify_*_close`, and what happens then? The problem is that invoking any kind of trio methods from `__del__` is fraught, since `__del__` is like a signal handler that might run at an arbitrary moment, or in an arbitrary thread. The only Trio function that's safe to call from a `__del__` method is `TrioToken.run_sync_soon`. So... we could define `__del__` methods that do that, e.g. on trio sockets.

This patch doesn't, though. Why not? Well... if you're calling `wait_*` on a socket, then you'd better be holding a reference to that socket! And in fact the `wait_socket_*` functions actually do hold a reference to the socket that's passed in. So... `__del__` won't run while there are any `wait_socket_*` calls outstanding. Or put another way, if `__del__` is being called, then it means we don't need to worry about calling `notify_socket_close`.

There might be some super-obscure edge cases where this breaks down? But implementing and testing `__del__` would be super tricky, and I can't think of any now, so I think we can let it be and worry about it if it ever comes up (or if someone comes up with a case where it's actually a problem).

njsmith:
@Fuyukai also pointed out that we can always handle further exception name changes as a separate PR. So I think I've now addressed all comments and this is ready to merge.

njsmith:
@oremanj thanks!

PR #462, opened 2018-02-27
Add better KeyboardInterrupt handling with broken threads.

### LINKED ISSUE #461: Maybe don't trust threading.current_thread() to tell us whether we're in the main thread for signal handling purposes
As @SunDwarf discovered ([chat log](https://gitter.im/python-trio/general?at=5a943b2a888332ee3ad47c41)), in some extremely weird circumstances it's possible to be  in the main thread, yet `threading.current_thread() == threading.main_thread()` returns `False`. This is problematic because the main thread is special with respect to signal handling, and we use this in several places to check which kind of signal behavior to expect.

Specifically the case that broke is: `import logbook`, which internally imports some internals of gevent, including `gevent.threading`. In gevent 1.2.2 (and probably earlier versions), this triggers some monkeypatching of threading internals. (You're not supposed to import this module directly; logbook does anyway.) In particular, it breaks `threading.current_thread()`, so it gives the wrong value.

This will be fixed in the next version of gevent, but that's not out yet. (See: https://github.com/gevent/gevent/issues/984) Also it's possible that there are other cases where the threading and signal modules end up with different ideas of whether they're running in the main thread? It seems hard to imagine though. So it's not entirely clear whether this is worth fixing.

The fix would be to stop checking `threading.current_thread()` before calling signal functions, and instead when we want to know whether the signal module thinks we're in the main thread, ask it directly by calling some signal function. So either "EAFP" style where we just `try: signal.signal(...) except ValueError: not_in_main_thread_fallback()`, or we could keep the current code style by adding a `trio._util.in_main_thread` that does something like, trying to call `signal.signal(signal.SIGINT, signal.getsignal(signal.SIGINT))` and checks whether it raises an exception or not.

Fuyukai:
Closes #461.

### DIFF
--- a/newsfragments/461.feature.rst
+++ b/newsfragments/461.feature.rst
@@ -1,3 +1,0 @@
-Add :func:`trio._util.current_thread` as a safer alternative to
-:func:`threading.current_thread`, and fix KeyboardInterrupt handling when
-threading state has been modified by a 3rd-party library.
--- a/newsfragments/461.feature.rst
+++ b/newsfragments/461.feature.rst
@@ -1,0 +1,3 @@
+Add :func:`trio._util.current_thread` as a safer alternative to
+:func:`threading.current_thread`, and fix KeyboardInterrupt handling when
+threading state has been modified by a 3rd-party library.
--- a/trio/_core/_ki.py
+++ b/trio/_core/_ki.py
@@ -1,12 +1,13 @@
+import inspect
+import signal
+import sys
 import threading
 from contextlib import contextmanager
 from functools import wraps
-import signal
-import sys
-import inspect
 
-import attr
 import async_generator
+
+from .._util import is_main_thread
 
 __all__ = [
     "enable_ki_protection",
@@ -177,7 +178,7 @@ disable_ki_protection.__name__ = "disable_ki_protection"
 @contextmanager
 def ki_manager(deliver_cb, restrict_keyboard_interrupt_to_checkpoints):
     if (
-        threading.current_thread() != threading.main_thread()
+        not is_main_thread()
         or signal.getsignal(signal.SIGINT) != signal.default_int_handler
     ):
         yield
--- a/trio/tests/test_util.py
+++ b/trio/tests/test_util.py
@@ -8,8 +8,9 @@ import threading
 from async_generator import async_generator, yield_
 
 from .. import _core
-from .._util import acontextmanager, signal_raise, ConflictDetector, fspath, \
-    is_main_thread
+from .._threads import run_sync_in_worker_thread
+from .._util import (acontextmanager, signal_raise, ConflictDetector, fspath,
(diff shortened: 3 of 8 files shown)


trio/tests/test_util.py
```diff
@@ -310,3 +310,18 @@ def test_bad_pathlike_implementation(self, exception, method):
...
+    def _():
+        nonlocal _thread_result
+        _thread_result = is_main_thread()
+
+    assert is_main_thread()
+    thread = threading.Thread(target=_)
+    thread.start()
+    thread.join()
+    assert _thread_result is not None
+    assert _thread_result is False
```
njsmith: Can you write this as

```python
async def test_is_main_thread():
    assert is_main_thread()

    async def not_main_thread():
        assert not is_main_thread()

    await run_sync_in_worker_thread(not_main_thread)
```

(I keep forgetting this is an option too and having to remind myself :-). But it really is a better habit I think -- not just because it's simpler, but e.g. in the test as originally written if there was an exception inside `_` then it would get lost.)

newsfragments/461.feature.rst
```diff
@@ -0,0 +1,3 @@
+Add :func:`trio._util.current_thread` as a safer alternative to
```
njsmith: `trio._util.main_thread` isn't a public function that we'll document, so it shouldn't be mentioned in the end-user release notes. Here we should just be saying that KeyboardInterrupt handling is now robust against third-party libraries that break `threading.current_thread`. (And I guess that makes this a `bugfix`, not a `feature`?)

trio/_core/_ki.py
```diff
@@ -177,7 +178,7 @@ def wrapper(*args, **kwargs):
 @contextmanager
 def ki_manager(deliver_cb, restrict_keyboard_interrupt_to_checkpoints):
     if (
-        threading.current_thread() != threading.main_thread()
+        not is_main_thread()
```
njsmith: Can you grep for `current_thread`? I think there are 3-4 other places where we use this check and should switch to this. (Definitely in `catch_signals` and `_io_windows.py` at least.)

SunDwarf:
Well, this was an interesting test failure: `PyEval_EvalFrameEx returned a result with an error set`

njsmith:
Whoops, sorry about the extra `async`. And wow, how did you get that error message exactly? It sounds like an interpreter bug...

SunDwarf:
Seemingly by passing an async function to run_sync_in_worker_thread? Otherwise, no idea.

njsmith:
Ok, looks good to me, thanks!

Looking at [the travis results](https://travis-ci.org/python-trio/trio/builds/347248237?utm_source=github_status&utm_medium=notification) I guess they must have fixed that `PyEval_EvalFrameEx` issue already in 3.6 or so? Pretty weird error to get...

Issue #242, opened 2017-07-10
Discussion: can we steal stuff from Concurrent ML?


njsmith:
So from @wingo's [blog I have learned about Concurrent ML](https://wingolog.org/archives/2017/06/29/a-new-concurrent-ml). (His [concurrency tag](https://wingolog.org/tags/concurrency) is also worth perusing generally.)

The core insight, IIUC, is to take the atomic blocking operations – roughly, the same ones our current [cancellation guarantee](https://trio.readthedocs.io/en/latest/reference-core.html#cancellation-and-primitive-operations) applies to – and systematically split them into a few consistent sub-operations:

* attempt to perform the operation synchronously, possibly wrapping the result in some thunk
* publish that we're waiting for the operation to be doable and should be woken up if it might succeed
* give up on waiting ("unpublish")

We already implicitly have this basic structure open-coded in a bunch of places, e.g. the `_try_sync` helper in `trio/socket.py`, the classes in `trio/_sync.py`, etc. Pretty much anywhere you see the `yield_if_cancelled`/`yield_briefly_no_cancel` pair in trio fits this general pattern, and "unpublish" is basically just our abort callback. So the advantages of reifying this would partly be just to simplify the code by having a single implementation of the overall pattern that we could slot things into – but even more, so because given the above pieces, you can create generic implementations of three variants:

* Try to perform the operation, but fail if it would block (like the `*_nowait` variants that we currently implement in an ad hoc way)
* Perform the operation, blocking as necessary (what we do now)
* A thing like golang's `select`, where you can say "perform exactly one of the following operations: ..."

(The first is done by just calling the "try this" sub-operation; the second you do by trying and then blocking if you fail, in a loop, with the unpublish operation as the abort callback; the third is done by calling a bunch of "try this" sub-operations and if one succeeds you're done, otherwise publish *all* the operations and go to sleep. There's some subtleties around knowing which operation woke you up, and when unpublish happens, etc., but that's the basic idea.)

Right now we have a bunch of manual implementations of `await x()` / `x_nowait()` primitives. It's not clear that we have enough, either; #14 has an example of a case where you semantically need `accept_nowait`, and for HTTP client connection pooling when you pull a connection out of the pool you need something like `receive_some_nowait` to check whether the server has closed it before you try to use it.

[...]

Issue #61, opened 2017-02-26
In epoll IOManager make it possible to wait on POLLPRI


njsmith:
Useful for e.g. GPIO support: https://www.kernel.org/doc/Documentation/gpio/sysfs.txt

`EPOLLWAKEUP` might be interesting too, but is pretty exotic and a little more subtle (unlike our other events, removing an `EPOLLWAKEUP` from the set is actually semantically meaningful).

njsmith:
I'm not sure if this is actually useful – e.g. the GPIO documentation linked above says "THIS ABI IS DEPRECATED ... NEW USERSPACE CONSUMERS ARE SUPPOSED TO USE THE CHARACTER DEVICE ABI." Going to close this until the use cases are clearer.

If we do find a reason to do this, I think it should be as a low-level Linux-specific API, similar to the low-level kqueue API (#578). There's just no consistency across systems about what counts as "priority" or "exceptional" data, or how it's handled – for example, kqueue doesn't even have a (portable) way to do `POLLPRI`, and then there was that time that Apple tried to [add support for urgent-data...](https://sandstorm.io/news/2015-04-08-osx-security-bug)

smurfix:
No problem, I've written `trio-gpio` which wraps libgpiod and does exactly that.

njsmith:
Noting for the record, since I was looking into OOB data handling for sockets, and might some day want to find this info again:

Using the BSD socket "OOB" APIs to handle TCP "urgent data" is a reasonably well-defined and portable thing. It's almost completely useless – literally the only use case I know if is to give telnet slightly better flow control when running over an unencrypted TCP connection – so I don't know that we should care. But if we did decide to care, then we would need a way to wait for "OOB" data. On Linux you can get notifications through `EPOLLPRI`, on Windows you can use `AFD_POLL_RECEIVE_EXPEDITED`, and on macOS you can use `EVFILT_EXCEPT` + `NOTE_OOB`. (This is a macOS extension though – on most BSDs there's no way to use kqueue to wait for OOB data at all. Which says something about how useless it is.)

We'd also need to think a bit about how to expose the concept: `EPOLLPRI` covers a few things besides OOB data (like the legacy GPIO use case that originally motivated this issue), while `AFD_POLL_RECEIVE_EXPEDITED` and `EVFILT_EXCEPT`+`NOTE_OOB` are very specific to OOB data.

PR #463, opened 2018-02-28
Add ability to wait for POLLPRI

smurfix:
This extension is required for some interesting things. Personally I need it for accessing GPIO pins (specifically, epoll()ing for changing inputs).

I still have to figure out how to reliably test this, given that the TCP URGent flag is less than well supported these days, Travis is unlikely to run on a machine with two connected GPIO pins, and I haven't done anything with PTYs in a decade.

### DIFF
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -194,6 +194,30 @@ Unix-like systems provide the following functions:
        become writable.
 
 
+.. function:: wait_priority(fd)
+   :async:
+
+   Block until there is an "exceptional condition" on the given file
+   descriptor.
+
+   .. warning::
+
+      This method uses the ``POLLPRI`` flag to the ``epoll`` system call.
+      Its semantics are highly specific to the type of file descriptor
+      you're passing to it.
+
+   This test may be used to
+   * wait for changes to hardware inputs (Linux GPIO)
+   * notice changed state of a PTY slave by the master
+   * …
+
+   :arg fd:
+       integer file descriptor, or else an object with a ``fileno()`` method
+   :raises trio.ResourceBusyError:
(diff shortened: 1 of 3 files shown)


njsmith:
The issue for this is #61.

Looking at the docs linked there, I see that the described gpio API is now considered obsolete and broken. Out of curiosity, does the new "character device API" also use POLLPRI? (I realize it will be a while before everyone has kernels with the new API, and that there are other possible use cases for this, but I find it never hurts to know more about how things are used :-).)

A fun thing to deal with is that on Windows when a non-blocking connect fails, this is reported by marking the socket as "priority". Everyone else uses "readable" for this. Currently we hack around this the same way that some other libraries do, by making the windows `wait_socket_readable` function actually wait for either read or priority. If we wanted to support `wait_socket_priority` there as well then we'd need to figure out how let them coordinate.

I guess one option would be to remove the hack from the io layer, and instead have our `connect` method wait for both read and priority. With the current io API, this would require that `connect` open a nursery to call both `wait` functions simultaneously. That feels a little weird, but I guess would be ok? Are there workloads that care about getting maximum speed out of `connect` on Windows? Another option to consider would be redoing the API to allow a task to wait for several events at a time, like `wait_io_ready(flags)`. (Some connection here to https://github.com/python-trio/trio/issues/242, though that gets into much more complicated issues.)

Another Windows-related thing to consider is that if we switch to iocp-only (#52), then we won't have a way to wait for priority at all on that platform. Though fortunately we won't need to, either, since the iocp version of `connect` works differently.

njsmith:
Github is telling me you just pushed some strange things here?

njsmith:
This is very stale, so going to close for now... feel free to re-open if you want to pick it up again.

Issue #178, opened 2017-05-28
Maybe make task-local storage work from run_in_worker_thread functions


njsmith:
This is handy for logging, and safe because we know that *either* the task *or* the thread is running, but never both - so no locking is needed. But maybe confusing (it can't work for run-locals, or `run_in_worker_process`)?

What should `current_task()` return in a `run_in_worker_thread` function? Currently we effectively find the task-locals by looking up the current task, so the most naive way to make task-locals work would to make `current_task()` work and then task-locals would work for free. Is that good or bad?

PR #478, opened 2018-03-20
Add contextvars support.

### LINKED ISSUE #417: Make trio.TaskLocal work in non-async scope
As the Trio documentation rightly says about including a request identifier in log messages:

> Now, here’s the problem: how does the logging code know what the request identifier is? One approach would be to explicitly pass it around to every function that might want to emit logs… but that’s basically every function, because you never know when you might need to add a log.debug(...) call to some utility function buried deep in the call stack, and when you’re in the middle of a debugging a nasty problem that last thing you want is to have to stop first and refactor everything to pass through the request identifier! 

And indeed, task local storage provides a nice solution for retrieving the request identifier :

```
task_info = trio.TaskLocal(conn_id=None)
def log_filter(record):
    if task_info.conn_id is None:
        return

    record.msg = '(conn %d) ' + record.msg
    record.args = (task_info.conn_id,) + record.args
```

But there is a problem: the log_filter needs to be added to every log handler - it's not enough to just add it to the local logger because then it won't be applied all the way down to the call chain, and we can't add it to the root logger because it is not inherited. This is a little ugly, but can be worked around by running something like:

```
     # Add connection information
    for handler in root_logger.handlers:
        handler.addFilter(server.log_filter)
```

..after all logging has otherwise been initialized.

But, there is always going to be that annoying bit of code that tries to log something before the event loop has started, giving us a `RuntimeError: must be called from async context` when accessing task_info.

Thus my question: can't access to the task local object outside of an async context just return the default value? Having to catch a `RuntimeError` when accessing an attribute is nasty.

### LINKED ISSUE #420: Support for PEP 567
[PEP 567 got accepted](https://www.python.org/dev/peps/pep-0567/) on the 22nd, adding context-local variables to Python 3.7, and it would be super useful to have implemented (especially for something I'm working on).

However, 3.5/3.6 compat is an issue, and if it wasn't for that I would've PR'd in already (since the change is probably super trivial, `msg = task.coro.send(next_send)` -> `context = contextvars.copy_context(); msg = context.run(task.coro.send, next_send)` or similar.), and I don't know how that would be solved.

Fuyukai:
This adds PEP 567 contextvars support to the core of Trio. 

Todo:

 - [x] Tests  
 - [x] Deprecate TaskLocal
 - [x] Wait for a backport for 3.5 and 3.6.

Closes #420 and closes #417, see also: #178.


Fuyukai:
Note the tests will not pass for this at all because I import contextvars which obviously doesn't work on 3.5/3.6 yet.

njsmith:
I guess we'll also want to replace `RunLocal` with some sort of `RunVar` interface that's modeled after `ContextVar`.

@1st1 would you have any objection to is releasing a `contextvars` package on pypi, that (initially) just has a pure python implementation of the contextvars API?

1st1:
Would you mind to wait a couple of days for this?  I will try to find time to do this.

1st1:
Here we go: https://pypi.python.org/pypi/pep567/0.1.0

1st1:
And here's the repo: https://github.com/MagicStack/pep567  Please feel free to file issue/requests. I can make you a co-maintainer Nathaniel if you need that.

1st1:
"contextvars" name is taken.

njsmith:
Heh, whoops. Though, the `import` name doesn't have to match the pypi name, so it could still be `import contextvars`. Or I bet that person could be convinced to give up their not-touched-in-the-last-6-years package given that the stdlib has claimed the name anyway...

1st1:
> Though, the import name doesn't have to match the pypi name, so it could still be import contextvars

Yeah, but that can be really confusing to some users (I'd be confused for one).

> Or I bet that person could be convinced to give up their not-touched-in-the-last-6-years package given that the stdlib has claimed the name anyway...

Maybe, the problem is that someone might still be using that old package. In the end, I thought that the name of the module is not that important...

njsmith:
@1st1 if it were just trio/asyncio/... who had to notice this I wouldn't care, but everyone who wants to use a `ContextVar` has to do the same import dance, so we have to include it in all our docs examples, etc. And if anyone is using that old package then, well, they're screwed when 3.7 comes out anyway :-). Also, it looks like it's only getting ~1 download/week. So... would you mind if I sent the owner a quick email to find out if they'd be amenable?

1st1:
Sure, please do. I'd be totally OK to rename the package.

njsmith:
Email sent!

1st1:
Thanks! :)

njsmith:
@Fuyukai If you want to keep making progress on this while we're waiting to hear back, I guess you can start with using `pep567` – should be easy to switch if the name changes.

You'll want to add an entry like:
```python
":python_version < '3.7'": ["pep567"]
```
to the `extras_require` dict in `setup.py`, and then it should start installing automatically on the older test runs.

We will want to deprecate `TaskLocal`; this requires:

* adding the warning (I think just decorating `TaskLocal.__init__`)

* Adding a `removal` newsfragment

* Deleting the `TaskLocal` docs

Also, do you want to handle the analogous `RunLocal` → `RunVar` transition in this PR too, or leave that for later?

(And thanks for working on this btw!)

Fuyukai:
> `":python_version < '3.7'": ["pep567"]`

TIL this is a thing. 

But sure, I'll do those changes.

1st1:
Alright, thanks to Nathaniel, the backport is now available under "contextvars" name :)

So please update the dependency in your PR to "contextvars".

njsmith:
Sweet!

@1st1 it looks like now the PR tests here are failing because your backport package only supports 3.6, not 3.5. Is that intentional? I know f strings are hard to give up... :-)

1st1:
> @1st1 it looks like now the PR tests here are failing because your backport package only supports 3.6, not 3.5. Is that intentional? I know f strings are hard to give up... :-)

They are indeed hard to give up :) contextvars@2.1 is now 3.5 compatible.

Are many trio users still on 3.5?  I'm considering dropping its support from uvloop/asyncpg, but will probably wait until 3.7 comes out.

Fuyukai:
> Deleting the TaskLocal docs

I don't think this should be done; they should be changed to mention ContextVar objects instead, since the majority of the section deals with the issue more abstractly and just autoclasses TaskLocal after (Unless this is what you meant?)

njsmith:
@Fuyukai

Yes, that's what I mean, sorry. The point is that we don't mention deprecated features in the docs, which might not be obvious. (Except for in the change log, of course.)

@1st1
> Are there many trio users still using 3.5? 

I'm not sure, but we do still support it for now. The main reasons are (1) it ships with ubuntu 16.04, so it does make trio easier to try for some people, (2) PyPy is still on 3.5 (though they're getting closer to 3.6), and twisted devs keep telling me that their users all deploy on PyPy. Plus it's much easier to compete with uvloop if we can cheat and use PyPy ;-).

Fuyukai:
The tests don't pass because TrioDeprecationWarning is warned on test_locals; doing a `with pytest.warns(TrioDeprecationWarning)` fails also because it doesn't warn because of course.

njsmith:
@Fuyukai Maybe try adding a `recwarn` fixture to the tests for the deprecated code, like in `trio/tests/test_socket.py::test_deprecated_resolver_methods`?

Fuyukai:
Docs updated, task locals deprecated, and I rebased for a slightly cleaner commit tree.

njsmith:
Doc build is failing with:
```
/home/travis/build/python-trio/trio/docs/source/reference-core.rst:1125:py:mod reference target not found: contextvars
```
because we're currently using the stable python 3 docs for intersphinx links, which means 3.6, and `contextvars` doesn't exist until 3.7.

I guess the slickest solution would be to use some [intersphinx cleverness](http://www.sphinx-doc.org/en/master/ext/intersphinx.html) to direct just that link to the python 3.7 docs. Alternatively it wouldn't be the end of the world though if we dropped the link, or made it a regular URL link instead of an automagic intersphinx link.

Fuyukai:
Given that it'll automatically fix itself in a few months time, I think the easier solution is to tell sphinx to ignore the warning and wait for the docs to point to 3.7 instead.

Fuyukai:
RunVar support has been added. The build fails because the doc can't look up trio.TaskLocal (??).

1st1:
I'm curious what's the motivation behind `RunVar`? Why `ContextVar` is not enough?

Fuyukai:
Well, ContextVar is task-local, whereas RunVar is shared across all tasks during a `trio.run`, so modifiying a RunVar modifies it for all whereas modifying a ContextVar only modifies it for that particular task.

1st1:
Right, but how is it different from a global/thread-local variable? Why does this specific use case deserve extra complexity?

Fuyukai:
No clue. I was just asked to add it.

njsmith:
@1st1 Another way to think of run-local data is that it's equivalent to attaching an attribute to an asyncio loop object, except that you can safely do it from third-party code. I agree that it's kind of a weird and obscure use case.  But `run_sync_in_worker_thread` needs it to track some data to avoid spawning too many threads simultaneously (you can think of it as holding the run's thread pool), and `run_sync_in_worker_thread` is built using only public APIs, so we need something like this as a public API :-). (And the reason `run_sync_in_worker_thread` is written using public APIs is because as a general policy we try to do this whenever possible. This is like how when asyncio needs something from the Python core, then you don't add a secret API that only asyncio can use, you add a new public API and asyncio becomes one of the users of it. Features like `run_sync_in_worker_thread` ship bundled together with trio, just like asyncio ships with the interpreter, but we want to keep the door open for users to invent their own better replacements 5 years from now.)

I guess `RunVar` might also be useful for other kinds of "system service" code... for example, if we were trying to implement trio's version of `call_soon_threadsafe` using only public APIs, then you'd need `RunVar` (and also `spawn_system_task`, which is another weird and obscure API). The only reason `call_soon_threadsafe` has to be built in is that `run` needs to start it automatically on every call.

Fuyukai:
The jenkins build hiccoughed on installing wheel it seems (??????) and I don't know how to resolve the doc error at all.

njsmith:
Yeah, the jenkins build currently has an old pip that only speaks some ancient version of TLS, and right now PyPI is randomly breaking old TLS versions to encourage folks to upgrade before their CDN service forces them to turn it off permanently: https://status.python.org/ , https://pyfound.blogspot.com/2017/01/time-to-upgrade-your-python-tls-v12.html

So, it worked! I've pinged the guy who maintains that server about it.

The doc error is because the docs for `RunLocal` reference the docs for `TaskLocal`, and the PR removes the `TaskLocal` docs, so that reference doesn't work anymore. But, I've realized I was unclear about the big picture here – sorry about that! Let me try to explain the motivation for the `RunVar` thing.

In 0.3.0, we had `TaskLocal`, whose API was designed to be like `threading.local` but for tasks, and we had `RunLocal`, whose API was designed to be like `threading.local`, but for runs. This makes things easy, because once you know how to any of these you know how to use all of them. But now we're replacing `TaskLocal` with `ContextVar`, which ruins the symmetry. Solution: replace `RunLocal` with `RunVar` to keep things symmetric.

So getting back to the doc error: we should deprecate `RunLocal` and delete its docs entirely, which will certainly fix the error as a side-effect :-). Does that make sense?

I'll do a proper code review in a few minutes too.

trio/_core/_run.py
```diff
@@ -1,43 +1,37 @@
...
     WaitTaskRescheduled,
 )
-from ._entry_queue import EntryQueue, TrioToken
-from ._ki import (
-    LOCALS_KEY_KI_PROTECTION_ENABLED, currently_ki_protected, ki_manager,
-    enable_ki_protection
-)
-from . import _public
+from .. import _core
+from .._util import acontextmanager
```
njsmith: I guess your editor is set up to automatically sort imports? I'm not going to stress over it here, but it does make it hard to tell what's changed... do you think you could get it to not do this in future PRs? (Or alternatively, if you absolutely can't stand unsorted imports, I guess you could set up [`isort`](https://github.com/timothycrosley/isort) alongside `yapf` so everyone does it? :-))

trio/_core/tests/test_local.py
```diff
@@ -166,10 +166,93 @@ def test_local_outside_run():
...
     for cls in _core.TaskLocal, _core.RunLocal:
         local = cls(default1=123, default2="abc")
         assert local.default1 == 123
         assert local.default2 == "abc"
         del local.default1
         assert not hasattr(local, "default1")
+
+
+# scary runvar tests
+def test_runvar_sanity():
```
njsmith: Style nit: can you rename this `test_runvar_basics` or `test_runvar_smoketest`? I try to avoid using "sanity" in the jokey "not horribly broken" sense these days b/c the jokey sense seems to come from actually pretty gross stereotypes about mental health. (And same comment on the other uses of "sanity".)

trio/_core/_run.py
Fuyukai: Oh, this was an accident; I have PyCharm set to auto-optimize imports by default on my global scope which removes unused ones and re-arranges them into a different order. I forgot to turn it off for this project.

Fuyukai:
I made the requested changes, as well as fixing the RunVar export. The build is (still) failing because the changelog refers to RunLocal which doesn't exist in the docs anymore. (Seems like a problem that's gonna hit in the future too).

njsmith:
@Fuyukai I did a last pass, and noticed a few more trivial things, so I just changed :-). I think this is good to merge – want to double-check my changes and then click merge if you agree?

njsmith:
Hooray! And thanks for your patience :-)

PR #483, opened 2018-03-31
Error for user-created Cancelled()

### LINKED ISSUE #342: Make Cancelled propagation/catching less error prone
You can't really catch `Cancelled` errors yourself, because you never know where they came from -- it might be some outer cancel scope your code knows nothing about. This also means that you can't really raise `Cancelled` errors yourself, because if you do that then they won't be tagged with the magic metadata that lets Trio know when to catch them, and if neither you can catch them nor Trio can catch them, they'll just crash your program, which is probably not what you wanted. But this is not really obvious. So we should probably make it so attempting to `raise Cancelled` gives a useful error message (e.g. by making `Cancelled.__init__` require some private value be passed).

RazerM:
Resolves #342.

- Is `RuntimeError` the correct type?
- Should the error message be more helpful? We probably want to point people to `CancelScope.cancel()`

## Todo
- [x] Add tests
- [x] Add note in `Cancelled`'s documentation that users should not raise it themselves.

### DIFF
--- a/newsfragments/342.removal.rst
+++ b/newsfragments/342.removal.rst
@@ -1,0 +1,3 @@
+Attempting to raise :exc:`trio.Cancelled` will cause a :exc:`RuntimeError`.
+:meth:`cancel_scope.cancel() <trio.The cancel scope interface.cancel>` should be
+used instead.
--- a/trio/_core/_exceptions.py
+++ b/trio/_core/_exceptions.py
@@ -55,6 +55,12 @@ class Cancelled(BaseException):
 
     then this *won't* catch a :exc:`Cancelled` exception.
 
+    Attempting to raise :exc:`Cancelled` yourself will cause a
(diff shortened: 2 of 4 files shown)


njsmith:
Thanks for looking at this. I think I just want to point at your first comment and nod vigorously :-). (`RuntimeError` is fine, I don't think it matters too much; mentioning `cancel` is probably a good idea; :+1: to docs and test (not that much is needed))

Fuyukai:
You seem to have generated history rather than added a newsfragment?

RazerM:
@Fuyukai oops, I was previewing the output and added the wrong file.

Fuyukai:
@njsmith your thoughts or can this be merged?

njsmith:
Looks good to me too. @Fuyukai want to do the honors?

PR #488, opened 2018-04-02
Fix wheel install for Jenkins sierra-python3.5 build

### LINKED ISSUE #487: Upgrade pip on MacOS builder
It looks like we still have pip 9.0.1 on our MacOS environments, which is causing tests to fail:
https://ci.cryptography.io/blue/organizations/jenkins/python-trio%2Ftrio/detail/PR-486/1/

I guess @reaperhulk is busy, but I think we could add a command like

`curl https://bootstrap.pypa.io/get-pip.py | python`

to `ci/travis.sh` before the first call to `pip`.

jab:
Fixes #487.

### DIFF
--- a/ci/travis.sh
+++ b/ci/travis.sh
@@ -34,6 +34,11 @@ if [ "$USE_PYPY_NIGHTLY" = "1" ]; then
     source testenv/bin/activate
 fi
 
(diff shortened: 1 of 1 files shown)


njsmith:
:+1:

Issue #472, opened 2018-03-16
add collection of worked examples to tutorial


njsmith:
We should refactor the tutorial into an initial part that's similar to what we have now, or the middle part of my talk, and then a collection of examples that also serve as excuses to go into more depth on particular topics.

I expect the list will grow over time, but here are some ideas. (Actually the main reason I'm filing this is to have a place collect these so I don't lose them.)

* The current tracing demo should move here. It's a good intro to trio introspection and to co-op concurrency.

* Happy eyeballs

* Multiplexing rpc

* Catch-all exception handler

* Custom nursery, like a race function or ignore-errors nursery (maybe both)

* Some standard stuff like echo server, proxy, fast web spider, ... Whatever doesn't end up in the main tutorial

* trio-asyncio example?

* `nursery.start`

Possibly some of these could be combined or form sequences, eg echo server -> catch all handler -> nursery.start

Fuyukai:
An `asyncio.gather`-like (collect the results of all tasks and return the results) would be a good example (as you've said before on Gitter iirc)

njsmith:
Oh yeah, good idea!

njsmith:
Oh, see also #421, which is a partial duplicate and has some more discussion of the nursery-based examples.

njsmith:
Oh duh, here's another one: an example of implementing a custom protocol, by combining a sansio protocol with the stream interface. (Probably some simple line-oriented or netstring-oriented thing. This is one of the motivations for why I started working on `sansio_toolbelt`. I should make it actually usable...)

smurfix:
A walkthrough for converting a sync protocol to Trio might also make sense.

WRT trio-asyncio: maybe simply refer to it. I do need to add some example that shows how to convert from asyncio to trio-asyncio to trio, and how that improves the code. ;-)

nicoddemus:
>  I do need to add some example that shows how to convert from asyncio to trio-asyncio to trio, and how that improves the code. ;-)

Would love to see that. 😁

njsmith:
Something like @jab's HTTP CONNECT proxy from https://github.com/python-trio/trio/pull/489#issuecomment-379455747 might be interesting too.

(Possibly rewritten to use h11 ;-).)

njsmith:
@oremanj's `as_completed` here might be interesting as the basis for a `gather` equivalent: https://gitter.im/python-trio/general?at=5ad186345d7286b43a29af53

[...]

PR #489, opened 2018-04-02
more small tutorial fixes (take 2)

jab:
- Fix outdated "no HTTP library yet" comment
- Fix typo (s/mean time/meantime)

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -841,10 +841,12 @@ An echo client
 
 The traditional application for demonstrating network APIs is an "echo
 server": a program that accepts arbitrary data from a client, and then
-sends that same data right back. Probably a more relevant example
+sends that same data right back. (Probably a more relevant example
 these days would be an application that does lots of concurrent HTTP
-requests, but trio doesn't have an HTTP library yet, so we'll stick
-with the echo server tradition.
+requests, but for that `you need an HTTP library
+<https://github.com/python-trio/trio/issues/236#issuecomment-310784001>`__
+such as `asks <https://asks.readthedocs.io>`__, so we'll stick
+with the echo server tradition.)
 
 To start with, here's an example echo *client*, i.e., the program that
 will send some data at our echo server and get responses back:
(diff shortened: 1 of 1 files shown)


njsmith:
This is an extremely finicky and unimportant issue, but since it comes up a lot I guess it's still worth mentioning: despite what you may have heard, the basic "make a hyperlink please" syntax in ReST uses *two* trailing underscores. If you use just one trailing underscore, then ReST treats it as, I don't even know, defining a term or something, so if your document happens to have two links with the same text and different targets, ReST freaks out. To avoid all this nonsense, it's a good habit to use the double underscores everywhere. (See the other links in the docs for examples.)

jab:
Good catch, thanks! Amended in 83a7fc0.

jab:
@njsmith Happy to work up a PR integrating an example like this into the tutorial if that would be helpful, or maybe even just a quick line like the following?
> For another example, check out how you can build an [HTTP CONNECT proxy on top of Trio](https://github.com/jab/trio_http_proxy/blob/master/trio_http_proxy.py) in about 70 lines of code.

jab:
Cool you've been collecting examples in #472 (hadn't seen that), and super cool to see your talk (hadn't seen that either) – live coding happy eyeballs in 40 lines was amazing! Reassuring to see that my HTTP CONNECT proxy example has the same basic structure as your TCP proxy example too. On that note, I think the TCP proxy example is great for its simplicity, so definitely worth including too, but even with the change to redirect to httpbin, I'm not sure the payoff comes close – like when I saw that I could point my browser through this tiny 60-line proxy I'd just written with Trio, and browsing any HTTPS site (including streaming YouTube) just worked, it felt like a super power.
screen shot 2018-04-16 at 08 01 12
For whatever it's worth 😊

P.S. I saw 0.4 just came out, so I reworked the CONNECT example to use the new contextvar support instead of passing custom `log` functions around. Nice work on 0.4!

njsmith:
Okay, using the trio talk on youtube as the demo URL definitely wins an extra prize for slickness :-)

Issue #1, opened 2017-01-22
Missing feature: sendfile


njsmith:
On Unix, `os.sendfile` is present and just needs a simple (but slightly finicky) wrapper.

On Windows, `os.sendfile` doesn't exist, but we can wrap [TransmitFile](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740565(v=vs.85).aspx) - shouldn't be too hard.

njsmith:
*ahem* *tap tap* is this thing on? Github doesn't give any way to see who's subscribed to an issue, so maybe I'm just talking to myself? Anyway, if anyone is listening, then welcome to the first update on backwards incompatibility in Trio. I think as far as the code goes what's in master right now is pretty much what will be in 0.2.0, but I still want to do some doc updates so it'll still probably be a few days before the actual release. That means that now would be a great time to read this and check out master and let me know if I've made any embarrassing mistakes or am gratuitously inconveniencing anyone.

In general, it turns out that there are very few actual breaking changes in 0.2.0; almost all the changes were easy enough to slap a small layer of back-compat shims on top of. The shims all print noisy `TrioDeprecationWarning`s that attempt to contain helpful information – suggestions for how to make them more helpful are welcome.

# If you read just one section of this comment then read this one

My plan is to be quite aggressive and drop the back-compat shims in 0.3.0. So with a few exceptions (see below), the idea is that:

* if your code works on 0.1.0, it should work on 0.2.0, possibly with warnings
* if you fix the warnings in 0.2.0, then your code should also work on 0.3.0
* jumping directly from 0.1.0 to 0.3.0 will probably break

**Important exception for anyone writing network servers**: any code using `bind` (the socket method) is going to break when going from 0.2.0 → 0.3.0, because it needs to become async (#241), BUT there's no way to provide a gentle transition here. [Feh.](https://github.com/python-trio/trio/issues/79#issuecomment-329944699) See below for more details and how to mitigate this.

[...]

PR #499, opened 2018-04-14
Rip out TaskLocal objects and fix system task context inheritance.

### LINKED ISSUE #289: spawn_system_task should not propagate TaskLocals
Probably we should make `spawn_impl` have a `system_task` flag instead of its current `ki_protection_enabled` flag. We already have to special case init as not inheriting TaskLocals, so this would remove that special case (since init is a system task).

Fuyukai:
Closes #289, and removes all the old TaskLocal code deprecated in the last version.


smurfix:
Umm …
* we should delay ripping out TaskLocal until after the deprecation time has passed. I want to impose some rules for minimum time until a deprecated feature is removed, but I don't want to *have* to delay a release just because that would otherwise be violated. We didn't decide on a timeframe yet but IMHO a month or two after the release in which the deprecation notice appeared *and* an intervening release is appropriate.
* this patch has at least two other ideas in it, and at least one of them causes massive failure. I don't have a problem with test failues, that's why we test, but could you please limit yourself to one feature per commit? ;-)

Fuyukai:
Hm, alright. It's just that ripping it out and then updating system tasks at the same time is the (imo) logical process, but otherwise I'll close this for now.

njsmith:
It's true it would be easier to review as two separate changes. I think they should be mostly orthogonal anyway?

OTOH the big advantage of the "one release of deprecation" model is indeed that it lets you immediately rip out the old stuff. Like discussed in https://github.com/python-trio/trio/issues/1#issue-202378594, there is a real cost to more complex deprecation processes, and if that's helping our users then that's fine, but we should also be careful not to be making unnecessary work for ourselves. We have no shortage of necessary work :-).

I get why it would make sense to have a minimum waiting period between the release deprecating something and the release removing it. I don't understand why it should also depend on how many releases happen in between – if something is deprecated for, say, a month, then anyone who upgrades during that period will see the warnings, and anyone doesn't, won't. So why does it matter whether a second release happened in week two or not? But @Fuyukai I know you [suggested it in the first place](https://gitter.im/python-trio/general?at=5ac9fc6f109bb043329d912f) and now @smurfix, you're keen on it too, so maybe I'm missing something.

smurfix:
"Deprecate for a month" doesn't work if there was no release within that month, because those of our users who get the code from PyPI won't see it – and there's going to be more of these as Trio gains traction.
"Deprecate, release, wait one month, then rip out the code" would work for me. I don't think we need to go the PendingDeprecation route Python itself uses … at least not yet. ;-)

jab:
Quoting https://semver.org/#spec-item-4:
> 4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

Based on this, as long as Trio is pre-1.0, I think the Trio maintainers should do whatever is the least work to move the project forward. So e.g. intervening releases, deprecation warnings, mandatory waiting periods, etc. are all more than users should expect. And for whatever it's worth, this particular user would rather see maintainers have less work at this point in the project so that it can realize its ultimate potential all the sooner.

smurfix:
Well, finding and removing deprecated code isn't any more or less work whether you do it sooner or later. So yes, semver says 0.x.y versions may change basically whenever, but that doesn't prevent us from adopting slightly stronger stability guidelines.

njsmith:
> "Deprecate for a month" doesn't work if there was no release within that month

Right, if we have a "deprecate for X time" policy, then the timer should start counting from the first release that includes the deprecation, not from the first commit that includes the deprecation.

> finding and removing deprecated code isn't any more or less work whether you do it sooner or later.

There is some overhead to having deprecated code hanging around the codebase though, and the bookkeeping to keep track of which stuff can be removed when is overhead too. (The more times we go "hey can we remove this? Oh not yet", the more attention we're wasting.) It's not necessarily a big deal, but like here we've just had a contributor's PR closed because of it; Laura is doing extra work because of this policy right now!

> server

For this project in particular, our challenge is to become a stable foundation library as quickly as we can. I would love it if we could release a 1.0 within, say, a year. But to do this we need more users to gain confidence in our API, which means that between now and 1.0 we'll need to do a juggling act where as more users gradually come on board, we gradually get stricter about stability so that the next round of users feels comfortable joining us... So it's a little more complex than "semver says that anything goes".

smurfix:
> server
semver ;-)

> like here we've just had a contributor's PR closed because of it;
> Laura is doing extra work because of this policy right now!

Except that, well, the PR had other issues …

As the release manager, I'm prepared to do that go-through-deprecations work.

Is everybody OK with setting that timer to a month, for now?

smurfix:
@Fuyukai even if this particular PR didn't work out, thanks for doing busywork like going through issues that may or may not be closed by now!

njsmith:
> Is everybody OK with setting that timer to a month, for now?

Works for me. Let's move the discussion over to #500.

PR #502, opened 2018-04-17
Improve trio.Path.iterdir

### LINKED ISSUE #501: trio.Path.iterdir wrapping is broken
Given `pathlib.Path.iterdir` returns a generator that does IO access on each iteration, `trio.Path.iterdir` is currently broken given it currently only generates the generator asynchronously (which I suppose is pointless given there is no need for IO at generator creation)

The solution would be to modify  `trio.Path.iterdir` to return an async generator, however this means creating a special case given the current implementation is only an async wrapper on `pathlib.Path.iterdir`.

touilleMan:
Closes #501

### DIFF
--- a/trio/_path.py
+++ b/trio/_path.py
@@ -128,6 +128,19 @@ class Path(metaclass=AsyncAutoWrapperType):
 
         self._wrapped = pathlib.Path(*args)
 
+    async def iterdir(self):
+        """
+        Almost (see below) like :meth:`iterdir.Path.iterdir`, but async.
+
+        Unlike :meth:`iterdir.Path.iterdir`, this method loads the entire
+        directory in memory once, then iter over it.
+        """
+        def _load_items():
+            return list(self._wrapped.iterdir())
+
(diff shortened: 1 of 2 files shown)


trio/_path.py
```diff
@@ -128,6 +128,19 @@ def __init__(self, *args):
 
         self._wrapped = pathlib.Path(*args)
 
+    async def iterdir(self):
+        """
+        Almost (see below) like :meth:`iterdir.Path.iterdir`, but async.
+
+        Unlike :meth:`iterdir.Path.iterdir`, this method loads the entire
```
Fuyukai: Did you mean `pathlib.Path.iterdir`?

touilleMan: Good catch ^^

njsmith:
Code looks good to me. I edited the docstring, and also edited your initial comment to say "closes #501 ", so when this is merged it will automagically close the issue.

Please go ahead and merge after confirming you like my edited docstring :-)

touilleMan:
Your comments&docstrings are always the best anyway ;-)

PR #503, opened 2018-04-17
Switch to using standalone Outcome library.

### LINKED ISSUE #494: Use standalone result library instead
As said in #477, the Result objects got split out into a separate library (https://github.com/python-trio/outcome). There's no point duplicating all the code in Trio - better to depend on it directly.

Todo:  
 - [ ] Gut result code inside of Trio  
 - [ ] Use the results from `outcome` instead.

Maybe make the current objects in Trio simply redirect to the outcome objects until the next version, deprecating them in 0.5 (with a redirect) and removing in 0.6.

### LINKED ISSUE #505: Trio fails when using single-use Outcome
```
_________________________________________ test_epoll_statistics _________________________________________

self = Error(AlreadyUsedError())

    def unwrap(self):
        self._set_unwrapped()
>       raise self.error

../outcome/src/outcome/_sync.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/lib/python3/dist-packages/contextvars/__init__.py:66: in run
    return callable(*args, **kwargs)
trio/_core/_run.py:938: in init
    self.entry_queue.spawn()
/usr/lib/python3/dist-packages/async_generator/_util.py:42: in __aexit__
    await self._agen.asend(None)
/usr/lib/python3/dist-packages/async_generator/_impl.py:274: in asend
    return await self._do_it(self._it.send, value)
/usr/lib/python3/dist-packages/async_generator/_impl.py:290: in _do_it
    return await ANextIter(self._it, start_fn, *args)
/usr/lib/python3/dist-packages/async_generator/_impl.py:202: in send
    return self._invoke(self._it.send, value)
/usr/lib/python3/dist-packages/async_generator/_impl.py:209: in _invoke
    result = fn(*args)
trio/_core/_run.py:313: in open_nursery
    await nursery._nested_child_finished(nested_child_exc)
trio/_core/_run.py:411: in _nested_child_finished
    await wait_task_rescheduled(aborted)
trio/_core/_traps.py:159: in wait_task_rescheduled
    return (await _async_yield(WaitTaskRescheduled(abort_func))).unwrap()
../outcome/src/outcome/_sync.py:84: in unwrap
    self._set_unwrapped()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Value(None)

    def _set_unwrapped(self):
        if self._unwrapped:
>           raise AlreadyUsedError
E           outcome.AlreadyUsedError

```

Fuyukai:
Closes #494. Closes #505.

Not sure what to do about the Result object section in hazmat docs yet.


trio/_core/_result.py
```diff
@@ -1,125 +1,29 @@
...
 
+__all__ = ["Result", "Value", "Error"]
 
-@attr.s(frozen=True, repr=False)
-class Error(Result):
-    """Concrete :class:`Result` subclass representing a raised exception.
 
-    """
+class Result(outcome.Outcome):
+    @deprecated(version="0.5.0", issue=494, instead="outcome.Outcome")
```
oremanj: `Result` was previously an ABC, so I wouldn't expect anyone to be trying to construct it; I think the deprecation notice would need to go on capture(), acapture(), and the constructors of Value and Error in order to have any effect.

njsmith: For `Value` and `Error` I'd suggest using `trio._deprecate`'s trick for module level deprecations. It lets you say "when someone accesses `trio.hazmat.Value`, then give them [this] deprecation message and then return [this] object." It looks like the diff for 089c6ff02e624190d4cd3c4f3f2543cabcddfdec has an example.

It might even be simplest to use this for `Result` too. The advantage is that it gives a warning for `isinstance(x, trio.hazmat.Result)`.

njsmith:
> Not sure what to do about the Result object section in hazmat docs yet.

You can delete the documentation of the `Result` objects themselves, and update the docs as if we always used `outcome`. (So e.g. `wait_task_rescheduled`'s docstring will need some changes.)

Fuyukai:
So, this doc.py nitpicky thing isn't very sustainable since we have to add every deprecated item there as it's in the changelog by a reference still.

njsmith:
I think it would also be fine to convert the old changelog links into regular literal strings, if that's easier? (I know it feels weird to edit old changelogs, but it's just a formatting tweak, not a content change.) Unfortunately without the nitpicky thing it's very very easy to have broken links all over the place :-/.

smurfix:
+1 on converting old changelog entries to literals. The link is dead, no sense in keeping it.

Fuyukai:
Shoutouts to jenkins for arbitrarily refusing to install dependencies sometimes

Fuyukai:
The build is failing because Travis hates me, I guess. Other than that, is all ready.

njsmith:
The build failure was sphinx trying to fetch the intersphinx inventories, and then failing with the gnomic error message:

```
Warning, treated as error:
failed to reach any of the inventories with the following issues:
```

Seems like some weird transient thing... I hope? I just restarted the job anyway.

njsmith:
Hmm. It failed again, while `loading intersphinx inventory from https://outcome.readthedocs.org/objects.inv...`, and indeed if I go to that URL I get a 404...

Maybe if you make the intersphinx link be `https://outcome.readthedocs.io/en/latest/` it'll work better?

trio/_core/_result.py
```diff
@@ -1,125 +1,48 @@
...
+        ),
+    "Value":
+        _deprecate.DeprecatedAttribute(
+            Value, "0.5.0", issue=494, instead="outcome.Value"
+        ),
+    "Error":
+        _deprecate.DeprecatedAttribute(
+            Error, "0.5.0", issue=494, instead="outcome.Error"
+        )
+}
```
njsmith: Sorry, this `__deprecated_attributes__` thing is pretty confusing, we need some better docs here...

Since it's based on module attribute lookup, using this inside `trio._core._result` means that we'll issue warnings when someone mentions `trio._core._result.<whatever>`, which isn't really what we want – it's references to `trio.hazmat.<whatever>` that people might doing and we want to deprecate. So we want to apply the `__deprecated_attributes__` stuff to `trio.hazmat`. In the version we used to have here we did that in `trio/__init__.py`, which seems as good as anything: https://github.com/python-trio/trio/commit/089c6ff02e624190d4cd3c4f3f2543cabcddfdec#diff-9bcb0dac3a23179589f66a8646fa38faL107

Also, this isn't really relevant because of the previous issue, but as an FYI: I think this also wouldn't work because the `__deprecated_attributes__` stuff is only checked if there isn't a real variable of the same name. (This is an optimization to avoid slowdowns on accessing non-deprecated attributes; it works using `__getattr__`.) So here where are actual variables named `Value`, `Error`, `Result` in the same module, it makes `__deprecated_attributes__` a no-op.

njsmith:
So having read over the diff, I think the complete todo list is:

* intersphinx url
* fixing `__deprecated_attributes__`
* figuring out the plan to handle #505

Fuyukai:
#505 was caused by The Curse of the Mutable Default. `reschedule()` had a default value for next_send of Value(None), which was being passed if there was no send - and promptly being unwrapped on the very first run loop. Then, when it was being yielded back up later on, it would be unwrapped again, and would explode.

Fuyukai:
Ready to merge.

trio/_core/_result.py
```diff
@@ -30,19 +30,4 @@ def capture(cls, sync_fn, *args):
...
-    "Value":
-        _deprecate.DeprecatedAttribute(
-            Value, "0.5.0", issue=494, instead="outcome.Value"
-        ),
-    "Error":
-        _deprecate.DeprecatedAttribute(
-            Error, "0.5.0", issue=494, instead="outcome.Error"
-        )
-}
+Result.register(Error)
```
smurfix: Ugh. Could you please keep the newline?

Fuyukai: Whoops, missed that. Surprised yapf didn't catch it.

njsmith:
I filed #508 for the annoying jenkins failures, and once #509 is merged it might help us track them down.

Issue #508, opened 2018-04-21
'pip install' on our MacOS python 3.5 CI is flaky


njsmith:
Example:

https://ci.cryptography.io/blue/organizations/jenkins/python-trio%2Ftrio/detail/PR-503/12/pipeline/

```
Collecting pyOpenSSL (from -r test-requirements.txt (line 4))

  Could not find a version that satisfies the requirement pyOpenSSL (from -r test-requirements.txt (line 4)) (from versions: )
```

Example:

https://ci.cryptography.io/blue/organizations/jenkins/python-trio%2Ftrio/detail/PR-503/8/pipeline/

```
Collecting pytest>=3.3 (from -r test-requirements.txt (line 1))

  Could not find a version that satisfies the requirement pytest>=3.3 (from -r test-requirements.txt (line 1)) (from versions: )
```

This is the same kind of message that you get when using an old pip on an old MacOS, and it can't handle PyPI's new TLS requirements. And it's suspicious that it seems to happen on the Python 3.5 builder only (so far). However... in both of those runs you can see that the first thing we do is use `get-pip.py` to upgrade to pip 10.0.1. The log verifies it:

```
++ pip --version
pip 10.0.1 from /Users/jenkins/workspace/-503-TALZSOV5BV3A6O3/trio/.venv/lib/python3.5/site-packages/pip (python 3.5)
```

Which makes sense, because if we didn't have working TLS, then we would be 100% unsuccessful at downloading things. Here we're only... intermittently, apparently at random, unsuccessful at downloading things.

I guess the first thing to do is to add some `-vvv` switches to the pip calls and wait for it to happen again...

dstufft:
That was only the behavior of pip during the brownouts, since we're at 100% unavailability, we were able to have Fastly hard fail old TLS versions, and they give a OpenSSL error message now instead of just not being able to find versions.

njsmith:
@dstufft Thanks for the information. That's mysterious, though... hopefully it'll happen again soon and we'll get more info. In the mean time, should I be filing a bug on `pip` to give better error messages when there's some kind of communication failure? That we're getting this weird "(from versions: )" message from two different issues is unfortunate...

njsmith:
Okay, it happened again, this time with `pip -vvv`: https://ci.cryptography.io/blue/organizations/jenkins/python-trio%2Ftrio/detail/PR-512/1/pipeline

Relevant logs:

```
Collecting idna (from trio==0.4.0+dev)
[...]

PR #509, opened 2018-04-21
[CI] Temporarily increase pip verbosity during CI runs

njsmith:
Attempting to debug gh-508


njsmith:
I've visually confirmed that this is successfully causing pip to dump lots of debug output on the MacOS builders (and probably travis too, but I don't care about that).

Fuyukai:
I like how codecov decided this increased coverage on a file that wasn't even touched.

PR #465, opened 2018-03-13
docs: Bump year automatically

pquentin:

PR #510, opened 2018-04-23
Docs bump year

pquentin:
As a follow-up to #465 after @Fuyukai's https://github.com/python-trio/trio/pull/465#issuecomment-379252669.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -84,7 +84,7 @@ master_doc = 'index'
 # General information about the project.
 project = 'Trio'
 year = datetime.datetime.utcnow().year
-copyright = '{}, Nathaniel J. Smith'.format(year)
+copyright = '2017–{}, Nathaniel J. Smith'.format(year)
 author = 'Nathaniel J. Smith'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -17,6 +17,7 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
+import datetime
 import os
 import sys
 # For our local_customization module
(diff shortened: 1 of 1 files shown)


njsmith:
I'm sorry I've been procrastinating so much on making a decision here! I am being indecisive because this is obscure legal stuff that probably doesn't matter too much or... does it? Well, probably not enough to actually talk to a lawyer. But then I have to decide! Etc. **But** that's still no excuse to leave you hanging.

So, as far as I understand:

* Copyright notices don't really... do that much. They don't need to be present, or accurate, for copyright to apply. But they can give you a somewhat better position in court if you do have to go to court, and they're cheap, so people generally use them: https://en.wikipedia.org/wiki/Copyright_notice#Reasons_to_include_an_optional_copyright_notice
* Once a copyright notice is added, they can never ever be removed (b/c all the standard OSS licenses require that they be kept as a condition of having a license at all). Technically this even applies to copyright statements that are flat-out inaccurate. Unless the courts decided it didn't, I guess, because of some public interest consideration or something, look I'm not an IP lawyer.
* There is a rule for what they're supposed to look like: https://en.wikipedia.org/wiki/Copyright_notice#Form_of_notice_for_visually_perceptible_copies
  * Basically: "ⓒ [year] [owner]", where "year" is the "year of first publication", which is awkward here because it was originally intended for like, books, which are generally published once and then never changed. I think for our purposes basically any content changes count as a "new publication". Maybe?

Options:

* Merge this. I have some nervousness that automatically changing the year will lead to an incorrect notice that can never be removed. I guess in fact it's sort of inevitable since all software eventually becomes unmaintained. OTOH that's a long time from now and by definition will be someone else's problem. Also, seriously Nathaniel no-one cares about these technicalities. Practically every package on PyPI has broken licensing because [setuptools doesn't provide any way to comply with a term that's present in every OSS license](https://github.com/pypa/packaging-problems/issues/41). This has literally zero practical consequences.

* Decide to just keep it 2017 forever because again, who cares, and that's certainly not wrong. This is also a very popular strategy (more because of laziness than intention).

* Put something like "Copyright Ⓒ 2017 and later, Nathaniel J. Smith and contributors", and figure that that will be accurate forever.

smurfix:
> Put something like "Copyright Ⓒ 2017 and later, Nathaniel J. Smith and contributors", and figure that that will be accurate forever.

+1

PR #511, opened 2018-04-27
Add flake8 to the testing

cclauss:
Import os, sys, or socket would solve several of these _undefined name_ issues.

Output: https://travis-ci.org/python-trio/trio/jobs/372189626#L3183

### DIFF
--- a/ci/travis.sh
+++ b/ci/travis.sh
@@ -46,7 +46,12 @@ alias pip="pip -vvv"
 pip install -U pip setuptools wheel
 
 if [ "$CHECK_FORMATTING" = "1" ]; then
-    pip install yapf==${YAPF_VERSION}
+    pip install flake8 yapf==${YAPF_VERSION}
+    # stop the build if there are Python syntax errors or undefined names
+    - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+    # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+    - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+
     if ! yapf -rpd setup.py trio; then
         cat <<EOF
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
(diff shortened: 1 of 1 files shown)


njsmith:
Hey, thanks for the contribution!

In general I have very mixed feelings about flake8; I feel like when I've contributed to other projects that enforced it, then it often became a frustrating roadblock. Obviously I like tools that help us increase quality, but I don't want to frustrate contributors with nitpicking and false positives.

Are any of flake8's complaints here valid, or are they all false positives? From a quick look at the log they seem to mostly involve it getting confused by `import *`...

cclauss:
I agree with you...  E901,E999,F821,F822,F823 are the "showstopper" flake8 issues that can halt the runtime with a SyntaxError, NameError, etc. The other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.  Pay attention to the showstoppers.

njsmith:
I'm not as much interested in flake8 message categories in the abstract; I'm interested in whether it helps us find actual bugs in this actual project :-). So my question is, are any of the errors/warnings it gives here real issues? My impression from a quick skim is that they're all noise, but I haven't gone through systemically.

cclauss:
__1     E999 IndentationError: expected an indented block__
* This is a syntax error that will halt your program — it should be fixed

__27    F821 undefined name 'os'__
* Each of these has the potential to raise NameError at runtime halting your program.  You can ignore the first one about version but the other 26 should be fixed

__47    F822 undefined name 'Cancelled' in `__all__`__
* These are promises that are being made but that can not be fulfilled.  If anyone trys to access these variables then they will probably halt the program with a NameError

njsmith:
Looking more closely, the `indented block` and `undefined name 'os'` are both in the `notes-to-self/` directory, which as the name suggests is just random notes that isn't expected to necessarily be running code. We shouldn't run linters/formatters/etc. on this directory. (Notice that the existing yapf CI just runs on `setup.py trio/`.)

Most of the undefined name warnings in the actual code base are due to flake8 not being clever enough to understand `import *`. We have >99% test coverage, so if we were actually referencing dozens of undefined names all over the codebase, then we would know it :-).

```
./trio/_core/_io_windows.py:304:26: F821 undefined name 'obj'
        handle = _handle(obj)
                         ^
./trio/_core/_io_windows.py:310:26: F821 undefined name 'obj'
        handle = _handle(obj)
```

These two appear to be (the only?) real issues. (It should be `handle = _handle(handle)`.) They're in some [undocumented/placeholder methods](https://trio.readthedocs.io/en/latest/reference-hazmat.html#windows-specific-api) that don't have tests yet.

Anyway, if that's right then it suggests this flake8 configuration has a >97% false positive rate on our codebase, which is way too high to be useful :-(.

PR #512, opened 2018-04-28
Import os, sys

cclauss:
Discovered via #511


njsmith:
This code isn't intended to be a working script; it's more like an example snippet you'd see in docs or something, where these imports are a distraction.

It looks like this PR did manage to trigger #508 though, with our new more verbose pip logging! That's helpful.

PR #513, opened 2018-04-28
Clear undefined names by using _stdlib_socket

cclauss:
Discovered via #511


njsmith:
This code was already working, and this PR breaks it :-(. It also [introduces style issues](https://travis-ci.org/python-trio/trio/jobs/372319164). And even if these are fixed, it won't actually make the code any easier to read, work with, more correct, etc.

I really appreciate your enthusiasm, but I feel like there are more meaningful things you could be spending this energy on?

PR #517, opened 2018-05-05
Factor Python 3.5.2 check out of aiter_compat implementation

jab:
Just happened to notice the Python 3.5.2 check could be moved out of every aiter_compat-decorated function call. In case it's worth it!

### DIFF
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -74,16 +74,16 @@ else:
 
 
 # Decorator to handle the change to __aiter__ in 3.5.2
-def aiter_compat(aiter_impl):
-    if sys.version_info < (3, 5, 2):
+if sys.version_info < (3, 5, 2):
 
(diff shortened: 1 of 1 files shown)


njsmith:
Build failures are (a) #508, which hopefully has been worked-around by the just-merged #519, (b) [yapf complaining that you forgot to run it :-)](https://travis-ci.org/python-trio/trio/jobs/375330150).

Re: the actual change: I don't think this matters a whole lot either way, because `@aiter_compat` is only called a few times at startup, but sure, it's not really any more complicated and I guess it's a bit faster, so why not. You do need to run yapf though :-)

jab:
> You do need to run yapf though

Sorry about that, fixed in acd758c. Looks like we're all green now.

njsmith:
Thanks!

Issue #275, opened 2017-08-09
Allow setting source interface in open_tcp_stream


njsmith:
Right now `open_tcp_stream` has no option to control which outgoing interface the connections are made from. We should make this possible.

Subtleties:

* It needs to be possible to set this separately for AF_INET and AF_INET6
* We can't support setting the actual outgoing port, because we might need to make multiple outgoing connections in parallel

On Linux 4.2+, there's `sock.setsockopt(IPPROTO_IP, IP_BIND_ADDRESS_NO_PORT, 1); sock.bind((address, 0))` which means "use this host, but delay picking the port until I call `connect`". We should use this if available. We should *not* use `SO_REUSEADDR`, because then it's possible to bind to a port that will fail when we actually call `connect` (because there might be a TIME_WAIT with the same 4-tuple).

So maybe something like: `source_host={AF_INET: "...", AF_INET6: "..."}`?

Should we also allow `source_host="1.2.3.4"` and auto-expand it into `{AF_INET: "1.2.3.4"}`?

If one is not specified, what happens? Maybe we should treat `INADDR_ANY` as meaning "you can use this protocol, and I don't care what host you originate from", and missing as meaning "don't use this protocol"? That could be nice since it would also give a way to say "I just want you to use IPv6", albeit not an obvious one.

njsmith:
Question: how does `IP_BIND_ADDRESS_NO_PORT` work with IPv6? it's only documented as an `IPPROTO_IP` option, but the [patch cover letter](https://www.mail-archive.com/netdev@vger.kernel.org/msg66879.html) says it works on IPv6; maybe you can just use `IPPROTO_IP` on IPv6 sockets, at least in this case? Should check this.

#39 has some discussion of SO_REUSEPORT details, for reference.

njsmith:
Here's a possible API that might be a bit simpler to use: allow specifying a list of source addresses, and each socket walks down the list and uses the first one that it can.

So `["1.2.3.4", "1:2:3::4"]` would use `1.2.3.4` for outgoing IPv4 connections, and `1:2:3::4` for outgoing IPv6 connections, with random ports. (Using `IP_BIND_ADDRESS_NO_PORT` if available, otherwise falling back on port 0.) Specifying `["1.2.3.4"]` would force the use of IPv4 for this connection.

[...]

PR #518, opened 2018-05-05
Support binding to a source address.

miracle2k:
I want to set the source ip address for a request. I see that there is a TODO in the code. This seems to work for me, which seems almost too easy. Is there something else?


njsmith:
There's also an open issue at #275, which has a bit more discussion of the subtleties here. This works in simple cases, but it forces a choice of IPv4/IPv6 (whichever one you use as the source), and there's some problem around picking a port given that we might be making multiple connections... let's move the discussion there.

PR #519, opened 2018-05-05
Attempt to workaround pip cache race condition on Jenkins (gh-508)

njsmith:


Fuyukai:
Codecov seems to really hate that file.

njsmith:
From inspecting the build output this seems to have done what I intended, and since it's a hacky workaround for an intermittent issue there's no way to know if it works except to merge it and wait, so I guess let's do that.

PR #520, opened 2018-05-05
Make system tasks live inside their own contextvars context.

### LINKED ISSUE #289: spawn_system_task should not propagate TaskLocals

Fuyukai:
Closes #289. 

Re-do of my previous PR, which was a bit too ambitious.


njsmith:
In case you were curious, the weird yapf stuff is because yapf uses the presence of a trailing comma on the last item as a hint that you want vertical formatting.

newsfragments/289.feature.rst
```diff
@@ -0,0 +1 @@
+Make system tasks run in their own context defined at the ``trio.run`` call.
```
njsmith: How about:

```
Tasks created by :func:`~trio.hazmat.spawn_system_task` now no longer inherit the creator's :mod:`contextvars` context.
```

?

Fuyukai: I wasn't sure what to put for the note, so this sounds a lot better.

PR #521, opened 2018-05-05
Make the default resolver more ABC-compliant.

Fuyukai:
This makes the default threaded resolver a class with an instance, rather than magic fallbacks.

I think I mentioned this on Gitter, but it's being a pain and won't search properly.


Fuyukai:
Given that there's a mechanism to define real objects that can be used to change the hostname resolving machinery, making the functions fall back on a special case seems wrong. This just changes it so that the fallback is compliant with the machinery that normal APIs would have to go through.

Issue #522, opened 2018-05-06
Coverage flapping in test_highlevel_open_tcp_listeners.py


njsmith:
There's a line in `test_highlevel_open_tcp_listeners.py` that codecov keeps complaining about, for example here (and in lots and lots of PRs): https://codecov.io/gh/python-trio/trio/pull/519/changes

The code is:

```python
            if cancel_scope.cancelled_caught:
                break
            if len(client_streams) >= limit:  # pragma: no cover
                return "lots"
    finally:
        for client_stream in client_streams:  # <---- THIS LINE
            await client_stream.aclose()

    return len(client_streams)
```

and the marked line was being reported as "partially covered". This was baffling to me, because sure, `for` loops have a branch: sometimes it executes the loop body, and sometimes it hits the end of the iterator and jumps ahead to whatever comes after the loop body. But here it was clearly doing both of those things.

So I hassled @nedbat on IRC, and he very helpfully explained: it's because of that `return "lots"` line up above. What, really? Yeah. After the `for` loop finishes, that also finishes the `finally` block, and then Python has decide where to go next. There are three possibilities:

* If the `try` block exited normally, then we move on to the next line (`return len(client_streams)`). Coverage.py saw this happen.
* If the `try` block exited with an exception, then we re-raise the exception. I guess coverage.py didn't see this happen – but coverage.py's branch coverage tracking doesn't count raising an exception as a branch, because otherwise you'd have to treat this as a possible outcome at basically every line in Python, and that would make its report useless.
* If the `try` block exited with a `return`, then after the `finally` block we immediately return, instead of executing the next line. Coverage.py didn't see this happen, and it does count it as a branch, so that was the problem.

This is quite surprising, especially since the `return` itself is marked "no cover" (which was our first attempt to solve this coverage flapping problem!), but it does make sense once you see it :-).

A major clue was clicking through codecov to find the actual XML files that coverage.py generated, where this line is described as:

```xml
<line branch="true" condition-coverage="66% (2/3)" hits="1" missing-branches="exit" number="84"/>
```

so the clues here are: (1) this `for` has *three* possible exits, instead of the normal two, and (2) the missing branch was an "exit", which I guess means an exit from the function.

Anyway, the solution is pretty easy, we should just mark more of this function as `no cover`, because it's just a silly little utility anyway. But I wanted to write this down because it was bugging me, and maybe we'll run into something like this again later.

Oh yeah and also we should fix it.

PR #523, opened 2018-05-06
Fix coverage flapping in test_highlevel_open_tcp_listeners

njsmith:
Fixes: gh-522


trio/tests/test_highlevel_open_tcp_listeners.py
```diff
@@ -81,7 +81,9 @@
             if len(client_streams) >= limit:  # pragma: no cover
                 return "lots"
     finally:
-        for client_stream in client_streams:
+        # The need for "no cover" here is subtle: see
+        # https://github.com/python-trio/trio/issues/522
+        for client_stream in client_streams:  # pragma: no cover
```
nedbat: You might prefer `#pragma: no branch`, which indicates that the line is expected to be covered, but that one of its possible branches will be missing.

PR #525, opened 2018-05-10
Fix recv_exactly example

hjbyt:

PR #526, opened 2018-05-10
pip doesn't respect XDG_CACHE_HOME on MacOS; use PIP_CACHE_DIR instead

njsmith:
Closes: gh-508, again.


njsmith:
Okay, I checked the jenkins output and it's saying "no cache found" for all the pypi requests, so it looks like *now* this is working properly.

njsmith:
I'm going to bed, but if someone wants to merge this and then nudge #525 along while I'm asleep, feel free :-).

pquentin:
Nice catch!

PR #533, opened 2018-05-15
WIP: Tracepoints.

ssanderson:
This is a tentative port of some tracepoint based tools I've used in gevent to trio. I'm not sure if this is actually something you're interested in, or if there's a more useful powerful version of this that could be implemented using more of trio's internals.

At any rate, I was curious to see what an async port would look like, so I figured it can't hurt to hack out a prototype to see what people think.

### DIFF
--- a/trio/trace.py
+++ b/trio/trace.py
@@ -1,0 +1,92 @@
+"""
+Utilities for tracing and testing concurrent executions.
+
+This module exposes three functions: trace(), pause(), pausing(). Together,
+they can be used to force one task to pause from another task, which can be
+used for things like testing correct handling of cancellations and potential
+race conditions.
+
+Usage
+-----
+Suppose that we have two tasks, A and B, and we want to run some code in Task B
+while Task A is paused at a known location. We can achieve that goal using
+tools provided by this module by writing something like::
+
+async def task_A():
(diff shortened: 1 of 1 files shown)


ssanderson:
Example script that shows how you might use this:

```python
from trio import run, open_nursery, Event, sleep
from trio.trace import pausing, trace


e = Event()


async def foo():
    await trace("start of foo")
    print("in foo")


@run
async def main():
    print("starting")
    async with open_nursery() as n:

        n.start_soon(foo)

        async with pausing('start of foo'):
            print("in pausing block")

            await sleep(1)

            print("finishing pausing block")

        print("end of pausing block")
        await sleep(1)
```

auvipy:
@njsmith what do you think about this?

njsmith:
@auvipy It's a neat idea! For context, I should mention @ssanderson and I were talking about it some at PyCon, so he opened the PR to demonstrate. I guess though that it isn't an obvious fit for the Trio core library (as opposed to living inside your code, or in a utility library, etc.), and it isn't obviously The One True Way That Everyone Should Use, so I don't think we're going to merge it here. But you should totally use it if you find it useful, and let us know how it goes :-).

Issue #422, opened 2018-01-27
make SocketType.type do the right thing by default


njsmith:
Python 3.7 is fixing a long standing and annoying bug in the handling of socket object `type` attributes:

https://bugs.python.org/issue32331#msg308604

We have some delicate workarounds for this ourselves, but they're hidden because I wanted to see what the upstream API would look like. Now that we know what the fix will look like, we should copy it.

njsmith:
Whoops, right, that's my fault for being unclear, sorry!

And right, we don't want to get rid of `real_socket_type` entirely, because we still need to work on 3.5 and 3.6. Instead, the idea would be to make Trio's socket object – on all versions – act like Python's socket object acts in 3.7+. And conveniently, Python's socket object in 3.7+ sets its `.type` attribute to be the same thing that Trio is already calculating with its `real_socket_type` hack.

So basically this means defining the `.type` attribute on *trio* socket objects to be something like `real_socket_type(self._sock.type)`.

And then we could go through all the places in trio that currently use `real_socket_type` and switch them to using `trio_socket.type` (it looks like there are currently 3 places, and they all have a trio socket object available).

Does that make sense?

scottbelden:
Yes, that definitely makes sense. I'll see if I can put this together. Thanks!

PR #534, opened 2018-05-16
make SocketType.type match what it will be in python 3.7+

scottbelden:
Resolves https://github.com/python-trio/trio/issues/422

I'm not sure if the test is correct so please let me know if that (or anything else) needs to be modified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
+.pytest_cache/
 nosetests.xml
 coverage.xml
 
--- a/trio/_highlevel_socket.py
+++ b/trio/_highlevel_socket.py
@@ -5,7 +5,6 @@ from contextlib import contextmanager
 
 from . import _core
 from . import socket as tsocket
-from ._socket import real_socket_type
 from ._util import ConflictDetector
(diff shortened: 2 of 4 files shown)


njsmith:
> I'm not sure if the test is correct so please let me know if that (or anything else) needs to be modified.

The actual modified test here doesn't do a great job of testing that this is working on its own, but I think it's fine because we actually have lots of tests that depend on `sock.type` working correctly. If I take this change and undo the fix to `SocketType.type`, then I get: `31 failed, 391 passed`. So if we ever break this we'll definitely notice :-)

However, can you update the comment on `real_socket_type`, and maybe add a comment on the `type` property? The existing comment is wrong, plus we should note that we're emulating the python 3.7 behavior (and that this code can eventually be removed when we drop support for versions <3.7).

Otherwise looks good to me, thanks!

scottbelden:
Updated. Let me know if you would like it worded differently. I was sure if the comments about the `_SOCK_TYPE_MASK` should be kept...

njsmith:
Looks fine to me, thanks!

...and... welcome! :tada: :birthday: As per [our standard policy](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team), I'm sending you a Github invite now :-).

Issue #200, opened 2017-06-11
testing.py needs to get broken down into submodules


njsmith:
It's waaay out of control.

njsmith:
Regarding the weird pypy nightly freeze in test_local: I downloaded the pypy-c-jit-91601-609a3cdf9cf7-linux64 nightly, and have let it loop running the trio testsuite for the last a few hours on my laptop, and I haven't been able to reproduce the problem so far. (Though I did get two failures in `test_ki.py`, both on [this line](https://github.com/python-trio/trio/blob/63110690bd9566ac7ca17126291745ac66ebec96/trio/_core/tests/test_ki.py#L362), which is getting "did not raise KeyboardInterrupt". The test sessions otherwise finished normally.)

njsmith:
#119 is now fixed, I think / hope. Demoting it to "on the radar".

njsmith:
Got fed up and fixed #140 :-)

njsmith:
#140 [came back](https://codecov.io/gh/python-trio/trio/pull/292/changes). This makes no sense at all.

njsmith:
Here's a weird one: https://travis-ci.org/python-trio/trio/jobs/298164123

It looks like our test for CPython 3.6.2 on MacOS, one of our calls to the *synchronous, stdlib function* `SSLSocket.unwrap()` raised an `SSLWantWriteError`. Which should be possible for a synchronous call, I think? Maybe this is some weird intermittent bug in the stdlib ssl module?

njsmith:
Here's a new one, I guess somehow introduced by #358:
https://ci.appveyor.com/project/njsmith/trio/build/1.0.768/job/3lbdyxl63q3h9s21
#361 attempts a diagnosis and fix.

njsmith:
The [weird SSL failure](https://github.com/python-trio/trio/issues/200#issuecomment-342323185) happened again: https://travis-ci.org/python-trio/trio/jobs/311618077
Again on CPython 3.6 on MacOS.

Filed upstream as [bpo-32219](https://bugs.python.org/issue32219). Possibly for now we should ignore `SSLWantWriteError` there as a workaround.

njsmith:
Freeze on PyPy nightly, we'll see if it repeats: https://travis-ci.org/python-trio/trio/jobs/319497598

njsmith:
Sigh: https://github.com/python-trio/trio/issues/447

njsmith:
#508

[...]

PR #535, opened 2018-05-16
fix pygments style in docs

sorcio:
Fixing `pygments_style` according to https://github.com/rtfd/sphinx_rtd_theme/issues/633#issuecomment-387594617

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -126,7 +126,7 @@ language = None
 exclude_patterns = []
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+pygments_style = 'default'
 
 highlight_language = 'python3'
 


sorcio:
AppVeyor Python35-x64 Windows build failed ([see logs](https://ci.appveyor.com/project/njsmith/trio/build/1.0.1085/job/iy6v6weexe9kksx4)):

```
pip install codecov
Collecting codecov
  Downloading https://files.pythonhosted.org/packages/8b/28/4c1950a61c3c5786f0f34d643d0d28ec832433c9a7c0bd157690d4eb1d5f/codecov-2.0.15-py2.py3-none-any.whl
Requirement already satisfied: coverage in c:\python35-x64\lib\site-packages (from codecov)
Collecting requests>=2.7.9 (from codecov)
  Could not find a version that satisfies the requirement requests>=2.7.9 (from codecov) (from versions: )
No matching distribution found for requests>=2.7.9 (from codecov)
You are using pip version 9.0.1, however version 10.0.1 is available.
You should consider upgrading via the 'python -m pip install --upgrade pip' command.
```

Closing/reopening to see if the problem magically goes away.

smurfix:
You could just go to appveyor and click the Rebuild button …

sorcio:
@smurfix right. If I remember correctly though @njsmith suggested to close/reopen so the build logs are not lost and the issue is documented somewhere.

smurfix:
@sorcio Actually that turns out not to be an issue with Appveyor – they keep the logs of superseded attempts around and assign new build IDs.

I'll ask the Travis people to please do the same thing.

njsmith:
That appveyor failure makes no sense :-(. If it was on pip 10, and on MacOS, it'd be #508, but in this case I have no idea :-(. Well, I'll add it to #200 anyway...

I also don't understand at all what's going on with RTD and sphinx and pygments and why this patch would fix anything, but I agree that people are claiming that it helped them and it seems harmless otherwise, so I guess let's merge it and see what RTD does.

njsmith:
I'm looking at trio.readthedocs.io now, and it doesn't look like highlighting is fixed :-(

sorcio:
Nope :( Maybe we should revert this but I have no idea what I'm doing.

I think the issue is still the missing CSS. `_static/pygments.css` maybe? It's there but it's not included in the generated HTML document. But it's included unconditionally in the theme (https://github.com/rtfd/sphinx_rtd_theme/blame/master/sphinx_rtd_theme/layout.html#L47) and it's been since 0.3.0. Does that mean RTD is building docs with a very old version of the theme? But Trio explicitly includes the latest theme version (https://github.com/python-trio/trio/blob/master/ci/rtd-requirements.txt) which as of now is 0.3.1 on PyPI.

Other updates to the theme are missing, like alt text on the logo (https://github.com/rtfd/sphinx_rtd_theme/commit/ed48532da6b96c2f0ec621401eb76d1f0d10a9ac#diff-1336f80431f476a232e3c48abb277c51) so I'm pretty sure that we're still using an older theme version. I don't know why yet. I don't have a clear memory of this: did syntax highlighting ever work? I will investigate further later.

sorcio:
Ah, now I get it. `rtd-requirements.txt` is installed after RTD installs its requirements. At that stage `sphinx-rtd-theme<0.3` is installed, so when it comes to our local requirements an unspecified `sphinx-rtd-theme` is already satisfied and therefore nothing is installed. We can override this by pinning a more specific version (e.g. `>=0.3,<0.4` or `==0.3.1`). Or, even better, just ignore it altogether because RTD has already merged the fix that will change the default theme version to `<0.4` and it's a matter of days before it gets deployed and our theme will automatically be updated next time we build documentation.

sorcio:
RTD deployed with the updated theme and now we have syntax highlighting! 🎉 

BTW this possibly means that we can do away with some hacks and fixes (but I'm not going to look into that right now)

njsmith:
Well, I guess it's nice when ignoring a problem makes it go away :-) Thanks for following up!

PR #539, opened 2018-05-21
Minor doc fixes

hjbyt:


trio/_abc.py
```diff
@@ -486,10 +486,10 @@ class Listener(AsyncResource):
...
 
         Returns:
-          AsyncResource: an object representing the incoming connection. In
-              practice this is almost always some variety of :class:`Stream`,
-              though in principle you could also use this interface with, say,
-              SOCK_SEQPACKET sockets or similar.
+          AsyncResource: An object representing the incoming connection. In
+          practice this is almost always some variety of :class:`Stream`,
+          though in principle you could also use this interface with, say,
+          SOCK_SEQPACKET sockets or similar.
```
njsmith: Does this still render correctly with the changed indentation? I can never predict what Sphinx will do with these kinds of things...

hjbyt: Yes, i built the docs locally to make sure it is fixed. I could upload a pic of the fixed layout later.
This is the current layout to be fixed:
current_layout
(notice the first line of "Returns" section is in its own box, and the rest is idented)

njsmith: Haha, whoops, that does look terrible. Thanks :-)

PR #541, opened 2018-05-26
Use immutable default for argument

### LINKED ISSUE #514: Is trio.run(..., instruments) a mutable default argument?
https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L1133
https://docs.python-guide.org/en/latest/writing/gotchas

Would it be possible to have __instruments__ modified in one call to __trio.run()__ and still have those modifications present in a future call that did not set the __instruments__ arg?

Zac-HD:
This is not currently a problem, but fixing it makes it *obviously* safe and avoids future bugs.  Closes #514.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1140,7 +1140,7 @@ def run(
     async_fn,
     *args,
     clock=None,
-    instruments=[],
+    instruments=(),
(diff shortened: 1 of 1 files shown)


touilleMan:
thanks @Zac-HD  ;-)

PR #547, opened 2018-06-12
Raise a more explicit async generator error

### LINKED ISSUE #546: Erroring in Runner.spawn_impl() with an async generator should explain itself better
Ran into the situation where I made a mistake and called `nursery.start_soon(async_fn)` where `async_fn` was an async generator but got the following error:

```
# Give good error for: nursery.start_soon(some_sync_fn)
    raise TypeError(
         "trio expected an async function, but {!r} appears to be "
         "synchronous"
>               .format(getattr(async_fn, "__qualname__", async_fn))
 )
TypeError: trio expected an async function, but functools.partial(<function test_local_actor_async_gen.<locals>.print_loop at 0x7f45a6e78d90>) appears to be synchronous
```

Instead I propose having a separate explicit error message for the case where a user mistakenly passes an async generator:

```
TypeError: start_soon expected an async function but got an async generator <async_generator object test_local_actor_async_gen.<locals>.print_loop at 0x7f00a5504e80>
```

tgoodlet:
Alert the user that that passing an async generator to a task spawning method is not allowed.

Fixes #546

I wasn't sure where to add a test for this as it doesn't seem the slew of `TypeError`s and corresponding messages are verified anywhere?

### DIFF
--- a/trio/_core/tests/test_run.py
+++ b/trio/_core/tests/test_run.py
@@ -11,6 +11,7 @@ from math import inf
 import attr
 import outcome
 import pytest
+from async_generator import async_generator
 
 from .tutil import check_sequence_matches, gc_collect_harder
 from ... import _core
@@ -1589,6 +1590,15 @@ def test_nice_error_on_bad_calls_to_run_or_spawn():
             with pytest.raises(TypeError) as excinfo:
                 bad_call(len, [1, 2, 3])
             assert "appears to be synchronous" in str(excinfo.value)
+
+            @async_generator
+            async def async_gen(arg):  # pragma: no cover
+                pass
+
+            with pytest.raises(TypeError) as excinfo:
+                bad_call(async_gen, 0)
+            msg = "expected an async function but got an async generator"
+            assert msg in str(excinfo.value)
 
             # Make sure no references are kept around to keep anything alive
             del excinfo
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -14,7 +14,9 @@ from math import inf
 from time import monotonic
 
 import attr
-from async_generator import async_generator, yield_, asynccontextmanager
+from async_generator import (
+    async_generator, yield_, asynccontextmanager, isasyncgen
+)
 from sortedcontainers import SortedDict
 from outcome import Error, Value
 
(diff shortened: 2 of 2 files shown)


tgoodlet:
Heh, well the python 3.5 thing is unfortunate...

tgoodlet:
Regarding testing I'm assuming I should write a test somewhere in `_core/tests`?

njsmith:
Yeah, I'd add it next to the other tests for these error messages. The easy way to find them: comment out some of the existing checks and see which tests start failing :-).

I think you might want `isasyncgen` instead of `isasyncgenfunction`.

tgoodlet:
@njsmith oh right :+1: 

Ok made the change and added the test.
Let me know if there's anything else!

tgoodlet:
just as soon as I get it working on 3.5...

njsmith:
Yeah, you'll want to use an `async_generator`-based AG, instead of a native one.

tgoodlet:
@njsmith ahh ok will do

tgoodlet:
Dahh `yapf`.. so picky about your `formats` eh?

trio/_core/tests/test_run.py
```diff
@@ -1590,6 +1591,15 @@ def generator_based_coro():  # pragma: no cover
                 bad_call(len, [1, 2, 3])
             assert "appears to be synchronous" in str(excinfo.value)
 
+            @async_generator
+            async def async_gen(arg):
+                yield_(await sleep(arg))
```
pquentin: This line is never executed, you probably want to use the nocover + pass tricked as above.

tgoodlet: Gotcha. Pushed it. Hopefully that works.

trio/_core/tests/test_run.py
```diff
@@ -1590,6 +1591,15 @@ def generator_based_coro():  # pragma: no cover
                 bad_call(len, [1, 2, 3])
             assert "appears to be synchronous" in str(excinfo.value)
 
+            @async_generator
+            async def async_gen(arg):  # pragma: no cover
+                yield_(await sleep(arg))
```
pquentin: I guess using the `pass` statement would make it clear that this line is never executed?

pquentin: I guess using the pass statement would make it clear that this line is never executed?

tgoodlet:
Ahh, right. Ok adding `pass` as well!
This might get my top prize for most re-written commit ever lol.

pquentin:
Thank you @tgoodlet! Since this is your first contribution, you should [receive an invitation to the trio organization](http://trio.readthedocs.io/en/latest/contributing.html#joining-the-team) in a few days.

njsmith:
>  This might get my top prize for most re-written commit ever lol.

This is part of why we suggest small fixes to start with – it always takes a little while to adapt to a project's conventions :-).

Also, invitation sent! (One of these days we should make a bot or something for this, but for now I'm still doing it manually.)

PR #551, opened 2018-06-30
Fix link to node.js code review docs

njsmith:
They moved it.

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -451,7 +451,7 @@ time you review a PR, you want someone else to check over your work
 before you hit the merge button.
 
 For general tips on doing code reviews, the `node.js guide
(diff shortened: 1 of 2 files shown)


pquentin:
LGTM!

However to get the merge button, we have to fix this first! 🙂 

```Warning, treated as error:
/home/travis/build/python-trio/trio/trio/ssl.py:docstring of trio.ssl.SSLStream:92:py:meth reference target not found: ssl.SSLContext.set_servername_callback```

Actually it seems to exist: https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_servername_callback. Not sure what that means.

njsmith:
Added an unrelated commit that should hopefully fix the CI. (Well, both are trivial doc changes to update links, actually, so I don't even feel bad about putting them in the same PR.)

pquentin:
Thanks, merged!

> so I don't even feel bad

😄

Issue #315, opened 2017-09-05
Should we call the abort_fn when rescheduling in general?


njsmith:
Currently the most fundamental pattern for blocking in trio is:

* Register your task in one or more places where someone will eventually find it and call `reschedule`
* Write an abort function that knows how to unwind those registrations so that no-one will call `reschedule`
* Go to sleep
* Later, someone finds the task, and unwinds all its registrations so no-one else will call `reschedule`
* 

(with a few wrinkles I'm glossing over to support primitives with asynchronous cancellation like IOCP.)

While writing this: https://gist.github.com/njsmith/40b7b7f65e5f433789153c7b668ce643

...the

oremanj:
The discussion on #242 has been pretty fruitful in terms of coming up with a slightly higher-level API for common cases of blocking operations. If we do wind up adopting something like that for the trio core, I think remaining direct users of abort_fn will probably mostly want to do different things when aborting vs completing anyway, which would reduce the relevance of this issue. Definitely worth leaving it open until we see what happens with #242 though.

PR #553, opened 2018-07-04
Always force tasks off the queue

smurfix:
When queue.get() is cancelled, the task must not be left on the queue.
Otherwise an interesting crash will result when the next item is pushed.

Best demonstrated by:
```
    with trio.move_on_after(0.01) as timeout_scope:
        await q.get()
    assert timeout_scope.cancelled_caught
    await q.put("Test for PR #553")
```

### DIFF
--- a/trio/_sync.py
+++ b/trio/_sync.py
@@ -968,7 +968,10 @@ class Queue:
             return _core.Abort.SUCCEEDED
 
         self._get_wait[task] = None
-        value = await _core.wait_task_rescheduled(abort_fn)
+        try:
+            value = await _core.wait_task_rescheduled(abort_fn)
+        finally:
+            self._get_wait.pop(task, None)
         return value
 
     @aiter_compat
--- a/trio/tests/test_sync.py
+++ b/trio/tests/test_sync.py
@@ -5,6 +5,7 @@ import weakref
 from ..testing import wait_all_tasks_blocked, assert_checkpoints
 
 from .. import _core
+from .. import _timeouts
 from .._timeouts import sleep_forever
 from .._sync import *
 
@@ -406,6 +407,11 @@ async def test_Queue():
     with pytest.raises(_core.WouldBlock):
         q.get_nowait()
     assert q.empty()
+
+    with _timeouts.move_on_after(0.01) as timeout_scope:
+        await q.get()
+    assert timeout_scope.cancelled_caught
+    await q.put("Test for https://github.com/python-trio/trio/pull/553")
 
 
 async def test_Queue_iter():


sorcio:
@smurfix Ouch, sorry this made you waste time :( Thanks for adding a test and providing a fix!

I'm late for review but I want to add a note: I think that cleanup line pertains in the abort function, rather than in a finally clause. The abort function is only called on cancellation, while the finally clause is in the success path. And hopefully `wait_task_rescheduled()` only raises on cancellation.

smurfix:
Umm, no, the "finally" is in the "in any case" path. The await will raise an exception (most likely trio.Cancelled).
The proof is in the pudding, i.e. if you run the second patch without the first, the test will fail.
In fact I probably should use `except BaseException` instead of `finally` because int the success case the task is guaranteed to have been removed from the dict. That should be somewhat faster, given that `.pop()` is likely to raise+catch an internal `KeyError`.

sorcio:
@smurfix I meant that the `finally` sits in the way of the success path, because it's executed every time instead of just when it's needed. As you correctly say, this can make make the success case (slightly) slower.

What I suggest is:

```python
def abort_fn(_):
    self._get_wait.pop(task, None)
    return _core.Abort.SUCCEEDED

value = await _core.wait_task_rescheduled(abort_fn)
```

smurfix:
OK, thanks for clarifying. I didn't do that on purpose because there may be other exceptions (KeyboardInterrupt, anybody?) which would trigger the same problem.

njsmith:
@smurfix I see that @pquentin already reminded you about the don't-merge-your-own-PR guideline in chat. No harm done, but just want to mention it here too in case others watching get confused. As you can see, even a small and fundamentally correct fix like this still can have some fiddly bits to get right...

@sorcio is correct that `abort_fn` is the canonical place to do this kind of cleanup. There's even an issue open (#315) suggesting that it might be simpler to unconditionally call `abort_fn` when rescheduling, to reduce code duplication between the normal and cancellation wakeup paths...

Note also that `get` is decorated with `enable_ki_protection`, so the only way `KeyboardInterrupt` can be raised there is through the cancellation mechanism (i.e., `KeyboardInterrupt` can happen, but it's guaranteed to call `abort_fn`).

Also, in the test, that 0.01 second timeout is making me itch. I think it does deterministically work right now, but the only reason there's no race condition is because of some subtle details about how the body of `get` interacts with the scheduler... I think it would be clearer and more reliable to write it like:

```python
async def test_553(autojump_clock):
    q = trio.Queue()
    with trio.move_on_after(10) as timeout_scope:
        await q.get()
    assert timeout_scope.cancelled_caught
    await q.put("Test for PR #553")
```

njsmith:
> given that `.pop()` is likely to raise+catch an internal `KeyError`

Apparently not:

```
# Popping a key that's present
In [3]: %timeit d = {1: None}; d.pop(1, None)
187 ns ± 8.37 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

# Popping a key that's missing
In [4]: %timeit d = {1: None}; d.pop(2, None)
176 ns ± 3.32 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

PR #554, opened 2018-07-04
queue.get: Speed up the "no error" case

smurfix:
This is a follow-up to https://github.com/python-trio/trio/pull/553
The problem is that in the no-error case the task has already been removed, so the `.pop()` will always return its default anyway.

### DIFF
--- a/trio/_sync.py
+++ b/trio/_sync.py
@@ -965,13 +965,11 @@ class Queue:
         task = _core.current_task()
 
         def abort_fn(_):
+            self._get_wait.pop(task)
             return _core.Abort.SUCCEEDED
 
         self._get_wait[task] = None
-        try:
-            value = await _core.wait_task_rescheduled(abort_fn)
-        finally:
-            self._get_wait.pop(task, None)
+        value = await _core.wait_task_rescheduled(abort_fn)
         return value
 
     @aiter_compat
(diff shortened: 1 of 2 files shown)


trio/_sync.py
```diff
@@ -970,8 +970,9 @@ def abort_fn(_):
         self._get_wait[task] = None
         try:
             value = await _core.wait_task_rescheduled(abort_fn)
-        finally:
+        except BaseException:
             self._get_wait.pop(task, None)
```
sorcio: As stated in the discussion in #553, I think the place for this cleanup is inside `abort_fn` and we don't need a try block here.

smurfix:
fixed as per comments in #553

trio/_sync.py
```diff
@@ -965,13 +965,11 @@ def get_nowait(self):
         task = _core.current_task()
 
         def abort_fn(_):
+            self._get_wait.pop(task, None)
```
njsmith: This is a small nitpick, but: is the `None` necessary? I think it isn't, and if it isn't then I'd prefer not to have it here, because this code is already confusing and it's better not to trick the reader into thinking we're trying to handle a case that doesn't actually exist. (Also it could potentially hide a bug.)

Other than that, looks good to me, so feel free to merge after resolving that nitpick one way or the other.

njsmith:
:+1: thanks!

PR #556, opened 2018-07-05
Test Python 3.7 on Travis

RazerM:
- Use xenial on GCE for 3.7+
- Add 3.7 and 3.8-dev

I changed `pypy3.5-5.10.1` to `pypy3.5` (which happens to be 5.10.1 just now). Is there a reason it was pinned? The Travis PyPy documentation isn't clear on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,31 @@
 language: python
-python:
-  # The pypy tests are slow, so we list them first
-  - pypy3.5-5.10.1
-  - 3.5.0
-  - 3.5.2
-  - 3.6
-  - 3.5-dev
-  - 3.6-dev
-  - 3.7-dev
 sudo: false
 dist: trusty
 
 matrix:
   include:
-    - os: linux
-      language: generic
-      env: USE_PYPY_NIGHTLY=1
-    - os: linux
-      language: python
-      python: 3.6
+    - python: 3.6
       env: CHECK_DOCS=1
-    - os: linux
-      language: python
-      python: 3.6
+    - python: 3.6
       env: CHECK_FORMATTING=1
+    # The pypy tests are slow, so we list them first
+    - python: pypy3.5
+    - language: generic
+      env: USE_PYPY_NIGHTLY=1
+    - python: 3.5.0
+    - python: 3.5.2
+    - python: 3.6
+    - python: 3.7
+      dist: xenial
+      sudo: required
+    - python: 3.5-dev
+    - python: 3.6-dev
+    - python: 3.7-dev
+      dist: xenial
+      sudo: required
+    - python: 3.8-dev
+      dist: xenial
+      sudo: required
 
 script:
   - ci/travis.sh


njsmith:
> I changed pypy3.5-5.10.1 to pypy3.5 (which happens to be 5.10.1 just now). Is there a reason it was pinned? The Travis PyPy documentation isn't clear on this.

IME the Travis docs for their Python versions are constantly out of date and inaccurate. But there is a secret! If you run

```
s3cmd ls s3://travis-python-archives/binaries/ubuntu/16.04/x86_64/
```

then you can see the actual list of Python versions available: `python: XX` on a xenial build causes Travis to fetch `XX.tar.bz2` from that directory. (Or use `14.04` for trusty, etc.) From this we can see that currently, `pypy3.5.tar.bz2` exists and has the same size and timestamp as `pypy3.5-5.10.1.tar.bz2`, so `pypy3.5` is indeed an alias for `pypy3.5-5.10.1`. I suspect the reason for the pin was that last time I looked the alias was broken, or something like that.

It also looks like they have `3.7` as an alias for `3.7.0`, which is good, and that `3.8-dev` exists and is up to date, which is also good.

Why do the 3.7 and 3.8-dev versions require sudo? I'd expect that us to need sudo if we were using some hack to install them ourselves, but it looks like we're just using the same mechanisms as the other versions...

RazerM:
> Why do the 3.7 and 3.8-dev versions require sudo?

I'm not clear on that either. I saw it in https://github.com/travis-ci/travis-ci/issues/9815#issuecomment-401756442. I tried it without sudo and the jobs [failed](https://travis-ci.org/RazerM/outcome/jobs/400606832) when trying to download 3.7 for **14.04** rather than **16.04** (which was running).

RazerM:
And thanks for the s3 tip 😄

njsmith:
> I tried it without sudo and the jobs failed when trying to download 3.7 for 14.04 rather than 16.04 (which was running).

Okay, I dug into this some more and here's what I think is going on. [Let's look at this big table of their supported build environments.](https://docs.travis-ci.com/user/reference/overview/)

The first thing we notice is that 16.04 xenial isn't mentioned *at all*. Obviously they are working on it behind the scenes, given that they have Python builds available for it and everything, but using xenial is not actually a documented/supported option yet. We also see that they have two main platforms they run on – virtual machines on GCE, and container images on EC2. It looks like `sudo: required` is how you request to be routed to GCE, and also that GCE is the only platform where they support multiple operating systems. So my guess is that *if* we don't set `sudo: required`, then the `dist: xenial` part is simply ignored, and we're running on trusty in a container on EC2.

And in fact i fyou look at the [log that we got](https://travis-ci.org/RazerM/outcome/jobs/400606832), and click on the little arrow to expand "Worker information" we see that it's running on ec2, and if we click to expand "Build system information" it says "Build dist: xenial" which sounds good but then you look down a little further and find that under "Operating system description" it says we're actually running on 14.04 trusty.

And 3.7, 3.8-dev aren't supported on trusty, because it uses an ancient version of openssl – this part *is* documented on their python page, and can also be confirmed by checking the s3 bucket directly.

OKAY SO what does this mean for us. Obviously we have to use `sudo: required; dist: xenial` for our 3.7 and 3.8-dev builds, because that's the only thing that works.

There's no point in using `dist: xenial` *without* `sudo: required`, because that actually ends up running on trusty and it's just going to confuse us later. (Or possibly break stuff if they ever do roll out support for xenial on their container infrastructure.)

We want to avoid using `sudo: required` whenever possible, because if we consult that table again we see that (a) starting jobs is substantially slower (boot time: "20-50s" vs "1-6s"), and (b) IPv6 is disabled so all our IPv6 tests can't run.

CONCLUSION: For now I think we want to use `dist: xenial, sudo: required` for only the 3.7 and 3.8 jobs where we have no choice, and for all the other jobs we should stick with `dist: trusty`.

RazerM:
> if we click to expand "Build system information" it says "Build dist: xenial" which sounds good but then you look down a little further and find that under "Operating system description" it says we're actually running on 14.04 trusty.

I got as far as the "Build dist: xenial" part and assumed it was correct.

I'll fix it.

njsmith:
:+1:

PR #558, opened 2018-07-12
[WiP] PoC for IterableStream

alexchamberlain:
Just wanted to float the idea of an IterableStream; that is a stream for which you can iterate over using `async for`. It encourages people not to rely on a whole message arriving together as it exposes a single byte at a time.

Before I write docs etc, I was wondering what everyone thought of the principle of the class?

### Open Questions
- What type should be returned by `__anext__`? Naive solution exposes `int`, but should it be `bytes`?

### TODO

* [ ] Code formatting
* [ ] Documentation
* [ ] Think about tests; this has 100% coverage, but is it sufficient?

### DIFF
--- a/trio/_highlevel_generic.py
+++ b/trio/_highlevel_generic.py
@@ -165,3 +165,25 @@ class StapledStream(HalfCloseableStream):
             await self.send_stream.aclose()
         finally:
             await self.receive_stream.aclose()
+
+
+class IterableStream:
+    def __init__(self, stream, *, bufsize=16384):
+        self.stream = stream
+        self.bufsize = bufsize
+        self.data = iter(())
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        try:
+            return next(self.data)
+        except StopIteration:
+            self.data = await self.stream.receive_some(self.bufsize)
+            if not self.data:
+                raise StopAsyncIteration
+            else:
+                self.data = iter(self.data)
+
+            return next(self.data)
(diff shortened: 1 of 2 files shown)


smurfix:
Iterating on bytes wastes a lot of CPU. I wouldn't want to do that, at least not as default.

However, what I think would have merit is to add an async iterator to `ReceiveStream` which simply periodically calls `receive_some` (with some buffer size) and yields that. We could then layer other iterators on top which modify that (line-buffer, Unicode-decode, paragraph-block, header+value decode, etc.).

alexchamberlain:
I actually wrote this whilst writing a Redis protocol parser; it's basically new line delimited, except for 1 exception for raw byte data. When I wrote this, I was separating the concern of managing the buffer from iterating over it, but may have gone too far looking for an elegant solution and disregarded CPU usage. I'll go back to the drawing board and see if I can do a better job.

Maybe I can submit a line delimited stream parser instead.

tjstum:
@alexchamberlain There are probably a zillion I/O-less redis parsers (hiredis is one, I [wrote](https://gitlab.com/tjstum/rediparse) a pure python one). Maybe that would help your project, unless you were interested in writing one for your own edification!
But I was looking at implementing a Redis client using trio, so I would be happy to help collaborate on one!

alexchamberlain:
@tjstum Thanks for the feedback. I was writing a Redis clone to better understand async programming in trio, so wrote the parser for server side, rather than client side. Mainly a learning exercise.

alexchamberlain:
Closing this out; switching out my byte-based parser for basic receive_some loop and rediparse is 4x faster. Clearly, separating the network from protocol handling is the right way to go...

Thanks @smurfix and @tjstum for your feedback!

PR #572, opened 2018-07-28
Add support for sniffio

njsmith:
See https://github.com/python-trio/sniffio

And also https://github.com/vxgmichel/aiostream/issues/17

This shouldn't be merged (and tests will fail) until sniffio is
released on PyPI. But it demonstrates how to do it (and the tests pass
locally with my current sniffio checkout).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ setup(
         "async_generator >= 1.9",
         "idna",
         "outcome",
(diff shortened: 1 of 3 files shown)


njsmith:
Okay, sniffio is now up on pypi, so this should be ready to review/merge in a few minutes when the tests are finished...

Issue #497, opened 2018-04-14
We should have some kind of better closure tracking for Queues


njsmith:
Right now, if you want to do clean shutdown of a `Queue`, you have to do some messy thing where you pass in a sentinel value and the other side has to watch for it. This is inherited from Python's `queue.Queue`, but ... `queue.Queue` sometimes feels like a neat v1 that never got fully reviewed or polished. (We already got rid of some of its unwise decisions, see #322.)

For streams, there's a nice clear well-defined way to do this: the sender closes the stream, the receiver gets a notification (a special sentinel value in that case), any further attempts to send on the stream raise an error. Or, the receiver closers their end of the stream, and then any further attempts to send/receive raise an error.

For `Queue` it's a little more complicated, because we have the sending and receiving merged into a single object. The most important case is a sender notifying the receiver that they're done. So the simplest thing would be to have a `close` method that makes further calls to `get`/`put` raise an error, and `__anext__` to raise `StopAsyncIteration`, and document that generally it should be the sender that uses it.

(What if there are multiple tasks putting to the same queue? We could have `close` cause any blocked `put`s to immediately wake up (compare #460), or have an `aclose` that waits for any current `put` calls to finish.)

Here's another option, that's more disruptive but might ultimately be better: have the `put` and `get` sides be two different objects, so you'd do something like `put_queue, get_queue = trio.open_queue()`. `put_queue` would have the `put` methods, `get_queue` would have the `get` methods and async iteration protocol, and they could be closed independently, with the same detailed semantics as the stream APIs.

Maybe we'd even want to make an ABC for this, like `ObjectSendStream` and `ObjectReceiveStream` (names to be bikeshedded later); then people could make versions that use pickle to implement the interface on top of a `Stream`, etc. If doing this we might want to also change the names to better match the `Stream` conventions, like have the two main methods be `send` and `receive` instead of `put` and `get`.

But we shouldn't take this analogy *too* far... regular in-memory `Queue`s are often used in a way that byte streams aren't. Are there any cases where what you really want is a *bidirectional* `Queue`? If so then that would be a strong argument for splitting the two endpoints into two different objects... but I'm not sure if this really happens? Another way `Queue`s are different from streams is that fan-in and fan-out is common. For fan-in/fan-out, maybe we could do even better than having two objects: maybe we want to have lots of objects, one for every send/receiver! For example this might look like:

```python
async def sender(object_send_stream):
    async with object_send_stream:
        await object_send_stream.send(None)

[...]

PR #573, opened 2018-07-28
Add the ability to close a queue.

Fuyukai:


sorcio:
I'm not sure about this change. The one part I'm specifically worried about is what is supposed to happen when the queue still holds data in its buffer. In this implementation, the getters never get a chance to consume that data. I'm not entirely sure of the full range of use cases for this API, but I can imagine that some might instead want the queue to be closed on the consuming side only when it's empty.

smurfix:
@sorcio Seconded. A closed queue should behave like a closed file descriptor – you can read what's been sent before closing but you can't send new things.

njsmith:
There's an ambiguity that's baked into Queue combining both the send and receive ends onto the same object. If you think of `close` as closing the put side, then it should be possible to get already-queued data. If you think of it as closing the get side, then it should discard existing data and wake up any putters with something like a `BrokenQueueError`.

One option would be to have separate `close_get` and `close_put` methods (though then you have to decide which, if any, is invoked by `with queue:`). Another would be to split the queue into two objects, as discussed in #497.

I think that probably in the long run we do want to refactor queue into a two-object design, but if you need this functionality urgently then maybe it'd be easiest to call it `close_get` (assuming that's the version you need)?

Fuyukai:
I made it so that the get side automatically closes if there's no data and the put side is closed, because otherwise it would let the queue block forever.

Fuyukai:
(The failing Jenkins is because Python segfaulted.)

sorcio:
@Fuyukai there are a few cases that you're not testing and I suspect there might be a bug there. I cannot thing straight enough right now but let me just drop some notes.

1) Why not two explicitly named flags? I found it hard to track what is 0 and what is 1. You can't use `any()` but a simple `or` will do its job (and is maybe faster?)

2) You don't check for closed state in the `close_*` methods. What happens if I close twice? What happens if I call one and then the other?

3) You don't clean up the `_put_wait` and `_get_wait` collections after you send an error to waiters. I suspect there might be a situation where another operation will try to get a waiter from the list and reschedule it and break everything because we already rescheduled it with the QueueClosed exception. Maybe you can add more tests where there are multiple concurrent getters/putters.

4) small typo in the comments (size/side)

njsmith:
I'm going to sleep on this one... in the mean time, probably anyone interested in this will also be interested in #586 :-)

njsmith:
@Fuyukai oh right, sorry that took longer than expected... thanks for your patience here.

njsmith:
(Oh right, and the context for the record: this is being closed in favor of #586)

Issue #4, opened 2017-01-22
Subprocess support


njsmith:
Edit: if you're just coming to this issue, then this comment has a good overview of what there is to do: https://github.com/python-trio/trio/issues/4#issuecomment-398967572

----

Original description

Lots of annoying and fiddly details, but important!

I think for waiting, the best plan is to just give up on SIGCHLD (seriously, SIGCHLD is the worst) and park a thread in `waitpid` for each child process. Threads are lighter weight than processes so one thread-per-process shouldn't be a big deal. At least on Linux - if we're feeling ambitious we can do better on kqueue platforms. On Windows, it depends on what the state of our WaitFor{Multiple,Single}Object-fu is.

njsmith:
Looked at this some more today. We want to provide our own version of the `subprocess` API, in particular `run` and `Popen`. (We can drop the legacy APIs like `check_call` in favor of `run`.) `run` needs to be copied and rewritten, but it's only a few lines. `Popen` is the interesting bit. I'm thinking we should use the real `subprocess.Popen` to do the actual spawning and setting up of pipes – it's just way too complicated to try and replicate. Then once it's spawned, we can steal the file descriptors and do our own `wait`ing.

Communication: we should expose pipes as `Stream` objects (which means that this issue depends on the `Stream` part of #73). We'll need to write our own `communicate` but this should be extremely easy with trio :-).

Waiting: this we definitely want to implement ourselves. I think we might need to hack at `subprocess.Popen` to stop it from doing this itself – normally it tries very hard to reap the child, but we're going to have to take that over. (Well... I guess the alternative is: (a) Linux spawns a thread in `popen.wait()`, (b) other systems block until the process is done and then call `popen.wait()`?) And then we need our actual implementation, which relies on `WaitForSingleObject` on Windows, kqueue cleverness on MacOS, and `waitpid` on Linux.

What else: there are other attributes, in particular `terminate`-and-friends, but I think everything else can be pretty much passed through directly.

njsmith:
Might be worth checking out some alternative subprocess APIs like [delegator.py](https://github.com/kennethreitz/delegator.py) and [pexpect](https://github.com/pexpect/pexpect) (which relies on [ptyprocess](https://github.com/pexpect/ptyprocess/) for a lot of the heavy lifting).

[...]

Issue #542, opened 2018-05-30
Use less magic in constructing our public API exports, to make trio more intelligible to static analysis


ksamuel:
Despite the well defined `__all__`, several important tools (e.g: pylint, mypy, jedi) cannot understand the code structure of trio. They then report error in the user code. 

E.G:

On the following snippet, pylint will report trio not having the `run()` member, jedi won't allow code completion, vscode intellisense won't let you go to definition and mypy won't report any error if I try to add `+1` after `run()`
```python
import trio
...
trio.run(parent)
```

I understand this is inconvenient, as those tools should be able to pick up the `__all__` declaration and use that instead of requiring your to type it all again, but they don't. Since they are the most popular in the Python ecosystem, and are used by a lot of editors (vi, sublime text, vscode, etc) but also in tasks (tox, travis, git hooks, etc), it's important that they work.

For now, that means putting `# noqa` a bit everywhere or disabling a lot of warning, which is not practical.

parity3:
FYI, there has been [some work](#316 ) on this in the past. Maybe there could be a program to read import modules, introspect and spit out a new set of sources that are more static?

`python make-nice-for.py --tool=pylint --output-path=new-trio-sources/`

njsmith:
Yeah, we should probably give up on most of this cleverness and switch to doing the more boring approach.

There was some discussion of this in the chat today also, starting about here: https://gitter.im/python-trio/general?at=5b1842e099fa7f4c064e93a0

This is probably best handled as an incremental set of cleanups, since there are a lot of different subtle bits of trickery here. "Does this cleanup help pylint/mypy/pycharm/jedi understand what's going on" is probably a good metric to use to decide where to start...

Also relevant: #475, which I think is a change we probably should do, and will reduce the number of auto-generated wrapper functions in `trio/_core/_run.py`.

njsmith:
This comment describes how to write a strong test for pylint being able to understand our public API: https://github.com/python-trio/trio/pull/594#issuecomment-415336167

Zac-HD:
Closed by #594 :tada:

njsmith:
I like your enthusiasm, but I'm afraid we're not done yet :-). #594 made the main trio namespace work with tab completion, but we still need to figure out what to do with the submodules, and think about whether we can get rid of any of the *other* runtime namespace manipulation that we do...

Zac-HD:
[...]

Issue #233, opened 2017-06-22
Provide an API that does WaitForSingleObject on Windows


njsmith:
As noted in #52, there are some events on Windows that can only be detected by calling one of the `WaitFor` family of functions (`WaitForSingleObject`, `WaitForSingleObjectEx`, `WaitForMultipleObjects`, etc. etc.). Some examples include child process exit (required for #4), and detecting console input (required for #174).

To resolve this issue, we should provide a `trio.hazmat.WaitForSingleObject` function that takes a `HANDLE`, and waits for it, and is cancellable.

The most straightforward way to do this is to allocate a thread to sit in `WaitForSingleObjectEx`, which can do an "alertable wait", which means that it's possible to cause the function to wake up early by calling `QueueUserAPC`. So we would use `QueueUserAPC` from inside our abort callback to implement cancellation. There are some more notes in `_io_windows.py`.

sorcio:
Looks like Windows already implements an asynchronous signaling mechanism around `WaitForSingleObject` with [`RegisterWaitForSingleObject`](https://docs.microsoft.com/en-us/windows/desktop/api/WinBase/nf-winbase-registerwaitforsingleobject). If I understand the documentation correctly, it will implicitly use threads from a thread pool. The cool thing is that you get a wait handle back, which is cancellable with a call to [`UnregisterWait`](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-unregisterwait). If we can pass a callback that can call into trio (through a portal?) then this could be the "fancy solution".

njsmith:
Yeah, there are lots of options for how to tackle this. In fact Windows actually has [two different thread pool APIs](https://docs.microsoft.com/en-gb/windows/desktop/ProcThread/thread-pool-api) that both have special support for delegating `WaitForSingleObject` to a thread – `RegisterWaitForSingleObject` is the older API, and [`SetThreadpoolWait`](https://docs.microsoft.com/en-gb/windows/desktop/api/threadpoolapiset/nf-threadpoolapiset-setthreadpoolwait) is the newer API. Also I notice that at the top of this issue I suggest cancelling `WaitForSingleObjectEx` using APC, while in https://github.com/python-trio/trio/issues/4#issuecomment-398967572 I suggested doing cancellation by passing a dedicated `Event` to `WaitForMultipleObjects`.

Some considerations:

[...]

PR #575, opened 2018-07-29
Implement cancelable WaitForSingleObject for Windows

almarklein:
This is a stab at step 1 in https://github.com/python-trio/trio/issues/4#issuecomment-398967572.

Right now, it is implemented by calling `kernel32.WaitForSingleObject ()` with a zero timeout. This avoids the need for a second event object + thread etc. as mentioned in the referenced comment. But perhaps there was a good reason for that suggested approach -- maybe avoiding the arbitrary `sleep` reolution? If there is, I'm happy to dive a bit deeper and implement that. If not, I can remove some of the kernel32 stuff.

The tests make use of `trio.sleep()`. The docs mention something about an `autojump_clock` fixture, but I could not find what that does. Could that be of use here to make the sleeping in this test not affect the test duration?

The test also do some timing measurements and assert based on the result. There is a pretty good margin, but it feels a bit flaky (e.g. what if the CI it runs on is busy with other things?) Any ideas to improve that would be good.

PS: ~~running the auto style formatter made a few changes is unrelated files.~~ the latest yapf formats differently in a handful of places than the one used by CI.

### DIFF
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -12,7 +12,6 @@ import signal
 import attr
 
 from .. import _core
-from .. import _timeouts
 from . import _public
 from ._wakeup_socketpair import WakeupSocketpair
 from .._util import is_main_thread
@@ -108,60 +107,6 @@ def _handle(obj):
         return ffi.cast("HANDLE", obj)
     else:
         return obj
-
-
-async def WaitForSingleObject(handle):
-    """Async and cancellable variant of kernel32.WaitForSingleObject().
-
-    Args:
-      handle: A win32 handle.
-
-    """
-    # Quick check; we might not even need to spawn a thread. The zero
-    # means a zero timeout; this call never blocks. We also exit here
-    # if the handle is already closed for some reason.
-    retcode = kernel32.WaitForSingleObject(handle, 0)
-    if retcode != ErrorCodes.WAIT_TIMEOUT:
-        return
-
-    # :'(  avoid circular imports
-    from .._threads import run_sync_in_worker_thread
-
-    class StubLimiter:
-        def release_on_behalf_of(self, x):
-            pass
-
-        async def acquire_on_behalf_of(self, x):
-            pass
-
-    # Wait for a thread that waits for two handles: the handle plus a handle
-    # that we can use to cancel the thread.
-    cancel_handle = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
-    try:
-        await run_sync_in_worker_thread(
-            WaitForMultipleObjects_sync,
-            handle,
-            cancel_handle,
-            cancellable=True,
-            limiter=StubLimiter(),
-        )
-    finally:
-        # Clean up our cancel handle. In case we get here because this task was
-        # cancelled, we also want to set the cancel_handle to stop the thread.
-        kernel32.SetEvent(cancel_handle)
-        kernel32.CloseHandle(cancel_handle)
-
-
-def WaitForMultipleObjects_sync(*handles):
-    """Wait for any of the given Windows handles to be signaled.
-    
-    """
-    n = len(handles)
-    handle_arr = ffi.new("HANDLE[{}]".format(n))
-    for i in range(n):
-        handle_arr[i] = handles[i]
-    timeout = 0xffffffff  # INFINITE
-    kernel32.WaitForMultipleObjects(n, handle_arr, False, timeout)  # blocking
 
 
 @attr.s(frozen=True)
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -22,6 +22,7 @@ from ._windows_cffi import (
     INVALID_HANDLE_VALUE,
     raise_winerror,
     ErrorCodes,
+    _handle,
 )
 
 # There's a lot to be said about the overall design of a Windows event
@@ -94,19 +95,6 @@ def _check(success):
     if not success:
         raise_winerror()
     return success
-
-
-def _handle(obj):
-    # For now, represent handles as either cffi HANDLEs or as ints.  If you
-    # try to pass in a file descriptor instead, it's not going to work
-    # out. (For that msvcrt.get_osfhandle does the trick, but I don't know if
-    # we'll actually need that for anything...) For sockets this doesn't
-    # matter, Python never allocates an fd. So let's wait until we actually
-    # encounter the problem before worrying about it.
-    if type(obj) is int:
-        return ffi.cast("HANDLE", obj)
-    else:
-        return obj
 
 
 @attr.s(frozen=True)
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -12,6 +12,7 @@ import signal
 import attr
 
 from .. import _core
+from .. import _timeouts
 from . import _public
 from ._wakeup_socketpair import WakeupSocketpair
 from .._util import is_main_thread
@@ -107,6 +108,24 @@ def _handle(obj):
         return ffi.cast("HANDLE", obj)
     else:
         return obj
+
+
+async def WaitForSingleObject(handle):
+    """Async and cancelable variant of kernel32.WaitForSingleObject().
+
+    Args:
+      handle: A win32 handle.
+
+    """
+    # Wait here while the handle is not signaled. The 0 in WaitForSingleObject()
+    # means zero timeout; the function simply tells us the status of the handle.
+    # Possible values: WAIT_TIMEOUT, WAIT_OBJECT_0, WAIT_ABANDONED, WAIT_FAILED
+    # We exit in all cases except when the hadle is still unset.
+    while True:
+        await _timeouts.sleep_until(_core.current_time() + 0.01)
+        retcode = kernel32.WaitForSingleObject(handle, 0)
+        if retcode != ErrorCodes.WAIT_TIMEOUT:
+            break
 
 
 @attr.s(frozen=True)
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -160,13 +160,8 @@ def WaitForMultipleObjects_sync(*handles):
     handle_arr = ffi.new("HANDLE[{}]".format(n))
     for i in range(n):
         handle_arr[i] = handles[i]
-    timeout = 1000 * 60 * 60 * 24  # todo: use INF here, whatever that is, and ditch the while
-    while True:
-        retcode = kernel32.WaitForMultipleObjects(
-            n, handle_arr, False, timeout
-        )
-        if retcode != ErrorCodes.WAIT_TIMEOUT:
-            break
+    timeout = 0xffffffff  # INFINITE
+    kernel32.WaitForMultipleObjects(n, handle_arr, False, timeout)  # blocking
 
 
 @attr.s(frozen=True)
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -111,19 +111,60 @@ def _handle(obj):
 
 
 async def WaitForSingleObject(handle):
-    """Async and cancelable variant of kernel32.WaitForSingleObject().
+    """Async and cancellable variant of kernel32.WaitForSingleObject().
 
     Args:
       handle: A win32 handle.
 
     """
-    # Wait here while the handle is not signaled. The 0 in WaitForSingleObject()
-    # means zero timeout; the function simply tells us the status of the handle.
-    # Possible values: WAIT_TIMEOUT, WAIT_OBJECT_0, WAIT_ABANDONED, WAIT_FAILED
-    # We exit in all cases except when the hadle is still unset.
+    # Quick check; we might not even need to spawn a thread. The zero
+    # means a zero timeout; this call never blocks. We also exit here
+    # if the handle is already closed for some reason.
+    retcode = kernel32.WaitForSingleObject(handle, 0)
+    if retcode != ErrorCodes.WAIT_TIMEOUT:
+        return
+
+    # :'(  avoid circular imports
+    from .._threads import run_sync_in_worker_thread
+
+    class StubLimiter:
+        def release_on_behalf_of(self, x):
+            pass
+
+        async def acquire_on_behalf_of(self, x):
+            pass
+
+    # Wait for a thread that waits for two handles: the handle plus a handle
+    # that we can use to cancel the thread.
+    cancel_handle = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    try:
+        await run_sync_in_worker_thread(
+            WaitForMultipleObjects_sync,
+            handle,
+            cancel_handle,
+            cancellable=True,
+            limiter=StubLimiter(),
+        )
+    finally:
+        # Clean up our cancel handle. In case we get here because this task was
+        # cancelled, we also want to set the cancel_handle to stop the thread.
+        kernel32.SetEvent(cancel_handle)
+        kernel32.CloseHandle(cancel_handle)
+
+
+def WaitForMultipleObjects_sync(*handles):
+    """Wait for any of the given Windows handles to be signaled.
+    
+    """
+    n = len(handles)
+    handle_arr = ffi.new("HANDLE[{}]".format(n))
+    for i in range(n):
+        handle_arr[i] = handles[i]
+    timeout = 1000 * 60 * 60 * 24  # todo: use INF here, whatever that is, and ditch the while
     while True:
-        await _timeouts.sleep_until(_core.current_time() + 0.01)
-        retcode = kernel32.WaitForSingleObject(handle, 0)
+        retcode = kernel32.WaitForMultipleObjects(
+            n, handle_arr, False, timeout
+        )
         if retcode != ErrorCodes.WAIT_TIMEOUT:
             break
 
--- a/trio/_core/tests/test_windows.py
+++ b/trio/_core/tests/test_windows.py
@@ -1,4 +1,5 @@
 import os
+from threading import Thread
 import pytest
 
 on_windows = (os.name == "nt")
@@ -9,7 +10,7 @@ from ... import _core
 from ... import _timeouts
 if on_windows:
     from .._windows_cffi import ffi, kernel32
-    from .._io_windows import WaitForSingleObject
+    from .._io_windows import WaitForSingleObject, WaitForMultipleObjects_sync
 
 
 async def test_completion_key_listen():
@@ -40,17 +41,52 @@ async def test_completion_key_listen():
             print("end loop")
 
 
+async def test_WaitForMultipleObjects_sync():
+    # One handle
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    t = Thread(target=WaitForMultipleObjects_sync, args=(handle1,))
+    t.start()
+    kernel32.SetEvent(handle1)
+    t.join()  # the test succeeds if we do not block here :)
+    kernel32.CloseHandle(handle1)
+
+    # Two handles, signal first
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    handle2 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    t = Thread(target=WaitForMultipleObjects_sync, args=(handle1, handle2))
+    t.start()
+    kernel32.SetEvent(handle1)
+    t.join()  # the test succeeds if we do not block here :)
+    kernel32.CloseHandle(handle1)
+    kernel32.CloseHandle(handle2)
+
+    # Two handles, signal seconds
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    handle2 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    t = Thread(target=WaitForMultipleObjects_sync, args=(handle1, handle2))
+    t.start()
+    kernel32.SetEvent(handle2)
+    t.join()  # the test succeeds if we do not block here :)
+    kernel32.CloseHandle(handle1)
+    kernel32.CloseHandle(handle2)
+
+    # Closing the handle will not stop the thread. Initiating a wait on a
+    # closed handle will fail/return, but closing a handle that is already
+    # being waited on will not stop whatever is waiting for it.
+
+
 async def test_WaitForSingleObject():
 
     # Set the timeout used in the tests. The resolution of WaitForSingleObject
     # is 0.01 so anything more than a magnitude larger should probably do.
     # If too large, the test become slow and we might need to mark it as @slow.
-    TIMEOUT = 0.1
+    TIMEOUT = 0.5
 
-    # Test 1, handle is SET after 1 sec in separate coroutine
     async def handle_setter(handle):
         await _timeouts.sleep(TIMEOUT)
         kernel32.SetEvent(handle)
+
+    # Test 1, handle is SET after 1 sec in separate coroutine
 
     handle = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
     t0 = _core.current_time()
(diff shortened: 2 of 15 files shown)


almarklein:
BTW: about [that question on stackoverflow](https://stackoverflow.com/questions/50959733/does-closing-an-event-object-cause-threads-waiting-on-that-object-to-wake-up): I added tests to verify that it stops waiting when the handle is closed. I am not sure if closing *sets* the handle, it can also be that `kernel32.WaitForSingleObject()` will return WAIT_ABANDONED or WAIT_FAILED (the code only waits on WAIT_TIMEOUT).

njsmith:
The "polling loop with short sleep" trick can be a really useful hack, but unfortunately doesn't play nicely with modern computers. The problem is that each time we poll, we have to wake up the CPU, usually to do nothing. Imagine a process that has 10 tasks blocked in `WaitForSingleObject` – now it's waking up 1000 times/second, which means your CPU never goes into energy saving mode, which means you laptop's battery life gets cut in two or something. That's why I was saying we should use a thread – it lets our process actually go to sleep until something happens. (Though by all means do a zero timeout call first before starting a thread, just to make sure there's actually something to wait for.)

sorcio:
Linking to the more specific open issue: https://github.com/python-trio/trio/issues/233

almarklein:
Thanks for the explanation. In that case I will do the thread-thing.

trio/_core/_io_windows.py
```diff
@@ -109,6 +110,65 @@ def _handle(obj):
...
+
+def WaitForMultipleObjects_sync(*handles):
+    """Wait for any of the given Windows handles to be signaled.
+    
+    """
+    n = len(handles)
+    handle_arr = ffi.new("HANDLE[{}]".format(n))
+    for i in range(n):
+        handle_arr[i] = handles[i]
+    timeout = 1000 * 60 * 60 * 24  # todo: use INF here, whatever that is, and ditch the while
```
njsmith: Windows defines the special value `INFINITE` to be `0xFFFFFFFF`. We should probably add this as a constant in `_windows_cffi.py`, next to the error codes.

trio/_core/_io_windows.py
```diff
@@ -109,6 +110,65 @@ def _handle(obj):
...
+        return
+
+    # :'(  avoid circular imports
+    from .._threads import run_sync_in_worker_thread
+
+    class StubLimiter:
+        def release_on_behalf_of(self, x):
+            pass
+
+        async def acquire_on_behalf_of(self, x):
```
njsmith: Since nothing here depends on access to the internals of the `WindowsIOManager`, I think we can move it out into some place like `trio/_wait_for_single_object.py`.

trio/_core/_io_windows.py
```diff
@@ -109,6 +110,65 @@ def _handle(obj):
...
+    handle_arr = ffi.new("HANDLE[{}]".format(n))
+    for i in range(n):
+        handle_arr[i] = handles[i]
+    timeout = 1000 * 60 * 60 * 24  # todo: use INF here, whatever that is, and ditch the while
+    while True:
+        retcode = kernel32.WaitForMultipleObjects(
+            n, handle_arr, False, timeout
+        )
+        if retcode != ErrorCodes.WAIT_TIMEOUT:
+            break
```
njsmith: The docs say that `WaitForMultipleObjects` can return `WAIT_FAILED`, and that in this case we should check `GetLastError`. I can't think of why this would ever fail, but probably best to check anyway to be safe. Fortunately `_windows_cffi.py` already has the machinery you need; just do something like:

```python
if retcode == WAIT_FAILED:
    _windows_cffi.raise_winerror()
```

trio/_core/tests/test_windows.py
```diff
@@ -1,13 +1,16 @@
 import os
+from threading import Thread
```
njsmith: I've [learned the hard way](https://github.com/python-trio/trio/pull/214) that it's usually a bad idea to use `threading.Thread` directly in tests – it can do things like swallow exceptions :-). If you need threads, then Trio's primitives are actually easier to use and more reliable, so ... that's what I use now.

trio/_core/tests/test_windows.py
```diff
@@ -38,5 +41,109 @@
...
 
 
+async def test_WaitForMultipleObjects_sync():
+    # One handle
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    t = Thread(target=WaitForMultipleObjects_sync, args=(handle1,))
+    t.start()
+    kernel32.SetEvent(handle1)
+    t.join()  # the test succeeds if we do not block here :)
+    kernel32.CloseHandle(handle1)
```
njsmith: I don't think this test needs a thread at all – there's no particular reason to think that the thread is actually running before we call `SetEvent`, so you can capture the important part of the test by doing:

```python
handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
kernel32.SetEvent(handle1)
WaitForMultipleObjects_sync(handle1)
kernel32.CloseHandle(handle1)
```

trio/_core/tests/test_windows.py
```diff
@@ -38,5 +41,109 @@
...
+
+    # Two handles, signal first
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    handle2 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    t = Thread(target=WaitForMultipleObjects_sync, args=(handle1, handle2))
+    t.start()
+    kernel32.SetEvent(handle1)
+    t.join()  # the test succeeds if we do not block here :)
+    kernel32.CloseHandle(handle1)
+    kernel32.CloseHandle(handle2)
```
njsmith: Same comment applies.

trio/_core/tests/test_windows.py
```diff
@@ -38,5 +41,109 @@
...
+
+    # Two handles, signal seconds
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    handle2 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    t = Thread(target=WaitForMultipleObjects_sync, args=(handle1, handle2))
+    t.start()
+    kernel32.SetEvent(handle2)
+    t.join()  # the test succeeds if we do not block here :)
+    kernel32.CloseHandle(handle1)
+    kernel32.CloseHandle(handle2)
```
njsmith: Same comment applies.

trio/_core/tests/test_windows.py
```diff
@@ -38,5 +41,109 @@
...
+    kernel32.CloseHandle(handle)
+    t1 = _core.current_time()
+    assert TIMEOUT <= (t1 - t0) < 1.1 * TIMEOUT
+    print('test_WaitForSingleObject test 1 OK')
+
+    # Test 2, handle is CLOSED after 1 sec - NOPE, wont work unless we use zero timeout
+
+    pass
+
+    # Test 3, cancelation
```
njsmith: spelling: should be `cancellation`

(In American spelling, there's some inconsistency about whether you use one or two l's in words like "cancelled". But in trio we always use two, because that's acceptable everywhere, and it's easier to remember than trying to keep track of which words have two and which have one.)

trio/_core/tests/test_windows.py
```diff
@@ -38,5 +41,109 @@
...
+    handle = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    kernel32.CloseHandle(handle)
+    t0 = _core.current_time()
+
+    with _timeouts.move_on_after(TIMEOUT):
+        await WaitForSingleObject(handle)
+
+    t1 = _core.current_time()
+    assert (t1 - t0) < 0.5 * TIMEOUT
+    print('test_WaitForSingleObject test 5 OK')
```
njsmith: Does this one actually require a timeout? I would have thought that `WaitForSingleObject` would fail immediately, before even starting a thread...

trio/_core/tests/test_windows.py
```diff
@@ -38,5 +41,109 @@
...
+    # closed handle will fail/return, but closing a handle that is already
+    # being waited on will not stop whatever is waiting for it.
+
+
+async def test_WaitForSingleObject():
+
+    # Set the timeout used in the tests. The resolution of WaitForSingleObject
+    # is 0.01 so anything more than a magnitude larger should probably do.
+    # If too large, the test become slow and we might need to mark it as @slow.
+    TIMEOUT = 0.5
```
njsmith: The comment about the resolution is out of date, right?

If this is a test that has to sleep – and I think it probably is – then we're probably going to mark it `@slow` :-). I try to keep the quick test run you do over and over while hacking as fast as possible, so right now the slowest single test is <0.3s. We always run the full test suite during CI runs though, so all the `@slow` tests still get run on every commit.

Maybe we could have one test that's not marked `@slow` and checks the trivial cases of (1) `WaitForSingleObject` on an already-ready object, (2) `WaitForSingleObject` that gets preemptively cancelled, and then a second test that is marked `@slow` and has the tests for (3) we wait a bit and then signal the object, (4) we wait a bit and then cancel the call?

trio/_core/tests/test_windows.py
```diff
@@ -38,5 +41,109 @@
...
+    kernel32.SetEvent(handle)
+    t0 = _core.current_time()
+
+    with _timeouts.move_on_after(TIMEOUT):
+        await WaitForSingleObject(handle)
+
+    kernel32.CloseHandle(handle)
+    t1 = _core.current_time()
+    assert (t1 - t0) < 0.5 * TIMEOUT
+    print('test_WaitForSingleObject test 4 OK')
```
njsmith: This is already signaled, right, not already cancelled?

trio/_core/_io_windows.py
almarklein: Done. This also gets rid of that circular import.

trio/_core/tests/test_windows.py
almarklein: 👍

trio/_core/tests/test_windows.py
almarklein: This tests a synchronous function, so afaik I cannot use a Trio primitive here, right? (I am not that familiar with Trio yet though :)). 

I now changed the test, so that the main thread waits for the handle, and a thread is spawned only to set the handle after a short timeout (there is thus very little that can go wrong in that thread).

trio/_core/tests/test_windows.py
almarklein: > there's no particular reason to think that the thread is actually running before we call `SetEvent`

Oh dear, I overlooked that :)

Signaling before waiting is one option (and I added it for the fast tests), but I think it would be good to also test actually "acquiring the wait" and then setting the handle to "release the wait". Put that in a `@slow` test.

trio/_core/tests/test_windows.py
almarklein: Not really, was mostly there as a fail-safe. Removed it because it makes the test much shorter.

trio/_core/tests/test_windows.py
almarklein: yes

trio/_core/tests/test_windows.py
almarklein: Yes, that comment was outdated.

almarklein:
@njsmith I addressed your comments, and the tests are now split into slow and fast tests.

trio/_core/tests/test_windows.py
njsmith: You can use an async test function to test a synchronous function :-).

```python
async def test_...():
        # This runs 'sync_fn' in a background thread
        nursery.start_soon(trio.run_sync_in_worker_thread, sync_fn)
        # ... we can do other stuff here while it's running ...
    # When we de-dent to close the nursery, it automatically joins the background thread
```

So this automatically makes sure that you join the background thread (it can't accidentally keep running while other tests go, even if this test crashes), it automatically detects exceptions in the background thread and makes sure that if they happen it causes the test to fail, and it's actually simpler to use (IMO) than the raw `threading.Thread` API :-).

trio/_core/tests/test_windows.py
njsmith: Makes sense :-)

trio/_wait_for_single_object.py
```diff
@@ -0,0 +1,56 @@
...
+    """Async and cancellable variant of kernel32.WaitForSingleObject().
+
+    Args:
+      handle: A win32 handle.
+
+    """
+    # Quick check; we might not even need to spawn a thread. The zero
+    # means a zero timeout; this call never blocks. We also exit here
+    # if the handle is already closed for some reason.
+    retcode = kernel32.WaitForSingleObject(handle, 0)
```
njsmith: This should have an `if retcode == ErrorCodes.WAIT_FAILED: raise_winerror()` branch too, I think.

trio/_wait_for_single_object.py
```diff
@@ -0,0 +1,56 @@
...
+def WaitForMultipleObjects_sync(*handles):
+    """Wait for any of the given Windows handles to be signaled.
+    
+    """
+    n = len(handles)
+    handle_arr = ffi.new("HANDLE[{}]".format(n))
+    for i in range(n):
+        handle_arr[i] = handles[i]
+    timeout = 0xffffffff  # INFINITE
+    kernel32.WaitForMultipleObjects(n, handle_arr, False, timeout)  # blocking
```
njsmith: And this should check for `WAIT_FAILED` too.

trio/tests/test_wait_for_single_object.py
```diff
@@ -0,0 +1,168 @@
...
+    kernel32.CloseHandle(handle1)
+    WaitForMultipleObjects_sync(handle1, handle2)
+    kernel32.CloseHandle(handle2)
+
+    # Two handles, close second
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    handle2 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    kernel32.CloseHandle(handle2)
+    WaitForMultipleObjects_sync(handle1, handle2)
+    kernel32.CloseHandle(handle1)
```
njsmith: I expect that if we do check for `WAIT_FAILED`, then these will start raising errors, which is good.

trio/tests/test_wait_for_single_object.py
```diff
@@ -0,0 +1,168 @@
...
+        kernel32.SetEvent(handle)
+
+    # One handle
+    handle1 = kernel32.CreateEventA(ffi.NULL, True, False, ffi.NULL)
+    t0 = _core.current_time()
+    Thread(target=signal_soon_sync, args=(handle1,)).start()
+    WaitForMultipleObjects_sync(handle1)
+    t1 = _core.current_time()
+    assert TIMEOUT <= (t1 - t0) < 1.2 * TIMEOUT
+    kernel32.CloseHandle(handle1)
```
njsmith: So... this is weird. If you look at the Appveyor results, the `sync_slow` test failed on a bunch of trials, though the other tests passed, including the other `_slow` test below. And the reason they failed is that `t1 - t0` ended up being like 0.297, which is less than 0.3. Which seems like it should be impossible.

I think that's going on is that on Windows, `time.sleep` actually uses a slightly different clock than trio does. The Windows clock APIs are really complicated, but trio currently uses `time.monotonic`, which is not quite the same as other clocks. If you use `trio.sleep`, that's always fully in sync with `trio.current_time`, so the assertion should pass – and that's what we see in the other slow test below. But if you use `time.sleep`, then apparently there can be a bit of skew.

So, I think you should rewrite this test to use `await trio.sleep`, like the other one below. Fortunately this fits in nicely with my other comment about not using the raw threading API :-).

Also, I'd switch all the timeouts checks to use `< 2 * TIMEOUT`, based on what we've ended up doing in `test_run.py` and `test_ki.py`. (We used to use smaller values and then would get false failures when some CI server was running slow, and eventually ended up bumping the values up to that.)

trio/_wait_for_single_object.py
```diff
@@ -0,0 +1,56 @@
...
+
+class StubLimiter:
+    def release_on_behalf_of(self, x):
+        pass
+
+    async def acquire_on_behalf_of(self, x):
+        pass
+
+
+async def WaitForSingleObject(handle):
```
njsmith: Okay, now that this is going to be a public API, we should think about what kind of object people will pass in as this `handle` object. Internally we use our ffi object's HANDLE objects, but the ffi object isn't part of our public API, so we can't expect users to do that. We should also let users pass in plain old integers, and then cast them to our HANDLE objects internally to use them.

The `_io_windows.py` has a `_handle` helper function that does this. We should move that helper into `_windows_cffi.py` (since it's part of our general "accessing the Windows API" code, not part of `WindowsIOManager` specifically), and then in this function add something like `handle = _handle(handle)`.

trio/_core/tests/test_windows.py
almarklein: Of course. I don't understand why I did not think to use `run_sync_in_worker_thread` 🤔

trio/_wait_for_single_object.py
almarklein: Good point. I initially though that just returning on closed handled would be practical, but if someone somehow passes a wrong/invalid handle, they'd want to know.

almarklein:
@njsmith 

> We need to export our new function, so people can use it! This is low-level, so it should be part of the trio.hazmat namespace. Unfortunately this is a little more complicated than it should be, because previous to this trio.hazmat only re-exported stuff from trio._core, so the code will need a bit of tweaking. Feel free to ask here or in gitter if you're not sure what to do.

Just wondering if this might be a hint that it should be in `trio._core` after all?

trio/tests/test_wait_for_single_object.py
almarklein: (I think I fixed it (yet unpushed) by checking `time.monotonic`. Am happy to change to make it more trio-ish, though not sure how that would work)

trio/_wait_for_single_object.py
njsmith: Yeah... I was also confused about this until I thought about it more, and remembered that there isn't really such a thing as a "closed handle" – once you close a handle, the next new handle you open has a chance of being assigned the same handle id. So doing anything with a handle after you close it is really error prone – you might end up doing it on some other random object without realizing it! Better to get an error early in testing and fix the bug.

trio/tests/test_wait_for_single_object.py
njsmith: Right, you can't use `trio.sleep()` inside a sync function. But you can move `trio.sleep()` into the main thread, and put the `WaitForMultipleObjects_sync` into a thread :-).

trio/_core/tests/test_windows.py
njsmith: Hey, don't feel bad, this concurrency/IO stuff is complicated. I didn't think of it either when I was writing the `trio.ssl` testsuite, even though I *wrote* `run_sync_in_worker_thread` like the month before... so you get to learn from my fail :-)

njsmith:
> Just wondering if this might be a hint that it should be in `trio._core` after all?

That's a great question! I thought about that too. But it turns out these are slightly different distinctions. 

The point of having `trio/*.py` versus `trio/_core/*.py` is to make sure that third-party libraries can make major extensions to trio without needing to add code to trio itself. Basically keeping ourselves honest: whenever possible, we put stuff in `trio/*.py`, and stuff in `trio/*.py` is only allowed to use APIs that anyone else could use – it doesn't have any special privileged access. So built-in features like `trio.socket` and `trio.run_sync_in_worker_thread` and `trio.Lock` are all pretty solid and useful, but if you don't like them then you can implement your own versions and put them up on PyPI without asking anyone for permission. `trio/_core/*.py` OTOH is the stuff that *has* to be built-in and deal with private internal data structures, and everything it exports to the rest of trio (the `trio._core` namespace) is also made available to users.

The point of `trio` vs. `trio.hazmat` is to give users a hint about which stuff they really should use every day, versus stuff that you only need if you're extending trio in some lowish-level way. So unsurprisingly, `trio._core` has a lot of fairly low-level stuff in it that ends up in `trio.hazmat` :-). (Though it also has lots of stuff that's in the main `trio` namespace, like `trio.open_nursery`.) `WaitForSingleObject` is unusual in that it's the first thing to fall into that venn diagram intersection of "doesn't actually need access to deep internals" and "actually super low-level functionality that only experts will use directly", but there it is.

trio/tests/test_wait_for_single_object.py
almarklein: Right :)

almarklein:
Ok, I've made the requested changes. A few points:

> WaitForSingleObject is unusual in that it [...] "doesn't actually need access to deep internals"

Just to be sure, these don't count as deep internals? `from ._core._windows_cffi import ffi, kernel32, ErrorCodes, raise_winerror, _handle`

I added module docstings to `hazmat` and `_core` to explain their purpose (as I understand it). I find module docstrings quite useful (e.g. hitting `somemodule?` in an IDE or notebook and getting docs). However, I have note seen module docstrings elsewhere, is there a reason that Trio avoids them?

I renamed the new module to `_wait_for_object.py`, because I would not be surprised if at some point someone needs a `WaitForMultipleObjects()`.

This new feature is Windows only. I've added this to the function docstring. What to do wrt exposing? I now wrote it such that the function is only in `hazmat` on Windows, but maybe its better to always have it present and just fail on non-windows?

Related, in `hazmat.py` it says "# Some hazmat symbols are platform specific". I checked, and these names are initially in `__all__`, but are dropped (on Windows): wait_writable, wait_readable, notify_fd_close, current_kqueue, monitor_kevent, wait_kevent. These do not sound platform specific, in fact, for some I cannot find any definition. Are these artifacts?

I understand the reason for the trio/_core/hazmat namespaces. But the loops that the code has to jump through is not the prettiest part of Trio :)  This is outside of the scope of this PR, but has it been considered to do this differently? For instance with a script that generates `__all__` lists, or just using manual lists backed up by tests?

njsmith:
> Just to be sure, these don't count as deep internals? from `._core._windows_cffi import ffi, kernel32, ErrorCodes, raise_winerror, _handle`

Nope! The raw Win32 bindings aren't connected to Trio's internals at all. In a perfect world they'd be part of the stdlib (similar to how the `os` module has bindings for lots of low-level POSIX APIs), or they could be trivially moved into a third-party library. For now there's no advantage to doing that, and they need to live somewhere, so they're in `trio/_core` next to `_io_windows.py`, which was the only user. But using them in `WaitForSingleObject` doesn't create any coupling between your code and Trio's internals.

> However, I have note seen module docstrings elsewhere, is there a reason that Trio avoids them?

Mostly I just don't use them myself and forget to write them :-). I guess another reason is that my manuals tend to have a lot more narrative text around each module than I would want to put in a module docstring, and keeping module docstrings up to date is a hassle. But I don't have any objection to adding docstrings to public module that are like, a few sentences long and give a brief overview of what the module's purpose is.

> This new feature is Windows only. I've added this to the function docstring. What to do wrt exposing? I now wrote it such that the function is only in hazmat on Windows, but maybe its better to always have it present and just fail on non-windows?

There are arguments either way, but for now Trio follows the stdlib's lead here and makes it so that system-specific functions only exist where they're supported. It's kind of convenient because you can do `if hasattr(trio.hazmat, "WaitForSingleObject"): ...` and select the appropriate system-specific code path without having to guess what's supported or try actually invoking the function.

> I checked, and these names are initially in `__all__`, but are dropped (on Windows): `wait_writable`, `wait_readable`, `notify_fd_close`, `current_kqueue`, `monitor_kevent`, `wait_kevent`. These do not sound platform specific,

They are all platform specific. The `wait_writable` and `wait_readable` should really be called `wait_fd_writable` and `wait_fd_readable`, and Windows doesn't have functions that work on arbitrary fd's the same way Unixes do. And the kqueue and kevent functions are related to the BSD/MacOS async I/O primitive [`kqueue`](https://www.freebsd.org/cgi/man.cgi?kqueue).

> in fact, for some I cannot find any definition. Are these artifacts?

It's, umm, like you said, not the prettiest part of Trio. They're defined in `trio/_core/_io_*.py`, and exported via a convoluted chain of code that involves the `@_public` decorator, some code generation at import time in the bottom of `trio/_core/_run.py`, some code in `trio/_core/__init__.py`, etc.

> has it been considered to do this differently?

Yeah, this dates back to the very beginning of trio, and has turned out to be pretty overengineered. We really should simplify it, which is #542.

almarklein:
Thanks for your explanations and patience, Nathaniel! This PR looks good from my end.

almarklein:
Done!

njsmith:
Noticed two more little doc nits, and just pushed changes myself – do they look good to you?

(The cffi thing is because our `kernel32` and cffi `HANDLE` objects are things that are defined in `_windows_cffi.py`; our users can make their own if they want, but I don't think they can pass them directly into `WaitForSingleObject`.)

almarklein:
> do they look good to you?

Yes! (was not sure if you'd seen the thumbsup :))

njsmith:
Oh thanks, I probably would have noticed the thumbsup eventually but it would have taken a bit :-). Okay, let's do this!

njsmith:
And as per our [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team), I also sent you an invitation to join the project :-). Up to you whether you accept or not, but I hope you do!

Are you planning to keep working on this, e.g. by moving on to `ReadFile`/`WriteFile`? Let me/us know if you need any help...

almarklein:
> Are you planning to keep working on this, e.g. by moving on to ReadFile/WriteFile? 

Not now; I should really focus on other things for the moment (and holidays). I got excited about Trio by @touilleMan's talk at EuroPython and decided to joint the sprint to familiarize myself with it more. I should congratulate you; the codebase is really nice, and the people even more so! So it's likely that when I have some free cycles I'll contribute more :)

njsmith:
Makes sense! Have a great holiday, and come back any time :-).

PR #576, opened 2018-07-29
Fix CapacityLimiter memory leak.

### LINKED ISSUE #548: Check CapacityLimiter's handling of _pending_borrowers
I was just reading the code for an unrelated reason, and realized that this code looks very suspicious:

https://github.com/python-trio/trio/blob/65729b121c55ff0aa1f5d43bd9cdcd6a81e54f0c/trio/_sync.py#L286-L296

Specifically, if the call to `await self._lot.park()` gets cancelled, it looks like we don't clear `self._pending_borrowers[task]`, which could be a memory leak (since it pins task objects in memory).

Fuyukai:
Closes #548.

### DIFF
--- a/trio/tests/test_sync.py
+++ b/trio/tests/test_sync.py
@@ -154,7 +154,7 @@ async def test_CapacityLimiter_memleak_548():
 
     async with _core.open_nursery() as n:
         n.start_soon(limiter.acquire)
-        await _core.checkpoint()  # give it a chance to run the task
+        await wait_all_tasks_blocked()  # give it a chance to run the task
         n.cancel_scope.cancel()
 
     # if this is 1, the acquire call (despite being killed) is still there in the task, and will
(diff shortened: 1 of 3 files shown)


trio/tests/test_sync.py
```diff
@@ -147,6 +147,21 @@
...
 
 
+# regression test for issue #548
+async def test_CapacityLimiter_memleak_548():
+    limiter = CapacityLimiter(total_tokens=1)
+    await limiter.acquire()
+
+    async with _core.open_nursery() as n:
+        n.start_soon(limiter.acquire)
+        await _core.checkpoint()  # give it a chance to run the task
```
njsmith: A more robust way to do this is `wait_all_tasks_blocked`.

njsmith:
Closing/reopening to tickle travis, hopefully clear #584 failures.

Issue #251, opened 2017-07-27
On Linux, set up socket object correctly when creating from an fd


njsmith:
Python has APIs for wrapping a bare socket fd into a socket object: `socket.socket(fileno=...)` to wrap, and `socket.fromfd(fileno, ...)` to dup-and-wrap. We currently inherit these with no changes.

There's a fairly nasty bug in Python though: to work with a socket correctly, it's not enough to know the fd; Python also needs to know the address family (most important), the socket type, and the protocol (least important). ~~On Windows, it correctly pulls this information out of the passed-in socket.~~ [Edit: actually [Windows is broken too](https://bugs.python.org/issue27377#msg270669).] On other platforms, it doesn't (see [bpo-28134](https://bugs.python.org/issue28134), [bpo-27377](https://bugs.python.org/issue27377)).

AFAICT on MacOS this is mostly a lost cause. You can use `getsockname` to get the address family, but that only works on connected stream sockets, and there's no way to query for the other attributes. I guess you might be able to distinguish Unix/IPv4/IPv6 families by making `getsockopt` calls with the appropriate level and see if you get an error or not? Maybe `LOCAL_PEERCRED` is a reliable signature of AF_UNIX+SOCK_STREAM, and then there's a bunch of IPv6-specific sockopts... (see `unix(4)`, `ip(4)`, `ip6(4)`).

But on Linux – which is the point of this bug! – this is actually pretty easy. You can just ask for all the key pieces of information like:

```python
sock_family = getsockopt(fd, SOL_SOCKET, SO_DOMAIN)
sock_type = getsockopt(fd, SOL_SOCKET, SO_TYPE)
sock_proto = getsockopt(fd, SOL_SOCKET, SO_PROTOCOL)
```
(more detailed example: https://github.com/tiran/socketfromfd/blob/master/socketfromfd.py)

And this actually matters, because to support systemd socket activation, we need to be able to take bare fds and wrap them into trio socket objects, on Linux.

So: we should teach `trio.socket.socket` to do these checks on Linux when `fileno=` is given, and we should implement `trio.socket.fromfd` in terms of `trio.socket.socket`. (The standard library's implementation of `socket.fileno` in terms of `socket.socket` is ~2 lines of code, we can just copy them into trio.)

njsmith:
[...]

PR #577, opened 2018-07-29
Teach trio.socket.fromfd and friends to autodetect socket configuration on Linux

2easy:
Hi, I'd like to propose this solution to the issue #251 .

### DIFF
--- a/trio/_socket.py
+++ b/trio/_socket.py
@@ -1,3 +1,5 @@
+import ctypes as _ctypes
+from ctypes.util import find_library as _find_library
 import os as _os
 import socket as _stdlib_socket
 import sys as _sys
@@ -298,15 +300,6 @@ def from_stdlib_socket(sock):
     return _SocketType(sock)
 
 
-@_wraps(_stdlib_socket.fromfd, assigned=(), updated=())
-@_add_to_all
-def fromfd(*args, **kwargs):
-    """Like :func:`socket.fromfd`, but returns a trio socket object.
-
-    """
-    return from_stdlib_socket(_stdlib_socket.fromfd(*args, **kwargs))
-
-
 if hasattr(_stdlib_socket, "fromshare"):
 
     @_wraps(_stdlib_socket.fromshare, assigned=(), updated=())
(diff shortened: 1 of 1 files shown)


2easy:
There is problem with `SO_PROTOCOL` on mac, will fix that in a moment.

njsmith:
Hello!

I'm afraid I have misled you and made this much more complicated than it has to be :-(. The code you used for a reference jumps through all kinds of hoops using ctypes, because it's trying to call `getsockopt` on a socket that it only has a fileno for, not a socket object. But on all the versions of Python we care about, we can create a socket object (maybe with incorrect values), and then we can directly use Python's regular `getsockopt` method and skip all that ctypes stuff. Also, the versions of Python we care about have the relevant `SO_WHATEVER` magic constants on the `socket` module on all the systems where they work at all, so we don't need to have the magic numbers like "39" for `SO_DOMAIN`. (And besides, those magic numbers are only correct for Linux.)

So my suggestion would be to delete all the ctypes stuff, and instead have a function like this, that takes whatever the user gave us, fixes it up as best we can (which might be more or less depending on the OS), and returns the fixed up values.

```python
def _fix_attrs_for_fileno(family, type, proto, fileno):
    # Wrap the raw fileno into a socket object
    sockobj = _stdlib_socket.socket(fileno=fileno)
    try:
        if hasattr(socket, "SO_DOMAIN"):
            family = sockobj.getsockopt(SOL_SOCKET, SO_DOMAIN)
        if hasattr(socket, "SO_TYPE"):
            type = sockobj.getsockopt(SOL_SOCKET, SO_TYPE)
        if hasattr(socket, "SO_PROTOCOL"):
            proto = sockobj.getsockopt(SOL_SOCKET, SO_PROTOCOL)
    finally:
        # Unwrap it again, so that sockobj.__del__ doesn't try to close our socket
        sockobj.detach()
    return family, type, proto
```

And then in `fromfd` and `socket` we can do something like:

```python
family, type, proto = _fix_attrs_for_fileno(family, type, proto, fileno)
```

Does that make sense?

njsmith:
@2easy Hey, just wanted to check in to see whether you're still interested in working on this :-)

2easy:
Sure thing, sorry for the delay. I'll fix it in the evening.

njsmith:
Haven't heard anything for a week, so I'm going to close this for now to take it off the list of active PRs. Feel free to reopen it, or open a new PR, if you do decide to come back to it later! We've all been there :-)

Issue #578, opened 2018-07-29
Arbitrary kevent support


sorcio:
This is part of the long term plan to implement subprocess suport as outlined in https://github.com/python-trio/trio/issues/4#issuecomment-398967572 (see "Step 3", copied/pasted below for convenience).

---

On MacOS and the BSDs, there's this neat thing called "kqueue", which lets you efficiently wait on all kinds of different events in an async-event-loop-friendly way. Currently Trio just uses this to wait for file descriptors to become readable/writable, but in general kqueue is much more flexible. For example, the `EVFILT_PROC` event type lets you wait for a subprocess to exit.

`trio/_core/_io_kqueue.py` should provide the ability to wait on arbitrary event types. (Then we'll use that in our subprocess support to wait for `EVFILT_PROC` events.)

There's already a stub implementation of this, in the `monitor_kevent` method. This isn't tested or used currently though, so if we're going to start using it we'd want to make sure it actually works. (And possibly change how it works, if there's some other semantics that make it easier to use.)

arthur-tacca:
For subscribers to this issue: there is related discussion #3500 (later comments) and #2922.

PR #579, opened 2018-07-29
monitoring arbitrary kevents

sorcio:
So, this is just to start thinking how we can implement support for arbitrary kevent events (#578). I'm not a kqueue expert at all, I just happen to run macOS. @njsmith is this similar to what you had in mind?

I think there are weird interactions between flags that are not apparent from the documentation. For example it appears that waiting for process exit will get an event with EV_ONESHOT which makes sense (you can't wait for an event from a pid that doesn't exist anymore). So maybe making it actually *arbitrary* involves some more tricks that I haven't explored yet.

But it seems to work ok to wait for a child process to terminate. I'm afraid I might be missing something for that use case too, though, because there is a race condition between process creation and opening a monitor. If the process has already terminated we get an exception, or we might even monitor the wrong process. I'm not sure this is relevant to this issue. Maybe the API should be changed for the bigger picture but right now I cannot suggest how.

### DIFF
--- a/trio/_core/tests/test_kqueue.py
+++ b/trio/_core/tests/test_kqueue.py
@@ -1,0 +1,20 @@
+import select
+import subprocess
+
+from ... import _core
+
+
+async def test_kqueue_monitor_proc_event():
+    p = subprocess.Popen(['sleep', '0.1'])
+    ident = p.pid
+    filter = select.KQ_FILTER_PROC
+    fflags = select.KQ_NOTE_EXIT
+    with _core.monitor_kevent(ident, filter, fflags) as q:
+        async for batch in q:
+            for evt in batch:
+                assert evt.ident == ident
+                assert evt.filter == filter
+                assert evt.fflags == fflags
+                assert evt.data == 0  # exit code
+                # we only expect one event for KQ_NOTE_EXIT
+                return
--- a/trio/_core/tests/test_kqueue.py
+++ b/trio/_core/tests/test_kqueue.py
@@ -1,7 +1,12 @@
 import select
 import subprocess
 
+import pytest
+
 from ... import _core
+
+use_kqueue = hasattr(select, 'kqueue')
+pytestmark = pytest.mark.skipif(not use_kqueue, reason="kqueue platforms only")
 
 
 async def test_kqueue_monitor_proc_event():
--- a/trio/_core/_io_kqueue.py
+++ b/trio/_core/_io_kqueue.py
@@ -79,19 +79,32 @@ class KqueueIOManager:
 
     @_public
     @contextmanager
-    def monitor_kevent(self, ident, filter):
+    def monitor_kevent(self, ident, filter, fflags=0):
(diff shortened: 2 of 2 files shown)


njsmith:
> there is a race condition between process creation and opening a monitor. If the process has already terminated we get an exception, or we might even monitor the wrong process. I'm not sure this is relevant to this issue

So this is one of those classic weird and subtle Unix things. You've heard of "zombie processes"? On Unix when a process exits, the kernel cleans up most of it (the address space, the threads, etc.), but it keeps around a little bit of metadata (the process name, the return value) – this is a "zombie process". This metadata *stays in the process table under the original process's PID*, until someone calls `waitpid`, which consumes the zombie (om nom nom) and frees up that slot in the process table.

So after you fork a child process, the pid you get back is guaranteed to keep referring to that exact process – or at least, its undead remains – until you call `waitpid`. This avoids the race condition. (Unless someone else calls `waitpid` on our process without our knowledge, of course, but hey, that's life.)

This does mean that you have to be careful that for each child process, you call `waitpid` exactly once, and that once you've called `waitpid` you can't use `EVFILT_PROC`. But that's a concern for higher-level code to worry about, not something that we need handle down here in the low-level kevent API.

trio/_core/tests/test_kqueue.py
```diff
@@ -0,0 +1,25 @@
...
+import pytest
+
+from ... import _core
+
+use_kqueue = hasattr(select, 'kqueue')
+pytestmark = pytest.mark.skipif(not use_kqueue, reason="kqueue platforms only")
+
+
+async def test_kqueue_monitor_proc_event():
+    p = subprocess.Popen(['sleep', '0.1'])
```
njsmith: So all that stuff about zombies means you don't need this process to sleep. Maybe `true` and `false` and then you can also check the exit code?

sorcio:
Not sure if Popen does something I don't understand compared to a naked fork/exec, but I cannot register a kevent for a child process pid after that process has terminated. I will definitely look further into it, because I too expected it to provide some way to deal with zombies. Now that I think of it, I was so confused by this that I forgot to reap the child :)

Thanks for the review and the excellent write-up! I will follow up on that later.

njsmith:
I guess `subprocess` could be automatically reaping the zombie before we notice, but looking at the code I think it would only do this if you let the `Popen` object get garbage collected, so probably not.

It's also possible that kqueue simple doesn't let you register a filter for a process that's already dead. Maybe you're expected to take the error message as your notification that the process is finished and you should call `waitpid` :-). If that's true, then testing purposes we could do something like spawning a `sleep 9999`, registering the kqueue filter, and then killing the process.

njsmith:
We should probably still do a pass over the low-level kqueue APIs to clean them up, but this PR has stalled and the discussion is way out of date (e.g. trio mainline is now actually using kqueue to monitor process termination). So I'm gonna close it, and we can always re-open if that turns out to be useful.

PR #587, opened 2018-07-30
Add CapacityLimiter.UNLIMITED

njsmith:
Sometimes we have an API that wants a CapacityLimiter (like
run_sync_soon_in_worker_thread), and we want to have unlimited
capacity (for example, when putting a long-running blocking call into
a background thread, like WaitForMultipleObjects or waitpid). This can
be done by writing a little stub class that implements the
CapacityLimiter interface, but doing this all the time is annoying.
Here's a full-fledged version, available whenever you need it.

### DIFF
--- a/newsfragments/587.feature.rst
+++ b/newsfragments/587.feature.rst
@@ -1,0 +1,2 @@
+New singleton object :data:`trio.CapacityLimiter.UNLIMITED`, for when
+you need a :class:`CapacityLimiter` with infinite capacity.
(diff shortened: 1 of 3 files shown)


sorcio:
Worth trying to document why it might be *needed*? Or maybe mention one example? I can’t think of a short and effective explanation, and there are going to be more use cases than we can show; so maybe this is a hint that we shouldn’t and just leave it to users to understand whether this is what they need. On the other hand, this has the potential to cause some confusion and can be misused, e.g. as a shortcut not to think too hard about what limit to impose on thread spawning.

njsmith:
You know, that's an excellent question. I spent a while trying to document when you would use this, and couldn't figure out how to do it. I guess that means it's a bad idea.

Maybe we should consider allow `CapacityLimiter(inf)` as a more obscure version? I'm not sure. I'll reject this PR anyway.

Issue #580, opened 2018-07-29
test_SocketType_resolve fails


sorcio:
Apparently on Mac, on *some* setups, `getaddrinfo()` will always return IPv4 mapped addresses regardless of whether the AI_V4MAPPED is actually set. This causes failures in this assertion https://github.com/python-trio/trio/blob/fd0d8770a120a09a506ccd8a3f9969b26a22f38e/trio/tests/test_socket.py#L420-L422

This is probably a system configuration issue and maybe there is not much we can do about it. But it's causing our tests to fail on some machines. I've seen this happen regularly on my work laptop for some time (macOS 10.13.5, Python 3.6 and 3.7) and a couple other sprinters at EuroPython had the same issue. Good news is that it doesn't affect our Mac CI.

One potential (maybe not ideal) solution is to verify whether the system has this quirk and in that case skip the test.

### Details

Calling getaddrinfo() on an IPv4 address, with AF_INET6 and the AI_V4MAPPED flag returns an IPv4-mapped IPv6 address. This is documented and works in all platforms.

```python
>>> getaddrinfo('1.2.3.4', 0, AF_INET6, SOCK_STREAM, 0, AI_V4MAPPED)
[(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('::ffff:1.2.3.4', 0, 0, 0))]
```

If I only want IPv6 addresses I won't set the AI_V4MAPPED and I won't get any result back. For example, this is what happens on Linux:

```
>>> s.getaddrinfo('1.2.3.4', 0, s.AF_INET6, s.SOCK_STREAM, 0, 0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/socket.py", line 748, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -9] Address family for hostname not supported
```

But if I try the same on my Mac I still get the IPv4-mapped results, even though I shouldn't!

```python
>>> getaddrinfo('1.2.3.4', 0, AF_INET6, SOCK_STREAM, 0, 0)
[(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('::ffff:1.2.3.4', 0, 0, 0))]
```

njsmith:
Huh, weird.

[...]

PR #589, opened 2018-07-31
Skip a test on systems with buggy AI_V4MAPPED handling

njsmith:
Closes gh-580

@sorcio: This is an unusual bug in that we actually do not know which
systems are affected or why, and we don't have any of the affected
systems in our CI setup. So even if the tests pass here I have no idea
if it's actually working :-). Can you test this branch and confirm it
fixes the failure you were seeing?

### DIFF
--- a/newsfragments/580.bugfix.rst
+++ b/newsfragments/580.bugfix.rst
@@ -1,0 +1,3 @@
+Some version of MacOS have a buggy ``getaddrinfo`` that was causing
+spurious test failures; we now detect those systems and skip the
+relevant test when found.
--- a/trio/_socket.py
+++ b/trio/_socket.py
@@ -523,7 +523,7 @@ class _SocketType(SocketType):
         # Special cases to match the stdlib, see gh-277
         if host == "":
             host = None
-        if host == "<broadcast>" and self._sock.family == AF_INET:
+        if host == "<broadcast>":
             host = "255.255.255.255"
         # Since we always pass in an explicit family here, AI_ADDRCONFIG
         # doesn't add any value -- if we have no ipv6 connectivity and are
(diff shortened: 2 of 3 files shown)


sorcio:
It was still failing on another assertion because 255.255.255.255 also gets v4-mapped. I pushed directly to your branch because it was open to commits. It's all green on my machine, so I guess that did the trick. I'd be more comfortable if I actually understood what is going on but I'm happy with a working test suite.

My added bit changes public behavior (to more closely match stdlib). Does it need a mention in the newsfragment, or even an issue of its own?

njsmith:
Oh, good catch. Hmm. I'm not sure why we support this `<broadcast>` thing at all, given that (a) people can just type `255.255.255.255` if that's what they want, and (b) `255.255.255.255` is not really useful these days anyway. But if we're going to make `<broadcast>` be a magic string, I guess we should make it magic in all cases, instead of actually passing it through to `getaddrinfo` and relying on that to fail, which I guess is what's happening with your patch? The stdlib actually raises `OSError` if you try to use `<broadcast>` with an IPv6 socket (even if the socket can handle other IPv4 addresses):

https://github.com/python/cpython/blob/944451cd8d3e897138f4b43569de13cd081ee251/Modules/socketmodule.c#L1065-L1072

Given all this, I guess my preference would be to have `<broadcast>` act exactly like `255.255.255.255`. Which is what we are actually doing now, but the test has an incorrect idea about how `255.255.255.255` works, and then gets confused when `<broadcast>` acts like `255.255.255.255` instead of acting like it expects.

I've just pushed a simple fix for this: run that test on the socket with `AI_V4MAPPED` enabled, so this stuff doesn't affect it :-).

sorcio:
Ah, uh. I had only compared to stdlib `getaddrinfo`, but not general socket operation (`bind`, `connect`). Of course, this is what the stdlib does. All green on my quirky macOS, I will merge!

PR #594, opened 2018-08-07
Simplify imports

jmfrank63:
As advised I replaced all import * with explicit imports in `__init__.py` of the trio root.

I tested against vscode code completion and it seems to work. I did not run any tests yet.

Also pylint does not complain anymore about missing attributes.

The issue was #542.


Zac-HD:
This looks great!  You can see the formatting issues Travis is complaining about [here](https://travis-ci.org/python-trio/trio/jobs/413179559), and fix them by running 

    pip install yapf==0.20.1
    yapf -rpi setup.py trio

After that, we'll need a test that every name in the `__all__` list appears in the trio namespace, i.e.

```python
def test_dunder_all():
    for name in trio.__all__:
        assert hasattr(trio, name)
```

This will ensure that we keep the imports up to date when new names are added at lower levels.

The optional last step would be to check that the dynamically-created `__all__` works for pylint and mypy; and if not to declare it up front and write another test that it's consistent with the lower-level `__all__`s.  However, that step can be left to a future pull request by someone else.

Finally - and most importantly - thanks for contributing!  😍 🎉

trio/_version.py
```diff
@@ -1,3 +1,3 @@
 # This file is imported from __init__.py and exec'd from setup.py
 
-__version__ = "0.5.0+dev"
+__version__ = "0.5.1+dev"
```
njsmith: No need to change the version here :-) We'll do that when we release.

Zac-HD:
@njsmith - at runtime, `__all__` just defines which names are imported by `from foo import *`.  However it's also widely used for static analysis (eg pylint and mypy), because PEP8 defines names which appear in `__all__` as public API.

Since the goal of #542 was to improve the static analyse-ability of Trio, I'd like to keep or write `__all__` literals for all our public modules.  (this is actually the main blocker for static typing, FWIW)

njsmith:
@Zac-HD Huh, the rule for `import *` is that if there's an `__all__`, it uses that, and if there isn't, it defaults to "all non-underscored names". I would have assumed that pylint/mypy/etc. knew that rule. Are you sure pylint etc. really treat these two modules differently?

```python
# module 1
def f():
    pass
```

```python
# module 2
def f()
    pass
__all__ = ["f"]
```

And if so, I'm super curious what the difference is...

trio/__init__.py
```diff
@@ -14,49 +14,85 @@
...
+                                        MultiError,
+                                        run,
+                                        open_nursery,
+                                        open_cancel_scope,
+                                        current_effective_deadline,
+                                        TASK_STATUS_IGNORED,
+                                        current_time,
+                                        TaskLocal,
+                                        __all__)
+
```
dhirschfeld: Not sure what the trio style is but rather than have towers of arguments at various arbitrary columns I personally prefer to drop down to the next line and only indent one level:
```python
from ._toplevel_core_reexports import (
    ClosedResourceError,
    MultiError,
    ResourceBusyError,
    RunFinishedError,
    TrioInternalError,
    Cancelled,
    WouldBlock,
    TaskLocal,
    TASK_STATUS_IGNORED,
    current_effective_deadline,
    current_time,
    open_cancel_scope,
    open_nursery,
    run,
)
```
...this also allows for a more compact form:
```python
from ._toplevel_core_reexports import (
    ClosedResourceError, MultiError, ResourceBusyError, RunFinishedError,
    TrioInternalError, Cancelled, WouldBlock, TaskLocal, TASK_STATUS_IGNORED,
    current_effective_deadline, current_time, open_cancel_scope, open_nursery,
    run,
)
```

Zac-HD: Yep, the compact form is [enforced by `yapf`](https://travis-ci.org/python-trio/trio/jobs/413179559#L2944); see my comment above re: Travis.

njsmith: Yapf allows either of the two forms @dhirschfeld showed – IIRC you get the first one if you put a trailing comma on the last item, and the "more compact" one if you don't use a trailing comma.

njsmith:
@Zac-HD
> In particular I've encountered a few weird issues with non-literal `__all__` in e.g. `pydocstyle`

Right, that's one of the motivations here (and see also #314, which has some commentary from one of the PyCharm devs on what it can/can't understand).

Let me try phrasing my question a different way. For `trio/__init__.py`, I think we basically have these four options:

```python
# option A
from ._submodule1 import *
__all__ += _submodule1.__all__
from ._submodule2 import *
__all__ += _submodule2.__all__

# option B
from ._submodule1 import X, Y
__all__ += _submodule1.__all__
from ._submodule2 import Z
__all__ += _submodule2.__all__

# Option C
from ._submodule1 import X, Y
from ._submodule2 import Z
__all__ = ["X", "Y", "Z"]

# option D
from ._submodule1 import X, Y
from ._submodule2 import Z
# (__all__ is left undefined)
```

Trio master currently uses option A, and that hasn't worked out so well. So now we're picking between B, C, or D.

With B, the `__all__ += _submodule.__all__` thing may or may not be understood by static analysis tools. Options C and D don't have that problem.

With B and C, the definition of `__all__` is more annoying for us to maintain, because now every time we add a new export, we have to manually update both the `import` line *and* the definition of `__all__`, and make sure they stay in sync. (With B we have to manually update `trio._submodule.__all__`; with C we have to manually update `trio.__all__`, but either way it's two places that need updating.)

Option D, it seems to me, doesn't have either problem: the exports are clearly visible to static analysis tools, there's no chance for `__all__` to confuse static analysis because there is no `__all__`, and the exports are only listed once, so maintenance is easier.

In fact I assume mypy has to mostly ignore `__all__`, because when it sees something like `trio.open_nursery` it has to track that back to a definition so it can find the type signature – and only the `import` line can help it do that.

Zac-HD:
I'd be happy with either C or D and leaning towards D as the less-effort option. 

Either way, it would be good to write a test that everything in the lower-level modules `__all__` is exported at the top level.

njsmith:
> Either way, it would be good to write a test that everything in the lower-level modules `__all__` is exported at the top level.

I'd be tempted to get rid of the lower-level `__all__`'s too. If we're not using them, then what's the point of maintaining and testing them?

I guess the case where this test would catch a bug is: someone adds a new symbol, and remembers to add it to the module's `__all__`, but forgets to add it to the higher-level `import` line – AND if we didn't have the `__all__` then they would have forgotten to export it anywhere. I guess this scenario could happen? We all forget stuff sometimes. But I feel like if someone's going to forget to export something, then it doesn't help to have *two* places they're supposed to export it.

jmfrank63:
Hello Zac, Nathaniel,

thanks for accepting me as a contributor. I am still a freshman, so I will
follow your recommendations:

Use yapf and fix the formating (I used the parenthesis because of PEP 328.
I think we should keep this, but put more imports on each line not
exceeding col 80)
Add a test (I wonder if this will still work with Option D in place)
Replace all * imports with explicit versions. (Should I confine this only
to __init__.py or cover all modules like (abc.py e.g))
Revert the version string to the release version

Is there some mechanism I can mark this as a dev branch of my own (besides
the git branch). I used this to make sure I really was on my branch and it
helped me to distinguish the two.

Thanks

Johannes

njsmith:
@dhirschfeld This is one of the frustrating things about this yeah – it's really hard to know what works and what doesn't :-(. To make sure I understand correctly: you're saying that if you write:

```python
# submodule.py
# has a mix of public and private with literal __all__
a = 1
b = 2
__all__ = ["a"]

# __init__.py
from .submodule import *
```

... Then both Jedi and VSCode think that now `b` is available in the `__init__.py` namespace?

I was going to say well, maybe that's the one case that *works* and we can use it for the simpler parts of trio where that pattern applies, but if even that doesn't work then never mind.

njsmith:
@jmfrank63 I think first let's settle how we want to handle `trio/__init__.py`. Then we'll probably want to apply the same change to the other files that currently follow the same pattern:

* `trio/hazmat.py`
* `trio/testing/__init__.py`
* `trio/_core/__init__.py`

For `hazmat` and `_core` this will be a little tricky because their exports are different on different platforms. But we can at least apply this trick to the parts that aren't platform-dependent.

And then after that, we'll want to think about what to do with the `trio.abc` and `trio.socket` modules, which are a bit different.

dhirschfeld:
> Then both Jedi and VSCode think that now b is available in the __init__.py namespace?

Yep, that appears to be the case - i.e. `val = mymodule.b` didn't show as an error using static analysis and `b` was a tab-completion option. In a python terminal with live objects `b` correctly didn't show up as an option. I was surprised by that, so much so that I'd call it a bug in static analysis!

I wonder if PyCharm does any better... might check that out.

jmfrank63:
Hi I did the formatting and reverted the version. Currently there is not a test for this. I will need some guidance on the tests, as testing is an absolute weak spot with me.
I left `__all__`intact, but I moved the call to the globals into a function so we can import it and call it for the reexport. I thought this will not hurt but might be beneficial for the future.
I don't think it is called anyway now that we do an explicit import and not a *, but I could be wrong.
Pycharm still works with the git repro installed and shows all the completions vscode with the python extension shows.
Jedi vim is currently in progress, setup is tricky due to the lack of virtualenv picked up by jedi.

njsmith:
@jmfrank63 Thanks for checking in with the pylint developers – that's super interesting to hear that they explicitly *don't* use `__all__`. (Link to the pylint issue: https://github.com/PyCQA/pylint/issues/2382#issuecomment-411825973. By the way, as a general tip, I like to link between issues like this as much as I can, because it helps when someone is trying to track down why a decision was made like 5 years later and they have no idea what's going on. I guess github has some official syntax for linking to issues in other repositories, but I can't be bothered to learn it – I just cut and paste the actual url into my comment, and then github notices that it's a github url and renders it nicely.)

For this PR:

* It sounds like it's more or less settled now that `__all__` isn't doing anything useful. Can you delete all the `__all__` update lines in `trio/__init__.py`, and also in all the `trio/_something.py` files that the `__init__.py` file imports from? (Just those files, not any others – for example `trio/_abc.py` will probably need some changes eventually but for this PR we shouldn't change it.)

Regarding `trio/_toplevel_core_reexports.py`: this whole file only exists because of an earlier attempt to fix this same issue. I guess it helps with pycharm, but apparently not with any other analyzer (because all it does is try to make `__all__` more visible). Our new strategy replaces it. So we should delete this file entirely, and in `__init__.py` we should write:

```python
from ._core import (
   # ... everything we currently import from _toplevel_core_reexports
)
```

For writing a test: the basic idea is to open one of the files in the `tests` directory – for this I would use `trio/tests/test_exports.py` – and add a function named `test_<whatever you like>`. The function should assert that whatever it's testing is true. Then to run the tests, use pytest (there's some more details here: https://trio.readthedocs.io/en/latest/contributing.html#pull-request-tests). If you haven't done this before, I'd try just locally playing around with writing tests like `assert 1 + 1 == 2` and `assert 1 + 1 == 987` to get the hang of it.

Then for the actual test, I suppose we want to use the pylint and/or Jedi APIs to query for what exports they see, and compare that to what we see when looking at the same namespace dynamically. For an example of how to use the pylint API, [see here](https://gitter.im/python-trio/general?at=5b69c2c05d1362758b44afcb). Jedi has an example of how to get completions [on the first page of its docs](https://jedi.readthedocs.io/en/latest/); we could ask it for completions for a script like `"import trio; trio."`.

Please do ask for help if you get stuck!

jmfrank63:
The changes just as they are get a difference in jedi of
```
{'__builtins__', '__cached__', '__loader__', '__path__', '__spec__'}
```
and zero difference with pylint.

njsmith:
We could move the `fixup_module_metadata` stuff to a submodule, if it makes this easier to test. Or just `import fixup_module_metadata as _fixup_module_metadata`.

jmfrank63:
I wrote a first test against pylint:

```
def test_pylint_sees_all_names_in_namespace():
    # Test pylints ast to contain the same content as dir(trio)
    
    linter = PyLinter()
    ast_set = set(linter.get_ast('trio/__init__.py', 'trio'))
    trio_set = set([symbol for symbol in dir(trio) if symbol[0] != '_'])
    trio_set.remove('tests')
    assert trio_set - ast_set == set([])
```
However since testing is not included by default (commented out) the test fails on the testing module not being in the namespace.

However this one passes the test:
```
TESTING = False
if TESTING:
    from . import testing
```
So we could define an environment variable that if not present skips the import otherwise imports.
Any suggestions?

njsmith:
I think it'd be fine for the test to special-case `"testing"` as a name that pylint/jedi might not pick up.

trio/tests/test_exports.py
```diff
@@ -18,3 +23,21 @@ def test_core_is_properly_reexported():
...
+    linter = PyLinter()
+    ast_set = set(linter.get_ast(trio.__file__, 'trio'))
+    trio_set = set([symbol for symbol in dir(trio) if symbol[0] != '_'])
+    trio_set.remove('tests')
+    assert trio_set - ast_set == set([])
+
+
+def test_jedi_sees_all_completions():
+    # Test the jedi completion library get all in dir(trio)
+    script = jedi.Script(path=trio.__file__)
```
njsmith: I think the way `jedi.Script` is supposed to be used is maybe more like `jedi.Script("import trio\ntrio.")`? (As in, we imagine somebody typed those characters into a file and then hit tab to get a list of completions.)

trio/tests/test_exports.py
```diff
@@ -18,3 +23,21 @@ def test_core_is_properly_reexported():
...
         print(symbol, found)
         assert found == 1
+
+
+def test_pylint_sees_all_non_underscore_symbols_in_namespace():
+    # Test pylints ast to contain the same content as dir(trio)
+    linter = PyLinter()
+    ast_set = set(linter.get_ast(trio.__file__, 'trio'))
+    trio_set = set([symbol for symbol in dir(trio) if symbol[0] != '_'])
+    trio_set.remove('tests')
```
njsmith: Probably worthwhile to factor out some of the namespace cleanup code from the two functions – I think in both cases we want to test something like: "the two sets are equivalent, if you ignore names that `.startswith("_")`, `"tests"`, and `"testing"`"?

jmfrank63: Ok, I'll work myself through all  the comments. Will take some time as this is all new stuff.

njsmith: No worries, take your time, and don't be afraid to ask for help if you get stuck :-)

njsmith:
@jmfrank63 Ah, so what's happened is you've discovered some places where we were unexpectedly relying on `__all__`. Exciting! :-)

In `test_exports`, it checks for `symbol in source.__all__`. Here `symbol` is the name of something exported from `trio._core`, like `"TrioExternalError"`, and `source` is one of the modules that's supposed to re-export those symbols, like `trio`. But now we've gotten rid of `trio.__all__`, so this test no longer makes sense – we *don't* want `"TrioExternalError"` to show up in `trio.__all__`.

The quick fix is replace `symbol in source.__all__` with `hasattr(source, symbol)`. Later on when we clean up `trio/_core/__init__.py` we might also need to adjust the `for symbol in _core.__all__` loop that's a few lines up, but for right now that's not necessary.

In `test_module_metadata_is_fixed_up`, the problem is an actual bug in `fixup_module_metadata`. Here's the source:

https://github.com/python-trio/trio/blob/ffb5bca1009d84716719b675345b5adc0aed2ac4/trio/_util.py#L171-L182

Notice how on line 180, we've got another of those references to `__all__`. The goal of this function is to take a module namespace, go through all the objects it exports, and fix up their `__module__` attribute so that when someone later tries to figure out the public name of this object, they get the right thing. For example, at the bottom of tracebacks it shows you the name of the exception type:

```
  File "<stdin>", line 1, in <module>
trio.TooSlowError: timeout expired
```

Notice how it says `trio.TooSlowError` there? That `trio` is coming from the exception type's `__module__` attribute. If we didn't rewrite it, then it would instead say `trio._timeouts.TooSlowError: timeout expired`, which exposes internal implementation details and is also kinda weird and distracting.

Anyway, those details don't really matter right now, just giving you some context :-). The specific issue is that the function needs to iterate over all the public exports in a namespace, and it's trying to do that using `__all__`, but of course that doesn't work once we stop using `__all__` to control public exports. Instead, we can rewrite that loop like:

```python
for objname, obj in namespace.items():
    if not objname.startswith("_"):  # ignore private attributes
        fix_one(obj)
```

jmfrank63:
Great, I'll try this and see how it works. Currently moving my home so some delay may occur.

njsmith:
Filed a bug upstream with jedi to ask about the skipped test: https://github.com/davidhalter/parso/issues/47

njsmith:
Github's changes view was being super annoying, so I used [this trick](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch) to reset the PR's branch point. (hat-tip to @pquentin for finding that.)

trio/__init__.py
```diff
@@ -13,61 +13,81 @@
...
+    "ClosedResourceError",
+    "MultiError",
+    "run",
+    "open_nursery",
+    "open_cancel_scope",
+    "current_effective_deadline",
+    "TASK_STATUS_IGNORED",
+    "current_time",
+    "TaskLocal",
+]
```
njsmith: So we don't actually need `__all__` at all :-)

If you do `from module import *`, then Python:

1. Checks for an `__all__`. If it finds one, then that's what it uses.
2. Otherwise, it imports all names that are defined in the module, and don't start with an underscore.

So we can (and should) delete this `__all__` entirely. So long as we're careful about what names we export – which is something we need to do anyway – then it's totally redundant.

trio/__init__.py
```diff
@@ -13,61 +13,81 @@
...
-__all__ = []
+from ._version import __version__
 
-from ._toplevel_core_reexports import *
-__all__ += _toplevel_core_reexports.__all__
+from ._core import (
+    TrioInternalError, RunFinishedError, WouldBlock, Cancelled,
+    ResourceBusyError, ClosedResourceError, MultiError, run, open_nursery,
+    open_cancel_scope, current_effective_deadline, TASK_STATUS_IGNORED,
+    current_time, TaskLocal, cancel_shielded_checkpoint
```
njsmith: I'm pretty sure `cancel_shielded_checkpoint` is not supposed to be listed here?

trio/__init__.py
```diff
@@ -13,61 +13,81 @@
...
 
-from ._highlevel_generic import *
-__all__ += _highlevel_generic.__all__
+from ._highlevel_generic import (
+    aclose_forcefully, BrokenStreamError, StapledStream
+)
 
-from ._signals import *
-__all__ += _signals.__all__
+from ._signals import catch_signals
```
njsmith: Since #619 just got merged, you'll want to update this to also import `open_signal_receiver`.

trio/_core/__init__.py
```diff
@@ -14,7 +14,22 @@ def _public(fn):
...
+    "ClosedResourceError",
+    "MultiError",
+    "run",
+    "open_nursery",
+    "open_cancel_scope",
+    "current_effective_deadline",
+    "TASK_STATUS_IGNORED",
+    "current_time",
+    "TaskLocal",
+]
```
njsmith: Not sure what this is trying to do?

Let's put this back to `__all__ = []` for now, and then make fixing `_core/__init__.py` another PR :-)

.gitignore
```diff
@@ -63,3 +63,15 @@ coverage.xml
...
+<<<<<<< HEAD
+Pipfile Pipfile.lock
+=======
+Pipfile
+Pipfile.lock
+>>>>>>> Formatting with yapf applied.\nVersion reverted to orig version 0.5.0+dev\nPipenv files added to gitignore
+=======
+Pipfile
+Pipfile.lock
+>>>>>>> 47197bf4922eb92510defd465e78bb6d375ccbe9
```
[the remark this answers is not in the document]
jmfrank63: Yes, I had something like this with the version file as well. As I said the rebase was frantic. I might have typed the wrong command. 
What should I do to prevent this the next time and more important fix this. What worries me is that this went undetected by the testing. Is there some kind of test configuration testing?

trio/__init__.py
```diff
@@ -13,61 +13,81 @@
...
 
 # Imported by default
 from . import hazmat
 from . import socket
 from . import abc
 from . import ssl
 # Not imported by default: testing
+TESTING = False
+if TESTING:
+    from . import testing
```
njsmith: If we do this then it should be `if False: ...` instead of defining a variable `TESTING`. (Remember: every variable we define in this file becomes part of trio's public API!)

I *think* we probably don't need this? From some playing around, it looks like `jedi` is happy to suggest `testing` as a completion even if we don't mention it here:

```python
In [15]: "testing" in [v.name for v in jedi.Script("import trio; trio.").completions()]
Out[15]: True
```

And if I put this into a file and run `pylint` on it, then it doesn't warn on the mention of `trio.testing`, but does warn on the mention of `trio.testingxxx`, so it seems that pylint has also figured out that `trio.testing` is a valid thing:

```python
import trio
print(trio.testing)
print(trio.testingxxx)
```

(These are both with trio v0.6.0, since I happened to have it lying around.)

So there's some puzzle about how pylint actually decides what names are in a namespace: it knows about `testing`, but somehow the way we're querying it currently doesn't...

For now please either make this `if False: ...` or else special-case the name `testing` in our pylint test, and we should look to see if we can figure out pylint's API better...

trio/tests/test_exports.py
```diff
@@ -18,3 +25,25 @@ def test_core_is_properly_reexported():
...
+    assert trio_set - ast_set == set([])
+
+
+@pytest.mark.skipif(
+    sys.version_info.minor > 7 and sys.version_info.releaselevel != 'final',
+    reason="jedi 3.8-dev not supported"
+)
+def test_jedi_sees_all_completions():
+    # Test the jedi completion library get all in dir(trio)
+    script = jedi.Script(path=trio.__file__)
```
njsmith: I think this should be `script = jedi.Script("import trio; trio.")`

Right now you're telling jedi: "hey, if I had `trio/__init__.py` open in my editor, and went to the last line (or something like that?), and hit TAB, what completions should be displayed?" Which is a bit strange. My snippet says "hey, if I typed `import trio; trio.` in my editor and then hit TAB, what should be displayed?" Which seems like what we want :-)

.gitignore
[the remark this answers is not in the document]
njsmith: We don't have any testing for `.gitignore`. I'm not sure how to do that.

I guess we could have a test for conflict markers like this being checked in. In practice it's unlikely to slip by reviewers, though, so it doesn't seem like a high priority.

trio/tests/test_exports.py
```diff
@@ -18,3 +25,25 @@ def test_core_is_properly_reexported():
...
+    ast_set = set(linter.get_ast(trio.__file__, 'trio'))
+    trio_set = set([symbol for symbol in dir(trio) if symbol[0] != '_'])
+    trio_set.remove('tests')
+    assert trio_set - ast_set == set([])
+
+
+@pytest.mark.skipif(
+    sys.version_info.minor > 7 and sys.version_info.releaselevel != 'final',
+    reason="jedi test for python 3.8-dev not supported"
+)
```
njsmith: On further thought... instead of using `skipif` here, I think this is a case where it's better to follow the ["EAFP principle"](https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python). So we should always run the test, but wrap the call to `jedi.Script` in a `try`, like:

```python
try:
    script = jedi.Script("import trio; trio.")
except NotImplementedError:  # pragma: no cover
    pytest.skip("jedi does not yet support this version of python")
```

The advantage of doing it this way, is that it automatically works for all versions of Python – as soon as jedi starts supporting 3.8, the test will start running, and when we start testing on 3.9-dev, we won't have to adjust anything...

njsmith:
Regarding pylint and `trio.testing`: I dug into this a little more.

It looks like the library that pylint actually uses to do static analysis is [astroid](https://astroid.readthedocs.io/en/latest/).

With astroid, I can get a list of all the names that it thinks are in the `trio` module like this:

```python
# Get a node representing accessing a variable named "trio"
ref_node = astroid.extract_node("import trio; trio")
# Ask astroid for the object that it thinks this variable refers to – the actual trio module
# .infer() is a generator, so we need to use list() to force its evaluation
module = list(ref_node.infer())[0]
# Get a list of the names astroid thinks are exported by this module
module.public_names()
```

This list does *not* include `testing`, with current trio.

However, we can also ask a slightly different question: "if someone mentions `trio.testing`, can you figure out what that means?"

```python
# This succeeds (and returns a module object for trio.testing)
ref_node = astroid.extract_node("import trio; trio.testing")
list(ref_node.infer())

# If we try the same trick on a bad name, we get an 'InferenceError' exception
ref_node = astroid.extract_node("import trio; trio.testingxxx")
list(ref_node.infer())
```

So it looks like astroid's two ways of accessing this information don't quite agree. I guess this makes sense – pylint is primarily designed to try to make sense of existing code that contains references like `trio.testing`, and mapping those back to their origin; starting from a module and generating an accurate completion list for it isn't really something it ever needs to do.

This does suggest an interesting thing we could do though: for each name exported by the `trio` namespace, we could ask whether astroid can figure out what it refers to, by doing something like:

```python
for name in trio.__dict__:
    if name.startswith("_"):
        continue
    source = "import trio; trio.{}".format(name)
    # I'm not sure why sometimes this raises an error and other times it return Uninferable
    found = list(astroid.extract_node(source).infer())
    assert len(found) > 0
    assert found[0] is not astroid.Uninferable
```

This is a much stronger test than just "can you figure out that this name is defined?". It actually requires that astroid can track its import chain all the back to the original object. So we might not want to do this now, because we probably can't pass this test until we finish fixing `trio._core`. But eventually this is probably a useful test, because this stronger kind of analysis is what you need to do things like type checking.

jmfrank63:
I did already try to make _core to be explicit already, but threw it away because I got stuck with the `__all__` dependency. It was moderate work. If you don't mind I try to take on this too, and try to use your test suggestion and see if we can pass it.

njsmith:
> If you don't mind I try to take on this too, and try to use your test suggestion and see if we can pass it.

That would be awesome BUT I think first we should finish off this PR and get it merged, and then do that as a second PR :-). The longer you have a single PR open the more time there is for it to diverge from mainline and you get issues like the one with `open_signal_receiver`. So better to split it into pieces when you can.

jmfrank63:
Makes sense, thanks for all the help. I'll try to GSD now ;-)

jmfrank63:
Hmmh, the `open_signal_receiver` still around. Any hints how to resolve this? Has to do with the if hasattr import but ...

pquentin:
@jmfrank63 open_signal_received is still around in your branch, but not in the master branch, that's the problem! To see the problem, you need another rebase on top of the latest master, or you need to merge master into your branch.

Does that make sense?

.gitignore
```diff
@@ -63,3 +63,6 @@ coverage.xml
 
 # Sphinx documentation
 doc/_build/
+.vscode/
+Pipfile
```
belm0: Pipfile and Pipfile.lock files are newly added in this PR, which seems at odds with having .gitignore entries.  Are the files supposed to be in the PR?

njsmith:
The `CHECK_DOCS` failure is because sphinx is trying to look up `trio.open_signal_receiver` (added in #619), and your `trio/__init__.py` isn't currently re-exporting it, so sphinx can't find it.

You mentioned in gitter that you're running yapf locally and getting a different result than when the CI runs yapf, even though you're using the correct yapf version. I have no idea why that would happen! But I would ignore that for now – once things are ready to merge otherwise, either you can run yapf then, or someone else can do it for you.

jmfrank63:
@njsmith @belm0 @pquentin First thanks for all the help. 
Yes, it seems I learn git the hard way. As the recommendation is it is salvageable, I'll do this. I'll remove the three files and work on the recommendations. Nathaniel, sorry when I overlooked a recommendation, I'll walk through the whole thread again and make sure nothing is missed.

njsmith:
Hi @jmfrank63 – are you still working on this? (No worries if not – we figured out a lot about what needs to be done that someone else could pick up from here, and there are plenty of other things to do if you want to get involved again later :-).)

belm0:
I think this is the top user-happiness issue of Trio currently (watchdog seems like a distant 2nd).

@jmfrank63 do you have an idea of time frame?  I'm happy to take a stab this weekend otherwise.

trio/__init__.py
jmfrank63: Decided to go for if False: currently to minimise changes.

docs/source/conf.py
```diff
@@ -70,7 +70,7 @@ def setup(app):
 
 intersphinx_mapping = {
     "python": ('https://docs.python.org/3', None),
-    "outcome": ('https://outcome.readthedocs.org/en/latest/', None),
+    "outcome": ('https://outcome.readthedocs.io/en/latest/', None),
```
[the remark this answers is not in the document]
jmfrank63: Changed this back to .org as it wasn't the problem.

docs/source/contributing.rst
```diff
@@ -347,7 +347,7 @@ Documentation is stored in ``docs/source/*.rst`` and is rendered using
 `Sphinx <http://www.sphinx-doc.org/>`__ with the `sphinxcontrib-trio
 <https://sphinxcontrib-trio.readthedocs.io/en/latest/>`__ extension.
 Documentation is hosted at `Read the Docs
-<https://readthedocs.org/>`__, who take care of automatically
+<https://readthedocs.io/>`__, who take care of automatically
```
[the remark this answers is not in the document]
jmfrank63: Fixed this by reverting back to .org.

trio/__init__.py
```diff
@@ -15,59 +15,62 @@
...
 
-from ._threads import *
-__all__ += _threads.__all__
+from ._highlevel_generic import (
+    aclose_forcefully, BrokenStreamError, StapledStream
+)
 
-from ._highlevel_generic import *
-__all__ += _highlevel_generic.__all__
+from ._signals import catch_signals
```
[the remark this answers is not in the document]
jmfrank63: It seems by merging with 0.70 this resolved itself.

trio/_core/__init__.py
```diff
@@ -14,7 +14,21 @@ def _public(fn):
...
+    "ResourceBusyError",
+    "ClosedResourceError",
+    "MultiError",
+    "run",
+    "open_nursery",
+    "open_cancel_scope",
+    "current_effective_deadline",
+    "TASK_STATUS_IGNORED",
+    "TaskLocal",
+]
```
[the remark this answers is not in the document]
jmfrank63: Removed the list.

test-requirements.txt
```diff
@@ -1,4 +1,4 @@
-pytest >= 3.3  # for catchlog fixture
+pytest == 3.7.4  # for catchlog fixture
```
dhirschfeld: Is there a particular reason for pinning to *exactly* 3.7.4 rather than just using `>=`?

jmfrank63: This was just a workaround to see the real errors. See https://github.com/python-trio/trio/pull/650
The change has been reverted and is done in PR#650

trio/__init__.py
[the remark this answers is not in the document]
jmfrank63: No, it didn't but I added it.

test-requirements.txt
dhirschfeld: Thanks! When making a change such as pinning to a particular version it's useful to users to make a comment (in the commit message) as to why so they know whether or not they can safely ignore it e.g. because the underlying issue has been fixed.

Fixing it in #650 is great because now the blame https://github.com/jmfrank63/trio/blame/528a2986434d644be8138219f0f3087e111d79e8/test-requirements.txt#L1 is linked to the PR and so a user can follow that all the way to #651 to see if the problem has been fixed and they're safe to use pytest 3.8.1 with trio.

njsmith:
We don't need `_toplevel_core_reexports.py` anymore, right? it doesn't look like anything is using it now...

trio/_signals.py
```diff
@@ -177,7 +177,7 @@ def __aiter__(self):
         return self
 
     async def __anext__(self):
-        return { await self._signal_queue.__anext__()}
+        return {await self._signal_queue.__anext__()}
```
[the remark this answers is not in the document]
njsmith: Yeah, this is some weird thing where yapf is apparently doing different for @jmfrank63 than for the rest of us, even with the same version and settings. ¯\\_(ツ)_/¯

And yeah, simplest thing is to manually delete the space...

njsmith:
@Zac-HD Thanks for reviewing! Re:

> Refactoring namespace cleanup code in your excellent tests

I think I'd also be OK with putting this off to a follow-up PR... this one has been dragging on a long time and we don't want it to get stuck again :-)

belm0:
My project just started using pylint for static checking in the continuous build, and many on our team rely on the PyCharm IDE, so this change will be fantastic-- thank you!

For the sake of those looking through historical commits, I'd like to ask again that the "squash" option be used on merge since 50 commits is a lot for what's ultimately a relatively small change.

njsmith:
You know, in all that, I think we forgot to add a newsfragment. We should do that so is will realize how hard we're working.

jmfrank63:
New to this one, too. It’s probably just a sentence or two, is there any
guideline or best practice?

njsmith:
@jmfrank63 For the mechanics, see: https://trio.readthedocs.io/en/latest/contributing.html#pull-request-release-notes

Basically: make a file called `newsfragments/542.bugfix.rst` (or `feature`? Idk, use your judgement :-)), and write a sentence or two to explain to users why they should be interested in upgrading to 0.8.0. For examples, you can use the previous release notes: https://trio.readthedocs.io/en/latest/history.html

jmfrank63:
Now, that the pull request is merged, how to add a newsfragment? Via an additional PR?

njsmith:
@jmfrank63 yep, exactly

PR #595, opened 2018-08-12
correction to "Spawning tasks without becoming a parent"

belm0:


belm0:
sorry, I realized that the document is likely written as intended

I do think it can still be improved by showing an example that doesn't trivially reduce.  Then there's no need to waste words defending it.

njsmith:
Yeah, that's an awkward bit for sure... the history is that before v0.3.0, it *didn't* reduce, because you weren't allowed to do arbitrary amounts of work inside a nursery block, and this really was a case where passing nurseries around was necessary. But then we fixed that, so the example stopped making much sense, and I stuck in an awkward patch instead of rethinking the section to find a new example...

I'm actually having trouble thinking of a simple case where passing nurseries around is actually useful... I can think of some exotic ones, like "implementing trio-asyncio", but that's not a great way to introduce the concept :-).

Of course if there aren't a lot of obvious use cases that users will run into, then maybe the concept isn't so important, and just a simple foo/bar-style example would be enough to give the idea.

PR #596, opened 2018-08-12
Add task blocked watchdog.

### LINKED ISSUE #591: Add a blocked task watchdog
In [this stackoverflow question](https://stackoverflow.com/questions/51680089/trio-lib-queue-get-and-put), someone gets very confused because they're using `time.sleep` and never using `await`, so one of their tasks seems to run correctly but the other doesn't. There's no indication of what's going on. Easy to fix once you know how, but if you don't, then how are you supposed to figure it out?

We could, pretty cheaply, add a watchdog to catch such situations. What I'm imagining is that when entering `trio.run`, we'd spawn a system task that spins up a thread and watches for the main loop to become stuck. In the main run loop, we'd send it messages occasionally, whenever we (a) start processing a batch of tasks, (b) stop processing a bunch of tasks. If the watchdog thread detects that the time between (a) and (b) exceeds, say, 5 seconds, then it should print an informative message and use the [`faulthandler`](https://docs.python.org/3/library/faulthandler.html) module to dump everyone's stack traces. I think this might be cheap enough it could be enabled-by-default.

Fuyukai:
Closes #591.

TODO:
  * [ ] rebase to master
  * [ ] fix deadlock (see https://github.com/python-trio/trio/pull/596#issuecomment-415270221 and https://github.com/python-trio/trio/pull/596#issuecomment-415339649)
  * [ ] improve tests: code coverage, reduce time scale
  * API review
    * [ ] `watchdog_timeout=None` instead of additional `use_watchdog` option?
    * [ ] best default?  Perhaps "max block time of a healthy Trio program x 10".
  * [ ] documentation
  * [ ] newsfragment
  * [ ] future work: include e.g. 99.9 percentile observed block time in Trio stats, to help users decide appropriate watchdog_timeout for their app

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1158,7 +1158,8 @@ def run(
     clock=None,
     instruments=(),
     restrict_keyboard_interrupt_to_checkpoints=False,
-    use_watchdog=True
+    use_watchdog=True,
+    watchdog_timeout=5
 ):
     """Run a trio-flavored async function, and return the result.
 
@@ -1220,6 +1221,9 @@ def run(
           and if so will notify you and print the stack traces of all
           threads to show exactly where the program is blocked.
 
+      watchdog_timeout (int): The number of seconds the watchdog will wait
+          before notifying that the main thread is blocked.
+
     Returns:
       Whatever ``async_fn`` returns.
 
@@ -1260,7 +1264,7 @@ def run(
     locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
 
     if use_watchdog:
-        watchdog = TrioWatchdog()
+        watchdog = TrioWatchdog(watchdog_timeout)
     else:
         watchdog = None
 
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -34,6 +34,7 @@ from ._traps import (
     CancelShieldedCheckpoint,
     WaitTaskRescheduled,
 )
+from ._watchdog import TrioWatchdog
 from .. import _core
 
 # At the bottom of this file there's also some "clever" code that generates
@@ -1156,7 +1157,8 @@ def run(
     *args,
     clock=None,
     instruments=(),
-    restrict_keyboard_interrupt_to_checkpoints=False
+    restrict_keyboard_interrupt_to_checkpoints=False,
+    use_watchdog=True
 ):
     """Run a trio-flavored async function, and return the result.
 
@@ -1213,6 +1215,11 @@ def run(
           main thread (this is a Python limitation), or if you use
           :func:`catch_signals` to catch SIGINT.
 
+      use_watchdog (bool): Enables the Trio task watchdog. This will spawn
+          a separate thread that will check if any tasks are blocked,
+          and if so will notify you and print the stack traces of all
+          threads to show exactly where the program is blocked.
+
     Returns:
       Whatever ``async_fn`` returns.
 
@@ -1252,6 +1259,11 @@ def run(
     GLOBAL_RUN_CONTEXT.runner = runner
     locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
 
+    if use_watchdog:
+        watchdog = TrioWatchdog()
+    else:
+        watchdog = None
+
     # KI handling goes outside the core try/except/finally to avoid a window
     # where KeyboardInterrupt would be allowed and converted into an
     # TrioInternalError:
@@ -1263,7 +1275,9 @@ def run(
                 with closing(runner):
                     # The main reason this is split off into its own function
                     # is just to get rid of this extra indentation.
-                    result = run_impl(runner, async_fn, args)
+                    result = run_impl(
+                        runner, async_fn, args, watchdog=watchdog
+                    )
             except TrioInternalError:
                 raise
             except BaseException as exc:
@@ -1271,6 +1285,8 @@ def run(
                     "internal error in trio - please file a bug!"
                 ) from exc
             finally:
+                if use_watchdog:
+                    watchdog.stop()
                 GLOBAL_RUN_CONTEXT.__dict__.clear()
             return result.unwrap()
     finally:
@@ -1286,7 +1302,7 @@ def run(
 _MAX_TIMEOUT = 24 * 60 * 60
 
 
-def run_impl(runner, async_fn, args):
+def run_impl(runner, async_fn, args, watchdog):
     __tracebackhide__ = True
 
     runner.instrument("before_run")
@@ -1298,6 +1314,8 @@ def run_impl(runner, async_fn, args):
         "<init>",
         system_task=True,
     )
+    if watchdog is not None:
+        watchdog.start()
 
     # You know how people talk about "event loops"? This 'while' loop right
     # here is our event loop:
@@ -1370,6 +1388,8 @@ def run_impl(runner, async_fn, args):
             task = batch.pop()
             GLOBAL_RUN_CONTEXT.task = task
             runner.instrument("before_task_step", task)
+            if watchdog is not None:
+                watchdog.notify_alive_before()
 
             next_send = task._next_send
             task._next_send = None
@@ -1387,6 +1407,9 @@ def run_impl(runner, async_fn, args):
                 final_result = Value(stop_iteration.value)
             except BaseException as task_exc:
                 final_result = Error(task_exc)
+
+            if watchdog is not None:
+                watchdog.notify_alive_after()
 
             if final_result is not None:
                 # We can't call this directly inside the except: blocks above,
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1264,6 +1264,8 @@ def run(
     else:
         watchdog = None
 
+    GLOBAL_RUN_CONTEXT.watchdog = watchdog
+
     # KI handling goes outside the core try/except/finally to avoid a window
     # where KeyboardInterrupt would be allowed and converted into an
     # TrioInternalError:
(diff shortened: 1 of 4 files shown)


njsmith:
Interesting approach! I was imagining something using a wake-up fd, but I think you're right and `threading.Event` could work too, and more simply.

This version does have a problem though: if the main loop goes to sleep for 5 seconds because there's nothing to do, then that's fine... but the watchdog still fires. We only want the watchdog to fire if we spend a long time in the task-running part of the loop; it should ignore the IO-handling part.

Of course, doing that between two threads without race conditions is not trivial... Here's my idea. Keep two counters: one that we increment whenever we start running user tasks, and one that we increment whenever we stop. So when we're in the task running part of the code, the counter values are unequal, and when we're in the io-handling part they're equal. Only the main thread mutates these counters; the watchdog thread observes them. The watchdog should fire if it ever makes two observations that are at least X seconds apart, where the values haven't changed, and are unequal to each other. That combination means that we're stuck in a user task.

Since python's memory model is pretty strongly sequential and consistent, and assignment is atomic, I think we can skip any locks on these counters, which is nice.

With this setup, a correct-but-slightly-inefficient implementation would be:

```python
while True:
    orig_starts = self._starts
    orig_stops = self._stops
    time.sleep(X)
    if orig_starts = self._starts and orig_stops = self._stops and orig_starts != orig_stops:
        # watchdog fires
```

The downside to this implementation is that it wakes up every X seconds even if the main thread is sleeping, which isn't a huge flaw but it does waste some power and besides, it's just not elegant. Can we do better?

```python
while True:
    self._event.clear()
    orig_starts = self._starts
    orig_stops = self._stops
    if orig_starts == orig_stops:
        # main thread asleep; nothing to do until it wakes up
        self._event.wait()
    else:
        self._event.wait(timeout=5)
        if orig_starts = self._starts and orig_stops = self._stops:
            # watchdog fires
```

And I think whenever we start running tasks we do `self._started += 1; self._event.set()` (in that order), and when we stop running tasks we can skip the event and just do `self._stopped += 1`.

Maybe this would be simpler if we used a `Condition` directly? `Event` is a thin shim around it, and not adding much value here...

Fuyukai:
Made the changes so now it should work even when there's nothing to do (basically copy-pasted your code...)

trio/_core/_watchdog.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _print_all_threads(self):
+        # separated for indent reasons, damned 80 char limit
+        for thread in threading.enumerate():
+            print(
+                "Thread {} (most recent call last):".format(thread.name),
+                file=sys.stderr
+            )
+            # scary internal function!
+            traceback.print_stack(sys._current_frames()[thread.ident])
```
njsmith: What's the advantage of doing this by hand instead of using faulthandler?

(This seems like a pretty delicate and race-prone operation, so I'm nervous about trying to get it right ourselves. E.g., what happens if a thread exits while `print_stack` is walking its stack?)

trio/_core/_watchdog.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def start(self):
+        self._thread = threading.Thread(
+            target=self._main_loop, name="<trio watchdog>", daemon=True
+        )
+        self._thread.start()
+
+    def stop(self):
+        self._stopped = True
```
njsmith: This should also set the event, so that the thread wakes up promptly and notices that `self._stopped` has been set.

trio/_core/tests/test_watchdog.py
```diff
@@ -0,0 +1,30 @@
...
+    watchdog = _run.GLOBAL_RUN_CONTEXT.watchdog
+    target = StringIO()
+    with contextlib.redirect_stderr(target):
+        time.sleep(7)
+
+    assert target.getvalue().startswith(
+        "Trio Watchdog has not received any "
+        "notifications in 5 seconds, main "
+        "thread is blocked!"
+    )
```
njsmith: Maybe assert that `time.sleep` and `test_watchdog` show up in the output too, to make sure that the watchdog is printing tracebacks?

trio/_core/tests/test_watchdog.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # checkpoint to ensure the watchdog gets a notification, sees that its
+    # dead, and winds down its thread
+    watchdog.stop()
+    await _run.checkpoint()
+    time.sleep(6)  # ensure if the watchdog is waiting for 5s, it wakes
+    await _run.checkpoint()
+    assert not watchdog._thread.is_alive()
```
njsmith: This looks very delicate... calling `self._event.set()` from `Watchdog.stop()` will help, but even so...

Here's another idea: make the watchdog thread *not* a daemon, and then join it in `Watchdog.stop`. That way every time we call `trio.run`, we're implicitly checking the watchdog shutdown logic (and also making sure we don't leave stray background threads running, even temporarily). What do you think?

trio/_core/tests/test_watchdog.py
```diff
@@ -0,0 +1,30 @@
...
+from .. import _run
+
+
+@slow
+async def test_watchdog():
+    watchdog = _run.GLOBAL_RUN_CONTEXT.watchdog
+    target = StringIO()
+    with contextlib.redirect_stderr(target):
+        time.sleep(7)
+
```
njsmith: Maybe it would also be good to have a test where we `await trio.sleep(7)` and make sure the watchdog doesn't fire? (I.e., a test to catch the bug that was in the first draft of this code.)

Also, these sleeps are kind of long, even for a `@slow` test... maybe it would be nicer to make the argument `watchdog_timeout=5`, so then in tests we can adjust it? (And we get rid of the unconfigurable magic constant.)

trio/_core/_run.py
```diff
@@ -1254,6 +1261,13 @@ def run(
     GLOBAL_RUN_CONTEXT.runner = runner
     locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
 
+    if use_watchdog:
+        watchdog = TrioWatchdog()
+    else:
+        watchdog = None
+
+    GLOBAL_RUN_CONTEXT.watchdog = watchdog
```
njsmith: It doesn't really matter, but... any reason in particular this isn't an attribute on `runner`?

trio/_core/_watchdog.py
Fuyukai: It prints the code being ran rather than an unhelpful list of lines (backwards, too).

trio/_core/_run.py
Fuyukai: No paticular reason - just where I put it at first.

Fuyukai:
For some reason, only on PyPy nightly, joining the thread blocks forever randomly.

njsmith:
> For some reason, only on PyPy nightly, joining the thread blocks forever randomly.

So the traceback shows the watchdog thread is blocked in `event.wait`, and it's blocked for a minute, so probably a call without a timeout. Probably the one on line 42 of `_watchdog.py`:

https://github.com/python-trio/trio/blob/753753be65de75c8dc1369d3738ed01045fbaa8d/trio/_core/_watchdog.py#L42

And the main thread is stuck in `join`, called from inside our `stop` method.

https://github.com/python-trio/trio/blob/753753be65de75c8dc1369d3738ed01045fbaa8d/trio/_core/_watchdog.py#L81-L84

That's weird, it looks like our `stop` method sets the event, so how could the watchdog be blocked waiting for it? Assuming `Event` actually works, the only way this could happen is if someone calls `clear`...

https://github.com/python-trio/trio/blob/753753be65de75c8dc1369d3738ed01045fbaa8d/trio/_core/_watchdog.py#L34-L42

Ah crud, the watchdog thread calls `clear` just before it goes to sleep. So if we happen to call `stop` *just* in that interval between the `if self._stopped` and `self._notify_event.clear`, then we'll deadlock. I'm not 100% sure that this is the problem we're seeing on pypy, but it's certainly *a* problem in the algorithm I suggested, and could explain it.

Man I hate threads.

I suspect that instead of messing around with `Event` like this, we want to switch to using a `Condition` (like how `threading.Event` is implemented, if you look), so that we can atomically check the stopped flag and counter values before sleeping.

njsmith:
To elaborate a bit more on how a `Condition` could help here: a `Condition` always has an attached `Lock`. And the basic pattern of using a `Condition` is: one thread acquires the lock, checks some stuff, and then blocks in `Condition.wait` – which temporarily releases the lock. Another thread acquires the lock, mutates some stuff, and then calls `Condition.notify` – which can only be done with the lock held. So for example, we might write:

```python
# initialize
self._cond = threading.Condition()

# going to sleep
with self._cond:
    if self._stopped:
        return
    self._cond.wait()

# waking up
with self._cond:
    self._stopped = True
    self._cond.notify()
```

This is the same pattern as our current code, *except* that before we had a problem because the wake-up thread could run in between when the sleeping thread checks `self._stopped`, and when it calls `wait` to go to sleep. Now, the sleeping thread holds the lock for this period, so the wake-up thread can't run during this time. So that race condition is solved. Does that make sense?

My intuition says that we can extend this pretty straightforwardly to the slightly more complex full watchdog code, but right now I'm too sleepy to figure it out, so maybe give it a try and see how far you get?

belm0:
What's left to do here?  Is this something I can pick up?

Fuyukai:
I haven't touched this in a week or so (been more focused on the subprocess support). Afaik the only thing left here is to prevent that deadlock (using a Condition).

njsmith:
Closing stale PR... we can always re-open if someone wants to pick it up again, but if anyone is looking to pick this up again I'd probably start with the new idea described in this comment: https://github.com/python-trio/trio/issues/591#issuecomment-463130577

Issue #469, opened 2018-03-16
raven doesn't like LOCALS_KEY_KI_PROTECTION_ENABLED


lilydjwg:
```py
import trio

async def main():
  from raven import Client
  client = Client(SENTRY_DSN)
  raise Exception

if __name__ == '__main__':
  trio.run(main)
```

The cause is:

```pybt
Error in sys.excepthook:
Traceback (most recent call last):
  File ".../venv/lib/python3.6/site-packages/raven/utils/json.py", line 48, in dumps
    return json.dumps(value, cls=BetterJSONEncoder, **kwargs)
  File "/usr/lib/python3.6/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
TypeError: keys must be a string
```

It's because `trio._core._ki.LOCALS_KEY_KI_PROTECTION_ENABLED` is used as a key and raven collects it.

njsmith:
Whoops, I guess I [won my bet](https://github.com/python-trio/trio/blob/aee2676fd5096daf2849db9756242a30e0aa1f30/trio/_core/_ki.py#L81-L83).

I'm not sure if this should be considered a bug in trio or in raven – we can work around it in trio by using a string key, but raven should perhaps be robust against this anyway :-).

You're going to run into some other issues with raven too, I think:

* Make sure to `import trio` before you `import raven`, because it looks like they both mess with `sys.excepthook`. If you `import raven` first, then it'll take over `sys.excepthook`, and then trio will notice that someone has taken it over and skip installing its hook, which means you won't get good tracebacks for `MultiError`s. If you `import trio` first, then raven will notice this and delegate to trio's hook after running its own, so that works better. (Unfortunately trio can't do the same thing, because it wants to *replace* the usual hook, while raven wants to *augment* the usual hook.)

* If raven is trying to capture frame information from unhandled exceptions, then it's not going to return very useful results for `MultiError` exceptions, because it won't know to peek inside and find the inner exception traceback frames.

It looks like raven has some [precedent for shipping integration code for specific libraries](https://docs.sentry.io/clients/python/integrations/), but I haven't looked any closer to see if its extension points are powerful enough to fix these issues...

I guess as a first step I'll file a bug on raven and see what they think.

[...]

PR #599, opened 2018-08-13
Fix crash when used with Raven.

miracle2k:
See #469.


njsmith:
Looks like you need to run yapf.

Can you throw in an entry in `newsfragments/` too?

miracle2k:
CI pass!

Issue #220, opened 2017-06-15
Better tooling for releases and development


njsmith:
Let's have an issue for collecting discussion about improving our tooling for release automation and similar.

# Release notes

I think it would be nice to move to [towncrier](https://github.com/hawkowl/towncrier), and possibly adopt the twisted rule that non-trivial PRs should always close an issue. (The idea is that the issue records what the problem is and what was considered for fixing it, while PRs are used for discussing a particular candidate patch.) Basically the way this works is that whenever you submit a PR, you make a new file in a special directory that contains a description of that change. This way you can generate your user-oriented release notes as you go, in a standard format, without hitting merge conflicts. (Then you also need the rule that PRs have to actually contain one of these files, possibly enforced by a linting bot.)

Alternatives include openstack's reno which seems substantially more complicated (it involves yaml and a special tool that contributors need to use to generate the file, etc.). CPython had a [long discussion](https://github.com/python/core-workflow/issues/66) about this and decided to go with their own tool that they can tweak to match the peculiarities of their workflow, but it's a useful survey (if very long).

# Blue sky

The Rust community has this great bot that does their merging called "homu" (or sometimes "bors", but the modern implementation is "homu"). The way it works is that when a change has passed review, the reviewer tells the bot to go ahead and merge it, and then the bot *makes sure the tests pass on current master* before merging. (With github status checks, the normal thing is: (1) PR submitted, (2) status checks test that it works against current master, (3) other stuff gets merged into master, (4) PR gets merged into master. So the actual commits on master are generally not tested until *after* they get merged, which is kind of absurd if you think about it.)

Unfortunately homu itself doesn't currently support our setup with appveyor+travis. It looks like this is the PR for that: https://github.com/barosl/homu/pull/134 (sitting unmerged for ~16 months now, woo. maybe the servo/homu fork is the real one now?)

(I'd also be nervous about using the hosted version of homu while [this bug](https://github.com/barosl/homu/issues/155) remains unfixed.)

What I'd really love is a bot that works like homu, but whose sequence goes like:

* Reviewer says "let's do it"
* Bot checks out the PR
* Bot automatically fixes formatting (probably by running [yapf](https://github.com/google/yapf) over everything)
* Bot tests the reformatted PR
* If tests pass, bot merges reformatted PR into mainline
* If some appropriate condition is met, bot tags this commit as a release and uploads to PyPI.

[...]

PR #601, opened 2018-08-13
Release 0.6.0

smurfix:


smurfix:
598 is already part of his.

pquentin:
@smurfix That was my point, we included a commit in the release that has not been approved yet

pquentin:
I guess the main benefit is that you can easily see on GitHub all the commits that will make it into the release. I think this workflow is pretty standard in the corporate world and is named Gitflow, but that does not mean we have to adopt it

@smurfix thoughts?

pquentin:
(It's also surprising that trio was released even though this pull request has not been merged yet)

njsmith:
Huh, I forgot that gitflow actually does this.

Reading through some of the posts about it, I think their release branch stuff is designed to support a slow-release workflow, where the mainline development branch isn't always releasable, releases are rare, and each release involves some stabilization ceremony. And also maybe your git repo is your release distribution mechanism, rather than using something like PyPI. That's a valid approach, but I'd rather move more towards a quick-release low-ceremony model (like we discussed some in #220), and the canonical way to get our latest release is to check PyPI, so a `release` branch is redundant (violates DRY). So... I don't think having a release branch provides much value here.

pquentin:
Sounds good 👍 

So that means we should close this PR and delete the release branch?

njsmith:
This has been sitting for a while with no followups, so I guess I'm going to go ahead and do what I said!

Issue #400, opened 2018-01-13
Portable way of waiting for sockets when all you have is a socket descriptor, not a socket object


Fuyukai:
I'm writing an async wrapper over psycopg2 for my ORM to work with trio, and I've hit a snag with it - as far as I can see in the docs, there's no universal way of waiting until a file descriptor is readable.

There's `trio.hazmat.wait_readable`, but that only works on *nix, and `wait_socket_readable` (as shown in the docs) only works on `socket.socket` objects. 

I took a quick search of the issues and couldn't find anything related to this either.

njsmith:
Yes, that's right. At that level of the API trio is just focused on
exposing the raw OS capabilities, and "wait for arbitrary file descriptor"
is a thing that Unix supports and Windows doesn't. (And even on Unix there
are some funny limitations. Most notably: files on disk will claim to
always be readable/writable.) I wish it were otherwise, but that's
something we'd have to take up with Microsoft about 20 years ago.

What are you trying to do? Maybe there's some alternative?

On Jan 13, 2018 06:29, "Laura F. D." <notifications@github.com> wrote:

I'm writing an async wrapper over psycopg2 for my ORM to work with trio,
and I've hit a snag with it - as far as I can see in the docs, there's no
universal way of waiting until a file descriptor is readable.

There's trio.hazmat.wait_readable, but that only works on *nix, and
wait_socket_readable (as shown in the docs) only works on socket.socket
objects.

I took a quick search of the issues and couldn't find anything related to
this either.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/python-trio/trio/issues/400>, or mute the thread

SunDwarf:
The asynchronous support for psycopg2 requires [selecting on the file descriptors](http://initd.org/psycopg/docs/advanced.html#asynchronous-support), which requires the ability to select on the connection objects returned.

njsmith:
Oh, I see – so it's only sockets that you want to wait on, but psycopg2 is giving handing you the raw socket descriptors, while `trio.hazmat.wait_socket_*` expect Python socket objects.

You can convert psycopg2's socket descriptors into Python socket objects by using [`socket.fromfd`](https://docs.python.org/3/library/socket.html#socket.fromfd). So something like:

```python
class TrioPgConn:
    def __init__(self, ...):
        self._conn = psycopg2.connect(..., async=True)
        self._sock = socket.fromfd(self._conn.fileno())

[...]

PR #610, opened 2018-08-18
wait_socket_*, notify_socket_close now accept raw socket descriptors

njsmith:
Fixes gh-400

CC @Fuyukai

### DIFF
--- a/trio/_core/__init__.py
+++ b/trio/_core/__init__.py
@@ -44,23 +44,9 @@ from ._local import *
 __all__ += _local.__all__
 
 if hasattr(_run, "wait_readable"):
-    import socket as _stdlib_socket
-
-    async def wait_socket_readable(sock):
-        if type(sock) != _stdlib_socket.socket:
-            raise TypeError("need a socket")
-        await wait_readable(sock)
-
-    async def wait_socket_writable(sock):
-        if type(sock) != _stdlib_socket.socket:
-            raise TypeError("need a socket")
-        await wait_writable(sock)
-
-    def notify_socket_close(sock):
-        if type(sock) != _stdlib_socket.socket:
-            raise TypeError("need a socket")
-        notify_fd_close(sock)
-
+    wait_socket_readable = wait_readable
+    wait_socket_writable = wait_writable
+    notify_socket_close = notify_fd_close
     __all__ += [
         "wait_socket_readable", "wait_socket_writable", "notify_socket_close"
     ]
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -129,11 +129,11 @@ All environments provide the following functions:
 
    Block until the given :func:`socket.socket` object is readable.
 
-   The given object *must* be exactly of type :func:`socket.socket`,
-   nothing else.
+   On Unix systems, sockets are fds, and this is identical to
+   :func:`wait_readable`. On Windows, ``SOCKET`` handles and fds are
+   different, and this works on ``SOCKET`` handles or Python socket
+   objects.
 
-   :raises TypeError:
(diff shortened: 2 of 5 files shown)


trio/_core/__init__.py
```diff
@@ -44,23 +44,9 @@ def _public(fn):
...
-        if type(sock) != _stdlib_socket.socket:
-            raise TypeError("need a socket")
-        await wait_writable(sock)
-
-    def notify_socket_close(sock):
-        if type(sock) != _stdlib_socket.socket:
-            raise TypeError("need a socket")
-        notify_fd_close(sock)
-
+    wait_socket_readable = wait_readable
```
Fuyukai: Is there any reason for these to be under a different name now?

njsmith:
Whoops, just realized there's a bug here: on Windows, we store and look up sockets by Python object id, not by fileno, so if one task does `wait_socket_readable(sockobj)` and another does `notify_socket_closed(sockobj.fileno())`, trio won't realize that those are the same socket. We should be like the Unix versions of these functions, and always use the `.fileno()` as the lookup key.

trio/_core/__init__.py
njsmith: excellent question :-)

The reason I did it this way in the first place is that on Windows, there are two totally different things: handles, and file descriptors. They're both integers that refer to OS objects via some entry in a process-local table, but it's two separate tables, so you have to know which kind of integer you have :-). Because of Python's unix heritage, it's sloppy about this distinction – e.g. in the stdlib, `file.fileno()` returns a file descriptor, but `socket.fileno()` returns a handle.

Python gets away with this because each operation either interprets arguments as fds (`os.read`), or as handles (`select.select`). And in Windows there is no wait to wait for an arbitrary fd to be readable or writable, or even an arbitrary handle to be readable or writable; it's only defined for socket handles.

So... my default for the low-level APIs is to expose the system's capabilities, whatever those are, as directly as possible. And on Unix there's a generic concept of "wait for a thing to be readable/writable", so I defined `wait_readable` and `wait_writable` functions, and on Windows there's a socket-specific concept of "wait for socket to be readable/writable", so I defined `wait_socket_readable` and `wait_socket_writable`. And then when I implemented `trio.socket`, I was like hey it's very annoying to have to call one function on Window and a different one on Unix, given that when you're working on sockets they're exactly the same. So I added the `*_socket_*` functions to Unix too, as a tiny little platform abstraction layer.

Another approach would be to follow Python's lead, and say eh, no-one's ever going to want to wait for readability/writability on an fd on Windows because Windows just doesn't do that, or notify that an fd is closed, so let's pretend that Windows handles are the same as Unix fds and that Windows fds don't exist, and have a single set of functions `wait_readable`, `wait_writable`, `notify_closed`. (And document that on Windows, they only work on sockets.) I'm not super excited about this though because (1) it feels inelegant, (2) there are ways to check if other objects are readable on Windows that wouldn't use `wait_readable`, e.g. you can check whether the console is readable by doing `WaitForSingleObject` on a special handle.

At some point it might also make sense at some point to have `notify_handle_closed` on Windows, that wakes up `WaitForSingleObject` as well as `wait_socket_*`? And then make `notify_socket_closed` an alias for it?

So... you're right that the difference is smaller now. Having written all this out, I guess I'm still vaguely slightly in favor of having the different functions, and if we do decide to deprecate the `*_socket_*` functions there's no particular reason to do it in this PR, so in conclusion I guess I'll leave this PR as it is for now.

njsmith:
Oh, and that bug is why the appveyor tests aren't running... it's causing the windows tests to freeze, so every appveyor job takes an hour to timeout, so I've totally killed the appveyor queue. Whoops.

njsmith:
Okay, fixed that bug and tests are green.

PR #617, opened 2018-08-21
trio.Event docs

belm0:
Make it more clear that trio.Event is level-triggered, and give a hint at how to make it behave as edge-triggered.

### DIFF
--- a/trio/_sync.py
+++ b/trio/_sync.py
@@ -23,15 +23,18 @@ class Event:
     """A waitable boolean value useful for inter-task synchronization,
     inspired by :class:`threading.Event`.
 
-    An event object manages an internal boolean flag, which is initially
-    False, and tasks can wait for it to become True.
-    
-    The flag is useful because it helps avoid race conditions and lost
-    wakeups: it doesn't matter whether :meth:`set` gets called just
-    before or after :meth:`wait`.
-    
-    If you want a lower-level wakeup primitive that doesn't have this
-    protection, consider :class`Condition` or :class:`trio.hazmat.ParkingLot`.
+    An event object with an internal boolean flag, representing whether
+    the event has happened yet. The flag is initially False, and the
+    :meth:wait method waits until the flag is True. If the flag is
+    already True, then :meth:wait returns immediately. (If the event has
+    already happened, there's nothing to wait for.) The :meth:set method
+    sets the flag to True, and wakes up any waiters.
+
+    This behavior is useful because it helps avoid race conditions and
+    lost wakeups: it doesn't matter whether :meth:set gets called just
+    before or after :meth:wait. If you want a lower-level wakeup
+    primitive that doesn't have this protection, consider :classCondition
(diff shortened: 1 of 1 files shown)


belm0:
I'm OK dropping the newly added sentence if it doesn't seem like a pattern we want to promote.

pquentin:
As a non-native speaker, I don't understand "though still a checkpoint" in "If it's already True, then this method returns immediately (though still a checkpoint).". Is a word missing?

belm0:
I can add "it's":
> If it's already True, then this method returns immediately (though *it's* still a checkpoint)

however I was trying to avoid duplication of "it's".

I think the grammar is OK as written, since it's just a shuffling of:
> If it's already True, then this method-- though still a checkpoint-- returns immediately.

njsmith:
@belm0 I just reworked this a bit... what do you think of my approach?

belm0:
@njsmith removing "then this method is still a checkpoint" is great-- this is what tripped up several casual trio-using colleagues because they don't know what checkpoints are.

The top level wording is a little strange.  I'd like to see it really explicit about the level-trigger behavior (even if we don't use that term).  Having an internal flag doesn't imply level-trigger, which is one way the proposed wording throws me off.  How about:

> An event object with an internal boolean flag, which is initially False, and tasks can wait for it to become True.  Attempts to wait when it's already True will return immediately.
>    
> This behavior is useful because it helps avoid race conditions and lost wakeups: it doesn't matter whether :meth:`set` gets called just before or after :meth:`wait`.
>
> If you want a lower-level wakeup primitive that doesn't have this protection, consider :class`Condition` or :class:`trio.hazmat.ParkingLot`.

njsmith:
Oh I see, so it's specifically the "wait for it to be *become* true" part that threw them...

Here's another attempt (based on yours above, plus some inspiration [from the stdlib docs](https://docs.python.org/3/library/threading.html#threading.Event)):

> An event object with an internal boolean flag, representing whether the event has happened yet. The flag is initially False, and the :meth:`wait` method waits until the flag is True. If the flag is already True, then :meth:`wait` returns immediately. (If the event has already happened, there's nothing to wait for.) The :meth:`set` method sets the flag to True, and wakes up any waiters.
>    
> This behavior is useful because it helps avoid race conditions and lost wakeups: it doesn't matter whether :meth:`set` gets called just before or after :meth:`wait`. If you want a lower-level wakeup primitive that doesn't have this protection, consider :class`Condition` or :class:`trio.hazmat.ParkingLot`.

belm0:
@njsmith I've updated to your proposed description and will merge it-- thank you.

njsmith:
@belm0 It looks like at some point in the copy-pasting, the ReST syntax got mangled and all the `:meth:` stuff is rendering wrong: https://trio.readthedocs.io/en/latest/reference-core.html#trio.Event

Also, this is a nitpick, but it reads weird to me to have *two* sentences with no verb at the beginning. Maybe "An event object *has* an internal"?

Issue #354, opened 2017-11-29
Remove batching from `catch_signals`?


njsmith:
I was just looking at [`catch_signals`](https://trio.readthedocs.io/en/latest/reference-io.html#trio.catch_signals), and it occurred to me maybe we can improve on its funky batching API.

The original motivation was [the same as for `UnboundedQueue`](https://trio.readthedocs.io/en/latest/reference-hazmat.html#unbounded-queues): since we have no way to put backpressure onto signals as they arrive, we nudge users to handle a whole batch at once.

One alternative I considered were to use a regular `UnboundedQueue`, which has the problem that its memory use actually be unbounded, which feels wrong given that signals are coalescable and thus actually can be queued with bounded memory. The other was to store incoming signals in a `set`, and then pop an element each time. The problem with this is that it can lead to a kind of starvation where signals of type A keep coming in and getting popped, while a signal of type B sits in the set indefinitely without being reported. (For example, a policy that would guarantee this worst-case behavior would be to always report the lowest-number pending signal.)

What I realized now though is that there's another option: we could use the `set`-type approach, but with a policy that's guaranteed to cycle through all the different signals in bounded time. For example, if on each iteration we report the signal with the lowest number *that is strictly larger than the one we reported last time* (wrapping around as necessary), then we'd guarantee that any incoming signal was reported within 64 event loop iterations at worst (at least on Linux which has 64 signals, but other OSes are similar). (And of course in practice it would be much, much quicker, since you never receive all 64 signals at once.)

Probably the nicest way to accomplish this would be to use an `OrderedDict` for queueing incoming signals, like we do for `TrioToken.run_sync_soon(..., idempotent=True)` calls.

There's some transition issue here, since this change would be backwards incompatible in a way that we can't warn for. We can either come up with a new name for `catch_signals` and transition to using that instead, or else rip off the bandaid by just changing it.

njsmith:
Regarding the transition: Maybe we could take this as an opportunity to standardize this on an `open_*` name, like we use for everything else.

PR #619, opened 2018-08-22
New signal API: trio.open_signal_receiver

njsmith:
Fixes gh-354

Other changes:

- deprecate trio.catch_signal
- fix a few small edge-cases I noticed along the way

### DIFF
--- a/trio/_signals.py
+++ b/trio/_signals.py
@@ -1,11 +1,15 @@
 import signal
 from contextlib import contextmanager
+from collections import OrderedDict
 
 from . import _core
-from ._sync import Semaphore
-from ._util import signal_raise, aiter_compat, is_main_thread
+from ._sync import Event
+from ._util import (
+    signal_raise, aiter_compat, is_main_thread, ConflictDetector
+)
+from ._deprecate import deprecated
 
-__all__ = ["catch_signals"]
+__all__ = ["open_signal_receiver", "catch_signals"]
 
 # Discussion of signal handling strategies:
 #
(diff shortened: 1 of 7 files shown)


trio/tests/test_signals.py
```diff
@@ -35,23 +34,98 @@
...
+    orig = signal.getsignal(signal.SIGILL)
+    with open_signal_receiver(signal.SIGILL) as receiver:
+        # Raise it a few times, to exercise signal coalescing, both at the
+        # call_soon level and at the SignalQueue level
+        signal_raise(signal.SIGILL)
+        signal_raise(signal.SIGILL)
+        await _core.wait_all_tasks_blocked()
+        signal_raise(signal.SIGILL)
+        await _core.wait_all_tasks_blocked()
+        async for signum in receiver:  # pragma: no branch
```
tgoodlet: Is it worth ensuring that `receiver` only *once* delivers a signal raised multiple times prior to iteration?
I guess I'm just thinking down the road if the implementation is changed (internal `set`-like data structure or whatever) and this behaviour should remain.

trio/tests/test_signals.py
```diff
@@ -60,50 +134,44 @@ def test_catch_signals_wrong_thread():
     def direct_handler(signo, frame):
         delivered_directly.add(signo)
 
-    async def wait_call_soon_idempotent_queue_barrier():
-        ev = Event()
-        token = _core.current_trio_token()
-        token.run_sync_soon(ev.set, idempotent=True)
-        await ev.wait()
-
     print(1)
```
tgoodlet: Maybe print something more descriptive?

tgoodlet: Hmm, actually, have you considered just adding some branch logic for each 
`await wait_run_sync_soon_idempotent_queue_barrier()` location and then just use `pytest.mark.parametrize()` over signals and that?

Might reduce some code unless the duplication is intentional.

trio/_signals.py
```diff
@@ -1,11 +1,15 @@
 import signal
 from contextlib import contextmanager
+from collections import OrderedDict
```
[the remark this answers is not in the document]
njsmith: PyPy has actually had ordered `dict`s by default for a while (CPython copied their implementation). But annoyingly, we also need `popitem(last=False)`, which even in 3.7 is still an `OrderedDict`-only method.

trio/tests/test_signals.py
```diff
@@ -1,17 +1,16 @@
...
+import trio
 from .. import _core
 from .._util import signal_raise
-from .._signals import catch_signals, _signal_handler
-from .._sync import Event
+from .._signals import catch_signals, open_signal_receiver, _signal_handler
 
 
-async def test_catch_signals():
+# Delete when catch_signals is removed
```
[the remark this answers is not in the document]
njsmith: Eh, when we remove `catch_signals` the test will start failing, so we won't miss it :-)

trio/_signals.py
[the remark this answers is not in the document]
tgoodlet: >which even in 3.7 is still an OrderedDict-only method.

Ahh what!?

trio/tests/test_signals.py
njsmith: We are implicitly checking that signals are coalesced, because if they weren't then when we exited the `with` block they'd be redelivered and cause the test process to dump core :-).

It would make sense to test this a little more explicitly... I'm going to add a little private helper to do that.

trio/tests/test_signals.py
njsmith: This is a pretty complicated test that I didn't really touch here, so honestly I'm not sure what it does right now :-). I doubt it's worth messing with – if it breaks we can easily add better print statements or whatever then.

njsmith:
Added a more explicit test of signal coalescing. All comments addressed.

trio/tests/test_signals.py
tgoodlet: Ahh right of course. Cool, I like explicit :+1:

pquentin:
Merging since @tgoodlet approved this

PR #621, opened 2018-08-22
Add support for pipes.

Fuyukai:
This is Step 4 of https://github.com/python-trio/trio/issues/4#issuecomment-398967572, and adds support for reading/writing from os.pipe objects to Trio.


njsmith:
Highlevel comment on both this and #622: this functionality is great, and obviously an important step towards #4. One question we need to answer along the way is how we're going to expose it – eventually it'll be part of some full-fledged subprocess spawning API, but should we expose the individual pieces, and if so, how? For `WaitForSingleObject`, this was pretty easy: it's an OS-specific core primitive, and the proper API was obvious, so, make it public in `trio.hazmat`, done. For these, it's a little more complicated: these particular implementations are specific to Unix/Linux, but the concepts are more general; and, I think we'll probably want to fine-tune the API as we go. (For example, with pipes we need to create a pipe where one end is wrapped in trio and non-blocking, and the other is in blocking mode to be passed to the child; and with `waitpid`, I suspect we might adjust how we track the `WaitpidResult` once we have `Popen` objects that could hold them – like maybe we'll kick off the `waitpid` call immediately when the process is created, and stash the `WaitpidResult` in the `Popen` object?)

So, my suggestion for right now is:

* Go ahead and add the implementations, maybe even in a subpackage like `trio/_subprocess/...`
* Go ahead and write the tests
* Don't export them publically
* Don't write newsfragments (because there's nothing for users yet)

That way we can incrementally write the different pieces, review them, test them, etc., and eventually we'll figure out how to hook them together into a coherent subprocess API.

Does that make sense?

njsmith:
Two more quick comments before I go to bed:

* We're going to need `__del__` methods here... most of trio doesn't bother, because e.g. `SocketStream` holds a `trio.socket.SocketType`, which holds a `socket.socket`... and `socket.socket` already has a `__del__` that makes sure the OS-level socket object is eventually closed. `os.pipe` doesn't provide that kind of wrapper, so we have to take care of `__del__` ourselves.

* Check out [`trio.testing.test_one_way_stream`](https://trio.readthedocs.io/en/latest/reference-testing.html#testing-custom-stream-implementations)

Fuyukai:
Re: check_one_way_stream - this can't be used, because it does a check that closing side W of a pipe ensures side R cancels reading (which, well, it doesn't, because one half of the fds are closed on a fork-with-exec).

trio/_highlevel_pipes.py
```diff
@@ -0,0 +1,63 @@
...
+
+class _PipeMixin:
+    def __init__(self, pipefd: int):
+        if not isinstance(pipefd, int):
+            raise TypeError("PipeSendStream needs a pipe fd")
+
+        self._pipe = pipefd
+
+    async def aclose(self):
+        os.close(self._pipe)
```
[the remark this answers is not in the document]
njsmith: The -1 thing is also baked into the `_translate_socket_errors_to_stream_errors`... on Unix, it's whole job is to convert EBADF into a `ClosedStreamError` (because EBADF is what you get when you try to use -1 as an fd), and convert everything else into `BrokenStreamError`.

It might make sense to write a `_translate_pipe_errors_to_stream_errors`, even if it's very similar to the socket version, just because both error-converters are so tightly coupled to internal implementation decisions that it's nice to not have to think about code in other files when looking at them.

trio/_highlevel_pipes.py
```diff
@@ -0,0 +1,63 @@
...
+        length = len(data)
+        # adapted from the SocketStream code
+        with memoryview(data) as view, \
+                _translate_socket_errors_to_stream_errors():
+            total_sent = 0
+            while total_sent < length:
+                with view[total_sent:] as remaining:
+                    total_sent += os.write(self._pipe, remaining)
+
+                await self.wait_send_all_might_not_block()
```
njsmith: This is going to crash if the pipe buffer fills up – `os.write` will raise `BlockingIOError` (EWOULDBLOCK), and we don't have anything to catch it. In `SocketStream.send_all`, it has a slightly easier job, because it doesn't call the raw OS `send` function directly – it calls the `trio.socket` `send` method, and that takes care or retrying and waiting. See `_nonblocking_helper` and `_try_sync` in `trio.socket` for the gory details... not that they're perfect, they're probably a bit over-elaborate themselves.

trio/_highlevel_pipes.py
```diff
@@ -0,0 +1,63 @@
...
+
+    async def wait_send_all_might_not_block(self) -> None:
+        await _core.wait_socket_writable(self._pipe)
```
njsmith: We shouldn't be using `socket` functions on non-socket fds, that's just confusing :-). Of course we can get away with it on Unix, and this is Unix-only code, so it's more a style thing. But if this code were accidentally run on Windows, then calling the `socket` functions would invoke undefined behavior, while using the fd versions would give a clean error.

trio/_highlevel_pipes.py
```diff
@@ -0,0 +1,63 @@
...
+            await _core.wait_socket_readable(self._pipe)
+            data = os.read(self._pipe, max_bytes)
+
+        return data
+
+
+def make_pipe() -> Tuple[PipeReceiveStream, PipeSendStream]:
+    """Makes a new pair of pipes."""
+    (r, w) = os.pipe2(os.O_NONBLOCK)
+    return PipeReceiveStream(r), PipeSendStream(w)
```
njsmith: I suspect it'll end up being a bit nicer if we make the stream `__init__` methods accept arbitrary pipes, and set them to non-blocking mode at that point. Then when we only want to use one side of a pipe with trio, we can write

```python
r_raw, w_raw = os.pipe()
r_trio = PipeReceiveStream(r_raw)
# ... now use r_trio and w_raw ...
```

Also, elsewhere in trio where we return a pair like this, we do `(sender, receiver)`. Which yeah is the opposite of `os.pipe`. But trio's way is easier to remember ("data flows from left to right, like you read").

Neither of these are big issues right now since we're keeping this private, and once we start trying to use it for real we'll probably find other things we want to tweak. But figured I'd mention.

trio/_highlevel_pipes.py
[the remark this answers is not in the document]
njsmith: Oh, and we should be calling `notify_fd_close` here too.

njsmith:
> Re: check_one_way_stream - this can't be used, because it does a check that closing side W of a pipe ensures side R cancels reading (which, well, it doesn't, because one half of the fds are closed on a fork-with-exec).

Not sure I follow... we wouldn't be doing fork/exec in the middle of a call to `check_one_way_stream`, right? Obviously when we actually use the pipe streams it'll be in a more complicated situation involving subprocesses, but that doesn't mean we can't use `check_one_way_stream` in a simple in-process case to validate that the stream code works and handles the edge cases correctly.

Fuyukai:
The check_one_way_stream runs a test that ensures that closing the S side of the stream stops the R side from working - with pipes, this can't happen due to child processes and what-not.

njsmith:
> The check_one_way_stream runs a test that ensures that closing the S side of the stream stops the R side from working - with pipes, this can't happen due to child processes and what-not.

This isn't a problem. It would be a problem if we stopped in the middle of a test to spawn a child process and shared our pipe object with us, but we won't do that :-). If you're just using a pipe object within a process, then it follows the usual rule where closing the send side causes the receive side to get an EOF:

```python
In [1]: import os

In [2]: R, S = os.pipe()

In [3]: os.write(S, b"x")
Out[3]: 1

In [4]: os.close(S)

In [5]: os.read(R, 10)
Out[5]: b'x'

In [6]: os.read(R, 10)
Out[6]: b''
```

trio/_subprocess/tests/conftest.py
```diff
@@ -0,0 +1,25 @@
+import pytest
```
njsmith: Instead of making another copy of this, I'm thinking we should put the subprocess tests into `trio/tests/subprocess/`? Or `_subprocess/` or `test_subprocess/`, I guess, whichever looks nicest. (And at some point we should move the `_core` tests into `trio/tests/` too. And probably rename it to `trio/_tests/`. But that can wait for another PR...)

trio/_subprocess/tests/test_highlevel_pipes.py
```diff
@@ -0,0 +1,78 @@
+import os
+import pytest
+
+from ... import _core
+from ..unix_pipes import PipeSendStream, PipeReceiveStream, make_pipe
+from ...testing import (wait_all_tasks_blocked, check_one_way_stream)
+
+pytestmark = pytest.mark.skipif(
+    not hasattr(os, "pipe2"), reason="pipes require os.pipe2()"
+)
```
njsmith: Now that we don't use `pipe2` this looks a little odd :-). Maybe `skipif(os.name != "posix", reason=...)` would be a bit clearer. (I think it has the same effect.)

trio/_subprocess/unix_pipes.py
```diff
@@ -0,0 +1,119 @@
...
+    def __init__(self, pipefd: int, *, set_non_blocking: bool = True):
+        if not isinstance(pipefd, int):
+            raise TypeError(
+                "{0.__class__.__name__} needs a pipe fd".format(self)
+            )
+
+        self._pipe = pipefd
+        self._closed = False
+
+        if set_non_blocking:
```
njsmith: I don't think there's any reason for this to be configurable? We can just do it unconditionally.

trio/_subprocess/unix_pipes.py
```diff
@@ -0,0 +1,119 @@
...
+            return
+
+        try:
+            os.close(self._pipe)
+        except AttributeError:
+            # probably in interpreter shut down
+            pass
+        except OSError as e:
+            # already closed from somewhere else
+            if e.errno != 9:
```
njsmith: Whoops, you should never hard-code errno values like this – they're arbitrary, and can be different on different platforms. (Though 9 does seem to be EBADF on both Linux and macOS... weird!) Instead, use the stdlib `errno` module, e.g. `e.errno != errno.EBADF`.

Except... in this case, I actually think we shouldn't be filtering out EBADF. Instead we should follow the rule that once you give an fd to one of these pipe objects, the pipe object "takes ownership" of it, and is the only object that should be used to close it. Closing the fileno somewhere else is a bug.

In fact, this probably explains some of your weird test failures. I bet what's happening is:

1. The test allocates the fd, and wraps it in a trio pipe object
2. The test closes the fd by hand
3. Later, some other fd is opened, and is assigned the same numerical value as the earlier pipe fd.
4. Finally, the trio pipe object gets GCed, and its `__del__` method closes whatever random fd was opened in step 3.
5. Chaos ensues.

File descriptors are a weird kind of global state with no guard-rails. They're very C-ish. It's important to know exactly who is responsible for managing them.

trio/_subprocess/unix_pipes.py
```diff
@@ -0,0 +1,119 @@
...
+            await _core.checkpoint()
+            raise _core.ClosedResourceError("this pipe is already closed")
+
+        if not isinstance(max_bytes, int):
+            await _core.checkpoint()
+            raise TypeError("max_bytes must be integer >= 1")
+
+        if max_bytes < 1:
+            await _core.checkpoint()
+            raise ValueError("max_bytes must be integer >= 1")
```
njsmith: You can save a bit of repetition by making it `if not isinstance(max_bytes, int) or max_bytes < 1`

newsfragments/621.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for pipes (os.pipe).
```
njsmith: No newsfragment for now, since the API's not public yet.

trio/_subprocess/unix_pipes.py
Fuyukai: Can't, must be different errors.

trio/_subprocess/unix_pipes.py
```diff
@@ -0,0 +1,107 @@
...
+                with view[total_sent:] as remaining:
+                    try:
+                        total_sent += os.write(self._pipe, remaining)
+                    except BrokenPipeError as e:
+                        await _core.checkpoint()
+                        raise BrokenStreamError from e
+                    except BlockingIOError:
+                        pass
+
+                await self.wait_send_all_might_not_block()
```
njsmith: I think the way I'd structure this is:

* Just do a `checkpoint` as the very first thing, given how this is structured it doesn't seem worth the hassle of avoiding this.

* Structure the main loop like:

```python
while True:
    <try to write some data>
    <check if we're done; if so, break out of the loop>
    <wait for the pipe to be writable>
```

trio/tests/subprocess/test_unix_pipes.py
```diff
@@ -0,0 +1,88 @@
...
+
+pytestmark = pytest.mark.skipif(
+    os.name != "posix", reason="pipes are only supported on posix"
+)
+
+
+def nonblock_pipe(p: int):
+    import fcntl
+    flags = fcntl.fcntl(p, fcntl.F_GETFL)
+    fcntl.fcntl(p, fcntl.F_SETFL, flags | os.O_NONBLOCK)
```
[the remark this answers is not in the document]
Fuyukai: I clearly wasn't thinking clearly when I wrote this, whoops.

trio/_subprocess/unix_pipes.py
```diff
@@ -58,7 +58,15 @@ class PipeSendStream(_PipeMixin, SendStream):
...
-                        await self.wait_send_all_might_not_block()
+                        try:
+                            await self.wait_send_all_might_not_block()
+                        except BrokenPipeError as e:
+                            # kqueue: raises EPIPE on wait_writable instead
+                            # of sending, which is annoying
+                            # also doesn't checkpoint so we have to do that
+                            # ourselves here too
+                            await _core.checkpoint()
+                            raise BrokenStreamError from e
```
njsmith: Maybe put this inside `wait_send_all_might_not_block`?

trio/_subprocess/unix_pipes.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, pipefd: int):
+        if not isinstance(pipefd, int):
+            raise TypeError(
+                "{0.__class__.__name__} needs a pipe fd".format(self)
+            )
+
+        self._pipe = pipefd
+        self._closed = False
+
+        import fcntl
```
njsmith: Can we move this import to the top of the file, like other imports?

trio/_subprocess/unix_pipes.py
njsmith: Oh yeah, good point :-)

trio/_subprocess/unix_pipes.py
```diff
@@ -0,0 +1,111 @@
...
+            try:
+                await _core.checkpoint_if_cancelled()
+                data = os.read(self._pipe, max_bytes)
+            except BlockingIOError:
+                await _core.wait_readable(self._pipe)
+            else:
+                await _core.cancel_shielded_checkpoint()
+                break
+
+        return data
```
njsmith: I guess technically you could move the checkpoint stuff out of the loop but in practice it doesn't really matter.

trio/tests/subprocess/test_unix_pipes.py
```diff
@@ -0,0 +1,105 @@
...
+            # If you do a write of <= PIPE_BUF bytes, then it's guaranteed
+            # to either complete entirely, or not at all. So if we tried to
+            # write PIPE_BUF bytes, and the buffer's free space is only
+            # PIPE_BUF/2, then the write will raise BlockingIOError... even
+            # though a smaller write could still succeed! To avoid this,
+            # make sure to write >PIPE_BUF bytes each time, which disables
+            # the special behavior.
+            # For details, search for PIPE_BUF here:
+            #   http://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html
+            buf_size = getattr(select, "PIPE_BUF", 8192)
```
njsmith: Please add a comment on this with a link to https://bitbucket.org/pypy/pypy/issues/2876/selectpipe_buf-is-missing-on-pypy3 , to improve our chances of noticing later when it becomes unnecessary.

Fuyukai:
Well, if `__del__` doesn't close the fds, what else can it do?

trio/_subprocess/unix_pipes.py
Fuyukai: The import was originally there because windows tests would import test_unix_pipes and crash when fcntl was imported. I rearranged some stuff in the test so now this won't happen.

trio/_subprocess/unix_pipes.py
```diff
@@ -0,0 +1,127 @@
...
+
+        flags = fcntl.fcntl(self._pipe, fcntl.F_GETFL)
+        fcntl.fcntl(self._pipe, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        await self.aclose()
+        return False
```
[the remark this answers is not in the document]
Fuyukai: Oh, yeah, whoops.

PR #622, opened 2018-08-22
Add support for asynchronous waitpid on Linux systems.

Fuyukai:
This completes step 6 of https://github.com/python-trio/trio/issues/4#issuecomment-398967572.

### DIFF
--- a/trio/_subprocess/linux_waitpid.py
+++ b/trio/_subprocess/linux_waitpid.py
@@ -4,9 +4,9 @@ import os
 import outcome
 from typing import Dict, Any
 
-from . import _core
-from ._sync import Event
-from ._threads import run_sync_in_worker_thread
+from .. import _core
+from .._sync import Event
+from .._threads import run_sync_in_worker_thread
 
 # type: Dict[int, WaitpidResult]
 _pending_waitpids = {}
--- a/trio/_subprocess/linux_waitpid.py
+++ b/trio/_subprocess/linux_waitpid.py
@@ -2,34 +2,18 @@ import attr
 import functools
 import os
 import outcome
-from typing import Dict, Any
+from typing import Any
 
 from .. import _core
 from .._sync import Event
 from .._threads import run_sync_in_worker_thread
 
-# type: Dict[int, WaitpidResult]
-_pending_waitpids = {}
-
 
 @attr.s
-class WaitpidResult:
+class WaitpidState:
+    pid = attr.ib()
     event = attr.ib(default=attr.Factory(Event))
     outcome = attr.ib(default=None)
-    _cached_result = attr.ib(default=None)
-
-    def unwrap(self):
-        if self._cached_result is not None:
-            if isinstance(self._cached_result, BaseException):
-                raise self._cached_result
-            return self._cached_result
-
-        try:
-            self._cached_result = self.outcome.unwrap()
-            return self._cached_result
-        except BaseException as e:
-            self._cached_result = e
-            raise e from None
 
 
 # https://github.com/python-trio/trio/issues/618
@@ -46,40 +30,26 @@ waitpid_limiter = StubLimiter()
 
 # adapted from
 # https://github.com/python-trio/trio/issues/4#issuecomment-398967572
-async def _task(pid: int) -> None:
+async def _task(state: WaitpidState) -> None:
     """The waitpid thread runner task. This must be spawned as a system
     task."""
     partial = functools.partial(
         os.waitpid,  # function
-        pid,  # pid
+        state.pid,  # pid
         0  # no options
     )
-    try:
-        tresult = await run_sync_in_worker_thread(
-            outcome.capture,
-            partial,
-            cancellable=True,
-            limiter=waitpid_limiter
-        )
-    except Exception as e:
-        result = _pending_waitpids.pop(pid)
-        result.outcome = outcome.Error(e)
-        result.event.set()
-        raise
-    else:
-        result = _pending_waitpids.pop(pid)
-        result.outcome = tresult
-        result.event.set()
+
+    tresult = await run_sync_in_worker_thread(
+        outcome.capture, partial, cancellable=True, limiter=waitpid_limiter
+    )
+    state.outcome = tresult
+    state.event.set()
 
 
 async def waitpid(pid: int) -> Any:
     """Waits for a child process with the specified PID to finish running."""
-    try:
-        waiter = _pending_waitpids[pid]
-    except KeyError:
-        waiter = WaitpidResult()
-        _pending_waitpids[pid] = waiter
-        _core.spawn_system_task(_task, pid)
+    waiter = WaitpidState(pid=pid)
+    _core.spawn_system_task(_task, waiter)
 
     await waiter.event.wait()
-    return waiter.unwrap()
+    return waiter.outcome.unwrap()
--- a/trio/tests/subprocess/test_waitpid_linux.py
+++ b/trio/tests/subprocess/test_waitpid_linux.py
@@ -16,33 +16,24 @@ async def test_waitpid():
     pid = os.spawnvp(os.P_NOWAIT, "/bin/false", ("false",))
     result = await waitpid(pid)
     # exit code is a 16-bit int: (code, signal)
-    assert result == (pid, 256)
+    assert result[0] == pid
+    assert os.WIFEXITED(result[1]) and os.WEXITSTATUS(result[1]) == 1
 
     pid2 = os.spawnvp(os.P_NOWAIT, "/bin/true", ("true",))
     result = await waitpid(pid2)
-    assert result == (pid2, 0)
+    assert result[0] == pid2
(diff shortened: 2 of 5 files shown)


njsmith:
Highlevel comment on both this and #621: https://github.com/python-trio/trio/pull/621#issuecomment-415344689

trio/_subprocess/linux_waitpid.py
```diff
@@ -0,0 +1,85 @@
...
+            outcome.capture,
+            partial,
+            cancellable=True,
+            limiter=waitpid_limiter
+        )
+    except Exception as e:
+        result = _pending_waitpids.pop(pid)
+        result.outcome = outcome.Error(e)
+        result.event.set()
+        raise
```
njsmith: The only way to hit this `except` block (modulo bugs) is if the `run_sync_in_worker_thread` gets cancelled. Cancellation is always a `BaseException`, not an `Exception`, so right now I think this is a no-op? Also, 

It's really not clear whether we even care about cancellation here... since this is a system task, the only way it can get cancelled is if the entire `run` is shutting down, either because the main task has already finished, or because there was an internal error and we're crashing. So... I guess the only time it matters whether we call `result.event.set()` here is when someone is doing a *cancel-shielded* wait for a subprocess to finish? And if they do, we won't handle it correctly *anyway* – the user would expect a cancel-shielded wait to actually wait for the subprocess to exit, but we'll still abort the `run_sync_in_worker_task`, so we won't know when the subprocess exits. Maybe we need to resurrect #303 to handle this corner case? Or toggle the system task's shielding on and off depending on whether anyone is currently blocked in `Popen.wait`? Or use `threading.Thread` directly instead of trying to go through `run_sync_in_worker_task`? (...can we even reliably re-enter the trio thread once the system nursery is cancelled? The re-entry queue processor is also a system task...)

Man, I hate `waitpid`.

This is such an obscure use case that I don't think we need to worry about it right now. Trying to do a cancel-shielded wait for a subprocess isn't *too* ridiculous – `nursery.__aexit__` is one of the standard use cases for cancel-shielding, and people will want to do nursery-like things for managing subprocesses. So we might end up caring eventually. But right now we should just shrug and accept that it won't do exactly the right thing here. And if we don't care about that, we can make this whole function a lot simpler, like...

```python
try:
    result.outcome = await run_sync_in_worker_thread(...)
finally:
    # Paranoia to make sure we unblock any waiters even if something goes horribly wrong
    result.event.set()
```

trio/tests/subprocess/test_waitpid_linux.py
```diff
@@ -0,0 +1,48 @@
...
+
+pytestmark = pytest.mark.skipif(
+    sys.platform != "linux", reason="linux waitpid only works on linux"
+)
+
+
+async def test_waitpid():
+    pid = os.spawnvp(os.P_NOWAIT, "/bin/false", ("false",))
+    result = await waitpid(pid)
+    # exit code is a 16-bit int: (code, signal)
```
njsmith: In theory it's a bit more complicated than that, there's a "was a core dumped?" flag in there and the ability to distinguish between stop signals and termination signals ([see](https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/waitstatus.h;h=a1c232612cd13e758f222ade6b27bc2c85ab8f8f;hb=HEAD)).

In practice this is a pretty pedantic distinction. If we want to be really POSIX-ly correct, though, I guess the tests should make assertions like

`assert os.WIFEXITED(code) and os.WEXITSTATUS(code) == 1`
`assert os.WIFEXITED(code) and os.WEXITSTATUS(code) == 0`
`assert os.WIFSIGNALED(code) and os.WTERMSIG(code) == signal.SIGKILL`

?

trio/tests/subprocess/test_waitpid_linux.py
```diff
@@ -0,0 +1,48 @@
...
+        n.start_soon(waiter)
+        n.start_soon(waiter)
+
+        os.kill(pid, signal.SIGKILL)
+
+
+async def test_waitpid_no_process():
+    with pytest.raises(ChildProcessError):
+        # this PID probably doesn't exist
+        await waitpid(100000)
```
njsmith: oo I thought of a trick to make this deterministic.

You can only wait for your child processes. So we need the pid of a process that we know is not a child process. How about... `waitpid(os.getpid())`? I think that deterministically raises the error.

Or I guess `waitpid(1)` would also work, since `init` has to exist and is never a child of any other process.

trio/_subprocess/linux_waitpid.py
njsmith: > Trying to do a cancel-shielded wait for a subprocess isn't too ridiculous – `nursery.__aexit__` is one of the standard use cases for cancel-shielding, and people will want to do nursery-like things for managing subprocesses.

Actually, on further thought, this is still overstating the importance of this edge case. If someone does a cancel-shielded wait for a subprocess, that will mostly work fine, even with my "simplified" code above. The only way this task gets cancelled is if we're crashing, or if the main task has already exited. That means that the only ways a cancel-shielded `Popen.wait` can fail to actually wait are:

* We're in the middle of crashing with a `TrioInternalError`: well, sorry, that means internal invariants have been violated and we're blowing up the world, so it's OK if your task manager doesn't wait for child processes correctly. At this point all guarantees are void.

* If we're not crashing with `TrioInternalError`, and the system task is cancelled, and someone is doing a cancel-shielded `Popen.wait`, then the main task has already exited, so they must be doing it from *inside a system task*. But doing *anything* inside a cancel-shield in a system task is highly dubious, because by the time a system task gets cancelled the world is being torn down around you. Don't do that please.

Fuyukai:
I have *no clue* what the hell happened to the tests here.

njsmith:
Huh, bizarre. Could be a change in a third-party dependency, like pytest-cov or coverage...?

njsmith:
There's something weird with your rebase too... did you accidentally rebase master onto this, or something? Somehow the commits in this branch that came from master have different hashes than they do on master...

smurfix:
Owch. You've been rebasing in the wrong direction, i.e. instead of rebasing your work on top of the release you've been rebasing the release tree on top of your work. :-(

smurfix:
Frankly I don't like rebasing anyway, just reset to 133e46b45, merge up, and be done with it.

smurfix:
(and then cherry-pick cfbedd00b)

Fuyukai:
Oh I think I accidentally rebased onto master then master from my fork onto this. Whoops.

njsmith:
Still getting those `coverage combine` failures though :-(. What on earth...

Lacking better ideas, at this point I'd probably try debugging by trying to isolate what exactly is triggering that, e.g. by temporarily pushing a commit that turns off all the tests added in this PR, and by opening a trivial no-op PR to confirm whether the issue even is specific to this PR.

pquentin:
OK, the coverage combine failures were sorted out in #647, thanks @njsmith!

trio/_subprocess/linux_waitpid.py
```diff
@@ -0,0 +1,58 @@
...
+    state.event.set()
+
+
+async def waitpid(pid: int) -> Any:
+    """Waits for a child process with the specified PID to finish running."""
+    waiter = WaitpidState(pid=pid)
+    _core.spawn_system_task(_task, waiter)
+
+    await waiter.event.wait()
+    return waiter.outcome.unwrap()
```
njsmith: I guess we'll eventually need to split this up into two functions, but that's fine, no particular reason to do that now.

trio/tests/subprocess/test_waitpid_linux.py
```diff
@@ -0,0 +1,36 @@
...
+pytestmark = pytest.mark.skipif(
+    sys.platform != "linux", reason="linux waitpid only works on linux"
+)
+
+
+async def test_waitpid():
+    pid = os.spawnvp(os.P_NOWAIT, "/bin/false", ("false",))
+    result = await waitpid(pid)
+    # exit code is a 16-bit int: (code, signal)
+    assert os.WIFEXITED(result[1]) and os.WEXITSTATUS(result[1]) == 1
```
njsmith: We should probably `assert result[0] == pid` as well

njsmith:
Shouldn't need a rebase – the CI systems don't test the head of the PR branch, they make a temporary merge between the PR branch and master and then test the merge. So the next time they run on this PR, they should pick up the fix from #647.

njsmith:
Tests are failing (typo in the new asserts)

njsmith:
Jenkins seems to be confused... Maybe this will tickle it?

njsmith:
That was a new one: apparently Jenkins just didn't notice the latest push or something; it didn't even create a job record, never mind actually run anything. But close/reopen seems to have fixed it. (Which is funny because usually Jenkins ignores close/reopen.)

njsmith:
Ok code looks good, CI is green, I'm going to merge this PR quick before it gets hit *another* bizarre mishap.

PR #623, opened 2018-08-23
Rip out all the old deprecated stuff.

Fuyukai:


njsmith:
Can you throw in a newsfragment? Can be simple, just `623.removal.rst` and something like "Removed all APIs that were deprecated in v0.4.0 or earlier." (Assuming that's true, I haven't checked :-).)

This will also trigger jenkins to re-run and hopefully clear the transient failure...

njsmith:
Ah, codecov is actually useful this time! It looks like there's still some task-local related code left behind in `_run.py`: https://codecov.io/gh/python-trio/trio/pull/623/src/trio/_core/_run.py

I guess we should get rid of the `Task._locals` attribute entirely.

njsmith:
Thanks!

PR #625, opened 2018-08-23
Rename MacOS to macOS

pquentin:
https://en.wikipedia.org/wiki/MacOS

We fixed illumos, so hey, why not fix this one too? The only occurrence I left untouched is in setup.py because the PyPI classifier still uses the old name: https://pypi.org/search/?c=Operating+System+%3A%3A+MacOS+%3A%3A+MacOS+X.


njsmith:
...sure why not

Issue #604, opened 2018-08-14
Introduce performance measurements


sorcio:
Obligatory background: we all agree (or at least nodded our heads in agreement when this was mentioned many times in the past) that Trio has a number of priorities over speed of execution. Optimizing for speed or any other performance metric should never hinder simplicity, "humanness", correctness, maintainability, predictability, security, debuggability, or other design principles. Premature optimization _might_ be the root of _some_ evil, and focusing on efficiency carries the risk of getting in the way of innovation. Even just identifying what is performance critical takes a lot of effort and is a moving target.

Given all of this, I believe we should **provide some tools to make sure that Trio doesn't add too much overhead and allows efficient user code to run efficiently**. This is more complicated than "create a bunch of benchmarks" but I want to start by creating a bunch of benchmarks, at least to start thinking about this. I had found myself caring about performance in a few cases and I want to do more work in that direction. It's relevant to me and it might be to someone else so it would be nice to share.

When I say a bunch of benchmarks I think of something that should reproduce a realistic workload for a Trio application. I don't think we are in a position to actually understand what a "realistic workload" is. I have a couple examples that I can derive from things I do for my paid job, and we can think of some more more (including something from projects you are working on, porting asyncio benchmarks and examples, adding trio-asyncio examples, etc). We can also produce some microbenchmarks. Also "performance" is a broad term that can mean many different things. In an I/O application it can mean throughput, latency, bandwidth usage, fairness, CPU impact, memory impact, or anything I did not think about. Covering all of these aspects can be hard (although sometimes some is a good proxy for others, e.g. a microbenchmark for execution time of some operation may provide hints about CPU, latency and throughput).

In terms of process, I don't think that performance measurements should get in the way of approving a PR or anything like that. Yeah, it can be nice to have comparisons but, really, I don't think that we are anywhere close to ready to make performance part of the development process. Benchmarks are lies that can have a mild informative value that is often hard to interpret. Still, like a good test suite, a benchmark suite can sometimes make it obvious that something is a regression, or an improvement, and it can be improved over time to provide better coverage of real-world use cases.

Anything we can learn from other projects? Tools (benchmark runners and the such) we can borrow? General feedback?

@njsmith would you see the `trio` repo as the space for this or would it better live in a different project?

njsmith:
[...]

PR #628, opened 2018-08-23
Add checks for if instruments exist before running instrument functions.

Fuyukai:
This provides a small speed-up.

Microbenchmark script:

```py
import time

import trio


async def main():
    q = trio.Queue(0)

    async def i1():
        await q.put("123")

    async def i2():
        i = await q.get()

    start = time.time()
    print("Start time", start)
        for x in range(0, 5_000):
            n.start_soon(i1)
            n.start_soon(i2)
            if x % 500 == 0:
                await trio.sleep(0)

    end = time.time()
    print("End time", end)
    print("Elapsed", end - start)


trio.run(main)
```

Before:
```
Start time 1535036274.015487
End time 1535036274.5630355
Elapsed 0.5475485324859619
```

After:
```
Start time 1535036229.3229249
End time 1535036229.8288205
Elapsed 0.5058956146240234
```

I generally got a ~40-50ms speedup with this change.


njsmith:
eh, in the long run I think we should try to approach these things systematically based on benchmarks and bottlenecks people experience in real life (#604), but this particular change is so trivial and harmless that sure, why not.

Fuyukai:
I actually discovered this by a profile - a suspicious amount of time was being spent inside instrument() so I added these guards and they seemed to reduce it by a bit.

njsmith:
Oh interesting, what were you profiling?

njsmith:
Neat. BTW, if it's possible to convert your bot code into a benchmark (like literally: take the existing code and wrap a harness around it), that could be a good way to start at #604 ... the more quirky real-world stuff is in a benchmark, the better it is for guiding these kinds of decisions :-).

Issue #408, opened 2018-01-20
MultiError.catch should take an exception type, like except: does


njsmith:
In regular Python, a catch-all exception handler looks like:

```python
try:
    ...
except Exception as exc:  # Don't catch KeyboardInterrupt, etc.
    logger.error(..., exc_info=exc)
```

In Trio we need to handle `MultiError`s, which might contain a mix of exceptions we want to catch and ones we don't, etc. So you have to use `MultiError.catch`, and the above becomes:

```python
def handler(exc):
    if isinstance(exc, Exception):
        logger.error(..., exc_info=exc)
        return None  # swallow this exception
    else:
        return exc  # let other exceptions propagate

with MultiError.catch(handler):
    ...
```

This is pointlessly cumbersome. We should make it look like this:

```python
def handler(exc):
    logger.error(..., exc_info=exc)

with MultiError.catch(Exception, handler):  # Note the extra argument
    ...
```

WindSoilder:
For the newer version of catch, if we have
```python
with MultiError.catch(RuntimeError, handler):
   ...
```
Does it means that we want the MultiError filter out the `RuntimeError` (and `Cancelled`) ?

If so, what if we want to keep the behavior default? (just filter out the `Cancelled` exception)....May be just pass `None` to catch?  So the code would be something like:
```python
with MultiError.catch(None, handler):
   ...
```
Err... Is the form of this ok ?

njsmith:
@windsoilder the idea is that

```python
with MultiError.catch(RuntimeError, handler):
    ... body ...
```

is a substitute for

```python
try:
    ... body ...
except RuntimeError as exc:
    handler(exc)
```

The work the same except the first one handles `MultiError`s.

When you talk about filtering `Cancelled`, you're thinking of what cancel scope blocks do. `MultiError` itself doesn't know anything about cancellation. In fact, `open_cancel_scope` uses `MultiError.catch` to catch `Cancelled` exceptions. So that's a higher level than all this.

WindSoilder:
@njsmith Thanks for your comment!  It makes me more clear about this.

But there is something still confuse me, take the code as example:
```python
with MultiError.catch(Cancelled, handler):
    ... body ...
```

According to your description, when handler is called, it's argument type(`Exception`) can be `RuntimeError` or `MultiError`. 

[...]

Issue #611, opened 2018-08-18
MultiError v2


njsmith:
`MultiError` is the one part of trio's core design that I'm really not satisfied with. Trying to support multiple exceptions on top of the language's firm one-exception-at-a-time stance raises (heh) all kinds of issues. Some of them can probably be fixed by changing the design. But the hardest problem is that there are lots of third-party packages that want to do custom handling of exception tracebacks (e.g. ipython, pytest, raven/sentry). And right now we have to monkeypatch all of them to work with MultiError, with more or less pain and success.

Now @1st1 wants to add nurseries to asyncio, and as part of that he'll need to add something like `MultiError` to the stdlib. We talked a bunch about this at PyCon, and the plan is to figure out how to do this in a way that works for both asyncio and trio. Probably we'll start by splitting `MultiError` off into a standalone library, that trio and uvloop can both consume, and then add that library to the stdlib in 3.8 (and then the library will remain as a backport library for those using trio on 3.7 and earlier). This way asyncio can build on our experience, and trio can get out of the monkeypatching business (because if `MultiError`s are in the stdlib, then it becomes ipython/pytest/sentry's job to figure out how to cope with them).

But before we can do that we need to fix the design, and do it in writing so we (and Yury) can all see that the new design is right :-).

# Current design

[If this were a PEP I'd talk more about the basic assumptions underlying the design: multiple errors cna happen concurrently, you need to preserve that fact, you need to be able to catch some-but-not-all of those errors, you need to make sure that don't accidentally throw away any errors that you didn't explicitly try to catch, etc. But this is a working doc so I'm just going to dive into the details...]

Currently, trio thinks of `MultiError` objects as being ephemeral things. It tries as much as possible to simulate a system where multiple exceptions just happen to propagate next to each other. So it's important to keep track of the individual errors and their tracebacks, but the `MultiError` objects themselves are just a detail needed to accomplish this.

[...]

Issue #607, opened 2018-08-16
idea: unbound cancel scopes


njsmith:
I've recently run into a few places where I want a cancel scope for some code that may or may not already be running. For example, in pytest-trio, if a fixture crashes you want to cancel the main test... but these run in different tasks, so it's tricky to find the main task's cancel scope and put it somewhere that a fixture can get at it, without race conditions.

It's possible to create an object that sort of acts like a cancel scope, but where you can call `cancel` before or after the code inside the scope starts running. But it's fairly tricky to get all the cases right, e.g.:

```python
@attr.s
class UnboundCancelScope:
    cancel_called = attr.ib(default=False)
    _cancel_scope = attri.ib(default=None)

    def cancel(self):
        self.cancel_called = True
        if self._cancel_scope is not None:
            self._cancel_scope.cancel()

    def __enter__(self):
        self._cancel_scope = trio.open_cancel_scope().__enter__()
        if self.cancel_called:
            self._cancel_scope.cancel()

    def __exit__(self, *args):
        return self._cancel_scope.__exit__(*args)

# Creation:
unbound_cancel_scope = UnboundCancelScope()

# Entering:
with unbound_cancel_scope:
    ...
```

Maybe we should make this just... how cancel scopes work, always? Right now `open_cancel_scope` is a context manager, so it forces you to immediately enter the scope it creates. But we could reinterpret it as returning an unbound cancel scope object, and then the `with cancel_scope: ...` as entering that scope – it'd even be backwards compatible!

Implementation-wise, I think it'd be almost trivial. The one thing to watch out for is that it'd become possible to attempt to re-enter a scope that you're already inside, which would be complicated (e.g. instead of keeping a set of which tasks are inside the scope, we'd have to keep a dict of task → refcount). For now we should just error out if someone tries to do this. (OTOH, I think having multiple independent tasks entering the same scope is fine and would Just Work.)

Maybe we should also make `CancelScope` public? Right now it's hidden in order to keep the constructor private, but that would be unnecessary in this approach – in fact `open_cancel_scope(...)` would just be `return CancelScope(...)`, so maybe we'd even want to deprecate it or something.

One limitation of this approach is that `cancelled_caught` would become ambiguous if multiple tasks can enter the same scope. It might not matter.

# Alternatives

[...]

PR #630, opened 2018-08-24
add Exception type for MultiError.catch

WindSoilder:
which added for feature #408 .  And change some relative unit test

### DIFF
--- a/newsfragments/408.feature.rst
+++ b/newsfragments/408.feature.rst
@@ -1,0 +1,1 @@
+For now :meth:`MultiError.catch` should take an exception type, And the ``handler`` function is called when the `exception` is `exception type`.
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -92,7 +92,7 @@ def _filter_impl(handler, root_exc, exc_type):
                 return exc
         else:
             # only handle for exception when it's exc_type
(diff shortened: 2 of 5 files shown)


njsmith:
Hey, thanks for jumping in to work on this!

Implementing the ideas in #611 is going to be a bit complicated. We're going to have to make several compatibility-breaking changes:

* nurseries start raising `MultiError`s, even if there's only one exception
* `MultiError.catch` starts taking the exception type (as in this PR)
* `MultiError.catch` starts calling the handler once with a combined exception, instead of a bunch of times on individual exceptions

Also it'd be nice to coordinate a bit with the asyncio folks before committing to something here, since we're hoping to converge with them on this part.

Given all that, I think we probably want to move a little slowly here, and try to batch up all these changes together, instead of doing multiple rounds of compatibility-breaking changes. So, I'm afraid I'm rejecting this PR in its current form.

#607 is somewhat similar in terms of refactoring some of trio's core APIs, and there's a pretty concrete plan for what to do (especially in this comment: https://github.com/python-trio/trio/issues/607#issuecomment-413739258, once you dig through all my verbiage). Or if you want to keep working on the `MultiError` stuff, I have a half-written prototype locally that I'll try to at least push up somewhere in the next few days, and we could work together on that?

Issue #56, opened 2017-02-20
Better ergonomics for exceptions


njsmith:
Our error-reporting story is tolerable at this point (see #49 for some history), but could definitely still be improved. This collects some misc ideas:

* [ ] Our tracebacks are very cluttered and hard to read (lots of internal helper functions etc. all over the place) – can we cut this down?
  * [x] Rewrite nursery context manager without `@asynccontextmanager` and `@async_generator` to reduce extraneous frames
  * [ ] Avoid raising internally to communicate gathered exceptions up to nursery exit (see https://github.com/python-trio/trio/pull/612#issuecomment-414201886 and https://github.com/python-trio/trio/issues/56#issuecomment-414887864)
  * [ ] Maybe elide trio-internal stuff with [...4 frames skipped...] to help users focus on their code?  May be at odds with pushing our traceback logic down to the interpreter.
* [ ] Tracebacks should include task context, like: as an exception propagates in (`Error.unwrap`) and out (`_task_finished`) of tasks, update a record like `exc.__trio_task_context__` with entries like `(len(tb), "exit", task)`, and use them to add context to printed tracebacks. (The need to compute `len` is a bit unfortunate though... I guess ideally these should be attributes on the tb objects themselves? But tb objects don't have `__dict__`s. Or we could just stash the tb object.) `MultiError.filter` needs to be careful to propagate these correctly when propagating tracebacks.
* [ ] `MultiError.catch` is OK, but could be better
  * [ ] We don't propagate `__context__` on `MultiError`s
  * [ ] We should really have an async version... but [bpo29600](https://bugs.python.org/issue29600) makes this difficult!
  * [ ] The filter interface is maybe a bit awkward and error prone?
  * [ ] If people try to work around the awkwardness by doing some sort of `try: raise exc except: ...` thing, then they get corrupt tracebacks (we assume that the filter callback doesn't touch the traceback). We could potentially detect new tb frames appearing and fix them up.
* [ ] Better IPython integration: right now we override the printing of `MultiError`s and use our regular `MultiError` printing code. This is good b/c it gives full information, but bad b/c it loses out on IPython's enhanced exception formatting. #49 has some notes on how this might be possible to get both. (Note that this becomes more of an issue if we do as suggested above and start customizing *all* exception output and not just `MultiError`!) Possibly should chat with the IPython devs about this.

bluetech:
Are you familiar with C#'s `AggregateException`?

[...]

PR #631, opened 2018-08-25
Elide task.context.run() and contextvars.callable() frames from tracebacks

belm0:
for #56 

TODO:
  * [x] write tests

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -3,8 +3,8 @@ import logging
 import os
 import random
 import select
-import sys
 import threading
+import traceback
 from collections import deque
 import collections.abc
 from contextlib import contextmanager, closing
@@ -1378,12 +1378,13 @@ def run_impl(runner, async_fn, args):
                 final_result = Value(stop_iteration.value)
             except BaseException as task_exc:
                 # Store for later, removing uninteresting top frames:
-                #   1. trio._core._run.run_impl()
-                #   2. contextvars.Context.run() (< Python 3.7 only)
-                tb_next = task_exc.__traceback__.tb_next
-                if sys.version_info < (3, 7):  # pragma: no cover
-                    tb_next = tb_next.tb_next
-                final_result = Error(task_exc.with_traceback(tb_next))
+                tb = task_exc.__traceback__
+                for path, _, _, _ in traceback.extract_tb(tb):
+                    if '/trio/_core/' in path or '/contextvars/' in path:
+                        tb = tb.tb_next
+                    else:
+                        break
+                final_result = Error(task_exc.with_traceback(tb))
 
             if final_result is not None:
                 # We can't call this directly inside the except: blocks above,
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -3,6 +3,7 @@ import logging
 import os
 import random
 import select
+import sys
 import threading
 from collections import deque
 import collections.abc
@@ -1376,7 +1377,13 @@ def run_impl(runner, async_fn, args):
             except StopIteration as stop_iteration:
                 final_result = Value(stop_iteration.value)
             except BaseException as task_exc:
-                final_result = Error(task_exc)
+                # Store for later, removing uninteresting top frames:
+                #   1. trio._core._run.run_impl()
+                #   2. contextvars.Context.run() (< Python 3.7 only)
+                tb_next = task_exc.__traceback__.tb_next
+                if sys.version_info < (3, 7):  # pragma: no cover
+                    tb_next = tb_next.tb_next
+                final_result = Error(task_exc.with_traceback(tb_next))
 
             if final_result is not None:
                 # We can't call this directly inside the except: blocks above,
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -3,8 +3,8 @@ import logging
 import os
 import random
 import select
+import sys
 import threading
-import traceback
 from collections import deque
 import collections.abc
 from contextlib import contextmanager, closing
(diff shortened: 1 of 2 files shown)


trio/_core/_run.py
```diff
@@ -1376,7 +1377,13 @@ def run_impl(runner, async_fn, args):
             except StopIteration as stop_iteration:
                 final_result = Value(stop_iteration.value)
             except BaseException as task_exc:
-                final_result = Error(task_exc)
+                # Store for later, removing uninteresting top frames:
+                #   1. task.context.run()
+                #   2. contextvars.callable() (< Python 3.7 only)
```
njsmith: Huh, I thought the top two frames were `trio._core._run.run_impl` and (on <3.7 only) `contextvars.Context.run`. Do you know where the discrepancy comes from? am I confused? :-)

belm0: here's what you wrote in #56:

```
  File "/.../site_packages/trio/_core/_run.py", line 1374, in run_impl
    msg = task.context.run(task.coro.send, next_send)
```
> This is noise... when we catch a user exception in run_impl, we should unconditionally discard the top frame in the traceback (which will be this one).

```
  File "/.../site-packages/contextvars/__init__.py", line 38, in run
    return callable(*args, **kwargs)
```
> This line will disappear on 3.7, where contextvars.Context.run is implemented in C, so maybe not worth worrying about. Or we could discard it at the same time we fix the run_impl frame above.

belm0: I see, I'm enumerating the child functions rather than the parents.

njsmith: Ahhhh yeah, right, so we're just looking at different lines in the traceback :-). that makes sense!

belm0:
Thanks for the test guidance.  It almost works, but appears to be testing the frame at the wrong end of the trace.  Investigating...

njsmith:
(`import traceback; traceback.print_tb(exc.__traceback__)` may be a useful incantation to have up your sleeve.)

belm0:
The first exception of the MultiError is as follows:
```
  File "/Users/john/dev/trio/trio/_core/_run.py", line 201, in open_cancel_scope
    yield scope
  File "/Users/john/dev/trio/trio/_core/_run.py", line 319, in __aexit__
    await self._nursery._nested_child_finished(exc)
  File "/Users/john/dev/trio/trio/_core/_run.py", line 425, in _nested_child_finished
    raise MultiError(self._pending_excs)
  File "/Users/john/dev/trio/trio/_core/tests/test_run.py", line 1903, in my_child_task
    raise KeyError()
```

trio/_core/_run.py
```diff
@@ -1376,7 +1377,13 @@ def run_impl(runner, async_fn, args):
             except StopIteration as stop_iteration:
                 final_result = Value(stop_iteration.value)
             except BaseException as task_exc:
-                final_result = Error(task_exc)
+                # Store for later, removing uninteresting top frames:
+                #   1. trio._core._run.run_impl()
+                #   2. contextvars.Context.run() (< Python 3.7 only)
+                tb_next = task_exc.__traceback__.tb_next
+                if sys.version_info < (3, 7):  # pragma: no cover
```
njsmith: `pragma: no cover` is for cases where we don't even want to have test coverage. Here, not only is it useful to have tests that cover both branches of the `if`, we actually have those tests :-). (We test on multiple python versions and combine the coverage.) So we shouldn't use `pragma: no cover` here.

trio/_core/tests/test_run.py
```diff
@@ -1897,6 +1899,27 @@ def handle(exc):
...
+        # force it by raising two exceptions.
+        async with _core.open_nursery() as nursery:
+            nursery.start_soon(my_child_task)
+            nursery.start_soon(my_child_task)
+    except _core.MultiError as exc:
+        first_exc = exc.exceptions[0]
+        assert isinstance(first_exc, KeyError)
+        tb_text = ''.join(traceback.format_tb(first_exc.__traceback__))
+        for r in ('/trio/_core/.* in run_impl$', '/contextvars/.* in run$'):
+            assert not re.search(r, tb_text, re.MULTILINE)
```
njsmith: If you're doing it this way, then I think you can simplify to just

```python
try:
        nursery.start_soon(my_child_task)
except KeyError as exc:
    ...
```

(Of course that will then need adjusting when we're working on #611, but everything will, so I wouldn't worry about that.)

trio/_core/_run.py
belm0: Oh that's good--  I assumed the coverage testing was looking at each version run individually rather than the aggregate.

njsmith:
I mentioned this a bit in chat, but let me say it again here with more carefully collected thoughts :-).

So the traceback frames that are relevant for this PR are:

* the frame that appears when `run_impl` catches the exception (always exactly one, guaranteed by the python language)

* The frame added by older versions of `Context.run` (doesn't apply on python 3.7+)

* The frames that older versions of `Context.run` might or might not add in the future, in case Yury changes the backport library (e.g. by adding another frame, or rewriting it in C so it stops adding any frames).

At one point I was confused and thought that this was also attempting to address the frames that the nursery aexit machinery is adding, but that was wrong, it can't and doesn't affect those either way. It's just about the frames mentioned above.

So some options include:

* Remove one frame, ignore `Context.run`. Extremely simple, works perfectly for recent enough python (and eventually that's everyone), degrades gracefully on older pythons. Not perfect but not too bad.

* Remove one frame on py37+, remove 2 frames otherwise. Provides perfect results on all current deployments. Might need adjusting in the future if `contextvars` changes. We could catch this with a test so we're at least alerted if it happens.

* Check how many frames `Context.run` actually uses at startup, and later remove 1+that many frames. Extremely robust, and keeps the main loop simple and tight, but at the cost of ~5-10 extra lines of code to do the calculation.

These all seem like viable options, with slightly different complexity/reliability tradeoffs.

The current PR has a different approach, based on calling `extract_tb` and looking at filenames. This unfortunately has a few issues:

* Unlike the options above, it has the potential for producing incorrect results in innocent-seeming circumstances. For example, what if the code in the task actually is part of `trio/_core`, like `nursery.start_soon(wait_readable, fd)`? What if someone has a module `mylib/contextvars/blah.py`? These aren't terribly common situations, but with the other options above these aren't even possibilities.

* Performance: `extract_tb` is *way* heavier than is appropriate for this part of the code. It does disk I/O! We could avoid this by writing our own filename extraction code, but that would add more complexity.

* Correctness: I'm pretty sure that the current PR doesn't work on Windows, where the filename contains `trio\_core`. Again, fixable, but only by adding more complexity. And there's something just... disconcerting... about the idea that core exception propagation code needs to know about `os.sep`.

So I don't think we should be checking filenames here.

On an unrelated note, the test isn't actually testing anything right now, I think? Like, it passes on current master, right, regardless of this change? If we want to test this change specifically we could iterate over the traceback and assert that `run_impl` and `contextvars` don't show up there... Or wait for the next change to land them both together, that's a little naughty but probably fine under the circumstances.

belm0:
@njsmith ready to go-- thanks for your review and patience

njsmith:
Thank you for *your* patience :-)

PR #632, opened 2018-08-26
Fix subtle bug in Happy Eyeballs

davidbrochart:
When a connection attempt succeeds in almost the same time as `happy_eyeballs_delay`, but just before, a next attempt is done anyway. This can be seen with the following code, just change `bug_fix = True` to get the bug fix. You might need to run the program several times to see the bug. When the bug happens, you see:

```
Connection to 0 (in 1 seconds)
Connected to 0
Connection to 1 (in 1 seconds)
```

Instead of (with bug fix):

```
Connection to 0 (in 1 seconds)
Connected to 0
```

When the bug doesn't happen, you see:

```
Connection to 0 (in 1 seconds)
Connection to 1 (in 1 seconds)
Connected to 0
```

```python
import trio

bug_fix = False

async def sock_connect(target, delay):
    print(f'Connection to {target} (in {delay} seconds)')
    await trio.sleep(delay)
    print(f'Connected to {target}')

async def connect(targets, happy_eyeballs_delay=1):
    winning_sockets = []
    targets_iter = iter(targets)

    async def attempt_connect(nursery, previous_attempt_failed):
        if previous_attempt_failed is not None:
            with trio.move_on_after(happy_eyeballs_delay):
                await previous_attempt_failed.wait()

        try:
            target = next(targets_iter)
        except StopIteration:
            return

        this_attempt_failed = trio.Event()
        nursery.start_soon(attempt_connect, nursery, this_attempt_failed)

        if (not bug_fix) or (bug_fix and not winning_sockets):
            try:
                await sock_connect(target, happy_eyeballs_delay)
            except OSError as exc:
                oserrors.append(exc)
                this_attempt_failed.set()
            else:
                winning_sockets.append(target)
                nursery.cancel_scope.cancel()

       nursery.start_soon(attempt_connect, nursery, None)

    if winning_sockets:
        first_prize = winning_sockets.pop(0)
        for sock in winning_sockets:
            print(f'close {sock}')
        return first_prize
    else:
        assert len(oserrors) == len(targets)
        msg = "all attempts to connect to {} failed".format(
            format_host_port(host, port)
        )
        raise OSError(msg) from trio.MultiError(oserrors)

async def main():
    targets = list(range(2))
    await connect(targets) 

trio.run(main)
```

### DIFF
--- a/trio/_highlevel_open_tcp_stream.py
+++ b/trio/_highlevel_open_tcp_stream.py
@@ -274,22 +274,25 @@ async def open_tcp_stream(
         this_attempt_failed = trio.Event()
         nursery.start_soon(attempt_connect, nursery, this_attempt_failed)
 
-        # Then make this invocation's attempt
-        try:
-            with close_on_error(socket(*socket_args)) as sock:
-                await sock.connect(target_sockaddr)
-        except OSError as exc:
-            # This connection attempt failed, but the next one might
-            # succeed. Save the error for later so we can report it if
-            # everything fails, and tell the next attempt that it should go
-            # ahead (if it hasn't already).
-            oserrors.append(exc)
-            this_attempt_failed.set()
-        else:
-            # Success! Save the winning socket and cancel all outstanding
-            # connection attempts.
-            winning_sockets.append(sock)
-            nursery.cancel_scope.cancel()
+        # Then make this invocation's attempt, but only if the previous one
+        # didn't succeed just now (and didn't have time to cancel all
+        # outstanding connection attempts).
+        if not winning_sockets:
+            try:
+                with close_on_error(socket(*socket_args)) as sock:
+                    await sock.connect(target_sockaddr)
+            except OSError as exc:
+                # This connection attempt failed, but the next one might
+                # succeed. Save the error for later so we can report it if
+                # everything fails, and tell the next attempt that it should go
+                # ahead (if it hasn't already).
+                oserrors.append(exc)
+                this_attempt_failed.set()
+            else:
+                # Success! Save the winning socket and cancel all outstanding
+                # connection attempts.
+                winning_sockets.append(sock)
+                nursery.cancel_scope.cancel()
 
     # Kick off the chain of connection attempts.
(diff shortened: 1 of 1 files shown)


njsmith:
This is so subtle that I'm not sure it's a bug at all :-). Does it cause any problems?

davidbrochart:
I'd say yes because it is wasting resources in trying to make another connection while we are already connected.

njsmith:
> I'd say yes because it is wasting resources in trying to make another connection while we are already connected.

Fair enough. The thing is, though, this is kind of an inevitable risk of using any happy eyeballs algorithm... say a packet comes in accepting a connection. It arrives in the network hardware, which notifies the network driver, which notifies the kernel networking subsystem, which notifies the trio I/O loop, which notifies the task that called `connect`, which then notifies the other tasks that they should give up. This all takes time – generally a tiny amount of time, like a millisecond or so, but some time. And that means that there's always going to be a tiny window where some part of the computer can know that our connection has succeeded, but where we still start a new connection anyway.

Also, the happy eyeballs timeout is pretty arbitrary anyway – if you set the timeout to 250 ms, and a connection succeeds at 251 ms, then whoops you just made an unnecessary connection, one you could have avoided if you'd set the timeout to 252 ms instead. It's just the nature of the beast, and a trade-off we accept in order to get happy eyeballs' robustness and latency benefits.

The upside of this patch is that it shaves a fraction of a second off of the effective timeout. The downside is that it makes the code harder to understand and to test. (Notice the codecov report saying that this adds some untested code – and it wouldn't be easy to write a new test that reliably hits this case!) Given that the timeout is already a vague and arbitrary thing, and that the happy eyeballs code starts out quite difficult to understand and test, I think in this case it's not worth it, so I'm declining this PR.

Thank you though for the contribution, even if it didn't work out this time! It's exciting to see people thinking deeply about these things, and if you have other suggestions I look forward to seeing them :-).

davidbrochart:
This all makes sense. Thanks for the detailed arguments.

Issue #624, opened 2018-08-23
Flaky test failure: test_ki_wakes_us_up on Windows


njsmith:
As noticed by @Fuyukai:

https://ci.appveyor.com/project/njsmith/trio/build/1.0.1302/job/nm14lkm4vpwgk6yw

I think this is the same issue we ran into here: https://github.com/python-trio/trio/pull/575#discussion_r206379315

Basically on Windows, `time.sleep` and `time.monotonic` use slightly different clocks, so it's possible to do `time.sleep(1)` and then have `time.monotonic` report that slightly less than 1 second has passed. (`trio.sleep` doesn't have this problem, fortunately ­– only `time.sleep` does.)

For this particular test, the timing isn't too important – the only reason we check is to make sure that we notice in case something weird happens and its taking like, tens of seconds.

So the easy solution is: make it `time.sleep(1.1)` (with a comment back to this issue!) and move on with our lives.

wgwz:
Hey! So I implemented a fix for this in #723. Let me know if I should do anything else. :-)

PR #633, opened 2018-08-26
Fix error messages for host not found when `host` is passed in as bytes

ncollins:
open_tcp_stream can take `host` as either bytes or str, but the formatting for the error message when a host is not found depends on host being a string. When trying to connect to a non-existent host, if passed as a string I would get the exception:

OSError: all attempts to connect to localhost:8081 failed

but when passed as bytes I would get the exception:

    if ":" in host:
TypeError: a bytes-like object is required, not 'str'

This PR adds a single line to convert bytes to a string before running the rest of the function (and some tests).

### DIFF
--- a/newsfragments/633.bugfix.rst
+++ b/newsfragments/633.bugfix.rst
@@ -1,0 +1,3 @@
+Updates the formatting of exception messages raised by :func:`trio.open_tcp_stream` 
+to correctly handle a hostname passed in as bytes, by converting the hostname to
+a string.
--- a/trio/_highlevel_open_tcp_stream.py
+++ b/trio/_highlevel_open_tcp_stream.py
@@ -132,6 +132,7 @@ def reorder_for_rfc_6555_section_5_4(targets):
 
 
 def format_host_port(host, port):
+    host = host.decode() if isinstance(host, bytes) else host
     if ":" in host:
         return "[{}]:{}".format(host, port)
     else:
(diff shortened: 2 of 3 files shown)


njsmith:
Nice catch!

The test failure is #624, nothing to worry about here.

The one thing that jumped out at me is that the interaction of hostnames and text encoding is stupidly complicated. So is a plain `.decode()` the right thing to do? On Python 2 that meant "use some random system-defined text encoding", which is definitely not right, but it looks like on Python 3 they made it so plain `.decode()` always uses utf-8, which is less wrong... but still probably wrong, because when a hostname is passed as `bytes`, we assume that it's already in the standard hostname "A-label" form, which is required to be all-ascii. Also, for hostnames we guarantee that `bytes` and `str` representations [behave identically if they're all ascii](https://github.com/python-trio/trio/blob/b926f5d0d251d4717026a061b2cd919f3679da35/trio/_socket.py#L228-L235), but for non-ascii characters idna gets involved and who knows what that even does.

So I think the right thing to do here is `.decode("ascii")`. Can you make that change?

And finally, can you add [a newsfragment](https://trio.readthedocs.io/en/latest/contributing.html#release-notes)? Just create a file called `newsfragments/633.bugfix.rst`, and put a brief description of what was fixed.

ncollins:
Thanks for the help. I've made those changes.

njsmith:
Awesome, thanks! Merging now :tada: :birthday: And, no pressure, but if you'd like to keep contributing then we'd love to have you, so I'm also sending you a github invite. You can read more about this in [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team).

PR #634, opened 2018-08-28
At startup, count how many frames Context.run adds to tracebacks

njsmith:
This is kind of ridiculous overkill, but I was thinking about how you
could do it, then wrote some code to experiment, then realized that
I had implemented it so we might as well use it...

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -64,6 +64,34 @@ _r = random.Random()
 INSTRUMENT_LOGGER = logging.getLogger("trio.abc.Instrument")
 
 
+# On 3.7+, Context.run() is implemented in C and doesn't show up in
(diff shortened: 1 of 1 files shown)


trio/_core/_run.py
```diff
@@ -64,6 +64,34 @@
...
+# currently implemented in Python and adds 1 frame to tracebacks. So this
+# function is a super-overkill version of "0 if sys.version_info >= (3, 7)
+# else 1". But if Context.run ever changes, we'll be ready!
+#
+# This can all be removed once we drop support for 3.6.
+def _count_context_run_tb_frames():
+    def function_with_unique_name_xyzzy():
+        1 / 0
+
+    ctx = copy_context()
```
belm0: I don't understand this part

belm0:
@njsmith please go ahead and merge

Issue #651, opened 2018-09-07
pytest 3.8.0 crashes before running the test suite because of a deprecation warning inside ipython


njsmith:
```
============================= test session starts ==============================
platform darwin -- Python 3.5.3, pytest-3.8.0, py-1.6.0, pluggy-0.7.1 -- /Users/jenkins/workspace/-594-SV5AYVYYLSFPHE6/trio/.venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/jenkins/workspace/-594-SV5AYVYYLSFPHE6/trio, inifile:
plugins: faulthandler-1.5.0, cov-2.6.0
collecting ... collected 434 items / 1 errors


==================================== ERRORS ====================================
 ERROR collecting .venv/lib/python3.5/site-packages/trio/_core/tests/test_multierror.py
../.venv/lib/python3.5/site-packages/trio/_core/tests/test_multierror.py:584: in <module>
    import IPython
../.venv/lib/python3.5/site-packages/IPython/__init__.py:55: in <module>
    from .terminal.embed import embed
../.venv/lib/python3.5/site-packages/IPython/terminal/embed.py:14: in <module>
    from IPython.core.magic import Magics, magics_class, line_magic
../.venv/lib/python3.5/site-packages/IPython/core/magic.py:20: in <module>
    from IPython.core import oinspect
../.venv/lib/python3.5/site-packages/IPython/core/oinspect.py:29: in <module>
    from IPython.lib.pretty import pretty
../.venv/lib/python3.5/site-packages/IPython/lib/pretty.py:91: in <module>
    from IPython.utils.signatures import signature
../.venv/lib/python3.5/site-packages/IPython/utils/signatures.py:9: in <module>
    DeprecationWarning, stacklevel=2)
E   DeprecationWarning: IPython.utils.signatures backport for Python 2 is deprecated in IPython 6, which only supports Python 3
```

For now we pinned to pytest 3.7.4: #650

njsmith:
Ah, so this is actually a combination of things:

* pytest 3.8.0 started displaying warnings that get triggered collection. This is a good thing, it means warnings are less likely to slip under the radar.
* in our CI, we run pytest with `-W error`. This is a good thing, it turns warnings into errors, so we notice them.

...but the combination is not so nice, because it turns out that IPython has been generating this meaningless warning all along, but now we're actually noticing it.

Maybe we can set up our warnings filter to explicitly ignore this warning?

njsmith:
[...]

PR #650, opened 2018-09-07
Pinned pytest version to 3.7.4

jmfrank63:
Due to this error we pin the pytest version to 3.7.4 to get CI to pass.
```
============================= test session starts ==============================
platform darwin -- Python 3.5.3, pytest-3.8.0, py-1.6.0, pluggy-0.7.1 -- /Users/jenkins/workspace/-594-SV5AYVYYLSFPHE6/trio/.venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/jenkins/workspace/-594-SV5AYVYYLSFPHE6/trio, inifile:
plugins: faulthandler-1.5.0, cov-2.6.0
collecting ... collected 434 items / 1 errors


==================================== ERRORS ====================================
 ERROR collecting .venv/lib/python3.5/site-packages/trio/_core/tests/test_multierror.py
../.venv/lib/python3.5/site-packages/trio/_core/tests/test_multierror.py:584: in <module>
    import IPython
../.venv/lib/python3.5/site-packages/IPython/__init__.py:55: in <module>
    from .terminal.embed import embed
../.venv/lib/python3.5/site-packages/IPython/terminal/embed.py:14: in <module>
    from IPython.core.magic import Magics, magics_class, line_magic
../.venv/lib/python3.5/site-packages/IPython/core/magic.py:20: in <module>
    from IPython.core import oinspect
../.venv/lib/python3.5/site-packages/IPython/core/oinspect.py:29: in <module>
    from IPython.lib.pretty import pretty
../.venv/lib/python3.5/site-packages/IPython/lib/pretty.py:91: in <module>
    from IPython.utils.signatures import signature
../.venv/lib/python3.5/site-packages/IPython/utils/signatures.py:9: in <module>
    DeprecationWarning, stacklevel=2)
E   DeprecationWarning: IPython.utils.signatures backport for Python 2 is deprecated in IPython 6, which only supports Python 3
```

PR #653, opened 2018-09-08
Filter out DeprecationWarning on IPython import in test suite

njsmith:
And unpin pytest again, since the pin in gh-650 was just a temporary
fix.

Fixes gh-651.


Zac-HD:
For posterity: IPython has [fixed this upstream](https://github.com/ipython/ipython/commit/29b75abae86bd186a91399518bbab6b9d36cf404) and I think prompt-toolkit has also been fixed (but by deleting the file, so it might just have been moved), so we won't need the filters forever.

Issue #543, opened 2018-06-06
Provide PEP 484-style type hints


njsmith:
Type hints are one of those things that's perpetually on my "it would be nice to learn about these someday" list, so I haven't done anything with them in trio myself, but people are interested in adding them (e.g. @Fuyukai, + [some discussion](https://gitter.im/python-trio/general?at=5b18452fb0d8ca5f1a852b7d)). Seems like a good idea to me.

Like I said, I don't know much about this. Looking at [PEP 561](https://www.python.org/dev/peps/pep-0561/), it sounds like the ideal is to use inline type hints where possible, and where it's not, to fall back to `.pyi` files that we maintain and ship alongside the corresponding `.py` files? Is that right?

That means that this interacts with #542: until we make our imports less magical, inline type hints probably aren't going to work. In some cases that might be fine and we should just use `.pyi` files, but it seems like right now there are lot of places where inline type hints *would* be the right approach, just we need to clean up the imports first, and that's a good idea anyway, so we might want to start there.

This is something we'd want to do incrementally. Maybe something like: (1) figure out how to get mypy to run on trio at all, (2) get that running in travis-ci so we don't regress, (3) start incrementally adding type hints, while making sure mypy can understand them, starting from the simplest/lowest-hanging fruit?

I know there's a growing body of experience with adding type hints to projects, but I'm not super familiar with it... if you know good articles please add links to them here :-). To get us started:

* How Zulip approached this: http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/
* [Clearer Code at Scale: Static types at Zulip and Dropbox](https://www.youtube.com/watch?v=0c46YHS3RY8) – talk that @gnprice gave at PyCon this year

smurfix:
IMHO `.pyi` files only make sense when we can't change the `.py` files, e.g. because they're shipped by somebody else. This doesn't apply to Trio.

See also https://github.com/dropbox/pyannotate for a way to auto-create (some) type annotations.

Fuyukai:
pyi files are good for describing the structure of magic objects that are created at runtime, as well.

smurfix:
Sure, but we don't actually have any of these AFAIK.

njsmith:
Oh right, something else we should keep in mind: PyPy doesn't yet support inline type annotation syntax. (I'm pretty sure they have it working in their 3.6 branch, so maybe this will become a non-issue soon. Or maybe not.)

Zac-HD:
[...]

PR #656, opened 2018-09-09
Initial work towards adding type hints

Zac-HD:
This PR contains some incremental work towards both #542 and #543.  It's definitely not finished, but we found on Hypothesis (and on #594!) that merging this kind of thing in small pieces helps keep it moving :smile: 

There are basically four parts to this:

- A fairly basic `mypy.ini`, with some notes on what to enable next
- Copying the #594 approach for more imports
- Completing pre-existing partial annotations
- Moving property setters to immediately follow getters
  (Mypy has special-case code to understand them but it relies on proximity)


njsmith:
Yeah, socket is going to be a problem. It's probably fairly straightforward to clean up the functions, but for the constants, the set exported by python changes unpredictably between versions and build configurations, so I don't know that there is any way to avoid the dynamic re-export. Maybe we can just annotate the constants we use, or something...

Anyway, this looks good. Do you think there should be a release note, or should we wait until it's further along?

njsmith:
[Staircase](https://en.wikipedia.org/wiki/L%27esprit_de_l%27escalier) review: this should have fixed tab completion for `trio.testing` and `trio._core`... it'd be nice to extend the jedi/pylint tests to make sure that stays true :-)

(not that tab completion is very important for `_core`, but why not)

PR #657, opened 2018-09-09
added newsfragment for issue 592 / 542

jmfrank63:
In the heat of the coding this had been forgotten.


njsmith:
Thanks!

>  Next would be to apply the same to `_core`?

I think @Zac-HD beat you to that in #656... let's discuss more in #542.

PR #659, opened 2018-09-11
Another increment for static analysis

Zac-HD:
I'm going in small stages to make review (and debugging) manageable, so:

- Internally get names as attributes of the stdlib `ssl` and `socket` modules, rather than by magic
- Remove a bunch of unused imports (mostly leftover from deleted code)
- Minor cleanups I found in the process

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -24,8 +24,7 @@ from . import _public
 from ._entry_queue import EntryQueue, TrioToken
 from ._exceptions import (TrioInternalError, RunFinishedError, Cancelled)
 from ._ki import (
-    LOCALS_KEY_KI_PROTECTION_ENABLED, currently_ki_protected, ki_manager,
-    enable_ki_protection
+    LOCALS_KEY_KI_PROTECTION_ENABLED, ki_manager, enable_ki_protection
 )
 from ._multierror import MultiError
 from ._traps import (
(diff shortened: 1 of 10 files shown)


trio/_core/_run.py
```diff
@@ -445,9 +444,8 @@ def _add_exc(self, exc):
         self.cancel_scope.cancel()
 
     def _check_nursery_closed(self):
-        if (
-            not self._nested_child_running and not self._children
-            and not self._pending_starts
+        if not any(
+            [self._nested_child_running, self._children, self._pending_starts]
```
sorcio: Do you think this is an improvement in readability? Consistency-wise I don't think Trio ever uses `any()` over a literal list of bools and I personally avoid that style because I feel it's less expressive.

Zac-HD: Personally I do find it reads more naturally this way, as starting with "not any" makes reading the three attributes easier. It's also closer to English grammar! 

However, if there's any doubt or disagreement I'd rather take it out and get the rest merged.

njsmith: Eh, I *guess* I'd do it the way I originally did it? But they're both fine and let's focus on more important things :-)

belm0:
regarding unused imports, does trio not having a lint check for this set up on the continuous build?

Zac-HD:
Belated cross-reference to #542 :sparkle:

PR #663, opened 2018-09-13
Run flake8 in CI, split out check script

Zac-HD:
This catches a whole pile of things that `yapf` doesn't, from trailing whitespace to unused variables or dodgy comparisons.  I've gone for a mixed strategy for existing issues - fixing or suppressing the less common issues so we'll catch any new instances, but globally ignoring a few pervasive lint errors where e.g. `flake8` and `yapf` disagree to favor the latter.

Of course all the import-related issues are behind the `--ignore=` flag, but we'll get #542 eventually!

Finally, I've split out the static analysis job into a new script `check.sh` which can be run locally for quick feedback.  This is also designed to make adding more analysis tools - e.g. `mypy` or `bandit` - as easy as possible.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -ex
+
+EXIT_STATUS=0
+
+# Autoformatter *first*, to avoid double-reporting errors
+yapf -rpd setup.py trio \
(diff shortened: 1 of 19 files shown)


pquentin:
Thanks!

PR #666, opened 2018-09-13
Tell Renovate not to update pinned yapf

Zac-HD:
Closes #665.  Also makes an attribute less magical.

For discussion: should we turn down the frequency of rennovate bot?  (*Can* we?  It looks to me like we can adjust hours of operation but not ask for batched updates...)


trio/_ssl.py
```diff
@@ -652,7 +652,7 @@ def _check_status(self):
                 # stream, and closing doesn't care about the state.
                 if (
                     self._https_compatible
-                    and isinstance(exc.__cause__, SSLEOFError)
```
Fuyukai: This seems out of place.

PR #667, opened 2018-09-13
Bandit - a security-oriented linter

Zac-HD:
[Bandit](https://bandit.readthedocs.io/en/latest/) is a security-focused static analysis tool, which looks for potential problems like use of `exec` or `subprocess`, or possibly binding to all interfaces (with `"0.0.0.0"`).  The argument for using it is basically that it's trivially easy to do so, and wouldn't we feel silly otherwise about avoidable security problems?  

That's not hypothetical by the way - we ran it against Hypothesis a while ago and it was fine, so we left it out of CI.  Subsequent versions allowed it to spot internal use of `pickle` in a cache file that may be shared from other users, and warn that deserialising pickle data gives the author arbitrary code execution.  That cache file is now written as JSON and Bandit runs in Hypothesis CI.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -13,6 +13,10 @@ flake8 trio/ \
     --ignore=D,E201,E402,E501,E722,E741,F401,F403,F405,F821,F822,W503 \
     || EXIT_STATUS=$?
 
+# Bandit is a security-oriented linter
+bandit --recursive --exclude=trio/tests --skip=B101,B311,B404 trio/ \
+    || EXIT_STATUS=$?
+
 # Finally, leave a really clear warning of any issues and exit
 if [ $EXIT_STATUS -ne 0 ]; then
     cat <<EOF
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -11,6 +11,7 @@ jedi           # for jedi code completion tests
 # Tools
 yapf == 0.22.0
 flake8
+bandit
 
 # https://github.com/python-trio/trio/pull/654#issuecomment-420518745
 typed_ast;python_version<"3.7" and implementation_name=="cpython"
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -10,11 +10,14 @@ astroid==2.0.4            # via pylint
 atomicwrites==1.2.1       # via pytest
 attrs==18.2.0             # via pytest
 backcall==0.1.0           # via ipython
+bandit==1.5.1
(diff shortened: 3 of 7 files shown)


Zac-HD:
There are four `# nosec` markers.  Two are in tests, and only exist because Bandit's `--exclude` option currently only allows for one path due to a bug.  Another is using `exec` to evaluate a statically generated string, which will go away as part of #542.  And the last is actually *preventing* the issue that Bandit is warning about, binding to `0.0.0.0`.

Zac-HD:
Ping @njsmith; what do you think of this?

njsmith:
Sorry for the slow response here!

I'm a bit dubious about the value of some of bandit's checks... flagging all use of `random` or `subprocess` as insecure? And I have a hard time imagining an accidental `exec` could slip through our code review! But, as a network library Trio is pretty security-sensitive, so I suppose we might as well give it a try – we can always turn it off again later if it turns out to be too annoying.

FYI, the `"0.0.0.0"` you noticed isn't actually preventing binding to all interfaces... in fact `trio.serve_tcp`'s default is to bind to all interfaces (which I think is pretty normal?). Rather, it's in the code for tests that want to connect to an open socket: if the socket is listening on `"0.0.0.0"`, then you can't do `connect(("0.0.0.0", port))`, so we do `connect(("127.0.0.1", port))` instead.

Would this be easier if moved `trio/_core/tests` into `trio/tests/_core`? We should do this anyway... (#274)

Can you add a comment next to the list of exclusions saying what `B101`, etc. are?

You'll need to re-run pip-compile to catch up with dependabot's updates.

I missed saying this on #663, but asking people to run `check.sh` locally is a bit awkward for Windows users, isn't it?

Zac-HD:
With some more space from the implementation, I'm actually inclined to leave this for now:

- It will be cleaner after we finish some known cleanup in test structure
- A future version of `bandit` will support `--quiet`, for no-output-unless-errors
- Trio's CI is a lot faster than Hypothesis and I'd like to keep it that way; and `bandit` hasn't actually found anything here.

So I'll probably revisit this in a few months but closing it for now.

webknjaz:
> and `bandit` hasn't actually found anything here.

I think that it's still useful for catching future violations, but it needs to be configured properly. In my opinion, bandit should be set up strictly and all “legal” violations would have to be consolidated in a dedicated module (and remain there) implementing safe wrappers.

PR #669, opened 2018-09-15
Document that the Nursery context manager blocks on exit

joernheissler:


njsmith:
Huh, there's something very weird happening here.

The text this adds is basically saying the same thing as the `open_nursery` docstring. This text is probably clearer :-). But still, we shouldn't say it twice in a row, so something should change. (I guess this should go into the docstring?)

But... for some reason sphinx is not showing the docstring at all: https://trio.readthedocs.io/en/latest/reference-core.html#the-nursery-api

Why not? In the REPL, `trio.open_nursery` seems to have a docstring just fine, and sphinx should be complaining if it can't find `trio.open_nursery` at all...

njsmith:
Oh I see, I had an old branch checked out. The docstring seems to have been accidentally lost in 0bc12ef585913daf050e5707624785b1f6a230a1.

Can you move this into the docstring instead, so that it's available in both the online docs and interactively at the REPL?

njsmith:

PR #681, opened 2018-09-24
Fixes 'invalid escape sequence \s' deprecation warning on python 3.7

jmfrank63:
Some strings contain content that needs these strings now to be raw strings.
This PR fixes this.

### DIFF
--- a/trio/_core/_ki.py
+++ b/trio/_core/_ki.py
@@ -97,7 +97,7 @@ def ki_protection_enabled(frame):
 
 
 def currently_ki_protected():
-    """Check whether the calling code has :exc:`KeyboardInterrupt` protection
+    r"""Check whether the calling code has :exc:`KeyboardInterrupt` protection
     enabled.
 
     It's surprisingly easy to think that one's :exc:`KeyboardInterrupt`
--- a/trio/_highlevel_serve_listeners.py
+++ b/trio/_highlevel_serve_listeners.py
@@ -57,7 +57,7 @@ async def serve_listeners(
     handler_nursery=None,
     task_status=trio.TASK_STATUS_IGNORED
 ):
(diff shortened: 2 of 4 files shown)


njsmith:
Fair enough! How did you catch these? Do we need to adjust something in our CI to catch them and stop them from sneaking back in over time?

jmfrank63:
To be honest I just got them on my AWS xenial machine. I try to build
slowly step by step a replica of the CI in order to learn how this works. I
am not sure if raw strings is the answer as not all comments are having
this issue. But for the collections.abc this PR is a good branch to debug
as there are no code changes.

njsmith:
I'm still puzzled about why you see the warnings on 3.7, but when the CI runs the tests on 3.7 with `-W error` then it doesn't produce any errors. But, oh well, we still should fix this I guess... and if/when they do actually make `\s` an error, I guess we'll definitely discover any new ones then!

This needs some kind of rebase or empty commit or something to trigger a CI re-run though, now that #683 is merged.

jmfrank63:
Sure, let me just do some work for my employer ;-), I'll take care of that later.

jmfrank63:
I think I know the difference now, I am not on xenial but on bionic. Nevertheless this might give us an idea what to expect in the future.

pquentin:
Perhaps using `\\s` instead of `\s` is a cleaner way of fixing the issue? It would feel less like a workaround and more like an actual fix, I think.

joernheissler:
https://www.python.org/dev/peps/pep-0257/

> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""".

pquentin:
Ah, nice find! Thanks, merging.

Issue #33, opened 2017-01-23
Which clock should we use on Windows?


njsmith:
Right now we use `time.monotonic` everywhere. This is a bit problematic on Windows, where `time.monotonic` is `GetTickCount64`, which has ~15 ms resolution. The other option is `QueryPerformanceCounter`, which has much higher resolution, is also monotonic, and is exposed as `time.perf_counter` (checked with `time.get_clock_info`).

The reason for this appears to be that `QueryPerformanceCounter` has had a troubled past: https://www.python.org/dev/peps/pep-0418/#windows-queryperformancecounter

But all these issues are like "Virtualbox had a bug that was fixed 6 years ago" or "Windows XP is flaky" which is probably true but irrelevant since we don't support it – it's not clear that any of them apply anymore.

Advantages of using a higher-precision clock:

* Right now if there's just one task running and it does a sleep for t seconds, the actual timeout passed to the underlying system call will be `(t + current_time()) - current_time()`, which can be pretty imprecise depending on whether the clock ticks over between the two calls. OTOH I don't know what the actual resolution of our sleeping syscall is (currently `GetQueuedCompletionStatusEx`), or whether anyone cares about millisecond accurate sleeps.

* I've had two bugs already with tests that assumed that time always, like, passes. These were trivial (like replacing a `<` with a `<=`), but it's always annoying to have the tests pass locally then fail on appveyor.

* If we implement a fancier scheduling system (#32) then we'll *definitely* need better than 15 ms precision to measure tasks running. (Though there's no reason that the clock we use for that has to match the clock we use for deadlines.)

njsmith:
See also [Microsoft's official guidance on clocks](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx)

njsmith:
@tkerwin [reported in chat](https://gitter.im/python-trio/general?at=5b92fea8728ddf02827675a6) that they have a program that misbehaves on windows if trio uses `time.monotonic`, but works correctly if trio uses `time.perf_counter`. This seems like a pretty good argument for switching over to me.

[...]

PR #682, opened 2018-09-25
Change to high-resolution clock per #33

tkerwin:
Per the discussion in issue #33 this changes the clock to use time.perf_counter() rather than time.monotonic() on all platforms.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -198,7 +198,7 @@ Time and clocks
 
 Every call to :func:`run` has an associated clock.
 
-By default, trio uses an unspecified monotonic clock, but this can be
+By default, trio uses the `time.perf_counter` clock, but this can be
 changed by passing a custom clock object to :func:`run` (e.g. for
 testing).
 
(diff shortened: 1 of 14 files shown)


trio/_core/_run.py
```diff
@@ -1233,7 +1236,7 @@ def run(
       args: Positional arguments to be passed to *async_fn*. If you need to
           pass keyword arguments, then use :func:`functools.partial`.
 
-      clock: ``None`` to use the default system-specific monotonic clock;
```
njsmith: This is just referring to the general concept of a "monotonic clock", so should stay the same.

docs/source/reference-core.rst
```diff
@@ -198,17 +198,17 @@ Time and clocks
 
 Every call to :func:`run` has an associated clock.
 
-By default, trio uses an unspecified monotonic clock, but this can be
```
njsmith: This should stay the same – it's talking about the general guarantees we make about the clock, and we still don't guarantee anything except that it has the property of monotonicity. (This is why we can change from `time.monotonic` to `time.perf_counter` :-).)

newsfragments/33.feature.rst
```diff
@@ -0,0 +1 @@
+The Trio main loop now uses a higher-precision clock by default.
```
njsmith: This should say "on Windows", since that's the only platform affected.

njsmith: My reasoning is: we use a different *function* on all platforms, but on most platforms this doesn't change behavior at all, so it's not of interest to end users. The only place where *precision* changes is on Windows.

njsmith:
@tkerwin I edited the newsfragment again, see what you think?

tkerwin:
Yeah, that's good.

njsmith:

PR #683, opened 2018-09-26
Don't use asyncio.sleep as a function that waits on an asyncio.Future

njsmith:
In this PR:

  https://github.com/python/cpython/pull/9415

asyncio.sleep switched to requiring a running asyncio event loop. This
is all very sensible, but broke our test that checks what happens when
some accidentally yields an asyncio.Future to the Trio run loop.
Switch the test to waiting on a Future directly.


njsmith:
I guess I'll probably also have to bump the pickleshare pin here to pick up https://github.com/pickleshare/pickleshare/pull/28 ...

Issue #671, opened 2018-09-17
Ideas for linting trio programs


njsmith:
I haven't given up on catching missing `await`s at runtime (see #495), but another strategy that could be useful is to have a linter/static analyzer that catches them for you. In fact, you really want both, for all the reasons runtime + static checks are useful for other kinds of errors: runtime checks can be 100% reliable when the error actually happens, and are especially useful for beginners who can't be expected to set up fancy linting tools; static checks can catch mistakes early, even in untested code-paths.

Examples of this being a real issue:

* https://github.com/python/cpython/pull/9168#issuecomment-420361009
* @wgwz gets bitten: https://gitter.im/python-trio/general?at=5b9dde717dca30650309ca3e
* Yury says that specifically when a function is alone on a line, he accidentally writes `foo()` instead of `await foo()` about once a month.

I know absolutely nothing about static analysis, so I have no idea how to actually implement a lint for this, but let's have an issue to discuss and maybe we'll figure it out together.

In theory, this should be pretty straightforward, because in Trio, the property to enforce is: any function call which returns an awaitable, should have an `await` to immediately consume that awaitable. This makes things *way* simpler than asyncio, where it's common to create coroutine objects and then pass them somewhere else, and sometimes this is what you want (passing them into `asyncio.run` or `asyncio.create_task` or `asyncio.gather`) and other times its an accident because you forgot an `await`. So a lint for asyncio would require some pretty sophisticated data flow inference and knowledge of different APIs; for trio, once we've figured out which calls are to async functions/methods, the check itself requires only superficial knowledge of the surface AST. OTOH, this may make it more difficult to convince tools like mypy or pylint to include our check, because it won't work for asyncio users; and, if they do include it there will need to be some kind of configuration to turn it on or off. Maybe it would need to be a plugin, at least to start out.

I got a bit [nerd-sniped](https://xkcd.com/356/) by this last night. Findings so far:

# pylint

Pylint has [docs on how to add a new check](http://pylint.pycqa.org/en/latest/how_tos/custom_checkers.html), and an [official plugin API](http://pylint.pycqa.org/en/latest/how_tos/plugins.html), so we could distribute a checker as a third-party package. And, (through [astroid](http://astroid.readthedocs.io/)), it has some reasonably sophisticated APIs to go from a call back to the definition of the function/method being called. So playing around, I came up with this: https://gist.github.com/njsmith/0ce8c661684ba4514b9ea4aac6182f3c

[...]

PR #700, opened 2018-09-30
Add pylint missing-await checker

belm0:
pylint support for #671 

TODO:
  * [x] fix async generator false positive
  * [x] tests
  * [ ] settle error name(s) and code numbers

This runs fine on trio codebase and my own codebase.  I haven't looked hard for false negatives.

Run on trio codebase:
```
PYTHONPATH=. python3 -m pylint --load-plugins=trio.testing.pylint-missing-await --disable=all --enable=missing-await trio/
```

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -14,7 +14,7 @@ flake8 trio/ \
     || EXIT_STATUS=$?
 
 # pylint to at least cover common omissions of "await" and "async" qualifiers
-find trio/ -name '*.py' |
+find trio/ -not -name 'test_pylint_missing_await.py' -name '*.py' |
     PYTHONPATH=. xargs pylint --load-plugins=trio.testing.pylint-missing-await \
     --disable=all --enable=missing-await,not-an-iterable,not-context-manager \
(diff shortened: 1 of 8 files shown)


njsmith:
It looks like there's a convention of having standalone `pylint-{projectname}` projects, e.g.:

* https://github.com/PyCQA/pylint-celery
* https://github.com/PyCQA/pylint-django
* https://github.com/jschaf/pylint-flask

Maybe we should follow, with a `pytest-trio` (that would include this, and potentially other lints if we think of any)?

In other news, it looks like py38-dev and astroid have become incompatible, so our CI is currently broken... maybe we should just disable the static analysis tests when `sys.version_info.releaselevel != "final"`? This isn't the first time we've had trouble with this...

belm0:
fixed false positive from async generators

belm0:
* added tests
  * missing-await run against trio code from `check.sh`

Fuyukai:
This is like five years old and the project switched to ruff since. Closing.

Issue #711, opened 2018-10-02
Broken coverage on 3.8-dev


njsmith:
So apparently codecov's weird behavior for the last few weeks is caused by an interaction between coverage.py and this change in 3.8-dev: https://github.com/python/cpython/pull/8774

Another example: https://github.com/marrink-lab/vermouth-martinize/issues/132

The coverage devs are aware of it: https://github.com/nedbat/coveragepy/issues/707#issuecomment-426455490

We should maybe disable coverage tracking on 3.8-dev builds until this gets fixed.

PR #717, opened 2018-10-04
Temporarily disable coverage reporting on 3.8-dev

njsmith:
Fixes gh-711

### DIFF
--- a/ci/travis.sh
+++ b/ci/travis.sh
@@ -75,5 +75,10 @@ else
     INSTALLDIR=$(python -c "import os, trio; print(os.path.dirname(trio.__file__))")
     pytest -W error -ra --run-slow --faulthandler-timeout=60 ${INSTALLDIR} --cov="$INSTALLDIR" --cov-config=../.coveragerc --verbose
 
-    bash <(curl -s https://codecov.io/bash)
+    # Disable coverage on 3.8-dev, at least until it's fixed (or a1 comes out):
+    #   https://github.com/python-trio/trio/issues/711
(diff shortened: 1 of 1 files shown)


njsmith:
Oh ugh, I bet the reason codecov hasn't commented here is because it knows that this 3.8-dev job ran, and is waiting for its coverage report to arrive before commenting.

njsmith:
Yeah, see https://docs.codecov.io/v5.0.0/docs/merging-reports#section-why-did-i-not-receive-a-notification-yet-

> Why did I not receive a notification yet?
> If CI is successful:
> * check if the reports failed to upload to Codecov.

njsmith:
Maybe we can somehow send an empty report?

njsmith:
So apparently if the codecov uploader doesn't find a `.coverage` file, then it doesn't upload anything (even to say "no coverage here"). (In particular: its log output on travis doesn't mention uploading a report, the way it does normally; and if you look at https://codecov.io/gh/python-trio/trio/commit/1f60eac7b0058a1af92709b7c5c431fae769971f/build there's no 3.8-dev box present.

OTOH, I guess codecov may have somehow magically figured out that the upload isn't coming? This time it did update promptly. And thinking about it, the `CHECK_FORMATTING` and `CHECK_DOCS` builds have never uploaded coverage, so it must have some kind of cleverness to handle that.... maybe it just got confused because it was the first build where 3.8-dev didn't upload coverage? Or just a random glitch that happened at a confusing moment?

pquentin:
Surprising, but it does seem to work! Thanks.

PR #725, opened 2018-10-08
ci: remove unneeded pyup references

pquentin:
We only use dependabot. Or at least I think so? I don't know how to check that. So this pull request is my way of asking the question.


njsmith:
Yeah, I added that while I was experimenting with the different options, then never cleaned it up again after disabling pyup.

PR #727, opened 2018-10-09
Closes #726 - Documentation should have previous and next button on t…

### LINKED ISSUE #726: Documentation should have previous and next button on top as well as bottom.
I have been going through the trio documentation and I have noticed that the page is really big and if someone just wants to get from one page to either previous or next then he/she will have to scroll down to the bottom of the page then click the next or previous button.

It would be nice to have previous and next button on top as well as on bottom so that anyone can navigate through the documentation more easily.

RatanShreshtha:
…op as well as bottom.

Updated configuration for sphinx rtd theme, next/previous navigation will be available on top and bottom.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -157,6 +157,7 @@ html_theme_options = {
     # versions/settings...
     "navigation_depth": 4,
     "logo_only": True,
+    'prev_next_buttons_location': 'both'
 }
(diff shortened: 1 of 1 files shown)


pquentin:
This looks like this:

top

...

bottom

Thanks! Merging.

PR #730, opened 2018-10-11
Added typechecking block for static analysis tools

jmfrank63:


njsmith:
The first commit looks good. The second commit doesn't make much sense to me. The old code was fine (reimporting `IPPROTO_IPV6` when it's already defined is harmless, and doing it unconditionally reduces the number of different configurations we have to test). And, `if not IPPROTO_IPV6` checks whether the value of that variable is falsey, not whether it's defined in the first place :-)

Still needs a test, like the existing ones in `test_exports`

trio/socket.py
```diff
@@ -9,6 +9,62 @@
 from . import _socket
 import sys as _sys
 
+# expose all uppercase names from standardlib socket to trio.socket
+import socket as _stdlib_socket
+
+try:
+    from typing import TYPE_CHECKING
+except ImportError:
+    TYPE_CHECKING = True  # Not precise but only for python 3.5.0
```
njsmith: We should actually set it to `False`! When we're running on the python interpreter, we don't want to execute the code inside the `if` block. And the static tools like mypy have hard-coded logic that when they see this variable, they act like it's `True` regardless of what it's actually set to.

trio/socket.py
```diff
@@ -9,6 +9,62 @@
...
+if TYPE_CHECKING:
```
njsmith: We should add a `# pragma: no cover` in this line. (I guess we should also consider adding it to the list of patterns that automatically trigger no-cover-mode in our `.coveragerc`? That's not urgent though, we can wait until we see how often we use this pattern.)

trio/socket.py
```diff
@@ -9,6 +9,62 @@
...
+        IPV6_RTHDR_TYPE_0, IPV6_TCLASS, TCP_NODELAY, TCP_MAXSEG, TCP_KEEPINTVL,
+        TCP_KEEPCNT, TCP_FASTOPEN, TCP_NOTSENT_LOWAT, EAI_ADDRFAMILY,
+        EAI_AGAIN, EAI_BADFLAGS, EAI_FAIL, EAI_FAMILY, EAI_MEMORY, EAI_NODATA,
+        EAI_NONAME, EAI_OVERFLOW, EAI_SERVICE, EAI_SOCKTYPE, EAI_SYSTEM,
+        EAI_BADHINTS, EAI_PROTOCOL, EAI_MAX, AI_PASSIVE, AI_CANONNAME,
+        AI_NUMERICHOST, AI_NUMERICSERV, AI_MASK, AI_ALL, AI_V4MAPPED_CFG,
+        AI_ADDRCONFIG, AI_V4MAPPED, AI_DEFAULT, NI_MAXHOST, NI_MAXSERV,
+        NI_NOFQDN, NI_NUMERICHOST, NI_NAMEREQD, NI_NUMERICSERV, NI_DGRAM,
+        SHUT_RD, SHUT_WR, SHUT_RDWR, EBADF, EAGAIN, EWOULDBLOCK, _LOCALHOST,
+        _LOCALHOST_V6, _GLOBAL_DEFAULT_TIMEOUT
```
njsmith: Oh heh, we actually shouldn't be reexporting these names that start with underscores! That's actually a bug in our current code... I didn't realize we were doing that.

I guess they should be removed here, and also the reexport loop should be rewritten to loop over `socket.__all__` instead of `socket.__dict__`.

trio/socket.py
```diff
@@ -9,6 +9,62 @@
 from . import _socket
 import sys as _sys
 
+# expose all uppercase names from standardlib socket to trio.socket
+import socket as _stdlib_socket
```
njsmith: Let's move this back down next to the loop where we use it.

trio/socket.py
```diff
@@ -44,24 +100,17 @@
...
     # Do not import for windows platform
     # (you can still get it from stdlib socket, of course, if you want it)
     del SO_REUSEADDR
 
 # get names used by trio that we define on our own
-from ._socket import IPPROTO_IPV6
+try:
+    IPPROTO_IPV6
+except NameError:
+    from ._socket import IPPROTO_IPV6
```
njsmith: Again... this code should work fine, but the old code also worked fine, and was 4x shorter and had fewer branches to test. Why change it?

jmfrank63:
Sorry, just saw your comments, had been sleeping ;-)
The problem I run into is python 3.5.0
https://github.com/python/mypy/issues/4714 shows why. I tried to workaround the problem but haven't had a chance to look at the 3.5.0 typing.py source code. The blind attempt failed.
Maybe we should do the same and drop 3.5.0 compatibility. Just 3.5.1 upwards. But that is something that very likely needs a discussion. I will try to get hold of the source code and see if I can find a workaround.

trio/socket.py
jmfrank63: Yes you are right, but 3.5.0 does not have that variable. Either we find a workaround or we have to drop 3.5.0 compatibility as https://github.com/python/mypy/issues/4714 did.

trio/socket.py
jmfrank63: I generated the list by printing. But I am not sure if I got all of them, because I created the list on a Mac. Windows and Linux will likely to generate a slightly different list. And I just created that list using python 3.7.0 Other python versions might also lead to different results? Could spin up some machines but to loop over all OS with all minor python versions is a lot of legwork.

trio/socket.py
jmfrank63: I am not sure I understood. I wrapped it because the CI pipeline failed on the double declaration. 
What should I do instead?

trio/socket.py
jmfrank63: Done!

trio/socket.py
jmfrank63: Ah, and I removed the underscore starting names.

jmfrank63:
I guess I have to learn how to read the coverage reports. It does not show any files so I don't know what to do. Also the tests have to be extended. But before I start this I'd like to know if the workaround for the missing `TYPE_CHECKING` variable in 3.5.0 is acceptable? The way this works is in any later module of `typing.py` `TYPE_CHECKING` is set to false before the `IO` class. I therefore took the nearest variable definition ( `_G_base`) touched by runtime code to set the `TYPE_CHECKING` back to false. There is a slight difference however, the classes `Generator` and `NamedTuple` are defined while `TYPE_CHECKING` is already false. Currently this doesn't matter, but later on this could matter if one of these classes are used with annotations.

jmfrank63:
Hehe, the whole idea is not working as it would require true parallel code to run. I guess setting `TYPE_CHECKING` simply to `false` is probably our best option. Any suggestions?

trio/socket.py
njsmith: > I generated the list by printing. But I am not sure if I got all of them...

After you add the test to make sure jedi/pylint see all the names, then the CI will tell us if you missed any names. And if you write the test right, you can even get it to tell you what names you're missing. Useful trick: you can freely `print` stuff in tests; pytest hides the output unless the test fails, in which case it shows it. So you can write a test that prints a list of all the missing items before it fails.

njsmith:
I tried checking this out and applying this patch:

```diff
--- a/trio/socket.py
+++ b/trio/socket.py
@@ -10,7 +10,7 @@ from . import _socket
 import sys as _sys
 
 try:
-    from ._socket import (
+    from socket import (
         CMSG_LEN, CMSG_SPACE, CAPI, AF_UNSPEC, AF_INET, AF_UNIX, AF_IPX,
         AF_APPLETALK, AF_INET6, AF_ROUTE, AF_LINK, AF_SNA, PF_SYSTEM,
         AF_SYSTEM, SOCK_STREAM, SOCK_DGRAM, SOCK_RAW, SOCK_SEQPACKET, SOCK_RDM,
```

...and the jedi test still seemed to work? (Well, it failed, because I'm on Linux, but jedi seems to be seeing all the symbols that are actually mentioned in `trio/socket.py`.)

What do you get?

jmfrank63:
I tried several ways of putting the import from sockets into try/catch blocks. The only way I was able to fool jedi was doing the import in _socket.py and then import from there.
But I did not leave the type checking out. Will conduct some more tests through the day. The linux symbols seem to be really long, I wasn't aware how different they are just between Mac and Linux. Is this a System V vs BSD difference?

njsmith:
> The linux symbols seem to be really long, I wasn't aware how different they are just between Mac and Linux. Is this a System V vs BSD difference?

Linux isn't any more a System V-derived system than it's a BSD-derived system, and I think most of those differences have smoothed out these days with POSIX and all. I think it's just a difference between how many people are working on networking on Linux versus other systems...

Suggestion: put something in the test like:

```python
missing_symbols = trio_set - jedi_set
if missing_symbols:
    print("Need to add the following symbols to trio/socket.py:")
    from textwrap import TextWrapper
    wrapper = TextWrapper(initial_indent=" " * 8, width=79)
    print(wrapper.wrap(", ".join(sorted(missing_symbols)))
    pytest.fail()
```

and then you can copy/paste directly from the test output into the source.

jmfrank63:
> > The linux symbols seem to be really long, I wasn't aware how different they are just between Mac and Linux. Is this a System V vs BSD difference?
> 
> Linux isn't any more a System V-derived system than it's a BSD-derived system, and I think most of those differences have smoothed out these days with POSIX and all. I think it's just a difference between how many people are working on networking on Linux versus other systems...
> 
> Suggestion: put something in the test like:
> 
> ```python
> missing_symbols = trio_set - jedi_set
> if missing_symbols:
>     print("Need to add the following symbols to trio/socket.py:")
>     from textwrap import TextWrapper
>     wrapper = TextWrapper(initial_indent=" " * 8, width=79)
>     print(wrapper.wrap(", ".join(sorted(missing_symbols)))
>     pytest.fail()
> ```
> 
> and then you can copy/paste directly from the test output into the source.

In my usual blindness I completely missed out your suggestion. I used the Travis and AppVeyor output that seemed to have done the job.

njsmith:
Yeah, codecov is a bit mysterious until you get used to it... the most reliable way I know to figure out what's going on with a PR is:

1. Click the "Continue to review full report at Codecov" link
1. Click on the "Files" tab (not anything else! You would think "diff" or "changes" are relevant, but they are just there to confuse you)
1. In the "files" tab, there's a tabular view of the source tree, with columns labeled "green dot", "yellow dot", "red dot". The numbers in those columns show you how many lines the PR caused to become covered (green), partially covered (yellow) or uncovered (red). I look for "+" entries in the yellow/red columns, and drill down through the source tree to find the individual files. In this case, I click on "trio", and then "tests", and then "test_exports.py".
1. Now I'm looking at [the actual file](https://codecov.io/gh/python-trio/trio/pull/730/src/trio/tests/test_exports.py?before=trio/tests/test_exports.py). At the top there is a message saying "Awesome! No coverage changes found when comparing this file." I don't know why codecov says this, because it is a blatant lie – we just got here because codecov said there were coverage changes in this file.
1. Scroll through the file looking for red/yellow lines. Here there are two sets of red lines, both of which look like:

```python
except NotImplementedError:  # pragma : no cover
        pytest.skip("jedi does not yet support {}".format(sys.version))
```

So... those are the lines that codecov is complaining about being uncovered. What's going on? It's not a surprise that those lines didn't get executed... it just means jedi didn't raise `NotImplementedError` in any of the configurations we currently test. But we have this `pragma` here that's supposed to tell coverage.py that we don't care about those lines. Yet they're still being reporting as uncovered, so our pragma isn't working.

The problem is that coverage.py's check for the pragma is very simple, and just does a literal string match. It's looking for `# pragma: no cover`, but here we have `# pragma : no cover` with an extra space before the colon, so it doesn't match

Solution: remove extra space before the colon, in both of the places where it happens :-)

jmfrank63:
Nice, now all the symbols are there and pylint only complains about the multiple star import ;-)
I was actually tempted to put the dynamic import into the except block. Should the import for some reason succeed we have all the symbols via this way, otherwise we do the dynamic import.
What do you think?

trio/socket.py
```diff
@@ -9,6 +9,100 @@
...
+import socket as _stdlib_socket
+
+globals().update(
+    {
+        _name: _value
+        for (_name, _value) in _stdlib_socket.__dict__.items()
+        if _name.isupper() and not _name.startswith('_')
+    }
+)
+
```
jab: You can pass a sequence of pairs (including a generator comprehension) directly to `dict.update()`, no need to build up this entire dictionary in memory before throwing it away, and it's a bit more concise to boot:
```suggestion
globals().update(
    (_name, _value)
    for (_name, _value) in _stdlib_socket.__dict__.items()
    if _name.isupper() and not _name.startswith('_')
)
```

njsmith:
> I was actually tempted to put the dynamic import into the except block. Should the import for some reason succeed we have all the symbols via this way, otherwise we do the dynamic import.
What do you think?

I think let's keep it the way it is now :-).

This is already really "clever" and confusing to understand, so we want to try to minimize it. Hopefully at some point we'll be able to switch back to `if TYPE_CHECKING`, which will make it more obvious, and if so we'll need to pull the imports back out of the `except` anyway. And anyway, if the `import` did succeed, it would break our test – if any new symbols appear in the socket module, that aren't in our list, then we won't notice, because we'll end up comparing out list of static imports against our list of static imports.

jmfrank63:
Yeah right, so I will take a look at one of the next modules to rework. I think we have done all we can and all checks are in including pylint, unless you want a test for mypy as well?

njsmith:
Yeah, we could keep nitpicking little things but basically this seems workable, and we can always tweak stuff later if it causes issues.

Codecov is complaining about +2 partial lines, but this seems to be a glitch; if I go looking through its report I can't find them anywhere.

Going to go ahead and merge. Thanks! And sorry for the slow review.

Issue #649, opened 2018-09-06
Should we have a way to let some other coroutine runner take temporary control of a task?


njsmith:
So there's a cute hack, which as far as I know @agronholm was the first to [discover](https://asyncio-extras.readthedocs.io/en/latest/#asyncio_extras.threads.threadpool), where you can use an `async with` block to temporarily switch between coroutine runners. For example:

```python
async def somefunc():
    await do_something_in_main_thread()
    async with in_worker_thread():
        do_something_blocking()
    await do_something_in_main_thread()
```

Another possible application would be to allow trio-asyncio to [switch back and forth between trio mode and asyncio mode within a single function](https://gitter.im/python-trio/general?at=5b90405ac2bd5d117a24e186).

This is kind of brain melting, but the implementation is surprisingly simple: a task's current execution is represented by a coroutine object. So, you suspend the coroutine, and from Trio's point of view it becomes "blocked"; then you take that coroutine object and give it to another coroutine runner to iterate for a while. Later, it gets handed back and Trio "resumes" it.

You can't actually implement something like this right now though, using Trio's public API. You can suspend a task, and get the coroutine object... but the coroutine is suspended inside `wait_task_rescheduled`, so to resume you need to send in the kind of value that `wait_task_rescheduled` is expecting, and that's not part of the public API. Similarly, to resume a task, `trio.hazmat.reschedule` will send in some kind of value, but you're not supposed to know how to interpret it – only `wait_task_rescheduled` knows that, so to resume the task you have to somehow get back into `wait_task_rescheduled`.

But it wouldn't be terribly hard to add a new suspend/resume API just for this. In `trio.hazmat`, obviously! This is extremely hazmat-y. But that's why `trio.hazmat` exists, so we can expose hazmat-y things. Something like:

```python
@types.coroutine
def detach_task(abort_fn):
    yield some_magic_value(abort_fn)

@types.coroutine
def reattach_task(yield_value):
    got = yield yield_value
    do_whatever_wait_task_rescheduled_would_do_with(got)
```

[...]

PR #732, opened 2018-10-13
Support handing off coroutines between Trio and other runners

njsmith:
I'm not sure if this is a good idea, but I guess we'll give it a try
and see how it goes.

Closes gh-649.

### DIFF
--- a/trio/_core/_traps.py
+++ b/trio/_core/_traps.py
@@ -1,12 +1,13 @@
-# These are the only 2 functions that ever yield back to the task runner.
+# These are the only functions that ever yield back to the task runner.
 
 import types
 import enum
 from functools import wraps
 
 import attr
+import outcome
 
-__all__ = ["cancel_shielded_checkpoint", "Abort", "wait_task_rescheduled"]
+from . import _run
 
 
 # Helper for the bottommost 'yield'. You can't use 'yield' inside an async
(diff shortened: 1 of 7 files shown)


trio/_core/tests/test_run.py
```diff
@@ -2015,3 +2016,108 @@ def test_sniffio_integration():
...
+        nonlocal unrelated_task
+        unrelated_task = _core.current_task()
+
+    async def reattachable_coroutine():
+        await sleep(0)
+
+        nonlocal task
+        task = _core.current_task()
+
+        def abort_fn(_):  # pragma: no cover
```
oremanj: Might be worth adding a test that cancels a detached coroutine, to validate that the abort_fn does actually get called?

trio/_core/_traps.py
```diff
@@ -163,3 +164,107 @@ def abort(inner_raise_cancel):
...
+
+    Args:
+      final_outcome (outcome.Outcome): Trio acts as if the current task exited
+          with the given return value or exception.
+
+    Returns or raises whatever value or exception the new coroutine runner
+    uses to resume the coroutine.
+
+    """
+    if _run.current_task().child_nurseries:
```
oremanj: Should this also check whether the detaching task is in any cancel scopes? I think task_exited will remove it from the scopes' task lists, but exiting the cancel scope in the now-detached task will then create some sort of confusing error.

trio/_core/tests/test_run.py
njsmith: Done

trio/_core/_traps.py
njsmith: I thought about that, but we don't really have an easy way to check if the detached task has any `with <cancel scope>` blocks open. We can see whether there are any cancel scopes active, but that includes inherited scopes, that don't correspond to a `with` block inside the exiting task. And, if someone does detach inside a `with <cancel scope>` block, then it's true that they'll eventually get some kind of explosion in the `__exit__` method, but I don't think this will corrupt Trio's internal state; it'll just be a regular exception.

For nurseries OTOH, it's easy to check for, and killing a task with active nurseries inside would break Trio's internal invariants.

Of course there are lots of ways to misuse `trio.hazmat` and these functions in particular to break Trio's internal invariants, but since the check was easy I figured why not.

oremanj: Makes sense! I hadn't checked in detail as to whether the cancel scope check would be feasible, and my memory of the bookkeeping in CancelScope was incorrect.

njsmith:
Guess I'll go ahead and merge this then. The real test will be to see how well it works in trio-asyncio...

PR #734, opened 2018-10-13
Belatedly fix a bunch of small issues in the v0.9.0 release note

njsmith:


njsmith:
Travis failure is just a job that failed to launch, just needs restarting. I can do it when I'm at a computer with credentials.

We did originally use fan-in/fan-out terminology, but then I did more research and find that that's mostly used for when you send duplicates of the same info to different places, so i switched it everywhere. Except i missed the release note. Not a big deal, but since I was fixing the other garbled stuff anyway I figured I'd get it too.

njsmith:
> Yeah, I was aware of the fan-in/fan-out -> producers/consumers change. I tried to be useful by reviewing quickly, but I was not: if I was going to leave a review, I should have checked those two things myself. Sorry about that. I'll try to improve on this front.

Hey, seriously, don't worry about it :-). There is literally nothing wrong with a review that's a bit quick and sloppy – it often catches stuff, it helps us realize which things are unclear to a hurried reader and pushes us to clarify them, it's way better than no review at all. If we had infinite energy and attention we could just get the changes right in the first place and wouldn't need review! The whole point of review is that we're messy and fallible and a few pairs of messy fallible eyes are better than one.

pquentin:
Thanks :)

Issue #735, opened 2018-10-13
Give a useful error if someone calls SSLStream.selected_alpn_protocol() before the TLS handshake


pgjones:
With this minimal example, trio 0.8.0, Python 3.7,
```python
import ssl
import trio

ssl_context = trio.ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
ssl_context.set_ciphers('ECDHE+AESGCM')
ssl_context.options |= (
    ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3 | ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1
)  # RFC 7540 Section 9.2: MUST be TLS >=1.2
ssl_context.options |= ssl.OP_NO_COMPRESSION  # RFC 7540 Section 9.2.1: MUST disable compression
ssl_context.set_alpn_protocols(['h2', 'http/1.1'])
ssl_context.load_cert_chain(certfile='cert.pem', keyfile='key.pem')

async def serve(stream):
    print(stream.selected_alpn_protocol())

async def main():
    await trio.serve_ssl_over_tcp(
        serve, ssl_context=ssl_context, host='localhost', port=5000,
        https_compatible=True,
    )

trio.run(main)
```
Using this `curl --http2 -v https://localhost:5000` command I would expect the program to print `h2`, but it prints `None`. I'm new to trio so it is equally likely I'm doing something wrong here...

njsmith:
Oh, I think I know why... `SSLStream` doesn't perform the handshake immediately, it waits until you try to send/receive data, or call `do_handshake`. (This is kind of necessary because you need to be able to apply timeouts to the handshake, and that would be difficult if Trio performed the handshake automatically before giving control to your connection handler.)

So try making it:

```python
async def serve(stream):
    await stream.do_handshake()
    print(stream.select_alpn_protocol())
```

pgjones:
Thank you, that has done the trick.

njsmith:
Awesome, glad that worked.

I wonder if there's anything we should do to make this less of a trap. We could update the docs, I guess. Or... actually, just because there stdlib has this annoying behavior where calling `selected_alpn_protocol` too early gives misleading results, doesn't mean we have to copy it. In our version of that method we could check whether we've done the handshake yet, and raise an error if not.

jxub:
Hi, I've just tried to address that in the PR, any feedback appreciated ;)

PR #743, opened 2018-10-20
Raise warning in case a SSLStream method returns none because no handshake was established

jxub:
References the issue #735 
The problem could be extended to the following methods in SSLStream which return none in case the handshake is not established:

- `get_channel_binding`: https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.get_channel_binding
- `selected_alpn_protocol`: https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.selected_alpn_protocol
- `selected_npn_protocol`: https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.selected_npn_protocol

These methods raise now a `NoHandshakeError` if the handshake wasn't established. Note that they still return `None` in the cases when:

- If the other party does not support ALPN/NPN.
- If `SSLContext.set_alpn_protocols()` or `SSLContext.set_npn_protocols()` was not called.

and in the case of `get_channel_binding` in the case when there is no connection.

### DIFF
--- a/trio/tests/test_ssl.py
+++ b/trio/tests/test_ssl.py
@@ -14,7 +14,7 @@ import trio
 from .. import _core
 from .._highlevel_socket import SocketStream, SocketListener
 from .._highlevel_generic import aclose_forcefully
-from .._core import ClosedResourceError, BrokenResourceError
+from .._core import ClosedResourceError, BrokenResourceError, NoHandshakeError
 from .._highlevel_open_tcp_stream import open_tcp_stream
 from .. import ssl as tssl
 from .. import socket as tsocket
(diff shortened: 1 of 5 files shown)


trio/tests/test_ssl.py
```diff
@@ -1109,6 +1109,76 @@ def close_hook():
             await client.do_handshake()
 
 
+async def test_selected_alpn_protocol_before_handshake():
+    client, server = ssl_memory_stream_pair()
+
+    with pytest.raises(NoHandshakeError):
+        client.selected_alpn_protocol()
+        server.selected_alpn_protocol()
```
sorcio: In order to make the test actually cover both lines, you need to have two separate `pytest.raises()` blocks. Otherwise what happens is that either the block exits at the first line because it raises (and it never executes the second one); or the first line doesn't raise, but you won't see that because the second one does raise.

```suggestion

    with pytest.raises(NoHandshakeError):
        server.selected_alpn_protocol()
```

The same applies to the other similar tests. This is the reason why codecov is complaining, and you fix it (assuming the tests pass) you will get full coverage again :)

jxub: I see, I'll fix that. Thanks for your help with that!

jxub:
@sorcio Just fixed the tests, and they seem to pass locally. Let's see what'll say codecov now :)

sorcio:
@jxub fantastic, thank you! :) Since this change is affecting public API I need to bother you a bit more for a couple small tasks:

- document that these methods can raise an exception in a given condition
- add a towncrier entry under `newsgragments/` briefly describing the change, as described in the contributor guide https://trio.readthedocs.io/en/latest/contributing.html#release-notes

I'm not sure what the best strategy to document this is. It could go in the docstring for `SSLStream` but it's easy to miss. You might need to explicitly define those three methods and add a docstring describing the behavior. I'm invoking @njsmith for a suggestion here.

jxub:
I added the things you suggested, both the documentation and the newsfragments entry, but I'll await any pending suggestions from both you @sorcio as well as @njsmith, as this documentation might not be enough in behaviour that varies from the stdlib one.

Fuyukai:
Failed build seems to just be a Travis hiccough, relevant build thing here: https://travis-ci.org/python-trio/trio/jobs/445524747

jxub:
Yay! :heart_eyes: Thanks.

njsmith:

PR #744, opened 2018-10-21
Removes call to getpeername in SocketStream

### LINKED ISSUE #609: SocketStream.__init__'s check for whether the socket is connected is unreliable on MacOS
On IRC, @dstufft told me about a strange crash he sometimes gets in [linehaul](https://github.com/pypa/linehaul) when dumping test data in using netcat, which looked like it might be a bug in trio's `SocketListener` or something.

Unfortunately, I didn't save the traceback, and I'm not really sure how to reproduce it... @dstufft, any idea?

wgwz:
Fixes #609.

I have a few questions:
- What should I do about this test which fails without the getpeername check? https://github.com/wgwz/trio/blob/wgwz/fix-for-609/trio/tests/test_highlevel_socket.py#L27-L30 I'm thinking it might be that the test just doesn't matter anymore. 
- What type of a newsfragment should I use? I'm guessing bugfix or removal..

### DIFF
--- a/newsfragments/609.bugfix.rst
+++ b/newsfragments/609.bugfix.rst
@@ -1,5 +1,2 @@
-This removes a check in the ``SocketStream`` constructor which was 
-making a call to ``getpeername`` method off of the trio socket object.
-This check was throwing an error in OSX due to an 
-`unreliable API <https://github.com/python-trio/trio/issues/609#issuecomment-414023026>`_
-in that platform. 
+Fixed a race condition on macOS, where Trio's TCP listener would crash if an
+incoming TCP connection was closed before the listener had a chance to accept it.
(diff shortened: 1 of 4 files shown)


njsmith:
* Yeah, you can just delete https://github.com/wgwz/trio/blob/wgwz/fix-for-609/trio/tests/test_highlevel_socket.py#L27-L30, it's testing for functionality we've decided to remove
* I'd go with bugfix – the main user-visible effect and motivation for doing this is that it fixes a bug on macOS

wgwz:
OK cleared up that test and added a newsfragment.

wgwz:
It's nice how the codecov report revealed those issues! I took a look at the reports. Thanks for suggestions on the newsfragments, all points noted. :-)

PR #751, opened 2018-10-27
Migrate trio.abc from star import to explicit imports, to aid static analysis

jmfrank63:
We continue to work towards making the code easier for static analysis tools.


njsmith:
Let's mention #542 in these PRs, so github will cross-link them and make it easier to find things later :-). I'm also going to change the title to make it easier to tell at a glance what's happening...

PR #752, opened 2018-10-28
migrate trio.ssl from star import to explicit imports, to aid static analysis issue #542

jmfrank63:
Use less magic in constructing our public API exports, to make trio more intelligible to static analysis #542


jmfrank63:
One test is failing but I cannot see where the import gets overwritten again.
```
____________________________________________ test_module_metadata_is_fixed_up ____________________________________________

    def test_module_metadata_is_fixed_up():
        import trio
        assert trio.Cancelled.__module__ == "trio"
        assert trio.open_cancel_scope.__module__ == "trio"
>       assert trio.ssl.SSLStream.__module__ == "trio.ssl"
E       AssertionError: assert 'ssl' == 'trio.ssl'
E         - ssl
E         + trio.ssl

trio/tests/test_util.py:73: AssertionError
```
Any ideas are welcome.

trio/ssl.py
```diff
@@ -4,5 +4,36 @@
 # temporaries, imports, etc. when implementing the module. So we put the
 # implementation in an underscored module, and then re-export the public parts
 # here.
-from ._ssl import *
-from ._ssl import __all__
+
+from ._ssl import (SSLStream, SSLListener, aclose_forcefully, ConflictDetector)
```
njsmith: This should just be `SSLStream` and `SSLListener`... the other ones were intentionally left out of `__all__` before :-)

trio/ssl.py
```diff
@@ -4,5 +4,36 @@
...
+# Fake import to enable static analysis tools to catch the names
+try:
+    from ssl import (
+        SSLError, SSLZeroReturnError, SSLSyscallError, SSLEOFError,
+        CertificateError, create_default_context, match_hostname,
+        cert_time_to_seconds, DER_cert_to_PEM_cert, PEM_cert_to_DER_cert,
+        get_default_verify_paths, Purpose, enum_certificates, enum_crls,
+        SSLSession, VerifyMode, VerifyFlags, Options, AlertDescription,
+        SSLErrorNumber
+    )
```
njsmith: These don't need to be fake! It's a well-defined set of things that should always be available. So we can just do a regular old `import` without any `try`/`except`.

trio/ssl.py
```diff
@@ -4,5 +4,36 @@
...
+        for (_name, _value) in _stdlib_ssl.__dict__.items() if _name in [
+            "SSLError", "SSLZeroReturnError", "SSLSyscallError", "SSLEOFError",
+            "CertificateError", "create_default_context", "match_hostname",
+            "cert_time_to_seconds", "DER_cert_to_PEM_cert",
+            "PEM_cert_to_DER_cert", "get_default_verify_paths", "Purpose",
+            "enum_certificates", "enum_crls", "SSLSession", "VerifyMode",
+            "VerifyFlags", "Options", "AlertDescription", "SSLErrorNumber"
+        ]
+    }
+)
```
njsmith: Once we switch to a regular import above, we can delete this.

trio/_ssl.py
```diff
@@ -159,53 +159,6 @@
...
-    "VerifyFlags",
-    "Options",
-    "AlertDescription",
-    "SSLErrorNumber",
-]:
-    _reexport(_name)
-
-for _name in _stdlib_ssl.__dict__.keys():
-    if _name == _name.upper():
-        _reexport(_name)
```
njsmith: This code is like `trio.socket`: it re-exports all the uppercased symbols. So we need some similar handling: a fake `try`/`import`, plus moving this loop into `trio/ssl.py`. With this PR right now it looks like we aren't re-exporting the uppercased symbols at all?

njsmith:
It might make sense to merge #756 before this, and then use #756's new tests instead of adding tests here.

njsmith:
This also needs a merge from master to pick up the changes in #756. Plus addressing the other comments above, of course :-)

jmfrank63:
> This also needs a merge from master to pick up the changes in #756. Plus addressing the other comments above, of course :-)

Right, I did this for #753 but totally forgot about this one. I'll also remove the extra imports. Might take a couple of days, I am really busy right now.

njsmith:
No worries :+1:

jmfrank63:
That's strange. I expected the test to fail, as I am only importing the ssl Mac symbols, so the jedi and pylint test for windows and linux should show some missing symbols. I'll checkout this repo on windows and see if it passes locally as well check which symbols are imported.
I also tried to import directly but only a subset of symbols was available on each system which made the tests to fail. Therefore I moved back to the try catch block which seems fine now.
@njsmith Let me know what you think.

jmfrank63:
I can confirm the tests did not catch`OP_NO_RENEGOTIATION` of the ssl module missing on windows. I assume this should be another issue, I'll try to get the missing symbols on linux via wsl.

njsmith:
I *think* that the main thing that determines which symbols appear on a given Python is (a) which version of Python you're using, and (b) which version of OpenSSL this Python was built against. I don't think it varies across platforms, except that if you grab a random Windows build and a random MacOS build then they may well have been built against different versions of OpenSSL.

trio/ssl.py
jmfrank63: Some actually don't but others do. I tried to import them directly but the test failed on some versions of python and with windows. I decided to revert back to being them all fake since doing a subset one way and the rest the other is more confusing and IMHO not worth the extra lines of code. Is this fine with you?

trio/_ssl.py
jmfrank63: They are now imported, but see my comment below about external dependencies like ssl library and version.

jmfrank63:
> I _think_ that the main thing that determines which symbols appear on a given Python is (a) which version of Python you're using, and (b) which version of OpenSSL this Python was built against. I don't think it varies across platforms, except that if you grab a random Windows build and a random MacOS build then they may well have been built against different versions of OpenSSL.

Understood, but that does not make it especially easier. Python can not only be build against different versions of OpenSSL it can also be build against other SSL libraries, e.g. LibreSSL.
Building tests for these scenarios will IMHO be not worth the effort.

njsmith:
1. It looks like this PR isn't actually enabling the generic exports test for the `trio.ssl` module? You need to remove this line from `trio/test/test_exports.py`:

```python
NAMESPACES.remove("trio.ssl")
```

2. I see what you mean about the different versions being frustrated... but I'm still wincing a lot at the duplicate lists :-(. As one last try before we give up and maintain two lists of symbols, can we try the following?

```python
# Always available
from ssl import (
     cert_time_to_seconds, CertificateError,
     create_default_context, DER_cert_to_PEM_cert,
     get_default_verify_paths, match_hostname,
     PEM_cert_to_DER_cert, Purpose, SSLEOFError, SSLError, 
     SSLSyscallError, SSLZeroReturnError
) 

# Added in python 3.6
try:
    from ssl import AlertDescription, SSLErrorNumber, SSLSession, VerifyFlags, VerifyMode, Options
except ImportError:
    pass

# Added in python 3.7
try:
    from ssl import SSLCertVerificationError, TLSVersion
except ImportError:
    pass

# Windows-only
try:
    from ssl import enum_certificates, enum_crls
except ImportError:
    pass
```

(It looks like the constants available depend on openssl version, but these lowercase things are mostly dependent on Python version.)

jmfrank63:
@njsmith Hi Nathaniel, thanks for the feedback and sorry for the delayed answer. I was really busy over the past days. I'll try to implement what you suggest and notify you once done. Thank you so much for letting me taking part of this awesome project.

jmfrank63:
??? I thought the second last test failed for pypy but now it is actually telling me that this test succeeded?
Well the difference is just two blocks vs one but the browser behaviour sometimes is really weird.
The split of blocks can easily be fixed by a rollback though. @njsmith Could you take a look?

njsmith:
This is looking good, except for some dead code left behind... let me try deleting that and check if it still works.

PR #756, opened 2018-10-29
Generalize namespace static analysis tests

njsmith:
This switches to a single parametrized test that covers all of Trio's
public namespaces, and if it fails, then it prints a nice readable
list of the missing symbols.

Also adds the symbols that were missing from trio.socket's static
import list on my laptop, and adds some comments to trio.socket while
I'm at it.

CC: @jmfrank63 @Zac-HD

### DIFF
--- a/trio/socket.py
+++ b/trio/socket.py
@@ -9,6 +9,16 @@
 from . import _socket
 import sys as _sys
 
+# The socket module exports a bunch of platform-specific constants. We want to
+# re-export them. Since the exact set of constants varies depending on Python
+# version, platform, the libc installed on the system where Python was built,
+# etc., we figure out which constants to re-export dynamically at runtime (see
+# below). But that confuses static analysis tools like jedi and mypy. So this
+# import statement statically lists every constant that *could* be
+# exported. It always fails at runtime, since no single Python build exports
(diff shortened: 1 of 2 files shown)


trio/socket.py
```diff
@@ -88,18 +98,21 @@
...
+# have:
 import socket as _stdlib_socket
 
 globals().update(
     {
-        _name: _value
-        for (_name, _value) in _stdlib_socket.__dict__.items()
-        if _name.isupper() and not _name.startswith('_')
+        _name: getattr(_stdlib_socket, _name)
+        for _name in _stdlib_socket.__all__ if _name.isupper()
```
pquentin: Any reason to change this again?

trio/tests/test_exports.py
```diff
@@ -27,54 +29,76 @@ def test_core_is_properly_reexported():
...
+    yield module.__name__
+    for name, value in module.__dict__.items():
+        if name.startswith("_"):
+            continue
+        if not isinstance(value, types.ModuleType):
+            continue
+        if not value.__name__.startswith(module.__name__):
+            continue
+        if value is module:
+            continue
```
pquentin: (This line does not seem to be covered by the tests.)

Issue #749, opened 2018-10-24
Our MacOS build logs are now hidden from everyone :-(


njsmith:
Our macOS builds are currently done by a Jenkins instance maintained by the PyCA folks.

Unfortunately, something has changed so that now when you try to click on a link to view a build log, you get a 404. For example: https://ci.cryptography.io/job/python-trio/job/trio/job/PR-728/10/display/redirect

I asked @reaperhulk about this, and apparently it isn't due to any intentional configuration change, it's something that just started happening after the last time they upgraded jenkins. And apparently you can configure it not to do this but... it just ignores that configuration.

For the moment, *I* can see the build logs, because @reaperhulk gave me admin rights on the Jenkins, because apparently that is the only thing it actually respects. But this is not a workable long-term solution. We all need to be able to see the logs.

Unless we have some Jenkins whisperer who wants to volunteer their time here, I guess we need to move macOS builds to Travis or Azure.

njsmith:
See also: #584 (and #583, #585)

njsmith:
This seems to be going OK, so I've turned off Jenkins for now (I think).

Notes:

- To re-enable, need to:
  - Go [here](https://github.com/orgs/python-trio/policies/applications/442515) and "grant access" again
  - Go [here](https://github.com/python-trio/trio/settings/branch_protection_rules/682095) and mark it required again

- To disable forever, need to:
  - Delete the `Jenkinsfile`

njsmith:
Apparently I only disabled one of the two places that Jenkins was hooked into the org settings. (I missed the webhook.) I've just disabled the webhook as well, so hopefully now it is actually disabled?

PR #759, opened 2018-10-31
Re-enable macOS testing on Travis

njsmith:
Since Jenkins is acting screwy and not letting people see the logs
:-(.

It's possible that this will cause gh-584 to start happening again,
so I guess we'll just have to watch out for that...

Fixes gh-749

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,15 +28,15 @@ matrix:
     - python: 3.8-dev
       dist: xenial
       sudo: required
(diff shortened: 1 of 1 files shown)


njsmith:
I guess we should merge this, and give it a day or two, and if it seems to be working out then I'll disable jenkins...

pquentin:
Thanks! 👍

PR #767, opened 2018-11-05
Exceptions don't need to be hashable.

smurfix:
The subject says it all …


njsmith:
Fair enough!

Can you add a test?

njsmith:
Continued in #1023

PR #768, opened 2018-11-05
Switch to black

pquentin:
We're currently using yapf to format our code, and this pull request is about switching to [black](https://black.readthedocs.io/en/stable/) instead.

The advantages of black:

 * the output is less likely to change when upgrading to a new version (this has been a problem for trio pull requests)
 * while `# fmt: off` / `# fmt: on` is supported, black indents nested dicts nicely, so I don't think we need that

The downsides:

 * contributors need to switch to the new tool
 * all pull requests will have conflicts :(
 * git blame will be less useful
 * it requires contributors to use Python 3.6+

The [first commit](https://github.com/python-trio/trio/pull/768/commits/8105b97bdfd31a031135416e572836e29a1f9161) prepares the switch, while the [second one](https://github.com/python-trio/trio/pull/768/commits/7ffbe9f0420c51ecee3e6fcd7a5c2a52ec394935) *only* runs `black setup.py trio`. Note that black uses 88 characters per line instead of 80.

What do you think?


Fuyukai:
I support this soley on having longer line lengths.

njsmith:
I am nervous about longer line lengths, because I have all my editor and terminal windows set up to be 80 chars wide at the largest font that will let me put two of them side-by-side on my laptop screen :-). You can configure black to use shorter lines, though, and I can take a look at how hard it would be to move my configuration out of the stone age...

I wonder how much of an issue the 3.6+ requirement will be. We definitely have users who are running on Debian stable or similar where 3.5 is the easiest python to use. OTOH yapf still doesn't work consistently on 3.7 (https://github.com/google/yapf/issues/624), so until that's fixed yapf is arguably worse than black in this regard. (At least black gives you a nice error if you run it on the wrong python version.) We can also potentially mitigate this with better tooling, e.g. a bot that runs black for you?

Last time I looked through the black output there were still some infelicities that jumped out at me. I was kind of hoping they'd get addressed before we switched:

https://github.com/ambv/black/issues/created_by/njsmith

But I guess @ambv has lots of things to worry about, and we're both distracted by this python governance mess...

Looking at the list of outstanding PRs, I think we want to get @jmfrank63's PRs merged first before doing anything here either way.

pquentin:
I don't have good answers to those questions. :)

I only know that if we do decide to switch to black, we do want to wait for @jmfrank63's PRs, yes, and any other active PRs: this PR is really easy to prepare again anyway.

I would be happy to close this and keep using yapf, I just wanted to make sure that the switch to black was not blocked because no one took the time to send a PR.

ambv:
> I was kind of hoping they'd get addressed before we switched:

Well, 9 out of 11 issues you opened are already closed, it's not like I'm not trying ;-) I responded on both open issues now. In those cases it's not lack of time, but lack of clarity that is holding me back.

pquentin:
Thank you @ambv for your efforts!

pquentin:
Closing for now, it will be easy to reopen if needed.

PR #782, opened 2018-11-21
[travis] Use xenial environment for all builds

njsmith:
This is now officially supported:
  https://blog.travis-ci.com/2018-11-08-xenial-release

It should make cryptography happier on our pypy builds because it has
a newer openssl:
  https://github.com/python-trio/trio/pull/781

Also dropped the sudo: tags, because Travis is in the process of
switching all builds to run on the old 'sudo: required' infrastructure
unconditionally:
  https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Conveniently, this also removes the penalty that used to exist for
using xenial.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
-sudo: false
-dist: trusty
+dist: xenial
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


pquentin:
Thanks! 👍 

3.5.0 no longer exists on Travis. I think we can simply test "3.5" and "3.5-dev" and drop 3.5.0 and 3.5.2.

njsmith:
Ugh yeah, it looks like xenial doesn't have anything before 3.5.2 (checked with: `s3cmd ls s3://travis-python-archives/binaries/ubuntu/16.04/x86_64/`)

Unfortunately due to a mistake in the early async/await spec, we have some special code that only runs on 3.5.0 and 3.5.1:

https://github.com/python-trio/trio/blob/91b54c4b4dc622a896113bc3842879672d2d4309/trio/_util.py#L76-L88

We could switch back to trusty just for the 3.5.0 job?

PR #783, opened 2018-11-22
unix_pipes.PipeSendStream: add send_some method

oremanj:
`send_some` does not block after its first write completes, and returns
an indication of how much data it sent, so it can be used in cases where
the amount of data sent before a cancellation must be known. This is
needed to implement stdlib-compatible `subprocess.Popen.communicate()`.

### DIFF
--- a/trio/_subprocess/unix_pipes.py
+++ b/trio/_subprocess/unix_pipes.py
@@ -46,28 +46,63 @@ class _PipeMixin:
 class PipeSendStream(_PipeMixin, SendStream):
     """Represents a send stream over an os.pipe object."""
 
-    async def send_all(self, data: bytes):
-        # we have to do this no matter what
-        await _core.checkpoint()
+    async def send_some(self, data: bytes, next_offset: int) -> int:
+        """Write data from ``data`` beginning at offset ``next_offset``
+        to the pipe until all data is written or an exception occurs.
+        This will block until some data can be written, but will return
+        a short write rather than blocking again after a previous
+        successful write.
+
+        Returns:
+          If some data is written, returns the index of the first byte in
+          ``data`` that was not written, or ``len(data)`` if all bytes
+          were written.
+
+        Raises:
+          If no data is written, raises the exception that caused the first
+          write to fail: :exc:`Cancelled`, :exc:`BrokenResourceError`, or
+          :exc:`OSError`.
+
+        """
+
         if self._closed:
+            await _core.checkpoint()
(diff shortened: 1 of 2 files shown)


njsmith:
I think we decided in chat that we weren't going to try to implement the stdlib `communicate` method's more exotic semantics (for now), which means we don't need this either (for now)?

njsmith:
Link to chat logs: https://gitter.im/python-trio/general?at=5bf5d64ced6bcf1ef83ccdb8

oremanj:
I just reviewed those chat logs again and I don’t think we actually decided anything - you pointed out that the stdlib behavior is unusual (which is true), I pointed out the usefulness of preserving at least the partial output across calls (which would require _some_ funkiness even if it wouldn’t need `send_some`), and I thought maybe if we were going to do something weird it would be better to preserve the full stdlib functionality rather than just part of it.

njsmith:
Subprocesses are pretty complicated, and `communicate` is especially complicated. As a general principle this makes me nervous, because when you (I) try to implement lots of complicated things at once, you (I) usually end up not thinking things through properly and committing to bad choices. So my intuition is that if we're worried, we should leave out `communicate` entirely from the first pass, and then come back to it later when we're not distracted by the other aspects.

(Storing unbounded amounts of string data as state on the process object itself? Maybe it's a good idea, I dunno, but it's definitely super weird.)

oremanj:
Fair enough. Making communicate() restartable seems like it would be a backwards-compatible change; I'm happy to leave that capability out of the initial version, which obviates the need for this diff.

njsmith:
Like, is `communicate` even a useful abstraction in the first place? It seems to me like the two main ways to interact with a process are:

* Interactively: spawn a process, send stuff, receive stuff, in a fine-grained way using Trio's regular I/O machinery
* One-shot: the equivalent of `subprocess.run`, where you want to run a process + send the input + get the output as a single string. If this is cancelled, kill the process and throw away the results.

Neither of these involves `communicate` and its tricky cancellation semantics...

oremanj:
Hmm, that's a good point. I had been thinking "well of course run() is implemented in terms of communicate()" but there's no particular reason that needs to be the case. I think a non-restartable communicate() is not going to be too much of a burden to maintain and leaving it out would surprise people, but I'm open to other alternatives.

njsmith:
Maybe! Let's move ahead with the stuff that we already know how to handle, and then revisit `communicate` later once the core subprocess functionality exists :-)

PR #790, opened 2018-11-27
docs: @njsmith is no longer lonely

pquentin:


njsmith:
True! :-)

njsmith:
...why is jenkins still running on this? I thought I disabled that...

PR #805, opened 2018-12-16
Make trio/_core/_run less magical and understandable for static code analysis issue(#542)

jmfrank63:
This is the first step on exec replacement. Manual code works (at least on OSX). Please take a look and add any comments.
The code still has the old code as dead commented code in it as a reference.

Addresses part of #542


njsmith:
Not sure what kind of feedback you're looking for here. Like we discussed in chat, I still think the strategy is unworkable compared to code generation – as written this pr adds lots of maintenance and runtime overhead. But maybe you have something more specific you're wondering about?

jmfrank63:
Hi Nathaniel,

the strategy is still unchanged, the code will be generated. This is only
to test the code that shall be generated. Generating code that does not
work is not a good idea imho.
Once I got this working and passing all tests, I can start to work out a
generation strategy. The feedback was more on "if this would be generated
code, could we go for it"

Sorry if I came unclear across.

Thanks

Johannes

njsmith:
Ah, that helps, thanks!

For generating code, we're going to want to generate the full code for each function, instead of using the `sync_helper` helper, e.g.:

```python
def current_time():
    try:
        meth = GLOBAL_RUN_CONTEXT.runner.current_time
    except AttributeError:
        raise RuntimeError("must be called from async context") from None
    return meth()
```

Since we're not maintaining the code by hand this should be just as easy to write/maintain as if we used a helper, and the helper adds a lot of runtime overhead as it has to do things like string parsing on every single call.

jmfrank63:
That's a neat idea, and also makes the code faster (well not by much). I
guess for generated code DRY is not important.
But first I want to make sure the windows part is working too. Got my dev
environment half set up, python and windows have sometimes a little
friction.

Once I got this setup I can run pytest locally and see what is going on.
Then I'll start my journey into AST and gentree.

jmfrank63:
@njsmith Now that the code is working, I'd like to start with AST. But I am a bit lost currently. I think I should put the generator into a script of it own, but who runs this? Jenkins, Travis, AppVeyor?

What was your idea?

njsmith:
Yeah, it should be a standalone script. I think the workflow will be similar to how we handle yapf: we'll check-in the generated source code, and we'll have a Travis job that runs the script to check that the output matches what's checked in. If not, it errors out, and tells the developer they need to regenerate the wrappers.

I'd start with just writing the script though, and we'll worry about hooking it up to CI later :-)

It looks like [astor](https://astor.readthedocs.io/en/latest/) is probably a good tool here – it can convert ast back to source code, so e.g. it should make it easy to copy the method signature and docstring into the generated code.

jmfrank63:
Great, @njsmith I have now two environments I can test on, I'll add linux later on. But I just wanted to share some failures I had last night. Maybe you have additional ideas.
I thought the `@_public` decorator might be a good place to replace the looping through the object attributes. Instead of marking the methods public I could export them directly from the decorator.
But what seemed to be an easy fix turned out to be a nightmare. Python3 dropped the unbound method and I am now at import time. While it is easy to get the class name and the function name with `__qualname__` I was unable to retrieve the class. Well I know for the `Runner` class where it lives but the aliased `TheIOManager` classes did show up as their module names and lived in different modules.
While I write these lines another idea sparks. I haven't tried to put _public into it's own class and have the other classes heritage from it. This way I should have access to self.
If I could solve the problem retrieving the right class to plug into the call I could export at import time. There is very likely a slim chance jedi and pylint will pick this up either. I'll be busy through the next two days so I thought I'll share this with you.
I will not drop the code generation as I found this will probably be very useful for my planned box async api based on trio. Instead of coding every endpoint by hand, generating them is very likely a better approach.

njsmith:
I would work from the AST, so not actually importing any Trio code at all. I'd start with hard-coding that I was looking for a `ClassDef` named `Runner`, inside the `trio/_core/_run.py` file, and then iterate over its contents looking for `FunctionDef` objects with a decorator named `_public`.

jmfrank63:
@njsmith This is work in progress, but I thought I'd update you on what had been happening.
I tried several approaches but all but the one you have here failed.
Main Failure: Trying to make use of the `@_public` decorator to dynamically export. While I got it working (advantage I already had the function name and didn't have to scan), jedi did not pick it up either.
But putting more code into the decorator proved to make my tests run longer, so I realized that this would have some performance impact. But the trial proved to be valuable in the end as I removed the decorator all together and just added an `_public = True` attribute removing an extra function call and achieving the same effect. It is just as easy to scan via AST. But yapf is complaining about the unused variable.
There are three ways around it: 
Tell yapf to ignore it
Use `_ = 'public'` instead
Put ` Public API` as the first line in the doc string and strip it out for the doc in the export function.
I plan to go for the last option (unless you think of something better)
I also tried to move the generated part into its own module, but that lead to test failures mainly in the `parking_lot` without any traces. I did not bother longer with this and decided to use the end of the `_run.py` file for the generation. I use `yapf: disable` for the split. This way I do not have to bother with the formatting complaints of yapf in my generated code.
`_run.py` is currently manually edited but passes all tests and shall be (unless you have some concerns) the version that is generated. Code completion works like a charm for trio.hazmat in VSCode and PyCharm.

njsmith:
> Main Failure: Trying to make use of the @_public decorator to dynamically export. While I got it working (advantage I already had the function name and didn't have to scan), jedi did not pick it up either. But putting more code into the decorator proved to make my tests run longer, so I realized that this would have some performance impact

The way I would do it is to make the actual decorator a no-op:

```python
def _public(fn):
    return fn
```

And then when scanning the AST we just look for methods that have a decorator named `_public`. (If you look at a `FunctionDef` or `AsyncFunctionDef` nodes, one of the attributes they have gives you a list of the decorators attached to the method.) That should be even easier than looking for variable assignments or scanning docstrings.

>  I also tried to move the generated part into its own module, but that lead to test failures mainly in the parking_lot without any traces. I did not bother longer with this and decided to use the end of the _run.py file for the generation. I use yapf: disable for the split. This way I do not have to bother with the formatting complaints of yapf in my generated code.

I think this is a bad idea. It's just too risky/fragile. We should put the generated code into its own file. I don't know what issue you ran into from the description, but I'm sure it's solvable.

One nice thing about putting it in its own file is that we never have to read this file at all -- the generator can just delete it and regenerate it from scratch every time.

jmfrank63:
@njsmith Thanks for the feedback I'll split the generated code into its own module and push it up. Maybe the CI gives us more info about the failures.
I had the empty decorator already in place but I am not sure if the python interpreter will optimize it away. The current version uses PUBLIC in the first line of the docstring which I intend to strip out for the exports. This way we save an extra function call that does nothing for each of the functions during run time.

jmfrank63:
Huh? It failed last time. But I guess I learned from when and where to import. Anyways makes it a lot easier.

njsmith:
> I had the empty decorator already in place but I am not sure if the python interpreter will optimize it away. 

Decorators only execute once, at import time. So technically an empty decorator isn't "optimized away" -- it does run! -- but it only runs once, and running an empty function an extra 10 or 20 times is not a measurable overhead.

jmfrank63:
Well, the code generation was unexpectedly put to a test. To my surprise the two new functions worked. What failed was the missing import from `trio/_core`. After adding the two public functions the local windows pytest succeeded.
@njsmith We still need some tests and some cleanup of the messy argument replacement code. Any wishes for changes can now be uttered :-)

jmfrank63:
@njsmith I have no idea why Mac / epoll is failing. Help much appreciated.

trio/_core/_gen_exports.py
```diff
@@ -0,0 +1,147 @@
...
+    raise RuntimeError('must be called from async context')
+    """.format(ctx, m.name + '(' + new_args + ')')
+        # Assemble function
+        ast_method = ast.parse(template)
+        m.body.extend(ast_method.body)
+        source = astor.to_source(m).replace('async ', '') + '\n\n'
+        with open(
+            os.path.join(m.module_path, '_public' + m.module_file), 'a'
+        ) as pub_file:
+            pub_file.writelines(source)
```
njsmith: We're also going to want a version of this that generates the text and then checks whether it matches what's already on disk (with just a simple string comparison). We'll run this in CI.

trio/_core/__init__.py
```diff
@@ -44,7 +40,22 @@ def _public(fn):
...
+
+# Windows imports
+try:
+    from ._run import (
+        wait_socket_readable, wait_socket_writable, notify_socket_close,
+        monitor_completion_key, current_iocp, register_with_iocp,
+        wait_overlapped, write_overlapped, readinto_overlapped
+    )
+except ImportError:
+    pass
```
njsmith: As noted in gitter, the reason the tests are failing on macOS is that you forgot to export the kqueue symbols in this file :-)

trio/_core/_gen_exports.py
```diff
@@ -0,0 +1,147 @@
...
+def gen_exports():
+    """ Sync the existing functions with the list
+    of exported functions imported from _public
+    """
+
+    imports = """from ._run import GLOBAL_RUN_CONTEXT, Runner
+from ._ki import LOCALS_KEY_KI_PROTECTION_ENABLED
+
+
+"""
```
njsmith: We should also put a big prominent comment at the top warning that this is an auto-generated file, and should not be edited directly.

trio/_core/_gen_exports.py
```diff
@@ -0,0 +1,147 @@
...
+    The function walks the given tree and extracts
+    all objects that are functions and have a
+    doc string that starts with PUBLIC
+    """
+    methods = []
+    for node in ast.walk(tree):
+        for child in ast.iter_child_nodes(node):
+            if (isinstance(child, ast.FunctionDef) or
+               isinstance(child, ast.AsyncFunctionDef)) and \
+               get_doc_string(child).startswith('PUBLIC '):
```
njsmith: I still think we're going to want to use the decorator to mark functions (`any(isinstance(dec, ast.Name) and dec.id == "_public") for dec in child.decorator_list`), instead of the docstring.

trio/_core/__init__.py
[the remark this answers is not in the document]
jmfrank63: Interesting, a missing import, and it only failed one test. I was looking for something like a default argument or something like this subtle. Anyways that fixed it. Thanks for helping out.

trio/_core/_gen_exports.py
```diff
@@ -0,0 +1,147 @@
...
+        arg_list = args.split(',')
+        new_arg_list = []
+        for arg in arg_list:
+            if '=' in arg:
+                key, value = arg.split('=')
+                key = key.strip()
+                new_arg_list.append('='.join([key, key]))
+            else:
+                new_arg_list.append(arg.strip())
+        new_args = ', '.join(new_arg_list)
```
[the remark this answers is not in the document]
jmfrank63: Oh, for sure, that was not meant to be final, just wanted it to work for now and have the CI passing. Thanks for giving me a generalized version, the arguments were of a major concern to me.
I'll test it out and let you know what the Outcome.outcome is ;-)

trio/_core/_gen_exports.py
jmfrank63: Oh, I had a different way in mind. The append is only because I do the imports first and then append to the file which has just the imports. I do not bother what was on disk before, I just create a new version every time. Why check the old version?

trio/_core/_gen_exports.py
jmfrank63: Well of course if you want decorators I will put them in again. One last remark on this. On the last pycon in Germany I talked to a peer and he took his working code and ported it from asyncio to trio. It went not smoothly but after figuring out what to do it worked.
However the performance was 30 % slower than on asyncio. Of course trio does a lot more than asyncio but that comes at a price. The impact of an empty decorator is not much but the functions you are exporting might be called a couple of thousand times a second and then even a few hundred nano seconds per call are up to a measurable millisecond.
I realized this when I tried to solve the problem from within the decorator, and my tests started to take longer.
Another suggestion to solve this is to just create a list where every name that should be exported is added. The decorator made perfectly sense when you did things at run time but now when stuff is done even before import you shouldn't use a decorator anymore.
As I write this the list is starting to get more and more my favorite as I do not even have to scan for public functions anymore I get them for free.

jmfrank63:
@njsmith Two bullet points are still open and I guess the code needs some tests. I'll put the decorator in if you want me but I don't understand why you want to compare with the disk version as I just throw the old away and regenerate a new one. And what are you going to do if there is a mismatch?
The tests will be probably the most difficult part, I am terribly bad at tests. But doing them here will hopefully get me nearer to my goal of true TDD.

smurfix:
It's a pure python module, so you want the code to be importable after checking it out with git as-is, without running any (re)generation. Thus you keep the generated code in git, which you ordinarily should not to, and verify that whoever last modified the source hasn't forgotten to also run the generator.

jmfrank63:
Hi Matthias,

yes, that's what currently is the case. It's not I don't want to do
anything about it, it is I don't understand what I am supposed to do. The
code is in git but when I run the generation code it just overwrites what
is already there (or not).

I tried to incorporate the code into the test suite but the first attempt
failed. Any hints would be much appreciated.

Thanks

Johannes

trio/_core/_gen_exports.py
jmfrank63: I'll go for the decorator. I still don't know how python does it, but I made some tests and there really was no difference. I have to admit the decorator looks much cleaner. You learn something every day ;-)

jmfrank63:
@njsmith Quick question, where should the script live? It's currently in `trio/_core/` but if it shall be used like yapf it should probably go somewhere else. What would be the best place?

trio/_core/_gen_exports.py
jmfrank63: @njsmith I now have the generated file formatted via yapf. Before I write the comparison, where should the current `_gen_exports.py` file live? What should be its name?
I still have to write the test though, but I'd like to know the final destination first, as it might not be part of code coverage.

trio/_core/_gen_exports.py
njsmith: Remember that decorators don't really have any magic... when you see:

```python
@_public
def fn(...):
    ...
```

that's just a shorthand for 

```python
def fn(...):
    ...
fn = _public(fn)
```

So if `_public` is a simple no-op like in our example above, then this becomes:

```python
def fn(...):
    ...
fn = fn
```

So it makes sense that a no-op decorator doesn't have any effect, because it literally has no effect at all after the module finishes importing.

njsmith:
> where should the script live?

Yeah, it probably shouldn't be inside the `trio/` package, because if we ever get into a state where we need to regenerate the code before we can import trio, we don't want regenerating the code to require importing trio first :-).

Let's just put it in the root of the git repo for now, and if it ever grows or we end up with more tools like it then we can create a `tools/` directory or something.

gen_exports.py
```diff
@@ -0,0 +1,223 @@
...
+def get_doc_string(func):
+    """ Returns the doc string of a function
+    or an empty sting if none
+    """
+    doc = func.body[0]
+    if isinstance(doc, ast.Expr):
+        if hasattr(doc, 'value'):
+            if hasattr(doc.value, 's'):
+                return doc.value.s
+    return ""
```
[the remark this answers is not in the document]
njsmith: And I especially don't understand why you needed to add empty docstrings to a bunch of methods in `_io_*.py` :-)

trio/tests/test_gen_exports.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+import ast
+
+import trio
+from trio import _core
+
+
+def test_docstring_is_found():
+    pass
```
njsmith: I'm actually not sure that there are a lot of tests to write here... we're testing the generated code pretty thoroughly by just running the regular Trio test suite.

I guess we could put in a test that the docstrings are passing through correctly, since that's something that doesn't affect normal functionality so the rest of the test suite wouldn't catch it. Could just be

```python
assert trio._core._run.Runner.current_time.__doc__ != ""
assert trio._core._run.Runner.current_time.__doc__ == trio.current_time.__doc__
```

Eventually we'll want to add tests that pylint/mypy can now trace references like `trio.current_time` back to find their signature, but that doesn't have to happen here.

The main test that's missing is: we should actually call `gen_exports.py --test` inside `check.sh`, so that we'll get alerted if someone submits a PR where the generated file is out-of-sync :-)

gen_exports.py
```diff
@@ -0,0 +1,223 @@
...
+    raise RuntimeError('must be called from async context')
+"""
+
+
+def is_function(node):
+    """Check if the AST node is either a function
+    or an async function
+    """
+    return isinstance(node, ast.FunctionDef) \
+           or isinstance(node, ast.AsyncFunctionDef)
```
njsmith: Minor style nit: it would be nicer to use parentheses instead of `\`

gen_exports.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def is_public(node):
+    """Check if the AST node has a _public decorator
+    """
+    return is_function(node) \
+           and node.decorator_list \
+           and isinstance(node.decorator_list[-1], ast.Name) \
+           and node.decorator_list[-1].id == '_public'
```
njsmith: again with the parentheses

gen_exports.py
```diff
@@ -0,0 +1,223 @@
...
+    marked to be public as AST module trees
+    """
+    export_modules = []
+    modules = get_module_trees_by_dir(source_dir)
+
+    for module in modules:
+        methods = get_public_methods(module)
+        if methods:
+            export_modules.append(module)
+    return export_modules
```
njsmith: This whole section feels more elaborate than we need... we know there are exactly 4 files we care about, and we could just hard-code it here. If the list ever changes, well, this lives in the source tree right next to the files, so we can just update it :-)

I guess if you prefer to do things this way it's not actually wrong? Just saying it could be simplified too.

gen_exports.py
```diff
@@ -0,0 +1,223 @@
...
+    Every default argument value is
+    replaced by its argument corresponding
+    argument name
+    """
+    call_args = []
+    for arg in funcdef.args.args:
+        call_args.append(arg.arg)
+    for _, arg in zip(funcdef.args.defaults[::-1], funcdef.args.args[::-1]):
+        call_args.remove(arg.arg)
+        call_args.append(arg.arg + "=" + arg.arg)
```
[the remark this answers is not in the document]
jmfrank63: I thought so first too, but the tests failed without the additional default arguments. Especially _NO_SEND did not work without this little trick. I could try to remove it again and check if I made a mistake, but I spend quite some time to figure this out, so I know I have to repeat the default argument otherwise some tests fail.
What do you suggest?

gen_exports.py
```diff
@@ -0,0 +1,223 @@
...
+        # Remove method body
+        del method.body[1:]
+
+        # Create export function body
+        template = TEMPLATE.format(ctx, method.name + new_args)
+
+        # Assemble function definition arguments and body
+        ast_method = ast.parse(template)
+        method.body.extend(ast_method.body)
+        snippet = astor.to_source(method).replace('async ', '')
```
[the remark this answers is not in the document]
jmfrank63: I thought I had tried this already, but again I must have had a different issue. Anyways, changed this, should be in the next commit.
This is anyways better for the documentation as the function shows up as an async function.

gen_exports.py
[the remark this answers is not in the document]
jmfrank63: Always having a doc string makes it easier to create the function body as I don't have to test if the doc string is present.
I actually had planned to add doc strings to all functions in _public_* as code completion shows the doc string, which will be very helpful.
Adding doc strings to all public api functions was on my list anyway. Do you want to leave the doc strings out?

trio/tests/test_gen_exports.py
jmfrank63: Added the test to `check.sh`. Doc string test not yet done, the function was more or less a prototype for the function inside `gen_exports.py`.

gen_exports.py
jmfrank63: Done

gen_exports.py
jmfrank63: We currently have three major OS but that might change in the future. Someone might port it to BSD or even to REDOX. Then you just have to add your modules without worrying where they have to live. IMHO we should just leave it now that it is in place instead of reverting, unless there are major disadvantages.

jmfrank63:
@njsmith Interesting it seems that yapf on python 3.6 fails, while on 3.7 it doesn't.

njsmith:
It's not yapf failing :-). The error is:

```
++python ./gen_exports.py --test
  File "./gen_exports.py", line 191, in <module>
    sources = gen_source()
  File "./gen_exports.py", line 174, in gen_source
    ast_method = ast.parse(template)
  File "/opt/python/3.6.7/lib/python3.6/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 3
    return await GLOBAL_RUN_CONTEXT.runner.wait_all_tasks_blocked(cushion, tiebreaker)
                                  ^
SyntaxError: invalid syntax
```

So what that says is that the CI script ran the command `python ./gen_exports.py --test`, and that errored out with that traceback. The traceback says that the crash came when calling `ast.parse(template)`.

So the problem is that the template has some invalid syntax in it. And... that's true, because now we're putting an `await` in there – which is good! – but, Python doesn't allow a bare `await` at the top level, only inside an `async def`. And the template code isn't inside an `async def`.

I think the easiest solution is just... put the template code inside an `async def` :-). That is, we should make the template like:

```python
TEMPLATE = """
async def template_function():
    try:
        return {} GLOBAL_RUN_CONTEXT.{}.{}
    except AttributeError:
        raise RuntimeError('must be called from async context')
"""
```

Then when we parse the template, this means that we'll have to reach a bit deeper into the AST to pull out our new method body. Something like:

```python
        ast_method = ast.parse(template)
        method.body.extend(ast_method.body[0].body)   # instead of ast_method.body like we do now
```

njsmith:
Oh, and as for the 3.6 versus 3.7 thing: this is kind of interesting actually! In both versions, `compile("await foo()", "", "exec")` gives a `SyntaxError`, because it's invalid Python. But they issue the errors at slightly different times: in 3.6, it's detected during the parse itself, so `ast.parse("await foo()")` gives a syntax error, but in 3.7, it's detected in a later part of the compiler, so `ast.parse("await foo()")` works.

jmfrank63:
I checked again, and it is valid syntax, because the template only contains the body of the function.

```
TEMPLATE = """locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
try:
    return {} GLOBAL_RUN_CONTEXT.{}.{}
except AttributeError:
    raise RuntimeError('must be called from async context')
"""
```
On a sync function there is no await, only when the function is async.

I went through all of the generated code manually and it seems to do exactly what it should do, only have an await in there when the function is async.
Also on 3.7.2 check.sh and pytest do not fail.

```
╰─ sh ./check.sh
+ EXIT_STATUS=0
+ python ./gen_exports.py --test
+ yapf -rpd setup.py trio
+ flake8 trio/ --ignore=D,E,W,F401,F403,F405,F821,F822
+ '[' 0 -ne 0 ']'
+ exit 0
```

njsmith:
It is invalid syntax. But like I said, Python 3.7 moves the check for this invalid syntax to a different place in the parsing pipeline, so that `ast.parse` doesn't catch the invalid syntax. But on 3.6, `ast.parse` does catch the invalid syntax.

It's true that the current code ends up working on 3.7, but we want it to work on 3.5/3.6 as well. And also, the only reason it works on 3.7 is because of this weird little edge-case, so it's better if we don't depend on that – maybe it will stop working on 3.8, who knows.

jmfrank63:
Looks like I am blind, I don't get it. I treated functions and async functions alike before. Then when adding the await keyword I only add it to async functions, sync function do not have an await in it.
` 'await' if isinstance(method, ast.AsyncFuncDef) else '' `
A manual check showed only awaits inside async function definitions. There isn't a single await inside a  sync function.
So what do I miss here?

njsmith:
When you do `ast.parse(string)`, then Python treats the string as if it's the contents of a `.py` file.

When we're trying to create the wrapper for an async function, currently we fill in the `TEMPLATE` and end up with a string like:

```python
try:
    return await GLOBAL_RUN_CONTEXT.runner.wait_all_tasks_blocked(cushion, tiebreaker)
except AttributeError:
    raise RuntimeError('must be called from async context')
```

To you, this is the body of an async function. But to `ast.parse`, this is the body of a module :-). And if you put that into a `.py` file and try running it, you'll get a `SyntaxError`, because you can't use `await` at the top-level inside a module.

Does that make sense?

jmfrank63:
Ok, got it, that means I have to delay the compile until the function header is in or use two different templates.

njsmith:
Or you can do what I suggested in my earlier comment :-)

Just add an `async def template_function():` to the top of the `TEMPLATE` string. That will work fine for both sync and async functions, since all we care about is makng sure that the template can be parsed. Then once it's parsed, we'll do `ast_method.body[0].body` to rip out the function body. The first `.body[0]` gives us the `AsyncFunctionDef("template_function", ...)`, then adding `.body` to that gives us the body of the function. Since we're throwing away the actual `AsyncFunctionDef("template_function", ...)` node, it doesn't matter whether it's `AsyncFunctionDef` or `FunctionDef`.

jmfrank63:
I like your idea it works fine as well. What I did is to parse the function header without the body, (now the empty doc string comes in very handy hehe ;-) and then just indented the body and concatenated the two. I want to create source code so why go back and forth?
But it is your library, so if you prefer the added async function line I'll go that way of course.
It is unlikely the indentation will ever change, but just in case the next commit will make sure a change will not break the concatenation.

jmfrank63:
Writing the tests proved to be a good idea. Already detected two hidden bugs. Another one is a potential candidate.
@njsmith What is the desired behaviour if a `_public` function has another decorator?
Is this decorator to be added to the api exposure as well?

jmfrank63:
@njsmith Need to do more research about codecov. Currently I do not understand why the generated files are still tracked although omitted.

njsmith:
Copying from gitter so it doesn't get forgotten:

suggestion: can you rename the generated files to something like `trio/_core/_generated/....py`, and make the coverage omit `*/_generated/*`?

jmfrank63:
Well, the code generator could surely bare some refactoring and some more tests, but it seems to work for now. I would open two more PRs after this:
One for adding doc strings to all public methods, and one for refactoring.
@njsmith @smurfix We are close now, but another review could not hurt.

jmfrank63:
@njsmith The names are now _generated but I just realised this name also shows up in pycharm code completion. I am not sure if we should expose this internal detail. Given this _public might be the better choice. Can be redone easily just my five cents of thought.

njsmith:
@jmfrank63 I don't use pycharm so I'm not sure what you're seeing. Would it be easy to post a screenshot here?

njsmith:
Showing-off internal implementation details like that is a bit ugly, but the `_generated` names don't strike me as particularly worse than the others?

jmfrank63:
> Showing-off internal implementation details like that is a bit ugly, but the `_generated` names don't strike me as particularly worse than the others?

I am not complaining, I just noticed and wanted to let you know. If this is fine for you we'll leave it as it is. 
I could refactor the code before we ship it, or we could ship and I open a new PR for the refactoring.

jmfrank63:
@njsmith @smurfix Keeping this up to date is quite some work, so please let me know if I have to change anything, I'll be happy to do so. Of course I have to admit I'll be really happy to see this one merged.

jmfrank63:
Hi Nathaniel,
in fact I already had an _tools in place but reverted in order to stick
what was agreed. Full support here.
I’ve tried the module level, but then some tests failed. I’ll try again, as
in the past it did always work and I probably did make something else wrong.

njsmith:
> what was agreed

yeah, sorry about that! Now that we have unit tests it's a bit more complicated :-)

Zac-HD:
> I feel like @Zac-HD might have thoughts on [how to arrange code and tests]?

Nope, sorry - I wouldn't usually distribute tests via PyPI, and haven't used on-disk codegen in anger, so I don't think my intuition is worth anything here :confused:

njsmith:
@jmfrank63 Hey, just wanted to check in – you're not waiting for anything from me, right? (If you're just busy, don't worry, we know it happens :-))

jmfrank63:
Hey Nathaniel, sorry yes, was sick and in the middle of a job transition.
The only thing I’d like to ask is Laura what’s her take on the NOSEND as a
module symbol. I don’t think that’s going to work given her comment in the
code ;-)

njsmith:
@jmfrank63 I think you've gotten misled by a confusing aspect of Python here :-). When you define a constant inside a `class` block, like:

```python
class Runner:
    _NO_SEND = object()
```

Then that *doesn't* create a special kind of per-object variable. It's exactly like declaring the constant at the module level, except that instead of accessing it like `trio._core._run._NO_SEND`, it's accessed like `trio._core._run.Runner._NO_SEND`. So all you need to do is move the original definition up out of the `class` block, and then refer to it everywhere as `_NO_SEND` or `_run._NO_SEND`, instead of `Runner._NO_SEND`.

jmfrank63:
> @jmfrank63 I think you've gotten misled by a confusing aspect of Python here :-). When you define a constant inside a `class` block, like:
> 
> ```python
> class Runner:
>     _NO_SEND = object()
> ```
> 
> Then that _doesn't_ create a special kind of per-object variable. It's exactly like declaring the constant at the module level, except that instead of accessing it like `trio._core._run._NO_SEND`, it's accessed like `trio._core._run.Runner._NO_SEND`. So all you need to do is move the original definition up out of the `class` block, and then refer to it everywhere as `_NO_SEND` or `_run._NO_SEND`, instead of `Runner._NO_SEND`.

Hehe, I should have known that you are probably right, but I tested it nevertheless. I am just a notorious unbeliever. At least I know have a better understanding of class variables.
I still don't know what kind of mistake I made when I attempted to do it in the past, this is the third time something like this happens. I tried it it failed and I thought ok, let's work around but the failure must have had some other reason.
_NO_SEND is now defined on module level and, surprise, surprise it works.
I am super busy the next two weeks, try to sneak in the relocation of tools and the code refactoring, but cannot promise this.

jmfrank63:
@njsmith Well, coverage again, I still don't grab their philosophy. I moved the test into trio/tests/tools expecting the test to be picked up by coverage. But that seems not to be the case.
What do you suggest?

jmfrank63:
Phew! We are getting closer. Only two more tests!

jmfrank63:
@njsmith I don't think without considerable amount of work I will get any closer. But you are always good for some fresh ideas. Anything that could improve coverage?
I must admit I am still confused. For example I improved coverage on my local test but the patch actually sank in percentage. I will have to implement this on my own with less complex code to get what they are doing.

jmfrank63:
Hi @njsmith just wanted to follow up on this. We are getting incrementally close, but as normal for limits there is still a non zero difference.  Besides the tests written, the code completion does show up in both VS Code and PyCharm. I know this is a daring PR, so I am willing to do further rounds of optimization.
But I think we should "release often and early" so users can test and report bugs.

trio/tests/tools/test_gen_exports.py
```diff
@@ -0,0 +1,233 @@
...
+def test_is_public(module):
+    for node in ast.walk(module):
+        if is_function(node) and node.name == 'public_func':
+            assert is_public(node) is True
+        if is_function(node) and node.name == 'non_public_func':
+            assert is_public(node) is False
+
+
+@pytest.mark.skipif(
+    sys.version_info >= (3, 8), reason="requires on dev version"
```
pquentin: Sorry, I have not read the whole PR, but I don't understand this comment. I think the test only works on stable versions, so maybe the reason should read "requires a stable version" instead? ditto for the other occurrences below.

jmfrank63: Hi @pquentin, right thanks for pointing this out, clearly a typo. I am not sure, but maybe I should reference https://github.com/berkerpeksag/astor/pull/121

which will hopefully help to remove the skipif.

I'll fix this.

jmfrank63:
Grrh, yapf, even switched to python 3.6.7 the exact same version as on travis but still getting different results.
I think we should put yapf into the build process, so developers do not have to care about running yapf.
Instead of running yapf and then checking it would just run and be the same for everybody independent of OS.
@pquentin @njsmith Any idea how to fix the formatting? Besides these issues, anything else you want me to do?

pquentin:
@jmfrank63 You simply got hit by the yapf upgrade from https://github.com/python-trio/trio/pull/920. In other words, you forgot to upgrade yapf with `pip install -r test-requirements.txt` after merging from master. You should not have included your "Fixed formatting". Does that make sense?

jmfrank63:
@pquentin Thanks, yes, I have to improve my workflow with pipenv. Since I reinstalled the whole virtualenv I did not think I get locked versions, but Pipfile.lock did fix me to the old yapf.

jmfrank63:
@njsmith @pquentin @Zac-HD Feeling like Achilles never being able to reach the tortoise. But getting closer and closer ;-) Maybe someone finds another gap I have to bridge.

jmfrank63:
@njsmith I just got word that there is progress on
https://github.com/berkerpeksag/astor/pull/121#issuecomment-465442783
That will then enable us to remove the python-3.8-dev skipping.
We are just 0.21% away from a full coverage, so I am happy for any suggestions to get this last fraction covered.

jmfrank63:
Hi @njsmith I am unsure what to do next. I know this touches the core of the library and if there is a major bug it will break things massively. However if we never merge, we will not get any bug reports. Maintaining this without a merge is quite some work. If things break we could still roll back.
It is fine if you decide not to go for code generation, I still learned a lot. Sometimes things change, so don't worry I am glad to be able to participate.

njsmith:
OK, I finally got back to this... sorry for the huge delay there. In hopes of getting it in ASAP, I went ahead and merged, did some cleanup, etc. But now I hit a wrinkle: I had missed that in the original version, the data flow was carefully constructed to let the tests run without ever invoking yapf to format the generated code. In my refactored version, the formatting happens automatically as part of the regular code generation, because, well... it's a mandatory part of generating the code!

But... that means that now the tests are trying to access `.style.yapf`, which we don't ship as part of the trio package, and we do ship the tests as part of the package, so the tests don't work, and ugggghh. Even I have trouble convincing myself that we should ship `.style.yapf`. I'm halfway tempted to just drop the tests and rely on the principle that if our generated code is broken, then we'll notice because everything is broken. (It's not like the code-generator-specific tests do any detailed checks that the generated code works, they just test a few edge cases + do an end-to-end smoke test to make sure the `-t` switch does something.) We could... run some self-tests as part of the regular `gen_exports` invocation? That would be hard for the end-to-end smoke-test though. We could... have two separate test suites, one for trio and one for the codegen, and invoke pytest twice? Kind of tempting in principle, but I bet it would be annoying to get the CI and coverage and everything working. We could... bite the bullet and move all of Trio's tests out of the source tree and somehow figure out how to make that work properly?

Maybe if I go to sleep something brilliant will occur to me.

jmfrank63:
Hi Nathaniel,

this brings me back to a recommendation I got from my former colleague a
very experienced developer advocate. He pointed out the generated code
shouldn't live in source control at all, but be generated as part of the
deployment process, as this is never changed anyway by humans.

i still find this idea not too bad, as the generated code does not do any
functionality changes, just bringing some convenience to code completion.
We can still test but can get rid of the styling, generated code does not
need to be styled. If we need to compare the code to some older code we
could do this via AST. In any case that would do the trick.

Let me know what you think of this.

Thanks

Johannes

njsmith:
> He pointed out the generated code shouldn't live in source control at all, but be generated as part of the deployment process, as this is never changed anyway by humans. 

This is good advice, but like most advice in programming, it applies in some situations and not others :-). In our case, we don't currently have a "deployment process" – you can run trio straight from the source tree, and to package it we just dump the source in a wheel. So in order to do this, we'd have to make significant workflow changes. And adding a new "deploy" step that has to be run before every test is much more of a hassle then having to run the code generator once every few months.

Anyway, I guess that is orthogonal to your main idea, which is to skip formatting the autogenerated code. I guess we could add a `.yapfignore` file to exclude all `_generated*.py` files from being formatted: https://github.com/google/yapf#excluding-files-from-formatting-yapfignore
That way we're checking in unformatted source, but it's fine, because yapf won't complain either.

Other options:

- We could fix the code generator's raw output so that it's correctly formatted to start with, so there's no need to run yapf on it. I spent a few minutes trying this, and I don't think it's viable: most of the stuff is easy to fix, but in some cases the generated "passthrough" calls have long lines that yapf wants to break up, and detecting and formatting those correctly ourselves is too much work.

- We could switch to black, which has way simpler configuration, which we could hard-code. (Basically just line length and maybe minimum supported python version.)

njsmith:
The `.yapfignore` thing seems to be working fine, so let's worry about black another time :-). Good idea, thanks!

I think that means this is ready to go once the tests pass...

jmfrank63:
Thanks Nathaniel,

great news. What would be your recommendation to be next on #452 ?

Cheers Johannes

Issue #762, opened 2018-10-31
Update happy eyeballs for RFC 8305


njsmith:
A new happy eyeballs RFC came out a few months ago: https://tools.ietf.org/html/rfc8305

Golang tracking issue: https://github.com/golang/go/issues/23841

It's interesting. The old one (RFC 6555) was incredibly vague; lots of "well maybe you could do this, or maybe that, there are lots of options ok", then with a paragraph stuck in at the end claiming (incorrectly) to document what browsers actually do. This one has actual concrete recommendations, and apparently corresponds to what MacOS/iOS do. So... anything we should take from it?

They recommend a 250 ms default timeout. Okay, sure, cool. There's some language about adjusting this based on history, within the bounds of 100 ms and 2 s, but I don't think we want to bother tracking history. And, we're not in a great place to do it either – if we were in the kernel or glibc then it would make more sense.

They recommend doing some cleverness with AAAA vs A records during DNS lookup, where if you get a DNS response for your preferred address family first, you should go ahead and start making connections while waiting for the other DNS response. It turns out that you actually can use `getaddrinfo` to do separate AAAA and A lookups (just specify `family=AF_INET6` vs `family=AF_INET`; in glibc at least this does correctly trigger the appropriate DNS queries, confirmed with wireshark). So in principle we could implement this pretty easily... IF we knew the local host's policy for preferring IPv4 vs IPv6. Which we don't, since it's some operating system level tuning parameter and there's no standard API for accessing it. Sigh. Maybe the new RFC will encourage OSes to implement better APIs here? There's some temptation to hard-code a preference for AAAA, because c'mon it's the future, and that might work fine on consumer devices, but I think this might annoy all the folks still using IPv4-only networks in the datacenter, and that's kind of an important target userbase for trio.

In fact a lot of the fancier recommendations really only make sense when the happy eyeballs algorithm is integrated into the OS. Does anyone know how Apple actually exposes their implementation? Is there some API we should be calling on macOS instead of using our implementation? Or do you have to switch over to some totally different networking API instead of BSD sockets?

It looks like in the short term the only action item here might be to switch our default delay to 250 ms. (I'm sort of tempted to refactor our happy eyeballs implementation a bit while I'm in there, to more closely match the later version I use in the talk, and to track unclosed sockets in a more automated way.)

joernheissler:
On my system the default is to return v6 addresses first, then v4. This can be changed in `/etc/gai.conf`.
[...]

PR #809, opened 2018-12-18
improve error handling in open_tcp_stream

smason:
improve handling of `open_tcp_stream` in the case of errors/exceptions being thrown.  previously it looked like there were a few places the code could be interrupted and it would end up leaking file descriptors

I've changed it to track everything in `open_sockets`, and only have a single winner.

I started writing this along with [Nathaniel's PyCon 2018](https://www.youtube.com/watch?v=oLkfnc_UMcE) talk but ended up structuring the retry behaviour differently as I found his confusing to follow.  Not sure if my code will catch all the cases appropriately or is actually better, hence the WIP.

Comments appreciated!

### DIFF
--- a/trio/_highlevel_open_tcp_stream.py
+++ b/trio/_highlevel_open_tcp_stream.py
@@ -113,6 +113,16 @@ def close_on_error(obj):
         raise
 
 
+@contextmanager
+def close_all():
+    sockets_to_close = set()
+    try:
+        yield sockets_to_close
+    finally:
+        for sock in sockets_to_close:
+            sock.close()
+
+
 def reorder_for_rfc_6555_section_5_4(targets):
     # RFC 6555 section 5.4 says that if getaddrinfo returns multiple address
     # families (e.g. IPv4 and IPv6), then you should make sure that your first
@@ -245,9 +255,6 @@ async def open_tcp_stream(
     # This list records all the connection failures that we ignored.
     oserrors = []
 
-    # list of all open sockets we need to clean up before returning
-    open_sockets = []
-
     # the socket that we're going to complete with, we need to make
     # sure this is excluded from any cleanup
     winning_socket = None
@@ -263,7 +270,7 @@ async def open_tcp_stream(
 
         try:
             sock = socket(*socket_args)
-            open_sockets.append(sock)
+            open_sockets.add(sock)
 
             await sock.connect(sockaddr)
 
@@ -279,7 +286,7 @@ async def open_tcp_stream(
             oserrors.append(exc)
             attempt_failed.set()
 
-    try:
+    with close_all() as open_sockets:
         # Kick off the chain of connection attempts.
         async with trio.open_nursery() as nursery:
             for *sa, _, addr in targets:
@@ -301,13 +308,6 @@ async def open_tcp_stream(
             )
             raise OSError(msg) from trio.MultiError(oserrors)
         else:
-            return trio.SocketStream(winning_socket)
-    except:
-        # make sure we also close the winning_socket on failure
-        winning_socket = None
-        raise
-    finally:
-        # close all open sockets appropriately
-        for sock in open_sockets:
-            if sock is not winning_socket:
-                sock.close()
+            stream = trio.SocketStream(winning_socket)
+            open_sockets.remove(winning_socket)
+            return stream
--- a/trio/_highlevel_open_tcp_stream.py
+++ b/trio/_highlevel_open_tcp_stream.py
@@ -252,16 +252,17 @@ async def open_tcp_stream(
     # This list records all the connection failures that we ignored.
     oserrors = []
 
-    # the socket that we're going to complete with, we need to make
-    # sure this is excluded from any cleanup
+    # Keeps track of the socket that we're going to complete with,
+    # need to make sure this isn't automatically closed
     winning_socket = None
 
-    # Sleep for the given amount of time, then kick off the next task and
-    # start a connection attempt. On failure, expedite the next task; on
-    # success, kill everything. Possible outcomes:
-    # - records a failure in oserrors, returns None
-    # - records a connected socket in winning_socket, returns None
-    # - crash (raises an unexpected exception)
+    # Try connecting to the specified address. Possible outcomes:
+    # - success: record connected socket in winning_socket and cancel
+    #   concurrent attempts
+    # - failure: record exception in oserrors, set attempt_failed allowing
+    #   the next connection attempt to start early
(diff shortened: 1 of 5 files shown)


njsmith:
Whoa, this is awesome.

Yeah, the old code could leak file descriptors in rare cases. It required an unlucky alignment of the stars, and only really matters on PyPy, but it's been bugging me for *months*. So thank you. It looks like I never actually opened an issue for it, but there's a brief mention [here](https://github.com/python-trio/trio/issues/762).

I never thought of moving the loop-and-wait logic into the parent task like that. That's really elegant! (I guess this might be a side-effect of my originally writing this back in prehistoric Trio, where you [couldn't put blocking logic into the parent task](https://trio.readthedocs.io/en/latest/history.html#trio-0-3-0-2017-12-28)? But it's kind of mind-blowing to realize I spent weeks tweaking the logic for the talk and still missed this.)

One quibble with the socket tracking: I think I'd factor it out into a helper, like:

```python
@contextmanager
def close_all():
    sockets_to_close = set()
    try:
        yield sockets_to_close
    finally:
        for sock in sockets_to_close:
            sock.close()

async def open_tcp_stream(...):
    with close_all() as sockets_to_close:
        ...
        # Opening a socket
        sock = socket.socket(...)
        sockets_to_close.add(sock)
        ...
        # Finishing up
        if winning_socket is None:
            ...
        else:
            sockets_to_close.remove(winning_socket)
            return winning_socket
```

This makes the main function's flow clearer, and lets you drop the awkward `except: winning_socket = None; raise` logic.

It would be great if we could add tests to confirm that the new code cleans up sockets correctly, in cases where the old code didn't. If you look in `trio/tests/test_highlevel_open_tcp_stream.py`, there's a pretty robust framework for setting up fake happy eyeballs scenarios, and it already has code to [check that sockets are closed properly](https://github.com/python-trio/trio/blob/master/trio/tests/test_highlevel_open_tcp_stream.py#L185-L186), so in principle it should just be a question of setting up a scenario where the connection is cancelled at just the wrong time. That might be difficult to do though, I'm not sure :-). (I don't remember off the top of my head exactly what the conditions are for provoking the old code to leak file descriptors.)

We'll also want a [news entry](https://trio.readthedocs.io/en/latest/contributing.html#release-notes), noting that we fixed some edge cases where `open_tcp_stream`  could leak file descriptors on PyPy.

smason:
thanks for the positive feedback!  I'll do some cleaning up and push again.  didn't want to go too far without some confirmation that this was a useful change

the "simplified nurseries" is an interesting feature, feels like an intuitive change

I'm not sure how I'd provoke any resource leakage, it was more of an academic concern…  I made sure tests succeeded before committing, but that's as far as I went into looking at your test framework — I'm generally afraid of concurrency-related heisenbugs

also, thanks for publishing Trio!  it's a very cool experiment, be interesting to see where it goes!

trio/tests/test_highlevel_open_tcp_stream.py
```diff
@@ -97,7 +97,7 @@ def test_format_host_port():
 # Now, thorough tests using fake sockets
 
 
-@attr.s
+@attr.s(cmp=False)
```
smason: I went with @njsmith's suggestion of a `set()` to track open sockets, but this requires that the socket is "hashable", which `attr` objects aren't by default

this seems to be the [suggested way](http://www.attrs.org/en/stable/hashing.html) of setting objects like this…

smason:
I've done the easy stuff…  just need to figure out a test that would leak in the old version and make sure my version catches it

I put dd2be12ed062f2f57cbf0408a42c9df9ecd27682 into its own commit as it feels somewhat gratuitous and might lead to unnecessary history/merge awkwardness down the line.

trio/tests/test_highlevel_open_tcp_stream.py
njsmith: :+1:

njsmith:
> I put dd2be12 into its own commit as it feels somewhat gratuitous and might lead to unnecessary history/merge awkwardness down the line.

Looks fine to me.

> I've done the easy stuff… just need to figure out a test that would leak in the old version and make sure my version catches it

I looked at it again, and I think one case where it could leak would be if `open_tcp_stream` gets cancelled at the same moment as a socket successfully connects. We already have a simple test for cancellation (`test_cancel`) – it might be enough to write a similar test with a cancel timeout set to match the same time as the scenario connect succeeds.

I'm not sure if this would work or not though... when two things are supposed to happen at "the same time" then in fact one of them ends up happening first, and it might not be the one we want :-).

Another approach, that's a little more complicated but that would definitely work, would be to set up a cancel scope around the whole call, and then arrange for the `FakeSocket.connect` method to cancel it just before it returns.

njsmith:
> it could also be fixed easily by moving the code around to do:

I don't think that would fix it... suppose that the cancellation happens after the `connect` succeeds. At that point it's too late to deliver a cancellation through the `await sock.connect` call. And there aren't any other `await`s inside the child task at that point, so the child task wouldn't see a `Cancelled` exception. The parent though is still blocked in `await nursery.__aexit__()`, so it will get the `Cancelled` and exit immediately without cleaning up `winning_sockets`.

But with the new code it's obvious that it handles all these cases correctly without needing any complicated analysis, so I'm not too worried.

> feel free to squash the last two fixup commits, I left them as is per nodejs suggestions

In general we're pretty relaxed about squashing-or-not. Keeping a nice clean history is nice, but I don't want people to feel like they need to spend a bunch of time fiddling with git to contribute.

trio/_highlevel_open_tcp_stream.py
```diff
@@ -242,71 +249,62 @@ def format_host_port(host, port):
...
 
-    # It's possible for multiple connection attempts to succeed at the ~same
-    # time; this list records all successful connections.
-    winning_sockets = []
+    # the socket that we're going to complete with, we need to make
+    # sure this is excluded from any cleanup
+    winning_socket = None
 
     # Sleep for the given amount of time, then kick off the next task and
     # start a connection attempt. On failure, expedite the next task; on
```
njsmith: This comment is a bit inaccurate now that the sleeping/task spawning has been moved into the parent – could you update?

njsmith:
:tada:

And per [our standard practice](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team), I'm sending you a Github invite now. No pressure of course, but if you want to keep contributing then you are very welcome! And [always happy to chat](https://gitter.im/python-trio/general) about what you're working on or what you might be interested in working on or... well, whatever, really :-)

PR #812, opened 2018-12-19
Remove APIs deprecated in 0.5.0

oremanj:
This covers `ClosedStreamError` and `ClosedListenerError` in the root of the trio package, and `Result`, `Value`, and `Error` in hazmat.


njsmith:
Awesome, thanks!

PR #814, opened 2018-12-22
Tweaks for #743

njsmith:
Thanks @jxub and @Fuyukai for handling #743 while I was swamped! I did
notice some tweaks I wanted to make in post-commit review, and finally
found the time to write them down... any thoughts?

----

- Move the exception from the main namespace into trio.ssl, since it's
  ssl-specific, and rename from NoHandshakeError → NeedHandshakeError
- Add the exception to the docs
- In the newsfragment, use the original issue number, and shorten
- Smooth out the docs a bit
- Add some more methods that require handshakes to the list


jxub:
@njsmith Looks good! I think that the new `NeedHandshakeError` is a better name indeed, and the descriptive exception message will help end users a lot in that corner-case.

njsmith:
@jxub in that case, want to do the honors?

Issue #819, opened 2018-12-23
Deadlocks caused by openssl's handling of TLS 1.3 session tickets


njsmith:
We have a test that tries running `SSLStream` over an unbuffered transport, in part to confirm we aren't making assumptions about buffering. It turns out that openssl 1.1.1/TLS 1.3 does make assumptions about buffering, and the test hangs. Now what?

Specifically: we [create an unbuffered transport, then call `do_handshake` on both the client and server sides](https://github.com/python-trio/trio/blob/5a03bee3fbf1a5364a3584ef8150b04143dcecdb/trio/tests/test_ssl.py#L762-L771). The `client.do_handshake()` completes successfully. The `server.do_handshake()` hangs.

I believe what's happening is that when openssl 1.1.1 is doing TLS 1.3, the server automatically sends some session tickets after the handshake completes. Technically these aren't part of the handshake, so the `client.do_handshake()` call doesn't wait for them. But on the server side we don't know this; all we can see is that we called `SSL_do_handshake` and it spit out some data to send, so we assume that the handshake won't complete until we send that data.

I guess there are two scenarios that the designers imagined: (1) the client is trying to read application-level data, so it's reading from the transport, and it gets the session tickets as a bonus. (2) the client isn't trying to read application-level data, so it's not reading from the transport, but the transport has enough intrinsic buffering that the session tickets can just sit there forever without causing any problems. But in our test neither is true, so it breaks.

As far as I know, this shouldn't have too much real-world impact, since real transports almost always have some buffering. (Empirically the session tickets appear to be 510 bytes, which is smaller than pretty much any real transport buffer.) And, there is a workaround available: modify `clogged_stream_maker` so the server does `await server.do_handshake(); await server.send_all(b"x")`, and the client does `await client.do_handshake(); await client.receive_some(1)`. Sending a bit of application level data from server→client like this should flush out the session tickets and get back to the state we want for the rest of the tests.

Nonetheless, it's not ideal that openssl is making subtle assumptions about buffering like this. Shoudl we do something?

Unfortunately, TLS 1.3's decision to move session tickets out of the handshake doesn't give us a lot of options here. We could disable session tickets, but that seems sub-optimal (and currently the `ssl` module doesn't expose a knob for this anyway).

[...]

Issue #817, opened 2018-12-23
Umbrella issue: openssl 1.1.1/TLS 1.3 support


njsmith:
Right now, if you run the trio testsuite on Ubuntu 18.10, the ssl tests fail, because Ubuntu 18.10 has openssl 1.1.1.

There are several aspects to this that I'll split into separate issues:

* [x] We should test against openssl 1.1.1 in CI, so we can tell whether it's working: #818
* [x] The way openssl 1.1.1/TLS 1.3 handles session tickets causes `test_SSLStream_generic` to hang: #819
* [x] There are new constants exposed in the ssl module, causing `test_exports.py` to fail
* [x] openssl now raises `SSLZeroReturnError` in several situations where it didn't before, and we need to handle them properly
* [x] we should make sure the renegotiation tests properly limit themselves to TLS v1.2. There's some code that tries to do that, but it looks like `OpenSSL.SSL.OP_NO_TLSv1_3` is missing, so it probably doesn't work?
* [x] we should possibly run at least some of the tests with both TLS v1.2 and TLS v1.3 on openssl v1.1.1? We do automatically get some coverage of both protocols by testing on multiple openssl versions, but v1.1.1 might be different.

Also, while not required to get the test suite passing, it would be good to:

* [ ] Support early data: #223 
* [ ] Support post-handshake auth: #741

njsmith:
All the top-line boxes here are checked off now, and the other items have their own dedicated issues, so I guess we can close this.

PR #820, opened 2018-12-23
Fix some test failures on openssl 1.1.1 / Ubuntu 18.10

njsmith:
See https://github.com/python-trio/trio/issues/817

We don't have CI for this yet, but at least on my laptop it helps.
However, it doesn't fix everything.

- Add OP_ENABLE_MIDDLEBOX_COMPAT to the list of static ssl constants
- Handle SSLZeroReturnError in more places
- Fix some -- but not all -- of the places where session tickets cause
  failures. See https://github.com/python-trio/trio/issues/819

Not sure if we should merge this as is, to at least cut down some of
the issues, or wait to try to deal with more of the session ticket
mess...


njsmith:
Closing in favor of #841

Issue #147, opened 2017-05-02
Add shield argument to move_on_after and other convenience functions


merrellb:
The documentation suggests assigning a cancel scope and setting shield as two separate steps:
```
with move_on_after(CLEANUP_TIMEOUT) as cleanup_scope:
    cleanup_scope.shield = True
    await conn.send_goodbye_msg()
```
However ```trio.open_cancel_scope``` has a shield argument and it seems like it would also make sense to add this to the convenience functions (```trio.move_on_after, trio.move_on_at, trio.fail_after and trio.fail_at```) as well.

njsmith:
For the record, the reason I didn't do this in the first place is that I feel like `shield` should be a pretty obscure and rarely used feature. It's​ very easy to use in a way that breaks your program in subtle ways (e.g. by making it unresponsive to rare network failure conditions), it breaks the rule that timeout policy should be set as far up the stack as possible, and I don't want it to become an attractive nuisance. So I included it in `open_cancel_scope` since that's a kind of low level "raw" API, but left it out of the high level timeout apis.

Maybe it will turn out the I'm wrong and `shield` really is something that gets used a lot, but I can't help but feel that that would indicate that we've messed up somehow.

In the mean time, I'm sort of feeling like maybe we should remove the kwarg from `open_cancel_scope` :-). It *should* be a bit annoying to use...

While we're at it, here's another idea: maybe we can have a way to say "yes, I know I'm cancelled, but please stop raising it unless an operation *actually* blocks". Like a shield with a zero second per operation timeout. So you could attempt to do things like send a goodbye message, but if it doesn't succeed immediately, oh well, let's move on. It would at least remove the issue of having to pick some magic timeout value way down inside the code...

... Though, I guess this doesn't actually help for your use case with websockets, since the main reason websockets have a goodbye message is so the initiator can then wait until they see the corresponding goodbye message from their peer, and know that both sides successfully processed all the data that was sent. (Since otherwise TCP can lose some data at the end of a connection.) Sending a goodbye message and then immediately closing the socket is pretty useless, but that's all my idea above would allow for.

[...]

PR #833, opened 2019-01-01
Subprocess support: high-level interface

### LINKED ISSUE #822: trio.run_process
So #791 added the core support for subprocesses, but at the last minute we had doubts about copying `subprocess.run` and decided to split out the high level convenience API into a separate PR: https://github.com/python-trio/trio/pull/791#issuecomment-449325056

Some issues to consider (it's worth reading the whole discussion linked above):

* probably make `check=True` as the default. (Might be worth considering making the name more descriptive, so people would write `allow_failure=True`?)
* Passing `stdout=PIPE` to mean "capture the output" is pretty confusing (basically just dumping an implementation detail into the public API). Maybe `capture_stdout=`/`capture_stderr=` as boolean kwargs? Should we have `capture=` as a shorthand for setting both at once?
* If we're changing the default for `check`, we should change the name, because that's just enough of a change for people to trip over. `trio.run_process` is a good name.
* Should we do anything about the weird mess around `shell=True` and string-versus-list commands?

oremanj:
Closes #822.

- Add `NullStream` to the generic streams interface, which acts like `/dev/null`
- Improvements to `Process`, mostly needed or wanted by higher-level functions:
  - Move from `trio.subprocess` to the base `trio` module, so `trio.subprocess` is just a reexport of constants and exceptions
  - Forbid lists with `shell=True` and strings with `shell=False` on UNIX due to the likelihood of user confusion
  - Add constructor options `shutdown_signal` and `shutdown_timeout` to allow orderly shutdown of subprocesses
  - Fix support for multiple tasks calling `wait()` simultaneously
  - `cancelled` attribute to track whether a subprocess was terminated due to `aclose`/etc cancellation
  - `failed` attribute to track whether a subprocess exit failure was "our fault" (if we killed it or closed its output streams before it had exited)
  - `join` method to do `aclose` without closing the pipes
  - `shutdown` method to do a cancelled `join`, since with `shutdown_signal`/`shutdown_timeout` this might be more complex than just `kill()`
- Add `ProcessStream`, a single `HalfCloseableStream` wrapper around `Process`
- Add `open_process`, a context manager that provides a `ProcessStream` and can raise `CalledProcessError` if the process fails
- Add our own `CompletedProcess` with `cancelled` and `failed` attributes and consistent naming of the command (we use `command` vs the stdlib `args`)
- Add `run_process`, like `subprocess.run` but with better defaults
  - Instead of the previous `timeout` and `deadline` arguments, add `preserve_result=True` option to return the `CompletedProcess` even if cancelled; this seems less invasive

I don't necessarily expect to merge this wholesale, but I wanted to provide a complete and polished interface as an example input to debate about what the interface ought to be. :-) Especial points of uncertainty for me include the advisability of adding `ProcessStream` instead of just having `open_process` return a `Process`; whether the process exiting failure should cancel the `open_process` block; and whether `preserve_result` is in fact a less invasive API choice than the previously proposed `timeout`/`deadline` arguments.


trio/_subprocess.py
```diff
@@ -155,67 +242,718 @@ def __init__(
...
+        Accessing this attribute does not check for termination; use
+        :meth:`poll` or :meth:`wait` for that.
         """
         return self._proc.returncode
 
+    @property
+    def failed(self):
+        """Whether this process should be considered to have failed.
+
+        If the process hasn't exited yet, this is None. Otherwise, it is False
```
Fuyukai: This seems tricky and error-prone (having it be True, False or None)

oremanj: I feel like this is the convention I usually see for "optional bool", and it mirrors the `returncode` attribute (which is None or an integer). Would you prefer that accessing `.failed` throw an exception if the process is still running?

oremanj:
The reported coverage miss is on the blank line after `test_not_a_failure_if_we_signal()` in test_subprocess.py, which seems questionable to me. I also saw a single stray coverage line on #835 - it was a comment there. Any clues for fixing would be appreciated.

oremanj:
Thanks for the feedback. I understand that there's a lot here and I appreciate the backpressure looking for cleaner ways to do things. Can I ask though that you read the documentation changes included in this PR before making decisions about whether something is worth the user-facing complexity or not? I put most of my explanatory effort there, not in the commit message, because docs reach a good deal more people than the commit messages do. :-)

> > Add NullStream to the generic streams interface, which acts like /dev/null

> Do we really need this? Why?

NullStream can be removed if we remove ProcessStream (I only needed it for `ProcessStream.stderr`). I think I prefer the `.stdio` solution there anyway.

> > Move from trio.subprocess to the base trio module, so trio.subprocess is just a reexport of constants and exceptions

> Let's split this out into #852 so we can discuss the trade-offs properly.

Sure. There doesn't seem to be anyone opposed, though, so maybe there's not much to discuss?

> > Fix support for multiple tasks calling wait() simultaneously

> Oh yeah we definitely want this! Again, maybe a good quick standalone PR?

#854 

> > Forbid lists with shell=True and strings with shell=False on UNIX due to the likelihood of user confusion

> Good idea. Maybe submit this as a standalone PR?

This sort of goes along with the section I added to the docs about quoting, and when trying to split it off I realized it's going to be troublesome to isolate those doc changes from the "we're not just reexporting `subprocess`" ones. So I'm going to leave it for the moment; happy to revisit later when we've decided more about the other stuff.

> > Add ProcessStream, a single HalfCloseableStream wrapper around Process
> > Add open_process, a context manager that provides a ProcessStream and can raise CalledProcessError if the process fails

> Maybe Process objects should simply have a .stdio attribute, which is StapledStream(self.stdin, self.stdout) if those are both Streams, and None otherwise? Would that be enough to remove the need for all this?

We can use that in place of `ProcessStream`, sure. I think there are still other reasons to have `open_process`, though, rather than just the base `Process`:
* better defaults (raise on failure and attach pipes to stdin/stdout/stderr out of the box)
* it can only be used as a context manager, which encourages better structured-concurrency type architecture, which I think is in line with Trio's goals

> > Add constructor options shutdown_signal and shutdown_timeout to allow orderly shutdown of subprocesses
> > cancelled attribute to track whether a subprocess was terminated due to aclose/etc cancellation
> > failed attribute to track whether a subprocess exit failure was "our fault" (if we killed it or closed its output streams before it had exited)
> > join method to do aclose without closing the pipes
> > shutdown method to do a cancelled join, since with shutdown_signal/shutdown_timeout this might be more complex than just kill()
> This is... a lot of stuff. What is it all for? Why should we believe that people will use it?

Shutdown stuff: If you start a subprocess that does nontrivial work, depending on the application it might be considered pretty rude to SIGKILL it without giving it any prior notice that you'd like it to clean up and exit. You might wind up with temporary files still on disk, sub-subprocesses getting reparented to init and running forever, etc. Probably whatever mechanism we use for grace periods (#147) could (should?) replace the explicit specification of a `shutdown_timeout`. I think it's still pretty important to be able to specify what signal gets used, though, since there's not one good choice for everyone: the venerable UNIX default is SIGTERM, but Python programs will out-of-the-box respond better to SIGINT, etc. (I don't feel nearly as strongly about supporting `shutdown_signal=0` meaning "close pipes and wait"; I added it mostly because it's pretty much the only way we can handle this at all for Windows. I don't actually care that much about Windows but I gather that some people do :P)

`Process.cancelled`: This is mostly (entirely?) to give `run_process()` something to fill in `CompletedProcess.cancelled` from. It could be made private.

`Process.failed`: Making errors loud by default is great! But if the user explicitly kills a process, I think they will be confused and saddened when they get a CalledProcessError out of that. And simply falling off the bottom of the `open_process()` context manager without reading EOF from both stdout and stderr makes it somewhere between possible and likely that the process will exit with a broken pipe error. I think it's better to have our determination of "should we throw an exception about this process having failed?" be fairly nuanced, than to have it be misled by common user actions in a way that might cause them to not check the return code at all.

`Process.join`: is mostly to avoid code duplication for the benefit of `open_process` and `run_process`; could easily be made private if you think it's too much clutter in the public API.

> > Instead of the previous timeout and deadline arguments, add preserve_result=True option to return the CompletedProcess even if cancelled; this seems less invasive

> Huh, that's a really interesting idea! Normally we're all dogmatic, Thou Shalt Not Catch A Cancelled Exception, but this is ... actually maybe a good reason to break that rule? Or would it be too error-prone in cases like: [...]

It's definitely a power user feature, and might merit a warning in the docs, and/or a suggestion to put an explicit checkpoint after the call. I don't mind going back to timeout/deadline if you think that's less confusing, but I do think we should have some way to support this use case. Maybe that works out to "we need a general way to report partial progress before a cancellation, and then run_process() can just use that".

I appreciate the simplicity of Trio's stance that "a cancelled operation didn't happen", but it doesn't necessarily compose very well -- if an operation is built out of multiple other underlying operations that can't readily be rolled back, either the "cancelled = didn't happen" rule has to break or the entire higher-level operation has to be uncancellable once started. I don't think we want to propose the latter, so maybe we should think about a user-friendly way to talk about the circumstances in which the rule gets bent?

njsmith:
> Can I ask though that you read the documentation changes included in this PR before making decisions about whether something is worth the user-facing complexity or not? I put most of my explanatory effort there, not in the commit message, because docs reach a good deal more people than the commit messages do. :-)

Ah cool, thanks for the heads up :-). Read them now.

At a high level: I think it's great you put this mega-PR together. Like you say in the first comment, it's great for identifying issues. Now that we have this laid out though, I think a divide-and-conquer strategy is going to work best for actually getting changes in :-). So I'm going to be looking for all the uncontroversial bits we can carve off and merge on their own.

In that vein, I see three "obvious" PRs, beyond the ones you already submitted. I'll discuss the other topics more below too, but these are probably the next things to actually work on :-):

* [ ] Adding `Process.stdio`
* [ ] Adding the prohibition on `shell=True`. I see what you mean about the docs, and I do like the idea of the trio docs teaching people general facts about how processes work, not just the minimal details of our API :-). But, this is such an obscure and useless edge case that I think it's OK to put the change in by itself, and then worry about teaching people about quoting in general later. In the unlikely case that anyone runs into this, the error message will explain it well enough.
* [ ] The "minimal core" of `run_process`, that doesn't (yet) do anything clever with cancellation/timeouts.

> I think there are still other reasons to have open_process, though, rather than just the base Process:
> * better defaults (raise on failure and attach pipes to stdin/stdout/stderr out of the box)
> * it can only be used as a context manager, which encourages better structured-concurrency type architecture, which I think is in line with Trio's goals

I feel like I still have a lot of questions here:

* Is capturing stderr the right thing to do? In the cases I'm thinking of like git hook programs, you usually want to let stderr go to the terminal, I think? And if you capture it but don't realize, and don't start a task to take care of reading from it, then when an error occurs you'll get nothing at best and a deadlock at worst, which is definitely not ideal...
* A number of the subtopics in this thread deal with the complexities around process shutdown and what should count as an error... for `run_process` it's simpler, but raising out of a `__aexit__` method is always fraught (e.g. should you do something different if you got there via an exception?). Do we understand it well enough to implement a `check=True` equivalent for a context manager like this, that actually DWIM? And if we do, would it make sense to add it to `Process` directly?
* We do like context managers around here, but the `Process` API already gives users the choice of either using a context manager or not. So even if *this* API makes the context manager mandatory, from the user perspective it's still optional. Is the value from the other `open_process` features enough to make this nudge helpful?
* [Auto-cancellation on process failure](https://github.com/python-trio/trio/pull/833/files#diff-051add6309baca2f3b2c7b870c816c44R590) seems dubious...
* Is spawning a child and interacting with it like this a common enough use case, and does this reduce the complexity of that use case, enough to justify adding this to the top-level trio namespace?

So to me this feels like something to come back to after we've dealt with all the easier stuff...

> Shutdown stuff

Okay yeah this is interesting and complicated. I feel like there are a few interlocking cases here: what `Process.aclose()` does normally, what `Process.aclose` does when cancelled, what `run_process` will do when cancelled... this feels like it deserves its own discussion thread? And we might want to defer some of it until after #147 has progressed more.

> `Process.cancelled`, `Process.failed`, `Process.join`

It sounds like these are all mostly helpers for other features, so we can defer worrying about them until we figure out those other features :-)

> I appreciate the simplicity of Trio's stance that "a cancelled operation didn't happen", but it doesn't necessarily compose very well -- if an operation is built out of multiple other underlying operations that can't readily be rolled back, either the "cancelled = didn't happen" rule has to break or the entire higher-level operation has to be uncancellable once started. I don't think we want to propose the latter, so maybe we should think about a user-friendly way to talk about the circumstances in which the rule gets bent?

Yeah, that stance is good (necessary) for fundamental operations like `Lock.acquire` or `sock.recv`, but it doesn't generalize. This also sounds like a good topic for its own discussion thread?

njsmith:
@oremanj I think this was superseded by #872? I'm going to mark it closed based on that belief, but it might be worth checking if there's anything else you want to salvage out of it.

Issue #320, opened 2017-09-06
Maybe move_on_after(0)/fail_after(0) should immediately cancel their scope


njsmith:
It's tempting to think that you can fake a "check once, but don't block" semantics by setting up a 0-timeout cancel scope:

```python
async def fake_accept_nowait(self):
    with fail_after(0):
        await self.accept()
```

This is not how Trio's cancellation semantics are supposed to work – there's no guarantee that `accept` won't raise `Cancelled` before it even attempts the operation.

@sorcio points out in chat that this actually works surprisingly well though. In fact, it probably succeeds deterministically right now (!), because `fail_after(0)` schedules its cancel scope to be cancelled "immediately", but since it uses a timeout to do this it doesn't actually happen until Trio processes timeouts, which it doesn't do until after this task yields. And if you look at `trio._socket._SocketType.accept`... it doesn't yield until after trying a non-blocking `accept`. So right now things work by accident.

This is dangerous, because this is a tempting thing to try, and the sort of thing where people might end up depending on it because it seems to work. I've been tempted and stopped b/c I thought it through and realized it wouldn't work – but we can't expect everyone to have my level of familiarity with the subtleties of Trio's semantics :-) – and @sorcio's tried it and found it worked...

So, if we want to keep the current semantics, maybe we should special-case `move_on_after(0)` and `fail_after(0)` to *immediately* cancel the scope they create, instead of going through the normal timeout machinery. Of course it would still be possible to get a similar effect with `move_on_after(1e-12)`, but then at least it's obvious that you're doing something potentially race-y.

Alternatively, could we change our semantics to make this work? The obvious problem would be IOCP on Windows. If you use IOCP to implement `accept`, then the underlying operation is actually "hey Windows, please start an `accept` running in the background", and then blocking and going through the event loop to check if it's done. In practice, if an `accept` is ready immediately, then it'll probably complete either synchronously (IOCP operations are allowed to do that) or at least before we notice the timeout and call `CancelIoEx`. But it makes me nervous? I dunno, maybe this would actually be fine.

The bigger problem is that the semantics of "don't deliver cancellations until a task actually *blocks* rather than just checkpoints" are fundamentally problematic: if you have some heavy computational work to do in the background, then the standard way is to go ahead and do it and just make sure to checkpoint frequently so other tasks can run. If we start delaying cancellations in general, then it becomes impossible to cancel a task like this.

[...]

PR #835, opened 2019-01-02
Add support for unbound cancel scopes

### LINKED ISSUE #607: idea: unbound cancel scopes

oremanj:
Add support for unbound cancel scopes as discussed in #607: the ability to create a `CancelScope` without entering it, so you can pass it to another task and retain a handle for cancelling some work in that other task.

This change exposes `trio.CancelScope` as a regular class that users can call to construct cancel scopes, which obviates the need for `trio.open_cancel_scope`. The latter is therefore deprecated.


oremanj:
The remaining reported coverage miss is on _run.py line 236, which is a comment. I can't reproduce locally - any idea what might be going on with that?

oremanj:
This diff mostly keeps the behavior of "deadlines don't result in calls to cancel() except on scopes that have at least one task attached to them". (It adds a bit of logic to also call cancel() on deadline expiry for scopes that have linked children, but that's unrelated to this question.) I'm not sure if keeping that behavior is necessarily a good thing, because it means that deadline expiries before the scope is entered still don't take effect immediately:
```
scope = trio.move_on_after(0.5)
await trio.sleep(1)
with scope:
    assert not scope.cancel_called
    await trio.hazmat.cancel_shielded_checkpoint()
    assert scope.cancel_called
```
I see three reasonable options here:
* Keep the behavior of this diff as it stands.
* Remove the logic that keeps cancel scopes out of `Runner.deadlines` if they have no tasks attached. This will require extra bookkeeping by the run loop to cancel scopes that have already completed successfully (because they're reusable, even if they usually won't get reused), but arguably provides more intuitive behavior if the scope's deadline expires before it's entered, and preserves the current behavior for things like `with trio.move_on_after(0)`. This would technically still be a behavioral change in terms of when `cancel_called` gets set, but that's less likely to be user-visible than option 3.
* Continue to keep cancel scopes out of `Runner.deadlines` if they have no tasks attached (and no linked children), but add logic to cancel the scope immediately upon entering its context if its deadline is in the past. This would make `trio.move_on_after(0)` immediately cancel its scope, which is a behavioral change, but one we've previously discussed being maybe desirable: #320 

Thoughts?

smurfix:
> I think there's a serious bug in the use of weakrefs: if you have B = A.linked_child(); C = B.linked_child(); del B then now A.cancel() won't cancel B.

The way to fix this is to have a non-weak `.link_parent` entry which links C back to B back to A and thus keeps B alive until C's refcount is zero.

njsmith:
Oh another note for the future linked scopes PR: updating `current_effective_deadline`

njsmith:
Oh, and regarding the issue of whether entering a scope with a deadline in the past should trigger an immediate cancellation:

I think there are arguments either way. But, there is also a nasty little subtlety that makes changing this more complicated than you'd expect. Consider:

```python
with move_on_after(0.5):
    # Then we do other stuff for a while, without checkpointing
    time.sleep(1)
    # Then we checkpoint, which we expect to cause the whole block to be aborted, because the
    # outer scope's deadline has expired
    await checkpoint()
    print("this should not be printed!")
```

Right now, the text is not printed. But if we made it so we checked deadlines when entering a cancel scope, it would be (!). That's because `checkpoint` is defined as:

```python
with CancelScope(deadline=-inf):
    await sleep_forever()
```

Right now, deadlines are all checked at the same time, in the main run loop, and we do some elaborate tricks to make sure that if two cancel scopes expire at the same time, then it's the outer one that's cancelled.

If we changed `CancelScope.__enter__` to check the deadline immediately, that would break this batching... so `checkpoint` would effectively become

```python
with CancelScope() as cs:
    cs.cancel()
    await sleep_forever()
```

and now in our original example *this* scope will cancel the sleep first, before the outer scope, so the message *will* be printed.

Anyway.... this is all a bit of a mess. I don't think we should try messing with it here in this PR. I suspect we should simplify it somewhat in any case. I'll open an issue for that.

oremanj:
Thanks for the commentary! I'll do this split and respond in more detail in a bit.

But first, I was looking at this in the context of #860 and noticed an additional subtlety:
```
>>> from trio import move_on_after
>>> from trio.hazmat import checkpoint
>>> import time
>>> with move_on_after(0.5):
...     # Then we do other stuff for a while, without checkpointing
...     time.sleep(1)
...     # Then we checkpoint, which we expect to cause the whole block to be aborted, because the
...     # outer scope's deadline has expired
...     await checkpoint()
...     print("this should not be printed!")
... 
>>> with move_on_after(0.5):
...     time.sleep(1)
...     await trio.hazmat.checkpoint_if_cancelled()
...     await trio.hazmat.cancel_shielded_checkpoint()
...     print("but should this?")
... 
but should this?
>>> 
```
And this isn't just an academic distinction - there's lots of code in trio that uses checkpoint_if_cancelled followed by cancel_shielded_checkpoint to do a checkpoint when it knows it doesn't need to block indefinitely. It's already documented that deadline expiry might not be noticed immediately. So, I'm not sure that "this should not be printed" getting printed in the above example is actually as bad as it looks?

If we don't want either print statement in the above examples to execute, I'd suggest:
* have `checkpoint_if_cancelled()` check whether the current effective deadline is less than the current time, rather than whether there's a pending cancel scope
* implement `checkpoint()` in terms of the two half-checkpoint operations, or as its own trap, rather than yielding inside a scope with a -inf deadline

trio/_core/_run.py
```diff
@@ -116,28 +118,172 @@ def deadline_to_sleep_time(self, deadline):
...
+        if self.cancel_called:
+            state = ", cancelled"
+        elif self.deadline == inf:
+            state = ""
+        else:
+            now = current_time()
+            if now >= self.deadline:
+                state = ", cancel soon"
+            else:
+                state = ", cancel in {:.2f}sec".format(self.deadline - now)
```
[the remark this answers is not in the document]
oremanj: I agree this is cleaner. I implemented a small variation: "deadline is X seconds from now" or "deadline is Y seconds ago".

oremanj:
I've removed the linked-child changes from this PR, and will repost them as a new PR once this one gets resolved.

> I think the interaction of shields and linked children needs a serious rethink... in fact maybe in between the two changes we should split off shielding to a different object, as suggested in the #607 thread.

The more I play with cancellation, the more I'm feeling like the ability to modify shielding dynamically is an _extremely_ powerful and useful feature that I don't want to give up. :-) If we drop it, I think we lose the ability to prototype a number of changes to cancellation semantics outside of _core. For example, I used both dynamic shielding and linked children in the "cancel nursery body before the rest of the children" implementation in https://github.com/python-trio/trio/issues/147#issuecomment-455824598. It might be possible to get the same semantics without dynamic shielding changes, but I think it would be a lot more cumbersome.

njsmith:
Yeah, the discussion of cancel-ordering [in #147](https://github.com/python-trio/trio/issues/147#issuecomment-453561373) made me think dynamic shielding may be important, too! But it could be dynamic while still being a separate object, and anyway, this PR probably isn't the best place to talk about it because we'll lose track once it's merged :-)

trio/_core/tests/test_run.py
```diff
@@ -823,20 +844,74 @@ def test_cancel_points():
...
+        )
+        async with _core.open_nursery() as nursery:
+
+            async def try_in_other_task():
+                with pytest.raises(RuntimeError) as exc_info:
+                    with scope:
+                        pass  # pragma: no cover
+                assert "while it is already active in another task" in str(
+                    exc_info.value
+                )
```
njsmith: This test feels a bit odd, given that the scope *is* active in the child task that tries to re-enter it (because it inherited it) :-). Maybe it would be clearer to do something like:

```python
async def enter_scope():
    with scope:
        await sleep_forever()

    nursery.start_soon(enter_scope)

    with pytest.raises(...):
        with scope:
            pass  # pragma: no cover

    nursery.cancel_scope.cancel()
```

docs/source/reference-core.rst
```diff
@@ -494,14 +494,11 @@ but *will* still close the underlying socket before raising
...
+by Trio are ultimately implemented in terms of :class:`CancelScope`
+objects.
 
-.. autofunction:: open_cancel_scope
-   :with: cancel_scope
-
-Cancel scope objects provide the following interface:
-
-.. interface:: The cancel scope interface
+.. autoclass:: trio.CancelScope
```
njsmith: Normally in trio, we put the detailed class/method/property docs into the code as docstrings, and then import them via `autoclass` and friends.

Historically, we haven't done that for `CancelScope`, because of this hacky `.. interface:: The cancel scope interface` thing. But now that it's a regular class, maybe we should move the docs into regular docstrings?

oremanj: I did so for `deadline`, `shield`, and `cancel()`. Since `cancelled_caught` and `cancel_called` are regular attributes, not properties, there's not a great place to document them; and using `Attributes:` in the class docstring seems to force those attributes to show up first in the Sphinx output, which isn't what we want here.

njsmith: Works for me

njsmith:
Looks great!

Issue #239, opened 2017-07-02
Tools for finding scheduler-dependent "heigenbugs"


njsmith:
Race conditions suck, but they're a common kind of bug in concurrent programs. We should help people test for them if we can.

Some interesting research out of MSR:

* [CHESS](https://www.microsoft.com/en-us/research/publication/finding-and-reproducing-heisenbugs-in-concurrent-programs/)
* [GAMBIT](https://www.microsoft.com/en-us/research/publication/gambit-effective-unit-testing-for-concurrency-libraries/)

CHESS assumes you have a preemptive multitasking system and works by hooking synchronization primitives, and explores different legal paths through the resulting happens-before graph. This makes it tricky to find data races, i.e., places where two threads access the same variable without using any synchronization primitive; you really want to be able to try schedules where threads get preempted in the middle of these. In practice, it sounds like the way they do this (see page 7) is to first run some heavyweight data race detector that instruments memory reads and writes, and then use the result to add new annotations for the CHESS scheduler.

For us, we don't AFAIK have any reasonable way to do data race detection in Python (maybe you could write something with PyPy? it's non-trivial), but we do have direct access to the scheduler, so we can in principle explore all possible preemption decisions directly instead of having to infer them from synchronization points. However, this is likely to be somewhat inefficient – for CHESS it *really* needs to cut down the space of preemption points because otherwise it has to consider a preemption at every instruction which is ridiculously intractable, and we're in a better position then that, but the average program does still have lots of cases where two tasks that aren't interacting at all have some preemption points and this generates an exponential space of boring possible schedules.

They have a lot of interesting stuff in the paper (section 4) about how to structure a search, recovering from non-determinism in the code-under-test, some (not enough) detail about fair scheduling (they want to handle code with spin-locks). It looks like [this is the fair scheduling paper](https://pdfs.semanticscholar.org/adec/c5f64d69031ee5048bac09d1da1fc7811e75.pdf); it looks like the main idea is that if a thread calls `sched_yield` or equivalent then it means it can't make progress, so you should lower its priority. That's... not a good signal in the trio context. Fortunately it's not clear that anyone wants to write spin-locks anyway... (we use the equivalent a few times in our test suite, but mostly only in the earliest tests I wrote before we had much infrastructure, and certainly it's a terrible thing to do in real code).

[...]

PR #840, opened 2019-01-06
Use module-level Random to shuffle tasks

Zac-HD:
This allows test helpers etc. to reproduce scheduling-dependent heisenbugs by seeding the `Random` instance.  (or at least ensures that *something else* will prevent reproductions instead)

I have draft patches against `pytest-trio` and Hypothesis to take advantage of this, but since this has no observable impact except to enable those future patches I thought this was a good place to start.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -671,7 +671,6 @@ class Runner:
 
     runq = attr.ib(default=attr.Factory(deque))
     tasks = attr.ib(default=attr.Factory(set))
-    r = attr.ib(default=attr.Factory(random.Random))
 
     # {(deadline, id(CancelScope)): CancelScope}
     # only contains scopes with non-infinite deadlines that are currently
@@ -1433,7 +1432,7 @@ def run_impl(runner, async_fn, args):
         # ordering of task._notify_queues.)
         batch = list(runner.runq)
         runner.runq.clear()
-        runner.r.shuffle(batch)
+        _r.shuffle(batch)
         while batch:
             task = batch.pop()
             GLOBAL_RUN_CONTEXT.task = task


njsmith:
Can you link to the issue thread where this was discussed, for context? (Not necessarily in the code, just here in the PR is fine.)

Actual change LGTM

Zac-HD:
Discussed here: HypothesisWorks/hypothesis#1709, and related to #239.

I considered a few other ways to make the scheduler deterministic-on-demand:

- This option, allowing Hypothesis to seed a random shuffle, is the smallest change to Trio and therefore leaves our future options open - the public API doesn't change at all.  It can be made flatly deterministic by default, or varied-but-reproducible with the `random_module` strategy (once the Hypothesis end lands, anyway).
- Adding a `seed=` argument to `trio.run` would involve public API that we might not want to keep (e.g. if we do move to fair or other deterministic scheduling later).  It would also be an *enormous* pain to connect it to Hypothesis for the varied-but-reproducible option.
- Adding an `order_tasks: callable=...` argument to `trio.run` would be very, very flexible but have all the same problems as `seed=` for this purpose.

njsmith:
> Adding a `seed=` argument to `trio.run` would involve public API that we might not want to keep (e.g. if we do move to fair or other deterministic scheduling later). It would also be an enormous pain to connect it to Hypothesis for the varied-but-reproducible option.

I'm fine with the current PR, but noting here:

The public API is an issue, but not a huge one: in the worst case, if we remove randomization, then the argument would just become a no-op, which would be totally backwards compatible. ("You asked for determinism, and you got determinism!") We could then deprecate it as lazily as we felt like.

For the "enormous pain" part: are you imagining that Hypothesis will have hard-coded knowledge of Trio, and automatically patch the seed here whenever Hypothesis and Trio are both loaded in the same process? If so then I agree :-). But in the other design, where Hypothesis remains ignorant of Trio and we leave it to pytest-trio to hook them up, then a `seed=` argument seems like it should be pretty easy to handle – pytest-trio already knows when the test is being run under hypothesis, and pytest-trio is already responsible for calling `trio.run`, so it seems like it should be pretty straightforward to plumb them together.

...I guess the version where Hypothesis has hard-coded knowledge of Trio is kind of compelling though, if it's acceptable to the Hypothesis devs. Not everyone uses pytest-trio, and even if you use pytest-trio you might still have some tests where you call `run` by hand, and I have no idea how stateful tests would fit into this, etc.

Another advantage of *not* exposing `seed` is that it won't tempt anyone to use a fixed `seed` to "solve" race conditions in their code :-). Though I doubt this will be an issue in practice...

Zac-HD:
Our policy for Hypothesis is that the core engine *cannot* have any knowledge of other libraries; the interface layer can have generic hooks (e.g. I added one for pytest-trio which is now also used by pytest-asyncio), and the strategy layer can have whatever dependencies *if* it's under the `hypothesis.extra` namespace.

So pytest-trio could easily inject a static `seed=` argument, but the approach I'm going with also supports varied-but-reproducible seeds via the `random_module` strategy.  From the Hypothesis side, it's also usable for simulations or anything else where an existing `Random` instance needs to be made deterministic.  

This approach will also work without pytest-trio; you just need to do some (idempotent) setup by calling `hypothesis.strategies.random_module.register(trio._core._run._r)` before running the test.  The pytest-trio implementation is literally just to do this setup when pytest imports the module :smile:

njsmith:
> So pytest-trio could easily inject a static `seed=` argument, but the approach I'm going with also supports varied-but-reproducible seeds via the random_module strategy.

Oh, I was imagining that it would sample the seed from some strategy, not use a static one.

But, anyway, I'm happy :-) Let's go ahead with this and we can always make it more complicated later if we need to.

Zac-HD:
> Oh, I was imagining that it would sample the seed from some strategy, not use a static one.

It does that for the stdlib, Numpy, and any registered random instances *iff* you use the `random_module` strategy - otherwise we run into complications about how to display failing examples etc.  

Otherwise, we use a static seed to ensure it's still reproducible.  In principle it would be nice to make using the random module without the corresponding strategy into a healthcheck error, but when we did that a few years ago it was a terrible user experience for not much gain.

njsmith:
>  In principle it would be nice to make using the random module without the corresponding strategy into a healthcheck error, but when we did that a few years ago it was a terrible user experience for not much gain.

Huh, interesting! The more you know :stars:

PR #850, opened 2019-01-11
Add support for poisoning channels

oremanj:


njsmith:
My problem here is that I'm just not sure if this is a feature that we want to have or not. Adding a whole new ABC is a fair amount of API weight. But OTOH mandating that all channels support this isn't obviously good either. (I still have a vague dream that we can use channels to model common stuff like websockets (`Channel[Union[str, bytes]]`) or sans-io protocols (`h11_channel = H11Channel(transport_stream): Channel[h11.Event]`)... I really don't know if this will turn out to be useful or not, but it's definitely incompatible with having poison as part of the base `Channel` API.) Or we could make it just a feature of memory channels, not part of the base interface? Or maybe we don't really need it at all? And in most cases if people really need this feature, they can add it as an ad hoc layer on top of a memory channel by defining their own "poison" object, so it shouldn't cause too much harm if we move slow here.

So... we should keep talking about this, and I definitely appreciate the work you put into this – we should keep it as a reference for that discussion! But I don't think we should merge anything until we have more experience with channels in general.

PR #854, opened 2019-01-13
Process: add a lock so that multiple tasks can wait() simultaneously

oremanj:
Previously this would fail on kqueue platforms.


trio/tests/test_subprocess.py
```diff
@@ -47,6 +47,9 @@ def got_signal(proc, sig):
 async def test_multi_wait():
     async with subprocess.Process(SLEEP(10)) as proc:
         async with _core.open_nursery() as nursery:
+            with move_on_after(0.1):
+                await proc.wait()
+
```
njsmith: The timeout here is a bit sloppy... it slows down the tests slightly, and if the OS scheduler decides it hates us then there's no guarantee that `proc.wait` will have reached the actual blocking point before the timeout fires. How about we do two rounds:

```python
# Check that wait (including multi-wait) tolerates being cancelled
    nursery.start_soon(proc.wait)
    nursery.start_soon(proc.wait)
    nursery.start_soon(proc.wait)
    nursery.cancel_scope.cancel()

# Now try waiting for real
    nursery.start_soon(proc.wait)
    nursery.start_soon(proc.wait)
    nursery.start_soon(proc.wait)
    proc.kill()
```

njsmith:
LGTM. There was a conflict with #862, whcih I think I resolved. If someone notices the CI passed while I wasn't looking, feel free to merge :-)

PR #858, opened 2019-01-16
Deprecate trio.ssl and trio.subprocess submodules

### LINKED ISSUE #852: Should we get rid of trio.ssl and trio.subprocess, and tell people to just import ssl/subprocess instead?
In the Beginning [1] there was `trio.socket`. Using a nested namespace seems like the only reasonable way to handle this, because we want to expose 99% of the functionality of the `socket` module, the `socket` module API is *huge*, and the changes we needed to make are complex and all over the place.

Then we copied this style for `trio.ssl`, and then `trio.subprocess`.

But for these two modules, it's not as compelling. `trio.ssl` is just `SSLStream` and `SSLListener`, plus a ton of complicated re-exports. `trio.subprocess` is just `Process`, plus a ton of complicated re-exports.

Given this, maybe it would be better to put `SSLStream`, `SSLListener`, and `Process` into the top-level `trio` namespace, and drop the `trio.ssl` and `trio.subprocess` modules entirely. Instead we'd just tell people they can `import ssl` or `import subprocess`.

CC: @oremanj since this is partly triggered by #833

[1] Actually, I stole the idea from `curio.socket`.

oremanj:
The interesting pieces have moved to the main trio namespace; the reexports will be going away, with users accessing the standard `ssl` or `subprocess` module directly.

Closes #852.


oremanj:
Weird coverage flapping strikes again. The reported misses are in `test_ssl.py`, in nonsensical places like comments, blank lines, and a `nonlocal` statement in a function the rest of whose body is covered just fine. I get 100% coverage when I run on my local machine.

njsmith:
Looks good, thanks!

PR #863, opened 2019-01-20
Subprocess support: tighten and document expectations for how the command is specified

oremanj:
On POSIX, don't permit two error-prone combinations that the underlying `subprocess` module supports:
* string command with `shell=False`: user might think it word-splits somehow, but actually it treats the entire string as the name of an executable to invoke with no arguments
* list command with `shell=True`: binds the extra arguments to shell variables `$1`, `$2`, etc, instead of passing them to the command being executed

Add a section on quoting to the subprocess documentation, explaining the above restrictions and acquainting users with the many pitfalls of using `shell=True` on Windows.


njsmith:
Thanks!

Issue #796, opened 2018-12-03
Provide some standard mechanism for splitting a stream into lines, and other basic protocol tasks


njsmith:
Most networking libraries provide some standard way to implement basic protocol building blocks like "split a stream into lines", "read exactly N bytes", or "split a stream into length-prefixed frames", e.g.:

* asyncio [`StreamReader.readline`](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readline), [`StreamReader.readexactly`](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readexactly), [`StreamReader.readuntil`](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readuntil]

* The classes in [`twisted.protocols.basic`](https://twistedmatrix.com/documents/current/api/twisted.protocols.basic.html)

* The stdlib socket module's [`makefile` method](https://docs.python.org/3/library/socket.html#socket.socket.makefile), that lets you get access to the full Python file API, including `readline` and friends

* Tornado `IOStream`'s [`read_until`](https://www.tornadoweb.org/en/stable/iostream.html#tornado.iostream.BaseIOStream.read_until)

We don't have anything like this currently, as I was reminded by [this StackOverflow question](https://stackoverflow.com/questions/53575979/how-can-i-read-one-line-at-a-time-from-a-trio-receivestream) from @basak.

# Use cases

* Simple protocols used in tutorials, to make it easy for beginners to get something working
* Using as a base for implementing substantial/standard protocols that happen to use one of these framing methods. This mostly applies to the line-based framing, e.g. twisted's [`LineReceiver`](https://twistedmatrix.com/documents/current/api/twisted.protocols.basic.LineReceiver.html) and [`LineOnlyReceiver`](https://twistedmatrix.com/documents/current/api/twisted.protocols.basic.LineOnlyReceiver.html) have subclasses implementing HTTP, IMAP, POP3, SMTP, Ident, Finger, FTP, Memcache, IRC, ... you get the idea.
[...]

PR #866, opened 2019-01-21
[WIP] Memory Streams

asmodehn:
A small step, to ultimately address https://github.com/python-trio/trio/issues/170
Exposing memory streams similarly to tcp streams to be able to test network code without using any actual sockets.

**Still WIP.**
No docs and basic tests (`check_half_closeable_stream`) still blocking around https://github.com/python-trio/trio/blob/master/trio/testing/_check_streams.py#L275

### DIFF
--- a/trio/__init__.py
+++ b/trio/__init__.py
@@ -42,6 +42,8 @@ from ._channel import open_memory_channel
 
 from ._signals import catch_signals, open_signal_receiver
 
+from ._highlevel_memory import memory_connect, MemoryStream, MemoryListener
+
 from ._highlevel_socket import SocketStream, SocketListener
 
 from ._file_io import open_file, wrap_file
@@ -54,7 +56,7 @@ from ._ssl import SSLStream, SSLListener, NeedHandshakeError
 
 from ._highlevel_serve_listeners import serve_listeners
 
-from ._highlevel_open_tcp_stream import open_tcp_stream
+from ._highlevel_open_tcp_stream import open_tcp_stream, format_host_port
 
 from ._highlevel_open_tcp_listeners import open_tcp_listeners, serve_tcp
 
--- a/trio/_highlevel_memory.py
+++ b/trio/_highlevel_memory.py
@@ -1,0 +1,61 @@
+# "High-level" in-memory networking interface
+
+from . import _core, sleep, open_memory_channel, StapledStream
+from ._sync import Event, Queue
+from .abc import HalfCloseableStream, Listener
+
+__all__ = ["memory_connect", "MemoryStream", "MemoryListener"]
+
+from . import testing
+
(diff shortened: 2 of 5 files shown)


njsmith:
Hi! Thanks for working on this!

Let's start by figuring out what the plan is at a high level. Right now, the idea in this PR seems to be that people will write `open_tcp_stream(..., testing=True)`, or `serve_tcp(..., testing=True)`. But that seems pretty awkward, especially in cases where the person writing the tests doesn't actually control the networking code. (For example, maybe I'm trying to write a test using an HTTP library like requests – I would need to have some way to tell requests that *it* should pass `testing=True`.)

So, the idea we've been working on is a bit different. Trio has hooks ([documented here](https://trio.readthedocs.io/en/latest/reference-testing.html#virtual-networking-for-testing)) that let you override *all* networking inside a Trio program, in a single step. It works by overriding how Trio's sockets work, so it works for programs that use requests, it works for programs that use `open_tcp_stream`, and it works for programs that ignore all those fancy abstractions and use `trio.socket` directly. I think this is what we want to end up doing. And the nice thing about this is that we don't have to expose a bunch of new public APIs like `testing=True`... the only public part is `with virtual_network(): ...` or something, and then all the magic can happen behind the scenes.

Does that make sense?

BTW, depending on how complex this is, it might make sense to split it out as a separate library. We don't have to decide that up front, and [cookiecutter-trio](https://github.com/python-trio/cookiecutter-trio) makes it pretty quick to set up a new library, but it's something to keep in mind.

njsmith:
I definitely recommend using the sans-io approach, it's great. (And you'll want to read #796 for more discussion.) But usually when we talk about sans-io, we don't mean asyncio protocols/transports; we mean libraries like h11/h2/wsproto, which use a fairly different API style. And the Trio equivalent of protocols/transports is the [Stream API](https://trio.readthedocs.io/en/latest/reference-io.html#the-abstract-stream-api), which is much more friendly to composition than the traditional protocol/transport approach.

If you want to see examples of sans-io libraries hooked up to trio, then I'd suggest looking at:

* [this example](https://github.com/python-hyper/h11/blob/master/examples/trio-server.py) from the [h11 docs](https://h11.readthedocs.io), especially the `TrioHTTPWrapper` that wires together a `trio.abc.Stream` with an `h11.Connection`.
* [trio-websocket](https://github.com/HyperionGray/trio-websocket/blob/master/trio_websocket/_impl.py) (which is based on [wsproto](https://wsproto.readthedocs.io/en/latest/))

njsmith:
I'm going to mark this closed for now b/c it sounds like you're rethinking things, so there isn't anything to review currently. Feel free to re-open after you do, or open a new PR, or whatever makes most sense :-).

PR #869, opened 2019-01-22
Enable codecov flag support

njsmith:
See: https://docs.codecov.io/docs/flags

When we get bizarro coverage results (e.g. complaints about comments
not executing), then I'm hoping this will help us track them down.


njsmith:
Well, judging from the comment this seems ot be working... but I'm a bit nervous about the thing where if I click "Continue to review full report at Codecov", then codecov like... freezes. The browser just times out eventually.

I'm hoping this is like, some weird side-effect of codecov not knowing how to compare a PR with flags to a master branch that doesn't have flags? Maybe if we merge this then codecov will magically fix itself?

pquentin:
Let's find out!

PR #872, opened 2019-01-23
Subprocess support: run_process()

### LINKED ISSUE #822: trio.run_process

oremanj:
Since we're now deviating from stdlib subprocess in a few different ways, also uppdate the subprocess documentation and docstrings to move away from the "just like stdlib subprocess, only async" framing.

Closes #822.


docs/source/reference-io.rst
```diff
@@ -651,153 +651,110 @@ Spawning subprocesses
...
+
 Trio always uses unbuffered byte streams for communicating with a
-process, so these options don't make sense. Text I/O should use a
-layer on top of the raw byte streams, just as it does with sockets.
-[This layer does not yet exist, but is in the works.]
+process, so it does not support the ``encoding``, ``errors``,
+``universal_newlines`` (alias ``text`` in 3.7+), and ``bufsize``
+options. Text I/O should use a layer on top of the raw byte streams,
+just as it does with sockets.  [This layer does not yet exist, but is
+in the works.]
```
njsmith: I'm having some more doubts about including this kind of forward-looking statement in the docs... maybe we should make it just: "Currently, Trio always uses unbuffered byte streams for communicating with a process, so it does not support the encoding, errors, universal_newlines (alias text in 3.7+) and bufsize options." And leave it at that?

I don't know for sure yet whether we will have `TextStream` wrapper objects in trio by default, and if we ever do add them then we might as well implement the arguments here (as a shorthand for wrapping the relevant `Stream`s in `TextStream`s).

trio/_subprocess.py
```diff
@@ -218,17 +230,187 @@ def returncode(self):
...
+          can send signals to the subprocess or wait for it to exit.
+          They shouldn't try to send or receive on the subprocess's
+          input and output streams, because :func:`run_process` is
+          already doing that. Note that signals which terminate a
+          subprocess often result in a nonzero return code; you
+          probably want to pass ``check=False`` and do your own
+          more specific error check if you're planning on sending any.
+      **options: :func:`run_process` also accepts any :ref:`general subprocess
+          options <subprocess-options>` and passes them on to the
+          :class:`~trio.Process` constructor.
```
njsmith: I'm having trouble thinking of cases where getting access to the `Process` object is the natural way to accomplish something... if someone's using `run_process` to pause a task to run a process, then it feels like the natural way to cancel that is to, y'know, cancel it :-). And TOOWTDI?

There's a long tradition in other libraries having lots of kinds of "thing you run", and giving each their own ad hoc cancellation API, that I've tried to avoid succumbing to...

docs/source/reference-io.rst
```diff
@@ -651,153 +651,110 @@ Spawning subprocesses
...
+.. autofunction:: trio.run_process
+
+If you want to run a process with no standard stream redirections
+(i.e., taking input from the user and showing output and errors to the user),
+you can use ``trio.run_process(..., stdin=None, stdout=None, stderr=None)``.
+This runs the child in the same process group as the parent Trio process,
+so a Ctrl+C will be delivered to both the child process and the parent Trio
+process simultaneously. If that's not what you want, you'll need to delve
+into platform-specific APIs like (on UNIX) :func:`os.setpgrp` and
+:func:`os.tcsetpgrp`.
```
njsmith: 1. Why is this up here, instead of in the `run_process` docstring? Output capturing semantics seems like a good thing to describe in the docstring :-).

2. I'm surprised at the switch to capturing by default. It doesn't seem obviously *wrong* to me, but it doesn't seem obviously right, either, and it's a pretty major change from `subprocess.run` so it needs some kind of rationale.

3. The mention of `setpgrp` and `tcsetpgrp` seems to come out of no-where here? I'm not sure if we should mention them at all, and if we do then probably it shouldn't be tacked onto the end of an unrelated paragraph :-)

docs/source/reference-io.rst
```diff
@@ -651,153 +651,110 @@ Spawning subprocesses
...
+
+   .. automethod:: wait
+
+   .. automethod:: poll
+
+   .. automethod:: kill
+
+   .. automethod:: terminate
+
+   .. automethod:: send_signal
```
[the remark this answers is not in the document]
oremanj: I wanted to be able to control the order, and I couldn't get them to go in non-alphabetical order using `:members:`. (This despite our docs/source/conf.py having `autodoc_member_order = "bysource"`, and also trying an explicit `:member-order: bysource`.)

trio/_subprocess.py
oremanj: I could imagine wanting to send SIGSTOP/SIGCONT, or to send some graceful shutdown signal (since we dropped the explicit support for that), or to have access to the PID for various reasons. I guess it's true that many of these wouldn't overlap with output capturing, and with no capturing and `check=False` run_process is basically just `async with Process(...) as proc: pass`, so maybe it's more of an attractive nuisance than it is a useful tool? Happy to remove the `start()` support if you think it's not worth it after reading this explanation.

docs/source/reference-io.rst
oremanj: I implemented the `passthrough` option; let me know how you think it reads?

docs/source/reference-io.rst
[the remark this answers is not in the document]
njsmith: Ah, yeah, that's https://github.com/python-trio/sphinxcontrib-trio/issues/13

This was already fixed back in August, we just never released it... whoops. I just released that, so hopefully bysource works now.

oremanj:
@njsmith - I think I've addressed all the outstanding comments here; could you take another look?

docs/source/reference-io.rst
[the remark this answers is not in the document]
njsmith: I think you can drop these now; `"bysource"` should work now. (And if it doesn't I want to know ;-))

trio/_subprocess.py
njsmith: I think let's drop the `task_status` for now, and we can always add it back later when we have more experience.

docs/source/reference-io.rst
njsmith: The `passthrough` option is exactly the same as the stdlib's `capture_output`, but backwards, right? If we do do this, then making the default `capture_output=True` seems strictly better than making it `passthrough=False` :-).

I think I might still prefer `capture_stdout=True, capture_stderr=False`, but before we dive into that can you clarify the handling of `input`? In the stdlib `run`, AFAICT `input=None` (the default) means we let the child inherit the parent's stdin, and you have to pass `input=b""` if you want to give it an empty pipe. Are we matching that, or are you thinking that `passthrough` would affect this, or...?

docs/source/reference-io.rst
```diff
@@ -680,80 +711,61 @@ Trio always uses unbuffered byte streams for communicating with a
 process, so these options don't make sense. Text I/O should use a
 layer on top of the raw byte streams, just as it does with sockets.
 [This layer does not yet exist, but is in the works.]
+>>>>>>> origin/master
```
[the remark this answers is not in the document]
oremanj: yeah, not sure how I missed these

oremanj:
For some reason github isn't letting me reply to or resolve your last two line comments inline

> I think you can drop these now; `"bysource"` should work now. (And if it doesn't I want to know ;-))

I just upgraded my sphinx and sphinxcontrib-trio (the latter to 1.0.2, which supposedly includes the fix?) and it still seems to enforce alphabetical ordering.

> I think let's drop the `task_status` for now, and we can always add it back later when we have more experience.

Ok!

docs/source/reference-io.rst
oremanj: `passthrough=True` gives the stdlib defaults. `passthrough=False` is equivalent to `input=b""` if no `input` is specified, plus `capture_output=True`. I explained above in this thread why I think it's error-prone to have different isolation defaults for input vs output.

My preference would be to have one flag that controls "are the standard streams for this process basically isolated or are they basically inherited from the parent Trio process", and then people can modify the treatment of individual streams with the stdin/out/err options if they don't want all three streams treated the same way. If the standard streams are isolated and you didn't provide any input, you probably want the subprocess's input to come from /dev/null. It seems confusing to me to distinguish `input=None` from `input=b""` for that purpose.

njsmith: I've been thinking about this a lot this week!

So first, an anecdote: I actually ran into a real case where I would have used `run_process` if we had it. It was a very simple process monitor – I just wanted a script that would start up a flask dev server + a worker daemon that it talked to, and tear them both down if either exited. So:

```python
async def main():
        async def run_process_then_quit(cmd):
            await trio.run_process(cmd)
            nursery.cancel_scope.cancel()
        nursery.start_soon(run_process_then_quit, [...])
        nursery.start_soon(run_process_then_quit, [...])

trio.run(main)
```

Earlier you wrote:

> If you're writing a concurrent app, you might want to run multiple processes at once. That's maybe even _more_ likely than that you'd be careful not to run multiple processes at once. If you run multiple processes at once, it's not really safe to not isolate their I/O from each other.

This makes a lot of sense. But... OTOH, in my example, the correct behavior is actually to not capture stdout/stderr :-). Both daemons are dumping debug chatter on their output, and I definitely want to see all of it on my terminal as it happens. So I guess both capturing and non-capturing use cases do happen in real life, even when you're writing concurrent apps!

OTOH, in this case it probably would be better to do `input=b""`. It doesn't actually matter that much in practice, because these processes never read from their input. But if I was wrong about that, and they actually did, then `input=b""` would make sure I noticed and had to figure out some more reasonable solution. So that to me suggests that maybe we shouldn't be using a single flag to control both input and output handling simultaneously.

I've also seen *not* passing `input=b""` cause problems in real life, because of this kind of incorrect assumption. For example, this fun bug, which was caused because it never occurred to the pip developers that someone would write a `setup.py` that prompted the user for input: https://github.com/pypa/pip/issues/2732

One more random thought: I figured out why I'm a bit uncomfortable with either capturing both stdout + stderr or neither. It's because the the original motivation for making them different things was exactly to encourage people to capture stdout for programmatic consumption while leaving stderr pointing at a terminal :-). Specifically, to make unix pipelines work: `foo | bar` sends `foo`'s stdout to `bar`, but `foo`'s stderr stays on the terminal for a human to read.

OK. So how do we decide on defaults.

One good heuristic for defaults is: whatever configuration is most common, should be the default. I guess that's implicitly what we're thinking of when we debate whether concurrent programs normally want to capture or not.

There's another heuristic, to consider too, though: if there are two configurations for two sorts of users, then the default should be the one where the users two don't want it will realize immediately that they have the wrong configuration. I.e., any default will be wrong sometimes, so pick the default that when it's wrong, is obviously wrong.

In this case – the two kinds of users are the ones who want to capture output, and the ones who don't. The ones who want to capture output are going to immediately *do* something with that output – parse it or write it to a file or something. As they type `run_process`, they are thinking about how they're going to capture that output. If our default is to *not* capture output, then they will immediately notice and turn on output capturing. OTOH, the users who don't want to capture output may not be thinking about the output at all. They weren't planning to write any code to handle it. Maybe they expect that the program won't print any output at all. If our default *is* to capture output, and they don't stop and think about that, then they'll end up silently discarding the output. That seems like a nastier failure mode. Especially for the stderr case, where there really is no output for the first month your system is in production, and you don't notice that you're discarding the stderr until after things blow up and you have no idea why.

So I am leaning towards: we should follow `subprocess.run`, and *not* capture stdout/stderr by default.

If we go with this, then we'll need some spelling for how to override the default. For this I'm still leaning towards `capture_stdout=`, `capture_stderr=` as separate switches, because of the intuition from `foo | bar`.

Now, what about the input. Here, the argument works a bit differently... `input=b""` actually seems like the safer default. Having multiple processes reading from the same pipe at the same time is a nightmare; ttys have [some protection](http://curiousthing.org/sigttin-sigttou-deep-dive-linux) but it's still a mess. And even if you're only running one process, you can have fun failure modes like the bug above complaining that "pip just freezes on some packages". So I'm leaning towards `input=b""` as the default.

If we go with this, the nI'm not quite sure how to get the ergonomics right for when people want to override it. In particular:

* what do you write when you want the child to inherit stdin? `input=None`? Not super intuitive, though I guess it's straightforward enough to document and it would get the job done.  `stdin=0`? That'd be pretty arcane.

* How do we resolve conflicts between `input` and `stdin`? I think we want to allow `run_process(cmd, stdin=<arcane subprocess stuff>)`. But if `input=b""` is the default, then there's no way to tell for sure that someone hasn't pass both `input=` and `stdin=`, which is awkward. I guess the best option is just to say, if `input == b""`, and `stdin` was given, then we ignore `input`? Not the most elegant thing, but maybe the alternatives are worse.

There are also other options, like making the default signature `input=None` and declaring that that *means* the same thing as `input=b""` (except possibly if `stdin=` is passed). But that seems like even more of a mess than the alternatives.

What do you think?

oremanj: Good points all!

The danger of running multiple processes with the same stdout/stderr is that you get their output unhelpfully intermingled. You know, `process oprocess two outputne output` type of thing. :-) But most things use line buffering for terminal output, and even when they don't, a human can usually reconstruct what happened. So, I can see how letting multiple processes write to the same stdout/stderr could be more useful-in-practice that it is unsafe-in-theory.

Meanwhile, running multiple processes with the same _stdin_ is almost always a terrible idea; if there's contention on reading from it, you get a pretty bad user experience, like only half your keystrokes seeming to work.

I agree with your point about it being easy to notice you failed to capture the output you wanted to capture (because those bytes you're about to inspect will be `None` instead) and hard to notice you accidentally captured the output you didn't want to capture. It's actually really obvious with that lens that my previous default was misguided, so thanks for laying it out like that!

How about:
* The defaults are `stdin=subprocess.DEVNULL, stdout=None, stderr=None`. These can be overridden by the user passing `stdin`, `stdout`, `stderr` arguments.
* `capture_stdout=True` means `stdout=subprocess.PIPE` and errors if a different `stdout` was specified by the user. Likewise for `capture_stderr=True` and `stderr=subprocess.PIPE`.
* For `subprocess.run()` compatibility, `capture_output=True` is equivalent to `capture_stdout=True, capture_stderr=True`.
* `input=<some bytes>` means `stdin=subprocess.PIPE` and tells the parent Trio process to write those bytes to the pipe, then close it. `passthrough_stdin=True` means `stdin=None`. It's an error to specify more than one of `input`, `passthrough_stdin`, and `stdin`.
* `input=None` is treated equivalently to "you didn't specify any input" and still means `stdin=subprocess.DEVNULL`. This is a discrepancy from `subprocess.run()`, but at least it fails fast. If we don't like that, we could have `input=None` be an error, and make the default be some private marker object instead of `None` in order to be able to distinguish.

oremanj:
@njsmith: heads up that this PR is currently awaiting a response to the proposed redirection semantics in https://github.com/python-trio/trio/pull/872#discussion_r255368338. No rush if you're busy, just wanted to poke since I find it easy to accidentally lose comments in github threads and I figured there was a chance that that had happened here.

docs/source/reference-io.rst
njsmith: > The danger of running multiple processes with the same stdout/stderr is that you get their output unhelpfully intermingled. You know, `process oprocess two outputne output` type of thing. :-) But most things use line buffering for terminal output, and even when they don't, a human can usually reconstruct what happened. So, I can see how letting multiple processes write to the same stdout/stderr could be more useful-in-practice that it is unsafe-in-theory.

I admit that in the real situation, the next thing I tried after getting the basic thing working was to mess around with proxying the process's output so I could make them different colors :-).

> [calling conventions]

Hmm.

I don't think we should allow or expose `PIPE` as a thing we let users see or pass. The pipes are an internal implementation detail, that the user can't get at, so we should abstract them away from the user's interface.

I kind of like `input=b""` better than `stdin=subprocess.DEVNULL`. Obviously they're equivalent semantically, but when we show the signature in the docs, `input=b""` seems easier to understand. Anyone who understands `input` in general will be able to figure out what `input=b""` means. But if we do `stdin=subprocess.DEVNULL`, then people still have to learn `input`, and then also know about the subprocess `stdin` arg (which is very complicated!), *and then* know what this weird `DEVNULL` thing is.

This thing where there are 3 different modes stdin could be in, and then two of the modes have more configuration (and one of them is really another like 7 modes, and includes passthrough as a special case)...... wedging this into Python's argument passing model is extremely awkward.

I almost wonder if we should drop `input`, and make it so `stdin` can be a bytestring, or else any of the things the `subprocess` would accept (except `PIPE`):

```python
def run_process(cmd, *, check=True, stdin=b"", capture_stdout=False, capture_stderr=False, **kwargs):
    ...
```

If we're splitting up `capture_output` and using the words `stdout` and `stderr`, then replacing `input` with `stdin` is nicely consistent.

This is where I paused to think more, and then left this comment window sitting open for several days :-). Thanks for the ping. I'm just going to hit post and see what you think :-).

oremanj:
Sorry for the delay in responding - how does this iteration look?

docs/source/reference-io.rst
Badg: Quick preface -- I'm hoping this comment is helpful, from someone who's working to get trio into an internal app we use at work (but nonetheless in prod there, which is pretty neat). But I know there's work been done here and there's a whole trope about people coming into open source projects with demands and just yuck I'm really not trying to do that!

That being said, one thing I've found *extremely* useful is to, instead of passing a `capture_stdout` etc, is to pass a `stdout_redirect`, something like this:

```python
async def subproc_run(command, *, stdin_redirect=None, stdout_redirect=None,
                      stderr_redirect=None, **options):
    '''Any of stdin_redirect, stdout_redirect, and stderr_redirect
    should be trio channels. We'll read (stdin) or write (stdout/stderr)
    to those channels as needed.
    '''
```

Originally I started with the ``subprocess.run()`` replica [in the docs](https://trio.readthedocs.io/en/latest/reference-io.html#running-a-process-and-waiting-for-it-to-finish), but one of the ways I'm using this is to call into terraform, and those commands can sometimes take on the order of a minute to run. So I'd be sitting there looking at my console for, like, a long time, wondering what was up, and then finally at the end it would dump the whole stdout into the terminal, by looking into the returned ``subprocess.CompletedProcess.stdout`` (or stderr), which was suboptimal.

So instead I retooled it a bit to accept trio channels instead. I kinda punted on the "what's an appropriate buffer size here" question, and this is still in-process code (also, there's some stuff I'm doing with an ``async with capture_output()`` helper object that's probably too cutesy-clever for something like trio)  -- but anyways, I put the implementation and a few example usages into [this gist](https://gist.github.com/Badg/d188f2c0c0bb684a1a08920bdac0d3eb).

oremanj: Thanks for the suggestion! I actually thought about that as well when I was doing the most recent revision, and decided that without a specific request it was probably going to be rejected as unnecessary complexity. I guess we have a specific request now. :-)

@njsmith, what do you think about this idea of making `run_process` accept channels as `stdin`, `stdout`, and/or `stderr` arguments? If we do that we might want to add back the nursery.start() support, since it becomes a lot more reasonable for someone to want to do stuff with the Process object while `run_process` is executing.

Another option: we could keep `run_process` as the black-box start-to-finish API like it is currently, and add back something like the `async with open_process(...) as proc:` from a previous diff of mine, with channels support and so on.

This doesn't all need to happen on this diff, and I suspect it could be prototyped fine in an external library.

njsmith: @badg @oremanj What does "channels support" mean? What do these channels do? (Do you mean channels or streams?)

If you want to run a subprocess while incrementally running some arbitrary function on the output, then what is it that makes `async with trio.Process(..., stdout=PIPE) as p: ...` unpleasant and this better?

oremanj: I didn't actually have a chance to look at @Badg's code yet, but my impression of the proposal was to support saying something like
```
write_stdout, read_stdout = trio.open_memory_channel[bytes](0)
# start some task that will consume data from read_stdout
await trio.run_process(..., stdout=write_stdout)
```
I guess the main utility is that you can use incremental processing for some streams and bulk processing for others? And the channel interface is nicer than the stream one for incremental processing -- `async for chunk in channel:` rather than
```
while True:
    chunk = await stream.receive_some(ARBITRARILY_CHOSEN_POWER_OF_TWO)
    if not chunk:
        break
    ...
```

smurfix: That `while` loop is ugly no matter where you see it, but nothing stops us from adding a `receive_loop` method to our streams that lets us do an `async for chunk in stream.receive_loop(POWER_OF_TWO)`.

The other choice would be to wait for Python 3.8 and use that infamous `:=` thingy. 🤪

Badg: > my impression of the proposal was to support saying something like

@njsmith @oremanj yep, that's the core of it -- I found it really useful to use channels (in my case, memory channels) as the high-level subprocess communication API. In my case the wrapper method doesn't even expose the ``capture_output``, ``input``, etc keywords; if you pass in channels, it captures the output there; if you don't, it discards it.

My use case is more eager processing than incremental processing, but that's sometimes a bit of a fine distinction, and I suspect that incremental processing is more likely to be the important one.

Badg: Thinking about this some more... I think actually that the mental distinction I'm having here is between pre-allocating consumers and post-allocating consumers, given the ``run_process`` high-level API.

I think at an intuitive level, I expect ``async def run_process`` to be atomic -- it's responsible for starting, managing, and closing the process; when it returns, the process is definitely finished, and before you call it, the process definitely hasn't started (plus all of the other usual trio semantics).

So then if you want to combine that high-level API with eager/incremental processing of output/input, then you're (by definition) forced to preallocate your stdin producers and stdout/err consumers using a nursery, which you then need to coordinate with the ``run_process`` API somehow so that they get hooked up correctly. That, ultimately, is the API problem my wrapper is trying to solve.

Also partly relevant to the discussion in #959, I think my choice for a channel over a stream, even though the stream is quite literally the thing returned by the trio ``Process``, is the result of two things (listed in order of importance):

1. at least according to the docs, there's nothing suggesting I can pass an existing stream object to the ``trio.Process`` constructor, which means I'd have no way of preallocating the stream
2. it seems much, much more natural (and easier!) to use channels for this, **because of the ``send_channel, receive_channel = trio.open_memory_channel()`` method**, which immediately allocates everything you need to do this coordination (and close it down easily)

Hopefully that explanation of my train of thought makes sense!

njsmith: @Badg 
> at least according to the docs, there's nothing suggesting I can pass an existing stream object to the trio.Process constructor, which means I'd have no way of preallocating the stream

That's correct. It's possible in some exotic cases, but not in general.

> Thinking about this some more... I think actually that the mental distinction I'm having here is between pre-allocating consumers and post-allocating consumers, given the run_process high-level API.

Why do you prefer pre-allocating consumers, though? It seems like this is just as easy?

```python
async with trio.Process(...) as p:
        ...
```

Or if not, why not? (I guess two possibilities that come to mind: that doesn't do `check` and it doesn't kill the process on error. I sort of wonder if `Process.__aexit__` should kill the process if there's an exception, by analogy with nurseries cancelling siblings on error.)

Badg: @njsmith I don't necessarily prefer preallocating consumers. The goal with my wrapper was to provide a high-level API that does ``check``, handles ``await process.wait()``, kills the process on error/cancellation/etc, all within a single ``await run_process`` call. As far as I'm aware, the only possible way to combine those semantics with incremental processing (of either input or output) is to preallocate, so that's what the API ends up exposing. Does that make sense?

njsmith: That's true, if the requirement is that it be a single `await my_run_process(...)`, then you have to preallocate. But what I'm still not sure of is: why is it important that it be a single `await my_run_process(...)`, as compared to a single `async with my_run_process(...): ...`?

trio/_subprocess.py
```diff
@@ -389,24 +391,47 @@ def kill(self):
...
         )
-
-    if options["stdin"] == subprocess.PIPE and not input:
-        options["stdin"] = subprocess.DEVNULL
+    if isinstance(stdin, bytes):
+        input = stdin
+        if input:
+            options["stdin"] = subprocess.PIPE
+        else:
+            options["stdin"] = subprocess.DEVNULL
```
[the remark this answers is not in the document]
smurfix: I can, but if you want DEVNULL instead of an empty pipe, then use DEVNULL. Explicit is better than premature micro-optimization, or something like that.

smurfix:
-1 on `capture`. +0 on `capture_both`; there's exactly one reason I can think of why you would do that, and that's to grab the output of "/usr/bin/whatever --help", because you never know whether it's going to be printer to stdout or stderr – but you can handle that case just as easily by capturing stdout and stderr separately.

njsmith:
> grab the output of "/usr/bin/whatever --help", because you never know whether it's going to be printer to stdout or stderr – but you can handle that case just as easily by capturing stdout and stderr separately.

Arguably the best way to do that is `capture_stdout=True, stderr=STDOUT`. (Actually this is what you want in a lot of cases – it loses the distinction between stdout and stderr, *but* in exchange you get them interleaved properly in time.)

oremanj:
All right, I'll concede that my bias towards output capturing seems to be in the minority around here. :-) Have removed `capture` entirely and addressed the other comments.

docs/source/reference-io.rst
Badg: I mean, at the end of the day it's just my own personal API preference, so it deserves a heaping mound of salt. But I think my preference for an atomic ``my_run_process(...)`` over an async context manager, is that the context manager introduces a second concurrency mechanism to trio -- it's behaving both like a subprocess spawner, and a nursery, and my personal preference would be to keep those two separate. It was also my experience that the context manager resulted in more boilerplate, but I think that's the weaker point so I'm not sure it's worth exploring very far.

Like I said though, I'm just one person, and I always reserve the right to be wrong haha. Just thought I'd add some thoughts into the mix!

dhirschfeld:
> *in exchange you get them interleaved properly in time*

For me this is critical and ~always what I do because it's the easiest way to debug when things go wrong. I've always found the stderr on it's own with no surrounding context to be pretty useless for debugging purposes.

Is there a way to do this with the current design? From a quick look I can't see how?

oremanj:
> Is there a way to do this with the current design? From a quick look I can't see how?

`run_process()` accepts all the Popen-style redirection options for stdin/stdout/stderr as alternatives to the higher-level `stdin=b"stuff"` / `capture_stdout` / `capture_stderr`. So you would say `await run_process(..., capture_stdout=True, stderr=subprocess.STDOUT)`, and look at the `.stdout` attribute of the returned object.

smurfix:
> in exchange you get them interleaved properly in time

Well, I would – if both stdout and stderr were either line buffered or unbuffered. Unfortunately, that's rarely the case; typically stderr is either line buffered or unbuffered, no matter where it's going, while stdout's buffer is some-power-of-two when it's not going to a terminal.

njsmith:
> Well, I would – if both stdout and stderr were either line buffered or unbuffered. Unfortunately, that's rarely the case; typically stderr is either line buffered or unbuffered, no matter where it's going, while stdout's buffer is some-power-of-two when it's not going to a terminal.

Off-topic but sometimes useful hack: If your subprocess is python, then you can pass `-u` or set the envvar `PYTHONUNBUFFERED=1` to force stdout and stderr into line-buffered mode.

The more general solution would be to allocate a tty instead of a pipe. That has some extra complications and I'm not sure if convenience functions for it are in-scope for trio or not, but if you want to open an issue to discuss then feel free.

Issue #830, opened 2018-12-28
Remove mixins in _windows_pipes.py and _unix_pipes.py?


njsmith:
I suspect this code will be nicer if we switch from inheritance → composition. This is a reminder to try it and find out :-)

njsmith:
Fixed in #874 and #884.

Issue #661, opened 2018-09-13
Subtle bugs around closure handling in _unix_pipes.py and _windows_pipes.py


njsmith:
This is something I just realized while thinking about a random question someone asked in #twisted:

In `unix_pipes.py`, suppose that the following sequence of events happens:

1. task A calls `receive_some`
2. the pipe isn't ready, so task A blocks in `wait_readable`
3. some data arrives on the pipe, so task A gets rescheduled
4. while task A is sitting on the run queue, task B closes the pipe
5. task A gets scheduled, and calls `os.read` on the closed fd, which if you're lucky raises an exception (EBADF), or if you're unlucky then a *new* fd got opened and assigned this value in between steps (4) and (5), and we end up reading from this random fd, which probably corrupts the state of some random other connection.

`send_all` has analogous issues.

I guess we need to re-check `self._closed` every time we use the fd, not just once on entry to the function.

This same issue can't happen with `SocketStream`, because when a socket is closed then it sets the underlying fd to `-1`, and so step 5 would call `recv(-1, ...)`, which is always an EBADF, and the `SocketStream` code knows to convert EBADF into `ClosedResourceError`.

With `trio.socket`, the `-1` thing means you can't get a wild read from a random fd, but there's no explicit handling of this case, so you will get an `OSError(EBADF)` instead of a proper `ClosedResourceError`. I guess that would probably be good to fix, though it's less urgent than the unix_pipes.py thing.

njsmith:
I think this could also affect `_windows_pipes.py`, though the way Windows handles work it's probably harder to cause actual corruption.

njsmith:
There's also a serious bug in `_unix_pipes.py` where `aclose` calls `notify_fd_closed` without checking `_closed`. This bit @Ninpo.

I think the root cause is back here, where I should have argued for clearing `self.fd` when we close it – that would have made it impossible to introduce bugs of this class:
https://github.com/python-trio/trio/pull/621#discussion_r212503766

We should use that strategy going forward.

njsmith:
It looks like `_socket.py` is OK, because:

* The stdlib socket module already uses the trick where as soon as you close the socket, it sets the internal fd/handle value to `-1`, guaranteeing that any future operations will raise `OSError` (either `EBADF` or `ENOTSOCK`).
* The lower-level trio socket module is allowed/expected to raise these `OSError`s when you perform operations on an already-closed socket. It can also raise `ClosedResourceError`, depending on its whims.
[...]

PR #874, opened 2019-01-23
Rewrite unix pipe fd handling logic

njsmith:
This addresses a number of issues:

- Fixes a major issue where `aclose()` called `notify_fd_close()`
  unconditionally, even if the fd was closed; if the fd had already
  been recycled this could (and did) affect unrelated file
  descriptors:
    https://github.com/python-trio/trio/issues/661#issuecomment-456582356
- Fixes a theoretical issue (not yet observed in the wild) where a
  poorly timed close could fail to be noticed by other tasks (gh-661)
- Adds `ConflictDetector`s to catch attempts to use the same stream from
  multiple tasks simultaneously
- Switches from inheritance to composition (gh-830)

Still todo:

- [x] Tests for these race conditions that snuck through
- [x] Audit `_windows_pipes.py` and `_socket.py` for related issues

### DIFF
--- a/trio/_unix_pipes.py
+++ b/trio/_unix_pipes.py
@@ -1,117 +1,161 @@
 import fcntl
 import os
 from typing import Tuple
+import errno
 
 from . import _core
 from ._abc import SendStream, ReceiveStream
+from ._util import ConflictDetector
 
 __all__ = ["PipeSendStream", "PipeReceiveStream", "make_pipe"]
 
 
-class _PipeMixin:
-    def __init__(self, pipefd: int) -> None:
-        if not isinstance(pipefd, int):
-            raise TypeError(
-                "{0.__class__.__name__} needs a pipe fd".format(self)
-            )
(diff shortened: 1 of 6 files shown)


trio/_unix_pipes.py
```diff
@@ -1,120 +1,158 @@
...
+            if max_bytes < 1:
+                raise ValueError("max_bytes must be integer >= 1")
+
+            while True:
+                try:
+                    data = os.read(self._fd_holder.fd, max_bytes)
+                except BlockingIOError:
+                    await _core.wait_readable(self._fd_holder.fd)
+                except OSError as e:
+                    await _core.cancel_shielded_checkpoint()
```
oremanj: Is this still meant to be here? You already did a full checkpoint when acquiring the conflict detector

njsmith: Good catch, thanks :-)

PR #881, opened 2019-01-26
switch to using codecov "names" instead of "flags"

njsmith:


njsmith:
Hey, this works pretty well actually!

Example: go here: https://codecov.io/gh/python-trio/trio/src/2d354c4f9a1dd810dcf9497ad5519738d43c19d0/trio/_file_io.py

And now just above the source code, there are three icons: [travis] [appveyor] [generic box]

If you hover over the travis icon, a menu drops down listing all the travis jobs. You can enable/disable them individually, or all at once. And, b/c of this PR, they actually have meaningful names, instead of being like "job 8". So this pretty much does what we wanted flags to do, but better!

Weird that codecov doesn't document this at all, but ¯\\_(ツ)_/¯

njsmith:
and... this last build seems to have worked great, except that for some reason all the travis and appveyor (but not azure!) coverage files were [apparently uploaded twice](https://codecov.io/gh/python-trio/trio/commit/7982456a5d693e5521747ab7f6c3af7b90135b14/build). codecov is baffling.

njsmith:
Looks like that *last* last build sorted things out again, so it was just a temporary glitch of some kind: https://codecov.io/gh/python-trio/trio/commit/05c2e91744c3082f7719f6bc2722b0b8547e7f2d/build

OK I think this is ready to go in! And hopefully it will also stop codecov from freezing when trying to look at PR reports...

pquentin:
Awesome, thanks!

Issue #883, opened 2019-01-27
On Windows, should we break up calls to `WriteFile` to reduce usage of non-paged memory?


njsmith:
In the initial version of the subprocess code, the Windows pipe `send_all` implementation breaks up large sends into mulitple small sends:

https://github.com/python-trio/trio/blob/360f1fb45e4a3edc2fb4251dd969d0496a68ae60/trio/_windows_pipes.py#L53-L67

This loop is obviously a bit tricky, and it also turns out that this trickiness [is hiding a subtle race condition](https://github.com/python-trio/trio/issues/661#issuecomment-457814382). We could fix that, but this also raises the question of whether we could just skip the trickiness altogether.

@oremanj [says](https://gitter.im/python-trio/general?at=5c4ce9ae54f21a71a1af8967) that his motivation for doing it this way is from reading in #52 about how servers sometimes want to avoid issuing lots of simultaneous `WSARecv` calls because the buffers all get pinned into memory at once. So this wasn't based on some special documentation somewhere about named pipes.

In libuv's named pipe code, the only thing it does with large buffers is check that they're not larger than UINT32_MAX, because apparently that's some kind of limit for `WriteFile`. (I am morbidly curious about how they discovered this. [1])

Poking around in [Windows Internals](https://www.microsoftpressstore.com/articles/article.aspx?p=2201309&seqNum=3) (search for "buffer management"), it does sound like in many cases, if you pass an N-byte buffer to `WriteFile`, then you end up with N-bytes of kernel memory pinned for the duration of the call. It also sounds like sufficiently clever drivers can work around this if they want to. I have no idea whether the named pipe drivers or the socket drivers do anything clever here; AFAICT the only way to know is to check the Windows source, which I don't have :-).

So if this is a problem, it sounds like it's a very general one: it applies to sending and receiving on sockets or pipes, to reading and writing files, to sync and async code, etc. etc. If we're going to worry about it, we should either worry about it everywhere, or figure out specific cases where it's likely to be hit and worry about it there.

[...]

PR #884, opened 2019-01-27
In Windows PipeSendStream.send_all, make just one call to WriteFile

njsmith:
See gh-883 for discussion.


njsmith:
After this is merged we can close: #661, #883, #830

(Also good news: codecov seems to be working again, presumably thanks to #881)

oremanj:
Looks good - thanks for fixing!

PR #890, opened 2019-01-28
Support deterministic scheduling

Zac-HD:
See python-trio/pytest-trio#73 - this is important for Hypothesis, and arguably for debugging in general.  

I've used the current time as primary sort key, with name and task cancel/schedule points to order tasks that started at the same time according to perf_counter.  This is all deterministic, and any remaining collisions are probably coming from random bitflips.

I've also designed the tests to be imported into pytest-trio's test suite, and the Hypothesis test added.

### DIFF
--- a/newsfragments/890.feature.rst
+++ b/newsfragments/890.feature.rst
@@ -1,0 +1,3 @@
+Added an internal mechanism for pytest-trio's
+`Hypothesis <https://hypothesis.readthedocs.io>`__ integration
+to make the task scheduler reproducible and avoid flaky tests.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1,4 +1,5 @@
 import functools
+import itertools
 import logging
 import os
 import random
@@ -58,6 +59,12 @@ elif hasattr(select, "kqueue"):
 else:  # pragma: no cover
     raise NotImplementedError("unsupported platform")
 
+# When running under Hypothesis, we want examples to be reproducible and
+# shrinkable.  pytest-trio's Hypothesis integration monkeypatches this
+# variable to True, and registers the Random instance _r for Hypothesis
+# to manage for each test case, which together should make Trio's task
+# scheduling loop deterministic.  We have a test for that, of course.
+_ALLOW_DETERMINISTIC_SCHEDULING = False
 _r = random.Random()
 
 # Used to log exceptions in instruments
@@ -682,6 +689,7 @@ class Task:
     name = attr.ib()
     # PEP 567 contextvars context
     context = attr.ib()
+    _counter = attr.ib(init=False, factory=itertools.count().__next__)
 
     # Invariant:
     # - for unscheduled tasks, _next_send is None
@@ -1557,6 +1565,12 @@ def run_impl(runner, async_fn, args):
         # change too, like the deadlines tie-breaker and the non-deterministic
         # ordering of task._notify_queues.)
         batch = list(runner.runq)
+        if _ALLOW_DETERMINISTIC_SCHEDULING:
+            # We're running under Hypothesis, and pytest-trio has patched this
+            # in to make the scheduler deterministic and avoid flaky tests.
+            # It's not worth the (small) performance cost in normal operation,
+            # since we'll shuffle the list and _r is only seeded for tests.
+            batch.sort(key=lambda t: t._counter)
         runner.runq.clear()
         _r.shuffle(batch)
         while batch:
--- a/trio/tests/test_scheduler_determinism.py
+++ b/trio/tests/test_scheduler_determinism.py
@@ -1,0 +1,42 @@
+import trio
+
+
+async def scheduler_trace():
+    """Returns a scheduler-dependent value we can use to check determinism."""
+    trace = []
+
+    async def tracer(name):
+        for i in range(10):
+            trace.append((name, i))
+            await trio.sleep(0)
+
+    async with trio.open_nursery() as nursery:
+        for i in range(5):
+            nursery.start_soon(tracer, i)
+
+    return tuple(trace)
+
+
+def test_the_trio_scheduler_is_not_deterministic():
+    # At least, not yet.  See https://github.com/python-trio/trio/issues/32
+    traces = []
+    for _ in range(10):
+        traces.append(trio.run(scheduler_trace))
+    assert len(set(traces)) == len(traces)
+
+
+def test_the_trio_scheduler_is_deterministic_if_seeded(monkeypatch):
+    monkeypatch.setattr(
+        trio._core._run, "_ALLOW_DETERMINISTIC_SCHEDULING", True
+    )
+    traces = []
+    for _ in range(10):
+        state = trio._core._run._r.getstate()
+        try:
+            trio._core._run._r.seed(0)
+            traces.append(trio.run(scheduler_trace))
+        finally:
+            trio._core._run._r.setstate(state)
+
+    assert len(traces) == 10
+    assert len(set(traces)) == 1


trio/_core/_run.py
```diff
@@ -1556,7 +1563,12 @@ def run_impl(runner, async_fn, args):
         # scheduling, then there are other things that will probably need to
         # change too, like the deadlines tie-breaker and the non-deterministic
         # ordering of task._notify_queues.)
-        batch = list(runner.runq)
+        #
+        # We sort tasks before shuffling them so that seeding the Random
+        # instance can make the scheduler deterministic, which is important
+        # for testing and debugging, especially with tools such as Hypothesis,
+        # without giving up the advantages of sets everywhere else.
+        batch = sorted(runner.runq, key=Task.sort_key)
```
njsmith: This is the very inner loop of trio's scheduler, so I'm slightly concerned about adding overhead. Of course the whole `_r.shuffle` thing is also dubious overhead... but maybe some simple measurements wouldn't be amiss? If it's non-trivial then we could hide it behind some flag that hypothesis (or someone) sets...

trio/_core/_run.py
```diff
@@ -682,6 +682,7 @@ class Task:
     name = attr.ib()
     # PEP 567 contextvars context
     context = attr.ib()
+    _creation_time = attr.ib(init=False, factory=perf_counter)
```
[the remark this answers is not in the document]
Zac-HD: I was thinking about more difficult concurrency models, mostly!  Trio is too easy :smile:

trio/_core/_run.py
Zac-HD: I just don't have enough of a sense for typical workloads to write a good benchmark - all my performance instincts are formed for out-of-memory array workloads or other heavy throughput-dominated stuff.  Very happy to run one and optimize accordingly though!

I don't think a heapq would help us here - to get a deterministic ordering we'd need to either run a n-log-n heapsort (and the `sorted` builtin is almost certainly faster), or else have the tasks stored in lists from creation time in which case we don't need to do anything (but we lose whatever benefits we currently get from sets).

njsmith: We don't really have good workloads either, which makes the whole question of performance a bit vague. But for this purpose I just meant something like a silly microbenchmark with 100 tasks that just reschedule themselves over and over, to get some kind of upper-bound on this. Maybe it doesn't matter at all.

Zac-HD:
OK, here's a terrible little microbenchmark, which at least demonstrates that it doesn't make a noticeable difference:

Using the exact `scheduler_trace` function that's in the tests for this PR (b675ef3e468600ef66dde43e230ca6f81f75a753), I used the ipython `%%timeit` magic:

	# In [4]: %%timeit
	#   ...: trio.run(scheduler_trace)
	# <with sorting and shuffling enabled>
	7.77 ms ± 1.54 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
	6.79 ms ± 154 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
	6.77 ms ± 119 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
	6.69 ms ± 47.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
	# <without sorting or shuffling>
	7.08 ms ± 1 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
	6.75 ms ± 337 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
	6.46 ms ± 26.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
	7.08 ms ± 1.47 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

So it's in the noise at that scale.  Now let's try with 1k tasks instead of five:

	# Enabled, 1k instead of five tasks
	808 ms ± 139 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	818 ms ± 134 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	763 ms ± 81 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	777 ms ± 82.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	# Disabled, same scale
	722 ms ± 98.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	666 ms ± 7.34 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	668 ms ± 14.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	670 ms ± 16.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

So we can at least construct workloads where the impact is noticeable (constant work per task and n-log-n scheduling make that easy), though still not pathological.  Judgement time: 

- Is this even a useful benchmark?  Is the "large" scale too large (or too small?), does an IO-free and dependency-free benchmark make sense, etc.  I think this is within a small constant factor of worst-case for a given number of tasks.
- Is a slight slowdown worth the maintenance cost of an internal "could be deterministic" flag and branch?  Probably yes, but I'd appreciate any thoughts.

njsmith:
So it looks like sorting+shuffling adds ~15% slowdown in a pure-scheduling workload. (Actually I bet we can get that even higher if we switch from `await checkpoint()` to `await cancel_shielded_checkpoint()`, since the latter is more optimized.)

Did you try sorting+shuffling versus shuffling alone, like we currently do?

Zac-HD:
Shuffling has a slight impact, but more importantly linear rather than log+linear complexity in the number of tasks.  Of course all of this depends on the number of tasks and the actual workload....

	# 1k tasks, shuffling but not sorting
	686 ms ± 6.67 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	698 ms ± 20.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	687 ms ± 12.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
	687 ms ± 12.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Since I expect `pytest-trio` to be the only consumer of determinism here, I've just gone and added a flag to enable the sorting step because it's basically free performance over the other option.

njsmith:
Yeah, since this is a super-janky API with exactly one consumer, I'm OK with hacky things like this.

I do wonder if it would be better to make the contract *slightly* less janky somehow. Would it be possible to have like `trio._core._run.hypothesis_mode()` that encapsulates both seeding the RNG and turning on sorting? What does the hypothesis RNG seeding functionality actually do, is it just calling `.seed(0)` before each test run or something fancier?

Zac-HD:
> What does the hypothesis RNG seeding functionality actually do, is it just calling `.seed(0)` before each test run or something fancier?

It *used* to do that, but it's now a bit fancier!  We eventually discovered that repeatedly seeding to zero meant that tests were affecting each other's state, and once we fixed it that there were some rare bugs that just never got triggered when the PRNG was "close" to the zeroed state.  So now we restore the previous state of every PRNG we manage after each test case is executed.

The *other* thing we do is the `random_module()` strategy, which you can use to tell Hypothesis "this test should be affected by randomness".  In that case, we actually draw a random seed value from an internal strategy, so test case executions are varied but still reproducible.  I wrote HypothesisWorks/hypothesis#1741 so we could use this for Trio too without Hypothesis needing to know the details of every such library - we make an exception for Numpy but try to keep everything else generalised.

So I don't think it's worth giving this a nicer API here (at least for now); it's kinda janky but I can add an autouse fixture to pytest-trio that does the monkeypatch for all and only tests with the Trio marker.  Like everything in the Hypothesis, Trio, and Pytest ecosystems it will be beautiful on top and... *pragmatic* underneath - magic hath it's price!

newsfragments/890.feature.rst
```diff
@@ -0,0 +1,4 @@
+The Trio scheduler can now be made deterministic by seeding the internal
+random.Random instance that is used to shuffle available tasks.
+This is not public API, but is important for some tools such as
+pytest-trio's `Hypothesis <https://hypothesis.readthedocs.io>`_ integration.
```
njsmith: The description here of how this works is a bit out of date, and it's also probably more detail than people need for a single-consumer API involving monkeypatching :-). How about just saying in general terms that the scheduler can now be made fully deterministic when running under hypothesis?

Also, in ReST links you almost always want two underscores: https://stackoverflow.com/questions/27420317/restructured-text-rst-http-links-underscore-vs-use

trio/_core/_run.py
```diff
@@ -63,6 +64,10 @@
 # Used to log exceptions in instruments
 INSTRUMENT_LOGGER = logging.getLogger("trio.abc.Instrument")
 
+# Private flag to sort tasks before shuffling them, so that seeding
+# the Random instance `_r` can make scheduling deterministic.
+_ALLOW_DETERMINISTIC_SCHEDULING = False
```
njsmith: Let's record a few more details here to help us figure it out again later. How about moving the definition of `_r` next to this, and the comment should say something like: "When running under hypothesis, we want scheduling to be as deterministic as possible. pytest-trio's hypothesis integration monkeypatches this variable to True, and seeds the `_r` instance, which together should be be enough to eliminate scheduling as a source of non-determinism."

Zac-HD:
Ping @njsmith; I've fixed the changelog and comment - ready to merge?

trio/_core/_run.py
```diff
@@ -1557,6 +1565,12 @@ def run_impl(runner, async_fn, args):
         # change too, like the deadlines tie-breaker and the non-deterministic
         # ordering of task._notify_queues.)
         batch = list(runner.runq)
+        if _ALLOW_DETERMINISTIC_SCHEDULING:
```
pquentin: I'm neither a native speaker or a fluent speaker, but I'm confused by `_ALLOW_DETERMINISTIC_SCHEDULING`. To me, that means "the scheduling could be deterministic, because it's allowed, but maybe it's not? who knows".

Is `_ENFORCE_DETERMINISTIC_SCHEDULING` a better name?

Zac-HD: > To me, [`_ALLOW_DETERMINISTIC_SCHEDULING`] means "the scheduling could be deterministic, because it's allowed, but maybe it's not? who knows".

That's exactly what it *does* mean!  If this flag is set to True, you *can* enable determinism by seeding the Random instance `_r`, but it's still not deterministic by default.

Issue #886, opened 2019-01-28
Cancel scopes with disjoint extents


njsmith:
So! Linked cancel scopes! This is an idea that was first raised in https://github.com/python-trio/trio/issues/607#issuecomment-413739258, and also discussed in https://github.com/python-trio/trio/pull/835. The latter also has some notes on implementation subtleties that I won't copy here, but are worth reviewing if/when this turns into a PR again.

Here's the problem: If you want to have a single `CancelScope` that applies to multiple disjoint tasks, then how do you do that? Even with unbound cancel scopes (#835), you can't, because a `CancelScope` tracks information about both what's been requested, and what actually happened to a specific block of code (`.cancelled_caught`, or maybe `.cancel_succeeded`, see #885). If the same scope object could be used for multiple pieces of code simultaneously, then this wouldn't work.

One idea is to let a cancel scope fork off "linked children": new `CancelScope` objects that are their own independent entity, but that whenever the parent is cancelled this automatically cancels the children as well.

(Do we have any example use cases? I feel like we did, but I don't remember them off the top of my head.)

There's a lot about this that I like, but there are also a lot of potential complexities to think through:

* Introspection: should there be a way to get the parent scope? The child scopes? The descendent scopes? The initial version of #835 had a `.linked_children` attribute that returned all descendents.

* `current_effective_deadline`: I guess this would need to walk up the parent chain for each cancel scope to collect up all their deadlines? (Which means we need to store parent links internally.)

* How do shield attributes and linked scopes interact? In the initial version of #835, changes to `parent.shield` propagated to `child.shield`, but this felt very weird to me. Conceptually, the shield component of a cancel scope is pretty much disjoint from the cancellation component. Maybe this is more evidence for the idea raised in #607, that shielding should be split off into a separate object from cancel scopes?

* If we add a "soft" cancellation concept (#147), then how will that interact with this?

* How tightly coupled should the parent and the children be? The two possibilities that come to mind are: (a) the idea described above, where the children are totally independent, full-fledged `CancelScope` objects, except that `parent.cancel()` triggers a call to `child.cancel()`. (b) making them "linked clones", where the parent's deadline is the child's deadline, mutating one deadline mutates them all, cancelling any of them cancels them all, etc. The only thing that would be independent is `.cancelled_caught`, and maybe in the future [some kind of data about what exceptions were caught](https://github.com/python-trio/trio/issues/285). 

[...]

Issue #860, opened 2019-01-17
Delay deciding which cancel scope a Cancelled exception belongs to


njsmith:
Right now, whenever we create a `Cancelled` exception, we immediately tag it with the cancel scope that's going to catch it.

There's another option that's briefly mentioned in #606: we could wait until `CancelScope.__exit__` to decide whether a given `Cancelled` exception "belongs" to this scope. The check would be "on the current task's stack, am I the outermost unshielded scope that's in the cancelled state?"

This can behave differently from what we have now in cases where there are multiple cancelled `CancelScope`s on the stack.

I'm not entirely sure if this would actually help with #606, because of complications noted there (hello `KeyboardInterrupt`). But it would probably simplify the cancel scope code in general, in particular `_cancel_no_notify`, and the issues described in this comment: https://github.com/python-trio/trio/pull/835#issuecomment-455107324

oremanj:
I just posted #900 which offers a less elegant, but also less invasive, solution to this problem. I prefer the solution in this issue on aesthetic grounds but I couldn't convince myself that it would be safe in the presence of a cancellation landscape that might be changing as the `Cancelled` exception propagates. Also, it makes the delivery of a cancellation be quadratic in the depth of the cancel stack, but probably that doesn't add up to much in practice?

njsmith:
> couldn't convince myself that it would be safe in the presence of a cancellation landscape that might be changing as the `Cancelled` exception propagates

Good question, yeah.

Right now, at the moment when we inject the `Cancelled` exception, we look at a single consistent snapshot of the whole stack, determine where the outermost-unshielded-cancelled-scope is, and then lock that in as the destination of the exception.

In this proposal, each time the exception passes through a cancel scope, we look at a single consistent snapshot of the whole stack, and determine whether the current scope is the outermost-unshielded-cancelled-scope. (So the naive implementation is O(stack depth * number of scopes in the cancel-requested state), which is probably much better than O(stack depth * stack depth), and I think this could be optimized further.)

Now, if we discover that this *is* the outermost-unshielded-cancelled-scope, and catch the exception, then that's definitely OK, because it was the right thing to do at the moment we did it, and then there's nothing afterwards that could go wrong.

[...]

PR #900, opened 2019-02-03
Add trio.hazmat.batch_cancellations()

oremanj:
This generalizes the "cancel multiple scopes without waking up the tasks, then wake up all the tasks at once" pattern that was previously a special case for handling cancellations on a deadline. It will simplify the
implementation of #886, and exposing it as part of hazmat lets users write their own cancellation
abstractions that work the way the native ones do.

### DIFF
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -492,6 +492,13 @@ this does serve to illustrate the basic structure of the
                trio.hazmat.reschedule(woken_task)
 
 
+Low-level cancellation control
+------------------------------
+
+.. autofunction:: batch_cancellations()
+   :with:
+
+
 Task API
 --------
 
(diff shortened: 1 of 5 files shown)


njsmith:
What do you think of #860 as a way to eliminate the need for this entirely?

oremanj:
I had looked at that, and I agree that it would be cleaner if we can get it to work, but I'm leery of lurking bugs if the cancellation landscape changes in just the wrong way while a `Cancelled` exception is propagating. I didn't feel confident in my own ability to correctly do the analysis that would disprove it, at any rate. OTOH, I also tried for a while to demonstrate a specific issue and couldn't find any, so maybe #860 is a better approach after all?

oremanj:
We decided #860 works after all, so let's go with that instead!

PR #901, opened 2019-02-04
Delay deciding which cancel scope a Cancelled exception belongs to

### LINKED ISSUE #860: Delay deciding which cancel scope a Cancelled exception belongs to

oremanj:
Fixes #860.

### DIFF
--- a/trio/_core/_exceptions.py
+++ b/trio/_core/_exceptions.py
@@ -46,13 +46,9 @@ class Cancelled(BaseException):
 
     then this *won't* catch a :exc:`Cancelled` exception.
 
-    Attempting to raise :exc:`Cancelled` yourself will cause a
-    :exc:`RuntimeError`. Unless raised within the context of a
-    cancelled cancel scope, it would not be caught by Trio and would
-    crash your program. Use :meth:`cancel_scope.cancel()
-    <trio.CancelScope.cancel>` instead, so Trio can maintain the
-    invariant that a :exc:`Cancelled` exception only gets raised when
-    there's a cancelled cancel scope that's intended to catch it.
+    You cannot raise :exc:`Cancelled` yourself. Attempting to do so
+    will produce a :exc:`RuntimeError`. Use :meth:`cancel_scope.cancel()
+    <trio.CancelScope.cancel>` instead.
 
     .. note::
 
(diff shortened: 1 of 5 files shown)


trio/_core/_exceptions.py
```diff
@@ -47,10 +47,12 @@ class Cancelled(BaseException):
...
-    :exc:`RuntimeError`. It would not be associated with a cancel scope and thus
-    not be caught by Trio. Use
-    :meth:`cancel_scope.cancel() <trio.CancelScope.cancel>`
-    instead.
+    :exc:`RuntimeError`. Unless raised within the context of a
+    cancelled cancel scope, it would not be caught by Trio and would
+    crash your program. Use :meth:`cancel_scope.cancel()
+    <trio.CancelScope.cancel>` instead, so Trio can maintain the
+    invariant that a :exc:`Cancelled` exception only gets raised when
+    there's a cancelled cancel scope that's intended to catch it.
```
njsmith: This is probably more details on the internals than we really want people thinking about. How about simplifying down to just: "You cannot raise Cancelled yourself. Attempting to do so will produce a RuntimeError. Use cancel_scope.cancel instead."

oremanj:
The 3.8-dev failure in Travis is unrelated, so I think this is ready.

njsmith:
LGTM. That travis failure is annoying... I just filed a bug on CPython ([bpo-35894](https://bugs.python.org/issue35894)), so hopefully it will be fixed soon. In the mean time though our CI is broken. I guess we need to temporarily disable the 3.8-dev test (or mark it "allowed to fail") until that's fixed?

njsmith:
Closing/opening to tickle CI, now that #905 is merged.

Issue #636, opened 2018-08-29
The Listener interface is at the wrong level of abstraction


njsmith:
We want to have an abstraction over the idea of "listening for incoming connection", because we want to be able to mix-and-match between protocol implementations and transport implementations – e.g. if you have an HTTP server or a Websocket server, you should be able to run it over TCP, over Tor, over TLS, over the haproxy PROXY protocol, over SSH tunneling, ... etc. Twisted gets a lot of mileage about of this idea with their "endpoints" system – one example that's particularly slick is [txacme](https://txacme.readthedocs.io/), which lets you transparently enable Let's Encrypt certs for any Twisted server. It works by defining a custom listener that detects when Let's Encrypt is connecting to validate a cert, and handles those connections automatically instead of passing them on to the application.

Our current abstraction is: a `Listener` is an object with `accept` and `aclose` methods. In practice, users mostly don't call these methods directly; the main user of this interface is `serve_listeners`, which calls `accept` in a loop, spawns tasks for incoming connections, and `aclose`s the listener when cancelled.

This basically works, but over time I've been building up a list of little issues that make me itch, and wonder if maybe we've drawn the abstraction boundary at the wrong place.

# Graceful shutdown of Unix-domain socket servers

If you have a server that's listening on a Unix socket, it's possible to do a zero-downtime upgrade, without any complicated protocol for handing off the socket between the old process and the new one. You can have the new process atomically bind to the listening address, and after that point all new incoming connections go to the new process, while existing connections stay with the old process. Then you tell the old process to shut down once all current connections are handled. (Haproxy for example does zero-downtime upgrades this way.)

To do this correctly, though, your accept loop needs a way to "drain" all the connections that have already happened, without waiting for any more to arrive. But that's not possible with the current `Listener` interface: we would need something like `accept_nowait`, which doesn't currently exist.

Maybe this doesn't matter because it's better to do zero-downtime upgrades via file descriptor passing, which is much more general solution (in particular, it can work for TCP services, which this can't!). Or we could add `accept_nowait`, which would work but is kind of awkward since then wrappers like `SSLListener` would have to wrap both `accept` and `accept_nowait`.

Another case where `accept_nowait` might be useful is in "batched accept" (see #14). Again, I'm not totally sure whether this is something we actually need, but maybe. Or batched accept might be something we need temporarily to work around other issues, but then eventually get rid of.

So... maybe `accept` alone isn't quite enough, but it isn't obvious what to do.

[...]

PR #908, opened 2019-02-05
Permit referring to some generic types in generic ways

oremanj:
Specifically, declare `SendChannel`, `ReceiveChannel`, `Listener`, and `RunVar` to
be generic in one type parameter, and also support the `open_memory_channel[T](bufsize)`
syntax at runtime.  Until trio is able to support typing directly, this change allows
users of external stubs to use correctly-typed channels without too many hacks.

### DIFF
--- a/trio/_abc.py
+++ b/trio/_abc.py
@@ -484,9 +484,18 @@ class HalfCloseableStream(Stream):
         """
 
 
-T = TypeVar("T")
+# The type of object produced by a ReceiveChannel (covariant because
+# ReceiveChannel[Derived] can be passed to someone expecting
+# ReceiveChannel[Base])
 T_co = TypeVar("T_co", covariant=True)
+
+# The type of object accepted by a SendChannel (contravariant because
+# SendChannel[Base] can be passed to someone expecting
+# SendChannel[Derived])
 T_contra = TypeVar("T_contra", contravariant=True)
+
+# The type of object produced by a Listener (covariant plus must be
+# an AsyncResource)
 T_resource = TypeVar("T_resource", bound=AsyncResource, covariant=True)
 
 
@@ -501,7 +510,7 @@ class Listener(AsyncResource, Generic[T_resource]):
     __slots__ = ()
 
     @abstractmethod
-    async def accept(self) -> T_resource:
+    async def accept(self):
         """Wait until an incoming connection arrives, and then return it.
 
         Returns:
@@ -542,7 +551,7 @@ class SendChannel(AsyncResource, Generic[T_contra]):
     __slots__ = ()
 
     @abstractmethod
-    def send_nowait(self, value: T_contra) -> None:
+    def send_nowait(self, value):
         """Attempt to send an object through the channel, without blocking.
 
         Args:
@@ -560,7 +569,7 @@ class SendChannel(AsyncResource, Generic[T_contra]):
         """
 
     @abstractmethod
-    async def send(self, value: T_contra) -> None:
+    async def send(self, value):
         """Attempt to send an object through the channel, blocking if necessary.
 
         Args:
@@ -577,7 +586,7 @@ class SendChannel(AsyncResource, Generic[T_contra]):
         """
 
     @abstractmethod
-    def clone(self: T) -> T:
+    def clone(self):
         """Clone this send channel object.
 
         This returns a new :class:`SendChannel` object, which acts as a
@@ -625,7 +634,7 @@ class ReceiveChannel(AsyncResource, Generic[T_co]):
     __slots__ = ()
 
     @abstractmethod
-    def receive_nowait(self) -> T_co:
+    def receive_nowait(self):
         """Attempt to receive an incoming object, without blocking.
 
         Returns:
@@ -644,7 +653,7 @@ class ReceiveChannel(AsyncResource, Generic[T_co]):
         """
 
     @abstractmethod
-    async def receive(self) -> T_co:
+    async def receive(self):
         """Attempt to receive an incoming object, blocking if necessary.
 
         It's legal for multiple tasks to call :meth:`receive` at the same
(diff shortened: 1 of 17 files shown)


njsmith:
What's the advantage of keeping the actual stubs out of trio here?

Oh, I see the force-push. We gotta fix sphinxcontrib-trio to handle these better...

oremanj:
Putting the type hints in trio currently has limited utility because in the absence of stubs mypy chokes when you try to `import trio` due to the `_core._run` magic (among other things). I wanted to include them on the generic classes because why not, but sphinx didn't do well with that. It might just fix itself if we enable the `sphinx_autodoc_typehints` extension, but that can be a separate PR, I think. :-)

More context here: I'm working on a `trio-typing` package which provides stubs for trio, async_generator, and outcome, and some other stuff like a mypy plugin to typecheck `nursery.start_soon(fn, *args)` correctly. I figure that until we're ready to fully tackle in-tree type hints (definitely desirable! but a much larger undertaking), the small change here can serve as a decent stopgap that lets people start experimenting.

oremanj:
Looks like 3.5.0 doesn't support instantiating generic slots classes. :-( I'll remove the RunVar genericization because it's much less usability-relevant than the others: not many people create RunVars, and no one inherits from them so the workaround is just quotes rather than some elaborate `if TYPE_CHECKING:` thing.

trio/_abc.py
```diff
@@ -483,7 +484,22 @@ class HalfCloseableStream(Stream):
...
+T_co = TypeVar("T_co", covariant=True)
+
+# The type of object accepted by a SendChannel (contravariant because
+# SendChannel[Base] can be passed to someone expecting
+# SendChannel[Derived])
+T_contra = TypeVar("T_contra", contravariant=True)
+
+# The type of object produced by a Listener (covariant plus must be
+# an AsyncResource)
+T_resource = TypeVar("T_resource", bound=AsyncResource, covariant=True)
```
njsmith: I think the `bound=AsyncResource` will disappear when we fix #636. We could start skipping it now, or we could put it in and then take it out later, doesn't make much difference IMO.

trio/_abc.py
```diff
@@ -483,7 +484,22 @@ class HalfCloseableStream(Stream):
...
+
+
+class Listener(AsyncResource, Generic[T_resource]):
```
njsmith: Perhaps we should also update `SocketListener` to inherit from `Listener[SocketStream]`, `SSLListener` to inherit from `Listener[SSLStream]`, etc.?

trio/_util.py
```diff
@@ -22,6 +22,7 @@
     "ConflictDetector",
     "fixup_module_metadata",
     "fspath",
+    "indexable_function",
```
njsmith: `generic_function`

trio/_util.py
```diff
@@ -177,7 +178,9 @@ def fix_one(obj):
             obj.__module__ = module_name
             if isinstance(obj, type):
                 for attr_value in obj.__dict__.values():
-                    fix_one(attr_value)
+                    # avoid infinite recursion when using typing.Generic
+                    if attr_value is not obj:
+                        fix_one(attr_value)
```
njsmith: Maybe it would be better to do more generic loop detection, by keeping a `set` of the ids of all objects we've already seen? OTOH this could wait until we actually need it...

trio/_util.py
```diff
@@ -242,3 +245,31 @@ def fspath(path) -> t.Union[str, bytes]:
...
+
+    def __init__(self, fn):
+        update_wrapper(self, fn)
+        self._fn = fn
+
+    def __call__(self, *args, **kwargs):
+        return self._fn(*args, **kwargs)
+
+    def __getitem__(self, _):
+        return self
```
njsmith: So you actually can make `open_memory_channel` directly typeable by mypy without any plugin, and it's not too terribly awful: https://github.com/python/mypy/issues/6073#issuecomment-447616140

But I think it does require inlining this class into the definition of `open_memory_channel`.

If that's where were going to end up, then maybe setting up a standalone `@generic_function` decorator, testing it, etc., is overkill, and we should instead move this code into `_memory_channels.py` right now?

trio/tests/test_abc.py
```diff
@@ -19,3 +19,24 @@ class MyAR(tabc.AsyncResource):
         assert myar.record == []
 
     assert myar.record == ["ac"]
+
+
+def test_abc_generics():
```
njsmith: It'd be nice to have a comment on this explaining why we thought such an odd-looking test is useful :-)

trio/_abc.py
oremanj: You think we'll have Listeners that provide objects not representable as an AsyncResource? I'm a bit surprised at that, can you give an example?

trio/_util.py
oremanj: The problem with that approach is that AFAICT, when you write `Type[T]`, mypy can't infer `T` as one of its "special forms" (Tuple, Union, Callable, etc). You just silently get `Any` if you use such a one. I figured that was error-prone enough that we should go this route instead, at least for now.

oremanj:
CI problem is codecov.io returning an invalid script

trio/_util.py
njsmith: Oh ick.

trio/_abc.py
njsmith: I don't have an example in mind. But the reason we restrict it to `AsyncResource` right now is because `serve_listeners` automatically closes the object after the handler returns: https://github.com/python-trio/trio/blob/0a815a015bd06fa6ff929d3bb881c9d44e6d9c8a/trio/_highlevel_serve_listeners.py#L25-L29

With #636, this code goes away, because the `Listener` itself is responsible for managing the handler lifetime. So... at that point it's really between the person developing a `Listener[X]` and their users, what kind of API `X` should support :-).

oremanj: Ah, that makes sense! I think it's probably sensible to keep the AsyncResource bound for now, since it reflects the current requirements, and we can easily change it later.

PR #909, opened 2019-02-05
Cancel a cancel scope upon entry if its deadline is in the past

### LINKED ISSUE #320: Maybe move_on_after(0)/fail_after(0) should immediately cancel their scope

oremanj:
Fixes #320. Now that #901 is implemented, the concerns discussed in #835 don't apply.

### DIFF
--- a/newsfragments/320.misc.rst
+++ b/newsfragments/320.misc.rst
@@ -1,0 +1,4 @@
+Entering a cancel scope whose deadline is in the past now immediately
+cancels it, so :exc:`~trio.Cancelled` will be raised by the first
+checkpoint in the cancel scope rather than the second one.
+This also affects constructs like ``with trio.move_on_after(0):``.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -201,6 +201,8 @@ class CancelScope:
             )
         self._scope_task = task
         self.cancelled_caught = False
+        if current_time() >= self._deadline:
+            self.cancel_called = True
         with self._might_change_effective_deadline():
             self._add_task(task)
         return self
--- a/trio/_core/tests/test_run.py
+++ b/trio/_core/tests/test_run.py
@@ -407,12 +407,13 @@ def test_instruments(recwarn):
 
     _core.run(main, instruments=[r1, r2])
 
-    # It sleeps 5 times, so it runs 6 times
+    # It sleeps 5 times, so it runs 6 times.  Note that checkpoint()
+    # reschedules the task immediately upon yielding, before the
+    # after_task_step event fires.
     expected = (
-        [("before_run",)] +
-        6 * [("schedule", task),
-             ("before", task),
-             ("after", task)] + [("after_run",)]
+        [("before_run",), ("schedule", task)] +
+        [("before", task), ("schedule", task), ("after", task)] * 5 +
+        [("before", task), ("after", task), ("after_run",)]
     )
     assert len(r1.record) > len(r2.record) > len(r3.record)
     assert r1.record == r2.record + r3.record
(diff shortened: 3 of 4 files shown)


njsmith:
Test failure is an intermittent failure in `test_open_tcp_listeners_backlog`:

* I made a note in https://github.com/python-trio/trio/issues/200#issuecomment-451885495
* I clicked the button on azure to "retain" the build, so we'll have the logs for reference
* I edited the top comment on #200 to note the requirement to "retain" the build
* Clicked "Queue" to queue a re-build (I think?)

njsmith:
OK, so the "Queue" button is a trap. It definitely didn't do what we want.

Now I tried clicking the "Rebuild" button that's hidden behind the "..." menu. Maybe that will work better?

oremanj:
It looks like the rerun worked (https://dev.azure.com/python-trio/trio/_build/results?buildId=152) but github didn't notice, so I guess I'll do the close-and-reopen dance.

njsmith:
Hmm, azure's UX could definitely use some improvement here :-)

Anyway, the PR seems fine :-). I'm not sure it makes a huge difference either way, but it's certainly a bit closer to what people expect, so why not.

PR #911, opened 2019-02-07
Disable cancel scope reuse

oremanj:
To keep our options open for adding a smarter `was_cancelled` property, per discussion in #886.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -164,8 +164,9 @@ class CancelScope:
     has been entered yet, and changes take immediate effect.
     """
 
-    _tasks = attr.ib(factory=set, init=False)
-    _scope_task = attr.ib(default=None, init=False)
+    # _tasks is None before we enter the cancel scope, and the set of
+    # tasks with this scope on their cancel stacks afterward
+    _tasks = attr.ib(default=None, init=False)
     _effective_deadline = attr.ib(default=inf, init=False)
     cancel_called = attr.ib(default=False, init=False)
     cancelled_caught = attr.ib(default=False, init=False)
@@ -177,12 +178,11 @@ class CancelScope:
     @enable_ki_protection
     def __enter__(self):
         task = _core.current_task()
-        if self._scope_task is not None:
+        if self._tasks is not None:
             raise RuntimeError(
-                "cancel scopes may not be reused or reentered "
-                "(first used in task {!r})".format(self._scope_task.name)
+                "Each CancelScope may only be used for a single 'with' block"
             )
-        self._scope_task = task
+        self._tasks = set()
         if current_time() >= self._deadline:
             self.cancel_called = True
         with self._might_change_effective_deadline():
@@ -214,14 +214,14 @@ class CancelScope:
                 value.__context__ = old_context
 
     def __repr__(self):
-        if self._scope_task is None:
+        if self._tasks is None:
             binding = "unbound"
+        elif not self._tasks:
+            binding = "exited"
         else:
-            binding = "bound to {!r}".format(self._scope_task.name)
-            if len(self._tasks) > 1:
-                binding += " and its {} descendant{}".format(
-                    len(self._tasks) - 1, "s" if len(self._tasks) > 2 else ""
-                )
+            binding = "bound to {} task{}".format(
+                len(self._tasks), "s" if len(self._tasks) > 1 else ""
+            )
 
         if self.cancel_called:
             state = ", cancelled"
@@ -322,7 +322,7 @@ class CancelScope:
         if not isinstance(new_value, bool):
             raise TypeError("shield must be a bool")
         self._shield = new_value
-        if not self._shield:
+        if self._tasks and not self._shield:
             for task in self._tasks:
                 task._attempt_delivery_of_any_pending_cancel()
 
@@ -337,8 +337,9 @@ class CancelScope:
             return
         with self._might_change_effective_deadline():
             self.cancel_called = True
-        for task in self._tasks:
-            task._attempt_delivery_of_any_pending_cancel()
+        if self._tasks:
+            for task in self._tasks:
+                task._attempt_delivery_of_any_pending_cancel()
 
     def _add_task(self, task):
         self._tasks.add(task)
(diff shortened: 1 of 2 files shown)


trio/_core/_run.py
```diff
@@ -193,14 +179,10 @@ def __enter__(self):
...
             raise RuntimeError(
-                "cancel scope may not be entered while it is already "
-                "active{}".format(
-                    "" if self._scope_task is task else
-                    " in another task ({!r})".format(self._scope_task.name)
-                )
+                "cancel scopes may not be reused or reentered "
+                "(first used in task {!r})".format(self._scope_task.name)
             )
         self._scope_task = task
```
njsmith: Do we use `self._scope_task` for anything else now? If it's effectively a bool, it would probably be better to make an an actual bool, so we don't inadvertently hold onto any stack frames longer than we need to.

oremanj: We use it in `_exc_filter` (testing whether `self._scope_task._pending_cancel_scope() is self`). I also don't think holding onto a `Task` can actually leak much memory -- if the task is running, it's referenced by the runner so our reference doesn't change the memory usage, and if it's done, it doesn't have any stack frames so the memory usage is small. Is there something I'm missing?

njsmith: Maybe `CancelScope._exc_filter` should use `current_task()`? ...and maybe it should be refactored into `CancelScope._close`. And actually, now that all `Cancelled` exceptions are the same, we can hoist the check up out of `_exc_filter` entirely; no reason `_close` should call `MultiError.filter` at all if it's not going to catch anything. That's another matter though...

I guess you're right that coroutine objects drop their `cr_frame` reference when the coroutine exits. Huh, I didn't know that. That does reduce the likelihood of this reference causing problems. So now it still makes me nervous to hold onto an unnecessary reference to a complex object, but it's only on general principles, not because of any specific issue I can point to :-).

njsmith:
Huh, I would have added a `self._has_been_entered = False` field. It wouldn't have occurred to me overload `self._tasks` to track this as well. Is there a reason for it I'm missing? Sorry to nitpick, I just find it easier to read code where the state is as decoupled as possible...

oremanj:
I figured it would be one fewer slot. I'll add the separate bool.

oremanj:
Azure Ubuntu timeout again, same reason as last night.

oremanj:
Even slower on a rerun.

... are we getting any value out of the lone Linux test environment on Azure? Should we just disable it?

njsmith:
> are we getting any value out of the lone Linux test environment on Azure?

Unfortunately yes, it's our only environment that has openssl 1.1.1, and so far our experience has been that 1.1.1 is cranky and easy to break.

We could move it to Travis, or bump up the timeout as a temporary workaround...

It's super weird that the two times this has happened were both at 10:00-11:00 PM west coast time. I wonder if that's accidental.

pquentin:
I don't think so: I think Azure Ubuntu is getting popular, and this is going to happen every day.

njsmith:
@pquentin my experience with other CI systems though is that they get overloaded in the middle of the USA working day. OTOH we're [hosted](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/change-organization-location?view=azure-devops&tabs=new-nav) in the Azure "Central US" region in Iowa, so the slowdown seems to starting right at midnight local time. Suspicious...

pquentin:
Ah, that explanation is better! 👍 Maybe they have cron jobs running at midnight.

njsmith:
Hmm, actually, it's not *quite* at midnight.

The first time we had this problem was 2019-02-05 (UTC):

*  The last quick build was [20190205.4](https://dev.azure.com/python-trio/trio/_build/results?buildId=122) (ran in ~2 min), which started at `2019-02-05 5:57:15Z`.
* The first slow build was [20190205.5](https://dev.azure.com/python-trio/trio/_build/results?buildId=123), which started at `2019-02-05T06:39:20Z`.
* Then we had a series of slow builds.
* Build [20190205.10](https://dev.azure.com/python-trio/trio/_build/results?buildId=128) was borderline – it didn't quite hit the timeout, but squeaked in at ~9:45 min. It started at `2019-02-05T07:07:27Z`.
* The next build, [20190205.11](https://dev.azure.com/python-trio/trio/_build/results?buildId=129) was definitely fast again – ~2 min runtime, started at `2019-02-05T07:35:19Z`.

So this episode started somewhere between 05:57 and 06:39 UTC, and ended somewhere between 07:07 and 07:35 UTC.

The next day, 2019-02-05 UTC, we didn't run any builds between [`2019-02-05T23:51:12Z`](https://dev.azure.com/python-trio/trio/_build/results?buildId=153) and [`2019-02-06T08:09:57Z`](https://dev.azure.com/python-trio/trio/_build/results?buildId=154), so nowhere near the period where the slowdown occurred.

Then then the day after that puts us at today, 2019-02-06 UTC:

* Build [20190207.11](https://dev.azure.com/python-trio/trio/_build/results?buildId=171) was quick (~2 min), and started at `2019-02-07T06:15:08Z`. In Iowa, that's 00:15, so today it was still fast a few minutes *after* local midnight.
* The next build, [20190207.12](https://dev.azure.com/python-trio/trio/_build/results?buildId=172), was slow, and started at `2019-02-07T06:35:21Z`
* Then we had a series of slow builds, the last one being [20190207.14](https://dev.azure.com/python-trio/trio/_build/results?buildId=174), which started at `2019-02-07T07:04:15Z`
* And then [20190207.15](https://dev.azure.com/python-trio/trio/_build/results?buildId=175) was only slightly slow (~4 min), and started at `2019-02-07T07:15:17Z`

So this time the slowdown started between 06:15 and 06:35 UTC, and ended around 07:15 UTC. That is still *remarkably* consistent with the previous window.

I also peeked back one day into the past, and it turns out that we did do a build at [`2019-02-04T06:42:28Z`](https://dev.azure.com/python-trio/trio/_build/results?buildId=117), which went fine. So either the temporal consistenty is a total coincidence, or else this is a new problem that just started on 2019-02-05.

PR #915, opened 2019-02-07
Trim down CI

njsmith:
Azure pipelines seems to be working fine for Windows and macOS, so
let's clean things up a bit:

- Removes appveyor.yml
- Removes Jenkinsfile
- Removes macOS testing from Travis

I also switched the repo config so now azure pipelines is required,
and appveyor is not. And I disabled the webhook for appveyor.


pquentin:
The behavior of yapf changes with Python 3.7:

```
--- trio/_signals.py	(original)
+++ trio/_signals.py	(reformatted)
@@ -177,7 +177,7 @@
         return self
 
     async def __anext__(self):
-        return { await self._signal_queue.__anext__()}
+        return {await self._signal_queue.__anext__()}
 
 
 @deprecated("0.7.0", issue=354, instead=open_signal_receiver)
```

Unfortunately, running yapf on Python 3.6 will add the space back.

njsmith:
Ugh, right, fine.

pquentin:
I was simply exposing the problem: we can also decide that the Python 3.7 version is better and expect that most contributions will use 3.7. Or switch to black. :)

Anyway, thanks for the CI improvements!

njsmith:
Yeah, I'm grumpy at software, not at you :-). We should do something about this at some point, but I'm not feeling inspired to deal with it tonight...

PR #919, opened 2019-02-09
Release v0.11.0

oremanj:


oremanj:
The wiki says "PR against `release` branch" but that doesn't seem to exist, so I sent a PR against master.

njsmith:
> The wiki says "PR against `release` branch" but that doesn't seem to exist, so I sent a PR against master.

Yeah, that's out of date. If you could fix it that'd be great :-)

This looks good. I guess usually we put the second bump to add `+dev` into this PR too?

njsmith:
Oh right, and what's your PyPI username?

pquentin:
(I think we actually need to delete the wiki page or make it point to https://trio.readthedocs.io/en/latest/releasing.html, and then fix any issues there.)

oremanj:
> I guess usually we put the second bump to add +dev into this PR too?

I assumed we would want a commit on master that was exactly 0.11.0, including the version number, so that we could tag it as 0.11.0. If I'm misunderstanding and this PR should bump from 0.10.0+dev straight to 0.11.0+dev, I can do that, but would love to understand why. :-)

> Probably https://pypi.org/user/oremanj/ but let's make sure it's the real @oremanj first!

Yep, that's me!

njsmith:
> I assumed we would want a commit on master that was exactly 0.11.0, including the version number, so that we could tag it as 0.11.0. If I'm misunderstanding and this PR should bump from 0.10.0+dev straight to 0.11.0+dev, I can do that, but would love to understand why. :-)

What I usually do is:

* make a branch
* in the branch, make a commit that bumps to 0.11.0 [this is the commit that will be tagged as 0.11.0]
* in the branch, make another commit that bumps to 0.11.0+dev
* merge the branch into master

As long as you don't do a "squash merge", both of those commits do actually end up in master.

You also have to push tags separately, because there's no way to include a tag in a PR...

njsmith:
> Yep, that's me!

OK, you have upload rights on trio now :-)

PR #937, opened 2019-02-19
Remove undocumented public API usage of UnboundedQueue

lojack5:
Long time lurker here, I was tinkering around with using IOCP on Windows (messing with ReadDirectoryChanges), dug into some of the undocumented public methods here, as well as the deprecated warning going along with its usage of UnboundedQueue.

So: I updated it to use channels instead, (as well as the kqueue IOManager version).  I have a few questions though:

1. The kqueue IOManager doesn't appear to have any tests for the similar API (`monitor_kevent`).  Should one be written? And if so maybe some guidance on tests, I'm pretty new to those, and don't have the appropriate machine setup ATM.  This also seems to be why coverage dropped?
2. My second commit adds a synchronous `close()` method to channels (does the exact same thing, just never awaits a checkpoint).  Not sure if this is a thing that makes sense to have or not, but it seemed like it was something that should be used to close the send side of the channel when the IOManager doesn't need it anymore.  Maybe there's a better way (queue the `aclose()` version in the scheduler?).
3. This changes a public API (`monitor_completion_key` and `monitor_kevent` yield `MemoryRecieveChannel`s instead of `UnboundedQueue`s).  But the API was undocumented to begin with...should I add documentation?  Should a news fragment go with this since *someone* may have been relying on the old interface?  And if so, should this wait until `UnboundedQueue` goes from deprecated to removed?
4. As far as I can tell `UnboundedQueue` isn't used anywhere else after these changes (I know, not a question).

This is pretty much just to squash the deprecation warning, as I tinker around I'll have more things I might want to change (ex: the context manager for `monitor_completion_key` is cumbersome to use some cases, though that could just be cause I'm designing my classes wrong for the API).

### DIFF
--- a/trio/_channel.py
+++ b/trio/_channel.py
@@ -168,8 +168,12 @@ class MemorySendChannel(SendChannel):
 
     @_core.enable_ki_protection
     async def aclose(self):
+        self.close()
+        await _core.checkpoint()
+
+    @_core.enable_ki_protection
+    def close(self):
         if self._closed:
-            await _core.checkpoint()
             return
         self._closed = True
         for task in self._tasks:
@@ -183,7 +187,6 @@ class MemorySendChannel(SendChannel):
                 task.custom_sleep_data._tasks.remove(task)
                 _core.reschedule(task, Error(_core.EndOfChannel()))
             self._state.receive_tasks.clear()
-        await _core.checkpoint()
 
 
 @attr.s(cmp=False, repr=False)
@@ -250,8 +253,12 @@ class MemoryReceiveChannel(ReceiveChannel):
 
     @_core.enable_ki_protection
     async def aclose(self):
+        self.close()
+        await _core.checkpoint()
+
+    @_core.enable_ki_protection
+    def close(self):
         if self._closed:
-            await _core.checkpoint()
             return
         self._closed = True
         for task in self._tasks:
@@ -266,4 +273,3 @@ class MemoryReceiveChannel(ReceiveChannel):
                 _core.reschedule(task, Error(_core.BrokenResourceError()))
             self._state.send_tasks.clear()
             self._state.data.clear()
-        await _core.checkpoint()
--- a/trio/_core/_io_kqueue.py
+++ b/trio/_core/_io_kqueue.py
@@ -1,4 +1,5 @@
 import select
+import math
 
 import outcome
 from contextlib import contextmanager
@@ -6,6 +7,7 @@ import attr
 
 from .. import _core
 from . import _public
+from .._channel import open_memory_channel
 
 
 @attr.s(frozen=True)
@@ -18,7 +20,7 @@ class _KqueueStatistics:
 @attr.s(slots=True, cmp=False, hash=False)
 class KqueueIOManager:
     _kqueue = attr.ib(default=attr.Factory(select.kqueue))
-    # {(ident, filter): Task or UnboundedQueue}
+    # {(ident, filter): Task or MemorySendChannel}
     _registered = attr.ib(default=attr.Factory(dict))
 
     def statistics(self):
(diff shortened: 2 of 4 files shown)


njsmith:
Hey, welcome!

So here's the deal with those low-level APIs: I stuck them in as placeholders when I was first figuring out how Trio's IO should work at all, as a kind of "statement of intent". But then no-one has actually used them so we haven't taken the time to figure out how they should *actually* work :-). If you want to pick that up, then that'd be awesome!

There's some discussion in #26, #578, #579, especially this comment: https://github.com/python-trio/trio/issues/578#issuecomment-408760107

Those are just random thoughts at different times though, not like, Pronouncements On The Proper Way To Do Things.

I guess the first question is, what is even useful. For Windows, we definitely want a way to handle the super-common pattern of "calling a function that takes an `Overlapped`". The `register_with_iocp` and `wait_overlapped` functions do currently allow that. Are they the nicest way? Not sure, they haven't been used a lot. (I think this is what you need for `ReadDirectoryChanges`?)

Do we also need a way to monitor a completion key? We have one because I was messing around with `PostQueuedCompletionStatus`, but I don't know if it's actually useful in general. Ditto with the `current_iocp` function.

There are similar questions for kevent.

And if we do want the `monitor` functions, then ... there's a few different ways to do it. We could use a custom iterable, a channel like in this PR, pass in a channel from outside (which allows multiple streams to be monitored on a single channel, and means we don't have to close it)... probably this would be easier to answer if we knew what the purpose of these monitor functions was :-)

Do you have any thoughts?

lojack5:
So my first attempt was in using `register_with_iocp` and `wait_overlapped`.  From what I know now I think I could get that to work fine, but here's where waiting on a completion key helped me out a lot:

1.  I'm new to IOCP, so I'm not 100% sure what will work, and been decyphering MSDN pages for structs, calls, etc.  Because of this, I found it extremely useful to get the `dwNumberOfBytesTransferrred` after the IOCP completes, which you need to wait on the completion key to get.  This really helped me to make have a second check in there to make sure I wasn't reading past actual data in my buffer into uninitialized data in my buffer.
2. Now that I've learned a bit more though, especially reading [here](https://docs.microsoft.com/en-us/windows/desktop/api/winnt/ns-winnt-_file_notify_information), I think I can get it safely working by just checking for `NextEntryOffset == 0`, and relying on IOCP only giving a result if one or more `FILE_NOTIFY_INFORMATION` struct has been written to the buffer...I think.  There might be some weird cases with error codes.

***
So with the above and responding to the rest of your thoughts:
1.  Definitely need `register_with_iocp` and `wait_overlapped`.  Currently I think the interface is fine with those.
2.  It'd be __nice__ to be able to also wait on completion keys, though the interface there can use a overhaul (I'll try to think up a good seeming way to do that based on what I had to work with).
3. `current_iocp` currently is needed if you want to wait on completion keys, since the user has to call `CreateIoCompletionPort` themselves.  This could probably be avoided by handling 2 above.
4. `kevent` stuff I have no knowledge of and don't have much experience with the platforms involved either.
5. I'll reread the linked issues to see if it sparks more ideas!

Thanks!  I'll try to get back within a week with at least a little bit of the above.

njsmith:
> I'm new to IOCP, so I'm not 100% sure what will work, and been decyphering MSDN pages for structs, calls, etc. Because of this, I found it extremely useful to get the `dwNumberOfBytesTransferrred` after the IOCP completes, which you need to wait on the completion key to get. This really helped me to make have a second check in there to make sure I wasn't reading past actual data in my buffer into uninitialized data in my buffer.

Oh good point! We should make `wait_overlapped` return the stuff in [`OVERLAPPED_ENTRY`](https://docs.microsoft.com/en-us/windows/desktop/api/MinWinBase/ns-minwinbase-_overlapped_entry), including the `dwNumberOfBytesTransferred`. We have the info, no reason to hide it from users.

lojack5:
Ok, after toying around with this I think going with making `wait_overlapped` will be a less intrusive change and I think it'll cover all the cases I'd want to use it for.  I'll go ahead an close this PR, and I've already implemented a test change on my fork, I'll tinker a bit more, then open a new PR once I'm happy with it.

njsmith:
Sounds good!

Issue #885, opened 2019-01-27
Can we talk about "cancelled" instead of "cancelled_caught" and "raised Cancelled" and similar circumlocutions?


njsmith:
In [this comment](https://github.com/python-trio/trio/issues/328#issuecomment-457643176), @mehaase wrote:

> > Most low-level operations in trio provide a guarantee: if they raise trio.Cancelled, this means that they had no effect
>
> I didn't understand this passage at first, because of the phrase "if they raise trio.Cancelled", which makes it sound like it got cancelled for some internal reason. From the user's point of view, it would make more sense to say, "if a call to `send_all()` is cancelled..."

I'm actually not sure what he's quoting, but it sure sounds like something I'd write in the docs, especially back when I was still struggling to come to grips with fine distinctions like `cancel_scope.cancel_called` versus `cancel_scope.cancelled_caught`. And I'm pretty sure the reason I didn't say "if a call to `send_all()` is cancelled" is because *technically* you could call `.cancel()` on a scope surrounding a call to `send_all` *but* the call didn't raise `Cancelled` because it had already passed the point of no return. Which... is *technically* accurate, but as Mark points out, *super confusing to ordinary humans.* And probably saying "if a call to `send_all()` is cancelled" would be totally clear to actual readers; if the tiny edge case was pointed out they'd just shrug and think "well, I guess it wasn't cancelled then".

And btw the distinction between `.cancel_called` and `.cancelled_caught` is also super confusing to regular humans.

Maybe we should change the official way of talking about this to more like: `cancel_scope.cancel()` *requests* a cancellation, but code might or might not actually *be cancelled* (= raise `Cancelled`).

That would suggest renaming the `CancelScope` attributes to something like `.cancel_requested` and `.cancel_succeeded`, and also probably updating the docs to use this language more uniformly (though I haven't read through the docs to check).

njsmith:
> (a) the called code didn't do anything, raising a Cancelled exception [...] Of course there's also the possibility of (c) the called code got part way through and left whatever it tried to accomplish in an inconsistent state.

Unfortunately I don't think the cancel scope machinery has any realistic way to tell the difference between these two cases. I guess we could try doing something wacky like counting how many checkpoints there were (cancelled on first checkpoint = didn't do anything?), but I think that'd be super fragile in practice.

Anyway, the whole partial cancellation topic has been coming up a lot, so I opened #889 to consolidate the threads...

-----

[...]

PR #941, opened 2019-02-19
WIP: Add graceful cancel to cancel scopes

nmalaguti:
See https://trio.discourse.group/t/graceful-shutdown/93/19?u=nmalaguti for the intended behavior.

No tests yet, just some code to illustrate intended behavior.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -168,11 +168,13 @@ class CancelScope:
     _has_been_entered = attr.ib(default=False, init=False)
     _effective_deadline = attr.ib(default=inf, init=False)
     cancel_called = attr.ib(default=False, init=False)
+    graceful_cancel_called = attr.ib(default=False, init=False)
     cancelled_caught = attr.ib(default=False, init=False)
 
     # Constructor arguments:
     _deadline = attr.ib(default=inf, kw_only=True)
     _shield = attr.ib(default=False, kw_only=True)
+    _graceful = attr.ib(default=None, kw_only=True)
 
     @enable_ki_protection
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


trio/_core/_run.py
```diff
@@ -366,8 +428,7 @@ def _exc_filter(self, exc):
     def _close(self, exc):
         scope_task = current_task()
         if (
-            exc is not None and self.cancel_called
```
nmalaguti: It looks like `self.cancel_called` is in this if statement to ensure that raising `Cancelled` won't be caught here unless `scope.cancel()` was called. Looking at the logic in `_pending_cancel_scope` it looks like you can't return a scope unless that is already true, so this check is redundant.

I have removed it because there are now cases where you can be cancelled gracefully without having been explicitly cancelled.

nmalaguti:
@oremanj I was surprised at how simple this was to implement, and it turns out it's because you've been doing a ton of work in the cancellation space.

Seeing how this relates to #886, #885, #910, and #147, I feel like I didn't do enough research before embarking on this change.

Also I just noticed you have an alternate approach in #921. I'd love to discuss further the differences between them.

trio/_core/_run.py
```diff
@@ -325,6 +327,19 @@ def shield(self, new_value):
...
                 task._attempt_delivery_of_any_pending_cancel()
 
+    @property
+    def graceful(self):
+        """If this is set to :data:`True`, then the code inside this
+        scope will receive :exc:`~trio.Cancelled` exceptions from scopes
+        that are outside this scope that have been gracefully canceled.
+        If this is set to :data:`None` then this scope will inherit its
+        graceful cancel behavior from its parent scope.
+
```
Tronic: This wording is hard to understand and should be clarified.

Issue #917, opened 2019-02-08
trio.Path inconsistencies


oremanj:
Noticed these doing typing:
* glob() and rglob() probably want the same treatment as iterdir(); currently they return an iterable of unwrapped pathlib.Path objects
* trio.Path.cwd() is synchronous - is that intended? On Linux it doesn't need to hit the disk (it has a dedicated syscall, or can use `readlink /proc/self/cwd`), but I think on some other Unixes it can...

njsmith:
Marking "good first issue" for the `glob` and `rglob` fixes. The issue is: `trio.Path.glob` returns an iterable of `pathlib.Path` objects. We should convert them into `trio.Path` objects instead. And the same for `trio.Path.rglob`. `trio.Path.iterdir` has an example of how to do this.

Regarding the `cwd` thing... well, does anyone know what macOS does? Does anyone with a Mac want to run a syscall tracer to find out?

It [looks like Windows is fine](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-getcurrentdirectory):

> The current directory state written by the `SetCurrentDirectory` function is stored as a global variable in each process, therefore multithreaded applications cannot reliably use this value without possible data corruption from other threads that may also be reading or setting this value.

pquentin:
> Regarding the `cwd` thing... well, does anyone know what macOS does? Does anyone with a Mac want to run a syscall tracer to find out?

I did not manage to get this working with Python: my syscall tracer (dtruss) stops at `posix_spawn()`. However, it looks like CPython just calls `getcwd()`. Here's what `getcwd()` does on macOS 10.14 inside "/private/tmp".

```
65059/0x42b8a:  open_nocancel(".\0", 0x0, 0x0)           = 3 0
65059/0x42b8a:  fstat64(0x3, 0x7FFEE819AB10, 0x0)                = 0 0
65059/0x42b8a:  fcntl_nocancel(0x3, 0x32, 0x7F8FC9800000)                = 0 0
65059/0x42b8a:  close_nocancel(0x3)              = 0 0
65059/0x42b8a:  stat64("/private/tmp\0", 0x7FFEE819AA80, 0x0)            = 0 0
```

[This Quora answer](https://www.quora.com/How-is-getcwd-implemented-on-Mac) mentions an existing implementation. I think the current one is https://opensource.apple.com/source/Libc/Libc-1272.200.26/gen/FreeBSD/getcwd.c.auto.html.

Is this helpful?

[...]

PR #967, opened 2019-03-05
Generate async classmethods for pathlib.Path

### LINKED ISSUE #960: trio.Path.home raises "AttributeError: _flavour"
I get the following on Mac and Linux:
```
>>> import trio
>>> trio.Path.home()
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/3.7.2_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pathlib.py", line 1055, in home
    return cls(cls()._flavour.gethomedir(None))
  File "/Users/zach/.allotropic/python/default/lib/python3.7/site-packages/trio/_path.py", line 157, in __getattr__
    raise AttributeError(name)
AttributeError: _flavour
```
As the traceback states, ``pathlib.Path`` is instantiating itself to dertermine its ``_flavour``, which is a class to delegate filesystem calls to differing Windows and Posix environments.

I think the explanation is that trio.Path does not override ``__new__``, so it doesn't get a chance to turn into a ``PosixPath`` or ``WindowsPath`` object like ``pathlib.Path`` objects do.  The ``_flavour`` attribute is only available from these system-specific classes, and the ``AsyncAutoWrapperType`` doesn't explicitly wrap these classes to make it an attribute of ``trio.Path``.

One workaround is to override ``home`` in ``trio.Path``:
```
    @classmethod
    def home(cls):
        return cls(cls()._wraps()._flavour.gethomedir(None))
```

There are other approaches and I'd like to check in with the team here.  Any thoughts?

zthompson47:
Fixes #960 and applies to #917.

### DIFF
--- a/newsfragments/960.bugfix.rst
+++ b/newsfragments/960.bugfix.rst
@@ -1,0 +1,2 @@
+As a result of fixing a bug in ``trio.Path.home``, the two ``trio.Path``
+classmethods, ``home`` and ``cwd``, are now async functions (see #917).
--- a/newsfragments/960.bugfix.rst
+++ b/newsfragments/960.bugfix.rst
@@ -1,2 +1,4 @@
-As a result of fixing a bug in ``trio.Path.home``, the two ``trio.Path``
-classmethods, ``home`` and ``cwd``, are now async functions (see #917).
+The :class:`trio.Path` classmethods, :meth:`~trio.Path.home` and
+:meth:`~trio.Path.cwd`, are now async functions.  Previously, a bug
+in the forwarding logic meant :meth:`~trio.Path.cwd` was synchronous
+and :meth:`~trio.Path.home` didn't work at all.
--- a/trio/_path.py
+++ b/trio/_path.py
@@ -54,7 +54,7 @@ def _forward_magic(cls, attr):
 
 
 def thread_wrapper_factory(cls, meth_name):
-    @async_wraps(cls, pathlib.Path, meth_name)
+    @async_wraps(cls, cls._wraps, meth_name)
     async def wrapper(self, *args, **kwargs):
         args = unwrap_paths(args)
         meth = getattr(self._wrapped, meth_name)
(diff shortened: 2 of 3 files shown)


trio/_path.py
```diff
@@ -65,6 +65,19 @@ def thread_wrapper_factory(cls, meth_name):
     return wrapper
 
 
+def classmethod_wrapper_factory(cls, meth_name):
+    @classmethod
+    @async_wraps(cls, cls._wraps, meth_name)
```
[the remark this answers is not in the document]
zthompson47: Using the generic ``cls._wraps`` in both also seems to match the generic nature of the other functions in this module, so I'll write it that way in the next commit.

zthompson47:
Thanks for the code reviews!  It's a fun project to work on.

I added this newsfragment as a 'bugfix' even though it made some breaking changes by going to async classmethods.  It seems an appropriate way to alert users, but I can change that as needed.

The sphinx documentation looks fine - it's grabbing the 'Like func(), but async' docstrings, and the links to ``pathlib.Path`` docstrings work.

newsfragments/960.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+As a result of fixing a bug in ``trio.Path.home``, the two ``trio.Path``
```
oremanj: Use Sphinx markup to link to the corresponding section of the documentation: that is, say
```
[...] a bug in :meth:`trio.Path.home`, the two :class:`trio.Path` classmethods,
:meth:`~trio.Path.home` and :meth:`~trio.Path.cwd`, [...]
```
I think this might also read better as two sentences, something like: "The trio.Path classmethods, home and cwd, are now async functions. Previously, a bug in the forwarding logic meant cwd was synchronous and home didn't work at all.

zthompson47: Nice suggestions.  Running ``towncrier --draft`` helped me see how things would look in the docs.

oremanj:
Meep, sorry for dropping this! Looks good, thanks again :-)

zthompson47:
> Meep, sorry for dropping this! Looks good, thanks again :-)

No worries.  It's exciting to see my first open source code contributions out there and I appreciate the help!

Issue #970, opened 2019-03-07
Document that Nursery.start_soon starts tasks in random order


rotu:
When calling `Nursery.start_soon`, submitted tasks are called in an apparently non-deterministic order. This is not intuitive (at least to me); since `nursery.start_soon` is synchronous I expected the functions to be started in FIFO order. If the order is non-deterministic on purpose, this should be documented.

Toy example:
```
>>> import trio
>>>
>>> print_lock = trio.StrictFIFOLock()
>>>
>>>
>>> async def print_soon(x):
...     async with print_lock:
...         print(x)
...
>>>
>>> async def amain():
...     for i in range(3):
...         print('-----')
...         async with trio.open_nursery() as nursery:
...             for j in range(3):
...                 nursery.start_soon(print_soon, j)
...
>>>
>>> trio.run(amain)
-----
0
2
1
-----
1
0
2
-----
2
1
0
```

njsmith:
Yes, this is intentional. They are started in FIFO order, in the sense that first task 1 is created and marked as runnable, then task 2, then task 3. And then the next time the parent task goes to sleep, the scheduler picks which task to run next. At this point those 3 tasks are all runnable – and in a more complex program, there might be arbitrary other tasks that were runnable as well – and the scheduler picks between them. And in general, Trio's scheduler doesn't make any guarantees about when it will run different tasks; all it guarantees is that all tasks will get a fair chance to run.

In practice, this is about the strongest guarantee you can usefully get from a scheduler. If you run tasks concurrently, then they run at different rates :-). In some cases you might be able to manually line up their execution in some consistent way for a while, but it's going to be quite fragile. You might also find #32 and #239 interesting.

If you want to make sure that task 1 has had a chance to finish starting up before starting task 2, then see [`nursery.start`](https://trio.readthedocs.io/en/latest/reference-core.html#trio.The%20nursery%20interface.start). (There's also [an example here](https://pytest-trio.readthedocs.io/en/latest/quickstart.html#running-a-background-server-from-a-fixture).)

Where would you have expected to see this in the docs?

rotu:
That makes sense. My naive expectation was that, because the docs emphasized that control gets nondeterministic at checkpoints, tasks would run in order *up until* the first checkpoint, at which point things would get wibbly-wobbly.

[...]

Issue #32, opened 2017-01-23
Design: alternative scheduling models


njsmith:
Currently we use a simple round-robin scheduling strategy. It isn't actually FIFO, because we randomize the order of execution within each "batch" in an attempt to force people not to make too many assumptions about the scheduling strategy :-).

Can we do better?

There's a rich literature on fancy scheduling, e.g. the weighted-fair queuing used by the Linux kernel's CFS ("completely fair scheduler"). And there are a lot of challenges in building async applications that come down to scheduling issues (e.g., #14 and https://stackoverflow.com/questions/40601119). AFAICT a lot of the existing literature assumes that have a pre-emptive scheduler; in some ways the network packet scheduling literature is more relevant to us, because packets come in different sizes and the scheduler doesn't get to change that. But there's very little out there on applying these ideas to cooperative scheduling systems.

In principle, a WFQ scheduler is actually very simple to implement: 

What's not so clear is whether this would actually help in practice. There are two obvious issues:

* If some task hogs the CPU for 100 ms, then there is no scheduling policy that can possibly avoid this causing a 100 ms spike for everyone else. You can get amortized fairness in the long run by exiling that task from the CPU for a comparatively long time afterwards, but the end result is still going to be jittery and spiky. Conclusion: nothing we do at the scheduler level is going to have a great effect unless user code is written to yield with a good frequency. I guess the solution here is to make sure we have great tools for profiling user code and warning of CPU hogs. The instrumentation API is a good start.

* In a cooperative scheduler, tasks run until they block and become non-runnable. In pure WFQ, the scheduler *only* cares about runnable tasks; a task that blocks effectively disappears, and when it becomes runnable again it's treat like a new task with zero history. This is okay in a pre-emptive scheduler where tasks don't have a chance to "overdraw" CPU time, but doesn't make much sense for us. For us the way you treat a task that wakes from sleep is the whole question.

And it's not entirely clear that "fairness" is actually the solution to the problems linked above, either!  Maybe what we really need is better knobs to let users set the priority of different tasks. (WFQ makes it easy to implement relative priorities, and relatively easy to implement hierarchical scheduling policies; it's also pretty straightforward to implement strictly tiered priorities like the Linux kernel's realtime priorities.) But "here's 100 knobs to tune, try tweaking some whenever your server performance degrades" is a pretty bad UX too. There's something to be said for the predictability of FIFO.

In general, this is a deep problem that will definitely require experience and data and visualizations of real apps on top of trio.

# Prior art

[...]

PR #972, opened 2019-03-07
clarified Nursery.start vs Nursery.start_soon

rotu:
#970

### DIFF
--- a/docs/source/history.rst
+++ b/docs/source/history.rst
@@ -5,42 +5,6 @@ Release history
 
 .. towncrier release notes start
 
-Trio 0.11.0+dev (2019-03-07)
-----------------------------
-
-Features
-~~~~~~~~
-
-- On Linux, when wrapping a bare file descriptor in a Trio socket object,
-  Trio now auto-detects the correct ``family``, ``type``, and ``protocol``.
-  This is useful, for example, when implementing `systemd socket activation
-  <http://0pointer.de/blog/projects/socket-activation.html>`__. (`#251 <https://github.com/python-trio/trio/issues/251>`__)
-
-
-Improved Documentation
-~~~~~~~~~~~~~~~~~~~~~~
-
-- Clarified documentation around Nursery.start_soon and Nursery.start, that start_soon does not guarantee task ordering, and when you should use each. (`#970 <https://github.com/python-trio/trio/issues/970>`__)
-
-
-Trio 0.11.0+dev (2019-03-07)
-----------------------------
-
-No significant changes.
-
-
-Trio 0.11.0+dev (2019-03-07)
-----------------------------
-
-Features
-~~~~~~~~
-
-- On Linux, when wrapping a bare file descriptor in a Trio socket object,
-  Trio now auto-detects the correct ``family``, ``type``, and ``protocol``.
-  This is useful, for example, when implementing `systemd socket activation
-  <http://0pointer.de/blog/projects/socket-activation.html>`__. (`#251 <https://github.com/python-trio/trio/issues/251>`__)
-
-
 Trio 0.11.0 (2019-02-09)
 ------------------------
 
--- a/docs/source/history.rst
+++ b/docs/source/history.rst
@@ -5,6 +5,42 @@ Release history
 
 .. towncrier release notes start
 
+Trio 0.11.0+dev (2019-03-07)
+----------------------------
+
+Features
+~~~~~~~~
+
+- On Linux, when wrapping a bare file descriptor in a Trio socket object,
+  Trio now auto-detects the correct ``family``, ``type``, and ``protocol``.
+  This is useful, for example, when implementing `systemd socket activation
+  <http://0pointer.de/blog/projects/socket-activation.html>`__. (`#251 <https://github.com/python-trio/trio/issues/251>`__)
+
+
+Improved Documentation
+~~~~~~~~~~~~~~~~~~~~~~
+
+- Clarified documentation around Nursery.start_soon and Nursery.start, that start_soon does not guarantee task ordering, and when you should use each. (`#970 <https://github.com/python-trio/trio/issues/970>`__)
+
+
+Trio 0.11.0+dev (2019-03-07)
+----------------------------
+
+No significant changes.
+
+
+Trio 0.11.0+dev (2019-03-07)
+----------------------------
+
+Features
+~~~~~~~~
+
+- On Linux, when wrapping a bare file descriptor in a Trio socket object,
+  Trio now auto-detects the correct ``family``, ``type``, and ``protocol``.
+  This is useful, for example, when implementing `systemd socket activation
+  <http://0pointer.de/blog/projects/socket-activation.html>`__. (`#251 <https://github.com/python-trio/trio/issues/251>`__)
+
+
 Trio 0.11.0 (2019-02-09)
 ------------------------
 
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -834,11 +834,11 @@ Nursery objects provide the following interface:
 
       Note that this is *not* an async function and you don't use await
       when calling it. It sets up the new task, but then returns
-      immediately, *before* it has a chance to run. At some future
-      time, (which might be closing of the nursery context, but no
-      later), all tasks requested with :meth:`start_soon` will begin,
-      in no particular order. If you want to run a function to
-      completion, then you don't need a nursery; just use
+      immediately, *before* it has a chance to run. All tasks requested
+      with :meth:`start_soon` will begin, at no particular time, in no
+      particular order (but no later than the closing of the nursery
+      context). If you want to run a function and block until it
+      finishes, then you don't need a nursery; just use
       ``await async_fn(*args)``. If you want to run some part of the
       child task before the parent task proceeds, see :meth:`start()`.
 
@@ -868,8 +868,8 @@ Nursery objects provide the following interface:
       :async:
 
       Like :meth:`start_soon`, but blocks until the new child task
-      signals that it may unblock, optionally returning additional
-      information.
+      indicates it may unblock, and optionally returns something back
+      to the parent.
 
       The ``async_fn`` must accept a ``task_status`` keyword argument,
       and it must make sure that it (or someone) eventually calls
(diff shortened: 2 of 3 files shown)


docs/source/history.rst
```diff
@@ -5,6 +5,42 @@ Release history
 
 .. towncrier release notes start
 
+Trio 0.11.0+dev (2019-03-07)
```
oremanj: The towncrier generated release notes shouldn't get checked in until we actually tag the release; the only change you would make is adding the new newsfragment file.

docs/source/reference-core.rst
```diff
@@ -834,13 +834,14 @@ Nursery objects provide the following interface:
...
-      If you want to run a function and immediately wait for its result,
+      immediately, *before* it has a chance to run. At the next
+      checkpoint, (which might be closing of the nursery context), all
+      tasks requested with :meth:`start_soon` will begin in no
+      particular order. If you want to run a function to completion,
       then you don't need a nursery; just use ``await async_fn(*args)``.
-      If you want to wait for the task to initialize itself before 
-      continuing, see :meth:`start()`.
+      If you want to ensure that the child task does something before
+      before the parent moves on to the next line, or that multiple
```
oremanj: duplicate "before"

docs/source/reference-core.rst
```diff
@@ -834,13 +834,14 @@ Nursery objects provide the following interface:
...
-      you execute a checkpoint and the scheduler decides to run it.
-      If you want to run a function and immediately wait for its result,
+      immediately, *before* it has a chance to run. At the next
+      checkpoint, (which might be closing of the nursery context), all
+      tasks requested with :meth:`start_soon` will begin in no
+      particular order. If you want to run a function to completion,
       then you don't need a nursery; just use ``await async_fn(*args)``.
-      If you want to wait for the task to initialize itself before 
-      continuing, see :meth:`start()`.
+      If you want to ensure that the child task does something before
```
oremanj: What do you find to be clear from "ensure that the child task does something before the parent moves onto the next line" that isn't clear from "wait for the task to initialize itself before continuing"? I don't really see a substantive difference, and tend to prefer the shorter/punchier version, but I might be missing something. (I think being clear about "the *child* task" and maybe "before continuing *execution of the parent*" are worthwhile changes, but I'm sketchy on the more complex grammatical structure.)

docs/source/reference-core.rst
```diff
@@ -834,13 +834,14 @@ Nursery objects provide the following interface:
 
       Note that this is *not* an async function and you don't use await
       when calling it. It sets up the new task, but then returns
-      immediately, *before* it has a chance to run. The new task won’t
-      actually get a chance to do anything until some later point when
-      you execute a checkpoint and the scheduler decides to run it.
-      If you want to run a function and immediately wait for its result,
+      immediately, *before* it has a chance to run. At the next
```
oremanj: I think this is a more stringent guarantee than we actually want to make -- #32 has some discussion of scheduling models that might want us to not switch tasks at every checkpoint.

newsfragments/970.doc.rst
```diff
@@ -0,0 +1 @@
+Clarified documentation around Nursery.start_soon and Nursery.start, that start_soon does not guarantee task ordering, and when you should use each.
```
oremanj: @njsmith - do we usually add newsfragments for documentation updates?

docs/source/reference-core.rst
```diff
@@ -878,9 +881,9 @@ Nursery objects provide the following interface:
       The conventional way to define ``async_fn`` is like::
 
          async def async_fn(arg1, arg2, *, task_status=trio.TASK_STATUS_IGNORED):
-             ...
+             ... # some work before the parent may proceed
```
oremanj: A task with `task_status=trio.TASK_STATUS_IGNORED` can also be started with `start_soon()`, in which case the parts before and after `task_status.started()` behave the same as each other. Is there a different way to word this comment that might be clearer about that?

docs/source/history.rst
rotu: Reverted

docs/source/reference-core.rst
rotu: Rewrote to not guarantee that tasks will launch at the next checkpoint.

docs/source/reference-core.rst
rotu: Initialize sounds technical, and implies to me that there's some sort of __init__ function being called, or that there's something other than a function being called. Rewrote to the punchier "If you want to run some part of the child task before the parent task proceeds, see :meth:`start()`."

docs/source/reference-core.rst
rotu: reworded to "some work the parent might wait for"

rotu:
Thank you, @oremanj for your thorough and patient feedback. I'm not used to contributing to open source, and I appreciate your help greatly. Aside from the newsfragment note (which I'd be happy to go either way on) I think I addressed all your concerns.

docs/source/reference-core.rst
```diff
@@ -867,9 +867,9 @@ Nursery objects provide the following interface:
    .. method:: start(async_fn, *args, name=None)
       :async:
 
-      Like :meth:`start_soon`, but blocks until the new task has
-      finished initializing itself, and optionally returns some
-      information from it.
+      Like :meth:`start_soon`, but blocks until the new child task
+      signals that it may unblock, optionally returning additional
```
oremanj: "signals" is a slightly overloaded word to use here due to Unix signals. Maybe "until the new child task indicates that it's finished starting up"? And I think the original "and optionally returns some information" is better than "optionally returning additional information" -- the more active phrasing tends to communicate more clearly IMO.

docs/source/reference-core.rst
```diff
@@ -834,13 +834,13 @@ Nursery objects provide the following interface:
...
-      If you want to run a function and immediately wait for its result,
-      then you don't need a nursery; just use ``await async_fn(*args)``.
-      If you want to wait for the task to initialize itself before 
-      continuing, see :meth:`start()`.
+      immediately, *before* it has a chance to run. At some future
+      time, (which might be closing of the nursery context, but no
+      later), all tasks requested with :meth:`start_soon` will begin,
+      in no particular order. If you want to run a function to
+      completion, then you don't need a nursery; just use
+      ``await async_fn(*args)``. If you want to run some part of the
```
oremanj: maybe "If you want to wait for the child task to finish starting up before the parent is allowed to proceed"

docs/source/reference-core.rst
```diff
@@ -834,13 +834,13 @@ Nursery objects provide the following interface:
...
-      you execute a checkpoint and the scheduler decides to run it.
-      If you want to run a function and immediately wait for its result,
-      then you don't need a nursery; just use ``await async_fn(*args)``.
-      If you want to wait for the task to initialize itself before 
-      continuing, see :meth:`start()`.
+      immediately, *before* it has a chance to run. At some future
+      time, (which might be closing of the nursery context, but no
+      later), all tasks requested with :meth:`start_soon` will begin,
+      in no particular order. If you want to run a function to
+      completion, then you don't need a nursery; just use
```
oremanj: The function always runs "to completion" (as opposed to being abandoned in the middle somewhere) regardless of whether you use `await` or `nursery.start_soon`. What are you trying to communicate here that you didn't feel was adequately communicated with "and immediately wait for its result"?

docs/source/reference-core.rst
```diff
@@ -834,13 +834,13 @@ Nursery objects provide the following interface:
...
-      immediately, *before* it has a chance to run. The new task won’t
-      actually get a chance to do anything until some later point when
-      you execute a checkpoint and the scheduler decides to run it.
-      If you want to run a function and immediately wait for its result,
-      then you don't need a nursery; just use ``await async_fn(*args)``.
-      If you want to wait for the task to initialize itself before 
-      continuing, see :meth:`start()`.
+      immediately, *before* it has a chance to run. At some future
+      time, (which might be closing of the nursery context, but no
+      later), all tasks requested with :meth:`start_soon` will begin,
```
oremanj: This still could be read to imply that all of the tasks will be started at once, which isn't a guarantee we want to make. What if anything do you think is unclear about taking the previous wording ("won't actually get a chance to do anything ... the scheduler decides to run it") and adding a sentence like "Tasks might start running in a different order than the calls to `start_soon()`"?

docs/source/reference-core.rst
rotu: No, because "finish starting up" is not a meaningful concept until you've already bought into the abstraction of Nursery.start(). A task scheduled with start_soon might never call a task_status.started() function - but any programmer would say that task starts and ends and infers that it "finishes starting" somewhere in between. There needs to be some concept of the child signaling to the parent, even if we don't use the word "signal".

docs/source/reference-core.rst
rotu: changing to "indicates"

docs/source/reference-core.rst
rotu: The start function runs a function and immediately waits for a result too. It just so happens that the result might be returned by an intermediate result from a .started call in the middle.. I wanted to convey the difference. How about "If you want to run a function and block until it finishes" ?

docs/source/reference-core.rst
rotu: You're right - it does imply a single time. Rephrased to "All tasks requested with :meth:`start_soon` will begin, at no particular time, in no particular order (but no later than the closing of the nursery context)."
I wanted to avoid mention of the scheduler here. The important thing is that ordering and timing get arbitrary when you call start_soon - I don't think the reader will get that unless they have a good grasp of what the scheduler does and does not promise.

rotu:
> sorry for being so nitpicky about this, but it's a pretty core aspect of the documentation

No problem! This documentation cost me personally hours of frustration and confusion. I'm more than happy to hammer it to death.

newsfragments/970.doc.rst
rotu: @oremanj @njsmith Any opinions? I don't care whether we leave it in or take it out, but I'd like this pull request resolved

rotu:
What's left to fix for this PR?

pquentin:
@rotu I'm sorry that you had a bad experience here despite a very promising start. I personally don't feel qualified enough to look at this, but hopefully @oremanj will be able to look at this again at some point.

PR #984, opened 2019-03-21
Use slots on whenever possible

thedrow:
Since we don't dynamically add any attributes to that class and attrs allow us to generate slots automatically, we should use them whenever possible.

### DIFF
--- a/newsfragments/984.feature.rst
+++ b/newsfragments/984.feature.rst
@@ -1,1 +1,1 @@
-Use __slots__ in more places internally, which should make Trio slightly faster.
+Use ``__slots__`` in more places internally, which should make Trio slightly faster.
--- a/newsfragments/984.feature.rst
+++ b/newsfragments/984.feature.rst
@@ -1,0 +1,1 @@
+Use __slots__ in more places internally, which should make Trio slightly faster.
--- a/trio/_core/_entry_queue.py
+++ b/trio/_core/_entry_queue.py
@@ -9,7 +9,7 @@ from ._wakeup_socketpair import WakeupSocketpair
 __all__ = ["TrioToken"]
 
 
-@attr.s
+@attr.s(slots=True)
 class EntryQueue:
     # This used to use a queue.Queue. but that was broken, because Queues are
     # implemented in Python, and not reentrant -- so it was thread-safe, but
--- a/trio/_core/_entry_queue.py
+++ b/trio/_core/_entry_queue.py
@@ -142,6 +142,8 @@ class TrioToken:
 
     """
 
+    __slots__ = ('_reentry_queue',)
+
     def __init__(self, reentry_queue):
         self._reentry_queue = reentry_queue
 
--- a/trio/_core/_io_epoll.py
+++ b/trio/_core/_io_epoll.py
@@ -6,14 +6,14 @@ from .. import _core
 from . import _public
 
 
-@attr.s(slots=True, cmp=False, hash=False, frozen=True)
+@attr.s(slots=True, cmp=False, frozen=True)
 class _EpollStatistics:
     tasks_waiting_read = attr.ib()
     tasks_waiting_write = attr.ib()
     backend = attr.ib(default="epoll")
 
 
-@attr.s(slots=True, cmp=False, hash=False)
+@attr.s(slots=True, cmp=False)
 class EpollWaiters:
     read_task = attr.ib(default=None)
     write_task = attr.ib(default=None)
--- a/trio/_core/_io_epoll.py
+++ b/trio/_core/_io_epoll.py
@@ -6,7 +6,7 @@ from .. import _core
 from . import _public
 
 
-@attr.s(frozen=True)
+@attr.s(slots=True, cmp=False, hash=False, frozen=True)
 class _EpollStatistics:
     tasks_waiting_read = attr.ib()
     tasks_waiting_write = attr.ib()
--- a/trio/_core/_io_kqueue.py
+++ b/trio/_core/_io_kqueue.py
@@ -8,14 +8,14 @@ from .. import _core
 from . import _public
 
 
-@attr.s(slots=True, cmp=False, hash=False, frozen=True)
+@attr.s(slots=True, cmp=False, frozen=True)
 class _KqueueStatistics:
     tasks_waiting = attr.ib()
     monitors = attr.ib()
     backend = attr.ib(default="kqueue")
 
 
-@attr.s(slots=True, cmp=False, hash=False)
+@attr.s(slots=True, cmp=False)
 class KqueueIOManager:
     _kqueue = attr.ib(factory=select.kqueue)
     # {(ident, filter): Task or UnboundedQueue}
--- a/trio/_core/_io_kqueue.py
+++ b/trio/_core/_io_kqueue.py
@@ -8,7 +8,7 @@ from .. import _core
 from . import _public
 
 
-@attr.s(frozen=True)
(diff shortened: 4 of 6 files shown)


njsmith:
My impression was that slots made attribute access a little bit slower, in return for reduced memory usage, so they only make sense when you have a lot of instances of a type. The PR description here suggests a different theory, that slots are always a benefit. Which one is right? :-) Can you elaborate on what benefits you think we'll get from this?

thedrow:
From my experience using `__slots__` is a good practice on libraries which are going to be used in very common large scale projects. What seems to be used rarely may be used extensively in large enough projects.
Access [should be](https://stackoverflow.com/a/28059785/85140) faster.
Avoiding [named method dispatch](http://thomas-cokelaer.info/tutorials/python/slots.html#more-about-sots) is also a good reason to use `__slots__`:

> Slots are very useful for library calls to eliminate the “named method dispatch” when making function calls. This is mentioned in the SWIG documentation. For high performance libraries that want to reduce function overhead for commonly called functions using slots is much faster.

If you expect your class to be inherited `__slots__` make it slightly more difficult to do so. If you expect to add more attributes dynamically, `__slots__` make this impossible.

If you have benchmarks in place, we could measure.

njsmith:
> named method dispatch

I wasn't sure what that this meant (and unfortunately the link didn't help much). After thinking about it a bit, though, I realized it might mean: normally when you do a method call like `a.f()`, then Python has to first check whether there's an attribute `f` on the object `a`, just in case you did something like:

```python
a = A()
a.f = lambda: "blah"
```

So this means checking for `a.__dict__["f"]`, and that has to fail before Python can go looking for a method on the type `A`. But, if `A` has `__slots__`, then you might be able to skip this, because if `f` isn't listed in `__slots__`, then you immediately know it can't be an attribute on the object, and can skip straight to looking for it on the class. And indeed, this seems to be the case – defining `__slots__` makes unrelated method calls faster!

```
In [2]: class A: 
   ...:     __slots__ = () 
   ...:     def f(self): 
   ...:         return 1                                                        

In [3]:                                                                         

In [3]: class B: 
   ...:     def f(self): 
   ...:         return 2                                                        

In [4]:                                                                         

In [4]: a = A()                                                                 

In [5]: b = B()                                                                 

In [6]: %timeit a.f()                                                           
90.3 ns ± 3.33 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: %timeit b.f()                                                           
129 ns ± 3.39 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

I'm a bit puzzled at how this works, because it seems like there should still be some overhead from checking whether `f` is present in `__slots__`. But even if I add tons and tons of entries to the `__slots__` table, then method calls remain super-fast:

```
In [12]: class C: 
    ...:     __slots__ = tuple(f"name{i}" for i in range(10000)) 
    ...:     def f(self): 
    ...:         return 1                                                       

In [13]: c = C()                                                                

In [15]: %timeit c.f()                                                          
87.8 ns ± 1.19 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

So... not 100% sure I trust this – maybe there's some caching or something going on that I don't know about? But defining `__slots__` certainly doesn't seem to make method calls slower...

> Access should be faster.

Looks like you're right. The effect is very small, but it goes the right direction!

```python
COUNT = 1000

class A: 
    __slots__ = tuple(f"name{i}" for i in range(COUNT))

class B:
    pass

a = A()
for i in range(COUNT):
    setattr(a, f"name{i}", i)

b = B()
for i in range(COUNT):
    setattr(b, f"name{i}", i)
```

```
# Reads
In [48]: %timeit a.name500                                                      
36.1 ns ± 1.42 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [49]: %timeit b.name500                                                      
41.7 ns ± 2.25 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

# Writes
In [50]: %timeit a.name500 = 10                                                 
42.3 ns ± 1.92 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [51]: %timeit b.name500 = 10                                                 
47.6 ns ± 1.54 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

I'm not sure why I thought it went the other way.

I guess it can slow down object creation:

```
In [52]: %timeit A()                                                            
1.49 µs ± 32.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [53]: %timeit B()                                                            
99.2 ns ± 2.94 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

But this is related to the number of slots... if I redo that with `__slots__ = ()`, then `A()` speeds up to 75 ns.

So I guess `__slots__` are just a win all around for us, so this PR is a good idea!

njsmith:
The patch looks fine. There are lots of classes that aren't included, but that's not necessarily a problem... this is still an improvement even if there are other improvements possible too.

Could you add a [newsfragment](https://github.com/python-trio/trio/blob/master/newsfragments/README.rst)? Something like `984.feature.rst`, "Use `__slots__` in more places internally, which should make Trio slightly faster."

thedrow:
Yes sure.

oremanj:
> I'm a bit puzzled at how this works, because it seems like there should still be some overhead from checking whether f is present in __slots__. But even if I add tons and tons of entries to the __slots__ table, then method calls remain super-fast:

I think the missing piece here is that if you define `__slots__`, then your instances don't have a `__dict__` at all, so there's no need to do a lookup there first. The slot descriptors go into the _class_ dictionary, which is the same place the methods are, so even with a lot of slots there's still only one dict lookup needed.

> I guess it can slow down object creation:

I think this is related to the fact that all slots need to be set to null when the object is created, while entries in the instance dict can simply not-exist.

njsmith:
> The slot descriptors go into the class dictionary

ah-hah, right that's what I was forgetting :-). I think the standard lookup path is actually:

1. check the class dict for an object that has a `__set__` method, and if found call its `__get__` method
2. otherwise, check the instance dict
3. otherwise, check the class dict for an object
  a. If it has a `__get__` method, call and return that
  b. otherwise, return the object unchanged
4. `__getattr__`

Normal instance variables are found at step 2. Slotted instance variables are found in step 1 and their descriptor object has a really fast `__get__` method, so they're faster, regardless of how many slots you have. Methods are found in step 3a (the `__get__` method is where the bound method object is created), but if the class is slotted than the lookup gets to skip step 2, so you can get to 3a faster.

So it does all make sense.

ziirish:
Hi,

Do you need some help to complete this PR?

thedrow:
I just haven't gotten a chance to write the news fragment.
Feel free to add it if I don't beat you to it.

njsmith:
FYI, in modern attrs, there's never any need to set `hash=`. Just `cmp=False` is enough on its own.

newsfragments/984.feature.rst
```diff
@@ -0,0 +1 @@
+Use __slots__ in more places internally, which should make Trio slightly faster.
```
njsmith: Since this is restructured text, I think you need to add double-backticks around `__slots__`, like:

```rst
Use ``__slots__`` in more places internally
```

thedrow:
I chose whatever seems the most elementary and used in trio as the first targets.

njsmith:
Thanks! And like [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team) says, you should have a Github invite :-).

(and @ziirish thanks for merging :-))

Issue #1009, opened 2019-04-17
return statement in a running nursery doesn't stop the function


math2001:
Firstly, thank you so much for trio, it's absolutely awesome ❤️ 

A return statement in a nursery doesn't exit the function straight away, it still waits for the nursery to be done. So for example:

```python
async def main():
    async with trio.open_nursery() as nursery:
        nursery.start_soon(trio.sleep, 5)
        await trio.sleep(1)
        print("returning")
        return # expected to cancel the nursery, but doesn't

    print("I'm never printed") # that's normal
```

This code would exit after 5 seconds, not 1 second. To make it stop at 1 second, we have to cancel it manually (which I thought was going to be done by trio):

```python
return nursery.cancel_scope.cancel()
```

So, I'm not sure if this is known, or if it should be changed, or if I'm missing something but I think it should be part of the documentation somewhere in the [nursery section](https://trio.readthedocs.io/en/latest/reference-core.html#tasks-let-you-do-multiple-things-at-once).

njsmith:
Yeah, that's working correctly (though maybe not intuitively!). The nursery is only cancelled if (a) there's an exception, or (b) if you use `nursery.cancel_scope` to cancel it explicitly.

We don't really have a choice in the matter, because the way Python works, there's no way for Trio to tell the difference between:

```python
async with trio.open_nursery() as nursery:
    nursery.start_soon(...)
    return
```

and 

```python
async with trio.open_nursery() as nursery:
    nursery.start_soon(...)
return
```

So I guess the question is how to make this clearer in the docs...

PR #1012, opened 2019-04-18
Add a warning about return statements in nurseries

math2001:
Regarding this issue: #1009

I'm not sure if it's too emphasized, though. It looks like this:

image

And I couldn't get reStructuredText to do a code block *inside* a warning using just indentation (like it's done in the rest of the docs). I had to add the prefix `.. code-block:: python`.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -651,19 +651,18 @@ Since all tasks are descendents of the initial task, one consequence
 of this is that :func:`run` can't finish until all tasks have
 finished.
 
-.. warning::
+.. note::
 
-  A return statement will not cancel the nursery if it still has tasks running.
+  A return statement will not cancel the nursery if it still has tasks running::
 
-  .. code-block:: python
-  
-      with trio.open_nursery() as nursery:
-          nursery.start_soon(trio.sleep(5))
-
-          await trio.sleep(1)
+    async def main():
+      async with trio.open_nursery() as nursery:
+          nursery.start_soon(trio.sleep, 5)
           return
 
-  This code will wait 5 seconds, NOT 1.
+    trio.run(main)
+
+  This code will wait 5 seconds (for the child task to finish), and then return.
 
 Child tasks and cancellation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


njsmith:
> I couldn't get reStructuredText to do a code block inside a warning using just indentation

It's not just indentation, it's a double-colon, like `For example::` ← notice the extra `:`. ReST is... quirky.

docs/source/reference-core.rst
```diff
@@ -656,6 +656,19 @@ Since all tasks are descendents of the initial task, one consequence
...
+
+  .. code-block:: python
+  
+      with trio.open_nursery() as nursery:
+          nursery.start_soon(trio.sleep(5))
+
+          await trio.sleep(1)
+          return
+
+  This code will wait 5 seconds, NOT 1.
```
njsmith: How about:

* Make it a `.. note::`

* It's `start_soon(trio.sleep, 5)`, not `start_soon(trio.sleep(5))`

* I think we can simplify the example to just:

  ```python
  with trio.open_nursery() as nursery:
      nursery.start_soon(trio.sleep, 5)
      return
  ```

  This will wait 5 seconds for the child task to finish, and *then* the enclosing function will `return`.

math2001:
> It's not just indentation, it's a double-colon, like For example:: ← notice the extra :. ReST is... quirky.

Oh... right.

I had to force push because I rebased the fix commit on top of the current master branch. I'm not sure that was needed though... Next time, I'll make sure not to send a pull request between `master`s.

math2001:
I also wrapped the whole thing in a function so the `return` would make sense. I'm wasn't sure if the `trio.run` should be there or not though...

njsmith:
Force-pushing to a PR is totally fine. But yeah, things generally go smoother if you use a dedicated branch instead of `master` :-)

Looks fine to me. I tweaked the indentation a little – first to line up the `.. note::` block contents with the `n` (I don't know if this matters, but that's how the rest of the docs do it, so it's more consistent this way), and second to use a consistent 4-space indent in the code block.

I think this is ready to go as soon as the CI finishes...

njsmith:
CI finished, so in it goes! :tada: :birthday: And, no pressure, but if you'd like to keep contributing then we'd love to have you, so I'm sending you a github invite now. You can read more about this in [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team).

PR #1018, opened 2019-04-23
CancelScopes are not meant to be reusable anymore (fix #1008)

### LINKED ISSUE #1008: Docs on CancelScope contradictory
In [https://trio.readthedocs.io/en/latest/reference-core.html](https://trio.readthedocs.io/en/latest/reference-core.html) the documentation on `CancelScope`s is contradictory. In the introduction it says:

> Cancel scopes are not reusable or reentrant; that is, each cancel scope can be used for at most one with block. (You’ll get a RuntimeError if you violate this rule.)

In the section on `cancelled_caught `it says:

> If the same CancelScope is reused for multiple with blocks, the cancelled_caught attribute applies to the most recent with block. (It is reset to False each time a new with block is entered.)

The first firmly states that the conditions described in the second are impossible.

ziirish:
As stated in #1008, The doc was not clear about the fact that CancelScopes are not meant to be reusable.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -514,11 +514,6 @@ objects.
       exception, and (2) this scope is the one that was responsible
       for triggering this :exc:`~trio.Cancelled` exception.
 
-      If the same :class:`CancelScope` is reused for multiple ``with``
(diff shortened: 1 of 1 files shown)


njsmith:
Thanks!

Issue #1005, opened 2019-04-15
TypeError: unhashable type: 'Argh'


joernheissler:
```python3
#!/usr/bin/env python3.7

from typing import Any
import trio

class Argh(Exception):
    code: int

    def __init__(self, code: int) -> None:
        super().__init__()
        self.code = code

    def __eq__(self, other: Any) -> bool:
        if not isinstance(other, Argh):
            return False

        return self.code == other.code

    def __repr__(self) -> str:
        return f'Argh(code={self.code})'

    def __str__(self) -> str:
        return repr(self)


async def die() -> None:
    raise Argh(844)


async def main() -> None:
    async with trio.open_nursery() as nursery:
        nursery.start_soon(die)


trio.run(main)
```

Results in:
```
Traceback (most recent call last):
  File "destroy.py", line 35, in <module>
    trio.run(main)
  File "/home/joern/.venv/trio/lib/python3.7/site-packages/trio/_core/_run.py", line 1444, in run
    raise runner.main_task_outcome.error
  File "destroy.py", line 32, in main
    nursery.start_soon(die)
  File "/home/joern/.venv/trio/lib/python3.7/site-packages/trio/_core/_run.py", line 496, in __aexit__
    combined_error_from_nursery = self._scope._close(new_exc)
  File "/home/joern/.venv/trio/lib/python3.7/site-packages/trio/_core/_run.py", line 372, in _close
    exc = MultiError.filter(self._exc_filter, exc)
  File "/home/joern/.venv/trio/lib/python3.7/site-packages/trio/_core/_multierror.py", line 209, in filter
    return _filter_impl(handler, root_exc)
  File "/home/joern/.venv/trio/lib/python3.7/site-packages/trio/_core/_multierror.py", line 111, in _filter_impl
    push_tb_down(None, root_exc, preserved)
  File "/home/joern/.venv/trio/lib/python3.7/site-packages/trio/_core/_multierror.py", line 104, in push_tb_down
    push_tb_down(new_tb, child_exc, preserved)
  File "/home/joern/.venv/trio/lib/python3.7/site-packages/trio/_core/_multierror.py", line 99, in push_tb_down
    if exc in preserved:
TypeError: unhashable type: 'Argh'
```

My Exception type is not hashable. In the real use case I get even uglier errors like `TrioInternalError`. Making the exception type hashable seems to resolve all issues.

But why does trio require hashable exceptions?

smurfix:
> My Exception type is not hashable.
[...]

PR #1023, opened 2019-04-25
Exceptions don't need to be hashable (see #767 and #1005)

ziirish:
This PR refers to #767 and #1005 in order to allow unhashable exceptions in `MultiError`s.

There is still a limitation though, you need to run python > 3.6.4 because `MultiError` relies on `traceback.TracebackException` which gained the support of unhashable Exceptions in 3.6.4.

### DIFF
--- a/trio/_core/tests/test_multierror.py
+++ b/trio/_core/tests/test_multierror.py
@@ -12,6 +12,24 @@ import warnings
 from .tutil import slow
 
 from .._multierror import MultiError, concat_tb
+from ..._core import open_nursery
+
+
+class NotHashableException(Exception):
+    code = None
+
+    def __init__(self, code):
+        super().__init__()
+        self.code = code
+
+    def __eq__(self, other):
+        if not isinstance(other, NotHashableException):
+            return False
+        return self.code == other.code
+
+
+async def raise_nothashable(code):
+    raise NotHashableException(code)
 
 
 def raiser1():
@@ -89,6 +107,16 @@ def test_MultiError():
         MultiError(object())
     with pytest.raises(TypeError):
         MultiError([KeyError(), ValueError])
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 6, 4), reason="Unsupported in python < 3.6.4"
+)
+async def test_MultiErrorNotHashable():
+    with pytest.raises(MultiError):
+        async with open_nursery() as nursery:
+            nursery.start_soon(raise_nothashable, 42)
+            nursery.start_soon(raise_nothashable, 4242)
 
 
 def make_tree():
--- a/trio/_core/tests/test_multierror.py
+++ b/trio/_core/tests/test_multierror.py
@@ -113,6 +113,12 @@ def test_MultiError():
     sys.version_info < (3, 6, 4), reason="Unsupported in python < 3.6.4"
 )
 async def test_MultiErrorNotHashable():
+    exc1 = NotHashableException(42)
+    exc2 = NotHashableException(4242)
+    exc3 = ValueError()
+    assert exc1 != exc2
+    assert exc1 != exc3
+
     with pytest.raises(MultiError):
         async with open_nursery() as nursery:
             nursery.start_soon(raise_nothashable, 42)
--- a/trio/_core/tests/test_multierror.py
+++ b/trio/_core/tests/test_multierror.py
@@ -67,6 +67,10 @@ def get_tb(raiser):
     return get_exc(raiser).__traceback__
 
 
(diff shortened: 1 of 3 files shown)


njsmith:
OK, I see what's going on now :-).

There are actually two unrelated bugs here... one is that `MultiError.filter` couldn't handle unhashable exceptions, and the other is the `traceback` module couldn't handle unhashable exceptions.

I think your fix for the first error is correct. But, this PR doesn't have any test for this :-). We should add one. I guess a test where we create a `MultiError` with an unhashable exception in it, and then try running `MultiError.filter` over it? (You'll know you got it right if the test fails without your patch :-).)

The fix for the second error is correct on Python 3.6.4+. But on earlier versions, it actually breaks things... on older Pythons, the `traceback` module is putting exception objects into the `seen` set. Before, we were doing `exc in seen`, which was at least correct for hashable exceptions. Now, we're doing `id(exc) in seen`, which is always wrong. So I think we need some kind of version-specific logic, so that we do `exc in seen` on older Python, and `id(exc) in seen` on newer Python.

The simplest way I can think of would be to do something like this at the top level:

```python
if sys.version_info < (3, 6, 4):
    exc_key = id
else:
    exc_key = lambda exc: exc
```

and then below do `exc_key(exc) in seen`.

And finally, the fact that this bug didn't trigger any test failures means that our test suite is missing something. I guess it would work to write a test like:

```python
exc1 = RuntimeError()
exc2 = KeyError()
# Note how this creates a loop, where exc1 refers to exc1
# This could trigger an infinite recursion; the 'seen' set is supposed to prevent this.
exc1.__cause__ = MultiError([exc1, exc2])
traceback.format_exception(exc1)
```

I'd start by writing a test like this, and making sure that with the current patch it passes on 3.6.4+ and fails on 3.5. Then I'd do the `exc_key` fix, and make sure that now the test passes on both versions.

ziirish:
> I think your fix for the first error is correct. But, this PR doesn't have any test for this :-). We should add one. I guess a test where we create a MultiError with an unhashable exception in it, and then try running MultiError.filter over it? (You'll know you got it right if the test fails without your patch :-).)

Although there is no explicit test for the `filter` part, I'm pretty sure the part I fixed are tested, because without the patch, the tests fail.
I'll be more explicit though.

I'll add the suggested test as well.

Thanks for the hints.

njsmith:
> Although there is no explicit test for the filter part, I'm pretty sure the part I fixed are tested, because without the patch, the tests fail.

Oh, I see... OK right, what's happening is that when you throw the exceptions through the nursery, the nursery's cancel scope is internally using `MultiError.filter` to check for `Cancelled` exceptions. So that does exercise `filter`. But, yeah, it's good to have an explicit check too :-).

The new test is *almost* there, but I have one more suggestion: there are actually two things we think are true:

* On all versions of python, `format_exception` should correctly handle circular tracebacks for regular hashable exceptions
* On newer versions of python, if you feed an unhashable exception into `format_exception` then it shouldn't crash

We know that before this PR, the test suite didn't check for either of properties – the second one because it was broken in master and the tests still passed, and the first one because it was broken in the earlier version of this PR and the tests still passed. So we should have tests for both of them.

The test you added mixes them together though – in particular, it doesn't test that `format_exception` handles circular tracebacks on older versions of Python. So I think it would still have allowed the original version of this PR through.

I'd split it into two tests: one that checks a circular exception with regular hashable exceptions, and then a second test that checks that `format_exception` doesn't crash on non-hashable exceptions. Strictly speaking, we don't really need a test that combines circular + non-hashable exceptions, because the two tests alone already exercise all the different paths, but I guess you could throw one of those in too if you want :-)

trio/_core/tests/test_multierror.py
```diff
@@ -91,6 +113,52 @@ def test_MultiError():
         MultiError([KeyError(), ValueError])
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 6, 4), reason="Unsupported in python < 3.6.4"
+)
```
njsmith: Also, I think this should actually pass on 3.5, because it's not calling the traceback logic? If so then the `skipif` should be removed.

ziirish: Actually, pytest must be calling `format_exception` implicitly or something, because even with the latest patch, this test wail on 3.5.

Now this test is actually redundant with the two others, so maybe we can just remove it entirely?

ziirish: The test actually passes on 3.5 so I removed the `skipif`

njsmith:
All right, looks good!

PR #1026, opened 2019-04-26
Add socket.bind to the list of async socket methods

tjstum:
I was recently using trio for a project to make a UDP listening
socket, and I didn't know to await on `socket.bind`. The warning that
I got about the coroutine never being awaited tipped me off, but it
would still be nice to have this in the docs.

I also wrote up about how to build the docs locally. I haven't
contributed to a lot of well-documented Python projects, so this was
instructive for me.


tjstum:
Please feel free to tell me:
- to go away
- to separate these into constituent parts
- to include one but not the other

This tripped me up the other day, so I figured I'd help the next person, but I will understand if this is too trivial. Thanks!

njsmith:
Hello! Thank you!

In general it's usually better to split things up into multiple PRs. This is as much for the submitter's sake as for the reviewer's... small individual things tend to get merged much quicker :-). These is pretty simple though so I think we can make it work like this.

The `bind` thing was definitely an oversight; nice catch.

Re: how to build the docs: this is excellent, thank you! It has also revealed an important reason why we should document it: we actually have a requirements file for this that you failed to find, because it isn't documented :-). Instead of listing the specific packages, we should tell people to `pip install -r ci/rtd-requirements.txt`. Maybe we should rename that to `docs-requirements.txt` or something? (If you do rename it please do make that a separate PR :-).)

Nitpick: there's a missing period at the end of the new section, after `browser`.

tjstum:
OK, I think I've incorporated all we've talked about here. I wasn't sure if the preferred way to bring the branch up to date was to use `merge` or `rebase` (I come from rebase workflows), so I picked merge. I am completely happy to rewrite the branch if that is the house style.

Thanks for your help and for making trio! :-)

oremanj:
You addressed all of @njsmith's comments, and I'm guessing he's busy at PyCon, so I'll go ahead and merge this. Thanks for contributing!

njsmith:
:+1:

> I wasn't sure if the preferred way to bring the branch up to date was to use merge or rebase (I come from rebase workflows), so I picked merge. I am completely happy to rewrite the branch if that is the house style.

Our house style is that it's more important to make things easy for contributors than to enforce a nitpicky house style. So use whichever is easier for you :-)

PR #1035, opened 2019-05-02
Fix each requirements.in to generate correct requirements.txt even when run on 3.7

### LINKED ISSUE #1034: dependabot broken by obscure pip-compile / py37 issues
I still don't fully understand what's going on here, but see #1032 for discussion and #1033 for proof that until we have a solution all our dependabot updates are broken. Opening a new issue too better reflect the problem and to improve our chances of finding it later if we have to.

oremanj:
Hopefully this will resolve #1034. I tested with `pip-compile` on 3.7 and we do get the `cpython` restriction on typed-ast + inclusion of `immutables` in the output.

### DIFF
--- a/docs-requirements.in
+++ b/docs-requirements.in
@@ -13,3 +13,6 @@ async_generator >= 1.9
 idna
 outcome
 sniffio
+
+# See note in test-requirements.in
+immutables >= 0.6
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -13,7 +13,7 @@ yapf ==0.27.0  # formatting
 flake8
 
 # https://github.com/python-trio/trio/pull/654#issuecomment-420518745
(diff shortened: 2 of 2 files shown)


oremanj:
"The remote name could not be resolved: 'mrtytfsprodcus3file1.queue.core.windows.net'"

Okay Azure...

njsmith:
Unfortunately typed-ast is incompatible with 3.8-dev, which I suspect will cause problems... maybe a `<3.8` restriction would help?

oremanj:
Come on, Azure, it's not even midnight in Iowa.

oremanj:
I get a perfectly fine codecov report by following the link to their website. I guess the submission to Github didn't work?

I think this ought to be OK to merge.

njsmith:
Well, it's a hacky patch, but hopefully by the time it breaks we'll have better options :-)

PR #1038, opened 2019-05-04
Update happy eyeballs default timeout for RFC 8305

brl1906:
first commit baby!  pycon2019


newsfragments/762.misc.rst
```diff
@@ -0,0 +1 @@
+To make the default timeout in trio.open_tcp_stream .250 which is consistent with RFC 8305.
```
njsmith:
```suggestion
Change the default timeout in :func:`trio.open_tcp_stream` to 0.250 seconds, for consistency with RFC 8305.
```

Issue #676, opened 2018-09-21
Always consider __del__ to be KeyboardInterrupt protected


alexshpilkin:
I just tried to interrupt a long-running program of mine and got the following message:
```
^CException ignored in: <function Nursery.__del__ at 0x7fb575698510>
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/trio/_core/_run.py", line 520, in __del__
    def __del__(self):
  File "/usr/lib/python3.7/site-packages/trio/_core/_ki.py", line 191, in handler
    raise KeyboardInterrupt
```
The interrupt was indeed ignored. A second interrupt a dozen seconds later did the job; there were several more web requests issued from the main task in between. After reading the docs, I was under the impression that this was never, ever supposed to happen (unless done on purpose).

What puzzles me is that my program didn’t create any nurseries or Tasks _at all_. It did use `asks` with connections > 1, though, but did not—itself—issue more than one request at a time.

I’m sorry that I can’t reproduce this reliably, but my only evidence is a failure of a single several-hour-long run of a web scraper, with several previous runs coming out OK. I’m just writing this up in case it can be fixed by staring at the code for `Nursery.__del__`.

njsmith:
Huh! That is an interesting corner case. Thanks for pointing this out.

So here are two general facts about python:

* `__del__` methods run in a strange context where they can execute arbitrary python code, but cannot raise exceptions

* Signal handlers, like the one that raises `KeyboardInterrupt`, can interrupt any python code

So i guess in any python program that uses `__del__`, there is some chance of `KeyboardInterrupt`s being lost, if you get really unlucky and hit control-C just as some `__del__` method runs. (Nothing special about `Nursery` here, it could be any python object with a `__del__`.) And trio does a lot of stuff to make control-C behave better, but since i never thought about this weird situation before, trio currently doesn't have any workaround for it, so it has the same flaw as regular python here. Neat! Also kind of annoying.

We could add a `@enable_ki_protection` decorator to all our `__del__` methods, but this still leaves a window where it can happen (because of `@enable_ki_protection`'s lack of atomicity), and also doesn't help with any other `__del__` methods in other libraries.

We could declare that all callables named `"__del__"` are automatically KI-protected. i guess this could go wrong in some edge cases:

* Someone calls `__del__` manually instead of through the GC: we can't tell who called it, so we have to assume it's the GC and defer raising the exception, even though in this case we could have safely raised

[...]

PR #1040, opened 2019-05-04
Enabled KeyboardInterrupt protection inside __del__ methods

StephenSchroeder:
Issue #676

### DIFF
--- a/newsfragments/676.bugfix.rst
+++ b/newsfragments/676.bugfix.rst
@@ -1,0 +1,1 @@
+Destructor methods (``__del__``) are now protected against ``KeyboardInterrupt``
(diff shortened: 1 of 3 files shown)


newsfragments/676.bugfix.rst
```diff
@@ -0,0 +1 @@
+Destructor methods (``__del__``) are now protected against ``KeyboardInterrupt``
```
oremanj: Nit: can you add a period to the end here, to match the style of the other newsfragments?

trio/_core/tests/test_ki.py
```diff
@@ -202,6 +202,18 @@ def protected_manager():
 def test_ki_enabled_out_of_context():
     assert not _core.currently_ki_protected()
 
+def test_ki_disabled_in_del():
+    def nestedfunction():
+            return _core.currently_ki_protected()
```
oremanj: Make sure you run `yapf -rpi setup.py trio` and add the resulting changes to your PR; this will update your code to reflect our whitespace and formatting conventions. (For example, here you have an 8-space indent where our convention is 4.)

PR #1041, opened 2019-05-04
Added readable str for Cancelled object

### LINKED ISSUE #674: trio.Cancelled missing readable str()
Trying to print a trio.Cancelled exception e.g. with `f'{exc}'` yields an empty string.  I usually resort to `type(exc)` when I need to see what's going on.

yayoiukai:
Resolve #674

### DIFF
--- a/newsfragments/674.misc.rst
+++ b/newsfragments/674.misc.rst
@@ -1,0 +1,1 @@
+Calling :func:`str` on a :exc:`trio.Cancelled` exception object returns "Cancelled" instead of an empty string.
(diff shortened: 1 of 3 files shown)


newsfragments/674.misc.rst
```diff
@@ -0,0 +1 @@
+Calling :func:`str` on a :exc:`trio.Cancelled` exception object returns "Cancelled" instead of an empty string.
```
oremanj: One of the tests failed on this line (https://travis-ci.org/python-trio/trio/jobs/528278782). I was apparently wrong about `:func:` being correct here (probably because `str` is also a type), and neither I nor @njsmith knows what it should be, so let's just write this as
```
``str``
```
(which won't link) for simplicity.

Issue #1021, opened 2019-04-24
Let's get serious about handling private constructors and preventing subclassing


njsmith:
In Trio we're pretty careful about our public API, because if it takes off then we want to defend ourselves against [Hyrum's law](http://www.hyrumslaw.com/). And because some of its internal state really is pretty delicate.

One consequence: we're kind of protective of some internal classes. For example, nurseries and cancel scopes both originally had no public class at all, because we didn't want to have a public constructor, or allow subclassing. (Even if not everyone shares my opinion that subclassing is a bad idea in general, I think it is at least generally agreed that you shouldn't subclass a type unless that type was intentionally written to enable subclassing, and these types definitely were not.)

Now, cancel scopes have gained a public constructor, and become a public class. And we want to make nurseries a public class too, to enable type annotations (#778). And I forget where now, but I saw someone post some sample code that subclassed `CancelScope`, and it was scary :-). So, we need to rethink this.

So, proposal: Come up with one metaclass/decorator/something that prevents subclassing, and one that prevents subclassing + hides the constructor. Use these for `CancelScope`, `Nursery`, etc. (And maybe everything else too!)

# Preventing subclassing

In 3.6+, this is easy using [PEP 487](https://www.python.org/dev/peps/pep-0487/):

```python
class Blah:
    def __init_subclass__(cls):
        raise TypeError("subclassing not supported")
```

You could also easily define a `@final` decorator to inject this into a class when it's defined. [PEP 591](https://www.python.org/dev/peps/pep-0591/) proposes to add a similar `@final` decorator, which works statically, rather than dynamically. You could combine the two using something like:

```python
if TYPE_CHECKING:
    # static version
    from typing import final
else:
    # runtime version
    from ._util import final
```

However, for now we still need to support Python 3.5, which is a bit trickier. On 3.5 it requires defining a custom metaclass. Something like:

```python
class Final(type):
    def __new__(cls, name, bases, cls_namespace):
        for base in bases:
            if isinstance(base, Final):
                raise TypeError(f"{base.__name__} doesn't support subclassing")
        return type(name, bases, cls_namespace)
```

It's not as easy to inject a metaclass using a decorator (it requires reconstructing the class object), but it's easy to use like:

```python
@typing.final
class Nursery(metaclass=Final):
    ...
```

# No public constructor

Say we have a normal class:

```python
class NormalClass:
    def __init__(self, ...):
        ...
```

[...]

PR #1043, opened 2019-05-06
Enforce private instantiation and inheritance through metaclass

ankitkumarr:
Issue: #1021 

Added two new metaclasses:
1. `Final`: Ensures that no class can inherit the class using the metaclass
2. `NoPublicConstructor`: Ensures that no object of the class using the metaclass can be directly created. This metaclass derives from `Final`, also ensuring 1

### DIFF
--- a/newsfragments/1021.misc.rst
+++ b/newsfragments/1021.misc.rst
@@ -1,0 +1,3 @@
+Any attempt to inherit from :class:`CancelScope` now raises TypeError.
+(Trio has never been able to safely support subclassing here;
+this change just makes it more obvious.)
--- a/trio/tests/test_util.py
+++ b/trio/tests/test_util.py
@@ -8,7 +8,8 @@ import pytest
 from .. import _core
 from trio import run_sync_in_worker_thread
 from .._util import (
-    signal_raise, ConflictDetector, fspath, is_main_thread, generic_function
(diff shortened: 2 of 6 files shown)


newsfragments/1021.misc.rst
```diff
@@ -0,0 +1,3 @@
+Inheriting from :exc:`trio.Cancelled` exception and :class:`CancelScope` is now blocked.
```
oremanj: Wording thoughts:
* we usually use the active voice
* I think it should just be "instantiating X", not "instantiating from X"
* Cancelled being not-publicly-instantiable is not a user-visible change, since it's always been like that

Maybe something like: "Any attempt to inherit from CancelScope now raises TypeError. (Trio has never been able to safely support subclassing here; this change just makes it more obvious.)"

trio/tests/test_util.py
```diff
@@ -184,3 +185,33 @@ def test_func(arg):
...
+
+    with pytest.raises(
+        TypeError, match="SpecialClass does not support subclassing"
+    ):
+
+        class SubClass(SpecialClass):
+            pass
+
+    # Private constructor should not raise
+    SpecialClass._create()
```
oremanj: Maybe assert you actually get an instance of SpecialClass back?

PR #1053, opened 2019-05-16
Consistently capitalize the word Trio

oremanj:


pquentin:
trio everywhere

Great work, thanks! I've noticed a few more, please tell me what you think?

```patch
From 9aeeb19c028547854a774159c9f474d04b0a76cd Mon Sep 17 00:00:00 2001
From: Quentin Pradet <quentin.pradet@gmail.com>
Date: Thu, 16 May 2019 10:15:44 +0400
Subject: [PATCH] Fix up a second set of stragglers

---
 ci.sh                                                           | 2 +-
 notes-to-self/blocking-read-hack.py                             | 2 +-
 notes-to-self/tiny-thread-pool.py                               | 2 +-
 trio/_abc.py                                                    | 2 +-
 .../tests/test_multierror_scripts/simple_excepthook_IPython.py  | 2 +-
 5 files changed, 5 insertions(+), 5 deletions(-)

diff --git a/ci.sh b/ci.sh
index 65f78b4..ec67196 100755
--- a/ci.sh
+++ b/ci.sh
@@ -69,7 +69,7 @@ if [ "$PYPY_NIGHTLY_BRANCH" != "" ]; then
         # server returns 4xx or 5xx")
         # - nonetheless, pypy.tar.bz2 does not exist, or contains no data
         # This isn't going to work, and the failure is not informative of
-        # anything involving trio.
+        # anything involving Trio.
         ls -l
         echo "PyPy3 nightly build failed to download – something is wrong on their end."
         echo "Skipping testing against the nightly build for right now."
diff --git a/notes-to-self/blocking-read-hack.py b/notes-to-self/blocking-read-hack.py
index 76eabc0..8f570c8 100644
--- a/notes-to-self/blocking-read-hack.py
+++ b/notes-to-self/blocking-read-hack.py
@@ -33,7 +33,7 @@ async def blocking_read_with_timeout(fd, count, timeout):
             except OSError as exc:
                 if cancel_requested and exc.errno == errno.ENOTCONN:
                     # Call was successfully cancelled. In a real version we'd
-                    # integrate properly with trio's cancellation tools; here
+                    # integrate properly with Trio's cancellation tools; here
                     # we'll just raise an arbitrary error.
                     raise BlockingReadTimeoutError from None
             print("got", data)
diff --git a/notes-to-self/tiny-thread-pool.py b/notes-to-self/tiny-thread-pool.py
index b718d5f..85afc4e 100644
--- a/notes-to-self/tiny-thread-pool.py
+++ b/notes-to-self/tiny-thread-pool.py
@@ -13,7 +13,7 @@ import Queue
 # we don't need to support job cancellation
 #
 # we do need to mark a thread as "available" just before it
-# signals back to trio that it's done, to maintain the invariant that all
+# signals back to Trio that it's done, to maintain the invariant that all
 # unavailable threads are inside the limiter= protection
 #
 # maintaining this invariant while exiting can be a bit tricky
diff --git a/trio/_abc.py b/trio/_abc.py
index ae4eb8c..2e586d7 100644
--- a/trio/_abc.py
+++ b/trio/_abc.py
@@ -195,7 +195,7 @@ class SocketFactory(metaclass=ABCMeta):
 
         Note that unlike :func:`trio.socket.socket`, this does not take a
         ``fileno=`` argument. If a ``fileno=`` is specified, then
-        :func:`trio.socket.socket` returns a regular trio socket object
+        :func:`trio.socket.socket` returns a regular Trio socket object
         instead of calling this method.
 
         """
diff --git a/trio/_core/tests/test_multierror_scripts/simple_excepthook_IPython.py b/trio/_core/tests/test_multierror_scripts/simple_excepthook_IPython.py
index 57a6b81..6aa1249 100644
--- a/trio/_core/tests/test_multierror_scripts/simple_excepthook_IPython.py
+++ b/trio/_core/tests/test_multierror_scripts/simple_excepthook_IPython.py
@@ -1,6 +1,6 @@
 import _common
 
-# To tickle the "is IPython loaded?" logic, make sure that trio tolerates
+# To tickle the "is IPython loaded?" logic, make sure that Trio tolerates
 # IPython loaded but not actually in use
 import IPython
 
-- 
2.20.1

```

oremanj:
Thanks for spotting the extras! I originally left out the notes-to-self but I'm not sure why :-) and the others were just oversights. Have applied your patch to this PR.

pquentin:
(For reference, a build has failed again because of a slow download: `2019-05-16T06:55:50.7228851Z Fetched 125 MB in 5min 58s (349 kB/s)`)

PR #1058, opened 2019-05-18
Stop requiring an async operation that raises an exception to act as a checkpoint

### LINKED ISSUE #474: Consider relaxing the await-always-checkpoints rule to await-always-checkpoints-on-non-error
Currently, all the async functions in trio proper attempt to always issue a checkpoint on every call. The idea is that you want to be able to tell from reading your code where the checkpoints are, so this makes it more predictable – it's a static guarantee, so you don't have to reason carefully about unclear runtime conditions, and it's easy to explain to users.

However, I'm wondering if we might want to relax this slightly: maybe the rule should be, trio's async functions always checkpoint, *unless they raise an exception* (in which case there's no guarantee either way).

The reasoning here is that guaranteeing the checkpoint invariant for error paths turns out to be very tiresome and error-prone, both to implement and to test. Also, in some sense it's actually impossible to get this perfect – e.g. if you accidentally do `await trio.sleep(1,5)` then Python will raise `TypeError: sleep() takes 1 positional argument ...` before even invoking the function, so this call doesn't checkpoint.

Furthermore, we can guess that *usually* users aren't going to be caring about checkpoints for operations that raise exceptions... for example, it'd be unusual to have a loop where the only `await` was on an operation that always raised an exception that you caught and then looped around on. And the proposed rule is easy to explain, though not quite as easy as the original rule.

Another nice thing is the rule is still sufficiently deterministic that we could check it automatically, as suggested in #152.

oremanj:
Fixes #474.

### DIFF
--- a/trio/_core/_unbounded_queue.py
+++ b/trio/_core/_unbounded_queue.py
@@ -127,8 +127,10 @@ class UnboundedQueue:
             await self._lot.park()
             return self._get_batch_protected()
         else:
-            await _core.cancel_shielded_checkpoint()
-            return self._get_batch_protected()
+            try:
+                return self._get_batch_protected()
+            finally:
+                await _core.cancel_shielded_checkpoint()
 
     def statistics(self):
         """Return an object containing debugging information.
--- a/trio/_core/_unbounded_queue.py
+++ b/trio/_core/_unbounded_queue.py
@@ -128,10 +128,8 @@ class UnboundedQueue:
             await self._lot.park()
             return self._get_batch_protected()
         else:
-            try:
-                return self._get_batch_protected()
-            finally:
-                await _core.cancel_shielded_checkpoint()
+            await _core.cancel_shielded_checkpoint()
+            return self._get_batch_protected()
 
     def statistics(self):
         """Return an object containing debugging information.
--- a/trio/_highlevel_socket.py
+++ b/trio/_highlevel_socket.py
@@ -96,7 +96,7 @@ class SocketStream(HalfCloseableStream):
     async def send_all(self, data):
         if self.socket.did_shutdown_SHUT_WR:
             raise trio.ClosedResourceError("can't send data after sending EOF")
-        with self._send_conflict_detector.sync:
+        with self._send_conflict_detector:
             with _translate_socket_errors_to_stream_errors():
                 with memoryview(data) as data:
                     if not data:
@@ -113,14 +113,15 @@ class SocketStream(HalfCloseableStream):
                         total_sent += sent
 
     async def wait_send_all_might_not_block(self):
-        with self._send_conflict_detector.sync:
+        with self._send_conflict_detector:
             if self.socket.fileno() == -1:
                 raise trio.ClosedResourceError
             with _translate_socket_errors_to_stream_errors():
                 await self.socket.wait_writable()
 
     async def send_eof(self):
-        async with self._send_conflict_detector:
+        with self._send_conflict_detector:
+            await trio.hazmat.checkpoint()
             # On macOS, calling shutdown a second time raises ENOTCONN, but
             # send_eof needs to be idempotent.
             if self.socket.did_shutdown_SHUT_WR:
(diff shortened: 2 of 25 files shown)


trio/_core/_unbounded_queue.py
```diff
@@ -128,10 +128,8 @@ def get_batch_nowait(self):
             await self._lot.park()
             return self._get_batch_protected()
         else:
-            try:
-                return self._get_batch_protected()
-            finally:
-                await _core.cancel_shielded_checkpoint()
+            await _core.cancel_shielded_checkpoint()
+            return self._get_batch_protected()
```
njsmith: I'm not sure this is correct – what if someone else steals the batch while we're rescheduling?

Of course this is `UnboundedQueue` so it doesn't really matter, we're deleting it ASAP anyway.

trio/_highlevel_socket.py
```diff
@@ -114,7 +113,7 @@ def __init__(self, socket):
                         total_sent += sent
 
     async def wait_send_all_might_not_block(self):
-        async with self._send_conflict_detector:
+        with self._send_conflict_detector.sync:
```
njsmith: We can probably get rid of the async version of `ConflictDetector` entirely... that could be done here, or we could open a followup issue if it's easier.

trio/testing/_checkpoints.py
```diff
@@ -13,26 +13,27 @@ def _assert_yields_or_not(expected):
...
         ):
-            raise AssertionError("assert_no_yields block yielded!")
+            raise AssertionError("assert_no_checkpoints block yielded!")
 
 
 def assert_checkpoints():
     """Use as a context manager to check that the code inside the ``with``
-    block executes at least one :ref:`checkpoint <checkpoints>`.
+    block either exits with an exception or executes at least one
+    :ref:`checkpoint <checkpoints>`.
```
njsmith: Technically this is a user-visible change in semantics too. I doubt anyone much will care, but we should probably mention it in the release notes anyway.

trio/_core/_unbounded_queue.py
oremanj: Good point -- fixed.

trio/_highlevel_socket.py
oremanj: I considered that too but wasn't sure if there was a call for it. Have now done so.

oremanj:
The lines codecov marks as uncovered are blank lines and comments -- not sure what's going on there.

njsmith:
> The lines codecov marks as uncovered are blank lines and comments -- not sure what's going on there.

There's one legitimate case I know of where this happens – if something gets merged into master while the CI is half-finished, then codecov can get confused.

That's not what happened here though, and codecov has done this several times recently. Maybe it's another bug, like the one we were hitting in 3.8?

[looks at coverage exclusions in `ci.sh`] Oh crud, I bet it's https://bitbucket.org/pypy/pypy/issues/2943/, now that we're *only* testing pypy3.6 stuff.

...anyway, nothing to do with this PR.

trio/testing/_memory_streams.py
```diff
@@ -67,17 +67,19 @@ def _get_impl(self, max_bytes):
...
 
     async def get(self, max_bytes=None):
-        async with self._fetch_lock:
+        with self._fetch_lock:
             self._check_max_bytes(max_bytes)
             if not self._closed and not self._data:
                 await self._lot.park()
+            else:
+                await _core.checkpoint()
             return self._get_impl(max_bytes)
```
njsmith: I think this has the same problem as `UnboundedQueue` – how do we know that after we wake up, `self._data` won't have disappeared?

oremanj: Well, anyone else consuming `self._data` would have to acquire the `_fetch_lock` ConflictDetector, which can't happen while we're holding it. I agree that it looks suspicious, though, so have cleaned it up.

njsmith: Oh yeah... in fact if this was a problem then the `park` version would have the same problem, because someone could steal the buffer in between the call to `unpark` and when the task actually woke up. And github helpfully hid the `with self._fetch_lock` line from me, which is the part that *actually* gives you critical information that makes it safe :-)

I'm going to revert back to what you had before your last commit...

PR #1059, opened 2019-05-20
Adopt `autoflake` and `pyupgrade` formatters

Zac-HD:
In the spirit of #220, here's some development tooling we've adopted in Hypothesis lately for your consideration.  I was going to write up an issue, but it seemed easier to explain by example 😉 

Both [`autoflake`](https://pypi.org/project/autoflake/) and [`pyupgrade`](https://pypi.org/project/pyupgrade/) aim to canonicalise source code; that is to ensure that equivalent constructs are represented in the same way wherever possible.  They also remove some redundancies!  For example, `autoflake` removes unused (standard library) imports and pointless `pass` statements; while `pyupgrade` standardizes set and dict comprehensions, string formatting, etc.

Personally, I've found the enforced consistency does make reading code somewhat easier!  If we want to take this further, `isort` and `black` would round it out, but IMO they deserve some discussion first.

Thoughts?

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -5,6 +5,10 @@ set -ex
 EXIT_STATUS=0
 
 # Autoformatter *first*, to avoid double-reporting errors
+# (we'd like to run further autoformatters but *after* merging;
+# see https://forum.bors.tech/t/pre-test-and-pre-merge-hooks/322)
+# autoflake --recursive --in-place .
+# pyupgrade --py3-plus $(find . -name "*.py")
 yapf -rpd setup.py trio \
     || EXIT_STATUS=$?
 
--- a/docs/source/reference-core/blocking-trio-portal-example.py
+++ b/docs/source/reference-core/blocking-trio-portal-example.py
@@ -1,5 +1,4 @@
 import trio
-import threading
 
 def thread_fn(portal, receive_from_trio, send_to_trio):
     while True:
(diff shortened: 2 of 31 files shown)


Zac-HD:
There's no change in the number of misses, just fewer vacuous hits and I don't think that should be counted as a bad thing ☹️  (hence my crusade against *percent* coverage as a measure)

pquentin:
I like the idea! I also think that we should switch to black to eventually.

But I believe the current plan is to get this proposal accepted: https://forum.bors.tech/t/pre-test-and-pre-merge-hooks/322 so that we can do those things automatically without requiring users to install and run more and more autoformatters, which is frustrating when submitting your first PR.

(Oh, and I think you need to update docs/source/contributing.rst for this PR to be complete.)

Zac-HD:
Sounds great!

If others agree, we can take out the bit where this runs in CI, but merge the rest to get most of the benefits - unlike Black, these changes are pretty rare.

Zac-HD:
@pquentin - given the CI tooling that we have coming, I've just dropped the CI portion of this PR entirely (beyond comments showing the relevant commands).  

That leaves us with a one-off application of `pyupgrade` and `autoflake`, which I think is still worth doing - indeed it's almost as good as enforcing it, as e.g. unused imports are quite rare!

pquentin:
@Zac-HD For the record, I don't know if we should wait for the CI tooling. Progress seems to be slow, and we're only at the Draft RFC step.

But, yes, the pull request as it is now is much easier to merge since I don't need to gather consensus on whether it's OK or not to continue adding more manual autoformatters.

Thank you!

Issue #1056, opened 2019-05-17
Make the TrioInternalError experience less confusing


njsmith:
We've had several bug reports recently where the filers basically said "I got this `TrioInternalError`, I know it's not a bug in Trio but rather 100% my fault for breaking things, but it said to file a bug so I'm doing that":

* https://github.com/python-trio/trio/issues/1055
* https://github.com/python-trio/pytest-trio/issues/82#issuecomment-490901943

At one level this is fine – a very common mistake that programmers make is to assume that when something goes wrong, it's their fault, when it's actually a fixable bug in the software they're using. I'd rather get some ambiguous cases filed in our tracker than miss out on real bugs. #882 is a good example. In that case it turned out that the reporter was implementing the context manager protocol by hand and their implementation had a bug in it, so the immediate problem was a bug in their code, not in Trio – but this wasn't obvious at first.

Still, it seems like the error message could be clearer and more accurate.

Cases where you get `TrioInternalError`:

* When there's a bug in Trio
* When you raise an exception from a place where the docs explicitly say you aren't allowed to raise an exception (especially `spawn_system_task`, `TrioToken.run_sync_soon`, but also potentially from weird places like the abort callback in `wait_task_rescheduled`)
* When you do something that violates Trio's assumptions, like calling `__enter__`/`__exit__` methods in the wrong order.

Ideally, when people misuse the API, they should get an error telling them that, like a `RuntimeError` or `TypeError` or something, not a `TrioInternalError`. (For example, I guess #882 is now basically "can we give a more specific error message when people do weird things with cancel scope/nursery entry/exit".) But we can't do that 100% of the time. Sometimes we can detect that something weird happened that we didn't anticipate, but by definition this means we don't know why it happened – could be a bug, could be someone messing with us, there's no general way to tell. #1055 is an example – @oremanj diagnosed that it has mis-nested cancel scopes, but in general `async_fn().send(None)` is going to create very weird internal corruption (consider `trio.hazmat.wait_readable(fd).send(None)` – Trio will think that the task has suspended itself and should be resumed when the fd becomes readable, but in fact the task isn't suspended at all and is going to sleep somewhere else instead...).

[...]

Issue #264, opened 2017-08-04
[discussion] What to do about cancel scopes / nurseries inside generators?


njsmith:
So here's a surprising thing that can happen in trio right now: if you have a generator (or async generator, doesn't matter) that yields inside a cancel scope, then the cancel scope remains in effect until you resume the generator and exit it.

For context managers, this is actually what you want, and we make use of this in lots of places, e.g.:
```python
@contextmanager
def fail_at(deadline):
    with move_on_at(deadline) as scope:
        yield scope
    ...

with fail_at(deadline):
    XX
```
So here we have: first `fail_at` runs for a bit, and it starts a cancel scope, and then it yields. And then the XX code runs *with the cancel scope in effect, even though `failt_at` is not on the stack*. And then `fail_at` is resumed, and it exits the cancel scope. A little confusing when you think about the details, but basically this is safe and useful and ultimately does what you'd expect.

However, in general this can cause strange behavior, and also crashes:

```python
import trio

def weird_generator():
    with trio.open_cancel_scope() as cs:
        yield cs

async def main():
    # hold onto a reference to prevent the gc from making this even more confusing:
    gen = weird_generator()
    for cs in gen:
        break
    cs.cancel()
    await trio.sleep(1)

trio.run(main)
```

Notice that even after we leave the loop and abandon the generator, the cancel scope is still in effect... and then the program crashes because trio detects that the cancel scope stack has become inconsistent (it tries to pop a scope that isn't at the top of the stack).

This all also applies to using nurseries inside async generators, because nurseries have an implicit cancel scope.

There are also some more generic issues with cleaning up (async) generators that call trio APIs after yielding, but that's another issue. The cancel scope leakage causes problems even while the generator is alive, and even if we had [PEP 533](https://www.python.org/dev/peps/pep-0533/) then this would still be an issue.

Really the ideal solution here would be to error out if someone tries to `yield` through a cancel scope, IFF they're not implementing a context manager.

I don't think there's any way to do this though without [PEP 521](https://www.python.org/dev/peps/pep-0521/), and even then we'd have to figure out somehow whether this particular generator is implementing a context manager. (I guess some annoying thread-local thing could work -- e.g. make `with trio.this_is_a_context_manager: ...` which sets some internal flag that disables the PEP 521-based checks.)

I don't currently have any other ideas how to improve this, except to put some prominent warning in the documentation.

njsmith:
[...]

PR #1061, opened 2019-05-21
Attempt to recover from cancel scope nesting violations

### LINKED ISSUE #882: Failing to close a CancelScope produces super confusing errors
(update: this appears to have been due to failing to close the CancelScope; see: https://github.com/python-trio/trio/issues/882#issuecomment-488817573)

I'm not sure either way if this is ultimately a dupe of #552, so feel free to close if so. I'm also having a hell of a hard time figuring out how to reliably reproduce this, and it may actually be an issue in trio_asyncio?

## What's the issue?

During teardown, there's a KeyError in ``trio._core._run:CancelScope._remove_task``.

## Steps to reproduce

I haven't been able to repro this outside of one specific script, but I've tried a lot of different things. For the sake of brevity, I'll call that script "voidreader", since that's what I'm actually calling it and if you can surmise any secret sauce from that name then you deserve to know anyways. Background:

+ Voidreader is starting from a call to ``trio.run``
+ It runs a [separate logging system in a companion thread](https://github.com/Badg/bescheid/tree/feature/mvp), also running via ``trio.run``, but I'm pretty confident that isn't related, because I still see the error when I comment out the logging system. However, that library does some patching of the stdlib logging module, which... may or may not be relevant?
+ Voidreader starts some async context managers pretty much immediately. One of them runs a Quart server inside of a call to ``aio_as_trio``; there's other uses of it as well
+ Voidreader is also using asks for some stuff, but I don't think that's relevant either
+ Control-C interrupts stop "just working" inside voidreader. It takes me several attempts to get it to actually exit, and then I see the traceback included below

I've tried reproing with stuff like this:

```python
from contextlib import asynccontextmanager
import asyncio

from trio_asyncio import aio_as_trio
from trio_asyncio import trio_as_aio
import trio
import trio_asyncio


@asynccontextmanager
async def nested():
    print('fly, you fools!')
    yield
    await aio_as_trio(asyncio.sleep(7))


async def main():
    async with trio_asyncio.open_loop() as loop:
        async with nested():
            await trio.sleep(1)


if __name__ == '__main__':
    trio.run(main)
```

but to no avail -- **note that control+C still works here, raising a keyboardinterrupt.** My hunch is that this may have something to do with the Quart stuff that's running in an asyncio context.

## Traceback

```
  File "C:\Users\Nick\venv\foo\lib\site-packages\trio\_core\_run.py", line 1323, in run
    run_impl(runner, async_fn, args)
  File "C:\Users\Nick\venv\foo\lib\site-packages\trio\_core\_run.py", line 1471, in run_impl
    runner.task_exited(task, final_outcome)
  File "C:\Users\Nick\venv\foo\lib\site-packages\trio\_core\_run.py", line 943, in task_exited
    task._cancel_stack[-1]._remove_task(task)
  File "C:\Users\Nick\venv\foo\lib\site-packages\trio\_core\_run.py", line 203, in _remove_task
    self._tasks.remove(task)
KeyError: <Task '__main__.main' at 0x27f9eaf5b00>

  File "C:\Python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "d:\dropbox\projekte\foo\bar\baz\__main__.py", line 36, in <module>
    trio.run(main)
  File "C:\Users\Nick\venv\foo\lib\site-packages\trio\_core\_run.py", line 1329, in run
    ) from exc
trio.TrioInternalError: internal error in trio - please file a bug!
Exception ignored in: <function Nursery.__del__ at 0x0000027F9CEE0730>
  File "C:\Users\Nick\venv\foo\lib\site-packages\trio\_core\_run.py", line 530, in __del__
AssertionError:
PS C:\Users\Nick>
```

## Expected behavior

At the very least, a better traceback -- at one point I thought I was seeing that traceback during *startup*, when ultimately the problem was a recursion issue in my logging patch. While I was trying to repro the trio issue, I stumbled upon the problem in my code, fixed it, and suddenly things ran happily again -- until shutdown, that is, which is where the traceback was originating from anyways. Having more information about surrounding exceptions would have been really helpful early on, to help me isolate which problems were mine vs which problems might be upstream.

Of course also, I'd expect a cleaner shutdown! :)

oremanj:
When exiting an outer cancel scope before some inner cancel scope nested inside it, Trio would previously raise a confusing chain of internal errors. One common source of this misbehavior was an async generator that yielded within a cancel scope. Now the mis-nesting condition will be noticed and in many cases the user will receive a single RuntimeError with a useful traceback. This is a best-effort feature, and it's still possible to fool Trio if you try hard enough, such as by closing a nursery's cancel scope from within some child task of the nursery.

Fixes #882. Relevant to #1056, #264.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -291,7 +291,12 @@ class CancelStatus:
         return min(self._scope.deadline, self._parent.effective_deadline())
 
 
-MISNESTING_ADVICE = """Typically this is caused by one of the following:
+MISNESTING_ADVICE = """
+This is probably a bug in your code, that has caused Trio's internal state to
+become corrupted. We'll do our best to recover, but from now on there are
+no guarantees.
+
+Typically this is caused by one of the following:
   - yielding within a generator or async generator that's opened a cancel
     scope or nursery (unless the generator is a @contextmanager or
     @asynccontextmanager); see https://github.com/python-trio/trio/issues/638
@@ -409,7 +414,7 @@ class CancelScope(metaclass=Final):
                 # without changing any state.
                 new_exc = RuntimeError(
                     "Cancel scope stack corrupted: attempted to exit {!r} "
-                    "from unrelated {!r}\n\n{}".format(
+                    "from unrelated {!r}\n{}".format(
                         self, scope_task, MISNESTING_ADVICE
                     )
                 )
@@ -423,7 +428,7 @@ class CancelScope(metaclass=Final):
                 # pass silently.
                 new_exc = RuntimeError(
                     "Cancel scope stack corrupted: attempted to exit {!r} "
-                    "in {!r} that's still within its child {!r}\n\n{}".format(
+                    "in {!r} that's still within its child {!r}\n{}".format(
                         self, scope_task, scope_task._cancel_status._scope,
                         MISNESTING_ADVICE
                     )
@@ -1281,7 +1286,7 @@ class Runner:
                 # traceback frame included
                 raise RuntimeError(
                     "Cancel scope stack corrupted: cancel scope surrounding "
-                    "{!r} was closed before the task exited\n\n{}".format(
+                    "{!r} was closed before the task exited\n{}".format(
                         task, MISNESTING_ADVICE
                     )
                 )
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -297,8 +297,8 @@ MISNESTING_ADVICE = """Typically this is caused by one of the following:
     @asynccontextmanager); see https://github.com/python-trio/trio/issues/638
   - manually calling __enter__ or __exit__ on a trio.CancelScope, or
     __aenter__ or __aexit__ on the object returned by trio.open_nursery();
-    doing so correctly is difficult and you should use ExitStack or
-    AsyncExitStack instead
+    doing so correctly is difficult and you should use @[async]contextmanager
+    instead, or maybe [Async]ExitStack
   - using [Async]ExitStack to interleave the entries/exits of cancel scopes
     and/or nurseries in a way that couldn't be achieved by some nesting of
     'with' and 'async with' blocks
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -173,7 +173,8 @@ class CancelStatus:
     # The CancelStatus whose cancellations can propagate to us; we
     # become effectively cancelled when they do, unless scope.shield
     # is True.  May be None (for the outermost CancelStatus in a call
-    # to trio.run(), or briefly during TaskStatus.started()).
+    # to trio.run(), briefly during TaskStatus.started(), or during
+    # recovery from mis-nesting of cancel scopes).
     _parent = attr.ib(default=None, repr=False)
 
     # All of the CancelStatuses that have this CancelStatus as their parent.
@@ -184,6 +185,12 @@ class CancelStatus:
     # this directly; instead, use Task._activate_cancel_status().
     # Invariant: all(task._cancel_status is self for task in self._tasks)
     _tasks = attr.ib(factory=set, init=False, repr=False)
+
+    # Set to True on still-active cancel statuses that are children
+    # of a cancel status that's been closed. This is used to permit
+    # recovery from mis-nested cancel scopes (well, at least enough
+    # recovery to show a useful traceback).
+    abandoned_by_misnesting = attr.ib(default=False, init=False, repr=False)
 
     def __attrs_post_init__(self):
         if self._parent is not None:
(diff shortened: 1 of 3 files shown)


belm0:
For the lazy, would you post an example traceback with and without this change?

oremanj:
```
$ cat t.py
import trio
from contextlib import ExitStack

@trio.run
async def example():
    outer = trio.CancelScope()
    inner = trio.CancelScope()
    with ExitStack() as stack:
        stack.enter_context(outer)
        with inner:
            stack.close()
```

With this change:
```
  File "t.py", line 4, in <module>
    @trio.run
  File "/Users/oremanj/rc/trio/trio/_core/_run.py", line 1687, in run
  File "t.py", line 11, in example
    stack.close()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 519, in close
    self.__exit__(None, None, None)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 511, in __exit__
    raise exc_details[1]
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 496, in __exit__
    if cb(*exc_details):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 377, in _exit_wrapper
    return cm_exit(cm, exc_type, exc, tb)
  File "/Users/oremanj/rc/trio/trio/_core/_ki.py", line 167, in wrapper
    return fn(*args, **kwargs)
  File "/Users/oremanj/rc/trio/trio/_core/_run.py", line 464, in __exit__
    raise remaining_error_after_cancel_scope
RuntimeError: Cancel scope stack corrupted: attempted to exit <trio.CancelScope at 0x1043c7320, active> in <Task '__main__.example' at 0x1043df208> that's still within its child <trio.CancelScope at 0x1043c7388, active>

Typically this is caused by one of the following:
  - yielding within a generator or async generator that's opened a cancel
    scope or nursery (unless the generator is a @contextmanager or
    @asynccontextmanager); see https://github.com/python-trio/trio/issues/638
  - manually calling __enter__ or __exit__ on a trio.CancelScope, or
    __aenter__ or __aexit__ on the object returned by trio.open_nursery();
    doing so correctly is difficult and you should use @[async]contextmanager
    instead, or maybe [Async]ExitStack
  - using [Async]ExitStack to interleave the entries/exits of cancel scopes
    and/or nurseries in a way that couldn't be achieved by some nesting of
    'with' and 'async with' blocks
  - using the low-level coroutine object protocol to execute some parts of
    an async function in a different cancel scope/nursery context than
    other parts
If you don't believe you're doing any of these things, please file a bug:
https://github.com/python-trio/trio/issues/new

```

Without this change:
```
$ python t.py
  File "/Users/oremanj/rc/triovenv37/lib/python3.7/site-packages/trio/_core/_run.py", line 1430, in run
    run_impl(runner, async_fn, args)
  File "/Users/oremanj/rc/triovenv37/lib/python3.7/site-packages/trio/_core/_run.py", line 1579, in run_impl
    runner.task_exited(task, final_outcome)
  File "/Users/oremanj/rc/triovenv37/lib/python3.7/site-packages/trio/_core/_run.py", line 1050, in task_exited
    task._cancel_stack[-1]._remove_task(task)
  File "/Users/oremanj/rc/triovenv37/lib/python3.7/site-packages/trio/_core/_run.py", line 347, in _remove_task
    self._tasks.remove(task)
KeyError: <Task '__main__.example' at 0x105074940>

  File "t.py", line 4, in <module>
    @trio.run
  File "/Users/oremanj/rc/triovenv37/lib/python3.7/site-packages/trio/_core/_run.py", line 1436, in run
    ) from exc
trio.TrioInternalError: internal error in trio - please file a bug!
Exception ignored in: <function Nursery.__del__ at 0x104e3f620>
  File "/Users/oremanj/rc/triovenv37/lib/python3.7/site-packages/trio/_core/_run.py", line 638, in __del__
AssertionError:
```

trio/_core/_run.py
```diff
@@ -185,6 +186,12 @@ class CancelStatus:
     # Invariant: all(task._cancel_status is self for task in self._tasks)
     _tasks = attr.ib(factory=set, init=False, repr=False)
 
+    # Set to True on still-active cancel statuses that are children
+    # of a cancel status that's been closed. This is used to permit
+    # recovery from mis-nested cancel scopes (well, at least enough
+    # recovery to show a useful traceback).
+    abandoned_by_misnesting = attr.ib(default=False, init=False, repr=False)
```
njsmith: Is this a new public API?

trio/_core/_run.py
```diff
@@ -213,9 +220,43 @@ def children(self):
     def tasks(self):
         return frozenset(self._tasks)
 
+    def encloses(self, other):
+        """Returns true if this cancel status is a direct or indirect
+        parent of cancel status *other*, or if *other* is *self*.
+        """
```
njsmith: Is this a new public API?

trio/_core/_run.py
```diff
@@ -245,6 +291,25 @@ def effective_deadline(self):
         return min(self._scope.deadline, self._parent.effective_deadline())
 
 
+MISNESTING_ADVICE = """Typically this is caused by one of the following:
```
njsmith: Let's start with a nice visible "This is probably a bug in your code, that has caused Trio's internal state to become corrupted. We'll do our best to recover, but from now on there are no guarantees. [blank line] Typically ..."

trio/_core/_run.py
oremanj: CancelStatus isn't part of the public API, so abandoned_by_misnesting and encloses() aren't either. They are non-underscore-prefixed because they're safe to use by other code in _run without worrying about CancelStatus's invariants.

oremanj:
Codecov report is nonsensical on this one too.

njsmith:
Looks good – let's see if it helps :-)

PR #1063, opened 2019-05-25
Fix shield code example

pohlt:
Add missing "trio." to move_on_after call.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -423,7 +423,7 @@ attribute to :data:`True`::
        try:
            await conn.send_hello_msg()
        finally:
-           with move_on_after(CLEANUP_TIMEOUT) as cleanup_scope:
+           with trio.move_on_after(CLEANUP_TIMEOUT) as cleanup_scope:
                cleanup_scope.shield = True
                await conn.send_goodbye_msg()
 


njsmith:
Nice catch, thanks!

And, no pressure, but if you'd like to keep contributing then we'd love to have you, so I'm sending you a github invite now. You can read more about this in [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team).

pohlt:
Sure, glad I could help, even if it's just a single word I added.

I've been using asyncio for some years now and never understood why there have to be so many classes and concepts which seem to do the same thing: encapsulate an async call. Thanks to your great blog posts I understood some more, but I'm still not knowledgeable enough for serious contributions to trio.

Anyway, for a new project I'm planning to use trio which I had on the radar (just like curio). As you know better than me, it's tricky to deal with exceptions in asyncio and I had lots of exceptions while talking to an MRI scanner using REST and Websockets. Back then I would have loved to have trio around.

So, thanks a lot for your great work on trio. I really like your style of communication in the blog and the trio docs. You are creating a very friendly and welcoming atmosphere, so I'm optimistic that this project will thrive. I'll post PRs where I feel competent enough like documentation.

Best,
Tom
________________________________

PR #1067, opened 2019-05-28
add MultiError.__init__ to avoid linter confusion

### LINKED ISSUE #1066: MultiError.exceptions foils linter
Access of `exceptions` on a MultiError instance elicits "no-member" error from Pylint.

I believe it's because MultiError `__new__` is merely mutating the result of `BaseException.__new__(MultiError)`.  I'd address it by making a proper `__init__()` constructor.

belm0:
Fixes #1066

### DIFF
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -176,6 +176,15 @@ class MultiError(BaseException):
 
     """
 
+    def __init__(self, exceptions):
+        # Avoid recursion when exceptions[0] returned by __new__() happens
+        # to be a MultiError and subsequently __init__() is called.
+        if hasattr(self, "exceptions"):
+            # __init__ was already called on this object
+            assert len(exceptions) == 1 and exceptions[0] is self
+            return
+        self.exceptions = exceptions
+
     def __new__(cls, exceptions):
         exceptions = list(exceptions)
         for exc in exceptions:
@@ -184,11 +193,18 @@ class MultiError(BaseException):
                     "Expected an exception object, not {!r}".format(exc)
                 )
         if len(exceptions) == 1:
+            # If this lone object happens to itself be a MultiError, then
+            # Python will implicitly call our __init__ on it again.  See
+            # special handling in __init__.
             return exceptions[0]
         else:
-            self = BaseException.__new__(cls)
-            self.exceptions = exceptions
-            return self
+            # The base class __new__() implicitly invokes our __init__, which
+            # is what we want.
+            #
+            # In an earlier version of the code, we didn't define __init__ and
+            # simply set the `exceptions` attribute directly on the new object.
+            # However, linters expect attributes to be initialized in __init__.
+            return BaseException.__new__(cls, exceptions)
 
     def __str__(self):
         return ", ".join(repr(exc) for exc in self.exceptions)
--- a/newsfragments/1066.feature.rst
+++ b/newsfragments/1066.feature.rst
@@ -1,0 +1,2 @@
+MultiError now defines its ``exceptions`` attribute in ``__init__()``
+to better support linters and code autocompletion.
--- a/trio/_core/tests/test_multierror.py
+++ b/trio/_core/tests/test_multierror.py
@@ -112,6 +112,16 @@ def test_MultiError():
         MultiError([KeyError(), ValueError])
 
 
+def test_MultiErrorOfSingleMultiError():
+    # For MultiError([MultiError]), ensure there is no bad recursion by the
+    # constructor where __init__ is called if __new__ returns a bare MultiError.
+    exceptions = [KeyError(), ValueError()]
+    a = MultiError(exceptions)
+    b = MultiError([a])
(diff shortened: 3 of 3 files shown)


njsmith:
```
=================================== FAILURES ===================================
______________________ test_system_task_crash_MultiError _______________________
    def test_system_task_crash_MultiError():
        async def crasher1():
            raise KeyError
    
        async def crasher2():
            raise ValueError
    
        async def system_task():
                nursery.start_soon(crasher1)
                nursery.start_soon(crasher2)
    
        async def main():
            _core.spawn_system_task(system_task)
            await sleep_forever()
    
        with pytest.raises(_core.TrioInternalError) as excinfo:
            _core.run(main)
    
        me = excinfo.value.__cause__
        assert isinstance(me, _core.MultiError)
>       assert len(me.exceptions) == 2
E       assert 1 == 2
E         -1
E         +2
../../../../virtualenv/python3.6-dev/lib/python3.6/site-packages/trio/_core/tests/test_run.py:1062: AssertionError
```

A news fragmnet would be appropriate too.

belm0:
I'm baffled by the test failure.   Somehow the exceptions field contains a MultiError instance.

Why only this particular test?

oremanj:
@belm0 Looks like you're running into the fact that Python calls `__init__` on the object returned by `__new__` if it's of the type that the `__new__` was called on. So let's say you have
```
assert type(e) is MultiError
assert e.exceptions == [a, b]
```
and you say `MultiError([e])`. There's only one exception there, so `MultiError.__new__([e])` returns `e`. But that's itself a MultiError, so its `__init__` gets called, setting `e.exceptions = [e]`.

I think this might be more trouble than it's worth to fix before MultiError v2 / ExceptionGroup arrives.

belm0:
Only an error in Python 3.8 now, but seems unrelated other that it's coming from the file to which I added a test.
```
_______________ ERROR collecting _core/tests/test_multierror.py ________________
../../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/trio/_core/tests/test_multierror.py:660: in <module>
    import IPython
../../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/IPython/__init__.py:55: in <module>
    from .terminal.embed import embed
../../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/IPython/terminal/embed.py:15: in <module>
    from IPython.core.interactiveshell import DummyMod, InteractiveShell
../../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/IPython/core/interactiveshell.py:339: in <module>
    class InteractiveShell(SingletonConfigurable):
../../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/IPython/core/interactiveshell.py:2909: in InteractiveShell
    def run_cell_async(self, raw_cell: str, store_history=False, silent=False, shell_futures=True) -> ExecutionResult:
/opt/python/3.8-dev/lib/python3.8/asyncio/coroutines.py:111: in coroutine
    warnings.warn('"@coroutine" decorator is deprecated since Python 3.8, use "async def" instead',
E   DeprecationWarning: "@coroutine" decorator is d
```

trio/_core/_multierror.py
```diff
@@ -186,9 +193,7 @@ def __new__(cls, exceptions):
         if len(exceptions) == 1:
             return exceptions[0]
         else:
-            self = BaseException.__new__(cls)
-            self.exceptions = exceptions
-            return self
+            return BaseException.__new__(cls, exceptions)
```
njsmith: Well.... okay, but this is super confusing and it still took me a good 5 minutes to figure out what was happening even in the context of the discussion above. So how about some more comments:

```python
if len(exceptions) == 1:
    # If this object happens to itself be a MultiError, then Python will
    # implicitly call our __init__ again, which we don't want and have
    # guard against.
    return exceptions[0]
else:
    # This always implicitly invokes __init__, which is what we want.
    return BaseException.__new__(cls, exceptions)
```

trio/_core/_multierror.py
```diff
@@ -176,6 +176,13 @@ class MultiError(BaseException):
 
     """
 
+    def __init__(self, exceptions):
+        # Avoid recursion when exceptions[0] returned by __new__() happens
+        # to be a MultiError and subsequently __init__() is called.
+        if self in exceptions:
+            return
+        self.exceptions = exceptions
```
njsmith: Let's also point to #1066 here to explain why we even have an `__init__` method.

Also, how about making the check:

```python
if hasattr(self, "exceptions"):
    # __init__ was already called on this object
    assert len(exceptions) == 1 and exceptions[0] is self
    return
```

njsmith:
(And the test failure will be fixed by #1076)

trio/_core/_multierror.py
belm0: done, though I explained the reason directly rather than reference a historical bug

njsmith:
Looks great, thanks!

PR #1075, opened 2019-06-01
Travis CI: Detect Python 3.7 syntax errors

cclauss:
https://travis-ci.org/python-trio/trio/jobs/540085064#L202

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,9 @@ matrix:
     - python: 3.5
     - python: 3.6
     - python: 3.7
+      before_script:
+        - pip install flake8
+        - flake8 --select=E9
     - python: 3.5-dev
     - python: 3.6-dev
     - python: 3.7-dev


njsmith:
Sorry, can you say a bit more about what this is and what prompted you to submit it?

cclauss:
Flake8 is a linter like pylint, Mypy but one of its tests (E999) builds an abstract syntax tree for each Python file and flags those files with syntax errors.  In a compiled language, these files would not compile.  In Python, they throw an exception at runtime.  I know that these are _notes to self_ but it is still interesting to know that they will crash.

njsmith:
I guess that is interesting, but we use the CI to warn us whether our code is broken, not to highlight general interesting facts :-). And we're already running flake8 over our actual code, as well as an extensive testsuite, so there's no risk that we'll miss syntax errors. So I think we'll pass on this, thanks!

PR #1077, opened 2019-06-02
Update pypy versions in travis

njsmith:
The "pypy3.5" tag points to pypy3.5 v5.10, while we actually want to
test against the latest pypy3 release, which is pypy3.6 v7.1. For now
at least the "pypy3" tag seems to be up to date, so let's use that and
cross our fingers that it stays that way.

We were also still testing against the pypy3.5 branch nightlies, even
though that branch is now abandoned in favor of the pypy3.6 branch. We
were already testing against pypy3.6 nightlies, so we can just drop
the pypy3.5 nightlies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,7 @@ matrix:
     - python: 3.6
       env: CHECK_FORMATTING=1
     # The pypy tests are slow, so we list them first
-    - python: pypy3.5
-    - language: generic
-      env: PYPY_NIGHTLY_BRANCH=py3.5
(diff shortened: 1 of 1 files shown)


njsmith:
It looks like codecov is confused because of merges happening during CI. It would be good to see if this changes coverage (maybe we have some pypy35 only code somewhere), so I guess I'll rerun after the dependabot queue clears out.

njsmith:
the coverage report is still broken and nonsensical

PR #1084, opened 2019-06-04
Sort out coverage issues on pypy3

njsmith:
split off the codecov pypy3 results because we're suspicious pypy might be causing the weird coverage issues we're seeing

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -124,9 +124,15 @@ else
 
     # Disable coverage on 3.8 until we run 3.8 on Windows CI too
     #   https://github.com/python-trio/trio/pull/784#issuecomment-446438407
-    if [[ "$(python -V)" != Python\ 3.8* ]]; then
-        # Special flag for pypy3.6 builds, so we can track weird coverage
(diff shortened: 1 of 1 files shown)


njsmith:
OK, confirmed that pypy 3.6 is the problem... in https://codecov.io/gh/python-trio/trio/src/06a497b16e625e70918be89a0b21ed0693c16dc2/trio/_ssl.py, like 306 is a comment that's marked as uncovered, and clicking the flag buttons at the top shows that both pypy3.6-release and pypy3.6-nightly are including this, but nothing else is. So it's a bug in pypy/coverage.

njsmith:
@cfbolz committed a patch, which should be in the pypy36 nightly now; let's see if it worked.

njsmith:
OK I think this is ready to go, and fixes the issues created by my merging #1077 without following up on the weird coverage I noticed.

Specifically it:

- disables coverage uploading for the pypy3 release build, because that's currently broken
- leaves coverage uploading enabled for the pypy3 nightly, because that's fixed now
- leaves some codecov "flags" in place so if this recurs it will be easier to debug

Issue #1081, opened 2019-06-04
Increase sleep time leniency in test_timeouts.py


njsmith:
Currently we allow 20% error, but this run on macos on azure had 40% error: https://dev.azure.com/python-trio/trio/_build/results?buildId=598&view=logs

(See also: https://github.com/python-trio/trio/issues/200#issuecomment-498509115)

I guess we should bump the error threshold to 50%, by modifying this line of code:

https://github.com/python-trio/trio/blob/5ba7c569dc897777588f11ab37e51a648b493d74/trio/tests/test_timeouts.py#L35

to say `1.5` instead of `1.2`.

epellis:
Just made a PR to fix this. Thanks for leaving easy issues like this for the newbies, I'll try to work on something bigger next!

njsmith:
@epellis Excellent! If you want some help brainstorming things to work on, feel free to ask in [our chat](https://gitter.im/python-trio/general).

PR #1086, opened 2019-06-06
[Sleep-time-leniency] Increase threshold to 50%

epellis:
As per njsmith's sugguestion in #1081. This should close issue #1081.

_Background_
This is an "arbitrary fudge factor" that gives an upper bound on when waiting for an async function to finish. Sometimes this upper bound proves too low, such as in [this CI build](https://dev.azure.com/python-trio/trio/_build/results?buildId=598&view=logs) where a test finished in 1.39 seconds, failing the build. To mitigate this, the factor has been increased to 50%.

### DIFF
--- a/trio/tests/test_timeouts.py
+++ b/trio/tests/test_timeouts.py
@@ -13,7 +13,7 @@ async def check_takes_about(f, expected_dur):
     result = await outcome.acapture(f)
     dur = time.perf_counter() - start
     print(dur / expected_dur)
-    # 1.2 is an arbitrary fudge factor because there's always some delay
+    # 1.5 is an arbitrary fudge factor because there's always some delay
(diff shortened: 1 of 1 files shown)


njsmith:
Looks good!

FYI, github has this clever feature for marking which issues a bug closes: https://help.github.com/en/articles/closing-issues-using-keywords

When you write "This should close issue #1081" it gets confused by the word "issue", but if you write "This should close #1081", then github takes care of closing the issue automatically when it's merged. Not a big deal, but a neat trick to know about :-)

And, as per [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team), I'm sending you a github invite now – you can read more there, but basically it means hey, if you want to stick around and keep helping then we'd love to have you.

PR #1088, opened 2019-06-06
Try installing pytest-azurepiplines on azure pipelines

njsmith:


njsmith:
Here's what it looks like currently: https://dev.azure.com/python-trio/trio/_build/results?buildId=623

I guess the main features are the summary of tests here: https://dev.azure.com/python-trio/trio/_build/results?buildId=623&view=ms.vss-test-web.build-test-results-tab

and the coverage summary in the overview, plus detailed coverage here: https://dev.azure.com/python-trio/trio/_build/results?buildId=623&view=codecoverage-tab

The coverage is totally useless currently. It only captures coverage from one run (no combining), so the data isn't meaningful, and then the coverage tab is the coverage.py HTML output from that one run, with what seems to be all the JS and CSS stripped out, which means all the navigation and red/green highlights are stripped out, so you can't even see the meaningless data!

The test summary might have some value? Generally the only thing that matters is (1) did they pass, (2) if not, then how many clicks does it take to get the detailed error message. There's a bunch of stuff to like, graph tests over time, and IMO for a project like ours its pure clutter. But it's not terrible to see an overview of which tests failed, and maybe it does something useful if a test fails? I guess we could check.

njsmith:
Output with some failing tests: https://dev.azure.com/python-trio/trio/_build/results?buildId=624&view=ms.vss-test-web.build-test-results-tab
Github's annoying summary page: https://github.com/python-trio/trio/pull/1088/checks?check_run_id=143529271

I can see how this could be useful... for each failing pytest test, it does pull out the actual failure message + traceback. There's a *lot* of clunky UI around it though – they've clearly never tested this on small monitors (on my laptop all the useful information is jammed into a 3-line-tall box in the the lower right of my screen), for some reason every run has two duplicate entries (one labeled with the actual run name, and one labeled `'Pytest results'`), even once you get to the test results you still have to do multiple clicks on each failing test to get the details... on balance the annoying "click three times to get to a log, and then scroll through the log to find the actual failure" is still better, because it doesn't take any more futzing about with the UI, and you end up on the full information-rich log.

Really what I want is one click from github to see a single-page summary of everything that failed *including all the available details directly on that page*, and then a second click to drop me into the full log for any individual failure. Note that this is the classic old-travis-without-github-apps workflow: one click gives you an [overview table with all your jobs marked as passed or failed](https://travis-ci.org/python-trio/trio/builds/542466891), and a second click puts you [straight into the log](https://travis-ci.org/python-trio/trio/jobs/542466896). AFAICT everything Github and MS have innovated since then just made this worse. But they're still working on it so maybe they'll figure it out.

Anyway, closing this for now, but we can keep an eye on it.

Issue #1044, opened 2019-05-07
Now that we have a mechanism for blocking subclassing, where should we use it?


njsmith:
Followup to #1021 

Possibly we should use it on... all classes? (Except ABCs.) Possibly not.

----

**Edit on 2020-05-07 to add more rationale:** The question here is about whether we should forbid subclassing on Trio classes.

Subclassing is a controversial topic in general – some people argue that it's an anti-pattern in general; some think there are situations where it's appropriate. But I think even subclassing advocates mostly agree that subclassing only works well when the base class is *intentionally designed* to be subclassed. In particular, without this, you can easily create fragile coupling between the internals of the base class and the subclass. For example, let's say we have a base class that uses the common trick where one method is implemented using another:

```python
class BaseClass:
    def my_op_basic(self, *basic_args):
        ...

    def my_op_extended(self, *extended_args):
        basic_args = process_extended_args(*extended_args)
        return self.my_op_basic(*basic_args)
```

If you make a subclass and want to customize how all the different `my_op` variations work, the obvious  probably just override `my_op_basic`, because that automatically covers both cases:

```python
class SubClass(BaseClass):
    def my_op_basic(self, *basic_args):
        # ... custom logic here ...

    # no override for my_op_extended; base class version is fine
```

We test it, and it works great: our custom logic runs on `SubClass.my_op_basic` *and* `SubClass.my_op_extended`, because `BaseClass.my_op_extended` internally invokes `SubClass.my_op_basic`. But then, someone refactors `BaseClass`, and moves the main logic into `my_op_extended`:

```python
class BaseClass:  # version 2
    def my_op_basic(self, *basic_args):
        extended_args = process_basic_args(*basic_args)
        return self.my_op_extended(*extended_args)

    def my_op_extended(self, *extended_args):
        ...
```

Now, `BaseClass`'s public API hasn't changed at all – but this internal refactoring is still a breaking change for `SubClass`! Now calls to `SubClass.my_op_basic` will invoke the custom logic, but calls to `SubClass.my_op_extended` won't!

The point is: subclasses don't just rely on base class public APIs; they also rely on details of how the base class methods are *implemented*. And then something that *looks* like an innocent refactoring suddenly becomes a breaking change. That can be fine in some cases: if both classes are in the same project, then the tests will probably catch it and you can refactor `SubClass` as well. Or, if the base class was explicitly designed for subclassing so these internal implementation details are documented as part of the public API, then it's not an issue.

[...]

Issue #778, opened 2018-11-17
Better expose Nursery and TaskStatus for type annotations


goodboy:
As per some [gitter banter](https://gitter.im/python-trio/general?at=5bef93919f5a9e6e41f9b0b4) we should more publicly expose the `Nursery`, `CancelScope` and `Taskstatus` types for use as [type annotations](https://www.python.org/dev/peps/pep-0484/).

Currently client code is stuck importing most of these from internal private modules which leads the following beauty:
```python
async def _async_main(                                                                                                                                                                              
   self,                                                                                                                                                                                           
   accept_addr: Tuple[str, int],                                                                                                                                                                   
   arbiter_addr: Optional[Tuple[str, int]] = None,                                                                                                                                                 
   parent_addr: Optional[Tuple[str, int]] = None,                                                                                                                                                  
   task_status: trio._core._run._TaskStatus = trio.TASK_STATUS_IGNORED,                                                                                                                            
) -> None:
     ...
```

Worth noting:
- `CancelScope` may already be exposed soon via python-trio/trio#607
- @njsmith and @Fuyukai already seems interested in adding them internally as per python-trio/trio#543

tgoodlet:
Oh, and let's not forget about the very important [`Task`](https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L552)!

oremanj:
`Task` is already public as `trio.hazmat.Task` -- how were you envisioning it getting more public?

For now I recommend using the `Nursery` and `TaskStatus` exposures in `trio-typing`. When we bring the type hints into Trio proper (#543), I imagine a resolution to this issue will go along with it. I'm going to close this for now, but leave a pointer at #543.

PR #1090, opened 2019-06-08
Hide Nursery constructor, finalize and expose Nursery

bengartner:
Issue: #1021

Seemed like a doable first issue for me since I've dabbled in metaclasses (and they classes were already written.)

Applied NoPublicConstructor to Nursery and added test cases for instantiation and subclassing. Worked pretty easily. Don't know the codebase well enough to know what else we might want to privatize. NurseryManager? CancelStatus? Do we want to protect everything or are there certain classes that are too magic to exist in nature?


njsmith:
We don't need to worry about special metaclasses for `NurseryManager` or `CancelStatus`, because those whole classes are private – you can only get them by doing `trio._core._run.NurseryManager`, and anyone who would do that already knows that they're cheating and isn't going to be stopped by having to write `._create` as well :-). And they're going to stay private going forward.

So far, `Nursery` has also been kept private, but that's actually a problem (#778). The idea was to stop people instantiating it directly, and that part worked, but it also stops people from using it in type hints, or `isinstance` checks, and it makes it annoying to document because we can't just say "here are the docs for `trio.Nursery`", so on.

This means that as it stands, this PR doesn't change anything in the public API – it takes something that was already hidden, and makes it even more hidden :-). But, it makes it possible to start exporting `Nursery` as part of the `trio.*` namespace. Do you want to do that too? Could be part of this PR, or could be a follow-up PR.

njsmith:
Oh, and there are quite a few other classes where we might want to use these metaclasses – #1044 is a placeholder for figuring that out. I'll comment there with more details.

The actual change in this PR looks good. Going forward, I don't think we need to add this many tests every time we mark a class as `Final` or `NoPublicConstructor` – we already have tests for those metaclasses. But `Nursery` is such a core thing, and it's so important that people not mess around with it, that I don't mind having some special tests just for it.

bengartner:
> Do you want to do that too? Could be part of this PR, or could be a follow-up PR.

Yes. I haven't used the __init__ and __all__ patterns for making symbols public before but seems straightforward enough.

Unless we need a newsfragment or explicit test for the visibility of trio.Nursery.

njsmith:
> Yes. I haven't used the `__init__` and `__all__` patterns for making symbols public before but seems straightforward enough.

Yep, you got it.

> Unless we need a newsfragment or explicit test for the visibility of `trio.Nursery`.

We don't need a test, but a newsfragment would be good. We should also add it to the docs.

The docs are more complicated than you might think, because we need to unwind the weird hack we're using to work around `Nursery` not being public. Usually, the way we arrange our docs is:

We put reference docs into  docstrings on each class/function/method, using [Google-style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google), and then in the sphinx manual we use the `autodoc` plugin to pull in the docstrings, and put more explanatory text around them.

For nursery, because it's not a public class, we couldn't do this, so instead we have some awkward workarounds. If you look at `docs/source/reference-core.rst` and search for `.. interface:: The nursery interface`, you'll see this weird thing:

* It's a `.. interface::`, which is a custom thing we invented just for this
* It has all the reference docs directly in the `.rst` file, instead of in the docstrings where they normally go
* They're all written in raw sphinx markup (using the `:param:` tag, etc.), instead of using Google docstring style (this is because autodoc knows how to process the google-style formatting, but sphinx proper doesn't, so if you don't use docstrings you can't use the google-style formatting)

So it would be good to move these docs into docstrings, convert to Google docstring style, and then pull them in with `.. autoclass:: Nursery`.

bengartner:
Okay, I'll take a look at untangling the Nursery workaround.

bengartner:
Okay, I understand what's going on with Nursery, but still figuring out the restructured text syntax. Notably Sphinx can't autoextract cancel_scope from the attribute list in the docstring because it's an instance variable and not a @property. Feels like it should be document with the other attributes of Nursery though.

bengartner:
> 
> 
> It looks like those few lines are some sort of Sphinx boilerplate that you forgot to replace?

Yes, sorry, I have more edits but I accidentally pushed my draft version. Is that a "draft" or "do not merge" label I should apply while I'm working on it?

njsmith:
Github has some kind of official way to mark a PR as "draft status" now, but I don't know if you can set it from an open PR – the docs just describe how to set it when creating a new PR.

The old convention is to edit the title to start with `[wip]`, and then it's obvious to everyone that it's a work-in-progress

bengartner:
I think this is ready now. (And I learned a lot about Sphinx.) Note that Nursery didn't have an actual docstring before, so someone should scrutinize that.

Not sure what's going on with the builds. Is Azure Pipelines having issues?

njsmith:
The macOS failure is weird... might just be some kind of glitch?

The Travis CHECK_DOCS failure is:

```
/home/travis/build/python-trio/trio/trio/__init__.py:docstring of trio.open_nursery:1:Inline interpreted text or phrase reference start-string without end-string.
```

pquentin:
Let's see if [the macOS failure](https://dev.azure.com/python-trio/6904dcc5-a6a9-4551-b6a2-5b3bdae84939/_build/results?buildId=632&view=logs&jobId=911aa8aa-3354-52e6-a2ae-0a0fa559dcbd) is a glitch or not by closing/reopening.

bengartner:
Okay, I think I had some wires crossed after a rebase but I think this looks good now.

pquentin:
This looks really good:

Capture d’écran 2019-06-14 à 15 32 07

Ironically we're now emphasizing the `trio.Nursery` constructor in the docs, but I don't think there's any way around that.

However, one thing that changed that we can probably improve is that start_soon is now at the bottom, but that's the method you want to show first. Ideally we could show `start_soon`, then `start`, then the three other things, as it's done currently. If there's no way to get `cancel_scope` after the two start methods, then so be it.

(@njsmith if you think none of that matters, please feel free to merge!)

bengartner:
> Ironically we're now emphasizing the `trio.Nursery` constructor in the docs, but I don't think there's any way around that.

Oh yeah, good point. And actually, I think I figured out how to deephasize this. If I just pushed this change
`.. autoclass:: Nursery` => `.. autoclass:: Nursery()`
and that hides the constructor params. I think this feature is more for hiding optional parameters in the constructor but it seems to work for obscuring the constructor altogether.

bengartner:
> Ideally we could show `start_soon`, then `start`, then the three other things, as it's done currently.

Yeah, I just have to include the methods/attributes explicitly in the rst, I can do that.

pquentin:
Nice find for the constructor!

Another option is to reorder the sources themselves, but I guess that introduces churn. Anyway, that's minor, so merging. Thank you!

njsmith:
Nice catch on the nursery constructor :-)

Issue #78, opened 2017-03-10
docs: set default_role = "obj" and refactor the docs to take advantage of it


njsmith:
See: http://www.sphinx-doc.org/en/stable/config.html#confval-default_role

oremanj:
This was mostly resolved by #1091, but I just noticed that referring to a function with bare backticks drops the trailing `()` that you get with `:func:` or `:meth:`. This seems like a regression to me... if it's not possible to get the `()` back, then maybe we want to stick with `:func:` and `:meth:` for referring to functions, and only use bare backticks where we would use `:attr:`, `:class:`, or `:data:`?

njsmith:
That does seem like a minor regression, but I'm hesitant to ask
contributors to understand a complicated rule like that. It doesn't seem
like the kind of thing you can apply consistently without wasting a lot of
energy on it. It would be better to see if we can fix Sphinx somehow...

PR #1091, opened 2019-06-08
Refactor to default role obj

bengartner:
Issue #78

Set default_role to obj and refactored out existing :obj: tags. Fixed 3 sphinx warnings for unresolved references and spot checked the docs. I did NOT look at all the defaults being parsed by sphinx, so it's possible some of them have silently changed meaning. I guess the way to check this would be to diff the html output and make sure it's small?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -39,6 +39,7 @@ nitpick_ignore = [
     ("py:exc", "Anything else"),
 ]
 autodoc_inherit_docstrings = False
+default_role = "obj"
 
 # XX hack the RTD theme until
 #   https://github.com/rtfd/sphinx_rtd_theme/pull/382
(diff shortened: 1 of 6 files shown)


njsmith:
> I guess the way to check this would be to diff the html output and make sure it's small?

I'm not super worried, but if you do this then I bet you'll learn something one way or the other :-)

There are also a *lot* of places in the docs where we use explicit role markers like ``:class:`trio.Lock` ``, that could become a plain `` `trio.Lock` `` instead. That doesn't necessarily have to happen as part of this PR, and maybe it's better if it doesn't. But that's where we'll get most of the benefit from having `default_role = "obj"`

bengartner:
Diff was small!

njsmith:
OK, let's put this in and we can worry about reducing `:class:`/`:meth:`/etc clutter in followups

I'm also sending a github invite, as described in [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team). You're not obligated to accept, but we'd certainly love to have you :-)

PR #1101, opened 2019-06-11
Take advantage of new features in cffi 1.12

njsmith:
We asked Armin to add these so it would be rude not to use them:
  https://bitbucket.org/cffi/cffi/issues/394/
  https://bitbucket.org/cffi/cffi/issues/395/


pquentin:
Thank you!

PR #1102, opened 2019-06-11
Show deprecations first in release notes

pquentin:
I've noticed this issue because of https://github.com/python-trio/trio/pull/1099. I believe deprecations should be shown before features and bugfixes. This pull request can be tested with `towncrier --draft`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,11 @@ issue_format = "`#{issue} <https://github.com/python-trio/trio/issues/{issue}>`_
 # tool.towncrier.type.misc.showcontent
 
 [[tool.towncrier.type]]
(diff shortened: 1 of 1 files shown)


njsmith:
Maybe after features but before bugfixes? It should be more prominent than it is, but I'm a little nervous about making it seem like all we do is break stuff :-)

pquentin:
Hmm, I thought it was common to show deprecations/breaking changes first (as a library user, it's always the first section I read) but it looks like it's not:
 * https://keepachangelog.com/en/1.0.0/
 * https://docs.djangoproject.com/en/2.2/releases/2.2/

Closing, thanks for the review.

Issue #928, opened 2019-02-14
Drop usage of SO_EXCLUSIVEADDRUSE on Windows


njsmith:
I recently saw a claim that contra the MSDN documentation, using `SO_REUSEADDR` on Windows no longer creates a bizarre security hole. We have a script in `notes-to-self/` that demonstrates that `SO_EXCLUSIVEADDRUSE` doesn't break things the way MSDN claims it does; but I never did do a test about whether it *fixes* things the way MSDN claims it does. So... for all I know, this might be true! In fact, it's possible that both `SO_REUSEADDR` and `SO_EXCLUSIVEADDRUSE` are no-ops these days.  *Is* this true? If so, when did it become true? Which versions of Windows are affected?

I suspect that the answers won't affect what we do in Trio itself, but it seems worthwhile to know in any case, so at least we know why we do the quirky platform-specific things we do.

See #39, #72 for more background.

njsmith:
# How do these things work?

The official docs on this are here: https://docs.microsoft.com/en-us/windows/desktop/WinSock/using-so-reuseaddr-and-so-exclusiveaddruse

Unfortunately, the formatting on that page is mangled. There are some big tables that include crucial information, but they're completely unreadable. Also, they changed the URL format at some point after the mangling happened, so if you just paste that URL into archive.org you don't get anything useful. But, anyway, here's the old URL, in archive.org, with working tables. I'll link to two versions, for reasons that will become apparent soon: [May 2015](https://web.archive.org/web/20150520052142/https://msdn.microsoft.com/en-us/library/windows/desktop/ms740621(v=vs.85).aspx), [March 2017](https://web.archive.org/web/20170307021231/https://msdn.microsoft.com/en-us/library/windows/desktop/ms740621(v=vs.85).aspx)

The first table is just a bunch of version info. The second purports to tell you how `SO_REUSEADDR` and `SO_EXCLUSIVEADDRUSE` work, but it applies to Windows XP and earlier, so we don't care about that. The *third* table, in the "Enhanced Socket Security" section, is the one we care about. According to the text, it tells you what happens on "Windows Server 2003 and later operating systems".

The most important entry in the table is at the intersection of row "Default / specific" and column "`SO_REUSEADDR` / specific". This tells you what happens if we make a regular socket, bind it to a specific hostname/port, and then later on some miscreant creates a socket with `SO_REUSEADDR` and attempts to hijack our socket.

[...]

Issue #72, opened 2017-03-07
What are the best socket option defaults?


njsmith:
Our default socket configuration is a little different than the classic BSD defaults. Are we making the right choices?

* `SO_REUSEADDR` on non-Windows, `SO_EXCLUSIVEADDR` on Windows: almost certainly correct for listening sockets. More dubious but mostly harmless for client sockets. (See also #39)

  On Windows the other option would be to leave both `SO_REUSEADDR` and `SO_EXCLUSIVEADDR` unset, which gives (a) the preferred `TIME_WAIT` semantics (same as setting `SO_REUSEADDR` on a Unix system), but also (b) weird socket hijacking issues. (This is what libuv does, e.g.) For now our reasoning is "when in doubt, err on the side of security", but really it would be better to make an actual decision :-). And it's not clear what the best trade-off here is. On recent versions of Windows (including I think all the versions Python officially supports), you can only hijack sockets belonging to the same user, which under the traditional Unix security model would be no big deal – if the process belongs to the same user you could also `ptrace` it or whatever, and we already have countermeasures against anyone setting `SO_REUSEADDR` by *accident*. So maybe the security issue is not worth worrying about in practice. But Windows is not Unix, and even on Unix the traditional security model is not what it used to be (e.g. you probably can't `ptrace` random other processes these days!), so ... I'm not confident I understand the risks.

  There's also the option of pushing this into the higher-level networking interface (#73), where it knows whether or not it's dealing with a listening socket.

* `IPV6_V6ONLY` unset: The important thing is to have a consistent default, because this varies across environments in practice. And I guess unset is a little better than set because it simplifies ipv6 support in simple cases?

* `TCP_NODELAY` set: I'm not sure if this is a good idea or not. `asyncio` sets it by default as well. My suspicion is that having `TCP_NODELAY` set is a good idea for most serious protocols (latency is important, Nagle sucks for that, most protocols aren't telnet, and hopefully most serious implementations are not hitting the OS with lots of tiny writes and relying on it to fix things up). But Nagle also serves as a safety belt for low-effort implementations, which are the ones that are least likely to override defaults...

* `TCP_NOTSENT_LOWAT` set to 16 KiB: well, it's... not obviously wrong? Here we're breaking new ground, so we'll have to see how it goes in practice. This is quite a complicated issue; see #76.

njsmith:
[...]

Issue #39, opened 2017-01-28
Handling of SO_REUSEADDR and friends


njsmith:
Before I looked into it, my assumption was that we should default `SO_REUSEADDR` to enabled. But it turns out that things are more complicated than that.

The situation as I understand it:

On Unix:
* For clients, `connect` on an unbound socket will automatically pick a good port + interface, and part of picking a "good" one is that it knows who you're connecting to, so it can strategically re-use local ports. (It's okay to have two client connections use the same local port so long as the peers have different addresses.)
* For servers, `bind` will by default disallow re-use of ports that are in TIME_WAIT, which is generally considered over-fussy these days. So generally it's recommended to enable `SO_REUSEADDR`, which allows to bind to ports that are in TIME_WAIT but otherwise unused.
* For clients that call `bind` before `connect`, you probably *don't* want to use `SO_REUSEADDR`, because it makes it possible to get a port that ends up failing when you call `connect` ([ref](https://idea.popcount.org/2014-04-03-bind-before-connect/)). Though really you're best off not calling `bind` at all, because `connect` can do a better job of `bind`ing than you can, because it has more information at hand.

On Windows:
* For clients, the plain `connect` function acts similar to Unix, AFAIK. (The WSA-level functions like `ConnectEx` are different and require you to `bind` first, but ATM we aren't using those.)
* For servers, you have to enable `SO_EXCLUSIVEADDRUSE` or else any program with the same uid can hijack your port. (Yes! At least this is better than it used to be – in XP and earlier, they didn't even have the uid check.) This is also required to prevent weird problems like being allowed to bind to a wildcard address + port where there is already another program bound to that port on all the concrete addresses. However, the downside is that it also prevents re-using ports that are in TIME_WAIT.
* Never ever ever use `SO_REUSEADDR`, it's totally broken

([Reference for the delightful Windows behavior](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740621(v=vs.85).aspx))

So one option would be to default-enable `SO_REUSEADDR` on Unix and `SO_EXCLUSIVEADDRUSE` on Windows. I'm a bit concerned about whether this will have a negative effect on clients, though – maybe we only want this to be the default for listening sockets? That's trickier. I guess we could set it in `bind` if not overridden? Or maybe we should keep it simple and say that it's default-enabled, and if you want to turn it off again then go for it.

[...]

PR #1108, opened 2019-06-12
Drop usage of SO_EXCLUSIVEADDRUSE on win32 (#928)

esnyder:
Tiny change to stop setting SO_EXCLUSIVEADDRUSE on sockets before binding them in trio.open_tcp_listeners when on win32, per all the messy details in issues 39, 72 and 928.

### DIFF
--- a/newsfragments/928.misc.rst
+++ b/newsfragments/928.misc.rst
@@ -1,0 +1,1 @@
+On win32 we no longer set SO_EXCLUSIVEADDRUSE when binding a socket in :exc:`trio.open_tcp_listeners`.
--- a/trio/_highlevel_open_tcp_listeners.py
+++ b/trio/_highlevel_open_tcp_listeners.py
@@ -120,11 +120,7 @@ async def open_tcp_listeners(port, *, host=None, backlog=None):
                     raise
             try:
                 # See https://github.com/python-trio/trio/issues/39
-                if sys.platform == "win32":
-                    sock.setsockopt(
-                        tsocket.SOL_SOCKET, tsocket.SO_EXCLUSIVEADDRUSE, 1
-                    )
-                else:
+                if sys.platform != "win32":
                     sock.setsockopt(
                         tsocket.SOL_SOCKET, tsocket.SO_REUSEADDR, 1
                     )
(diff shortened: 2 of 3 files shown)


newsfragments/928.misc.rst
```diff
@@ -0,0 +1 @@
+On win32 we no longer set SO_EXCLUSIVEADDRUSE when binding a socket in :exc:`trio.open_tcp_listeners`. See https://github.com/python-trio/trio/issues/39, https://github.com/python-trio/trio/issues/72, and https://github.com/python-trio/trio/issues/928 for the gory details.
```
njsmith: I think you can drop the second sentence. When we generate the release notes, the `928` in the filename will be automatically converted into a link to #928 (you can [see this in action in the release notes](https://trio.readthedocs.io/en/latest/history.html); it's configured by [this bit of magic in our pyproject.toml](https://github.com/python-trio/trio/blob/00de6cb8fdc7cf96d17d015369be11feb3fc4abe/pyproject.toml#L12)). And then if anyone follows the link to #928, they'll see links to #39 and #72 right in the first message.

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -121,9 +121,10 @@ def _compute_backlog(backlog):
             try:
                 # See https://github.com/python-trio/trio/issues/39
                 if sys.platform == "win32":
-                    sock.setsockopt(
-                        tsocket.SOL_SOCKET, tsocket.SO_EXCLUSIVEADDRUSE, 1
-                    )
+                    # See https://github.com/python-trio/trio/issues/928
+                    # We used to set SO_EXCLUSIVEADDRUSE here, but determined
+                    # that we shouldn't.
+                    pass
```
njsmith: Given that the conclusion is `SO_EXCLUSIVEADDRUSE` is kinda useless and no-one cares about it, I think we can drop all mention of it from the code. (The issue tracker, git history, and release notes will still be there to remind us if we ever need to look something up :-).) So I'd collapse the whole if/else down to just:

```python
if sys.platform != "win32":
    sock.setsockopt(tsocket.SOL_SOCKET, tsocket.SO_REUSEADDR, 1)
```

newsfragments/928.misc.rst
esnyder: Thanks, got it.

trio/_highlevel_open_tcp_listeners.py
esnyder: Thanks

njsmith:
Merged! :tada: :birthday: And, as per [our contributing documentation](https://trio.readthedocs.io/en/latest/contributing.html#joining-the-team), a github invite is on the way.

Issue #1109, opened 2019-06-13
Put process-creation into a thread


njsmith:
[Original title: Is process creation *actually* non-blocking?]

In https://github.com/python-trio/trio/issues/1104 I wrote:

> the actual process startup is synchronous, so you could just as well have a synchronous version

But uh... it just occurred to me that I'm actually not sure if this is true! I mean, right now we just use `subprocess.Popen`, which is indeed a synchronous interface. And on Unix, spawning a new process and getting a handle on it is generally super cheap – it's just `fork`. The `exec` is expensive, but that happens after the child has split off – the parent doesn't wait for it.

But on Windows, you call `CreateProcess`, which I think might block the caller while doing all the disk access to set up the new process? Process creation on Windows are notoriously slow, and I don't know how much of that the parent process has to sit and wait for before `CreateProcess` can return.

And even on Unix, you use `vfork`, in which case the parent process is blocked until the `exec`. And on recent Pythons, `subprocess` uses `posix_spawn`. On Linux this might use `vfork` (I'm not actually sure?). And on macOS it uses a native `posix_spawn` syscall, so who knows what that does. Again, this *might* not be a big deal... maybe the parent gets to go again the instant the child calls `exec`, or sooner, without having to wait for any disk access or anything. But I'm not sure!

So... we should figure this out. Because if process creation is slow enough that we need to treat it as a blocking operation, we might need to change the process API to give it an async constructor. (Presumably by making `Process.__init__` private, and adding `await trio.open_process(...)` – similar to how we handle files.)

njsmith:
It's easy to find benchmarks of how long it takes a Windows process to start up (i.e., from `CreateProcess` start → the process actually doing things). It's harder to find benchmarks of `CreateProcess`'s "return latency" (from `CreateProcess` start → `CreateProcess` returns in the parent, even if the child is still setting up). But here's a report of a `CreateProcess` call that occasionally takes 700 ms for no clear reason: https://social.msdn.microsoft.com/Forums/vstudio/en-US/80bcb045-c5ee-4514-b93d-694296747f91/function-createprocess-takes-almost-a-second-to-run?forum=windowsgeneraldevelopmentissues

@eryksun Sorry to bother, but you seem to know all kinds of mysterious things about Windows internals. Do you happen to know at what point during process startup `CreateProcess` returns?

asvetlov:
Interesting observation.
[...]

PR #1113, opened 2019-06-17
Add trio.open_process, and deprecate Process(...)

njsmith:
Fixes gh-1109

### DIFF
--- a/docs/source/reference-io.rst
+++ b/docs/source/reference-io.rst
@@ -662,12 +662,12 @@ for them to exit. The interface for doing so consists of two layers:
   the standard :func:`subprocess.run` with some additional features
   and safer defaults.
 
-* :class:`trio.Process` starts a process in the background and optionally
-  provides Trio streams for interacting with it (sending input,
-  receiving output and errors). Using it requires a bit more code
-  than :func:`~trio.run_process`, but exposes additional capabilities:
-  back-and-forth communication, processing output as soon as it is generated,
-  and so forth. It is modelled after the standard :class:`subprocess.Popen`.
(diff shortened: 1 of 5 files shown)


njsmith:
Oh bleh, Windows wants to do async stuff in `Process.__init__`. So instead of shoving the whole thing off into a thread we have to shove just the `Popen` call into a thread and move the rest into open_process.

njsmith:
...which is actually pretty non-trivial, if we want to share the rest of the setup code between `open_process` and the legacy `__init__` code during the deprecation period. I'm going to go to bed and see if I have any better ideas in the morning...

oremanj:
What's the additional async stuff that Windows needs in Process.__init__? Is it the creation of the named pipes?

njsmith:
yeah, and specifically registering them with IOCP, which requires finding the io manager's state

oremanj:
oh I see, not so much "wants to do blocking stuff" as "wants to do stuff in the Trio thread". Yeah that's a tough one :-( Will mull it over some.

njsmith:
Eh, no brilliant ideas and it's not really worth fussing over for some code we're going to delete after the next release anyway.

For now I scaled back the PR to only making the interface changes, without the actual switch to using a thread, and then in a few months after `Process(...)` is gone we can fix the actual issue.

oremanj:
I'd love to be able to represent `async with await ...` as just `async with ...`; asyncssh has a decorator that does this. It would probably require us to stop pretending that the awaitable protocol doesn't exist, though, and is in any event a separate change. This looks good; thanks for making the change!

Issue #1092, opened 2019-06-09
Should the Final and NoPublicConstructor classes inherit from typing.GenericMeta?


njsmith:
We have some metaclasses now to enforce restrictions on subclassing and constructors (#1043), and we should be using them more (#1044). One problem we're going to run into:

```
In [3]: class Blah(trio.abc.Stream, metaclass=trio._util.Final): 
   ...:     pass 
   ...:                                                                         
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

The problem is that:

- if you subclass any ABC (like `Stream`), then your metaclass must subclass `abc.ABCMeta`
- if subclass any generic class (in the `typing.Generic` sense – like our channel ABCs), then your metaclass must subclass `typing.GenericMeta`

Fortunately `GenericMeta` subclasses `ABCMeta`, so at least these don't conflict with each other, but it's still a pain.

I think maybe we can just make `Final` subclass `GenericMeta` and be done with it? It's a little awkward because it means we're dragging in the code for generic and abstract classes even when we don't need it, but in practice I don't *think* this affects much? Both metaclasses are prepared to cope with situations where the generic and abstract features are not actually used. The extra metaclasses probably slow down the initial class creation at import time slightly, but I don't think they affect any common operations on class *instances*.

njsmith:
> I think maybe we can just make Final subclass GenericMeta and be done with it? It's a little awkward because it means we're dragging in the code for generic and abstract classes even when we don't need it, but in practice I don't think this affects much? Both metaclasses are prepared to cope with situations where the generic and abstract features are not actually used.

I guess I was over-optimistic. If we make `NoPublicConstructor` inherit from `ABCMeta`, then `isinstance(exc, Cancelled)` fails:

```
  File "/home/njs/trio/trio/_core/_run.py", line 386, in _exc_filter
    if isinstance(exc, Cancelled):
  File "/home/njs/.user-python3.7/lib/python3.7/abc.py", line 139, in __instancecheck__
    return _abc_instancecheck(cls, instance)
AttributeError: type object 'Cancelled' has no attribute '_abc_impl'
```

njsmith:
False alarm – the problem was that `Final.__new__` was calling `type.__new__` instead of `ABCMeta.__new__`. Switching it to use `super().__new__` fixed the issue above.

Issue #719, opened 2018-10-05
Fine-tuning channels


njsmith:
Channels (#586, #497) are a pretty complicated design problem, and fairly central to user experience, so while I think our first cut is pretty decent, we'll likely want to fine tune it as we get experience using them.

Here are my initial notes (basically questions that I decided to defer while implementing version 1):

In `open_memory_channel`, should we make `max_buffer_size=0` default? Is this really a good choice across a broad range of settings? The docs recommend it right now, but that's not really based on lots of practical experience or anything. And it's way easier to add defaults later than it is to remove or change them.

For channels that allow buffering, it's theoretically possible for values to get "lost" without there ever being an exception (`send` puts it in the buffer, then receiver never takes it out). This even happens if both sides are being careful, and closing their endpoint objects when they're done with them. We could potentially make it so that `ReceiveChannel.aclose` raises an exception (I guess `BrokenResourceError`) if the channel wasn't already cleanly closed by the sender. This would be inconsistent with the precedent set by `ReceiveStream` (which inherited it from BSD sockets), but OTOH, `ReceiveStream` is a much lower-level API – to use a `ReceiveStream` you need some non-trivial engineering work to implement a protocol on top, while `ReceiveChannel` is supposed to be a simple high-level protocol that's usable out-of-the-box. Practically speaking, though, it would be pretty annoying if some code inside a consumer crashes, and then `ReceiveChannel.__aexit__` throws away the actual exception and replaces it with `BrokenChannelError`.

Should memory channels have a sync `close()` method? They could easily, they only reason not to is that it makes life simpler for hypothetical IPC channels. I have no idea yet how common those will be, or whether it's actually important for in-process channels and cross-process channels to have compatible APIs.

[...]

PR #1115, opened 2019-06-19
Refactor channel interfaces

njsmith:
Make MemorySendChannel and MemoryReceiveChannel into public classes,
and move a number of the methods that used to be on the abstract
SendChannel/ReceiveChannel interfaces into the Memory*Channel concrete
classes. Also add a Channel type, analogous to Stream.

See: gh-719

Still to do:

- Merge MemorySendChannel and MemoryReceiveChannel into a single
  MemoryChannel
- decide what to do about clone
- decide whether to add some kind of half-close on Channel
- ~~refactor this PR's one-off solution to gh-1092 into something more
  general.~~ done

### DIFF
--- a/trio/_abc.py
+++ b/trio/_abc.py
@@ -484,15 +484,18 @@ class HalfCloseableStream(Stream):
         """
 
 
+# A regular invariant generic type
+T = TypeVar("T")
+
 # The type of object produced by a ReceiveChannel (covariant because
 # ReceiveChannel[Derived] can be passed to someone expecting
 # ReceiveChannel[Base])
-T_co = TypeVar("T_co", covariant=True)
+ReceiveType = TypeVar("ReceiveType", covariant=True)
 
 # The type of object accepted by a SendChannel (contravariant because
 # SendChannel[Base] can be passed to someone expecting
 # SendChannel[Derived])
-T_contra = TypeVar("T_contra", contravariant=True)
+SendType = TypeVar("SendType", contravariant=True)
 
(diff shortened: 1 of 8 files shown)


trio/_channel.py
```diff
@@ -141,6 +147,9 @@ def send_nowait(self, value):
 
     @enable_ki_protection
     async def send(self, value):
+        """See `~trio.abc.SendChannel.send`.
```
oremanj: This will format as "See send()", won't it? I might remove the tilde so it looks less recursive.

trio/_abc.py
```diff
@@ -621,45 +581,22 @@ class ReceiveChannel(AsyncResource, Generic[T_co]):
...
-              channel is broken.
-
-        """
-
-    @abstractmethod
-    async def receive(self):
+    async def receive(self) -> ReceiveType:
         """Attempt to receive an incoming object, blocking if necessary.
 
-        It's legal for multiple tasks to call :meth:`receive` at the same
```
oremanj: It seems like it would be useful to explicitly document what is and is not guaranteed about tasks calling send() or receive() simultaneously for all Channels. Also maybe something about cancellation safety. If you want to stop guaranteeing that multiple simultaneous receive() works OK, maybe move this text to MemoryReceiveChannel.receive() instead of removing it entirely?

njsmith: Hmm, yeah. I'm thinking ahead to "framing channels" that just wrap line-breaking or whatever around a stream, so they'll have the same single-user rule as streams. But that's a good point.

Issue #878, opened 2019-01-25
Drop the distinction between wait_readable / wait_socket_readable


njsmith:
As discussed in #400, and specifically [this comment](https://github.com/python-trio/trio/issues/400#issuecomment-357478182), we should drop the distinction between `wait_readable` and `wait_socket_readable` (and similar for `writable`, `notify_close`). Instead, we should just document that on Unix they take file descriptors, and on Windows they take socket handles (not file descriptors).

(Should they also accept other handles? See also: https://github.com/python-trio/trio/issues/661#issuecomment-457502687)

PR #1119, opened 2019-06-20
Standardize the names of wait_* and notify_closing across platforms

njsmith:
Fixes gh-878


njsmith:
The coverage drop appears to be because some version of CPython/coverage/something has started reporting `if False: ...` as uncovered code, which didn't use to happen. I guess we could spend a bunch of time tracking down which versions and why, but probably it's easier to just add `# pragma: no cover` to all our `if False: ...` branches. In any case, it has nothing to do with this PR.

oremanj:
Thanks!

Issue #810, opened 2018-12-19
Rename run_sync_in_worker_thread


njsmith:
I was ok with this having a slightly-awkward name because threads are going to make your life awkward so, you know, it's fair warning. But when you're stuck using threads to hack around missing trio libraries, it can feel like a bit of extra punishment on top, which is not so nice.

Maybe we should rename it to `run_sync_in_thread`?

njsmith:
Well, that seems popular.

Related idea: the whole "portal" design seems a bit more awkward than I was anticipating. I was imagining that threads were a mostly-for-experts kind of thing, and "foreign" threads would be as common as trio-spawned threads, but I feel like maybe it's not turning out that way.

Alternative idea (inspired by curio and anyio): provide a standard way to re-enter trio from a trio thread. Basically stash the token in thread-local storage, and then use it.

One possible API:

```python
trio.run_from_thread(async_fn, *args, *, token=None)
trio.run_sync_from_thread(fn, *args, *, token=None)
trio.sync_cm_from_thread(cm, *, token=None)
trio.async_cm_from_thread(cm, *, token=None)

trio.run_sync_in_thread(fn, *args, *, ...)
trio.sync_cm_in_thread(cm, *, ...)
```

So the general idea is that `from_thread` is for going thread->trio, and `in_thread` is for going trio->thread. (It's taken for granted that one end is trio, because of course trio is the center of the universe. Also it's in the name of the functions.)

This pattern might be used for other projects too, like trio-asyncio – `run_in_aio`, `run_from_aio`. trio-asyncio has really struggled with making these names easy to understand – @smurfix, do you think this would do any better?

Alternatively, we could group these into a namespace, like:

```python
trio.from_thread.run(...)
trio.from_thread.run_sync(...)
trio.from_thread.sync_cm(...)
trio.from_thread.async_cm(...)

trio.in_thread.run_sync(...)
trio.in_thread.sync_cm(...)
```

This would group related functions together for tab-completion, and avoid "polluting" the main trio namespace. OTOH it looks a bit weird, and flat is better than nested...

Another alternative, or complement: we could do `async with trio.in_thread: ...`/`async with trio.from_thread` – see https://github.com/python-trio/trio-asyncio/issues/42 for a discussion of why this is nice – in particular it automatically works for context managers.

See also: #680 for context managers across threads, #606 for propagating cancellation across trio->thread->trio transitions, and #648 for preserving contextvars across thread switches

smurfix:
```
[...]

PR #1122, opened 2019-06-23
Deprecate BlockingTrioPortal in favor of from_thread.run and from_thread.run_sync

epellis:
This addressses issue #810 by implementing `trio.from_thread.run` and  `trio.from_thread.run_sync` and deprecating `BlockingTrioPortal`. There is probably some more stuff we need to add before this PR is fully complete, like integrating #1098 and the `run_sync_soon` API, but this is a good start.

Support re-entering the same `Trio.run()` loop by stashing the current
Trio Token in Thread Local Storage right before `run_sync_in_thread`
spawns a worker thread. When `trio.from_thread.run_sync` or
`trio.from_thread.run` are called in this thread, they can access the
token and use it to re-enter the `Trio.run()` loop.

This commit deprecates `BlockingTrioPortal`. For the majority of
use cases, using the thread local Trio Token should be sufficient. If
for any reason, another Trio Token needs to be used, it can be
passed as a kwarg to `from_thread.run` and `from_thread.run_sync`.

Here is a snippet from how the new API works:
```python3
import trio

def thread_fn():
    start = trio.from_thread.run_sync(trio.current_time)
    print("In Trio-land, the time is now:", start)
    trio.from_thread.run(trio.sleep, 1)
    end = trio.from_thread.run_sync(trio.current_time)
    print("And now it's:", end)

async def main():
    await trio.run_sync_in_thread(thread_fn)

trio.run(main)
```

Here is how the same code works in "old" Trio:
```python3
import trio

def thread_fn(token):
    portal = trio.BlockingTrioPortal(token)
    start = portal.run_sync(trio.current_time)
    print("In Trio-land, the time is now:", start)
    portal.run(trio.sleep, 1)
    end = portal.run_sync(trio.current_time)
    print("And now it's:", end)

async def main():
    token = trio.hazmat.current_trio_token()
    await trio.run_sync_in_thread(thread_fn, token)

trio.run(main)
```


trio/tests/test_threads.py
```diff
@@ -457,3 +458,88 @@ def bad_start(self):
...
+async def test_trio_from_thread_run_sync():
+    # Test that run_sync_in_thread correctly "hands off" the trio token to
+    # trio.from_thread.run_sync()
+    def thread_fn():
+        start = run_sync(_core.current_time)
+        end = run_sync(_core.current_time)
+        return end - start
+
+    duration = await run_sync_in_thread(thread_fn)
+    assert duration > 0
```
njsmith: This is more complicated than it needs to be... just checking that `run_sync(_core.current_time)` returns a `float` is probably good enough.

trio/tests/test_threads.py
```diff
@@ -457,3 +458,88 @@ def bad_start(self):
...
+    # Test that run_sync_in_thread correctly "hands off" the trio token to
+    # trio.from_thread.run()
+    def thread_fn():
+        start = time.perf_counter()
+        run(sleep, 0.05)
+        end = time.perf_counter()
+        return end - start
+
+    duration = await run_sync_in_thread(thread_fn)
+    assert duration > 0
```
njsmith: We don't `sleep` in Trio's tests unless we absolutely can't help it... 50 milliseconds is a pretty short sleep, so not really an issue, but since our only goal here is to check that the code runs, we might as well do something that doesn't involve sleeping at all. How about:

```python
record = []

async def back_in_trio_fn():
    trio.current_time()  # implicitly checks that we're in trio
    record.append("back in trio")

def thread_fn():
    record.append("in thread")
    run(back_in_trio_fn)

await run_sync_in_thread(thread_fn)
assert record == ["in thread", "back in trio"]
```

trio/tests/test_threads.py
```diff
@@ -457,3 +458,88 @@ def bad_start(self):
...
+
+
+async def test_trio_from_thread_raw_call():
+    # Test that a "raw call" to trio.from_thread.run() fails because no token
+    # has been provided
+
+    with pytest.raises(AttributeError):
+        run_sync(_core.current_time)
```
njsmith: Two things about this test:

- There are two reasons this might error out... because the thread local isn't set, or because we're trying to invoke it from inside trio. I think we're trying to test what happens when the thread-local isn't set. To make that unambiguous, I'd suggest making it a regular synchronous test (so it's just `def test_from_thread_no_token()` or similar), so it doesn't run inside trio.

- When someone does try to use this from a non-trio thread, maybe we should give a nice error message instead of just letting the `AttributeError` escape? Like `raise RuntimeError("this thread wasn't created by Trio, so you have to pass an explicit trio_token=...")`.

trio/tests/test_threads.py
```diff
@@ -457,3 +458,88 @@ def bad_start(self):
...
+    # the same thread together
+
+    def thread_fn():
+        start = run_sync(_core.current_time)
+        run(sleep, 0.05)
+        end = run_sync(_core.current_time)
+        return end - start
+
+    duration = await run_sync_in_thread(thread_fn)
+    assert duration > 0
```
njsmith: Another `sleep`... and I'm not sure this test really adds anything? We've already tested `run` and `run_sync` separately above, and I don't know any way we could mess things up so that they work separately but not together...

trio/_threads.py
```diff
@@ -385,6 +391,7 @@ def worker_thread_fn():
         thread = threading.Thread(
             target=worker_thread_fn, name=name, daemon=True
         )
+        setattr(thread, 'current_trio_token', trio.hazmat.current_trio_token())
```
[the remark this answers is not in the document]
smurfix: Umm, in what way is ``setattr(thread, 'current_trio_token', trio.hazmat.current_trio_token())`` any different from ``thread.current_trio_token = trio.hazmat.current_trio_token()``? I strongly suspect that it's not, thus the ``setattr`` way is needlessly verbose.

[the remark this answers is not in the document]
epellis: Good catch, I forgot that the dot operator worked like that for most attributes!

trio/tests/test_threads.py
epellis: You make a good point, I went ahead and removed this test.

trio/tests/test_threads.py
epellis: Good idea, `AttributeError` is quite vague and I think this kind of exception will happen frequently enough that a well worded error message will ease a lot of debugging pain!

trio/_threads.py
```diff
@@ -269,6 +269,12 @@ class ThreadPlaceholder:
     tasks to continue working while ``sync_fn`` runs. This is accomplished by
     pushing the call to ``sync_fn(*args)`` off into a worker thread.
 
+    ``run_sync_in_thread`` also injects the current ``TrioToken`` into the
+    spawned thread's local storage so that these threads can re-enter the Trio
+    loop by calling either :func: ``trio.from_thread.run`` or
+    :func: ``trio.from_thread.run_sync`` for async or synchronous functions,
```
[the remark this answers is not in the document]
epellis: Thanks for the hint! Yeah, I have never used RST, only Markdown, and also never Sphinx either so the docs are a bit of a learning curve for me.

epellis:
Okay, hello again! I have fixed up all of the issues mentioned in the last round of the PR and have also made sure all of the documentation and code coverage is at par with or better than where it was before.

epellis:
@njsmith mind giving any feedback on this? I think I made all of the requested changes but want to confirm before I merge this.

docs/source/reference-core/from-thread-example.py
```diff
@@ -1,29 +1,29 @@
 import trio
 
-def thread_fn(portal, receive_from_trio, send_to_trio):
+
+def thread_fn(receive_from_trio, send_to_trio):
     while True:
         # Since we're in a thread, we can't call methods on Trio
         # objects directly -- so we use our portal to call them.
```
njsmith: Leftover reference to a "portal" here.

trio/_threads.py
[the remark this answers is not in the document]
njsmith: It looks like these double backticks still need to be replaced with single backticks.

trio/_threads.py
```diff
@@ -369,21 +285,30 @@ def do_release_then_return_result():
...
+        TOKEN_LOCAL.token = trio_token
         result = outcome.capture(sync_fn, *args)
         try:
             token.run_sync_soon(report_back_in_trio_thread_fn, result)
         except trio.RunFinishedError:
             # The entire run finished, so our particular task is certainly
             # long gone -- it must have cancelled.
             pass
+        finally:
+            del TOKEN_LOCAL.token
```
njsmith: This is a finicky style comment; I think the code you have here works correctly. But the `try` block is doing two logically unrelated things here – cleaning up the thread local, and dealing with a missing trio thread. And the thread local cleanup logically has a larger scope – if `capture` raised an exception, we'd still want to clean up the thread local. Fortunately `capture` can't raise an exception, so we could get away with this, but for readability, I think it will be clearer what's happening if we split it into two try blocks, structured like:

```python
TOKEN_LOCAL = ...
try:
    result = ...
    try:
        token.run_sync_soon(...)
    except RunFinishedError:
        ...
finally:
    ...
```

trio/_threads.py
```diff
@@ -8,119 +8,29 @@
 import trio
 
 from ._sync import CapacityLimiter
-from ._core import enable_ki_protection, disable_ki_protection, RunVar
+from ._core import enable_ki_protection, disable_ki_protection, RunVar, _entry_queue
```
njsmith: There's a weird rule we follow: any code outside of the `_core/` directory is not allowed to "reach inside" the `_core` namespace to access private attributes. (Meaning, attributes with leading underscores.)

The rationale is: we want to make sure that if someone needs to implement something like `from_thread.*` in a third party library on top of trio, then that should be possible without using any private APIs. Everything in `_core`'s public API is re-exported as part of `trio`'s public API, so if we can implement our own stuff using only `_core`'s public API, it means other people will be able to do the same. There's a bit more discussion here: https://trio.readthedocs.io/en/latest/design.html#a-brief-tour-of-trio-s-internals

In this case, that means we can't import `_entry_queue`. But, it looks like the only thing we use it for is to look up the `TrioToken` class object, and that's already exported from `_core`, so we can just do `from ._core import TrioToken` directly.

trio/_threads.py
epellis: Hmm that is an interesting way to look at it, I never thought about the separation of concerns that way!

epellis:
@njsmith Got some more time to polish this up, can you let me know if the latest round of changes are up to spec? Thanks as always!

njsmith:
@epellis Looks great, thanks! BTW, I think github is listing your last few commits as coming from your work account – doesn't matter to me, but FYI in case you wanted to keep those separate.

Any thoughts on what you want to work on next?

epellis:
@njsmith Thanks for letting me know, I thought I signed out of that account on this laptop but I guess my git cache says otherwise 🤷‍♂. I'd love to continue working with the scheduler and threading/async interop so I'll take stock of how the rest of #1085 and friends are doing and try to comment there in the next week or so with chunks I think I could manage. Thanks again for your feedback, this PR was a really enjoyable process! :)

Issue #959, opened 2019-03-01
Should we merge the Stream and Channel interfaces?


njsmith:
This is a weird/radical idea, but @oremanj's comment [here](https://github.com/python-trio/trio/pull/872#discussion_r261459795) (in response to @Badg) raised some red flags for me:

> And the channel interface is nicer than the stream one for incremental processing -- `async for chunk in channel:` rather than
> 
> ```python
> while True:
>     chunk = await stream.receive_some(ARBITRARILY_CHOSEN_POWER_OF_TWO)
>     if not chunk:
>         break
>     ...
> ```

At the conceptual level, the output from a process is *exactly* a `Stream` (or `ReceiveStream` or whatever), and if people are trying to jump through hoops to avoid using our `Stream` ABC to represent the Stream concept, then that seems like a bad sign!

So, let's at least go through the thought experiment: what if we got rid of `Stream` and used `Channel[bytes]` instead?

# Basic usability

Remembering which is a "stream" and which is a "channel" is [super annoying](https://github.com/python-trio/trio/issues/719#issuecomment-453334770). Merging them would eliminate this problem. Also annoying: constantly going through the pointless ritual of inventing a made-up buffer size (@oremanj's `ARBITRARILY_CHOSEN_POWER_OF_TWO`). And writing that `while True` loop over and over is also annoying. Making it a plain `await channel.receive()` or `async for chunk in channel:` would eliminate these annoyances.

# Conceptual level

For me, the major conceptual difference is that I've thought of `Channel` as inherently preserving object boundaries, i.e., whatever I pass to `send` is what comes out of `receive`. In this way of thinking, a `Stream` is equivalent to a `Channel[single_byte]`, but since handling single bytes individual would be inefficient, it uses batched-send and batched-receive operations. If we do decide to merge `Stream` and `Channel`, then we'd have to change this, and start saying that some `Channel`s don't preserve the 1-to-1 mapping between `send` and `receive`.

[...]

PR #1123, opened 2019-06-25
Streams are iterable + receive_some doesn't require an explicit size

njsmith:
This came out of discussion in gh-959

### DIFF
--- a/docs/source/tutorial/echo-server.py
+++ b/docs/source/tutorial/echo-server.py
@@ -21,7 +21,6 @@ async def echo_server(server_stream):
             print("echo_server {}: received data {!r}".format(ident, data))
             await server_stream.send_all(data)
         print("echo_server {}: connection closed".format(ident))
-        return
     # FIXME: add discussion of MultiErrors to the tutorial, and use
     # MultiError.catch here. (Not important in this case, but important if the
     # server code uses nurseries internally.)
--- a/docs/source/tutorial/echo-server.py
+++ b/docs/source/tutorial/echo-server.py
@@ -8,9 +8,6 @@ from itertools import count
 # - can't be in use by some other program on your computer
 # - must match what we set in our echo client
 PORT = 12345
-# How much memory to spend (at most) on each call to recv. Pretty arbitrary,
-# but shouldn't be too big or too small.
-BUFSIZE = 16384
 
 CONNECTION_COUNTER = count()
 
@@ -20,14 +17,11 @@ async def echo_server(server_stream):
     ident = next(CONNECTION_COUNTER)
     print("echo_server {}: started".format(ident))
     try:
-        while True:
-            data = await server_stream.receive_some(BUFSIZE)
+        async for data in server_stream:
             print("echo_server {}: received data {!r}".format(ident, data))
-            if not data:
-                print("echo_server {}: connection closed".format(ident))
-                return
-            print("echo_server {}: sending data {!r}".format(ident, data))
             await server_stream.send_all(data)
+        print("echo_server {}: connection closed".format(ident))
+        return
     # FIXME: add discussion of MultiErrors to the tutorial, and use
     # MultiError.catch here. (Not important in this case, but important if the
     # server code uses nurseries internally.)
--- a/newsfragments/959.feature.rst
+++ b/newsfragments/959.feature.rst
@@ -1,0 +1,8 @@
+If you have a `~trio.abc.ReceiveStream` object, you can now use
+``async for data in stream: ...`` instead of calling
+`~trio.abc.ReceiveStream.receive_some` repeatedly. And the best part
+is, it automatically checks for EOF for you, so you don't have to.
+Also, you no longer have to choose a magic buffer size value before
+calling `~trio.abc.ReceiveStream.receive_some`; you can now call
+``await stream.receive_some()`` and the stream will automatically pick
+a reasonable value for you.
--- a/newsfragments/959.feature.rst
+++ b/newsfragments/959.feature.rst
@@ -1,8 +1,10 @@
 If you have a `~trio.abc.ReceiveStream` object, you can now use
 ``async for data in stream: ...`` instead of calling
-`~trio.abc.ReceiveStream.receive_some` repeatedly. And the best part
-is, it automatically checks for EOF for you, so you don't have to.
-Also, you no longer have to choose a magic buffer size value before
-calling `~trio.abc.ReceiveStream.receive_some`; you can now call
-``await stream.receive_some()`` and the stream will automatically pick
-a reasonable value for you.
+`~trio.abc.ReceiveStream.receive_some`. Each iteration gives an
+arbitrary sized chunk of bytes. And the best part is, the loop
+automatically exits when you reach EOF, so you don't have to check for
+it yourself anymore. Relatedly, you no longer need to pick a magic
+buffer size value before calling
+`~trio.abc.ReceiveStream.receive_some`; you can ``await
+stream.receive_some()`` with no arguments, and the stream will
+automatically pick a reasonable size for you.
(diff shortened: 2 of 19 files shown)


njsmith:
Codecov doesn't seem to be updating its comment, but if I click through now, it says that this isn't adding any new uncovered lines.

docs/source/tutorial/echo-server.py
```diff
@@ -20,14 +17,11 @@
...
-            data = await server_stream.receive_some(BUFSIZE)
+        async for data in server_stream:
             print("echo_server {}: received data {!r}".format(ident, data))
-            if not data:
-                print("echo_server {}: connection closed".format(ident))
-                return
-            print("echo_server {}: sending data {!r}".format(ident, data))
             await server_stream.send_all(data)
+        print("echo_server {}: connection closed".format(ident))
+        return
```
oremanj: I think this `return` is not necessary anymore?

newsfragments/959.feature.rst
```diff
@@ -0,0 +1,8 @@
+If you have a `~trio.abc.ReceiveStream` object, you can now use
+``async for data in stream: ...`` instead of calling
+`~trio.abc.ReceiveStream.receive_some` repeatedly. And the best part
+is, it automatically checks for EOF for you, so you don't have to.
```
oremanj: I might rephrase this ("And the best part is ...") as something like "Receiving EOF will cleanly terminate the `async for` loop, so you don't have to check for it explicitly."

newsfragments/959.feature.rst
```diff
@@ -0,0 +1,8 @@
...
+Also, you no longer have to choose a magic buffer size value before
+calling `~trio.abc.ReceiveStream.receive_some`; you can now call
+``await stream.receive_some()`` and the stream will automatically pick
```
oremanj: maybe add "with no arguments" after the example call to make it really obvious

trio/_abc.py
```diff
@@ -378,26 +378,26 @@ class ReceiveStream(AsyncResource):
     If you want to receive Python objects rather than raw bytes, see
     :class:`ReceiveChannel`.
 
+    `ReceiveStream` objects can be used in ``async for`` loops. Each iteration
+    will produce an arbitrary size
```
oremanj: finish sentence?

trio/_highlevel_socket.py
```diff
@@ -10,6 +10,12 @@
 
 __all__ = ["SocketStream", "SocketListener"]
 
+# XX TODO: this number was picked arbitrarily. We should do experiments to
```
oremanj: One wrinkle: AFAIK, each call to `socket.recv()` allocates a new `bytes` object that is large enough for the entire given chunksize. If large allocations are more expensive, passing a too-large buffer is probably bad for performance. (The allocators I know of use 128KB as their threshold for "this is big, mmap it instead of finding a free chunk" but if one used 64KB instead and we got a mmap/munmap pair on each receive, that feels maybe bad?)

My intuition favors a much lower buffer size, like 4KB or 8KB, but I also do most of my work on systems that are rarely backlogged, so my intuition might well be off when it comes to a high-throughput Trio application.

Another option we could consider: the socket owns a receive buffer (bytearray) which it reuses, calls `recv_into()`, and extracts just the amount actually received into a `bytes` for returning. Downside: spends 64KB (or whatever) per socket in steady state. Counterpoint: the OS-level socket buffers are probably much larger than that (but I don't know how much memory they occupy when the socket isn't backlogged).

trio/_ssl.py
```diff
@@ -621,9 +616,15 @@ def _check_status(self):
...
                 else:
                     raise
-            max_bytes = _operator.index(max_bytes)
-            if max_bytes < 1:
-                raise ValueError("max_bytes must be >= 1")
+            if max_bytes is None:
+                # Heuristic: normally we use DEFAULT_RECEIVE_SIZE, but if
+                # the transport gave us a bunch of data last time then we'll
+                # try to decrypt and pass it all back at once.
+                max_bytes = max(DEFAULT_RECEIVE_SIZE, self._incoming.pending)
```
oremanj: I'm a little confused at what the benefit is of having a `DEFAULT_RECEIVE_SIZE` for SSLStream at all. It seems like we could instead have a nice magic-number-free policy of "ask the transport stream to receive_some() with no size specified, then return all the data we decrypted from whatever we got in that chunk, or loop and receive_some() again if we didn't get any decrypted data".

trio/tests/test_ssl.py
```diff
@@ -1232,17 +1234,25 @@ def close_hook():
 
     ################
 
-    # Test https_compatible and max_refill_bytes
-    _, ssl_listener, ssl_client = await setup(
-        https_compatible=True,
-        max_refill_bytes=100,
-    )
+    # Test https_compatible
+    _, ssl_listener, ssl_client = await setup(https_compatible=True,)
```
oremanj: trailing comma probably superfluous?

trio/_highlevel_socket.py
njsmith: This is an interesting discussion but I don't want it to hold up merging the basic functionality, so I split it off into https://github.com/python-trio/trio/issues/1139

(Twisted has apparently used 64 KiB receive buffers for its entire existence and I can't find any evidence that anyone has ever thought twice about it. So we're probably not risking any disaster by starting with 64 KiB for now :-).)

trio/_ssl.py
njsmith: Yeah, this is complicated... what you say makes logical sense, but, openssl's API is super awkward. There isn't any way to say "please decrypt all the data in your receive buffer". You have to pick a value to pass to `SSLObject.read`. And even more annoying: you don't find out until after you've picked a value whether you have to go back to the underlying transport for more data. So you have to pick the value before you know how much data the underlying transport wants to give you. And once you've picked a value, you have to keep using that value until some data is returned.

So my logic was: well, if we already have a bunch of data in the receive buffer because the underlying transport was generous, then likely we can just decrypt and return that, and the size of the encrypted data is a plausible upper bound on the size of the decrypted data, so `self._incoming.pending` is a good value to pass to `SSLObject.read`.

But, sometimes there won't be a lot of data in the receive buffer – for example, because our heuristic worked well the previous time, and cleared everything out, or almost everything. Like, imagine there's 1 byte left in the receive buffer. The way TLS works, you generally can't decrypt just 1 byte – everything's transmitted in frames, and you need to get the whole frame with its header and MAC and everything before you can decrypt any of it. So if we call `ssl_object.read(1)`, then openssl will end up requesting another large chunk of data from the underlying transport, then our `read(1)` call will decrypt the first byte and return it, leaving the rest of the data sitting in the buffer for next time. And that would be unfortunate.

So my first attempt at a heuristic is: use the receive buffer size, but never anything smaller than `DEFAULT_RECEIVE_SIZE`.

I guess this has a weird effect if the underlying transport likes to return more than `DEFAULT_RECEIVE_SIZE`. Say it gives us 65 KiB, while `DEFAULT_RECEIVE_SIZE` is 64 KiB. On our first call to `SSLStream.receive_some`, the buffer size is zero, so we do `read(64 KiB)`. This drains 65 KiB from the underlying transport, then decrypts and returns the first 64 KiB. The next time we call `SSLStream.receive_some`, we do `read(64 KiB)` again, but there's already 1 KiB of data in the buffer, so we just return that immediately without refilling the buffer. Then this repeats indefinitely, so we alternate between doing a big receive and a small receive every time. Seems wasteful – it'd be better to return 65 KiB each time.

So maybe a better strategy would be to start with some smallish default receive size, and then increase it over time if we observe the underlying transport giving us more data.

njsmith: I rewrote the `SSLStream` stuff to hopefully address the above issues...

PR #1129, opened 2019-06-27
add hazmat.FdStream

### LINKED ISSUE #829: Should we somehow expose Pipe*Stream on Unix, and if so, how?
We have `PipeReceiveStream` and `PipeSendStream` classes on Unix, that are currently not exposed publicly; they're only used internally for subprocesses.

The actual code is a pretty generic building block: they can wrap around any Unix file-descriptor that supports non-blocking mode. So they could potentially be useful for talking to [FIFOs](http://man7.org/linux/man-pages/man7/fifo.7.html), TTYs, stdio streams (#174), etc.

Should we expose them more directly? If so, how? In `trio.hazmat`, with some sort of [`os.fdopen`](https://docs.python.org/3/library/os.html#os.fdopen) equivalent?

If we do we'll want to think through the ownership semantics: should they take ownership of a passed in fd, or should they dup it?

Also, they should restore the fd's blocking/non-blocking state before closing it.

belm0:
Closes #829.

TODO:
 * [x] port PipeSend/ReceiveStream tests to FdStream
 * [x] update docs

### DIFF
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -174,6 +174,40 @@ All environments provide the following functions:
    yourself afterwards.
 
 
+Unix-specific API
+-----------------
+
+`FdStream` supports wrapping Unix files (such as a pipe or TTY) as
+a stream.
+
+To be used as a Trio stream, an open file must be placed in non-blocking
+mode.  Unfortunately, this impacts all I/O that goes through the
+underlying open file description, including I/O that uses a different
+file descriptor than the one that was passed to Trio. If other threads
+or processes are using file descriptors that are related through `os.dup`
+or inheritance across `os.fork` to the one that Trio is using, they are
+unlikely to be prepared to have non-blocking I/O semantics suddenly
+thrust upon them.  For example, you can use ``FdStream(os.dup(0))`` to
+obtain a stream for reading from standard input, but it is only safe to
+do so with heavy caveats: your stdin must not be shared by any other
+processes and you must not make any calls to synchronous methods of
+`sys.stdin` until the stream returned by `FdStream` is closed. See
+`issue #174 <https://github.com/python-trio/trio/issues/174>`__ for a
+discussion of the challenges involved in relaxing this restriction.
+
+To obtain a bidirectional `~trio.abc.Stream` for sending and receiving on
+different file descriptors, use `trio.StapledStream`::
+
+    bidirectional_stream = trio.StapledStream(
+        trio.hazmat.FdStream(write_fd),
+        trio.hazmat.FdStream(read_fd)
+    )
+
+.. autoclass:: FdStream
+   :show-inheritance:
+   :members:
+
+
 Kqueue-specific API
 -------------------
 
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -180,8 +180,9 @@ Unix-specific API
 `FdStream` supports wrapping Unix files (such as a pipe or TTY) as
 a stream.
 
-To obtain a bidirectional `~trio.abc.Stream` for sending and receiving on
-different file descriptors, use `trio.StapledStream`::
+If you have two different file descriptors for sending and receiving,
+and want to bundle them together into a single bidirectional
+`~trio.abc.Stream`, then use `trio.StapledStream`::
 
     bidirectional_stream = trio.StapledStream(
         trio.hazmat.FdStream(write_fd),
--- a/docs/source/reference-hazmat.rst
+++ b/docs/source/reference-hazmat.rst
@@ -180,21 +180,6 @@ Unix-specific API
 `FdStream` supports wrapping Unix files (such as a pipe or TTY) as
 a stream.
 
-To be used as a Trio stream, an open file must be placed in non-blocking
-mode.  Unfortunately, this impacts all I/O that goes through the
-underlying open file description, including I/O that uses a different
-file descriptor than the one that was passed to Trio. If other threads
-or processes are using file descriptors that are related through `os.dup`
-or inheritance across `os.fork` to the one that Trio is using, they are
-unlikely to be prepared to have non-blocking I/O semantics suddenly
-thrust upon them.  For example, you can use ``FdStream(os.dup(0))`` to
-obtain a stream for reading from standard input, but it is only safe to
-do so with heavy caveats: your stdin must not be shared by any other
-processes and you must not make any calls to synchronous methods of
-`sys.stdin` until the stream returned by `FdStream` is closed. See
-`issue #174 <https://github.com/python-trio/trio/issues/174>`__ for a
-discussion of the challenges involved in relaxing this restriction.
-
 To obtain a bidirectional `~trio.abc.Stream` for sending and receiving on
 different file descriptors, use `trio.StapledStream`::
 
--- a/trio/_unix_pipes.py
+++ b/trio/_unix_pipes.py
@@ -33,9 +33,11 @@ class _FdHolder:
         if not isinstance(fd, int):
             raise TypeError("file descriptor must be an int")
         self.fd = fd
-        # Flip the fd to non-blocking mode
-        flags = fcntl.fcntl(self.fd, fcntl.F_GETFL)
-        fcntl.fcntl(self.fd, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+        # Store original state, and ensure non-blocking mode is enabled
+        self._original_flags = fcntl.fcntl(self.fd, fcntl.F_GETFL)
+        fcntl.fcntl(
+            self.fd, fcntl.F_SETFL, self._original_flags | os.O_NONBLOCK
+        )
 
     @property
     def closed(self):
@@ -53,6 +55,7 @@ class _FdHolder:
             return
         fd = self.fd
         self.fd = -1
+        fcntl.fcntl(fd, fcntl.F_SETFL, self._original_flags)
         os.close(fd)
 
     def __del__(self):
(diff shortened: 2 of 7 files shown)


trio/_unix_pipes.py
```diff
@@ -157,3 +160,25 @@ def __init__(self, fd: int):
 
     def fileno(self):
         return self._fd_holder.fd
+
+
+def fd_open_send_stream(fd: int):
```
oremanj: Name bikeshed: I feel like `open_` going at the beginning is maybe an important convention to maintain? Maybe `open_send_stream_to_fd` and `open_receive_stream_from_fd`, or even `open_stream_{to,from}_fd` if we think the send/receive-ness is obvious enough from the to/from-ness.

trio/_unix_pipes.py
```diff
@@ -157,3 +160,25 @@ def __init__(self, fd: int):
...
+    """Returns a send stream over an os.pipe.
+
+    The pipe must support non-blocking mode.  The stream takes ownership
```
oremanj: Elaborate what you mean here? Maybe
```
*fd* must refer to a file that is open for writing and supports non-blocking I/O (pipes and TTYs will
work, on-disk files probably not). The returned stream takes ownership of *fd*, so closing the stream
will close *fd* too. As with `os.fdopen`, you should not directly use *fd* after you have wrapped it
in a stream using this function.
```

njsmith:
I'm trying to think of the reasons why we use `open_*` functions:

- When we can't use `__init__` because we need a different calling convention (e.g. async `open_file`, or async context manager `open_nursery`)
- When we want to hide the actual type (`open_nursery`, the old `open_cancel_scope`) – but we're moving away from this (#778)
- When it's a convenience function that does some non-trivial setup (`open_tcp_stream`)

I don't think any of these apply here though? So we should probably expose the type(s) themselves instead of having `open_*` functions.

I also think it's awkward to have separate types for send and receive here... a generic fd supports both `read` and `write`, so it's weird to have to use `StapledStream` + `dup` to read-and-write to the same fd. Plus using `dup` here is a bit tricky, because we end up with two different objects trying to fiddle with the `fcntl` flags on the same underlying file description, so it increases the risk of them stomping on each other. And we don't have `SocketSendStream` and `SocketReceiveStream` either.

So I'm thinking maybe we should merge these together into a single `trio.hazmat.FdStream`.

This would mean that on Unix `Process.stdin` and friends are all `FdStream`s, even though they're unidirectional. But I think that's OK – if someone does accidentally try to call `stdin.receive_some` or `stdout.send_all` then at runtime that will raise an exception, and we'll statically declare those attributes to be a `SendStream` and `ReceiveStream` respectively, so mypy will also flag the error.

belm0:
> and we'll statically declare those attributes to be a SendStream and ReceiveStream respectively, so mypy will also flag the error.

I got lost here.  If `Process.stdin` is typed as a `FdStream`, how can it also be declared as a `SendStream`?

njsmith:
`FdStream` is a subtype of `SendStream`, so we can statically type `Process.stdin` as `ReceiveStream`, and then make it a `FdStream` at runtime. (And we certainly can't statically type it as `FdStream`, because it will be something else on Windows...)

belm0:
I've implemented `FdStream`.  Not quite sure how Process.stdio should be declared in the Unix case.

TODO:
 * [ ] port PipeSend/ReceiveStream tests to FdStream
 * [ ] Process.stdio
 * [ ] ?

belm0:
Aside:  I believe StapledStream `send_stream` and `receive_stream` attributes should be hidden.  Anyway changing Process.stdio to `FdStream` instance in the Unix case is going to break access to them.

njsmith:
I would expect that we'd have:

`Process.stdin`: typed as `SendStream`, on Unix actually an `FdStream`
`Process.stdout`, `Process.stderr`: typed as `ReceiveStream`, on Unix actually an `FdStream`
`Process.stdio`: `StapledStream(stdin, stdout)`

Note that on Unix we use pipes to talk to the child process, and pipes are unidirectional, so `stdio` has to be a `StapledStream`...

njsmith:
Oh I see, you're defining `FdStream` as managing two fds, so it kind of has the `StapledStream` functionality baked into it. I was suggesting making `FdStream` manage a single fd, similar to `SocketStream`.

belm0:
> Oh I see, you're defining `FdStream` as managing two fds

It did make it easy to ensure only one holder when receive and send streams were the same.  Any ideas for that given the single-fd FdStream?

njsmith:
If the send and receive streams are the same, then just... Use one `FdStream`? I don't think I'm understanding the question.

belm0:
OK-- I think the API is settled, so it's a matter of tests and docs.

I didn't do anything to preemptively check for accidental send_all / receive_some on the wrong mode at runtime.  I'm assuming it's fine to defer to OSError.

belm0:
another thing I wasn't quite sure about:  in the implementation I've used separate task conflict detectors for send & receive

NOTE: please squash on merge

trio/_unix_pipes.py
```diff
@@ -33,9 +33,11 @@ def __init__(self, fd: int):
...
             raise TypeError("file descriptor must be an int")
         self.fd = fd
-        # Flip the fd to non-blocking mode
-        flags = fcntl.fcntl(self.fd, fcntl.F_GETFL)
-        fcntl.fcntl(self.fd, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+        # Store original state, and ensure non-blocking mode is enabled
+        self._original_flags = fcntl.fcntl(self.fd, fcntl.F_GETFL)
+        fcntl.fcntl(
+            self.fd, fcntl.F_SETFL, self._original_flags | os.O_NONBLOCK
+        )
```
njsmith: Let's use `os.get_blocking` and `os.set_blocking`. Advantages:

- Less fcntl line noise
- In the unlikely event that someone else is fiddling with other unrelated fcntl flags while the `FdStream` is open, we won't inadverdantly overwrite their changes when we reset the blocking state

trio/_unix_pipes.py
```diff
@@ -53,6 +55,7 @@ def _raw_close(self):
             return
         fd = self.fd
         self.fd = -1
+        fcntl.fcntl(fd, fcntl.F_SETFL, self._original_flags)
```
njsmith: Likewise here

trio/_unix_pipes.py
```diff
@@ -65,21 +68,38 @@ def __del__(self):
...
     def __init__(self, fd: int):
         self._fd_holder = _FdHolder(fd)
-        self._conflict_detector = ConflictDetector(
-            "another task is using this pipe"
+        self._send_conflict_detector = ConflictDetector(
+            "another task is using this stream for send"
+        )
+        self._receive_conflict_detector = ConflictDetector(
+            "another task is using this stream for receive"
         )
```
njsmith: Yep, this is the right way to do it.

docs/source/reference-hazmat.rst
```diff
@@ -174,6 +174,40 @@ All environments provide the following functions:
...
+or processes are using file descriptors that are related through `os.dup`
+or inheritance across `os.fork` to the one that Trio is using, they are
+unlikely to be prepared to have non-blocking I/O semantics suddenly
+thrust upon them.  For example, you can use ``FdStream(os.dup(0))`` to
+obtain a stream for reading from standard input, but it is only safe to
+do so with heavy caveats: your stdin must not be shared by any other
+processes and you must not make any calls to synchronous methods of
+`sys.stdin` until the stream returned by `FdStream` is closed. See
+`issue #174 <https://github.com/python-trio/trio/issues/174>`__ for a
+discussion of the challenges involved in relaxing this restriction.
```
njsmith: Maybe this should be merged into the class docstring?

belm0:
> the only reason `_FdHolder` exists is to factor out some common code from `PipeSendStream` and `PipeReceiveStream`. Now that we're merging those two classes into one, it's no longer really serving any purpose, and we could merge it back into `FdStream`.

I'm inclined to keep the separation as it keeps FdStream readable.  FdHolder could have its own test.

njsmith:
I made a few trivial wording tweaks, and otherwise looks great! Feel free to merge once CI passes, in case I forget...

njsmith:
Actually, I take that back :-). The coverage report is saying that `from ._unix_pipes import FdStream` always succeeds, even on Windows: https://codecov.io/gh/python-trio/trio/pull/1129/src/trio/hazmat.py?before=trio/hazmat.py

Which makes sense, because there's nothing in `trio._unix_pipes` that depends on platform-specific stuff at import time. There used to be, when we were importing `fcntl`, but not anymore. So I guess right now we're exporting `trio.hazmat.FdStream` on Windows. But of course if anyone tries to use it on Windows then it will just crash.

(Side note: it's too bad the tests don't catch this... I guess they're doing `skipif(not os.name == "posix")`, not `skipif(not hasattr(trio.hazmat, "FdStream")`?)

The obvious solution would be to make it `if os.name == "posix": from ._unix_pipes import FdStream`, but from your commit comments, I guess jedi didn't like that. Which is irritating :-/. Annoyingly, mypy doesn't understand it either: https://mypy.readthedocs.io/en/latest/common_issues.html#version-and-platform-checks

One option would be to use `if os.name == "posix"`, and accept that jedi won't autocomplete correctly, and hack the jedi test to add a special-case exception

Another option, which is kind of a silly hack, but I guess would work: at the top of `_unix_pipes.py`, do `from os import get_blocking, set_blocking`, which will raise `ImportError` on Windows. Of course this would need a comment :-)

belm0:
I'm now manually raising ImportError from `_unix_pipes.py`, and added coverage for this.

njsmith:
Looks great, waiting for CI now...

PR #1130, opened 2019-06-30
improve BlockingTrioPortal documentation

belm0:
* improve `BlockingTrioPortal.run()` exception message
  * clarify the meaning of "Trio thread"
  * clarify that the `BlockingTrioPortal.run()` thread check applies to *any* Trio thread


belm0:
context:  https://gitter.im/python-trio/general?at=5d1800979cbde24b2f42b90b

belm0:
I'll reword the message now that I understand meaning of "trio thread" better, and perhaps expand this PR to define that better in the docs.

belm0:
so it turns out BlockingTrioPortal docs are not clear that the check applies to *any* Trio thread, so I clarified that as well

**to consider: if the rule is merely about preventing deadlock, perhaps the check could be loosened to only the Trio thread associated with the portal**

docs/source/reference-core.rst
```diff
@@ -195,7 +195,8 @@ The vast majority of Trio's API is *not* thread safe: it can only be
 used from inside a call to :func:`trio.run`. This manual doesn't
 bother documenting this on individual calls; unless specifically noted
 otherwise, you should assume that it isn't safe to call any Trio
-functions from anywhere except the Trio thread. (But :ref:`see below
+functions from anywhere except the Trio thread (i.e. the OS thread
```
oremanj: maybe: from anywhere except the thread that invoked `trio.run` (the "Trio thread"). But see below [etc].

trio/_threads.py
```diff
@@ -19,7 +19,7 @@
 
 class BlockingTrioPortal:
     """A portal that synchronous threads can reach through to run code in the
-    Trio thread.
+    Trio thread (i.e. the OS thread which invoked :func:`trio.run`).
```
oremanj: Nit: either remove "i.e." or put a comma after it

trio/_threads.py
```diff
@@ -81,7 +81,8 @@ def _do_it(self, cb, fn, *args):
             pass
         else:
             raise RuntimeError(
-                "this is a blocking function; call it from a thread"
+                "BlockingTrioPortal run() or run_async() was called from a "
```
oremanj: run_sync(), not run_async(). I don't think we want to call it "a trio.run() thread" because that confuses the terminology, so could I suggest instead something like the following? "must be called from a thread that is not inside a call to `trio.run`", or maybe "from a thread that is not running the Trio event loop".

trio/_threads.py
```diff
@@ -100,7 +101,7 @@ def run(self, afn, *args):
           Cancelled: if the corresponding call to :func:`trio.run` completes
               while ``afn(*args)`` is running, then ``afn`` is likely to raise
               :class:`Cancelled`, and this will propagate out into
-          RuntimeError: if you try calling this from inside the Trio thread,
+          RuntimeError: if you try calling this from inside a Trio thread,
```
oremanj: hmm. I see the reasoning here (you can have more than one Trio thread in a process) but I think that's quite an expert-level trick, and being consistent about calling it "the Trio thread" might be more understandable since for 99% of users there will only be one. (Once you have multiple threads you start needing to worry about which thread owns each channel/lock/etc, and you need to write your own, uhh, "NonBlockingTrioPortal" to communicate between them.)

I think my preference here would be to be consistent about saying "the Trio thread" in our docs, and maybe add a note somewhere that yes, you can have multiple threads in the same process each in their own call to `trio.run()`, but communicating between those threads will require dipping down into hazmat. Not a strong preference though.

smurfix: My preference too.

njsmith:
What do folks think about using phrasing like "inside `trio.run`" instead of "inside a/the Trio thread"?

smurfix:
Well, questions along the lines of "I've started that other thread from inside `trio.run` so why isn't this working" come to mind.

njsmith:
I'm afraid I've lost track of the status of this, and in the mean time `BlockingTrioPortal` has turned into `trio.from_thread` – though most of the docs were moved over verbatim. @belm0 is this still something you're interested in working on, and if so do you know what the status is?

Fuyukai:
BlockingTrioPortal got removed four years ago. Closing as this is no longer relevant.

PR #1131, opened 2019-06-30
update documentation of RunFinishedError

belm0:


njsmith:
I switched it to single-backticks, because (a) that makes a link, which is nice, (b) it will mean that if/when `BlockingTrioPortal.run` gets removed (not a theoretical concern: #1122), we'll notice...

njsmith:
Weird failure from azure pipelines (on Win py36 32-bit). Hopefully just a glitch; I hit "re-run":

```
+ nuget install pythonx86 -Version 3.6.8 -OutputDirectory /d/a/1/s/pyinstall -ExcludeVersion -Source https://api.nuget.org/v3/index.json -Verbosity detailed -DirectDownload -NonInteractive
NuGet.Config is not valid XML. Path: 'C:\Users\VssAdministrator\AppData\Roaming\NuGet\NuGet.Config'.
  Root element is missing.
```

njsmith:
Well, Azure Pipelines does have a "re-run" button now on Github, and I clicked it, but as far as I can tell it didn't do anything at all. So I guess we have to use a bigger hammer:

PR #1137, opened 2019-07-03
add a new is_readable method to SocketType (fix #760)

### LINKED ISSUE #760: Add a method to trio socket objects that checks whether the OS thinks the socket is readable
Say you're writing an HTTP/1.1 client library, that supports connection pooling. When you take an idle connection out of your pool, it might be fine to use... OR, the server might have gotten tired of waiting for that connection to do something while it was sitting idle in the pool, and have already closed it. Because of infelicities in HTTP/1.1's design, there is no 100% reliable way to detect this, because there's a race condition: you might send a new request down the connection just as the server is closing it, and then you have no idea what happened and whether the connection dropped before or after the server started processing your request.

So, we would like to minimize how often this happens. But, because there's actually no way to detect it at the protocol level, there really is no good way to do this using regular high-level APIs. The best approach is: when you take an idle connection out of the pool, before you start using it again, do a quick check whether the underlying socket is readable. You don't want to actually read from it; there's nothing to learn by doing so. If an HTTP/1.1 server sent anything on an idle connection, it's either a 408 Request Timeout or just a preemptory close. You especially don't want to block waiting to receive data, since that would defeat the whole purpose – we just want to check whether you *could* receive data. And you want to do this directly on the raw socket, ignoring things like any TLS encryption layer, because TLS does not really provide any way to do a pure check for whether *unencrypted* bytes are available – to figure that out you first have to read the encrypted bytes and then check if they make a complete frame, and we don't want to read any bytes.

So... for this specific, weird, and incredibly important use case, you don't want to use Trio's high-level `Stream` layer, you want to peek through all the abstraction layers, ignore Trio's socket I/O API entirely, and call `select` or `poll` directly on the underlying socket just to ask the OS whether it is readable.

So far, so good; Trio lets you do all that using supported, public APIs.

But! Suppose that for testing, you want to use a fake virtual networking layer. That's great: we have first-class support for that in Trio's socket I/O API (#170). Except... asks or urllib3 or whatever, for this one operation, currently has to do an end-run around Trio's socket I/O API – it has to directly call the OS's `select` or `poll` operation, which means that it needs a real OS socket, not a fake virtual socket. And that means that currently, our network faking API cannot handle HTTP clients, which is kind of an important use case.

Solution: we should add a method on Trio socket objects that does this readability check. It's pretty straightforward, [though there are a few complications](https://github.com/urllib3/urllib3/blob/master/src/urllib3/util/wait.py). (Mostly, you need to use `select` on windows and `poll` everywhere else.) And the whole point of this is that libraries like asks or urllib3 will call this method instead of doing their own call to `select`/`poll`. And it will do exactly the same thing they would have done by hand. Except..... since it's a method on the socket object, our fake virtual sockets will be able to override it to do a fake virtual `select`/`poll`. And that will solve the problem.

ziirish:
As explained in #760, this PR adds a new `is_readable` method to `trio.socket.SocketType` object.

### DIFF
--- a/newsfragments/760.feature.rst
+++ b/newsfragments/760.feature.rst
@@ -1,3 +1,2 @@
-We added a new ``is_readable`` method to the ``trio.socket.SocketType``
-object that allows you to check whether a socket is ready to be read
-or not.
+Trio sockets have a new method `~trio.socket.SocketType.is_readable` that allows
+you to check whether a socket is readable. This is useful for HTTP/1.1 clients.
--- a/newsfragments/760.feature.rst
+++ b/newsfragments/760.feature.rst
@@ -1,0 +1,3 @@
+We added a new ``is_readable`` method to the ``trio.socket.SocketType``
+object that allows you to check whether a socket is ready to be read
+or not.
--- a/docs/source/reference-io.rst
+++ b/docs/source/reference-io.rst
@@ -403,6 +403,10 @@ Socket objects
          left in an unknown state – possibly open, and possibly
          closed. The only reasonable thing to do is to close it.
 
+   .. method:: is_readable
+
+      Check whether the socket is readable or not.
+
    .. method:: sendfile
 
       `Not implemented yet! <https://github.com/python-trio/trio/issues/45>`__
--- a/trio/_socket.py
+++ b/trio/_socket.py
@@ -1,5 +1,6 @@
 import os as _os
 import sys as _sys
+import select
 import socket as _stdlib_socket
 from functools import wraps as _wraps
 
@@ -289,7 +290,7 @@ def socket(
 
 def _sniff_sockopts_for_fileno(family, type, proto, fileno):
     """Correct SOCKOPTS for given fileno, falling back to provided values.
-    
+
     """
     # Wrap the raw fileno into a Python socket object
     # This object might have the wrong metadata, but it lets us easily call getsockopt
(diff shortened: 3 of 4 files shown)


newsfragments/760.feature.rst
```diff
@@ -0,0 +1,3 @@
+We added a new ``is_readable`` method to the ``trio.socket.SocketType``
+object that allows you to check whether a socket is ready to be read
+or not.
```
njsmith: The double-backticks are raw code; since we're referring to objects that have entries in the documentation, we can use single-backticks to link directly:

```
Trio sockets have a new method `~trio.socket.SocketType.is_readable`, that allows you
to check whether a socket is readable. This is useful for HTTP/1.1 clients.
```

(The tilde is a shorthand that means "make a link to `trio.socket.SocketType.is_readable`, but just render it as `is_readable`".)

ziirish:
Thanks for the review.
I have addressed the comments.

njsmith:
Thanks!

ashburry-chat-irc:
How do I use the is_readable attribute? I get ("'SocketStream' object has no attribute 'is_readable'",) Any way to load it or something?

oremanj:
This is a method of the underlying socket object, not the SocketStream. Use `stream.socket.is_readable()`, not `stream.is_readable()`.

PR #1140, opened 2019-07-06
[WIP] pyOpenSSL-backed TLS

dholth:
I would like to be able to do OCSP stapling and other SSL tricks. Tried doing a ssl -> OpenSSL.SSL replace for `asyncio` https://bitbucket.org/dholth/bioopenssl/src/default/ , am now trying the same thing for `trio`. Amazed at how much easier to read `trio` is, but haven't gotten very far with this particular change.

OpenSSL.SSL is completely low level and seems to have no examples outside of the tests. Here are some examples I found: 

- https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/pyopenssl.py
- https://github.com/twisted/twisted/blob/twisted-16.4.1/twisted/internet/_sslverify.py#L1341

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ __pycache__/
 .installed.cfg
 *.egg
 /.pybuild
+pip-wheel-metadata
 
 # Installer logs
 pip-log.txt
--- a/trio/_openssl.py
+++ b/trio/_openssl.py
@@ -302,9 +302,6 @@ class SSLStream(Stream):
 
     """
 
-    _outgoing: _SSL.Connection.bio_read
-    _incoming: _SSL.Connection.bio_write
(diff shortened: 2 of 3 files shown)


dholth:
Alternative plan: use urllib3's stdlib API wrapper around pyOpenSSL; it would need to have the server-side bits and the BIO (in memory streams) API added.

dholth:
Second alternative possibility: pypy's ssl has the same interface as the stdlib's and is based on the cryptography package, which is also the base of pyopenssl.

njsmith:
@dholth can we discuss the overall strategy in #1145 first?

dholth:
How's trio's ssl hangin'

PR #1156, opened 2019-07-27
Add new trio.to_thread module for running worker threads

njsmith:
For consistency with trio.from_thread, and to give us a place for
future extensions, like utilities for pushing context managers into
threads.

See gh-810.


njsmith:
@epellis I know, for a library that tries to avoid threads we sure do use or mention threads a lot :-)

Want to go ahead and hit the green merge button?

njsmith:
woot :tada:

PR #1171, opened 2019-08-02
Workaround OpenSSL 1.1.1 session-ticket issue

### LINKED ISSUE #819: Deadlocks caused by openssl's handling of TLS 1.3 session tickets

njsmith:
Fixes #819

### DIFF
--- a/trio/tests/test_highlevel_ssl_helpers.py
+++ b/trio/tests/test_highlevel_ssl_helpers.py
@@ -40,7 +40,10 @@ class FakeHostnameResolver(trio.abc.HostnameResolver):
 
 
 # This uses serve_ssl_over_tcp, which uses open_ssl_over_tcp_listeners...
-async def test_open_ssl_over_tcp_stream_and_everything_else(client_ctx):
+# noqa is needed because flake8 doesn't understand how pytest fixtures work.
+async def test_open_ssl_over_tcp_stream_and_everything_else(
+    client_ctx,  # noqa: F811
+):
     async with trio.open_nursery() as nursery:
         (listener,) = await nursery.start(
             partial(
--- a/trio/tests/test_highlevel_ssl_helpers.py
+++ b/trio/tests/test_highlevel_ssl_helpers.py
@@ -7,7 +7,7 @@ import attr
 import trio
 from trio.socket import AF_INET, SOCK_STREAM, IPPROTO_TCP
 import trio.testing
-from .test_ssl import CLIENT_CTX, SERVER_CTX
+from .test_ssl import client_ctx, SERVER_CTX
 
 from .._highlevel_ssl_helpers import (
     open_ssl_over_tcp_stream, open_ssl_over_tcp_listeners, serve_ssl_over_tcp
@@ -40,7 +40,7 @@ class FakeHostnameResolver(trio.abc.HostnameResolver):
 
 
 # This uses serve_ssl_over_tcp, which uses open_ssl_over_tcp_listeners...
-async def test_open_ssl_over_tcp_stream_and_everything_else():
+async def test_open_ssl_over_tcp_stream_and_everything_else(client_ctx):
     async with trio.open_nursery() as nursery:
         (listener,) = await nursery.start(
             partial(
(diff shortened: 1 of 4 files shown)


njsmith:
The remaining test failure should be fixed by #1172

njsmith:
Not sure why codecov isn't updating its comment. If you click through then it says that everything is covered.

trio/tests/test_highlevel_ssl_helpers.py
```diff
@@ -40,7 +40,10 @@ class FakeHostnameResolver(trio.abc.HostnameResolver):
 
 
 # This uses serve_ssl_over_tcp, which uses open_ssl_over_tcp_listeners...
-async def test_open_ssl_over_tcp_stream_and_everything_else():
+# noqa is needed because flake8 doesn't understand how pytest fixtures work.
```
pquentin: I think you can fix the flake8 issue by removing the unused client_ctx import

njsmith: It's not unused, it's making the fixture function visible to pytest...

oremanj:
I'm a big fan of the analysis on #819 - thank you for fixing this!

trio/tests/test_highlevel_ssl_helpers.py
pquentin: Sorry about that, I'm used to projects where pytest discovers the fixtures through contest.py

PR #1172, opened 2019-08-02
Add new socket module constants from python 3.8

njsmith:

PR #1173, opened 2019-08-02
Use a longer sleep in test_run::test_timekeeping

njsmith:
This test has been the source of a few intermittent failures recently:

  https://github.com/python-trio/trio/issues/200#issuecomment-515667567

Not really clear if this will solve the problem, but it's worth a try.


njsmith:
Test failure should be fixed by #1172

PR #1176, opened 2019-08-05
Prevent uncancelled scopes from catching Cancelled exceptions

### LINKED ISSUE #1175: New cancellation logic doesn't handle in-flight shielding changes correctly
Reproducer:
```python
import trio

async def main():
    with trio.CancelScope() as outer:
        with trio.CancelScope() as inner:
            outer.cancel()
            try:
                await trio.hazmat.checkpoint()
            finally:
                inner.shield = True
    print(outer.cancel_called, inner.cancel_called)
    print(outer.cancelled_caught, inner.cancelled_caught)

trio.run(main)
```
Prints:
```
True False
False True
```
It shouldn't be possible for a scope that wasn't cancelled to swallow a Cancelled.

oremanj:
This could happen in practice if a `finally:` block set an inner scope to be
shielded while a cancellation targeting the outer scope was propagating.
(Or if a user used private APIs to raise `Cancelled` directly themselves,
which is how it was first discovered.)

Fixes #1175.

### DIFF
--- a/newsfragments/1175.bugfix.rst
+++ b/newsfragments/1175.bugfix.rst
@@ -1,0 +1,4 @@
+It is no longer possible for a cancel scope that isn't cancelled to
+catch a propagating :exc:`trio.Cancelled` exception, even if you
+change shielding while a :exc:`trio.Cancelled` exception is propagating.
+(This fixes a regression introduced with the cancellation changes in 0.12.)
--- a/newsfragments/1175.bugfix.rst
+++ b/newsfragments/1175.bugfix.rst
@@ -1,4 +1,4 @@
-It is no longer possible for a cancel scope that isn't cancelled to
-catch a propagating :exc:`trio.Cancelled` exception, even if you
-change shielding while a :exc:`trio.Cancelled` exception is propagating.
-(This fixes a regression introduced with the cancellation changes in 0.12.)
+Fix regression introduced with cancellation changes in 0.12.0, where a
+`trio.CancelScope` which isn't cancelled could catch a propagating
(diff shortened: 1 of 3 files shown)


trio/_core/tests/test_run.py
```diff
@@ -840,6 +840,16 @@ def test_cancel_points():
             with pytest.raises(_core.Cancelled):
                 await _core.checkpoint()
 
+    # change shielding while a cancellation is in flight
```
belm0: would it be difficult to make this a separate test?

Testing is hard and itself error-prone, so there's some advantage to testing one thing at a time.  Also some opportunity for self-description via test function name.

newsfragments/1175.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+It is no longer possible for a cancel scope that isn't cancelled to
```
belm0: I'd word it primarily as fixing a regression:
```
Fix regression introduced with cancellation changes in 0.12.0, where a CancelScope
which isn't cancelled would catch a propagating :exc:`trio.Cancelled` exception.
```

trio/_core/tests/test_run.py
```diff
@@ -849,6 +849,18 @@ def test_cancel_points():
     assert not scope.cancelled_caught
 
 
+# Regression test for https://github.com/python-trio/trio/issues/1175
+async def test_unshield_while_cancel_propagating():
```
belm0: should it have been `test_shield_while_cancel_propagating`?

PR #1178, opened 2019-08-07
Recover from a 'foreign await' by throwing in the TypeError

### LINKED ISSUE #552: Our code for recovering after a "foreign await" needs to be reworked
Just doing some testing to see if I can use `dask.distributed` with `trio` & `trio_asyncio` but have fallen at the first hurdle and hit a `TrioInternalError`:

```python
import asyncio
import trio
import trio_asyncio
from dask.distributed import Client


async def f():
    async with trio_asyncio.open_loop() as loop:
        client = await Client(asynchronous=True)
        future = client.submit(lambda x: x + 1, 10)
        return await future

trio.run(f)
```
```python-traceback
AssertionError                            Traceback (most recent call last)
C:\Miniconda3\lib\site-packages\trio\_core\_run.py in run(async_fn, clock, instruments, restrict_keyboard_interrupt_to_checkpoints, *args)
   1228                     # is just to get rid of this extra indentation.
-> 1229                     result = run_impl(runner, async_fn, args)
   1230             except TrioInternalError:

C:\Miniconda3\lib\site-packages\trio\_core\_run.py in run_impl(runner, async_fn, args)
   1356                 # other exceptions as __context__ in unwanted ways.
-> 1357                 runner.task_exited(task, final_result)
   1358             else:

C:\Miniconda3\lib\site-packages\trio\_core\_run.py in task_exited(self, task, result)
    854             # the init task should be the last task to exit
--> 855             assert not self.tasks
    856         else:

AssertionError: 

TrioInternalError                         Traceback (most recent call last)
<ipython-input-41-03c750b89f65> in <module>()
----> 1 trio.run(f)

C:\Miniconda3\lib\site-packages\trio\_core\_run.py in run(async_fn, clock, instruments, restrict_keyboard_interrupt_to_checkpoints, *args)
   1233                 raise TrioInternalError(
   1234                     "internal error in trio - please file a bug!"
-> 1235                 ) from exc
   1236             finally:
   1237                 GLOBAL_RUN_CONTEXT.__dict__.clear()

TrioInternalError: internal error in trio - please file a bug!
```

Happy to move this to the `trio_asyncio` repo if that's a more appropriate place...

oremanj:
This improves debuggability (since the traceback points to the location of the error, rather than just the nursery whose task was at fault) and fixes #552 by ensuring we don't abandon the child tasks of the errant one.

### DIFF
--- a/newsfragments/552.bugfix.rst
+++ b/newsfragments/552.bugfix.rst
@@ -1,0 +1,1 @@
+When a Trio task makes improper use of a non-Trio async library, Trio now causes an exception to be raised within the task at the point of the error, rather than abandoning the task and raising an error in its parent. This improves debuggability and resolves the `TrioInternalError` that would sometimes result from the latter strategy.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -984,7 +984,10 @@ class Task:
     # - for scheduled tasks, _next_send_fn(_next_send) resumes the task;
     #   usually _next_send_fn is self.coro.send and _next_send is an
     #   Outcome. When recovering from a foreign await, _next_send_fn is
-    #   self.coro.throw and _next_send is an exception.
+    #   self.coro.throw and _next_send is an exception. _next_send_fn
+    #   will effectively be at the top of every task's call stack, so
+    #   it should be written in C if you don't want to pollute Trio
+    #   tracebacks with extraneous frames.
     # - for scheduled tasks, custom_sleep_data is None
     # Tasks start out unscheduled.
     _next_send_fn = attr.ib(default=None)
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -980,10 +980,14 @@ class Task:
     _counter = attr.ib(init=False, factory=itertools.count().__next__)
 
     # Invariant:
-    # - for unscheduled tasks, _next_send is None
-    # - for scheduled tasks, _next_send is an Outcome object,
-    #   and custom_sleep_data is None
+    # - for unscheduled tasks, _next_send_fn and _next_send are both None
+    # - for scheduled tasks, _next_send_fn(_next_send) resumes the task;
+    #   usually _next_send_fn is self.coro.send and _next_send is an
+    #   Outcome. When recovering from a foreign await, _next_send_fn is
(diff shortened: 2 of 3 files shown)


njsmith:
Nice!

- I think we have a subtle invariant that the send/throw function has to implemented in C, to make tracebacks work right. Probably worth mentioning this in the comment with all the invariants?

- Can we add a regression test with something like this code in it? https://github.com/python-trio/trio/issues/552#issuecomment-457879973

oremanj:
Added the requested test. I don't think having send() written in C is a fundamental invariant; what in your model would require that? A Python implementation will add an extraneous traceback frame at each task boundary, but that seems like a reasonable accounting of what really happened. I tried running Trio's testsuite with `_next_send_fn` set to a Python wrapper around the C send(), and only one test failed (`test_traceback_frame_removal` which directly makes assertions about which frame comes first in the traceback).

oremanj:
(Not sure what codecov is on about -- it looks green on the website.)

njsmith:
Codecov is complaining that this line is "partially" covered:

```python
    assert any(entry.name == "child_xyzzy" for entry in excinfo.traceback)
```

I guess it's something to do with the embedded loop – maybe because it always exits early?

> I don't think having send() written in C is a fundamental invariant; what in your model would require that? A Python implementation will add an extraneous traceback frame at each task boundary, but that seems like a reasonable accounting of what really happened. I tried running Trio's testsuite with _next_send_fn set to a Python wrapper around the C send(), and only one test failed (test_traceback_frame_removal which directly makes assertions about which frame comes first in the traceback).

I guess I had two things in mind:

- We go to quite some trouble to only show user frames in user tracebacks, which is what that test is checking. I guess if for some reason we wanted to start supporting other methods besides `CoroutineType.send` and `CoroutineType.throw`, then we'd have to somehow update the traceback code?

- The KeyboardInterrupt protection logic is super subtle, and this method is a delicate moment: it needs to atomically transition from "trio internal mode" to "user mode". I guess any way of resuming the task has to have some atomic moment when it both puts the user's frame on the stack and also resumes the user's `try`/`except` blocks, even if you wrapped `send`/`throw` inside some Python wrapper, so maybe it's not an issue? But only using C functions here definitely makes the analysis simpler :-)

PR #1179, opened 2019-08-09
Update README to point to stable docs instead of dev docs

njsmith:
I recently switched the default docs version on RTD from 'latest' to
'stable', because people kept getting confused by reading about future
features that weren't released yet. But the README was still linking
directly to the 'latest' version.


pquentin:
Thank you for changing your mind on this! 👍

Issue #1180, opened 2019-08-09
StapledStream.send_eof should error out if you call it while send_all is pending


njsmith:
Noticed while working on #823: normally, if you try to call `send_eof` on an object at the same time that another task is already calling `send_all` or `wait_send_all_might_not_block`, then `send_eof` should raise `BusyResourceError`. But `StapledStream.send_eof` calls `self.send_stream.aclose`, which has a different semantics: it *interrupts* any pending calls. So right now `StapledStream.send_eof` has the wrong semantics.

To solve this, `StapledStream` should detect when another task is already calling a send method, and in that case make `send_eof` raise an exception instead of calling `self.send_stream.aclose`.

I have a fix for this in my #823 branch, that I'll post shortly.

Issue #823, opened 2018-12-27
Add send_eof to SSLStream and get rid of HalfCloseableStream


njsmith:
We have an awkward thing in our stream abstraction: there's the abstract [`Stream`](https://trio.readthedocs.io/en/latest/reference-io.html#trio.abc.Stream) interface, and then there's the abstract [`HalfCloseableStream`](https://trio.readthedocs.io/en/latest/reference-io.html#trio.abc.HalfCloseableStream) interface, which is just `Stream` with an added `send_eof` method.

In practice, every single `Stream` you'd ever meet in real life – TCP connections, [`StapledStream`](https://trio.readthedocs.io/en/latest/reference-io.html#trio.StapledStream), SSH tunnels, HTTP/2 streams, etc. etc. etc. – is a `HalfCloseableStream`. Except... `SSLStream`, because for some reason TLS doesn't support half-close. And unfortunately TLS is super important in practice, so even though in some abstract sense it's TLS that's broken and fails to implement a proper byte stream, we instead broke our `Stream` interface to accomodate TLS.

BUT! It turns out that TLS 1.3 fixed this! Quoting [RFC 8446 §6.1](https://tools.ietf.org/html/rfc8446#section-6.1):

>    Each party MUST send a "close_notify" alert before closing its write side of the connection, unless it has already sent some error alert. This does not have any effect on its read side of the connection. Note that this is a change from versions of TLS prior to TLS 1.3 in which implementations were required to react to a "close_notify" by discarding pending writes and sending an immediate "close_notify" alert of their own.  That previous requirement could cause truncation in the read side.  Both parties need not wait to receive a "close_notify" alert before closing their read side of the connection, though doing so would introduce the possibility of truncation.

Also @kroeckx [says](https://github.com/openssl/openssl/issues/7948#issuecomment-449723225):

> OpenSSL has supported that [= half-close] for a very long time, even when the older TLS standards didn't say that was supported.

So TLS 1.3 now has the same semantics as every other common byte-stream transport, and apparently in some cases we can get that with earlier versions too, if we cross our fingers and the moon is full.

I was already on the fence about whether we should have two separate types for `Stream` and `HalfCloseableStream`, versus merging them into a single type and just saying that `send_eof` might fail in some cases. This clearly tips things over, given that it means `SSLStream` is going to want to implement `send_eof`.

[...]

PR #1181, opened 2019-08-09
Make send_eof a standard part of the Stream interface

njsmith:
The only reason HalfCloseableStream existed as a separate interface
was because SSLStream couldn't support send_eof. But TLS 1.3 made it
possible to support send_eof! So now the static split doesn't make
sense. Instead, we move send_eof into the Stream interface (though it
can still fail at runtime with NotImplementedError), and get rid of
HalfCloseableStream. (Fixes: gh-823.)

This also includes a fix for StapledStream.send_eof, that was revealed
by the new enhanced check_two_way_stream. (Fixes: gh-1180).

### DIFF
--- a/trio/_highlevel_generic.py
+++ b/trio/_highlevel_generic.py
@@ -1,7 +1,8 @@
 import attr
 
 import trio
-from .abc import HalfCloseableStream
+from .abc import Stream
+from ._util import ConflictDetector
 
 
 async def aclose_forcefully(resource):
@@ -35,7 +36,7 @@ async def aclose_forcefully(resource):
 
 
 @attr.s(cmp=False, hash=False)
-class StapledStream(HalfCloseableStream):
+class StapledStream(Stream):
     """This class `staples <https://en.wikipedia.org/wiki/Staple_(fastener)>`__
     together two unidirectional streams to make single bidirectional stream.
 
@@ -54,9 +55,8 @@ class StapledStream(HalfCloseableStream):
           await echo_stream.send_all(b"x")
           assert await echo_stream.receive_some() == b"x"
 
-    :class:`StapledStream` objects implement the methods in the
-    :class:`~trio.abc.HalfCloseableStream` interface. They also have two
-    additional public attributes:
+    `StapledStream` objects implement the methods in the `~trio.abc.Stream`
+    interface. They also have two additional public attributes:
 
     .. attribute:: send_stream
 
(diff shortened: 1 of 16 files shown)


oremanj:
This looks great overall. One yapf nit to fix, and one backwards compatibility concern: adding `send_eof()` as an abstract method of Stream will break any user-provided Stream implementation that doesn't define it. Another approach with better properties there:
* make it non-abstract and have the default implementation raise NotImplementedError
* plus use `__init_subclass__` to raise a TrioDeprecationWarning when defining a subclass that doesn't override it
* could use `__new__` to warn on construction of such subclasses on 3.5 which doesn't call `__init_subclass__`, or could decide not to worry about that

I'm also OK leaving this incompatibility in, but it should probably be called out more explicitly in the newsfragments in that case.

njsmith:
@oremanj okay, fine, you're right :-)

I already figured out how to do this kind of deprecation properly while looking at #636, so might as well do it here too...

njsmith:
Still need to fix up some coverage.

njsmith:
OK, I think everything's been addressed now.

trio/_ssl.py
```diff
@@ -685,22 +686,19 @@ def _check_status(self):
...
-                # BROKEN. But that's actually fine, because after getting an
-                # EOF on TLS then the only thing you can do is close the
-                # stream, and closing doesn't care about the state.
-                if (
-                    self._https_compatible
-                    and isinstance(exc.__cause__, _stdlib_ssl.SSLEOFError)
-                ):
-                    await trio.hazmat.checkpoint()
+
+            def do_read():
```
wgwz: one thought here is that maybe it would be better to move this function out into it's own method? i don't have any really strong opinion on whether or not that makes that code more readable. but i just wanted to bring it up to see what others think. i definitely think that this is cleaner than the previous approach.

trio/_highlevel_socket.py
```diff
@@ -119,6 +119,8 @@ def __init__(self, socket):
                         total_sent += sent
 
     async def wait_send_all_might_not_block(self):
+        if self.socket.did_shutdown_SHUT_WR:
+            raise trio.ClosedResourceError("can't send data after sending EOF")
```
wgwz: i'm not exactly sure where these changes fit in to the issues this MR is fixing. perhaps there is something that can be added to the newsfragment? it seems related related to what you were fixing in #1180 but it's in a different class.

trio/_highlevel_generic.py
```diff
@@ -71,29 +71,42 @@ class StapledStream(HalfCloseableStream):
...
-        calls ``self.send_stream.aclose()``.
+        If ``self.send_stream.send_eof`` exists, then attempts to call it.
+        Otherwise, calls ``self.send_stream.aclose()``.
 
         """
-        if hasattr(self.send_stream, "send_eof"):
-            return await self.send_stream.send_eof()
-        else:
+        with self._send_conflict_detector:
+            if hasattr(self.send_stream, "send_eof"):
```
wgwz: now that all classes inheriting from `Stream` are required to implement `send_eof`, why do we need to check that the sending stream has the `send_eof` method here? i think i am probably misunderstanding a subtle point here. maybe this is for backwards compatibility or maybe there are other types of streams that i am unaware of, which can be "stapled" together.

trio/_highlevel_socket.py
njsmith: I don't remember exactly anymore, but I think the sequence was (1) I reworked the `send_eof` tests in `trio.testing.check_*`, (2) `SocketStream` started failing the more rigorous tests, (3) I fixed `SocketStream` to pass the tests again. I'll add a newsfragment...

trio/_highlevel_generic.py
njsmith: All `Stream`s have to implement `send_eof`, but unidirectional `SendStream`s don't have to, and in this method, all we know about `self.send_stream` is that it's a `SendStream`, not necessarily a `Stream`.

I'm not actually sure if this opportunistic attempt to call `send_eof` accomplishes anything useful... it would be simpler, and maybe better, to unconditionally call `aclose`. But I decided this PR was already doing enough, so I'd leave worrying about that to another time...

trio/_highlevel_socket.py
njsmith: I don't remember exactly anymore, but I think the sequence was (1) I reworked the `send_eof` tests in `trio.testing.check_*`, (2) `SocketStream` started failing the more rigorous tests, (3) I fixed `SocketStream` to pass the tests again. I'll add a newsfragment...

oremanj:
Sorry for dropping this. I took another look just now and I'm happy with it, except that the version numbers of the deprecations need to be increased from 0.13.0 to 0.15.0.

PR #1184, opened 2019-08-12
Use finally instead of reraising the exception.

thedrow:
If an exception was raised during the initialization of the thread we don't have to reraise it.
We just have to cleanup afterwards.
The exception will be raised anyway.

### DIFF
--- a/trio/_threads.py
+++ b/trio/_threads.py
@@ -296,9 +296,8 @@ async def to_thread_run_sync(sync_fn, *args, cancellable=False, limiter=None):
             daemon=True
         )
         thread.start()
-    except:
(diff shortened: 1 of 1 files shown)


njsmith:
This makes it so we release the limiter even if the thread initialization *succeeded*. That's wrong – if the thread starts running successfully, then we want to keep holding the limiter until the thread exits. That happens in `report_back_in_trio_thread_fn`.

So I think the original code is correct: we only want to release the limiter here if the thread failed to start.

thedrow:
Oh right. My bad.
Good to see that the tests capture that error.

PR #1185, opened 2019-08-13
Fix minor issues in CombinedLimiter example

tjstum:
- Typo in function passed to `run_sync`
- It looks like the example predates the dropping of kwargs, so limiter wasn't actually ever being passed


njsmith:
Nice catch!

Issue #550, opened 2018-06-26
Support Cython-based async functions


touilleMan:
The title say it all, some projects re-implement their own iscoroutine function by looking for `CO_COROUTINE` and `CO_ITERABLE_COROUTINE` flags ([for instance](https://github.com/ssanderson/interface/pull/17/files#diff-62977b350e62b018986bf62bcbace77aR63)) to address this...

This is more of a CPython issue of course, but this currently means the coroutine generated by cython are not compatible with trio:
```python
import trio

async def trio_main():
    print('hello...')
    await trio.sleep(1)
    print(' world !')

def main():
    trio.run(trio_main)

if __name__ == '__main__':
    main()
```

```
$ python cython_test.py 
hello...
 world !
$ cythonize -a -i cython_test.py 
Compiling /home/emmanuel/projects/parsec-lrw/cython_test.py because it changed.
[1/1] Cythonizing /home/emmanuel/projects/parsec-lrw/cython_test.py
running build_ext
building 'cython_test' extension
creating /home/emmanuel/projects/parsec-lrw/tmp0p_t5pyt/home
creating /home/emmanuel/projects/parsec-lrw/tmp0p_t5pyt/home/emmanuel
creating /home/emmanuel/projects/parsec-lrw/tmp0p_t5pyt/home/emmanuel/projects
creating /home/emmanuel/projects/parsec-lrw/tmp0p_t5pyt/home/emmanuel/projects/parsec-lrw
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/home/emmanuel/projects/cpython/build-install/include/python3.6m -c /home/emmanuel/projects/parsec-lrw/cython_test.c -o /home/emmanuel/projects/parsec-lrw/tmp0p_t5pyt/home/emmanuel/projects/parsec-lrw/cython_test.o
gcc -pthread -shared /home/emmanuel/projects/parsec-lrw/tmp0p_t5pyt/home/emmanuel/projects/parsec-lrw/cython_test.o -o /home/emmanuel/projects/parsec-lrw/cython_test.cpython-36m-x86_64-linux-gnu.so
$ python 
Python 3.6.3 (v3.6.3:2c5fed8, Feb  6 2018, 17:05:14) 
[GCC 5.4.1 20160904] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cython_test
>>> cython_test.main()
[...]

Issue #1191, opened 2019-08-17
TrioInternalError on trio.run(agen.__anext__)


njsmith:
Here's a strange crash, discovered by @florimondmanca in https://github.com/encode/httpx/pull/214:

```python
import trio
async def gen():
    yield "hi"
trio.run(gen().__anext__)
```

Results in:

```
Traceback (most recent call last):
  File "/tmp/asdf.py", line 6, in <module>
    trio.run(gen().__anext__)
  File "/home/njs/.user-python3.7/lib/python3.7/site-packages/trio/_core/_run.py", line 1769, in run
    run_impl(runner, async_fn, args)
  File "/home/njs/.user-python3.7/lib/python3.7/site-packages/trio/_core/_run.py", line 1918, in run_impl
    runner.task_exited(task, final_outcome)
  File "/home/njs/.user-python3.7/lib/python3.7/site-packages/trio/_core/_run.py", line 1404, in task_exited
    raise TrioInternalError
trio.TrioInternalError
```

Poking around, I see that `init` has set `self.main_task_outcome` before exiting:

```
(Pdb) p self.main_task_outcome
Error(AttributeError("'async_generator_asend' object has no attribute 'cr_frame'"))
```

So this is another variant of #550.

This is a pretty weird thing to do, so I guess this issue isn't super urgent, but two todo items:

- The error message is really bad... there are several measures in place here to try to make the error come out, and they all failed. The run loop tries to reveal exceptions from `init`, but  `init` exited successfully. It did this after setting `self.main_task_exited`... but even though `spawn_impl` failed, it didn't fail until after adding the new task to `self.tasks`, so `run` freaked out instead of re-raising `self.main_task_exited`. I guess we should move the `self.tasks` assignment down even further in `spawn_impl`?

- Also, this probably shouldn't crash! And #550 has another set of reasons to support it.

njsmith:
What's neat about this discovery though is that it gives us a way to test our fix for #550 without having to build a Cython extension in our testsuite!

njsmith:
Fixed by #1192

PR #1192, opened 2019-08-17
Don't crash on C coroutine objects

njsmith:
Fixes: gh-550, gh-1191


oremanj:
You'll need to hide the async generator test from the 3.5 builds. Otherwise looks good.

njsmith:
So apparently my first version didn't work at all, because we were replacing `coro` *after* creating the `Task` object, so the `Task` kept going with the original coroutine... fortunately codecov caught it. I fixed that and made the test more robust, so I *think* this is good now?

njsmith:
We did some code review for this PR on gitter, starting here: https://gitter.im/python-trio/general?at=5d58b8434e17537f525114d7

Issue #1190, opened 2019-08-16
Mock clock cannot jump into the past, but uses seconds_to_next_deadline which can !


touilleMan:
When using hypothesis with the `mock_clock` fixture, I sometime end up with a `raise ValueError("time can't go backwards")`

The thing is we have `MockClock._autojumper` doing this:
https://github.com/python-trio/trio/blob/4512ae3ca199f0a99d02f6ff2a6f8e94dd08f1ef/trio/testing/_mock_clock.py#L149-L152

However according to `current_statisticts` documentation:
https://github.com/python-trio/trio/blob/4512ae3ca199f0a99d02f6ff2a6f8e94dd08f1ef/trio/_core/_run.py#L1143-L1146

I guess `_autojumper` should be corrected to do something like this:
```python
statistics = _core.current_statistics() 
jump = statistics.seconds_to_next_deadline 
if 0 < jump < inf:
     self.jump(jump) 
```

What do you think ?

njsmith:
Oh interesting!

We certainly don't want to try to do a negative jump, so your proposed fix seems safe in that regard.

But I am confused about how that deadline can be in the past at the same time that we're getting told that all tasks are blocked. If time were advancing normally then I get it, maybe the deadline expired in the fraction of a second between when we determined that all tests were blocked vs when the clock tries to advance. But I assume you're using MockClock in the mode where time *only* moves forward when someone calls `jump`, in which case... The run loop *ought* to have processed those deadlines before waking up the clock task. So that makes me nervous that there's another bug here we don't understand yet.

Do you have a reproducer?

touilleMan:
> But I assume you're using MockClock in the mode where time only moves forward when someone calls jump

Maybe my use of MockClock is a bit exotic:
I'm testing a background task that does some synchronization. To do that I have two kind of hypothesis rules:
1) Rules that create data that need to be synchronized. During those rules I don't want the background task to wake up randomly and do concurrent stuff so I keep `MockClock.rate == 0`
2) A single rule that wait for the synchronization to occur and then check the state of the tested system. This rule is the only time I switch `MockClock.rate == 1` and wait for a long time (with `autojump_threshold == 0`, considering I've mocked all my real IOs with stapled streams) until I'm sure the sync should have occurred. 

So in the end I'm freezing/unfreezing the MockClock really often

> If time were advancing normally then I get it, maybe the deadline expired in the fraction of a second between when we determined that all tests were blocked vs when the clock tries to advance

[...]

PR #1194, opened 2019-08-19
Ensure MockClock autojump never feed jump with negative values

touilleMan:
Fixes: #1190

### DIFF
--- a/newsfragments/1190.bugfix.rst
+++ b/newsfragments/1190.bugfix.rst
@@ -1,0 +1,2 @@
+Fix a crash that could happen when using ``MockClock`` with autojump
+enabled and a non-zero rate.
--- a/trio/testing/_mock_clock.py
+++ b/trio/testing/_mock_clock.py
@@ -148,7 +148,7 @@ class MockClock(Clock):
                     )
                     statistics = _core.current_statistics()
                     jump = statistics.seconds_to_next_deadline
-                    if jump < inf:
+                    if 0 < jump < inf:
                         self.jump(jump)
                     else:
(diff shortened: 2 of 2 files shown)


touilleMan:
not sure why the CI failed, any idea ?

njsmith:
It looks like the return of the Curse of Iowa Midnight:

https://github.com/python-trio/trio/pull/911
https://developercommunity.visualstudio.com/content/problem/458410/repeated-failed-azure-pipelines-builds-due-to-incr.html

(Note: we don't actually know that Iowa is involved at all; the azure folks swear that their network looks fine at the times when this happens.)

Anyway it's an intermittent issue unrelated to your PR, so the solution here is just to rerun the tests, which I will trigger by closing and reopening the issue.

njsmith:
@oremanj hmm, is this why you did a close/reopen earlier too?

oremanj:
It is, and the issue persisted at that point, but probably I just hadn’t waited long enough after the failure.

njsmith:
The change looks fine – thanks! Since it's fixing a bug, I guess it should have a newsfragment – maybe 1190.bugfix.rst, "Fix a crash that could happen when using `MockClock` with autojump enabled and a non-zero rate."

Also, not a problem here, but as a tip for the future, consider writing "[Fixes: #1190](https://help.github.com/en/articles/closing-issues-using-keywords)" instead of "Fix for #1190"

touilleMan:
@njsmith I've added the newsfragment and corrected the "fix for" thing ;-)

njsmith:
Awesome, thanks!

PR #1195, opened 2019-08-20
Use slots for channel state and statistics.

thedrow:


njsmith:
Looks good. I edited the newsfragment to refer to memory channels specifically, since we also have `Channel` as an abstract API.

PR #1232, opened 2019-10-04
Set --connect-timeout when running curl, to make --retry work, and hopefully fix #1231

### LINKED ISSUE #1231: Flakiness downloading codecov-bash
So this is baffling to me, and I know it has nothing to do with anything we did, but it's super frustrating: for some reason, running `curl https://codecov.io/bash` has recently become *super* flaky and is making our CI runs fail all the time and we gotta do something.

Observations:

The most common failure seems to be a timeout during connect. I've also seen this, which is super weird:

```
+curl --retry 5 -o codecov.sh https://codecov.io/bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:15 --:--:--     0
curl: (35) gnutls_handshake() failed: Error in the pull function.
```

According to the curl docs, `curl --retry 5` should retry on timeouts but.... I can't tell whether it's actually doing that?

Codecov as a whole is kind of flaky, so it's tempting to blame this on them. But I'm not sure this is actually their fault... `codecov.io` appears to be hosted by google (based on running `whois` on the IP address), so I'm guessing it's some google CDN or something? (Or maybe it's just some random server running on google cloud with no CDN in front of it, who knows?)

The failures seem to all be from Travis, not Azure. And I'm pretty sure Travis is running on Google Cloud servers. Wild supposition: maybe Travis's traffic is considered "internal" to Google Cloud and doesn't hit the CDN, while Azure's traffic is "external" and does hit the CDN? I have no idea. I guess we could try running traceroute from both Travis and Azure, but that seems like it might be running off down an unproductive rabbit-hole.

...oh wait, and actually it just failed on azure too, so never mind: https://dev.azure.com/python-trio/trio/_build/results?buildId=1031&view=logs&jobId=872bf439-86bb-5ce5-edcd-c35619d700a0

so... what the heck can we do about this? the obvious answer is to retry, but curl's `--retry` isn't working. Is that because there's something wrong with `--retry`, or is it because when we hit one of these failures, it's somehow "sticky"? Would putting our own retry loop around curl help?

I guess we could also like, stash a copy of codecov-bash somewhere more reliable, but that seems like it would create all kinds of operational annoyances.

njsmith:
Link: #1231

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -143,7 +143,8 @@ else
         #   bash <(curl ...)
         # but azure is broken:
         #   https://developercommunity.visualstudio.com/content/problem/743824/bash-task-on-windows-suddenly-fails-with-bash-devf.html
-        curl -v --retry 5 -o codecov.sh https://codecov.io/bash
+        curl --version
+        curl -v --connect-timeout 5 --retry 5 -o codecov.sh https://codecov.io/bash
         bash codecov.sh -n "${CODECOV_NAME}" -F "$FLAG"
     fi
 fi
--- a/ci.sh
+++ b/ci.sh
@@ -143,8 +143,9 @@ else
         #   bash <(curl ...)
         # but azure is broken:
         #   https://developercommunity.visualstudio.com/content/problem/743824/bash-task-on-windows-suddenly-fails-with-bash-devf.html
-        curl --version
-        curl -v --connect-timeout 5 --retry 5 -o codecov.sh https://codecov.io/bash
+        # Also we have to set --connect-timeout to work around:
+        #   https://github.com/curl/curl/issues/4461
+        curl --connect-timeout 5 --retry 5 -o codecov.sh https://codecov.io/bash
         bash codecov.sh -n "${CODECOV_NAME}" -F "$FLAG"
     fi
 fi
--- a/ci.sh
+++ b/ci.sh
@@ -143,7 +143,7 @@ else
         #   bash <(curl ...)
         # but azure is broken:
         #   https://developercommunity.visualstudio.com/content/problem/743824/bash-task-on-windows-suddenly-fails-with-bash-devf.html
-        curl --retry 5 -o codecov.sh https://codecov.io/bash
+        curl -v --retry 5 -o codecov.sh https://codecov.io/bash
         bash codecov.sh -n "${CODECOV_NAME}" -F "$FLAG"
     fi
 fi


njsmith:
Huh apparently the azure failures are unrelated, because nuget is down right now? of course any attempt to debug flaky infrastructure must hit unrelated flaky infrastructure. I don't know why I didn't expect that.

pquentin:
(I edited the title so that we get a link to the issue.)

njsmith:
@pquentin I guess the problem was that GH doesn't parse titles for links. I edited it into the body instead.

njsmith:
I thought the issue might be an old version of curl, but that failure I pasted above was from a Azure windows machine, and further investigation reveals that the Azure Windows machines have curl 7.65.3, which is even newer than the 7.64.0 I have on my laptop. But my laptop's version of curl provides clear evidence that it's handling retries properly:

```
~$ curl --connect-timeout 2 --retry 5 -v http://240.0.0.1
* Expire in 0 ms for 6 (transfer 0x55614e3b35c0)
* Expire in 2000 ms for 2 (transfer 0x55614e3b35c0)
*   Trying 240.0.0.1...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x55614e3b35c0)
* Connection timed out after 2001 milliseconds
* Closing connection 0
Warning: Transient problem: timeout Will retry in 1 seconds. 5 retries left.
* Expire in 0 ms for 6 (transfer 0x55614e3b35c0)
* Expire in 2000 ms for 2 (transfer 0x55614e3b35c0)
* Hostname 240.0.0.1 was found in DNS cache
*   Trying 240.0.0.1...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x55614e3b35c0)
* Connection timed out after 2001 milliseconds
* Closing connection 1
Warning: Transient problem: timeout Will retry in 2 seconds. 4 retries left.
```

...and so on.

Hmm, I did cheat there and use `--connect-timeout` to speed up the test. Could that be important? [several minutes of staring at an unchanged terminal later] Ah-hah!

```
~$ curl --retry 5 -v http://240.0.0.1                 
* Expire in 0 ms for 6 (transfer 0x557c953145c0)
*   Trying 240.0.0.1...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x557c953145c0)
* connect to 240.0.0.1 port 80 failed: Connection timed out
* Failed to connect to 240.0.0.1 port 80: Connection timed out
* Closing connection 0
curl: (7) Failed to connect to 240.0.0.1 port 80: Connection timed out
```

So apparently curl *knows* that the connect timed out, but it doesn't *count* as the connect timing out for retry purposes unless you explicitly set `--connect-timeout`. Which would make this a bug in curl. Sweet.

https://github.com/curl/curl/issues/4461

njsmith:
OK, so hopefully this fixes the issue now?

Issue #1210, opened 2019-09-14
Hard to read abc table


jtrakk:
It's not easy to read this table https://trio.readthedocs.io/en/stable/reference-io.html#abstract-base-classes because of the horizontal scrollbar. I've seen on some project's docs (can't remember which) a wide code-block that expands horizontally when you mouse-over it.

njsmith:
I agree that that scrollbar is pretty awful. I'm just clueless about HTML/JS/graphic design so I don't know how to make it work better :-). If anyone has any ideas, please speak up...

stevenjackson121:
I'm also clueless about HTML/JS/graphic design, but my best google-fu led me to https://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html as the simplest "how to fix" guide. The code contained there is identical to what appears in this thread https://github.com/readthedocs/sphinx_rtd_theme/issues/117, which is referenced by many projects in commits fixing this same issue.

njsmith:
Oh great find! We already have some override CSS in [`docs/source/_static/hackrtd.css`](https://github.com/python-trio/trio/blob/master/docs/source/_static/hackrtd.css), so I guess we could try dropping that text in there and see how that looks? (I guess it can't look much worse...) Anyone want to try making a PR?

stevenjackson121:
I gave it a try. I'm not happy with how it looks right now, but I'm not sure how to make it better.

The overflow: visible is very ugly as the table hangs past the white background), but it may be more usable even if it's ugly. The wrapping within table cells is a definite improvement.  The wrapping of text within table headers is somewhere in between. It also looks awkward but is probably worth it to have a table that isn't so wide and make it easier to actually use (especially if the overflow visible gets removed).

I think the best way to avoid the overflow vs scrollbar issue is just to have narrower tables. The text wrapping helps, but because long function references like `wait_send_all_might_not_block()` are over 200 px (and cannot be broken across lines), 5 column tables can still end up over 1000 px after padding which is obviously way above the 800 px background.

I tried looking for a way to make these references smaller, but you get into so many tradeoffs I'm quickly getting out of my depth:
1.) You could try to display first 10 characters only with "<...>" appended as in `wait_send_<...>` to limit the worst case to something more manageable. I don't even know how to do this, but it would probably be my preference, hopefully it could expand on mouseover (not having seen it actually implemented, take that with a grain of salt).
[...]

PR #1233, opened 2019-10-05
Improve Wide Table Handling in Docs

stevenjackson121:
* Allows text to wrap in table rows
* Allows text to wrap in table headers
* Forces overflow to be visible (ugly, but at least the table is readable on large screens... Not sure if this makes things worse for small screens though)

https://github.com/python-trio/trio/issues/1210

### DIFF
--- a/docs/source/_static/hackrtd.css
+++ b/docs/source/_static/hackrtd.css
@@ -110,3 +110,27 @@ a.trio-help-hint, .trio-help-hint a:link, .trio-help-hint a:visited {
     text-decoration: none;
     border-bottom: 1px solid;
 }
+
+/* override table width restrictions */
+@media screen and (min-width: 767px) {
+
+   .wy-table-responsive table td {
+      /* !important prevents the common CSS stylesheets from overriding
+         this as on RTD they are loaded after this stylesheet */
+      white-space: normal !important;
+   }
+
+   .wy-table-responsive table th {
+      /* This is a trio specific override since we have long header names compared to other projects who've had issues
+         with wide tables (allowing wrapping just in the cells isn't enough to shrink tables meaningfully).
+         */
+      white-space: normal !important;
+   }
+
+   .wy-table-responsive {
+      /* This makes very wide tables look ugly and extend past their parent... Maybe allowing the scroll bar isn't so
+         bad in comparison (after the above 2 fixes).
+         */
+      overflow: visible !important;
+   }
+}


pquentin:
I asked ReadTheDocs to generate your PR so that we can easily compare the two versions:

 * before: https://trio.readthedocs.io/en/latest/reference-io.html#abstract-base-classes
 * after: https://trio-pquentin.readthedocs.io/en/1233/reference-io.html#abstract-base-classes

For some reason, this pull request does not make any difference on mobile. On large desktop screens, the result is better. Here are Firefox screenshots:

before

after

However, on a small dekstop screen (eg. if I have a browser on the left and a terminal on the right, the width of the window is smaller than the maximum ReadTheDocs width), then it's not possible to scroll left to see more data. :(

Maybe the solution here is to stop displaying this is as a table but with lists and headers? Something like:

paragraphs

stevenjackson121:
@pquentin I just realized a comment I thought I added to this PR, I had actually left on [the issue](https://github.com/python-trio/trio/issues/1210#issuecomment-538667257) instead:

**Original Text Follows**

I gave it a try. I'm not happy with how it looks right now, but I'm not sure how to make it better.

The overflow: visible is very ugly as the table hangs past the white background), but it may be more usable even if it's ugly. The wrapping within table cells is a definite improvement. The wrapping of text within table headers is somewhere in between. It also looks awkward but is probably worth it to have a table that isn't so wide and make it easier to actually use (especially if the overflow visible gets removed).

I think the best way to avoid the overflow vs scrollbar issue is just to have narrower tables. The text wrapping helps, but because long function references like wait_send_all_might_not_block() are over 200 px (and cannot be broken across lines), 5 column tables can still end up over 1000 px after padding which is obviously way above the 800 px background.

I tried looking for a way to make these references smaller, but you get into so many tradeoffs I'm quickly getting out of my depth:
1.) You could try to display first 10 characters only with "<...>" appended as in wait_send_<...> to limit the worst case to something more manageable. I don't even know how to do this, but it would probably be my preference, hopefully it could expand on mouseover (not having seen it actually implemented, take that with a grain of salt).
2.) You could make text smaller obviously, but I have no idea whether to make all table text smaller, all text in the offending table smaller, all text in the offending column smaller, or all text in just the offending cell smaller (consistency is nice, but so are tables actually being displayable all at once)... The advantage here is you can see (and copy paste) the entire function name at once, but it looks weirder IMO

I thought of other options I can't remember now, but none that seemed like clear winners.

**End Original Text**

In other words, I agree that basically as it stands it's probably not merge-able. Converting it to a list is a good idea! Do you think any of my other suggestions are worth exploring, or is using a list just superior?

I do like think the wrapping within cells is a good improvement over the current default (and maybe headers also)? Even if we end up avoiding tables altogether for this piece of the docs (because this table is really huge), would it still make sense to include that change in this PR so more borderline tables can stay as tables?

pquentin:
Yes, I'd like to see a pull request with only the wrapping so that we can merge that quickly.

I'm not a huge fan of making the font size smaller or shortening the function names. But using a list is certainly not just superior as it's more verbose, so in a sense harder to read. Oh, and we can probably make the visible overflow proposal less ugly by fixing the background color.

stevenjackson121:
I removed the forced overflow, so you can view the text-wrapping in isolation. After looking back at the code I copied and the discussions in other projects who adopted it, it appears that

```
@media screen and (min-width: 767px) {
```

is supposed to ensure the overflow only goes into effect if the width of the browser meets a minimum threshold (thus it shouldn't be a problem even on vertical monitors or with terminal beside the browser). If you shrink the window enough (down to about 749 px on my screen), two things happen: The left nav bar collapses, and the forced overflow automatically turns off (and you can scroll again to see the whole table. Hooray! So it actually works for both very large and very small screens.

Unfortunately, if you're above 767 px, but only by a little bit, overflow is turned off, but the left nav bar takes up 320 px, so you have overflow=True in the ~447 px area that remains (which isn't enough to display the table).

I tried playing with the numbers some, but even with 1087 (320 from left nav bar + 767 that other projects use) there's still some mid-size configurations where it looks weird (though more small screens look good, and it still helps on the very large). HTML/CSS is really not my strong suit, as I mentioned in the related issue, so while I'm sure there is a more intelligent way to enable it on only the screens where it helps, it's eluding me so far.

docs/source/_static/hackrtd.css
```diff
@@ -110,3 +110,20 @@ a.trio-help-hint, .trio-help-hint a:link, .trio-help-hint a:visited {
     text-decoration: none;
     border-bottom: 1px solid;
 }
+
+/* override table width restrictions */
+@media screen {
+
+   .wy-table-responsive table td {
+      /* !important prevents the common CSS stylesheets from overriding
+         this as on RTD they are loaded after this stylesheet */
```
pquentin: We don't need a comment to explain what !important is here, we're in a CSS file so we assume the readers who don't know what it is can look it up.

Comments are useful when they explain *why* we're doing a change.

docs/source/_static/hackrtd.css
```diff
@@ -110,3 +110,20 @@ a.trio-help-hint, .trio-help-hint a:link, .trio-help-hint a:visited {
     text-decoration: none;
     border-bottom: 1px solid;
 }
+
+/* override table width restrictions */
```
pquentin: Can we move the comment for `th` and place it here to explain both the td and th changes?

pquentin: Something like "Since we have long names compared to other projects, allow word-wrapping in header cells and data cells to avoid too large tables"? (I'm not a native speaker, feel free to rephrase or put something else entirely.)

PR #1235, opened 2019-10-06
Don't crash when CancelScopes are nested *very* deeply

njsmith:
This will probably never come up in real life, but I hit it while
messing around with some microbenchmarks, and the fix is easy.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -253,21 +253,23 @@ class CancelStatus:
         )
 
     def recalculate(self):
-        # This is basically a recursive algorithm, but we do it iteratively to
-        # avoid any issues with stack overflow.
+        # This does a depth-first traversal over this and descendent cancel
+        # statuses, to ensure their state is up-to-date. It's basically a
+        # recursive algorithm, but we use an explicit stack to avoid any
+        # issues with stack overflow.
         todo = [self]
         while todo:
-            cs = todo.pop()
+            current = todo.pop()
             new_state = (
-                cs._scope.cancel_called
-                or cs.parent_cancellation_is_visible_to_us
+                current._scope.cancel_called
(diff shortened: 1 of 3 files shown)


njsmith:
Note: I noticed this while messing around with microbenchmarks for https://github.com/python-trio/trio/pull/1234

trio/_core/_run.py
```diff
@@ -253,17 +253,21 @@ def parent_cancellation_is_visible_to_us(self):
...
-            if new_state:
-                for task in self._tasks:
-                    task._attempt_delivery_of_any_pending_cancel()
-            for child in self._children:
-                child.recalculate()
+        # This is basically a recursive algorithm, but we do it iteratively to
+        # avoid any issues with stack overflow.
+        todo = [self]
+        while todo:
+            cs = todo.pop()
```
wgwz: Hey, @njsmith. Just to clarify `cs` should read cognitively as `cancel_scope`, right?

njsmith: Yeah, that's what I was thinking.

wgwz: I think you should increase the verbosity here, and expand out to `cancel_scope` for readability. Especially because the docstring for this class mentions "cancelled status" and "cancel scope" -> both valid interpolations for `cs`.

oremanj: `cs` here is in fact a CancelStatus object, not a CancelScope -- another reason to spell it out? :-)

njsmith: Ok, I renamed it `current` to describe its purpose rather than its type :-)

njsmith: OK, fair points :-) I renamed it `current` to describe its role in the algorithm, rather than its type.

pquentin:
Thank you!

PR #1252, opened 2019-10-16
Make testing instructions less generic

zackw:
In `contributing.rst`, the instructions for running tests locally appear to have been cut-and-pasted from the pytest manual, or something like that, and it's not clear exactly what to type.  Replacing "project" with "trio" in two places fixes this.  In particular, I have verified that `pytest trio` is the correct command to run the test suite once everything is set up.

Also use a code-block instead of a bulleted list of verbatim strings.

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -184,10 +184,11 @@ Tests
 We use `pytest <https://pytest.org/>`__ for testing. To run the tests
 locally, you should run:
 
-* ``cd path/to/project/checkout/``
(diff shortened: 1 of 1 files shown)


zackw:
I don't understand how a change that touches only the documentation could possibly have caused code coverage changes :-(

pquentin:
We use codecov to merge coverage reports across platforms, and it's not 100% reliable. Sometimes we miss a platform or two in a commit, so the coverage drops, only to raise again in the next PR. That's unfortunate but there's no easy way to do better than this.

Anyway, thank you for your contribution!

PR #1254, opened 2019-10-17
[WIP] First pass of mypy-cleanliness

oremanj:
The goal of this PR is to make `mypy -p trio` work. 

WIP because there are still some errors:
* "cmp is deprecated, use eq and order" thrown by attrs plugin: will be resolved by upgrading attrs
* Path needs a stubs file
* 3-4 others that I ran out of time to fix on the flight

And we should add mypy to CI once it's working so it stays working.


pquentin:
> "cmp is deprecated, use eq and order" thrown by attrs plugin: will be resolved by upgrading attrs

We need a release of https://github.com/python-trio/outcome/pull/25 before we can fix attrs in trio proper

pquentin:
Okay, the attrs fix has landed in master. :)

oremanj:
Superseded by recent work

PR #1255, opened 2019-10-17
Test Python 3.8 release

pquentin:


njsmith:
It looks like the upstream bug for adding 3.8 to azure images is: https://github.com/microsoft/azure-pipelines-image-generation/issues/1317

pquentin:
Okay, closing for now. :) It's Windows that I want to see the most so that we can cover a few more lines.

Issue #1259, opened 2019-10-17
Path should get the `__hash__` from `pathlib.PurePath`


georgeaf99:
`trio.Path`s don't exhibit the same hashing behavior as `pathlib.PurePath`

Repro (`python3.6.5` and `trio (0.11.0)`):
```
>>> from trio import Path
>>> hash(Path("foo"))
8743626168503
>>> hash(Path("foo"))
8743626168510
```

oremanj:
Fixed by #1260. (@georgeaf99 FYI: if you put "Closes #1259" in the PR description, Github will close the issue automatically when the PR is merged.)

PR #1260, opened 2019-10-17
Fix Path's hash behavior to match `pathlib.Path`'s

georgeaf99:
See https://github.com/python-trio/trio/issues/1259

New Behavior:
- `hash(Path("foo/bar/baz")) == hash(Path("foo/bar/baz"))`
- `hash(trio.Path("foo/bar/baz")) == hash(pathlib.Path("foo/bar/baz"))`


oremanj:
Thanks for fixing this! The code change looks good but there are some associated updates we like to see with every change:
- update the unit tests so they would have noticed this bug
- run yapf to standardize formatting on the code you edited (this will fix the current CI failure)
- add a newsfragment describing the user-visible impacts of the change

More detail on all of these can be found in our contributing guide: https://trio.readthedocs.io/en/stable/contributing.html

Thanks again!

georgeaf99:
Thanks for the heads up. I've updated the PR to address your comments.

PR #1263, opened 2019-10-19
`serve_tcp` is used in comment while it should be  `trio.serve_tcp` in tutorial code

millefalcon:


njsmith:
Nice catch, thanks :-)

millefalcon:
Thanks : )

Issue #109, opened 2017-03-22
Maybe use signal.set_wakeup_fd on unix-likes


njsmith:
As of #108, we now use `set_wakeup_fd` to ensure that when a signal that we have a handler for is delivered, it causes our IO loop to actually wake up and run the handler ... on Windows.

On Unix, this is *mostly* unnecessary, but there are some edge cases where it matters:

* When entering epoll/kqueue, there's a race condition when a signal arrives in between the last time the interpreter checks for signals and when it calls `epoll_wait`. This is partially a bug in CPython – on Linux it could be eliminated by masking the signal, checking for the signal, and then doing `epoll_pwait`. There doesn't seem to be an equivalent for kqueue, though.

* If the kernel decides to deliver a signal to a non-main thread, then CPython doesn't run the handler until the main thread re-enters Python. I think this is a bug in CPython ([issue21895](https://bugs.python.org/issue21895)), but we might still want to work around it. The only time this is known to happen in practice is that when a child thread spawns a subprocess, then on Linux the SIGCHLD gets delivered to the child thread.

Neither of these are super-urgent (one's a rare race condition, and the other only seems to happen with SIGCHLD, which we don't plan to use), but it might be nice to fix them anyway.

Using `set_wakeup_fd` would fix them... except that it generates spurious warning messages when the wakeup fd buffer is full, and there's now way to stop it :-(.

njsmith:
https://github.com/python/cpython/pull/4792 has been merged, so I think we should:

- always use `set_wakeup_fd`
- use `warn_on_full_buffer=False` where available
- use large buffers everywhere else

We should also hassle PyPy about implementing `warn_on_full_buffer=False` (or just submit a patch)

PR #1267, opened 2019-10-24
Move set_wakeup_fd into common code, used on all platforms

njsmith:
Also switched signal_raise to always run the C-level signal handler in
the current thread, which gives us more control over signal delivery
and lets us test the new behavior.

Fixes gh-109

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1783,9 +1783,11 @@ def run(
         ):
             try:
                 with closing(runner):
-                    # The main reason this is split off into its own function
-                    # is just to get rid of this extra indentation.
-                    run_impl(runner, async_fn, args)
+                    with runner.entry_queue.wakeup.wakeup_on_signals():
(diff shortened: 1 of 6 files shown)


trio/_core/_run.py
```diff
@@ -1783,9 +1783,11 @@ def run(
...
             try:
                 with closing(runner):
-                    # The main reason this is split off into its own function
-                    # is just to get rid of this extra indentation.
-                    run_impl(runner, async_fn, args)
+                    with runner.entry_queue.wakeup.wakeup_on_signals():
+                        # The main reason this is split off into its own
+                        # function is just to get rid of this extra
+                        # indentation.
+                        run_impl(runner, async_fn, args)
```
Tronic: Wrapping a comment about indentation width to a third line because of the column limit? I wish the world already switched to 120 characters instead of the width of 1980's graphics adapters and terminals.

Issue #399, opened 2018-01-12
Using trio with Qt's event loop?


nicoddemus:
Hi,

Awesome library, I've been studying it and the care with its design and implementation is something to behold, congratulations to @njsmith and the team!

At work we use Qt to develop desktop applications and I was wondering if it is possible to use Qt's event loop instead of the internal `trio` event loop, something like [quamash](https://github.com/harvimt/quamash) does for `asyncio`.

njsmith:
Hello!

> Awesome library, I've been studying it and the care with its design and implementation is something to behold, congratulations to @njsmith and the team!

Thanks! And sorry for the slow response...

> At work we use Qt to develop desktop applications and I was wondering if it is possible to use Qt's event loop instead of the internal `trio` event loop, something like quamash does for asyncio.

Yeah, Trio+GUI is an important question! Also... kind of complicated. Right now there isn't anything implemented. Obviously this should change. I have very little experience writing GUI apps, so I'm not sure I know enough to recognize a good solution when I see it. So it's tricky.

Well, if you look around the issue tracker, you'll see that Trio follows a thinking-out-loud-driven development methodology, and we don't have a ticket yet for GUI stuff, so let's start with a brain dump of some things I've been thinking about :-).

First, if you just want to get something working, the shortest path to do that will be to run Qt in one thread and Trio in another. It's not too hard to build some infrastructure to call between them, similar to what `trio.BlockingTrioPortal` does for calling between Trio and a thread running synchronous code – I can give you some tips if you like. This isn't exactly ideal, basically because it involves threads and threads are never ideal. In particular, if you have actual logic running simultaneously in the Qt thread and the Trio thread, then you have to take all the usual care about them stepping on each other toes, which is too bad since one of the major advantages of cooperative scheduling is that it [makes it easier to reason about interleavings](https://glyph.twistedmatrix.com/2014/02/unyielding.html).

That said, people do manage to write threaded programs all the time, and this probably works great for lots of cases, especially if you do things like push all the complex state manipulation onto the Trio side and restrict the Qt side to just dealing with the GUI and dispatching events to the Trio side. I'll also CC @matham here, who's been doing something similar with Kivy instead of Qt, and might be able to offer some comments. If you just want to get something working, stop here, this is definitely the path of least resistance.

----

[...]

PR #1269, opened 2019-10-25
Rewrite Windows backend to use IOCP exclusively

njsmith:
This dramatically simplifies the code (no more janky background
thread!), and probably makes it substantially faster (though I haven't
measured), since we're not using `select` anymore or passing events
back and forth between threads.

The way we accomplish this is basically to reimplement `select`
ourselves, using the undocumented ``IOCTL_AFD_POLL`` mechanism to talk
to the kernel directly. Other projects that use a similar approach are
libuv, wepoll, and mio. See #52 for (lots) more discussion.

Since this is undocumented, it's hard to be 100% certain we've
understood everything, but let's see what the tests say :-). I also
added some tests that install "layered service providers", which are
this weird system on Windows that lets random libraries monkeypatch
how sockets work; I think we're doing all the stuff we need to to be
compatible with these things, but again, we need to check that
empirically.

This also unblocks the idea of allowing Trio to run "over" a foreign
event loop, as discussed in https://github.com/python-trio/trio/issues/399#issuecomment-406200369


trio/_core/_io_windows.py
```diff
@@ -97,219 +222,280 @@ class CompletionKeyEventInfo:
...
+            if entry.lpCompletionKey == CKeys.AFD_POLL:
+                lpo = entry.lpOverlapped
+                op = self._afd_ops.pop(lpo)
+                waiters = op.waiters
+                if waiters.current_op is not op:
+                    # Stale op, nothing to do
+                    pass
+                else:
+                    waiters.current_op = None
+                    # I don't think this can happen, so if it does let's crash
```
oremanj: Any reason not to raise it as the result of `wait_readable` or `wait_writable`? It's hard to track down what tasks were doing when a TrioInternalError occurred (although I know we're hoping to fix that more generally at some point).

trio/_core/_io_windows.py
```diff
@@ -97,219 +222,280 @@ class CompletionKeyEventInfo:
...
+    async def _afd_poll(self, sock, mode):
+        base_handle = _get_base_socket(sock)
+        waiters = self._afd_waiters.get(base_handle)
+        if waiters is None:
+            waiters = AFDWaiters()
+            self._afd_waiters[base_handle] = waiters
+        if getattr(waiters, mode) is not None:
+            raise _core.BusyResourceError
+        setattr(waiters, mode, _core.current_task())
+        self._refresh_afd(base_handle)
```
oremanj: If _refresh_afd raises for some reason, you're definitely not going to be in a good state. Do you think it's worth trying to recover? I was worried initially that there might be an issue if `sock` is invalid, but I think `_get_base_socket` will raise in that case.

trio/_core/_io_windows.py
```diff
@@ -350,19 +532,17 @@ def abort(raise_cancel_):
                     # does, we'll assume the handle wasn't registered.
```
oremanj: In this comment block there's a reference to the IOCP thread which needs to be updated

trio/_core/_io_windows.py
```diff
@@ -73,22 +82,138 @@
...
+#
+# There's also a AFD_POLL_LOCAL_CLOSE that we could wait on, to potentially
+# catch some cases where someone forgot to call notify_closing. But it's not
+# reliable – e.g. if the socket has been dup'ed, then closing one of the
+# handles doesn't trigger the event – and it's not available on Unix-like
+# platforms. So it seems like including it here would be more likely to mask
+# subtle bugs than to actually help anything.
+
+READABLE_FLAGS = (
+    AFDPollFlags.AFD_POLL_RECEIVE
```
oremanj: Should `RECEIVE_EXPEDITED` be in here too? (I'm guessing that's for TCP urgent data)

trio/_core/_io_windows.py
```diff
@@ -97,219 +222,280 @@ class CompletionKeyEventInfo:
...
+    async def wait_readable(self, sock):
+        await self._afd_poll(sock, "read_task")
+
+    @_public
+    async def wait_writable(self, sock):
+        await self._afd_poll(sock, "write_task")
+
+    @_public
+    def notify_closing(self, handle):
+        handle = _get_base_socket(handle)
```
oremanj: I wonder if it's worth marking these `_get_base_socket` calls with some indication (a comment or a chained exception) that failure usually means the argument is not a socket. I figure that's the first thing that will go wrong if someone tries to use a library that does Unix-specific `wait_readable` calls on Windows, so it might pay to be user-friendly here.

trio/_core/tests/test_io.py
```diff
@@ -273,3 +273,65 @@ def fileno_wrapper(fileobj):
...
+            nursery.start_soon(_core.wait_readable, a3)
+
+            await wait_all_tasks_blocked()
+
+            statistics = _core.current_statistics()
+            print(statistics)
+            # 1 for call_soon_task
+            check(statistics.io_statistics, 4, 2)
+
+            nursery.cancel_scope.cancel()
```
oremanj: `check(_core.current_statistics().io_statistics, 1, 0)` again after the nursery block?

I also wonder if it's worth having a version of this test that wakes up the waiters normally (write data on the other side to wake the readers, read from the other side to wake the writers) rather than cancelling them, so we get coverage on both paths.

trio/_windows_pipes.py
```diff
@@ -115,6 +115,12 @@ def __init__(self, handle: int) -> None:
                 # whenever the other end closes, regardless of direction.
                 # Convert this to the Unix behavior of returning EOF to the
                 # reader when the writer closes.
+                #
+                # And since we're not raising an exception, we have to
+                # checkpoint. But readinto_overlapped did raise an exception,
+                # so it might not have checkpointed for us. So we have to
+                # checkpoint manually.
+                await _core.checkpoint()
```
oremanj: Good catch!

trio/_core/_io_windows.py
njsmith: Right, `RECEIVE_EXPEDITED` is for urgent/OOB data. And we definitely don't want to treat the socket as readable every time we receive urgent/OOB data: https://sandstorm.io/news/2015-04-08-osx-security-bug

trio/_core/_io_windows.py
njsmith: For `notify_closing` I think we want to eventually accept non-SOCKET handles: #1270.

You're right though that we should probably have a test that `wait_*` on non-SOCKETs gives a sensible result...

trio/_core/tests/test_io.py
njsmith: I don't think we need to fret too much over this test; it's just a quick check that some rarely-used almost-trivial debug code works properly. If things are so broken that the IOManager can't track cancellation correctly, then hopefully we have other tests that will notice that :-)

njsmith: I did do a cleanup pass and add the check at the end though.

trio/_core/_io_windows.py
njsmith: It looks like the error we get currently is:

```
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

Which I guess is pretty clear on it not being a socket :-)

trio/_core/_io_windows.py
njsmith: We also call `_refresh_afd` from abort function context, and `handle_io` context, neither of which are the slightest bit recoverable. And we don't know any way that it could fail, I think?

....well, hmm. I guess in those two contexts we actually don't revalidate the handle before calling it. I guess the only way you could reach those points with an invalid handle is if you manually close a socket without calling `notify_closing`, so arguably we could say "not our problem". But I wonder if I can construct a case where it happens. I guess you need to do something like:

1. Task A calls `wait_readable`
2. Task B calls `wait_writable`
3. Task C closes the socket without telling us
4. Task A gets cancelled while leaving task B still running

I guess that could happen by accident – accidentally forget to close a socket, hit control-C to get out... hmm.

njsmith: ...It turns out that this is quite a rabbit-hole. This is already a big complicated PR, so let's defer worrying about this for a follow-up issue: #1272

njsmith: Though I should also say, for purposes of this PR: while we should follow up, merging this as-is won't be terrible. Our epoll backend is structured the same way, so merging this won't create any *new* issues, plus it means we know that no-one's ever hit the issues in practice.

trio/_core/_io_windows.py
njsmith: I thought about this a lot and made an attempt at implementing it, but it turns out to be non-trivial to wake everyone up here while keeping our internal state consistent. And that's code that we can't even test, because we don't know any way to trigger this error. Also, since we have no idea how we could hit this error, it's entirely possible that if it happens, things will be more generally broken, and trying to resume the scheduler will create a deadlock or something. So, I think I'd rather leave it as a `TrioInternalError` crash for now. (Fortunately I'm pretty sure exceptions from inside `handle_io` do come out directly from `trio.run`.)

njsmith:
OK, I think I've addressed all the comments and all the other cleanups I noticed, so this is ready for review again.

trio/_core/_io_windows.py
oremanj: Works for me!

trio/_core/_io_windows.py
oremanj: great!

newsfragments/52.feature.rst
```diff
@@ -1,4 +1,16 @@
...
-of a hybrid of ``select`` + IOCP. This should make it much faster and
-more scalable. It also simplifies the code internally, and paves the
-way for future improvements.
+On Windows, the `IOCP subsystem
+<https://docs.microsoft.com/en-us/windows/win32/fileio/i-o-completion-ports>`__
+is generally the best way to implement async I/O operations – but it's
+historically been weak at providing ``select``\-style readiness
+notifications, like `trio.hazmat.wait_readable` and
+`~trio.hazmat.wait_writable`. We aren't willing to give those up, so
+Trio's Windows backend used to use a hybrid of ``select`` + IOCP. This
```
oremanj: something seems confusing to me about "used to use". maybe "previously used"?

trio/_core/_io_windows.py
```diff
@@ -32,57 +33,136 @@
...
-#   - the "completion key" (an integer)
-#   - pointer to OVERLAPPED
-#   - dwNumberOfBytesTransferred
-# - and in addition, for regular I/O, the OVERLAPPED structure gets filled in
-#   with:
-#   - result code (named "Internal")
-#   - number of bytes transferred (named "InternalHigh"); redundant with
-#     dwNumberOfBytesTransferred *if* this is a regular I/O event.
+#
+# The general model is that you call some function like ReadFile or WriteFile
```
oremanj: Thank you for updating this comment, it's super helpful!

notes-to-self/afd-lab.py
```diff
@@ -0,0 +1,173 @@
+# A little script to experiment with AFD polling.
```
oremanj: Is it possible for the IOCP-based I/O manager implemented in this PR to run into the bug elicited by this script? If not, maybe comment why not?

njsmith: Right, this bug is why we have all this machinery to avoid ever issuing more than one IOCTL_AFD_POLL operation on the same socket at the same time :-). Updated comments in both `afd-lab.py` and `_io_windows.py`.

njsmith:
Test failure is codecov flakiness (#1231)

njsmith:
There was some discussion of scaling problems with `CancelIoEx`, starting here: https://github.com/python-trio/trio/issues/52#issuecomment-548215128

It looks like the conclusion for now is that this PR is still an improvement over what we had before, but we probably want a follow-up issue to consider splitting up our AFD_POLL operations across multiple AFD helper handles, to improve scalability even more.

njsmith:
OK, let's see how this goes! Thanks everyone for your comments and feedback so far!

PR #1285, opened 2019-11-03
Make Trio more robust to fds/handles being closed behind our back

### LINKED ISSUE #1272: Be more robust against file descriptors/handles disappearing under our feet
In [this discussion](https://github.com/python-trio/trio/pull/1269#discussion_r339242694) we discovered that the epoll and Windows backends can get into quite a mess if a file descriptor/socket handle gets closed under our feet without telling us – deadlocks, loop crashes, etc. Technically this isn't our problem, because the user should have called `notify_closing`. But the experience would be better if we can provide some better error messages and survive, instead of locking.

Here's an example of one of the nasty cases:

```python
a, b = stdlib_socket.socketpair()
with a, b, a.dup() as a2:
    a.setblocking(False)
    b.setblocking(False)
    fill_socket(a)
        nursery.start_soon(trio.hazmat.wait_readable, a)
        nursery.start_soon(trio.hazmat.wait_writable, a)
        a.close()
        nursery.cancel_scope.cancel()
```

Since we dup'ed `a`, the underlying kernel object survives the call to `a.close()`. So the `wait_readable`/`wait_writable` tasks keep going, and if we manipulated `b` we could even make them wake up as normal. But instead we cancel. Whichever task is cancelled first attempts to modify the underlying primitive to only wait for the other task, but the fd it uses to do this has disappeared under its feet, and things go bad quickly.

This case is particularly nasty on the epoll backend, because there it's actually impossible to deregister our interest in `a`: we're going to keep getting events for `a` whether we want them or not. The only way to survive is to close our epoll descriptor and rebuild it. This is possible, but I'm not sure whether it's actually better than crashing... The thing is that we can't really detect these situations reliably. It's possible in this particular test case, but in a more complex program the file descriptor might get reassigned between the call to `close` and the call to `cancel`, in which case the user will silently start notifications sent to the wrong task. So even if we can do some kind of best-effort recovery, we still need to somehow tell users to fix their code, not just pretend that everything is fine.

njsmith:
Fixes #1272

Todo:

- [x] Windows coverage is a bit wonky -- there's code in `wake_all` to handle the case where `_refresh_afd` failed inside an issuing task, but in fact that never happens in our tests, because in the cases where it would happen `_get_base_handle` crashes first. So this code seems to be unreachable. Should we remove it? Keep it but mark it `no cover`? Not sure. [Edit: better idea – the `wake_all` code was identical between `_io_windows.py` and `_io_epoll.py`, so I just factored it out into a shared file, and now they share testing.]

- [x] Newsfragment

- [x] See what happens on macOS – hopefully nothing because kqueue is simple and friendly, but we'll see what the CI says! [Edit: ok it wasn't quite that simple, but it's working now]

### DIFF
--- a/trio/_core/tests/test_io.py
+++ b/trio/_core/tests/test_io.py
@@ -418,6 +418,7 @@ async def test_can_survive_unnotified_close():
         b.setblocking(False)
         fill_socket(a)
         e = trio.Event()
+
         async def wait_readable_then_set():
             await trio.hazmat.wait_readable(a)
             e.set()
--- a/trio/_core/tests/test_io.py
+++ b/trio/_core/tests/test_io.py
@@ -438,7 +438,7 @@ async def test_can_survive_unnotified_close():
         async with trio.open_nursery() as nursery:
             nursery.start_soon(allow_OSError, trio.hazmat.wait_readable, a)
             nursery.start_soon(allow_OSError, trio.hazmat.wait_writable, a)
-            nursery.start_soon(allow_OSError, wait_readable_a2_then_set)
+            nursery.start_soon(wait_readable_a2_then_set)
             await wait_all_tasks_blocked()
             a.close()
             b.send(b"x")
--- a/trio/_core/tests/test_io.py
+++ b/trio/_core/tests/test_io.py
@@ -419,13 +419,20 @@ async def test_can_survive_unnotified_close():
         fill_socket(a)
         e = trio.Event()
 
-        async def wait_readable_then_set():
-            await trio.hazmat.wait_readable(a)
+        # We want to wait for the kernel to process the wakeup on 'a', if any.
+        # But depending on the platform, we might not get a wakeup on 'a'. So
+        # we put one task to sleep waiting on 'a', and we put a second task to
+        # sleep waiting on 'a2', with the idea that the 'a2' notification will
+        # definitely arrive, and when it does then we can assume that whatever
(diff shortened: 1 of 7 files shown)


njsmith:
OK, I think this is finished and ready for review!

It's a bit complex, sorry... though I've learned all kinds of interesting things in the process :-).

To review, I'd start with `test_io.py`, which adds the new tests. And then the changes to `_io_kqueue.py`, `_io_windows.py`, and `_io_epoll.py` are essentially independent changes needed to pass the new tests, and can be reviewed in whatever order you like.

trio/_core/tests/test_io.py
```diff
@@ -352,3 +351,93 @@ def check(*, expected_readers, expected_writers):
...
+        # definitely arrive, and when it does then we can assume that whatever
+        # notification was going to arrive for 'a' has also arrived.
+        async def wait_readable_a2_then_set():
+            await trio.hazmat.wait_readable(a2)
+            e.set()
+
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(allow_OSError, trio.hazmat.wait_readable, a)
+            nursery.start_soon(allow_OSError, trio.hazmat.wait_writable, a)
+            nursery.start_soon(allow_OSError, wait_readable_a2_then_set)
```
oremanj: You're not closing a2, so I don't think this needs allow_OSError?

njsmith: Good catch :-)

Issue #1241, opened 2019-10-08
use pidfd on Linux when available


njsmith:
Recent versions of the Linux kernel finally added a much more reasonable way to interact with subprocesses called "pidfd". The main advantages are:

- Using an fd as a handle for interacting with a process avoids a bunch of weird edge cases that pids have, like pids being recycled out from under you so you accidentally kill the wrong process (this isn't an issue for the specific way trio interacts with subprocesses, because we only work with child processes and the relationship between a parent and child is special, but in more general cases it's an issue)
- You can wait for child processes using `epoll`!!!!
- There may be more fancy features coming that are built on top of pidfd, like the ability to automatically kill a child when the last pidfd reference to it is closed.

Probably the epoll integration is the first thing we care about, since it provides an alternative to the rube goldberg thread-based construction we use now. Of course we'll still have to maintain that indefinitely because of RHEL and friends, but it'll be nice to avoid it when running on a modern system. That also means this isn't like, a huge urgent priority, because the user-visible impact is basically nil. But we'll want it eventually, so might as well have an issue to collect information until we're ready.

Initial pidfd support landed in kernel 5.1; the epoll integration landed in 5.3. And Ubuntu 19.10 (out next week) will have a 5.3 kernel, so pidfd is starting to actually filter out to users. I believe that in 19.10 there still aren't any glibc wrappers for the new syscalls, so if you want to access them you have to [call `syscall` directly](https://manpages.courier-mta.org/htmlman2/syscall.2.html). We could do that with ctypes or something though – it's janky but not difficult.

The biggest blocker here is that we will definitely need to test the new code, and AFAIK none of the CI services we have access to will ship a pidfd-capable kernel for... a while. They all seem to stick to LTS releases, and sometimes they're slow to upgrade even then... Azure is still on Ubuntu 16.04, though Travis and Github Actions both have 18.04. So probably we won't be able to actually *test* pidfd support until sometime after Ubuntu 20.04 LTS is released in April 2020. (Unfortunately since this is a kernel feature, using docker to pull down a newer release doesn't help, and I'm not aware of any hosted CI services that let you bring-your-own-kernel.)

njsmith:
Python 3.9 will have a `os.pidfd_open` wrapper built-in: https://github.com/python/cpython/pull/17063

It's pretty simple so we might end up ignoring this and using cffi instead, in order to support older Python versions, but wanted to mention for reference.

benjaminp:
[...]

Issue #932, opened 2019-02-14
Consider io_submit as an alternative to epoll


thedrow:
`io_submit()` is not async but it allows batching of syscalls which may or may not be better for a user's usecase.

Should we introduce an IOManager which utilizes it?

Read more [here](https://blog.cloudflare.com/io_submit-the-epoll-alternative-youve-never-heard-about/).

pquentin:
Relevant discussion in chat: https://gitter.im/python-trio/general?at=5c2ff4d65ec8fe5a850242ac:

> I am skeptical that micro-optimizations to reduce syscalls or copying are a measurable win in python
> But I could well be wrong :-)
> And the trick of using IOCB_CMD_POLL is super clever
> (and actually equivalent to what we're going to do on Windows)

njsmith:
There's also the upcoming `io_uring`: https://lwn.net/Articles/776703/

Some practical challenges: Python ships with a wrapper for epoll, but not for `io_submit`. Currently we don't use cffi or compiled modules on Linux, but we could reconsider that if it makes sense. Also... we don't currently have any good benchmarks, which are sort of important for figuring out whether this is a win or not :-)

As a general rule, Trio is much less keen to offer configurable/pluggable `IOManager`s than some of the more well-known Python event loop libraries. But if this turns out to be a worthwhile change, then I don't see why it wouldn't be worthwhile for everyone!

It's not something I'll personally be looking at in the foreseeable future, [because well see this list](https://trio.discourse.group/t/priorities-roadmap/34) :-). But if you do any experiments then I will be super curious to see how it goes!

thedrow:
I won't be looking into this soon.
Maybe after we finish rewriting Celery to use trio.

njsmith:
> Maybe after we finish rewriting Celery to use trio.

Is the discussion around this happening in public somewhere? Anything we can help with?

thedrow:
@njsmith It will soon. Right after we finish with releasing 4.3.

njsmith:
I asked Jens Axboe [about io_uring on twitter](https://twitter.com/vorpalsmith/status/1184963233368526848), and got some more details (click on each message in turn to see the subthreads).

The most interesting part is that he sounds serious about providing a solid async interface to disk I/O. So it will be interesting to see how that shapes up...

njsmith:
[...]

PR #1292, opened 2019-11-06
[wip] [ci] Run tests in a full-fledged ubuntu 19.10 VM

njsmith:
I'm not sure if this is actually useful... I just got curious about it
today and it came together more easily than I expected, so now I'm
curious if it actually works :-).

Rumor says that Travis supports nested KVM acceleration now, if you're
using their 'bionic' image, so in theory this should work and be
reasonably fast?

Right now, it isn't especially useful. But if we started adding
support for new Linux kernel features like pidfd (#1241) or
io_uring (#932), then this would become important, because AFAICT this
is the only way to run tests on new kernels on any of the hosted CI
services.

I guess we could also use it to like, test on ARM or something, if we
cared? I experimented with testing on 32-bit, but the test setup
failed because cryptography has stopped shipping 32-bit wheels, and I
couldn't be bothered to figure out how to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-dist: bionic
+dist: trusty
 
 matrix:
   include:
@@ -18,6 +18,9 @@ matrix:
     # early warning of any issues that might happen in the next Ubuntu
     # LTS.
     - language: generic
+      # We use bionic for the host, b/c rumor says that Travis's
+      # 'bionic' systems have nested KVM enabled.
+      dist: bionic
       env:
         - "JOB_NAME=Ubuntu 19.10, full VM"
         - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-dist: xenial
+dist: bionic
 
 matrix:
   include:
@@ -11,9 +11,20 @@ matrix:
     - python: pypy3
     - language: generic
       env: PYPY_NIGHTLY_BRANCH=py3.6
+    # Qemu tests are also slow
+    # The unique thing this provides is testing on the given distro's
+    # kernel, which is important when we use new kernel features. This
+    # is also good for testing the latest openssl etc., and getting
+    # early warning of any issues that might happen in the next Ubuntu
+    # LTS.
+    - language: generic
+      env:
+        - "JOB_NAME=Ubuntu 19.10, full VM"
+        - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
     # 3.5.0 and 3.5.1 have different __aiter__ semantics than all
     # other versions, so we need to test them specially. Travis's
-    # xenial dist only provides 3.5.2+.
+    # newer images only provide 3.5.2+, so we have to request the old
+    # 'trusty' images.
     - python: 3.5.0
       dist: trusty
     - python: 3.5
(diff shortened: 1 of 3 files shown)


njsmith:
Since the Trio run loop is essentially a kernel for a specialized python machine, I guess this is testing a virtual machine, running inside a virtual machine, running inside a virtual machine, running inside a real machine.

pquentin:
Just trying to understand: does that mean we were not testing the latest Ubuntu in the py37_latest_ubuntu Azure Pipelines job? I'm asking because the new setup seems to be more complicated

njsmith:
Because of how docker works, we were testing the latest Ubuntu userspace, but still using the CI system's kernel (which for Azure Pipelines I think is either from Ubuntu 16.04 or Ubuntu 18.04). For 99% of things that's fine, but if you want to test new kernel features, then you need a new kernel. This approach tests the latest Ubuntu userspace *running on the latest Ubuntu kernel*, so it does everything the docker approach does, and a bit more.

It's definitely more complicated and slower, though, so it depends on how much we care about testing on newer kernels. If we were to say, add a whole new `io_uring` backend, then we would care a lot :-). Right now though it's not so obvious.

pquentin:
Ah, yes, that makes sense! Okay, so we have a complicated thing that took some time to get right, but now works and might be useful in the feature. Per the sunken cost fallacy, I think we should merge this!

But first, are we sure that our codecov setup still works? I still see references to `CODECOV_NAME` in ci.sh.

auvipy:
does ubuntu LXD containers are of any use for it? https://github.com/fnordahl/travis-lxd , https://docs.travis-ci.com/user/reference/overview/ , https://fnordahl.com/2017/04/08/continuous-deployment-testing-of-github-hosted-code/

njsmith:
@pquentin 
> Ah, yes, that makes sense! Okay, so we have a complicated thing that took some time to get right, but now works and might be useful in the feature. Per the sunk cost fallacy, I think we should merge this!

Heh :-). I was also thinking maybe we should close it, and then in the future if we have a need for it we'll know it's here and can resurrect the branch.

> But first, are we sure that our codecov setup still works? I still see references to CODECOV_NAME in ci.sh.

Whoops, good catch, thanks!

@auvipy 
> does ubuntu LXD containers are of any use for it?

I believe that LXD containers use the same basic technology as docker, i.e., the containers all share the same underlying host kernel.

.travis.yml
```diff
@@ -11,9 +11,23 @@ matrix:
...
+    # kernel, which is important when we use new kernel features. This
+    # is also good for testing the latest openssl etc., and getting
+    # early warning of any issues that might happen in the next Ubuntu
+    # LTS.
+    - language: generic
+      # We use bionic for the host, b/c rumor says that Travis's
+      # 'bionic' systems have nested KVM enabled.
+      dist: bionic
+      env:
+        - "JOB_NAME=Ubuntu 19.10, full VM"
```
pquentin: codecov does not like spaces here, and only shows "Ubuntu" - see https://codecov.io/gh/python-trio/trio/commit/7cabd162b3fe04bc1465cecca0a7b07804567210/build

pquentin: However it can't be the exact issue as we have the spaces for the Azure Pipeliens jobs.

njsmith: Hmm, it looks like something is going wrong at the Travis level... at the very top of the file, it shows the initial environment for the job as created by travis, and [that says](https://travis-ci.org/python-trio/trio/jobs/608056847#L216):

```
JOB_NAME=Ubuntu
```

Maybe I need to add more quotes to the YAML or something?

njsmith: Yeah, adding more quotes seems to have fixed it.

pquentin: Nice!

pquentin:
Thanks, merged!

PR #1296, opened 2019-11-08
Use pidfd to implement trio.Process.wait, when available

### LINKED ISSUE #1241: use pidfd on Linux when available

njsmith:
Fixes #1241

The new QEMU-based CI run (#1292) *should* let us test this. I'm curious
to see what codecov says :-).

But unfortunately, even with that, we're still only testing the
ctypes-based configuration, not the configuration that uses the stdlib
`os.pidfd_open` that currently only exists in CPython master. We're
running tests on CPython master, and we're running tests on a new
kernel, but we aren't doing both at the same time. AFAICT the only way
to test that configuration would be to check out and build CPython
master from inside the nested VM, which sounds like a real hassle.
OTOH, I'm not super comfortable about shipping code that hasn't been
tested... `os.pidfd_open` is pretty simple but it's always possible my
ctypes-based implementation has some subtle behavioral difference I
missed.

Also, maybe we should add some tests to confirm the pidfd is cleaned
up properly?

CC: @benjaminp

### DIFF
--- a/trio/_subprocess.py
+++ b/trio/_subprocess.py
@@ -22,7 +22,7 @@ except ImportError:
         import ctypes
         _cdll_for_pidfd_open = ctypes.CDLL(None, use_errno=True)
         _cdll_for_pidfd_open.syscall.restype = ctypes.c_long
-        # fd and flags are actually int-sized, but the syscall() function
+        # pid and flags are actually int-sized, but the syscall() function
         # always takes longs. (Except on x32 where long is 32-bits and syscall
         # takes 64-bit arguments. But in the unlikely case that anyone is
         # using x32, this will still work, b/c we only need to pass in 32 bits
--- a/trio/_subprocess.py
+++ b/trio/_subprocess.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import sys
 from typing import Optional
 
 from ._abc import AsyncResource, SendStream, ReceiveStream
@@ -10,6 +11,39 @@ from ._subprocess_platform import (
     create_pipe_from_child_output
 )
 import trio
+
+# Linux-specific, but has complex lifetime management stuff so we hard-code it
+# here instead of hiding it behind the _subprocess_platform abstraction
+can_try_pidfd_open = True
+try:
+    from os import pidfd_open
+except ImportError:
+    if sys.platform == "linux":
+        import ctypes
+        _cdll_for_pidfd_open = ctypes.CDLL(None, use_errno=True)
+        _cdll_for_pidfd_open.syscall.restype = ctypes.c_long
+        # fd and flags are actually int-sized, but the syscall() function
+        # always takes longs. (Except on x32 where long is 32-bits and syscall
+        # takes 64-bit arguments. But in the unlikely case that anyone is
+        # using x32, this will still work, b/c we only need to pass in 32 bits
+        # of data, and the C ABI doesn't distinguish between passing 32-bit vs
+        # 64-bit integers; our 32-bit values will get loaded into 64-bit
+        # registers where syscall() will find them.)
+        _cdll_for_pidfd_open.syscall.argtypes = [
+            ctypes.c_long,  # syscall number
+            ctypes.c_long,  # fd
+            ctypes.c_long,  # flags
+        ]
+        __NR_pidfd_open = 434
+
+        def pidfd_open(fd, flags):
+            result = _cdll_for_pidfd_open.syscall(__NR_pidfd_open, fd, flags)
+            if result < 0:
+                err = ctypes.get_errno()
+                raise OSError(err, os.strerror(err))
+            return result
+    else:
+        can_try_pidfd_open = False
 
 
 class Process(AsyncResource):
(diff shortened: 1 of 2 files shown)


trio/_subprocess.py
```diff
@@ -11,6 +12,39 @@
...
+        # fd and flags are actually int-sized, but the syscall() function
+        # always takes longs. (Except on x32 where long is 32-bits and syscall
+        # takes 64-bit arguments. But in the unlikely case that anyone is
+        # using x32, this will still work, b/c we only need to pass in 32 bits
+        # of data, and the C ABI doesn't distinguish between passing 32-bit vs
+        # 64-bit integers; our 32-bit values will get loaded into 64-bit
+        # registers where syscall() will find them.)
+        _cdll_for_pidfd_open.syscall.argtypes = [
+            ctypes.c_long,  # syscall number
+            ctypes.c_long,  # fd
```
oremanj: I think this argument is called 'pid'?

trio/_subprocess.py
```diff
@@ -214,10 +263,14 @@ def returncode(self):
         if self.poll() is None:
             async with self._wait_lock:
                 if self.poll() is None:
-                    await wait_child_exiting(self)
-                    self._proc.wait()
+                    if self._pidfd is not None:
+                        await trio.hazmat.wait_readable(self._pidfd)
+                    else:
+                        await wait_child_exiting(self)
+                    self.poll()
```
oremanj: This is changing from `self._proc.wait()` to `self.poll()`. Any particular reason? I seem to recall that using `wait()` here was necessary for correctness on OS X, because kqueue and wait() aren't quite synchronized with each other, such that the waitpid(WNOHANG) used by poll() might return zero even though kqueue said "exiting" already.

njsmith: Oof, good catch, I'd forgotten about that `waidpid(WNOHANG)` synchronization thing.

The idea was to centralize the logic for cleaning up the pidfd inside our `poll` method. But you're right, this needs a different approach.

trio/_subprocess.py
```diff
@@ -31,7 +31,7 @@
         # registers where syscall() will find them.)
         _cdll_for_pidfd_open.syscall.argtypes = [
             ctypes.c_long,  # syscall number
-            ctypes.c_long,  # fd
+            ctypes.c_long,  # pid
```
pquentin: nit: The comment above still refers to `fd`

njsmith:
OK, I finally got back to this -- rebased on master, addressed the last review comment, I think it's ready to go.

njsmith:
Build failures are #1478, so this is blocked on #1479 or similar.

pquentin:
Now that #1479 is merged, is closing/reopening enough, or do we need to merge from master?

njsmith:
I think closing/reopening should be enough, since the CI always does a merge with master before running. But let's try it and find out :-)

pquentin:
GitHub won't let me merge because "Required statuses must pass before merging", but as far as I can tell everything passed. 🤔

njsmith:
Yeah, I'm seeing that too. What the heck, github. I also manually compared the list of statuses here to the ones listed in the branch protection settings, and everything matches.

PR #1299, opened 2019-11-09
[ci] Try switching to wget

njsmith:
In the hopes that it copes better with flaky servers/networks than
curl does.

The hope is that this will finally resolve gh-1231.


njsmith:
Apparently Azure's windows boxes don't have wget installed :-(

PR #1310, opened 2019-11-17
Attempt to clarify "same time" wording in tutorial

decentral1se:
> Hey, on the tutorial: it says: "it’s baked into Trio’s design that you never have two tasks running at the same time. " and then right after: "Trio can run 10,000+ tasks simultaneously without breaking a sweat". That's a bit confusing.

> that could use some better wording, yeah
> the same event loop will never run two tasks simultaneously
> but it can run a vast number of tasks concurrently

> yeah, they mean "same time" slightly differently, but that phrasing is confusing for sure... any chance you're up for making a PR to make it clearer?

I kinda don't still get it but I'm opening the PR to get the balling rolling.

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -770,11 +770,12 @@ programming, and – to add insult to injury – `pretty poor scalability
 Python just aren't that appealing.
 
 Trio doesn't make your code run on multiple cores; in fact, as we saw
-above, it's baked into Trio's design that you never have two tasks
-running at the same time. We're not so much overcoming the GIL as
-embracing it. But if you're willing to accept that, plus a bit of
-extra work to put these new ``async`` and ``await`` keywords in the
-right places, then in exchange you get:
+above, it's baked into Trio's design that when it has multiple tasks,
+they take turns, so at each moment only one of them is actively running.
+We're not so much overcoming the GIL as embracing it. But if you're
+willing to accept that, plus a bit of extra work to put these new
+``async`` and ``await`` keywords in the right places, then in exchange 
+you get:
 
 * Excellent scalability: Trio can run 10,000+ tasks simultaneously
   without breaking a sweat, so long as their total CPU demands don't


docs/source/tutorial.rst
```diff
@@ -771,12 +771,12 @@ Python just aren't that appealing.
 
 Trio doesn't make your code run on multiple cores; in fact, as we saw
 above, it's baked into Trio's design that you never have two tasks
-running at the same time. We're not so much overcoming the GIL as
+running simultaneously. We're not so much overcoming the GIL as
```
pquentin: I believe "at the same time" was a better choice of words. It means that at any given *point* of time, only one task is running. That being said, many tasks can still make progress within a given *period* of time. The words you can search for to understand the distinction are "concurrency vs. parallelism". Hope that helps!

decentral1se: Uhhhh, but "at the same time" does not work well with "simultaneously" below (see comments above), which was the whole point of the PR, to avoid this ambiguous wording.

pquentin: Sure, but now that the we use "concurrently" below, maybe that's fine?

Ultimately it's a subtle distinction that's hard to communicate to an audience that's not already aware of it :/ Maybe we should use "in parallel/concurrently"? I don't know.

In any case, this choice of words *is* better, but I won't merge this myself as I'm not a native English speaker.

njsmith:
I'm not sure using "concurrent" really clarifies things for most people either... maybe we should say something a little bit more elaborate, like

> it's baked into Trio's design that when it has multiple tasks, they take turns, so at each moment only one of them is actively running.

docs/source/tutorial.rst
decentral1se: See what you mean! Tough :)

decentral1se:
Tried to bake in both comments! See you what you think.

pquentin:
I think the version from @njsmith is better, you can remove "So, at any given point of time, only one task is running but many tasks can still make progress within a given period of time." :) Then it would be ready to merge!

(Oh, and do you know why your changes contain trailing whitespaces?)

decentral1se:
done done done

PR #1313, opened 2019-11-20
Take a stab at starting the tutorial cookbook

decentral1se:
Related to https://github.com/python-trio/trio/issues/472.

Specifically https://github.com/python-trio/trio/issues/472#issuecomment-555836888.

I just wrote an example of the itch I had to scratch and it should probably by no means be the first thing that is seen on that list but I thought it would be good to get this started and see what is what.


decentral1se:
(If this isn't immediately useful right now, please feel free to close. I made a vague stab at some wording and an example from my perspective, having a copy of that is maybe enough for when you do settle down to write this documentation?)

PR #1317, opened 2019-11-26
Add a "task_raised" instrumentation.

smurfix:
The use case: One of your tasks raises an exception. The nursery cancels
all other tasks and waits for them to end. If one of these tasks
swallows the cancellation, e.g. because of a deadlock, you need a way
to find the original exception.


oremanj:
I think it's unnecessary to have a separate `task_raised` instrument to capture tasks exiting with exceptions, when you can just define a `task_exited` instrument that checks whether its outcome argument is an Error.

It looks like you're also calling `task_raised` when an exception is raised out of the nested child of a nursery. That's a functionality that's not currently exposed to the Instrument API -- maybe this PR could provide just that, under a more specific name like `nested_child_exited`?

smurfix:
Fair enough, thanks for the feedback. Fixed.

smurfix:
Umm, Azure seems not to have picked up the latest change ?

smurfix:
> you can just define a task_exited instrument that checks whether its outcome argument is an Error.

… except that we don't have that yet.

@njsmith OK to simply add an `outcome` or `exception` argument to the `task_exited` instrumentation? Otherwise we'd need deprecation and a new method and all that.

smurfix:
… or I could simply set `task.outcome=outcome` before calling `task_exited` …

oremanj:
Ah, for some reason I thought we already passed the final outcome to the instrument... my bad!

I'm going to defer to @njsmith on the backwards compatibility policy for instruments.

Issue #1329, opened 2019-12-03
Accessing unavailable attributes gives a confusing error


pquentin:
When I access a variable that does not exist in Trio, I get this error message:

```
>>> import trio
>>> trio.stuff
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/quentin/.virtualenvs/connectors-watcher/lib/python3.6/site-packages/trio/_deprecate.py", line 125, in __getattr__
    raise AttributeError(name)
AttributeError: stuff
```

And Sentry displays it like this:

![Capture d’écran 2019-12-03 à 08 59 17](https://user-images.githubusercontent.com/42327/70035673-3f12a980-15ab-11ea-8190-cb7e9073ec5d.png)

Why is it confusing, you might say? The error is raised from `_deprecate.py` and I thought it was only a deprecation warning, but in fact it did not work at all.

I would have expected the standard behavior:

```
>>> import asyncio
>>> asyncio.stuff
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'asyncio' has no attribute 'stuff'
>>>
```

njsmith:
Hmm, that's tricky.

The reason it's being raised from `_deprecate.py` is because `_deprecate.py` checks access to undefined attributes on `trio`, because for deprecated attributes it needs to issue a deprecation warning and then return the real value. But once it's intercepted the attribute lookup to handle deprecated attributes, it *has* to handle it somehow, so if the name *isn't* a deprecated attribute then it has to raise `AttributeError` itself.

I don't know how to make the traceback match what you were expecting :-/ I guess we could do some hack, like drop the first frame from the traceback?

pquentin:
Yes, dropping the first frame would work for me. We already go out of our way to produce beautiful tracebacks in other cases (https://vorpus.org/blog/beautiful-tracebacks-in-trio-v070/), would you consider it to be acceptable in this case? I'm also happy to close the issue as I now understand it.

pquentin:
Closed in #1330 thanks to @dhirschfeld who is now part of the team. :)

PR #1330, opened 2019-12-03
Improve module-level AttributeError message

dhirschfeld:
This PR improves the error message by specifying the module name and also renames `_deprecate.py --> _check_attrs.py` as it might be somewhat misleading for an actual `AttributeError` to be raised from a module named `_deprecate.py`. This latter change might be controversial?

```python
In [1]: import trio

In [2]: trio.stuff

  File "<ipython-input-2-558a186a1b55>", line 1, in <module>
    trio.stuff

  File "c:\users\dhirschf\code\github\python-trio\trio\trio\_check_attrs.py", line 125, in __getattr__
    raise AttributeError(msg)

AttributeError: module 'trio' has no attribute 'stuff'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ __pycache__/
 .installed.cfg
 *.egg
 /.pybuild
+pip-wheel-metadata/
 
 # Installer logs
 pip-log.txt
@@ -64,3 +65,6 @@ coverage.xml
 
 # Sphinx documentation
 doc/_build/
+
+# PyCharm
+.idea/
--- a/trio/_deprecate.py
+++ b/trio/_deprecate.py
@@ -122,7 +122,8 @@ class _ModuleWithDeprecations(ModuleType):
             )
             return info.value
 
-        raise AttributeError(name)
+        msg = "module '{}' has no attribute '{}'"
+        raise AttributeError(msg.format(self.__name__, name))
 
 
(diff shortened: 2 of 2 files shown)


dhirschfeld:
Not sure if this is a useful change or not - I just felt like scratching an itch so feel free to take it or leave it...

If there is any interest in merging I'm happy to fix the usage of format strings (or make any other desired changes)

dhirschfeld:
Fixes #1329

pquentin:
This looks good to me and would solve my issue, thanks! Let's see what others think.

njsmith:
I like the new message, that definitely makes things clearer. Renaming `_deprecate.py` feels weird though... it has a bunch of deprecation-related features, and this is just one of them. And it doesn't really have attribute checking features at all.

I feel like the new message alone is likely enough to make things clear? If we really want to emphasize it, then we could even say "(accessed at {file}:{line})". (Or I mentioned removing the last traceback frame before, which would also work, except on further thought not I'm not sure it's possible in this case – it's hard to do with an exception until after you catch it.)

pquentin:
I would also be happy with only the message change :+1:

dhirschfeld:
> *I would also be happy with only the message change*

Done!

I kept the gitignore changes though under the assumption that if they're useful to me they must be useful to someone else also....

PR #784, opened 2018-11-26
ci: re-enable 3.8-dev coverage reports

pquentin:
The coverage.py bugs appear to have been fixed, let's see if that fixes our problem!

Issue #75, opened 2017-03-10
Dropping python 3.5 support?


njsmith:
Advantages of going to 3.6+:

* f strings
* the `__aiter__` mess goes away
* native async generators → slightly nicer tracebacks
* idempotent call_soon becomes guaranteed FIFO (not sure anyone cares but hey)

Disadvantages:
* 3.6 final won't be in any ubuntu release until 17.04
* there's no pypy 3.6 (but pypy 3.5 does have f-string support at least)
* some complications on RTD

My feeling is that it's not quite time to drop 3.5 support, but it wouldn't be too bad to start allowing f-strings in the medium term. And beyond that pypy becomes the gating factor.

dimaqq:
Odd opinion -- I live in 3.6 only world since b2.
wrt. distro -- that's where pyenv comes in :) or docker
pypy is a bummer, but then again I don't see many projects start with pypy.

What are the options?
* maintain 2 branches
* 1 branch, "dumb" code only
* recommend 3.6 (examples, etc.) just be careful to allow 3.5

njsmith:
We're definitely not going to maintain 2 branches :-). It's more about when the benefits outweigh the costs. The benefits are currently pretty minor, and as a practical-oriented project we want to avoid creating annoyances for users when possible (and "can't run on ubuntu without weird setup" is definitely an annoyance in these days when so many people run ubuntu servers).

Victor-Savu:
Personally, I don't even use the distro-provided python: `pyenv` ftw!

njsmith:
Well, here's a legitimate argument in favor of dropping support for CPython 3.5: the MacOS releases use an ancient openssl that's making it difficult to properly test our ssl support (https://github.com/pyca/pyopenssl/issues/624)

joernheissler:
> we want to avoid creating annoyances for users when possible (and "can't run on ubuntu without weird setup" is definitely an annoyance in these days when so many people run ubuntu servers).

Current Ubuntu LTS ships cpython 3.6.
But I use Debian on several of my systems and it won't be before end of 2019 that I can move them to Debian Buster which will have cpython 3.7 (or 3.8?)

Of course I could use pyenv, but I prefer to use the system's python installation.

Next CentOS/RHEL (probably released beginning of 2019) will have cpython 3.6.6 from what I can see.

In my opinion, a good date to drop python 3.5 support would be some time after next debian release, i.e. mid/end 2019.

njsmith:
[...]

PR #1339, opened 2019-12-16
Enable Python 3.8 coverage

pquentin:
We now test Python 3.8 on Windows, so I rebased the commits of https://github.com/python-trio/trio/pull/784 on top of master (without using coverage master). I don't know how to configure Azure Pipelines on my fork, so here's a PR. It's only a draft because I first want to see if we see the same things.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -267,14 +267,10 @@ else
         netsh winsock reset
     fi
 
-    # Disable coverage on 3.8 until we run 3.8 on Windows CI too
-    #   https://github.com/python-trio/trio/pull/784#issuecomment-446438407
-    if [[ "$(python -V)" = Python\ 3.8* ]]; then
-        true;
     # coverage is broken in pypy3 7.1.1, but is fixed in nightly and should be
     # fixed in the next release after 7.1.1.
     # See: https://bitbucket.org/pypy/pypy/issues/2943/
-    elif [[ "$TRAVIS_PYTHON_VERSION" = "pypy3" ]]; then
+    if [[ "$TRAVIS_PYTHON_VERSION" = "pypy3" ]]; then
         true;
     else
         # Flag pypy and cpython coverage differently, until it settles down...
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -72,9 +72,12 @@ else:
 if sys.version_info < (3, 5, 2):
 
     def aiter_compat(aiter_impl):
-        @wraps(aiter_impl)
+        # de-sugar decorator to fix Python 3.8 coverage issue
+        # https://github.com/python-trio/trio/pull/784#issuecomment-446438407
         async def __aiter__(*args, **kwargs):
             return aiter_impl(*args, **kwargs)
+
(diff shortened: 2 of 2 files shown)


pquentin:
> We could wait to re-enable coverage until 3.8 is actually released. That would at least handle the `fromshare` issue...
>
> _Originally posted by @njsmith in https://github.com/python-trio/trio/pull/784#issuecomment-446438407_

Indeed, we get the fromshare fix for free. However, that does not fix the aiter_compat issue that we use for 3.5.0 and 3.5.1. I don't really know how to fix it. Here is the full code:

https://github.com/python-trio/trio/blob/d47e018d84862eecd9a896bf127159436a473494/trio/_util.py#L72-L83

I don't even understand how that works. The idea is to fix https://bugs.python.org/issue27243, but with something more generic than https://github.com/aio-libs/aiohttp/commit/7276cc913741d79822e04adffc20905659c90a96. I cant't use a if inside the def since we want to define the functions in both case. And I tried to desugar but it didn't work.

I'm sure @njsmith can come up with a clever fix but I'd rather drop support for Python 3.5 as I think we reached consensus in #75.

ci.sh
```diff
@@ -267,14 +267,10 @@ else
         netsh winsock reset
     fi
 
-    # Disable coverage on 3.8 until we run 3.8 on Windows CI too
-    #   https://github.com/python-trio/trio/pull/784#issuecomment-446438407
-    if [[ "$(python -V)" = Python\ 3.8* ]]; then
-        true;
     # coverage is broken in pypy3 7.1.1, but is fixed in nightly and should be
     # fixed in the next release after 7.1.1.
     # See: https://bitbucket.org/pypy/pypy/issues/2943/
```
njsmith: I guess 7.2.0 is out, so possibly we can remove this too...

trio/_util.py
```diff
@@ -72,9 +72,11 @@ def signal_raise(signum):
 if sys.version_info < (3, 5, 2):
 
     def aiter_compat(aiter_impl):
-        @wraps(aiter_impl)
+        # de-sugar decorator to fix Python 3.8 coverage issue
+        # https://github.com/python-trio/trio/pull/784#issuecomment-446438407
         async def __aiter__(*args, **kwargs):
             return aiter_impl(*args, **kwargs)
+        __aiter__ = wraps(aiter_impl)
```
njsmith:
```suggestion
        __aiter__ = wraps(aiter_impl)(__aiter__)
```

ci.sh
pquentin: I want to try that next. 👍

trio/_util.py
pquentin: Thanks! I just did that. I'll worry about dropping 3.5 in another pull request.

pquentin:
>  python-trio.trio Expected — Waiting for status to be reported 

Not sure why it's not reported. Closing/reopening.

pquentin:
Closing/reopening to see if I get `python-trio.trio` check this time

pquentin:
@njsmith Could you please take another look? Thanks!

PR #1351, opened 2019-12-28
Deadline setting performance

Tronic:
A continuation of #1234.


Tronic:
Improved performance of CancelScope deadline management.

- Refactor deadlines into a separate class easily modified to support different algorithms
- Use `heapq` to track deadlines instead of `SortedDict`
- Removing entries in the middle is a slow operation, and thus must be performed only occasionally
   * Triggered by (arbitrary) condition: `len(heap) > active_count * 2 + 1000`

Tronic:
HTTP server benchmark
```python
import trio

async def http_server(sock):
    try:
        with trio.CancelScope() as cs:
            while True:
                cs.deadline = trio.current_time() + 15.0  # Idle
                buf = await sock.receive_some()
                cs.deadline = trio.current_time() + 30.0  # Request
                while b"\r\n\r\n" not in buf:
                    data = await sock.receive_some()
                    if not data: return
                    buf += data
                cs.deadline = trio.current_time() + 60.0  # Response
                await sock.send_all(
                    b"HTTP/1.1 200 OK\r\ncontent-type:text/plain\r\ncontent-length:6\r\n\r\nhello\n"
                )
    except trio.BrokenResourceError:
        pass

trio.run(trio.serve_tcp, http_server, 5000)
```

Benchmarking with `wrk http://127.0.0.1:5000/`:

Benchmark | req/s
--- | ---
master | 11000
faster-deadlines2 | 13800
no deadlines | 20800

The no deadlines test has the deadline setting rows of the above code commented out. In that, results do not differ between this PR and master.

dkwolfe4:
> * Removing entries in the middle is a slow operation, and thus must be performed only occasionally

Perhaps popping the end entry off and replacing the one to be removed?

Tronic:
> > * Removing entries in the middle is a slow operation, and thus must be performed only occasionally
> 
> Perhaps popping the end entry off and replacing the one to be removed?

Would you mind implementing that and benchmarking to see how it performs?

njsmith:
Closing in favor of #1629

PR #1365, opened 2020-01-13
Add note about cancellation and memory channels

decentral1se:
Ripped from the gitter channel. Might help.

### DIFF
--- a/trio/_channel.py
+++ b/trio/_channel.py
@@ -27,6 +27,11 @@ def open_memory_channel(max_buffer_size):
     on a channel when there's no-one on the other side. See
     :ref:`channel-shutdown` for details.
 
+    Memory channel operations are all atomic with respect to
+    cancellation, either `~trio.abc.ReceiveChannel.receive` will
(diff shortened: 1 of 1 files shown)


decentral1se:
Ah, don't think that coverage issue relates to my change then.

tjstum:
This looks good to me, and I agree with your assessment about coverage.

Do we want to mention this in the reference documentation for `SendChannel` and `ReceiveChannel` (i.e. mention it in `send` and `receive`)?
We also might want a similar mention about `send` being cancelled at _just_ the right moment causing the channel to be empty but no one's `receive` unblocking (unless I am misunderstanding the issue here)

njsmith:
Thanks, and sorry for the delayed review!

I don't think we want to make this kind of atomicity part of the generic `SendChannel`/`ReceiveChannel` contract, because those are also useful for things like TCP-stream-plus-framing, where atomicity would be extremely difficult.

decentral1se:
Thanks!

PR #1382, opened 2020-01-31
Add new Python 3.8 socket constants

pquentin:
As seen on https://dev.azure.com/python-trio/trio/_build/results?buildId=1458&view=logs&jobId=e9afaa34-1a0f-534e-78fc-fae528ccd915&j=e9afaa34-1a0f-534e-78fc-fae528ccd915&t=20613640-1190-5387-fcd9-276ced9a79eb


njsmith:
Huh, I wonder how those snuck in without our noticing. They must have been added in the 3.8.1 point release or something? Anyway, thanks for taking care of this :-)

PR #1387, opened 2020-02-04
Test PyPy 7.2

pquentin:
Upgrading to PyPy 7.2 allows removing the coverage workaround.

I also tried using PyPy 3.7 nightly, but there are a few TLS/SSL errors that I don't understand, so I'll try that in another pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 os: linux
 language: python
-dist: xenial
+dist: bionic
 
 jobs:
   include:
     # The pypy tests are slow, so we list them first
-    - python: pypy3
+    - python: pypy3.6-7.2.0
     - language: generic
       env: PYPY_NIGHTLY_BRANCH=py3.6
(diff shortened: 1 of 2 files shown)


pquentin:
The coverage is low because Windows coverage is broken since https://github.com/python-trio/trio/pull/1374. I think PyPy is not making things worse here? But I'm not sure.

njsmith:
The pypy issue that's linked to there gives the example of `test_ssl.py` as having random blank lines marked as uncovered. I did a spot check of that file with this PR (https://codecov.io/gh/python-trio/trio/src/fb2cd8d45c389a5a368af73f10922b5b025ed09a/trio/tests/test_ssl.py), and I don't see any mysterious red lines, so that also suggests that this is good :-) If there are any other problems with pypy coverage, I guess we will notice after windows coverage is fixed and then we can file another pypy bug...

pquentin:
Thank you for the review!

PR #1390, opened 2020-02-05
Drop Python 3.5

### LINKED ISSUE #75: Dropping python 3.5 support?

pquentin:
Closes #75 

I'll need to write newsfragments for 1/ the actual removal 2/ the call_soon FIFO ordering guarantee.

@oremanj Can `test_abc_generics` go away with Python 3.5 or is it still useful?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ jobs:
       env:
         - "JOB_NAME='Ubuntu 19.10, full VM'"
         - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
+    - python: 3.6.1  # earliest 3.6 version available on Travis
     - python: 3.6-dev
     - python: 3.7-dev
     - python: 3.8-dev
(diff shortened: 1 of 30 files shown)


pquentin:
Coverage is pointing more places where I need to change things. Unfortunately I can't put my PR in draft mode now. I'll request reviews when this is ready.

pquentin:
Closing for now as I'm not sure I'll be able to continue this, but keeping the branch alive in case someone wants to continue the work started here.

wgwz:
@pquentin do you have a list of things that need to be done for completion?

pquentin:
@wgwz Thanks for stepping up! There's nothing specific left, we just need to figure out why the tests don't pass (probably due to one of the last commits), and then make sure the coverage does not drop (which is a bit tricky since it's flaky and Windows coverage is currently broken).

pquentin:
I realized I forgot to close, doing this now so that the work can continue in another pull request.

njsmith:
It wouldn't be a terrible idea to stick in a deprecation warning for a
release, anyway – just check if we're on 3.5 at import, and issue a warning
if so. Do you have time for a smaller PR like that?

pquentin:
Yes, that would be a good first step.

PR #1391, opened 2020-02-05
Add back cffi on install_requires

pquentin:
The wheel bug was fixed in 2017 by setuptools. Indeed, the METADATA is correct now.

https://github.com/pypa/wheel/issues/181

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,9 +85,7 @@ setup(
         "idna",
         "outcome",
         "sniffio",
-        # PEP 508 style, but:
-        # https://bitbucket.org/pypa/wheel/issues/181/bdist_wheel-silently-discards-pep-508
-        #"cffi; os_name == 'nt'",  # "cffi is required on windows"
+        "cffi; os_name == 'nt'",  # "cffi is required on windows"
     ],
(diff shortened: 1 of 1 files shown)


njsmith:
This isn't quite right, one way or another – we have a workaround too, a bit further down the file, where we use this weird old extras syntax instead that generates the same wheel metadata. So we definitely don't need to do it twice :-) it's true that the workaround isn't doing much for us these days though...

pquentin:
Thanks! I added a commit to fix that.

I checked that the wheel METADATA file continues to contain this:

```
Requires-Dist: cffi (>=1.12) ; os_name == "nt"
Requires-Dist: contextvars (>=2.1) ; python_version < "3.7"
```

And that trio.egg-info/requires.txt in the sdist still contains this:

```
[:os_name == "nt"]
cffi>=1.12

[:python_version < "3.7"]
contextvars>=2.1
```

So as long as the person releasing trio is not using anything too ancient, the result will be exactly the same as before.

pquentin:
@njsmith Can you please take another look?

pquentin:
Thank you!

PR #1393, opened 2020-02-06
ci: use bash <(curl ...) again

pquentin:
[The Azure Pipelines bug is now fixed](https://developercommunity.visualstudio.com/content/problem/743824/bash-task-on-windows-suddenly-fails-with-bash-devf.html).


pquentin:
Thank you @sethmlarson!

PR #1396, opened 2020-02-10
Drop python3.5

### LINKED ISSUE #75: Dropping python 3.5 support?

wgwz:
Continues @pquentin's work from #1390
Closes #75

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ jobs:
       env:
         - "JOB_NAME='Ubuntu 19.10, full VM'"
         - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
+    - python: 3.6.1  # earliest 3.6 version available on Travis
     - python: 3.6-dev
     - python: 3.7-dev
     - python: 3.8-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,13 +21,6 @@ jobs:
       env:
         - "JOB_NAME='Ubuntu 19.10, full VM'"
         - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
-    # 3.5.0 and 3.5.1 have different __aiter__ semantics than all
-    # other versions, so we need to test them specially. Travis's
-    # newer images only provide 3.5.2+, so we have to request the old
-    # 'trusty' images.
-    - python: 3.5.0
-      dist: trusty
-    - python: 3.5-dev
     - python: 3.6-dev
     - python: 3.7-dev
     - python: 3.8-dev
--- a/README.rst
+++ b/README.rst
@@ -102,7 +102,7 @@ demonstration of implementing the "Happy Eyeballs" algorithm in an
 older library versus Trio.
 
 **Cool, but will it work on my system?** Probably! As long as you have
-some kind of Python 3.5-or-better (CPython or the latest PyPy3 are
+some kind of Python 3.6-or-better (CPython or the latest PyPy3 are
 both fine), and are using Linux, macOS, or Windows, then Trio should
 absolutely work. *BSD and illumos likely work too, but we don't have
 testing infrastructure for them. And all of our dependencies are pure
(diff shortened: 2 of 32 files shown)


wgwz:
I was able to reproduce the issues locally with both 3.7-dev and 3.8-dev. It looks like the issues with the tests are all related to this error:

```
RuntimeError: cannot reuse already awaited aclose()/athrow()
```

I'm not sure how to proceed from here. I found this:

- https://bugs.python.org/issue38559

Any pointers?

pquentin:
@wgwz Thanks for working on this! I think we should just revert that commit for now, and open an issue to worry about it later.

@njsmith Assuming we'll soon have a working pull request here, we should first release Trio with just a warning as mentioned in #1390, right?

njsmith:
> we should first release Trio with just a warning as mentioned in #1390, right?

I think that would be ideal. It could be like, merge the warning, release, and then immediately merge this PR within a 10 minute span, though.

njsmith:
The pypy failure looks real.

njsmith:
The correct bpo link for the new aclose/athrow thing is: https://bugs.python.org/issue39386

njsmith:
OK and I tracked down the `cannot reuse already awaited aclose()/athrow()` problem... it's a genuine regression in the CPython dev branches, that was introduced in a PR that I approved (whoops). Upstream bug filed here: https://bugs.python.org/issue39606

njsmith:
Phew, it's lucky we dug into that `cannot reuse already awaited aclose()/athrow()` thing... the problematic change was scheduled to be released in 3.8.2 on Monday. So it's pure luck that this PR came along this week and caught the regression before it could make it into a released Python... anyway I submitted a fix upstream at https://github.com/python/cpython/pull/18475, in case anyone is curious.

wgwz:
Is there any place that would be good to see how to issue the type of deprecation warning you are looking for here? I've never done that before.

@njsmith Glad this brought it to your attention! Going to take a look at your fix.

njsmith:
That fix is merged upstream now, so sometime in the next day or so the Travis builds should update to include it, and the tests here should start passing.

> Is there any place that would be good to see how to issue the type of deprecation warning you are looking for here? I've never done that before.

I don't have an example, no, but I guess it should just require adding something like this to the bottom of `trio/__init__.py`?

```python
import sys
if sys.version_info < (3, 6):
    _deprecate.warn_deprecated("Support for Python 3.5", "0.14", issue=75, instead="Python 3.6+")
del sys
```

It might also need some fiddling with our pytest warnings filters to keep the tests passing.

wgwz:
It looks like the pypy nightly branch is failing because of these 3.6 additions not being present: https://docs.python.org/3/whatsnew/3.6.html#socket

pquentin:
@wgwz Right, 4f3030f should go away if it does not exist on PyPI. Can you please revert it? Sorry about that.

Also, thanks for the deprecated warning commit! However it should go in another pull request, so that we can merge it and release it before merging this one.

njsmith:
I filed a bug with PyPy about the missing socket constants: https://foss.heptapod.net/pypy/pypy/issues/3171

njsmith:
We should add a comment next to the `SO_PROTOCOL`/`SO_DOMAIN` fallback definitions that links to that PyPy issue, so that later on we can figure out why the code is there :-)

wgwz:
I'm a bit puzzled by the [latest][1] CI failure..

[1]: https://dev.azure.com/python-trio/trio/_build/results?buildId=1548&view=logs&j=dcd50a69-7bd8-59f7-72b1-a119e6ffb32f&t=3369a326-68a0-5202-7c21-6671aa8f30ac&l=735

njsmith:
> I'm a bit puzzled by the latest CI failure..

It's a mysterious issue that happens occasionally for reasons unknown: https://github.com/python-trio/trio/issues/200#issuecomment-584485278

It doesn't seem to be related to this PR though, so we can just re-run the CI here.

wgwz:
Hey @njsmith, is there anything I can do to help here? I've somewhat lost track of where we left off :-)

pquentin:
The idea is to fix the conflict and remove code that's no longer covered. I can take over from here if you want.

wgwz:
Sure, feel free @pquentin!

pquentin:
Merged as part of https://github.com/python-trio/trio/pull/1481

PR #1399, opened 2020-02-11
Update function open_unix_socket to check the type of filename argument

lewoudar:
Hi,
this PR fixes an edge case where a user can pass a number when calling `open_unix_socket`. In this case he will get a `TypeError`. There are two possibilities:
- We can just let `socket` function raises the error or
- We can intercept the error and print a nicer message, which is what I did

And in all cases, we need to update the docstring :)

### DIFF
--- a/trio/_highlevel_open_unix_stream.py
+++ b/trio/_highlevel_open_unix_stream.py
@@ -36,12 +36,13 @@ async def open_unix_socket(filename,):
     Raises:
       OSError: If the socket file could not be connected to.
       RuntimeError: If AF_UNIX sockets are not supported.
+      TypeError: if filename is not str or bytes.
     """
     if not has_unix:
         raise RuntimeError("Unix sockets are not supported on this platform")
 
-    if filename is None:
-        raise ValueError("Filename cannot be None")
+    if not isinstance(filename, (str, bytes)):
+        raise TypeError("Filename must be str or bytes")
 
     # much more simplified logic vs tcp sockets - one socket type and only one
     # possible location to connect to
--- a/trio/_highlevel_open_unix_stream.py
+++ b/trio/_highlevel_open_unix_stream.py
@@ -41,8 +41,8 @@ async def open_unix_socket(filename,):
     if not has_unix:
         raise RuntimeError("Unix sockets are not supported on this platform")
 
-    if not isinstance(filename, (str, bytes)):
-        raise TypeError("Filename must be str or bytes")
+    if not isinstance(filename, (str, trio.Path, bytes)):
+        raise TypeError("Filename must be str, trio.Path or bytes")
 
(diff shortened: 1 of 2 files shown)


trio/_highlevel_open_unix_stream.py
```diff
@@ -36,12 +36,13 @@ def close_on_error(obj):
     Raises:
       OSError: If the socket file could not be connected to.
       RuntimeError: If AF_UNIX sockets are not supported.
+      TypeError: if filename is not str or bytes.
```
njsmith: I don't think we should mention this in the documentation. It's a general rule that pretty much any Python function can raise `TypeError` if you give it the wrong argument types, and mentioning it explicitly here distracts from more relevant information.

trio/_highlevel_open_unix_stream.py
```diff
@@ -36,12 +36,13 @@ def close_on_error(obj):
...
     """
     if not has_unix:
         raise RuntimeError("Unix sockets are not supported on this platform")
 
-    if filename is None:
-        raise ValueError("Filename cannot be None")
+    if not isinstance(filename, (str, trio.Path, bytes)):
+        raise TypeError("Filename must be str, trio.Path or bytes")
```
njsmith: This isn't *quite* correct... really the criterion is "is the object a valid path". And starting in Python 3.6, the proper way to check that is to use the [fspath protocol](https://www.python.org/dev/peps/pep-0519/). So if we want to do this check, then the right way to do it would be something like:

```python
filename = os.fspath(filename)
```

This automatically handles all "path-like" objects, including new ones defined in third-party libraries that we've never heard of. And if the user passes in something inappropriate, it gives a good error message:

```python
>>> os.fspath(4)                                                                      
TypeError: expected str, bytes or os.PathLike object, not int
```

Annoyingly, though, we still support Python 3.5 (though not for long, see #1396), so we can't just use `os.fspath`. Instead, use our compatibility wrapper: `trio._util.fspath`

trio/_highlevel_open_unix_stream.py
lewoudar: Ok for `TypeError` I removed it

trio/_highlevel_open_unix_stream.py
lewoudar: Thank you for the information, I didn't know the function `os.fspath`. To be honest, when I made the change, I told myself that a `pathlib.Path` object would also be valid but I was "comforted" by the fact that the tests only took into account `trio.Path` 😛
I changed the check

pquentin:
Cycling the pull request to fix the Travis checks.

pquentin:
Thanks!

lewoudar:
You are welcome!

PR #1400, opened 2020-02-11
Update error message in IO function open_tcp_listeners

lewoudar:
Hi,
I just change the message printed when a user passes a string to `open_tcp_listeners`, previously it was said that it can be a *string* which is not true


njsmith:
Ah right, good catch, thanks! (I think what happened is that at some point during development I wanted to support some of the more exotic ways of specifying ports, like service names etc., and then decided it wasn't useful enough to justify the complications and edge cases it creates, so we restricted it to just ints. But the error message didn't get updated.)

njsmith:
Ugh, tests are failing though for unrelated reasons. Apparently travis has broken all their python dev builds. (Somehow pip is missing from all of them.)

pquentin:
Opened https://github.com/python-trio/trio/issues/1402 to track the python dev failures

pquentin:
Cycling to PR to see if the checks pass.

pquentin:
Thanks @lewoudar!

PR #1433, opened 2020-03-11
RFC: UNIX server sockets

Tronic:
Implement high level API for UNIX server sockets, based on https://github.com/python-trio/trio/issues/279

This needs some more work but I'd like to invite others for comments at this point.


decentral1se:
Just tried to figure out how to serve a unix socket and ended up here, thanks for this effort ... I guess my only capacity to help this along is basically to bump the PR with this comment :sweat_smile: There is a whole bunch of stuff here I had no idea that should be taken care of when dealing with sockets...

decentral1se:
Oh yeah, could also take it for a test drive! Saw the following warning when running `serve_unix`:

> TrioDeprecationWarning: subclassing trio.SocketListener is deprecated since Trio 0.15.0; use composition or delegation instead (https://github.com/python-trio/trio/issues/1044) class UnixSocketListener(trio.SocketListener):

py-radicz:
I am also bumping this PR as I think it would be nice to have also UDS server support - thank you @Tronic + @decentral1se you saved me ton of time figuring this out

trio/_highlevel_open_unix_listeners.py
```diff
@@ -0,0 +1,298 @@
...
+    Args:
+
+      path (str): Filename of UNIX socket to create and listen on.
+
+          Absolute or relative paths may be used.
+
+          The socket is initially created with a random token appended to its
+          name, and then moved over the requested name while protected by a
+          separate lock file. The additional names use suffixes on the
+          requested name.
```
GalaxySnail, line 220: Why is it necessary here?

CoolCat467: Not sure, should we try removing this?

PR #1442, opened 2020-03-30
Turn codecov back on

njsmith:
Apparently this got accidentally disabled as part of a cleanup a while
ago. (The problem is that the bash <(...) syntax assumes that curl is
printing the file to stdout, but we're actually redirecting to a
temporary file in order to handle retries.)


pquentin:
Great catch, thanks!

PR #1448, opened 2020-04-07
[Fix][Docs] use trio.sleep instead of sleep

ianre657:
The trio library needs to use `trio.sleep()` instead of regular `sleep` function.


njsmith:
Nice catch, thanks!

PR #1458, opened 2020-04-19
Remove deprecated open_cancel_scope

pquentin:


pquentin:
The coverage only goes down because I remove 6 previously covered lines.

PR #1455, opened 2020-04-15
Remove redundent token variables.

matham:
`token` and `trio_token` are the same.

I guess we don't need to pass `trio_token` to the thread function as a arg parameter, but that seems more explicit so I removed `token` instead of removing `trio_token`.

PR #1467, opened 2020-04-27
Port Azure Pipelines configuration to Github Actions

njsmith:
Github Actions is probably a better choice anyway, because the free
tier gives us 2x the resources, plus it avoids all the janky stuff
with managing separate Pipelines user accounts. But the final straw is
that Azure Pipelines has been totally broken for unknown reasons for
the last week+, see: #1455, #1466, etc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [pull_request, push]
+on: [pull_request]
 
 jobs:
   Windows:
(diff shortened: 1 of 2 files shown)


njsmith:
So I don't think the actions will actually start running until this is merged to master, but you can see an example run here: https://github.com/njsmith/trio/runs/622090739?check_suite_focus=true

njsmith:
So *now* AP decides to work... maybe the secret was just to threaten it :-)

PR #1475, opened 2020-04-27
Removed reference to `portal` in from thread example

MicahLyle:
When I was reading the docs, I was unsure of why `portal` was there in the reference.

My assumption is that it's there to hint to the reader that calling `trio.to_thread.run_sync` creates a portal in a sense, but based in the function signature the comment was confusing to me.

Please correct me if I'm misunderstanding anything!


oremanj:
Good catch! This is a holdover from when getting back into Trio from a thread required constructing a `BlockingTrioPortal` object.

PR #1481, opened 2020-04-29
Drop Python 3.5 (third take)

### LINKED ISSUE #75: Dropping python 3.5 support?

pquentin:
Closes #1390, closes #75 

I went to great lengths to make this easy to review commit by commit, by reorganizing the previous pull request.

There's one major change compared to #1390: even if it's not needed in Python 3.6, I decided to keep support for `async_generator` in user code. I also added a test for native async generators.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ jobs:
       env:
         - "JOB_NAME='Ubuntu 19.10, full VM'"
         - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
+    - python: 3.6.1  # earliest 3.6 version available on Travis
     - python: 3.6-dev
     - python: 3.7-dev
     - python: 3.8-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,13 +21,6 @@ jobs:
       env:
         - "JOB_NAME='Ubuntu 19.10, full VM'"
         - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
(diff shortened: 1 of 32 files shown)


pquentin:
Unfortunately the coverage isn't really useful, but the important thing to notice is that we were [missing 24 lines before](https://codecov.io/gh/python-trio/trio/tree/a434a00f98d52430d768b3c9169e646ad2715fc1) and that [it's still the case now](https://codecov.io/gh/python-trio/trio/pull/1481/tree).

Also, @njsmith will need to stop requiring the Python 3.5 checks.

.travis.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       env:
         - "JOB_NAME='Ubuntu 19.10, full VM'"
         - "VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
+    - python: 3.6.1  # earliest 3.6 version available on Travis
```
njsmith, line 24: Is there a specific motivation for this?

njsmith:
One question and a few more small cleanups that I noticed while reviewing, but otherwise looks great.

Thanks to @wgwz for the initial work, and @pquentin for making this such a pleasure to review!

.travis.yml
pquentin: Not really, I think I just mirrored the fact that we were supporting early versions of Python 3.5. In fact it looks like `test_ki_wakes_us_up` could be simplified if we decide to only support 3.6.2+.

pquentin:
I can't approve my own pull request, but your additions look good to me, thanks!

PR #1482, opened 2020-04-29
Use pidfd to implement trio.Process.wait, when available (2nd try)

### LINKED ISSUE #1241: use pidfd on Linux when available

njsmith:
This is identical to #1296, but for some reason that PR has gotten
into a glitchy state where GH won't let it be merged. So I'm opening a
second PR in hopes that GH will somehow un-wedge themselves.

Fixes #1241

### DIFF
--- a/newsfragments/1241.feature.rst
+++ b/newsfragments/1241.feature.rst
@@ -1,0 +1,4 @@
+On Linux kernels v5.3 or newer, `trio.Process.wait` now uses `the
+pidfd API <https://lwn.net/Articles/794707/>`__ to track child
+processes. This shouldn't have any user-visible change, but it makes
+working with subprocesses faster and use less memory.
--- a/trio/_subprocess.py
+++ b/trio/_subprocess.py
@@ -22,7 +22,7 @@ except ImportError:
         import ctypes
         _cdll_for_pidfd_open = ctypes.CDLL(None, use_errno=True)
         _cdll_for_pidfd_open.syscall.restype = ctypes.c_long
-        # fd and flags are actually int-sized, but the syscall() function
+        # pid and flags are actually int-sized, but the syscall() function
         # always takes longs. (Except on x32 where long is 32-bits and syscall
         # takes 64-bit arguments. But in the unlikely case that anyone is
         # using x32, this will still work, b/c we only need to pass in 32 bits
--- a/trio/_subprocess.py
+++ b/trio/_subprocess.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import sys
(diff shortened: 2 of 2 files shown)


njsmith:
Failure was: https://github.com/python-trio/trio/issues/200#issuecomment-621442601

I hit re-run jobs.

njsmith:
Huh, now this is glitched out too. Maybe the problem is hitting re-run jobs?

njsmith:
Yeah, that seems to have cleared it... weird. Well, I am ready for this to be done :-) @pquentin approved it already in #1296 so I'll just self-merge.

Issue #476, opened 2018-03-19
Should we rename trio.hazmat?


njsmith:
On the one hand, the name `hazmat` is definitely effective at getting people to pay attention and be careful. OTOH, I'm not sure it's quite communicating what we want: people seem scared of it, and uncertain whether they can actually use it. A bit of that is fine, but... I'm not sure it's communicating what we want. The block of text at the beginning helps of course, but you can't really fix up a confusing name with good docs.

Maybe `trio.lowlevel`? Though `trio.socket` is kind of low-level too. Does `lowlevel` have the kind of "if you see this during a code review, ask for a comment justifying its usage" implication that we're going for?

`trio.foundation`? `trio.core`?

lisael:
Hi, and thanks for the great work with trio.

While reading the docs, I had the feeling that trio API is sometimes not non-english-native-speaker friendly. I read a lot of technical documentation and I usually barely struggle to understand technical English but to be honest, I have exactly no idea what `hazmat` means. Not scary at all for more than the 2/3 of potential users. I've also noticed `cushion` and `tiebreaker`, that don't sound very technical, even though I can infer the meaning of those, from my French background or by word composition.

OTOH, the naming in the API is somehow unusual, refreshing, and sometimes very explicit `do_stuff_before_foo`, but the refreshing part shouldn't hinder the readability. Don't make me wrong, I do think that the documentation is incredibly extensive, readable and sometimes amusing for such a young project.

May I suggest `trio.internals` ? `core` and `foundation` read good too. Maybe `internals` reads too much like `don't you dare using this, you mortal`

njsmith:
@lisael 
> I read a lot of technical documentation and I usually barely struggle to understand technical English but to be honest, I have exactly no idea what `hazmat` means. Not scary at all for more than the 2/3 of potential users.

That's a great point – thank you! In colloquial English, `hazmat` is "hazardous materials"; it suggests that when using this module you might [want to wear a suit like this](https://en.wikipedia.org/wiki/Hazmat_suit#/media/File:Hazmat_DEA.jpg). I definitely want things to be accessible, but as a native English speaker I'm at a disadvantage: it's hard for me to guess which things are common knowledge and which things are weird idioms that most people don't know :-). So, I really appreciate this kind of feedback! If you notice any other places where trio's naming is confusing, please speak up; it's a really valuable contribution. (And just to confirm: you're saying that `cushion` and `tiebreaker` surprised you, but were understandable, right?)

[...]

PR #1484, opened 2020-05-01
Rename trio.hazmat → trio.lowlevel

njsmith:
Closes: gh-476


newsfragments/476.removal.rst
```diff
@@ -0,0 +1,3 @@
+It turns out that everyone got confused by the name ``trio.lowlevel``.
```
pquentin, line 1: s/lowlevel/hazmat/s

docs/source/reference-lowlevel.rst
```diff
@@ -1,44 +1,40 @@
...
-Using :mod:`trio.hazmat` isn't really *that* hazardous; in fact you're
-already using it – it's how most of the functionality described in
-previous chapters is implemented. The APIs described here have
-strictly defined and carefully documented semantics, and are perfectly
-safe – *if* you read carefully and take proper precautions. Some of
-those strict semantics have `nasty big pointy teeth
+You don't need to be scared of :mod:`trio.lowlevel`, as long as you
+take proper precautions. These are real public APIs, with strictly
+defined and carefully documented semantics. But, you should read
+carefully and take proper precautions. Some of those strict semantics
```
georgepsarakis, line 30: Is the phrase repetition with line 28 intentional? Perhaps something like this could also work:

```
But, you should pay attention while using them and certainly read the documentation in detail.
```

newsfragments/476.removal.rst
njsmith: Haha, whoops! nice catch :-)

njsmith:
@pquentin 
> This no longer works, is it expected?
> ```
> >>> import trio.hazmat
> ```

Excellent question! I did not think about that.

I doubt many folks are writing `import trio.hazmat`, because `import trio` automatically gives you `trio.hazmat`. But I bet there are folks writing `from trio.hazmat import ...`, and it turns out that this change also broke that, for the same reason.

The issue is that when you refer to `trio.hazmat.something`, then it looks for the `hazmat` attribute on `trio`, and when the attribute isn't found, then it runs our `__getattr__` fallback code, which issues the warning and then returns the proper value. BUT, `import trio.hazmat` tries to find `trio/hazmat.py` directly, without checking for an existing `.hazmat` attribute on `trio`.

I did find a workaround though: if we make a `trio/hazmat.py` file that issues a warning and then does `from trio.lowlevel import *`, then everything mostly works:

- If you do `trio.hazmat.blah` then it works and gives a warning
- If you do `from trio.hazmat import blah` then it works and gives a warning

The one issue is that *after* you do `from trio.hazmat import ...`, then any *future* accesses to `trio.hazmat.whatever` will stop issuing warnings (because the import machinery adds a real `hazmat` attribute to the `trio` module, so our `__getattr__` fallback code never runs). But that's OK... it's technically not *quite* as nice because it defeats the part of the `warnings` module that tries to tell you about all the different places that are using the deprecated code, but we still preserve the key invariant that every program which uses `trio.hazmat` does print at least one warning.

docs/source/reference-lowlevel.rst
njsmith: Good point, tweaked the wording a bit.

njsmith:
Addressed review comments.

On further consideration, I think I want to make one more change: in the changelog, it's a bit awkward to handle all the old references to `trio.hazmat`, since on the one hand it's a historical document so you want to leave things alone, but on the other hand, it's something people read in the present.

The awkward compromise I did in the initial commit was to apply the search/replace to only the text that had live links. I think a better way would be to keep the text the same while fixing the link destinations only.

njsmith:
OK done, I think (hope?) this is ready now

oremanj:
Looks great, thank you!

pquentin:
Looks good to me too, thanks!

PR #1487, opened 2020-05-02
Fixed NotImplementedError for set_wakeup_fd on PyPy3/Windows

### LINKED ISSUE #1361: Trio does not work with PyPy3 on Windows
Apparently on pypy3+Windows, most trio functions raise `NotImplementedError: signal.set_wakeup_fd is not implemented on Windows`. Either somebody has to add that to pypy, or trio has to work around the limitation.

agronholm:
PyPy does not have `signal.set_wakeup_fd()` on Windows, but ctrl+c and ctrl+break work just fine without it.

Fixes #1361.

### DIFF
--- a/trio/_core/_wakeup_socketpair.py
+++ b/trio/_core/_wakeup_socketpair.py
@@ -60,7 +60,8 @@ class WakeupSocketpair:
 
     @contextmanager
     def wakeup_on_signals(self):
-        if not is_main_thread():
+        # PyPy does not have (or need) set_wakeup_fd() on Windows
+        if not is_main_thread() or not hasattr(signal, 'set_wakeup_fd'):
             yield
             return
         fd = self.write_sock.fileno()
--- a/trio/_core/_wakeup_socketpair.py
+++ b/trio/_core/_wakeup_socketpair.py
@@ -61,7 +61,7 @@ class WakeupSocketpair:
     @contextmanager
     def wakeup_on_signals(self):
         # PyPy does not have (or need) set_wakeup_fd() on Windows
-        if not is_main_thread() or not hasattr(signal, 'set_wakeup_fd'):
+        if not is_main_thread() or sys.implementation.name == 'pypy':
             yield
             return
         fd = self.write_sock.fileno()


agronholm:
Should we include pypy3 in the test suite as well?

njsmith:
I'm kind of uncomfortable merging something where we have no idea how or
why it works :-). PyPy can't magically change how Win32 APIs work...

The only thing I can think of is: is it possible that PyPy is running the
signal handler in a worker thread, instead of the main thread? I guess this
would be checkable with a little test program that printed the current
thread id from inside a signal handler?

njsmith:
Actually, let's start by just making sure we're on the same page. The
tricky case where you need set_wakeup_fd is when the user hits control-C
while the event loop is quiescent. So the test would be whether you can
break out of:

trio.run(trio.sleep_forever)

Is that what you observed with PyPy on windows?

agronholm:
I'm having some difficulties installing trio on pypy because it cannot find my compiler (which I do have installed).

njsmith:
That's weird... It shouldn't need a compiler at all, I thought. The only C
module we use is cffi, and that should be included with PyPy? (Though maybe
pip doesn't know that?)

agronholm:
It's trying to compile cffi because trio needs it and requires it if installing on Windows. Fixing the dependency so it doesn't install it on pypy would most likely do the trick.

agronholm:
Confirmed. Would you like a separate PR for that or should I amend this one?

agronholm:
This fix doesn't work. Somehow I thought the function didn't exist there but it was just a brain fart. A proper fix is coming right up.

agronholm:
Pressing ctrl+c twice during `trio.run(trio.sleep_forever)` does indeed exit the program.

njsmith:
Checked with the pypy devs:

Re: cffi: they say pypy does ship with cffi and the metadata so that pip will know that... but there was a new cffi release since the last pypy release, and trio wants that newest cffi to fix a memory leak, so pip is trying to upgrade it. The upgrade is doomed to futility (pypy always uses the built-in version of cffi even if you install a new one), so I guess restricting the cffi dependency to not-pypy would make sense. It's not a huge deal though because the next pypy release will catch up again anyway...

Re: signals: @arigo says he has no idea why this is working; pypy does the same thing as cpython, so it shouldn't work :-).

njsmith:
@agronholm Another random thought, just to double-check: can you confirm that when you say "Pressing ctrl+c twice during `trio.run(trio.sleep_forever)` does indeed exit the program.", you mean "exits with a nice `KeyboardInterrupt` traceback"?

agronholm:
> @agronholm Another random thought, just to double-check: can you confirm that when you say "Pressing ctrl+c twice during `trio.run(trio.sleep_forever)` does indeed exit the program.", you mean "exits with a nice `KeyboardInterrupt` traceback"?

Well...sometimes it does that with a single ctrl+c. If I have to press it twice, the process just exits without printing a traceback. If I press ctrl+break instead, it always gives me a KeyboardInterrupt traceback, originating from a random KeyError inside cffi.

agronholm:
> Re: signals: @arigo says he has no idea why this is working; pypy does the same thing as cpython, so it shouldn't work :-).

Did he then explain why PyPy did not implement `signal.set_wakeup_fd()`?

arigo:
@agronholm actually pypy3 has got ``signal.set_wakeup_fd`` on Windows.  Either you're using an old version of pypy3 or you know more (e.g. it doesn't work) which you should report in the pypy issue tracker.

agronholm:
> @agronholm actually pypy3 has got `signal.set_wakeup_fd` on Windows. Either you're using an old version of pypy3 or you know more (e.g. it doesn't work) which you should report in the pypy issue tracker.

Just to make sure, I upgraded to pypy3 v7.3.1 and tried `signal.set_wakeup_fd(5)` and got `NotImplementedError: signal.set_wakeup_fd is not implemented on Windows`. Are you getting a different result?

njsmith:
> which you should report in the pypy issue tracker.

https://foss.heptapod.net/pypy/pypy/issues/3220

agronholm:
> > which you should report in the pypy issue tracker.
> 
> https://foss.heptapod.net/pypy/pypy/issues/3220

Clearly you have done more digging into this than I have. Thank you for that :)

agronholm:
Since we have established that `signal.set_wakeup_fd()` does not work on Windows + PyPy3, would it not make sense to merge this PR? When the PyPy devs fix the issue on their end we can amend the code to conditionally allow this again based on the PyPy version. If we don't merge, then this combination remains completely unusable.

agronholm:
This was fixed in PyPy so we don't need this PR.

PR #1497, opened 2020-05-06
Better error messages from Final and NoPublicConstructor

njsmith:
This is a pretty trivial change, but I noticed it while working on
gh-1496 so I figured I'd just fix it.

### DIFF
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -216,7 +216,9 @@ class Final(BaseMeta):
         for base in bases:
             if isinstance(base, Final):
                 raise TypeError(
-                    "`%s` does not support subclassing" % base.__name__
+                    "the {!r} class does not support subclassing".format(
+                        base.__name__
+                    )
                 )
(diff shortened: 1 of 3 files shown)


trio/_util.py
```diff
@@ -216,7 +216,9 @@ def __new__(cls, name, bases, cls_namespace):
         for base in bases:
             if isinstance(base, Final):
                 raise TypeError(
-                    "`%s` does not support subclassing" % base.__name__
+                    "the {!r} class does not support subclassing".format(
+                        base.__name__
```
oremanj, line 220: Do we want to do `{__module__}.{__qualname__}` for maximum comprehensibility?

oremanj:
LGTM, one optional suggestion

trio/_util.py
njsmith: Good idea! Done.

njsmith:
(Also, whoa, I just realized we can use f-strings now!)

PR #1498, opened 2020-05-06
Remove deprecated `clear` method from `trio.Event`

pquentin:


pquentin:
The coverage drop is expected as I'm removing code.

Issue #1104, opened 2019-06-12
Should we have a conventional way to "spawn a process into a nursery"?


njsmith:
We have `run_process`, which is a convenient way to run a process like it's a regular subroutine, with error checking (`check=`), cancellation support, automatic waiting, etc. And we have `trio.Process`, which is a fully general way to spawn a process and then do *whatever* with it.

In the discussion in #872, @Badg [brought up](https://github.com/python-trio/trio/pull/872#discussion_r261434230) some dissatisfaction with both of these options. I *think* I mostly get the issue now, so I'll try to rephrase in my own words:

In some ways, when a task spawns a subprocess it's very similar to spawning a new task: the subprocess is a chunk of code that runs simultaneously with the Trio task, and they can potentially interact.

Trio has some strong and convenient conventions for spawning tasks:

- trio always makes sure that both tasks are finished before exiting the nursery block
- if the new task crashes, the exception propagates until caught, and automatically cancels sibling tasks
- if the original task crashes, the new task is automatically cancelled too
- if the surrounding context is cancelled, then this automatically cancels both tasks

It would be nice to be able to get this combination of features for subprocesses too. But we don't currently have any simple way to do that.

The closest thing we have is to use `async with` on a `Process` object. But that's stuck between two conflicting sets of conventions. The `Process` API mostly treats processes as independent *objects*, and our `async with process` mostly follows the general conventions for `async with <resource>`: it never cancels the body of the `with` (even if the process crashes). It doesn't care whether the body of the `with` block completed with an exception or not. And it doesn't do any particular checks on the subprocess's return code. This is also reasonable and what you'd expect for like, `with file_obj`. OTOH, the conventions above are specific to how Trio handles *code*, and really only show up with nurseries and nursery-like objects. Which a `Process`... kind of is, and kind of isn't.

I don't think it makes sense to try to wedge all those nursery semantics into `async with process_obj`. In particular, it would be very weird for `async with process_obj` to suddenly enable the equivalent of `run_process`'s `check=True`. And then we'd need some way to allow it to be overridden with `check=False` on a case-by-case basis, and there's no obvious way to spell that. And if we use `check=True` and the subprocess fails and the body fails, do we raise a `MultiError`? In multierror v2, does that mean we need to wrap all outgoing exceptions in `MultiError`? It would involve a lot of weirdness.

Another idea that's been proposed is to extend `run_process` so that it can be used like:

```python
process_obj = await nursery.start(run_process, ...)
```

[...]

PR #1501, opened 2020-05-08
Move towards making most public classes Final

njsmith:
This changes most of Trio's public classes so that attempts to
subclass them will produce a deprecation warning, with the idea that
once the deprecation period passes on we'll use metaclass=Final to
make subclassing an error.

See #1104 for discussion.

There are some special cases where we do allow subclassing: ABCs,
exceptions, enums, etc. I also added a test to make sure that every
public class is either Final, on its way to becoming Final, or else
falls into one of the special cases. The idea is that when we add new
classes in the future, this will force us to either mark them Final,
or else add a new special case, hopefully with a comment explaining
what makes the new class a special case.

### DIFF
--- a/newsfragments/1044.removal.rst
+++ b/newsfragments/1044.removal.rst
@@ -1,0 +1,17 @@
+Most of the public classes that Trio exports – like `trio.Lock`,
+`trio.SocketStream`, and so on – weren't designed with subclassing in
+mind. And we've noticed that some users were trying to subclass them
+anyway, and ending up with fragile code that we're likely to
+accidentally break in the future, or else be stuck unable to make
+changes for fear of breaking subclasses.
+
+There are also some classes that were explicitly designed to be
+subclassed, like the ones in `trio.abc`. Subclassing these is still
+supported. However, for all other classes, attempts to subclass will
+now raise a deprecation warning, and in the future will raise an
+error.
+
+If this causes problems for you, feel free to drop by our `chat room
+<https://gitter.im/python-trio/general>`__ or file a bug, to discuss
+alternatives or make a case for why some particular class should be
+designed to support subclassing.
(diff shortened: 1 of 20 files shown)


newsfragments/1044.removal.rst
```diff
@@ -0,0 +1,17 @@
+Most of the public classes that Trio exports – like `trio.Lock`,
+`trio.SocketStream`, and so on – weren't designed with subclassing in
+mind. And we've noticed that some users were trying to subclass them
+anyway, and ending up with fragile code that we're likely to
+accidentally break in the future, or else be stuck unable to make
+changes for fear of breaking subclasses.
+
+There are also some classes that were explicitly designed to be
+subclassed, like the ones in `trio.abc`. Subclassing these is still
```
pquentin, line 9: For what it's worth, the docs check is failing with `/home/runner/work/trio/trio/docs/source/history.rst:37:py:obj reference target not found: trio.abc`

njsmith: Ah, whoops, thanks for the heads up! I forgot we don't have a dedicated page for `trio.abc`.

trio/_highlevel_generic.py
```diff
@@ -35,7 +37,9 @@ async def aclose_forcefully(resource):
 
 
 @attr.s(eq=False, hash=False)
-class StapledStream(HalfCloseableStream):
+class StapledStream(
```
oremanj, line 40: I think StapledStream is safe enough to allow subclassing of, since it has no private attributes.

njsmith: Yeah, the cost of supporting subclasses would be relatively low compared to other things. But that's only the "cost" side of the cost/benefit analysis :-). Is there some reason `StapledStream` subclassing is particularly valuable? I'd rather keep things simple until there's a good reason not too.

oremanj: I don't have an immediate use case and am happy to leave it non-subclassable for consistency.

njsmith: :+1: We can always change it later.

PR #1505, opened 2020-05-11
Use Ubuntu 20.04 in Travis CI

pquentin:
I used `gsutil ls gs://travis-ci-language-archives/python/binaries/ubuntu/20.04/x86_64/` to see what versions of Python were available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 os: linux
 language: python
-dist: bionic
+dist: focal
 
 jobs:
   include:
     # The pypy tests are slow, so we list them first
     - python: pypy3.6-7.2.0
+      dist: bionic
     - language: generic
       env: PYPY_NIGHTLY_BRANCH=py3.6
     # Qemu tests are also slow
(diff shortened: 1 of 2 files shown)


njsmith:
Huh, interesting. I don't think this is official yet – it's certainly not listed as a legal environment here :-) https://docs.travis-ci.com/user/reference/overview/

So maybe we want to wait a bit? Or maybe we should say screw it, it seems to work, and I guess it even lets us drop the elaborate qemu stuff for now?

pquentin:
I'm in the "screw it" camp, yes. :-)

Why can we drop the elaborate qemu stuff? What would it be replaced with?

njsmith:
The point of the elaborate qemu stuff is to test on a newer kernel than the ones available on our CI providers. But if Travis is shipping 20.04, then they (I assume) now have a newer kernel than our qemu stuff :-)

We might want to shift the qemu stuff to something even newer, like Fedora 32 w/ kernel 5.6, e.g if we start playing with io_uring (#932). But jumping through hoops to test on 19.10 is pretty redundant if we can test on 20.04 directly :-)

pquentin:
Oh, yes, that makes sense, thanks. Indeed, we already have 5.4 with Travis.

```
Linux travis-job-07776598-16a0-4e37-958f-a7e733b9a94e 5.4.0-14-generic #17-Ubuntu SMP Thu Feb 6 22:47:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
Linux test-host 5.4.0-29-generic #33-Ubuntu SMP Wed Apr 29 14:32:27 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

I'll experiment with Fedora 32.

pquentin:
Included in https://github.com/python-trio/trio/pull/1507

Issue #733, opened 2018-10-13
Rethink KeyboardInterrupt protection APIs


njsmith:
Sometime before 1.0, we should go through the KeyboardInterrupt protection APIs and think them through again.

There seem to be a limited number of patterns: decorating functions/methods that should always have KI protection enabled; setting the default protection in a new task (can be overridden if the function has a preference); temporarily disabling and then reverting to what it was before (especially, reverting inside a call to `outcome.capture` or `acapture`). Some of these are kind of awkward with the current API, or impossible (in some places I think we do disable/enable, rather than disable/revert-to-previous-state).

It would be nice to fix #550. This might be doable on 3.7+ by exploiting the atomicity of `Context.run`?

We'll also want to look again at the issues around atomicity when exiting context managers, I guess.

Should tasks inherit their default KI protection state from the parent nursery?

oremanj:
I've been looking at this some yesterday and today. I have a few possible approaches that I'd like feedback on.

## The evolutionary one

I'd like to see us move away from the magic locals key if we can. It adds some runtime overhead at most protection status changes, adds extra stack frames in tracebacks, makes `@enable_ki_protected`-ified async functions return false from `inspect.iscoroutinefunction()`, and is noisy when inspecting locals in a debugger. Instead, I think we should use some sort of static mechanism (based on what code you're executing right now), plus a dynamic/policy-level mechanism (so the same function/context can allow KI in some cases and not in others). KI is only delivered immediately if both mechanisms say it's fine.

The dynamic mechanism can just be a contextvar as long as the static mechanism protects all the contextvar machinery. We'll use this to provide "system tasks start out KI-protected", and provide context managers for toggling it back and forth. Then roundtripping through a thread (the only place in Trio where we explicitly disable KI protection) will automatically preserve KI status once it preserves other contextvars (#178).

I've prototyped this with the static mechanism being a dictionary from code objects to protection state changes. The KI protection decorators just add the decorated function's code object to the dictionary. This works pretty well; it's a little more expensive at import time but effectively free when going from unprotected to protected or vice versa at runtime. It also lets us KI-protect code we don't control, such as the plumbing of `contextvars.Context.run`, `outcome.acapture`, `contextlib._GeneratorContextDecorator`, and so on.

## The revolutionary one

[...]

PR #1508, opened 2020-05-11
[WIP] Track KeyboardInterrupt protection per code object

oremanj:
Example implementation for the first idea from my comment on #733. Still somewhat unpolished (needs a newsfragment, doc updates, and additional tests) but I didn't want to put more work into it without feedback on the overall direction.


njsmith:
I like the overall approach. I wonder if we can simplify it though? In particular having 4 different protection states seems like a lot.

- "This and all its callees are protected" makes sense
- "This is protected but its callees aren't": I think this really only applies to `contextlib` and `contextvars`, so maybe we can just special-case those namespaces instead of tracking them via code objects?
- "Transparent": This is only for the benefit of `currently_ki_protected`, right, and only so it can give the right result when called from a "protected but its callees aren't" function? So (a) if we remove the general category of "protected but its callees aren't" then I think it's unnecessary, and (b) anyway, can't it just do `sys._getframe(1)` to test the caller's frame instead of its own?

The trickier one is "This and its callees are *not* protected". As noted in #733, I'm not convinced that `@disable_ki_protection` is really a feature we even want? Maybe instead what we want is a flag that says "if your stack walk reaches this frame, then stop and trust the contextvar instead"?

oremanj:
> "This is protected but its callees aren't": I think this really only applies to contextlib and contextvars, so maybe we can just special-case those namespaces instead of tracking them via code objects?

I also used it to remove one function call layer in `_thread`, but we could just put back the extra function call layer there.

> "Transparent": This is only for the benefit of currently_ki_protected, right, and only so it can give the right result when called from a "protected but its callees aren't" function? So (a) if we remove the general category of "protected but its callees aren't" then I think it's unnecessary, and (b) anyway, can't it just do sys._getframe(1) to test the caller's frame instead of its own?

Agreed on both counts, except that I think we would need to apply this to `signal_raise()` and its callees also, if we want to be able to fully test the "protected but callees aren't" case. But if we get rid of that case, we get rid of the need for this one too.

> The trickier one is "This and its callees are not protected". As noted in #733, I'm not convinced that @disable_ki_protection is really a feature we even want? Maybe instead what we want is a flag that says "if your stack walk reaches this frame, then stop and trust the contextvar instead"?

That's actually how it works here -- if you say @disable_ki_protection, then KI deliverability follows the contextvar. Probably the name should change at minimum.

More discussion in #733, which I'm working on writing up a response to now. :-)

PR #1507, opened 2020-05-11
Test Fedora 32 VM to get Linux 5.6

pquentin:
This includes the newer io_uring.

Interestingly, though, we get this failure on Fedora 32 with Linux 5.6:

```
[  178.198410] cloud-init[668]: =================================== FAILURES ===================================
[  178.210646] cloud-init[668]: __________________________________ test_pipes __________________________________
[  178.222895] cloud-init[668]:     async def test_pipes():
[  178.231072] cloud-init[668]:         async with await open_process(
[  178.240076] cloud-init[668]:             COPY_STDIN_TO_STDOUT_AND_BACKWARD_TO_STDERR,
[  178.250236] cloud-init[668]:             stdin=subprocess.PIPE,
[  178.258906] cloud-init[668]:             stdout=subprocess.PIPE,
[  178.267381] cloud-init[668]:             stderr=subprocess.PIPE,
[  178.275827] cloud-init[668]:         ) as proc:
[  178.282874] cloud-init[668]:             msg = b"the quick brown fox jumps over the lazy dog"
[  178.293632] cloud-init[668]:     
[  178.299695] cloud-init[668]:             async def feed_input():
[  178.307731] cloud-init[668]:                 await proc.stdin.send_all(msg)
[  178.316621] cloud-init[668]:                 await proc.stdin.aclose()
[  178.324972] cloud-init[668]:     
[  178.330520] cloud-init[668]:             async def check_output(stream, expected):
[  178.340674] cloud-init[668]:                 seen = bytearray()
[  178.349319] cloud-init[668]:                 async for chunk in stream:
[  178.358565] cloud-init[668]:                     seen += chunk
[  178.367056] cloud-init[668]:                 assert seen == expected
[  178.376660] cloud-init[668]:     
[  178.382834] cloud-init[668]:             async with _core.open_nursery() as nursery:
[  178.393086] cloud-init[668]:                 # fail quickly if something is broken
[  178.403041] cloud-init[668]:                 nursery.cancel_scope.deadline = _core.current_time() + 3.0
[  178.414754] cloud-init[668]:                 nursery.start_soon(feed_input)
[  178.424084] cloud-init[668]:                 nursery.start_soon(check_output, proc.stdout, msg)
[  178.435395] cloud-init[668]:                 nursery.start_soon(check_output, proc.stderr, msg[::-1])
[  178.446341] cloud-init[668]:     
[  178.452663] cloud-init[668]: >           assert not nursery.cancel_scope.cancelled_caught
[...]

PR #1510, opened 2020-05-11
Newer Linux kernels in CI

pquentin:
With this change, we'll use Ubuntu 20.04 in most Travis builds, which will allow us to continue testing pidfd, but without using a VM. We'll also use Linux 5.6 with Fedora 32, which will allow to add tests for io_uring in the future. It's not clear yet whether we want to do that now or not, but I wanted to see if it was possible. This is exactly https://github.com/python-trio/trio/pull/1507, but in #1507 GitHub did not see that Travis had succeeded.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ jobs:
       # 'bionic' systems have nested KVM enabled.
       dist: bionic
       env:
-        - "JOB_NAME='Ubuntu 19.10, full VM'"
-        - "LINUX_VM_IMAGE=https://cloud-images.ubuntu.com/eoan/current/eoan-server-cloudimg-amd64.img"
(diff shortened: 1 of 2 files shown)


pquentin:
@njsmith Thanks for the clarification! However, I think you edited the FreeBSD section of ci.sh instead of the Linux one here. Was it intended?

.travis.yml
```diff
@@ -1,17 +1,17 @@
 os: linux
 language: python
-dist: bionic
+dist: focal
```
auvipy, line 3: is it available on travis for all repo?

pquentin: Yes, but it's not official, I've only seen this used in Travis owned repos. See https://github.com/python-trio/trio/pull/1505 for more discussion about this

Issue #1504, opened 2020-05-10
Expose most of trio._util and trio._deprecate for the benefit of the Trio ecosystem


oremanj:
Trio has great tools for dealing with some of the minutiae of making a polished Python package in general, but they're private; libraries that depend on Trio need to copy-paste or reimplement these tools if they want to offer equivalent functionality. If they reimplement, they probably won't do as good of a job.

I think we should add a new public Trio submodule (`trio.libtools`?) which provides:
* From `trio._util`: Final, NoPublicConstructor, generic_function, fixup_module_metadata, async_wraps, ConflictDetector
* A factory for creating modules like `trio._deprecate`, substituting a project's own deprecation warning, issue URL, and name.

oremanj:
Also useful: extract and publicize the logic from Runner.spawn_impl() for "either get a coroutine from this async function+args or raise an error explaining why it's not in the form I wanted" and "turn this function/name into a string". They're very useful for duck-nurseries (I have a hack in tricycle.open_service_nursery() working around the absence of the error-reporting one), and I think would find other uses too (#1513 made me think of this).

njsmith:
> extract and publicize the logic from Runner.spawn_impl() for "either get a coroutine from this async function+args or raise an error explaining why it's not in the form I wanted"

Heh, I was just wondering whether to comment on #1513 to suggest this before I saw your post :-). Maybe `trio.lowlevel.trio_async_fn_to_coroutine_object(afn, *args, **kwargs)`?

> A factory for creating modules like trio._deprecate, substituting a project's own deprecation warning, issue URL, and name.

`trio._deprecate` is definitely a substantial piece of code, and it would make a ton of sense to extract into a standalone library. The reason I haven't yet is just because, y'know, making sure an interface is general enough to make public takes some work, and there are a lot of different ways that different projects like to approach to deprecation (e.g. there are multiple existing deprecation-helper libraries on PyPI already but I didn't like any of the ones I found). If someone wants to take this on though then I'd be supportive :-). Making it part of `trio`'s public interface would feel weird though. Even just within `python-trio`, we have lots of libraries that might need to deprecate things but that don't otherwise depend on trio :-)

> From trio._util: Final, NoPublicConstructor, generic_function, fixup_module_metadata, async_wraps, ConflictDetector

[...]

PR #1513, opened 2020-05-12
Fixed unhelpful error message in from_thread_run functions.

### LINKED ISSUE #1244: In trio.to_thread.run, give a better error message if passed a sync function
I needed to decouple sender and receiver and initialized the channel with ```max_buffer_size>0```. Lost elements are accepted when buffer is full, but ```sendchannel.send_nowait``` throws 

```
TypeError: object NoneType can't be used in 'await' expression
```
I had expected that the channel could be filled up to max_buffer_size. 
For demonstration I made minor changes to the threaded channel example: 

``` python
import trio

def thread_fn(receive_from_trio, send_to_trio):
    while True:
        # Since we're in a thread, we can't call methods on Trio
        # objects directly -- so we use trio.from_thread to call them.
        try:
            request = trio.from_thread.run(receive_from_trio.receive)
            response = request + 1
            trio.from_thread.run(send_to_trio.send_nowait, response)
        except trio.WouldBlock:
            print('would block')
        except trio.EndOfChannel:
            trio.from_thread.run(send_to_trio.aclose)
            return

async def main():
    send_to_thread, receive_from_trio = trio.open_memory_channel(1)
    send_to_trio, receive_from_thread = trio.open_memory_channel(1)

        # In a background thread, run:
        #   thread_fn(portal, receive_from_trio, send_to_trio)
        nursery.start_soon(
            trio.to_thread.run_sync, thread_fn, receive_from_trio, send_to_trio
        )

        # prints "1"
        await send_to_thread.send(0)
        
        # simulate a slow starting receiver
        await trio.sleep(2)
        print(await receive_from_thread.receive())       

        # prints "2"
        await send_to_thread.send(1)
        print(await receive_from_thread.receive())

        # When we close the channel, it signals the thread to exit.
        await send_to_thread.aclose()

        # When we exit the nursery, it waits for the background thread to
        # exit.


trio.run(main)
```

guilledk:
Added helpful error message in the case the user passed incorrect type of function to both "from_thread_run" functions.

Closes #1244


guilledk:
I'm not really familiar with the whole CI thing, so just point me to some docs and I'll fix the test fails.

Didn't expect a simple error message would break so many tests :P

trio/_threads.py
```diff
@@ -377,6 +379,9 @@ def from_thread_run(afn, *args, trio_token=None):
           "foreign" thread, spawned using some other framework, and still want
           to enter Trio.
     """
+    if not inspect.iscoroutinefunction(afn):
```
oremanj, line 382: Unfortunately, identifying async functions is rather subtler than just `inspect.iscoroutinefunction`, because async functions can be wrapped in a way that doesn't satisfy that test. (For example, Trio functions decorated with `@enable_ki_protection` are implemented using a wrapper that looks synchronous. `functools.partial` objects don't even satisfy `inspect.isfunction`, let alone `inspect.iscoroutinefunction`. All of these are perfectly valid to pass as an "async function" as long as the function being wrapped is async.)

The best way to see if something is an async function is to call it without `await` and use `inspect.iscoroutine` on the result. Of course, if it wasn't an async function, you just executed it, so it's best to do this only in the context where you were planning to call it anyway. In this case, that would be inside `unprotected_afn` below.

If you're not willing to call the function, you have to get a lot more special-cased to see if it's async or not. You can look inside the `.func` attribute of a `functools.partial` object, follow the `.__wrapped__` attributes generally added by function wrappers, etc, but there's no approach that will definitely succeed for any async callable.

trio/_threads.py
```diff
@@ -422,6 +427,9 @@ def from_thread_run_sync(fn, *args, trio_token=None):
           "foreign" thread, spawned using some other framework, and still want
           to enter Trio.
     """
+    if not (inspect.isfunction(fn) and not inspect.iscoroutinefunction(fn)):
```
oremanj, line 430: As discussed above, these won't tell you what you want. Better to wait and raise an error if the result of calling the function satisfies `inspect.iscoroutine`. (If it does, you'll also want to `close()` it before returning, so you don't get an extra "coroutine was never awaited" warning.)

trio/_threads.py
```diff
@@ -377,6 +379,9 @@ def from_thread_run(afn, *args, trio_token=None):
           "foreign" thread, spawned using some other framework, and still want
           to enter Trio.
     """
+    if not inspect.iscoroutinefunction(afn):
+        raise AttributeError("afn must be an asynchronous function")
```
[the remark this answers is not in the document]
guilledk: Great! I knew the error message needed some work.

guilledk:
Thanks a lot for the feedback!, I'll work on it and send another PR!

trio/_threads.py
guilledk: Just saw your comment on #1504, should I look into `Runner.spawn_impl()` for the "async fn or error" logic?

njsmith:
@guilledk Note that you don't have to actually create a new PR from scratch – you can just more commits to your `issue1244` branch on github, and they'll automatically appear here in this PR.

guilledk:
Ok good to know.

Thanks a lot for the patience, you guys rock!

trio/_threads.py
oremanj: That would be a useful extension if you feel like it! The relevant code is toward the top of spawn_impl(), with the comments about "Give good error for [...]"

guilledk:
So following what @oremanj said in #1504, I moved the "either get a coroutine from this async function+args or raise an error explaining why it's not in the form I wanted" logic to a function in `trio._util` called `coroutine_or_error` (WIP name).

Added the check to `trio._threads.from_thread_run`, but its still missing the check on `trio._threads.from_thread_run_sync`, I'll do that one tomorrow.

Also there is some code formatting issues I have to fix.

guilledk:
Cool, all that's missing are the tests!

guilledk:
@njsmith Hey so I could use some pointers to fix the code coverage issues, I check the codecov diff and I don't seem to get were the issue is. Thanks in advance!

pquentin:
The coverage drop is in trio/socket.py: https://codecov.io/gh/python-trio/trio/pull/1513/src/trio/socket.py

Screenshot 2020-05-14 at 15 26 38

The issue is that the macOS 3.6 build was cancelled after 1m30s, I don't know why: https://github.com/python-trio/trio/pull/1513/checks?check_run_id=672946337

Closing/reopening to re-run all jobs

guilledk:
Cool so this closes #1244

pquentin:
Great that the tests and coverage pass!

@guilledk Note that you need to edit your first message to say that, otherwise GitHub won't close the issue when we merge this.

guilledk:
@pquentin Done!

trio/_core/_run.py
```diff
@@ -1243,86 +1242,8 @@ def spawn_impl(self, async_fn, args, nursery, name, *, system_task=False):
         if nursery is None:
             assert self.init_task is None
 
-        ######
-        # Call the function and get the coroutine object, while giving helpful
```
oremanj, line 1247: This is a better description than the "Check if async_fn is callable coroutine" that you added below, so I'd suggest you keep it. (The function is not a "callable coroutine" -- it's a callable that returns a coroutine object, but that's just another way of saying "async function" which is Trio's preferred term. Coroutine objects themselves are not callable.)

trio/_threads.py
```diff
@@ -404,12 +408,13 @@ def from_thread_run_sync(fn, *args, trio_token=None):
         RunFinishedError: if the corresponding call to `trio.run` has
             already completed.
         Cancelled: if the corresponding call to `trio.run` completes
```
oremanj, line 410: Not your fault, but this is copypasta from the async version -- sync functions can't raise Cancelled. So you can just remove the description of Cancelled as an exception-that-might-be-raised entirely.

trio/_threads.py
```diff
@@ -404,12 +408,13 @@ def from_thread_run_sync(fn, *args, trio_token=None):
...
-            while ``afn(*args)`` is running, then ``afn`` is likely to raise
+            while ``fn(*args)`` is running, then ``fn`` is likely to raise
             :exc:`trio.Cancelled`, and this will propagate out into
         RuntimeError: if you try calling this from inside the Trio thread,
             which would otherwise cause a deadlock.
         AttributeError: if no ``trio_token`` was provided, and we can't infer
-            one from context.
+            one from context. Also if ``fn`` is not a sync function.
```
oremanj, line 416: This is leftover from your previous use of AttributeError for this condition, but you're now using TypeError.

trio/_threads.py
```diff
@@ -422,6 +427,13 @@ def from_thread_run_sync(fn, *args, trio_token=None):
           "foreign" thread, spawned using some other framework, and still want
           to enter Trio.
     """
+
+    if not callable(fn) or inspect.iscoroutinefunction(fn):
```
oremanj, line 431: I actually don't think you need to explicitly check whether it's callable, because the TypeError you get from calling a non-callable will properly propagate back into the caller of `trio.from_thread.run_sync` when you try to call the function below.

As I explained in chat, `inspect.iscoroutinefunction` will not catch all async callables. That's OK here, but you will give a better user experience (ie, catch more problem cases) if you wait until after you've called the function, and then throw an error if the value you got back from it satisfies `inspect.iscoroutine`.

trio/_util.py
```diff
@@ -85,6 +88,87 @@ def is_main_thread():
...
+
+    # We can't check iscoroutinefunction(async_fn), because that will fail
+    # for things like functools.partial objects wrapping an async
+    # function. So we have to just call it and then check whether the
+    # return value is a coroutine object.
+    if not isinstance(coro, collections.abc.Coroutine):
+        # Give good error for: nursery.start_soon(func_returning_future)
+        if _return_value_looks_like_wrong_library(coro):
+            raise TypeError(
+                "start_soon got unexpected {!r} – are you trying to use a "
```
oremanj, line 151: Suggest you change "start_soon" to "Trio" in this error and the one below. This is for consistency with the other error messages. start_soon was already incomplete in which cases it covered (you can get the same error from trio.run() or nursery.start()) but now it's even more so (since now from_thread.run(), and later potentially even more functions, will be using this helper too).

trio/_util.py
```diff
@@ -85,6 +88,87 @@ def is_main_thread():
         return False
 
 
+######
+# Call the function and get the coroutine object, while giving helpful
+# errors for common mistakes. Returns coroutine object.
+######
+def coroutine_or_error(async_fn, args=[]):
```
oremanj, line 95: Trio convention is to take `*args` in this case -- we almost never pass around the argument tuple as a tuple, even internally.

trio/_threads.py
```diff
@@ -422,6 +427,13 @@ def from_thread_run_sync(fn, *args, trio_token=None):
...
+        raise TypeError(
+            "Trio expected a sync function, but {!r} appears to not be "
+            "callable or asynchronous".format(getattr(fn, "__qualname__", fn))
```
oremanj, line 434: You know which it is (is it non-callable or is it an async function?), so the error message should say which it is.

Also, be extra careful about grammatical ambiguity in error messages, since we want them to be as comprehensible as possible. In this case "not be callable or asynchronous" could be read as "not be (callable or asynchronous)" where what you mean is "(not be callable) or (be asynchronous)". I would write that as "is not callable or is an async function".

trio/_core/tests/test_run.py
```diff
@@ -1696,74 +1696,6 @@ async def test_current_effective_deadline(mock_clock):
     assert _core.current_effective_deadline() == inf
 
 
-# @coroutine is deprecated since python 3.8, which is fine with us.
-@pytest.mark.filterwarnings("ignore:.*@coroutine.*:DeprecationWarning")
-def test_nice_error_on_bad_calls_to_run_or_spawn():
-    def bad_call_run(*args):
```
oremanj, line 1702: You've removed the test that passing these things to run() or start_soon() actually results in an error (ie, that your now factored-out helper actually gets called). You don't need to duplicate the full suite of tests, but you should still pass one or two error-provoking arguments and ensure that errors get produced.

trio/tests/test_threads.py
```diff
@@ -471,6 +471,15 @@ def thread_fn():
...
     assert isinstance(trio_time, float)
 
+    # Test correct error when passed async function
+    async def async_fn():  # pragma: no cover
+        pass
+
+    with pytest.raises(TypeError) as excinfo:
+        from_thread_run_sync(async_fn)
+
+    assert "expected a sync function" in str(excinfo.value)
```
oremanj, line 481: It's nicer to write this as `with pytest.raises(TypeError, match="expected a sync function"):`

trio/tests/test_threads.py
```diff
@@ -488,6 +497,18 @@ def thread_fn():
...
 
+    # Test correct error when passed sync function
+    def sync_fn():  # pragma: no cover
+        pass
+
+    def thread_fn():
+        from_thread_run(sync_fn)
+
+    with pytest.raises(TypeError) as excinfo:
+        await to_thread_run_sync(thread_fn)
```
oremanj, line 508: You don't need the separate thread_fn. You can say `await to_thread_run_sync(from_thread_run, sync_fn)`

trio/_threads.py
njsmith: Python already gives a decent error message if you try to call something that's non-callable, so we could drop that check entirely and just let Python raise the exception for us.

oremanj: Oh, whoops, I thought I had left a comment about that but it seems I didn't.

I was also going to suggest that instead of `inspect.iscoroutinefunction` (which won't catch async methods, async functions used with `partial`, etc), you wait until you call the function below and then raise an error if the return value satisfies `inspect.iscoroutine`.

trio/_core/tests/test_run.py
guilledk: Ok, added test for passing a sync function and an async generator.

trio/_core/tests/tutil.py
```diff
@@ -52,6 +54,24 @@ def gc_collect_harder():
         gc.collect()
 
 
+# Some of our tests need to leak coroutines, and thus trigger the
+# "RuntimeWarning: coroutine '...' was never awaited" message. This context
+# manager should be used anywhere this happens to hide those messages, because
+# when expected they're clutter.
+@contextmanager
+def ignore_coroutine_never_awaited_warnings():
```
guilledk, line 62: Moved this here because its used in several tests now.

trio/_threads.py
```diff
@@ -425,7 +427,16 @@ def from_thread_run_sync(fn, *args, trio_token=None):
     def callback(q, fn, args):
         @disable_ki_protection
         def unprotected_fn():
-            return fn(*args)
+            call = fn(*args)
+
+            if inspect.iscoroutine(call):
+                call.close()
```
guilledk, line 433: To avoid `RuntimeWarning`s

trio/tests/test_threads.py
```diff
@@ -13,7 +13,7 @@
 )
 
 from .._core.tests.test_ki import ki_self
-from .._core.tests.tutil import slow
```
guilledk, line 16: Unused import

trio/_core/_run.py
```diff
@@ -1247,86 +1246,7 @@ def spawn_impl(self, async_fn, args, nursery, name, *, system_task=False):
...
-            # Give good error for: nursery.start_soon(some_sync_fn)
-            raise TypeError(
-                "Trio expected an async function, but {!r} appears to be "
-                "synchronous".format(
-                    getattr(async_fn, "__qualname__", async_fn)
-                )
-            )
-
-        ######
-        # Set up the Task object
```
oremanj, line 1328: This comment ("Set up the Task object") is still applicable to the code below, so you shouldn't remove it.

trio/_threads.py
```diff
@@ -425,7 +427,16 @@ def from_thread_run_sync(fn, *args, trio_token=None):
     def callback(q, fn, args):
         @disable_ki_protection
         def unprotected_fn():
-            return fn(*args)
+            call = fn(*args)
```
oremanj, line 430: This is the value returned by the function, so `call` is not a very clear name for it. `result` or `res` or `ret` perhaps?

trio/_threads.py
oremanj: Please say that in a comment in the code, so that people reading the code will understand it. Comments in the PR don't help anyone except the people reviewing the PR.

trio/_threads.py
oremanj: You marked this as resolved but you didn't update the docstring.

trio/_core/tests/test_run.py
```diff
@@ -1709,59 +1693,25 @@ async def main():
...
+    for bad_call in bad_call_run, bad_call_spawn:
 
-    with ignore_coroutine_never_awaited_warnings():
-        for bad_call in bad_call_run, bad_call_spawn:
+        async def f():  # pragma: no cover
+            pass
 
-            async def f():  # pragma: no cover
-                pass
+        with ignore_coroutine_never_awaited_warnings():
```
oremanj, line 1701: You don't need this anymore since you added the logic to close the passed coroutine, right?

trio/_core/tests/test_run.py
```diff
@@ -1709,59 +1693,25 @@ async def main():
...
+            with pytest.raises(TypeError, match="expecting an async function"):
```
oremanj, line 1703: No need for an extra blank line after `with pytest.raises`.

trio/_core/tests/test_run.py
guilledk: Yep! you're right

trio/_threads.py
oremanj: You still have not updated the docstring despite having marked this as resolved again. The request is to remove the text "Also if fn is not a sync function" from the description for AttributeError, because you no longer raise AttributeError in that case.

trio/_threads.py
```diff
@@ -427,16 +427,18 @@ def from_thread_run_sync(fn, *args, trio_token=None):
     def callback(q, fn, args):
         @disable_ki_protection
         def unprotected_fn():
-            call = fn(*args)
+            ret = fn(*args)
 
-            if inspect.iscoroutine(call):
-                call.close()
+            if inspect.iscoroutine(ret):
+
```
oremanj, line 433: See comment above about extra blank lines after block introducers

trio/_core/tests/test_run.py
oremanj: You marked this as resolved but didn't change anything in your PR. Do you feel like the extra blank lines here are important for readability in a way that's related to the changes you're making here? It seems like you've added a number of them in this diff after block introducers (for, with, if, etc). That's not consistent with the style in the rest of the codebase.

trio/_threads.py
guilledk: Ups sorry about that, it's burned into my brain

guilledk:
Ok now I think I got all the formatting issues, sorry guys very used to my code style. Thanks again for the patience.

oremanj:
Thank you for *your* patience! I removed a few more of the extraneous blank lines, and will merge this once the CI finishes.

oremanj:
Also, since I made some trivial commits above, you'll want to pull from your PR branch before you make additional changes. Sorry for the trouble!

guilledk:
No trouble at all! Now I understand the workflow a lot more, next time I'll take less reviews I swear haha

newsfragments/1244.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added helpful error message in the case the user passed incorrect type of function to "from_thread_run" & "from_thread_run_sync" functions.
```
oremanj, line 1:
```suggestion
Added a helpful error message if a sync function is passed to `trio.from_thread.run_sync` or an async function to `trio.from_thread.run`.
```

Reworded for clarity and to link to the documentation of those functions (that's what the single backticks do).

pquentin:
Congratulations @guilledk! How was the experience? Would you like to work on another issue next? https://github.com/python-trio/trio/issues/1532, for example?

guilledk:
Thanks @pquentin!, Awsome experience, to be honest never wrote a test before this PR, now all my projects have tests, so I learned a thing already.

I was actually looking at #1532 to do next, I really like the `trio` community and technology, I'd like to keep learning about the backend and helping you guys. Right now I have project that runs on top of `trio` and might go into production soon so I'd like to give back!

pquentin:
@guilledk Thanks, that's great to hear! If you decide to work on #1532, please ask if you have any questions, that might avoid a round of feedback or two. :)

Issue #1500, opened 2020-05-07
Remove unused __all__ variables from Trio's modules


njsmith:
This is a minor issue, so let's leave it for someone who hasn't made any PRs yet and wants a simple one to practice one.

The issue: lots of Trio's modules have `__all__` variables defined. That's because once upon a time, `trio/__init__.py` used these `__all__` variables to figure out what to re-export as part of the main namespace. But we stopped doing that a long time ago, because it confused static analysis tools. So now the `__all__` variables are vestigial, aren't being updated, and don't actually do anything except clutter up the code, and potentially confuse readers. So we should do a sweep through and delete them all.

PR #1515, opened 2020-05-13
[#1500]: Remove unused __all__ variables from Trio modules

kaldown:
#1500 

In addition, added `pyvenv.cfg` into `.gitignore`

### DIFF
--- a/trio/_highlevel_open_unix_stream.py
+++ b/trio/_highlevel_open_unix_stream.py
@@ -10,8 +10,6 @@ try:
 except ImportError:
     has_unix = False
 
-__all__ = ["open_unix_socket"]
-
 
 @contextmanager
 def close_on_error(obj):
--- a/trio/testing/_network.py
+++ b/trio/testing/_network.py
@@ -1,7 +1,5 @@
 from .. import socket as tsocket
-from .._highlevel_socket import SocketListener, SocketStream
-
-__all__ = ["open_stream_to_socket_listener"]
+from .._highlevel_socket import SocketStream
 
 
 async def open_stream_to_socket_listener(socket_listener):
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ pip-log.txt
 htmlcov/
 .tox/
 .venv/
+pyvenv.cfg
 .coverage
 .coverage.*
 .cache
--- a/trio/_core/_entry_queue.py
+++ b/trio/_core/_entry_queue.py
@@ -7,8 +7,6 @@ from .. import _core
 from .._util import NoPublicConstructor
 from ._wakeup_socketpair import WakeupSocketpair
 
-__all__ = ["TrioToken"]
-
 
 @attr.s(slots=True)
 class EntryQueue:
--- a/trio/_core/_ki.py
+++ b/trio/_core/_ki.py
@@ -12,12 +12,6 @@ if False:
     from typing import Any, TypeVar, Callable
     F = TypeVar('F', bound=Callable[..., Any])
 
-__all__ = [
-    "enable_ki_protection",
-    "disable_ki_protection",
-    "currently_ki_protected",
-]
-
 # In ordinary single-threaded Python code, when you hit control-C, it raises
 # an exception and automatically does all the regular unwinding stuff.
 #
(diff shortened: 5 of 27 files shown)


trio/testing/_network.py
```diff
@@ -1,7 +1,5 @@
 from .. import socket as tsocket
-from .._highlevel_socket import SocketListener, SocketStream
-
-__all__ = ["open_stream_to_socket_listener"]
+from .._highlevel_socket import SocketStream
```
kaldown, line 2: mystery, looks like it's not used. Tests passed. `yapf` won't show anything before and after. But making an error in that code, by removing used `SocketStream`, breaks tests. So I think this is legit change? But want to make sure

trio/_highlevel_open_unix_stream.py
```diff
@@ -10,14 +10,12 @@
...
 
-__all__ = ["open_unix_socket"]
-
 
 @contextmanager
 def close_on_error(obj):
     try:
         yield obj
-    except:
+    except Exception:
```
njsmith, line 18: This is a surprising change to find in here, since it's not mentioned in the description :-). Also, I'm pretty in this case `except Exception:` is wrong, and the bare `except:` was correct?

kaldown: I did it automatically :) Sorry I forgot to mention.
As for the question, https://docs.python.org/2/howto/doanddont.html#index-0 + my default linter showed this is a bad idea.
Why do you think catching the base `Exception` is wrong? Is this a chance something unhandled can happen?

kaldown: You're right, it's better not to change something that works. I will remove that. Just curious, why this isn't a good idea to explicitly catch `Exception` rather than **everything**.

njsmith: There is a rule of thumb that you shouldn't use bare `except:`, but, that rule is assuming a specific context, that doesn't apply here :-).

Often, programmers want to say "try to do this, and if it fails, I don't care why, try something else". Like you might try the operation again, or try doing it another way. And in this case new Python programmers often write `except:`, because they think it means "I don't care why it failed, just that it failed". But, that's not really what bare `except:` does! The problem is that there are some special exceptions, like `KeyboardInterrupt`, that don't mean the thing you were trying failed, they mean something else – like for `KeyboardInterrupt`, it means "the user hit control-C and wants your program to exit". In that case, retrying the operation is definitely not the right response :-). So if you use bare `except:`, then maybe your program will accidentally swallow and hide `KeyboardInterrupt`. And if you use `except Exception:` instead, then that only catches exceptions that indicate some error performing the operation, not special weird exceptions like `KeyboardInterrupt`.

But, here we're doing something different! We're not catching an exception and handling it. We're just trying to clean up some resources before exiting. And we want to clean up those resources regardless of why we're exiting – if it's a `KeyboardInterrupt` and the program is shutting down, well, that's fine, we still want to close things. And we aren't going to accidentally hide any exceptions, because this code always re-raises every exception it catches.

So this is one of the specific situations where you really do want to catch *everything*.

Also, as a rule of thumb when making PRs, it's good to avoid making small unrelated changes like this. Maintainers have to read your PR to decide whether to merge it. If there are unrelated changes, it makes it harder to read and figure out if everything is correct. And, if there are unrelated changes, it increases the chance that one of the changes will have some issue, and then that has to be dealt with before the other good changes can go in.

You're fine here, this is a pretty simple PR to read, so I don't mind the `pyvenv` thing or the cleaning up an unused import. But in general it's a good thing to keep in mind.

trio/testing/_network.py
njsmith: Yeah, it looks like `SocketListener` wasn't actually used here, so this is fine.

njsmith:
Anyway, this looks good now, so merging – thanks!

trio/_highlevel_open_unix_stream.py
kaldown: Thank you for the exceptional feedback. I thought the same at first, but didn't understand why any related to sockets code would yield anything that can't be handled by an `Exception`. After your explanation that makes sense. I remember in one of your posts, you showed an example when we can fail even after catching the exception and `closing()` a resource. So yes, it's better to do our best to be prepared.

Ok, I will try to do more atomic commits in future.

smurfix: Note that in place of a bare `except:`, an `except BaseException:` does the exact same thing (because all exceptions are required to descend from it) and documents that you have thought about this issue and yes you really want to catch everything thank you very much.

PR #1517, opened 2020-05-13
Format MultiError without sys.excepthook

### LINKED ISSUE #1065: trio's excepthook competes with apport's, which is installed by default on Ubuntu
**Python version**: 3.6.7
**OS**: Elementary OS Juno (based on ubuntu 18.04)
**trio installation**: `pip3 install --user trio`

**Error**: RuntimeWarning when i import trio 

example:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import trio
```
$ ./test.py
```
/home/user/.local/lib/python3.6/site-packages/trio/_core/_multierror.py:450: 
RuntimeWarning: You seem to already have a custom sys.excepthook handler installed.
I'll skip installing trio's custom handler, but this means MultiErrors will not show full tracebacks.
category=RuntimeWarning
```

rotu:
I think this may be a better approach to MultiError. It doesn't involve clobbering `sys.excepthook`.

Initial thoughts?

Fixes #1065

### DIFF
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -1,7 +1,6 @@
 import sys
 import traceback
 import textwrap
-import warnings
 
 import attr
 
@@ -153,6 +152,21 @@ class MultiErrorCatcher:
                 value.__context__ = old_context
 
 
+class MultiErrorCause(BaseException):
+    def __init__(self, exceptions):
+        self.embedded = list(traceback.TracebackException.from_exception(e)
+            for e in exceptions)
+
+    def __str__(self):
+        def lines():
+            for i, exc in enumerate(self.embedded):
+                yield "\nDetails of embedded exception {}:\n\n".format(i + 1)
+                yield from (
+                    textwrap.indent(line, " " * 2) for line in exc.format()
+                )
+        return '\n'.join(lines())
+
+
 class MultiError(BaseException):
     """An exception that contains other exceptions; also known as an
     "inception".
@@ -183,6 +197,7 @@ class MultiError(BaseException):
             assert len(exceptions) == 1 and exceptions[0] is self
             return
         self.exceptions = exceptions
+        self.__cause__ = MultiErrorCause(self.exceptions)
 
     def __new__(cls, exceptions):
         exceptions = list(exceptions)
@@ -355,47 +370,3 @@ def concat_tb(head, tail):
         current_head = copy_tb(head_tb, tb_next=current_head)
     return current_head
 
-
-################################################################
-# MultiError traceback formatting
-#
-# What follows is terrible, terrible monkey patching of
-# traceback.TracebackException to add support for handling
-# MultiErrors
-################################################################
-
-original_TracebackException_new = traceback.TracebackException.__new__
-
-
-class MultiTracebackException(traceback.TracebackException):
-    def __new__(cls, *args, **kwargs):
-        return original_TracebackException_new(*args, **kwargs)
-
-    def __init__(self, exc_type, exc_value, exc_traceback, **kwargs):
-        self.embedded = tuple(traceback.TracebackException.from_exception(e,
-            **kwargs) for e in exc_value.exceptions)
-        super().__init__(exc_type, exc_value, exc_traceback, **kwargs)
-
-    def format(self, *, chain=True):
-        yield from super().format(self, chain=chain)
-
-        for i, exc in enumerate(self.embedded):
-            yield "\nDetails of embedded exception {}:\n\n".format(i + 1)
-            yield from (
-                textwrap.indent(line, " " * 2) for line in exc.format(chain=chain)
-            )
-
-
-def new_traceback_exception_new(cls, exc_type, *args, **kwargs):
-    print('b0', file=sys.stderr)
-
-    if exc_type and issubclass(exc_type, MultiError):
-        print('b1', file=sys.stderr)
-        return object.__new__(MultiTracebackException)
-    else:
-        print('b2', file=sys.stderr)
-        return object.__new__(traceback.TracebackException)
-
-
-
-traceback.TracebackException.__new__ = new_traceback_exception_new
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -364,110 +364,35 @@ def concat_tb(head, tail):
 # MultiErrors
 ################################################################
 
-traceback_exception_original_init = traceback.TracebackException.__init__
 
+class MultiTracebackException:
+    def __init__(self, str_, traceback_exceptions):
+        self.embedded = tuple(traceback_exceptions)
+        self.str_ = str_
 
-def traceback_exception_init(
-    self,
-    exc_type,
-    exc_value,
-    exc_traceback,
-    *,
-    limit=None,
-    lookup_lines=True,
-    capture_locals=False,
-    _seen=None
-):
-    if _seen is None:
-        _seen = set()
+    @classmethod
+    def from_multierror(cls, exc:MultiError, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


rotu:
Obviously a little more massaging is necessary - please let me know if I should psend the time. Here's the after error message from 'test_simple_excepthook'. I clearly need to remove the `trio._core._multierror.MultiErrorCause:` if possible and adjust the tests to match that causes are now in a different order (before instead of after the effect):

```
trio._core._multierror.MultiErrorCause: 
Details of embedded exception 1:


    File "/Users/dan/Documents/trio/trio/_core/tests/test_multierror_scripts/simple_excepthook.py", line 8, in exc1_fn
      raise ValueError

  ValueError


Details of embedded exception 2:


    File "/Users/dan/Documents/trio/trio/_core/tests/test_multierror_scripts/simple_excepthook.py", line 15, in exc2_fn
      raise KeyError

  KeyError


  File "/Users/dan/Documents/trio/trio/_core/tests/test_multierror_scripts/simple_excepthook.py", line 21, in <module>
    raise trio.MultiError([exc1_fn(), exc2_fn()])
trio.MultiError: ValueError(), KeyError()
`

**Comment by rotu** [comment]
@njsmith is this viable?

**Comment by rotu** [comment]
@njsmith is this viable (assuming I fix all the tests)? Or is there some deep reason it wasn’t done this way in the first place?

**Comment by oremanj** [comment]
Unfortunately I'm not sure this approach will be workable (though I appreciate you taking the time to prototype it!). Using `__cause__` for this prevents a MultiError from having a different cause or context displayed. (The same exception can have both a cause and a context, but if both are set then by default only the cause will be shown.) Also, semantically the MultiError is intended to *represent* multiple exceptions at once, rather than being "caused by" the combination of multiple exceptions at once. See #611 for much more discussion on this.

I think the best solution for #1065 is to directly detect and disable the apport hook.

**Comment by rotu** [comment]
Hmmm... I thought cause and context were *not* mutually exclusive. I can certainly see an example where a MultiError has a context, but not where it should have a cause that’s not in the list of exceptions it rounds up.

As for the multierror being a conjunction of errors versus *caused by* a conjunction of errors, I’m on the fence. Certainly there *is* a subtle change of semantics, but when errors X and Y fail in a nursery, I think it *does* make sense to say failures of X and Y mutually caused a single failure I’m the nursery.

I’ll meditate on this and read over #611 a few more times

**Comment by oremanj** [comment]
They’re not mutually exclusive (you can set both attributes) but they sort of are (if you set a cause, the context won’t be *shown* when printing the traceback).

Another issue with using the cause to represent the source exceptions: the cause is shown before the main exception, but the tracebacks look nicer if the MultiError’s constituent exceptions are shown after the main exception. That way you can read the tree straightforwardly from top to bottom, following whichever branch of the MultiError you’re interested in.

**Comment by njsmith** [comment]
Sorry for being slow to reply... I've been trying to figure out what I wanted to say. This is tricky!

I'm not certain, but I'm leaning towards rejecting this. Here's my reasoning:

I feel like this works better in some ways (avoiding monkeypatching) but worse in others (messes with `__cause__`). Either way we'll be in a place where `MultiError`... mostly kinda works, but is clunky, and will ultimately have to be replaced by a better system (probably built into the interpreter).

The one unambiguous issue this is fixing is #1065, but that's also pretty easy to fix with the current system (see #1528), and that's a much less disruptive change.

So, given that this isn't an obvious clear-cut win, I think it's probably better to leave it be, and put that energy towards the long-term fix instead.

Does that sound right to you? Let me know what you think.

**Comment by rotu** [comment]
@oremanj

> They’re not mutually exclusive (you can set both attributes) but they sort of are (if you set a cause, the context won’t be _shown_ when printing the traceback).

Wow! You're right and that seems like a huge problem for this PR! It seems a built-in feature of [PEP 3134](https://www.python.org/dev/peps/pep-3134/), with the underlying assumption that the exception chain is linear: "The chain of exceptions is traversed by following the `__cause__` and `__context__` attributes, with `__cause__` taking priority."

> Another issue with using the cause to represent the source exceptions: the cause is shown before the main exception, but the tracebacks look nicer if the MultiError’s constituent exceptions are shown after the main exception. That way you can read the tree straightforwardly from top to bottom, following whichever branch of the MultiError you’re interested in.

I thought of that, and I don't think it's undesirable. The order of a stack and of intermediate errors in Python is deliberate (as per PEP 3134 "In keeping with the chronological order of tracebacks, the most recently raised exception is displayed last"). `MultiError` breaks this convention.

@njsmith

> Sorry for being slow to reply... I've been trying to figure out what I wanted to say. This is tricky!

No worries. It is indeed tricky!!!

> I'm not certain, but I'm leaning towards rejecting this. Here's my reasoning:
> 
> I feel like this works better in some ways (avoiding monkeypatching) but worse in others (messes with `__cause__`). Either way we'll be in a place where `MultiError`... mostly kinda works, but is clunky, and will ultimately have to be replaced by a better system (probably built into the interpreter).

I wouldn't call it "messing with `__cause__`" rather than "using `__cause__`". The proposal in #611 seems to me a reimplementation (that essentially decays to `__cause__` when there's one exception).

But yeah, as @oremanj points out, the assumptions in the existing framework (linear error chain, either explicit or implicit chaining, i.e. `__cause__` vs. `__context__`) show that this is not really compatible with the error chain abstraction.

> The one unambiguous issue this is fixing is #1065, but that's also pretty easy to fix with the current system (see #1528), and that's a much less disruptive change.

I'm *really* glad to see #1528! This PR was born of the frustration with #1065.

> So, given that this isn't an obvious clear-cut win, I think it's probably better to leave it be, and put that energy towards the long-term fix instead.
> 
> Does that sound right to you? Let me know what you think.

Basically it comes down to: is this PR a stepping stone *towards* a better long-term fix of `MultiError` or a change that makes it harder to evolve `MultiError` in the future? It seems this is the latter, so closing.

Maaaybe it's possible to achieve the same aims by modifying the `traceback` object, which is something I hope to explore.

**Comment by pquentin** [comment]
> I'm _really_ glad to see #1528! This PR was born of the frustration with #1065.

#1528 is now on PyPI as part of 0.15.0, which also removes Python 3.5 support: https://pypi.org/project/trio/0.15.0/

@rotu Is there anything else you'd like to work on? We have lots of other things to fix :)

**Comment by rotu** [comment]
@pquentin unfortunately I can’t hack on trio much right now. The warning about error handlers became high priority, as it was displayed every time the user would start our robot utility program on Ubuntu https://pypi.org/project/openrover/

I’m sure I can dig into other issues if/when they get in my way or when I have the bandwidth

**Comment by oremanj** [comment]
> I thought of that, and I don't think it's undesirable. The order of a stack and of intermediate errors in Python is deliberate (as per PEP 3134 "In keeping with the chronological order of tracebacks, the most recently raised exception is displayed last"). MultiError breaks this convention.

If you think about the individual exceptions as the cause of the MultiError, then you're right. But Trio generally doesn't do that - it thinks of the MultiError as simply a *bundle* of exceptions, which did some propagating on their own and then some more propagating together. The traceback frames on the MultiError correspond to locations further up the call tree than the traceback frames on the individual exceptions, so they should be further up the traceback printout.

Another way to think about it: In all of the following examples, `outer()` calls `inner()`, so the Python convention is for the `inner` traceback frame to be lower on the stack.

```
def outer():
    inner()
def inner():
    raise RuntimeError
```

```
async def outer():
        nursery.start_soon(inner)
async def inner():
    raise RuntimeError
```

```
async def outer():
        nursery.start_soon(inner)
        nursery.start_soon(inner)
async def inner():
    raise RuntimeError
```

But in your PR, the last of these would have the `inner` frames above the `outer` frame.

PR #1524, opened 2020-05-16
Automatically run towncrier on RTD for in-development versions

njsmith:
This has several advantages:

- On RTD, folks using the 'latest' tag to look at the git version of
  the docs will be able to see what's changed in git.

- On pull requests, newsfragments will be included in the RTD CI
  build, so we have the option of previewing them in rendered form
  before merging.

This commit also configures RTD sphinx to fail on warnings, in hopes
that this will make RTD CI work as a replacement for our current doc
CI job.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -9,3 +9,6 @@ python:
   version: 3.7
   install:
     - requirements: docs-requirements.txt
+
+sphinx:
+  fail_on_warning: true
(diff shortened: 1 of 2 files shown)


njsmith:
@pquentin The RTD team said they're looking for feedback since it's still a beta feature, so I forwarded them your comment :-)

Issue #1209, opened 2019-09-13
What should run_process do if it's cancelled but it can't kill the child process?


dj-foxxy:
The async context manager tries to `SIGKILL` the subprocess, which causes a `PermissionError` if that subprocess is `sudo`
```python
import trio


async def main():
    with trio.move_on_after(1):
        await trio.run_process(['sleep', '2'])
    print('Cancelled 1')

    with trio.move_on_after(1):
        await trio.run_process(['sudo', 'sleep', '2'])
    print('Cancelled 2')


trio.run(main)
```

njsmith:
So the problem here is that the `sudo` process runs as root, so a regular user doesn't have the necessary permissions to kill it. As far as I know, there's nothing we can do about the underlying problem – we just can't kill it. (As a user you can generally kill it by hitting control-C, but to take advantage of that we'd have to allocate a pseudo-tty and set that as the subprocess's controlling terminal... and `sudo` tries to read its password from its controlling terminal, so if we inserted our own controlling terminal it would break the password prompt.)

So I guess `run_process`/`open_process` need to cope with the fact that their `kill` might fail, and we should decide what they should do when this happens.

Options:

- crash with a long traceback, like we do now
- just wait for the process to exit normally (which might take an arbitrarily long time)
- abandon the process, letting it continue running in the background

None of these are super appealing, but I think they're the only options we have? Anyone have any thoughts on why one would be better than another?

@dj-foxxy What would you expect to happen, given that what you'd *actually* expect (the process being killed) isn't possible?

njsmith:
In case anyone's curious, here's what the traceback looks like currently:

<details>

```
Traceback (most recent call last):
  File "/home/njs/.user-python3.7/lib/python3.7/site-packages/trio/_subprocess.py", line 489, in run_process
    await proc.wait()
  File "/home/njs/.user-python3.7/lib/python3.7/site-packages/trio/_core/_run.py", line 725, in __aexit__
    raise combined_error_from_nursery
trio.MultiError: Cancelled(), Cancelled()

Details of embedded exception 1:

  Traceback (most recent call last):
    File "/home/njs/.user-python3.7/lib/python3.7/site-packages/trio/_subprocess.py", line 489, in run_process
      await proc.wait()
    File "/home/njs/.user-python3.7/lib/python3.7/site-packages/trio/_subprocess.py", line 217, in wait
      await wait_child_exiting(self)
[...]

PR #1525, opened 2020-05-16
Allow customization of run_process cancellation behavior

njsmith:
And also tweak the default to something a bit nicer on Unix.

Fixes: #1209

### DIFF
--- a/trio/_subprocess.py
+++ b/trio/_subprocess.py
@@ -3,6 +3,7 @@ import subprocess
 import sys
 from typing import Optional
 from functools import partial
+import warnings
 
 from ._abc import AsyncResource, SendStream, ReceiveStream
 from ._highlevel_generic import StapledStream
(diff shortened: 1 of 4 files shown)


trio/_subprocess.py
```diff
@@ -377,13 +378,42 @@ async def open_process(
...
+
+
+async def _posix_deliver_cancel(p):
+    try:
+        p.terminate()
+        await trio.sleep(5)
+        warnings.warn(
+            RuntimeWarning(
+                f"process {p!r} ignored SIGTERM for 5 seconds. "
+                f"(Maybe you should custom deliver_cancel?) Trying SIGKILL."
```
oremanj, line 397: The grammar nitpicker in me wants this to be "customize deliver_cancel" or "specify a custom deliver_cancel".

trio/_subprocess.py
```diff
@@ -542,7 +614,20 @@ async def read_output(stream, chunks):
...
-            await proc.wait()
+            try:
+                await proc.wait()
+            except trio.Cancelled:
+                with trio.CancelScope(shield=True):
+                    killer_cscope = trio.CancelScope(shield=True)
+
+                    async def killer():
+                        with killer_cscope:
+                            await deliver_cancel(proc)
```
oremanj, line 625: What should happen if deliver_cancel raises an exception? In the current implementation, the shielded cancel scope will prevent it from propagating until the process exits, which might take a while if the crash occurred before signaling the process in any way. Maybe on exception from a user-specified deliver_cancel we should call the default deliver_cancel to kill the process? Or just kill() since we don't care much about polish at that point, only about letting the user know what they need to fix.

trio/tests/test_subprocess.py
```diff
@@ -425,3 +426,56 @@ def on_alarm(sig, frame):
...
+
+    with pytest.warns(RuntimeWarning, match=".*whoops.*"):
+        async with _core.open_nursery() as nursery:
+            nursery.start_soon(run_process, SLEEP(9999))
+            await wait_all_tasks_blocked()
+            nursery.cancel_scope.cancel()
+
+
+@pytest.mark.skipif(os.name != "posix", reason="posix only")
+async def test_warn_on_cancel_SIGKILL_escalation(autojump_clock, monkeypatch):
```
oremanj, line 474: This should be `@slow` since it'll wait 5 seconds

pquentin: It actually runs instantly, probably thanks to `autojump_clock`?

oremanj: D'oh, right you are!

trio/_subprocess.py
njsmith: Whoops, that was an editing fail, yeah. #1531

trio/_subprocess.py
njsmith: It isn't immediately obvious to me what the right answer is here, so splitting it off into #1532

PR #1527, opened 2020-05-16
[#1526]: Preventing the problem of catching an empty list of signals

### LINKED ISSUE #1526: trio.open_signal_receiver should give an error if no arguments are passed
User confusion spotted in the wild: https://gitter.im/python-trio/general?at=5ebfa9d613878c30b581b9fe

kaldown:
closes #1526

### DIFF
--- a/trio/_signals.py
+++ b/trio/_signals.py
@@ -129,6 +129,8 @@ def open_signal_receiver(*signals):
       signals: the signals to listen for.
 
     Raises:
+      TypeError: if no signals were provided.
+
       RuntimeError: if you try to use this anywhere except Python's main
           thread. (This is a Python limitation.)
 
@@ -144,6 +146,9 @@ def open_signal_receiver(*signals):
                  reload_configuration()
 
     """
+    if not signals:
+        raise TypeError("No signals were provided")
+
     if not is_main_thread():
         raise RuntimeError(
             "Sorry, open_signal_receiver is only possible when running in "
--- a/trio/tests/test_signals.py
+++ b/trio/tests/test_signals.py
@@ -41,6 +41,12 @@ async def test_open_signal_receiver_restore_handler_after_one_bad_signal():
     assert signal.getsignal(signal.SIGILL) is orig
 
 
+async def test_open_signal_receiver_empty_fail():
+    with pytest.raises(TypeError, match="No signals were provided"):
+        with open_signal_receiver():
+            pass
+
+
 async def test_open_signal_receiver_restore_handler_after_duplicate_signal():
     orig = signal.getsignal(signal.SIGILL)
     with open_signal_receiver(signal.SIGILL, signal.SIGILL):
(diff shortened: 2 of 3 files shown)


trio/tests/test_signals.py
```diff
@@ -41,6 +41,12 @@ async def test_open_signal_receiver_restore_handler_after_one_bad_signal():
     assert signal.getsignal(signal.SIGILL) is orig
 
 
+async def test_open_signal_receiver_empty_fail():
+    with pytest.raises(AttributeError):
```
pquentin, line 45: You can pass `match=...` to `pytest.raises` to make sure you get the expected message. It's not mandatory or anything though

trio/_signals.py
```diff
@@ -144,6 +144,9 @@ def open_signal_receiver(*signals):
                  reload_configuration()
 
     """
+    if not signals:
+        raise AttributeError("No signals were provided")
```
pquentin, line 148: AttributeError is reserved to attributes of objects, I think ValueError is more appropriate here.

njsmith: Or maybe `TypeError`? That's what you get if you try to call a function with too few mandatory arguments:

```
In [1]: def f(x): 
   ...:     pass 
   ...:                                                                                   

In [2]: f()                                                                               
TypeError: f() missing 1 required positional argument: 'x'
```

We should also add a newsfragment.

kaldown: What's that 'newsfragment' ?

kaldown: changed to `TypeError`

pquentin: See https://trio.readthedocs.io/en/stable/contributing.html#pull-request-release-notes for details

kaldown:
Finally, I won that `*.rst`.Tthis `__` at the end of the url, looks weird, but helpful.

newsfragments/1526.feature.rst
```diff
@@ -0,0 +1,2 @@
+Some users are confused with the usage of `open_signal_receiver <https://trio.readthedocs.io/en/stable/reference-io.html#trio.open_signal_receiver>`__ being able to operate without any signal.
```
oremanj, line 1: Just say
```
`open_signal_receiver`
```
and Sphinx will resolve the link for you automatically.

newsfragments/1526.feature.rst
```diff
@@ -0,0 +1,2 @@
...
+This issue being fixed.
```
oremanj, line 2: The wording you have here is not clear about what changed, which is the main point of a newsfragment. I would say something like: Calling `open_signal_receiver` with no arguments used to succeed without listening for any signals. This was confusing, so now it raises `ValueError` instead.

Also this isn't really a "feature" - I would probably put this in the "bug fixes" section of the release notes (so name the newsfragment "1526.bugfix.rst").

oremanj:
Also, please write "Closes #1526" in the PR description so that merging the PR will close the issue. Just listing the issue number isn't sufficient for that.

newsfragments/1526.feature.rst
kaldown: I followed this example https://raw.githubusercontent.com/python-trio/trio/master/newsfragments/1241.feature.rst

pquentin: (It also makes sure the function actually exists, which is handy to avoid dead links.)

kaldown:
Thank you for the patient :)

PR #1528, opened 2020-05-17
Make Trio's excepthook play nicely with Ubuntu's excepthook

njsmith:
Fixes gh-1065

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -407,6 +407,11 @@ else
     # Actual tests
     python -m pip install -r test-requirements.txt
 
+    # So we can run the test for our apport/excepthook interaction working
+    if [ -e /etc/lsb-release ] && grep -q Ubuntu /etc/lsb-release; then
+        sudo apt install -q python3-apport
+    fi
+
     # If we're testing with a LSP installed, then it might break network
     # stuff, so wait until after we've finished setting everything else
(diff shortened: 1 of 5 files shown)


pquentin:
I failed to reproduce the issue in Docker using Ubuntu and trio from PyPI. I think you're using Ubuntu locally, is this reproducing for you?

I also tried not installing python-apport in ci.sh and the tests still passed: see https://github.com/pquentin/trio/commit/ebcde15eb7ac58482ec0ce9123e82f3c730cecae and https://github.com/pquentin/trio/actions/runs/107173872

pquentin:
Hmm I managed to reproduce it, looks like it only happens outside of virtual environments. But in ci.sh we use a virtual environment?

Maybe none of this matters, I don't know.

njsmith:
Yeah, the original issue only affects the system python, so `sudo pip install trio` or `pip install --user trio`, with no venv, on Ubuntu, with python3-apport installed.

The test is marked to be skipped if you don't have Ubuntu + python3-apport, and then inside the test it does a gross hack to make a venv python temporarily act like a system python. The change to ci.sh to install python3-apport is technically redundant because it turns out the images already have it installed, but I guess it's still good for documentation and robustness?

PR #1535, opened 2020-05-19
Simplify releasing documentation

pquentin:
Following @njsmith's procedure here: https://gitter.im/python-trio/general?at=5ec36d49940fa238d6fffa33

### DIFF
--- a/docs/source/releasing.rst
+++ b/docs/source/releasing.rst
@@ -19,7 +19,7 @@ Things to do for releasing:
 
 * Do the actual release changeset
 
-   + update version number
+   + bump version number
 
       - increment as per Semantic Versioning rules
 
@@ -37,31 +37,18 @@ Things to do for releasing:
 
 * create pull request to ``python-trio/trio``'s "master" branch
 
-* announce PR on gitter
-
-   + wait for feedback
-
-   + fix problems, if any
-
 * verify that all checks succeeded
 
-* acknowledge the release PR
-
-   + or rather, somebody else should do that
-
-* tag with vVERSION
+* tag with vVERSION, push tag
 
 * push to PyPI
 
    + ``python3 setup.py sdist bdist_wheel upload``
 
-* announce on gitter
-
-* update version number
+* update version number in the same pull request
 
    + add ``+dev`` tag to the end
 
-* prepare another pull request to "master"
+* approve and merge the release pull request
 
-   + acknowledge it
-
+* announce on gitter
--- a/docs/source/releasing.rst
+++ b/docs/source/releasing.rst
@@ -41,14 +41,16 @@ Things to do for releasing:
 
 * tag with vVERSION, push tag
 
-* push to PyPI
+* push to PyPI::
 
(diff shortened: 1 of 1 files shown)


oremanj:
I think it would be a good idea for someone to approve the release PR before tagging and uploading to PyPI. It's our last chance to catch silly newsfragment typos and such. I don't feel strongly though, if others think this is too much process.

smurfix:
In the past we uploaded *after* merging the PR. That's not more process, let alone too much, but just a different (and IMHO safer) order of doing things.

njsmith:
We already review the newsfragments before they're merged in the first place, typos are rarely catastrophic, and do have the option of retroactively fixing the changelog in future releases if we really want to.

I don't have anything against process per se, but I feel like it should have a clear motivation and guidelines for what to do, and in this case I'm not quite sure what those are. Except that, well, releasing is scary, and more process feels safer than less process :-). But IME vague "this is scary" process tends to be counterproductive.

That said, when I'm releasing I do tend to skim over the changelogs since they're in front of me and sometimes rewrite bits. And if you do actually change them, then that's a clear thing you might want some review on!

> In the past we uploaded after merging the PR. That's not more process, let alone too much, but just a different (and IMHO safer) order of doing things.

I don't have a strong feeling either way here, but generally you're going to be doing all these within a few seconds of each other, right? Is there a realistic risk that you're going to notice some catastrophic error in between hitting merge and hitting enter on the `twine upload`?

docs/source/releasing.rst
```diff
@@ -37,31 +37,18 @@ Things to do for releasing:
...
-* acknowledge the release PR
-
-   + or rather, somebody else should do that
-
-* tag with vVERSION
+* tag with vVERSION, push tag
 
 * push to PyPI
 
    + ``python3 setup.py sdist bdist_wheel upload``
```
njsmith, line 46: I tend to do:

```
git clean -xdf   # maybe run 'git clean -xdn' first to see what it will delete
python3 setup.py sdist bdist_wheel
twine upload dist/*
```

Historically `twine` was better than `setup.py upload`, but I don't know if that's still true.

docs/source/releasing.rst
```diff
@@ -37,31 +37,18 @@ Things to do for releasing:
...
-
-* update version number
+* update version number in the same pull request
 
    + add ``+dev`` tag to the end
 
-* prepare another pull request to "master"
-
-   + acknowledge it
+* approve and merge the release pull request
```
njsmith, line 52: You can post an approving review if you want, but I don't know if it actually does anything :-). We don't have github set to enforce approvals before merge. (Actually, now that I think about it, I think you may not even be able to post an approval on your own PR?)

pquentin: Right, we can't approve our own PRs, but we can merge them, fixed

docs/source/releasing.rst
pquentin: This is what I use too! Updated

pquentin:
I don't have a strong opinion either way 🤷 

But I updated the review comments, please take another look

Issue #1521, opened 2020-05-14
Idea: service nurseries


njsmith:
Extracting this idea from #147, because that's a monster thread and this seems worth discussing separately.

Idea: a "service nursery" is one where the background tasks are "services" that don't do anything on their own, but provide some service to the main task. Therefore:

- background tasks are automatically cancelled when the nursery block exits
- before the nursery block exits, background tasks are automatically shielded from cancellation (and maybe `KeyboardInterrupt`?)
- mayyybe there's some special handling of exception propagation? if in the future we switch to always wrapping nursery exceptions in a multi-error-equivalent (#611), AND multi-errors remain awkward and folks want to avoid them, then we could potentially do that with a service nursery by deeming that background tasks should never fail, so if they do then you get a special `BackgroundServiceFailed` error or something, and that way the main task body exception doesn't need to be wrapped in a multi-error? Dunno, this seems dubious to me.

Examples that fit this pattern:

- Trio's "system" nursery
- The background nursery that trio-websocket creates for each connection, or any other protocol that requires some background task to drive the socket
- The `nursery` fixture in pytest-trio
- Independently suggested by @matham at https://gitter.im/python-trio/general?at=5ebceeaf2cf0da0ad90004d4

Straw man implementation: https://github.com/python-trio/trio/issues/147#issuecomment-455824598

Questions:

- Are those semantics right?
- Should the special "service-ness" flag be attached to the nursery, or to individual tasks?
- `open_nursery(service=True)` or `open_service_nursery()`?
- control-C handling? (see also #733)

oremanj:
This currently exists (without the "auto cancel when nested child is done" semantics) as `tricycle.open_service_nursery()` (implementation is effectively unchanged from the linked strawman one). I use it all the time. I'd be fine adding the "auto cancel when nested child is done" and "wrap background service exceptions so that service nurseries don't MultiError-ize and thus can be wrapped in abstractions without impacting exception handling" features; I think they naturally complement the typical use cases. 

Shielding the "service" tasks from KI makes sense to me, and mirrors what we currently do for system tasks (which would become tasks in just another service nursery). It feels consistent to me with the fact that they're shielded from cancellation. It does mean that an infinite loop in a service task is harder to debug, but we can compensate for that with better deadlock debugging tools. And this is a mildly advanced-level feature anyway.

> Should the special "service-ness" flag be attached to the nursery, or to individual tasks?

[...]

PR #1537, opened 2020-05-19
[WIP] Inject deferred KeyboardInterrupt in a new task, instead of waking up the main task to deliver it

oremanj:
Uploading now to fix the issue number for the deprecation message. Still todo: newsfragment, update KI tests, update docs, post on #1.

### DIFF
--- a/trio/_core/_exceptions.py
+++ b/trio/_core/_exceptions.py
@@ -1,6 +1,9 @@
+# coding: utf-8
+
 import attr
 
 from trio._util import NoPublicConstructor
+from trio import _deprecate
 
 
 class TrioInternalError(Exception):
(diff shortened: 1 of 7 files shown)


oremanj:
The coverage "failure" is because `trio._core._run` now contains fewer statements, so the same set of uncovered statements results in a lower coverage percentage.

pquentin:
Sorry for the merge conflict, I just merged the Black pull request

njsmith:
I'm not sure cancelling the system nursery is a great idea... until now, we had an invariant that system tasks would only be cancelled (a) after the main task exited, or (b) after a system task crashed. So they basically had the "service task" semantics discussed in #1521.

I feel like `KeyboardInterrupt` should immediately cancel the main task, but *not* immediately cancel system tasks.

oremanj:
That makes sense to me, but I think it will be easiest to implement by just doing #1521, because I don't see any way to wrap the main task in a dedicated cancel scope without adding another frame to every Trio traceback.

njsmith:
Maybe possibly we could do something like this?

```python
    async def init(self, async_fn, args):
        async with open_nursery() as system_nursery:
            self.system_nursery = system_nursery
            async with open_nursery() as main_task_nursery:   ##### <-----
                try:
                    self.main_task = self.spawn_impl(async_fn, args, main_task_nursery, None)
                except BaseException as exc:
                    self.main_task_outcome = Error(exc)
                    system_nursery.cancel_scope.cancel()
                self.entry_queue.spawn()
```

but yeah, I see your point

A5rocks:
> That makes sense to me, but I think it will be easiest to implement by just doing #1521, because I don't see any way to wrap the main task in a dedicated cancel scope without adding another frame to every Trio traceback.

@oremanj Well now we have `main_task_nursery`! So this can use that.

Also you will probably have to start the PR from scratch as it is out of date.

PR #1540, opened 2020-05-22
Switch to black

pquentin:
Trying again after https://github.com/python-trio/trio/pull/768 in 2018. Why now? black is much more popular, and Trio supports the same versions of Python as black. (It does not support PyPy, though.)

The benefits:

 * we're using a tool more contributors are familiar with
 * all editors have useful black integrations, unlike yapf
 * we won't have to be worried about yapf migration from lib2to3 which is deprecated
 * the output is less likely to change between versions
 * the output is generally more useful
 * we stay at 80 characters, so @njsmith can keep his stone age configuration :)

The downsides:

 * all pull request will get conflicts!
 * git blame will be less useful


trio/_core/tests/test_ki.py
```diff
@@ -537,9 +537,11 @@ def test_ki_wakes_us_up():
     #   https://bugs.python.org/issue31119
     #   https://bitbucket.org/pypy/pypy/issues/2623
     import platform
-    buggy_wakeup_fd = (
-        platform.python_implementation() == "CPython" and sys.version_info <
-        (3, 6, 2)
+
+    buggy_wakeup_fd = platform.python_implementation() == "CPython" and sys.version_info < (
```
oremanj, line 541: ugh, this is an ugly one. Doesn't really matter though. It might do a better job if you put the clauses in the opposite order, and surround with parens?

trio/_socket.py
```diff
@@ -338,10 +341,10 @@ def _make_simple_sock_method_wrapper(methname, wait_fn, maybe_avail=False):
     async def wrapper(self, *args, **kwargs):
         return await self._nonblocking_helper(fn, args, kwargs, wait_fn)
 
-    wrapper.__doc__ = (
-        """Like :meth:`socket.socket.{}`, but async.
+    wrapper.__doc__ = """Like :meth:`socket.socket.{}`, but async.
```
oremanj, line 344: Maybe make this into an f-string so it doesn't format so poorly?

trio/tests/test_util.py
```diff
@@ -53,11 +58,13 @@ async def wait_with_ul1():
 
 def test_module_metadata_is_fixed_up():
     import trio
+
     assert trio.Cancelled.__module__ == "trio"
     assert trio.open_nursery.__module__ == "trio"
     assert trio.abc.Stream.__module__ == "trio.abc"
     assert trio.lowlevel.wait_task_rescheduled.__module__ == "trio.lowlevel"
     import trio.testing
```
oremanj, line 66: With the extra blank lines, I think it looks better to put both imports next to each other.

oremanj:
Oh, and I believe you can fix the CI issues by only installing black on CPython. (It depends on typed-ast, which doesn't compile under PyPy.)

njsmith:
I actually did update my editor and terminal to be 90 columns...

pquentin:
@njsmith I would like to apologize for using the term "stone age", in my defense I just reused your words from #768.

Should we use 88 characters, then? I guess sticking to black defaults is better when possible. And more generally, are you in favor of this or do you still have concerns?

oremanj:
I'm neutral on whether we switch to 88 or stay at 80. (I also use an approx width=88 editor setup.) But I think if we plan to reflow we should do it now, to consolidate all the disruptive changes in one place.

pquentin:
> But I think if we plan to reflow we should do it now, to consolidate all the disruptive changes in one place.

Agreed. I'm happy to wait for consensus: it's easy to rebase this as the actual black changes are isolated in their own commit.

njsmith:
Oh, I wasn't bothered, just terse b/c I was replying from my phone while waiting for my laptop to upgrade to Ubuntu 20.04 :-).

I'm in favor of switching to black, and I guess mildly in favor of sticking to the defaults instead of fighting them.

pquentin:
@njsmith @oremanj Can you please take another look? I now use the default 88 characters. I had to rebase, in order to avoid two reformatting commmits, sorry about that!

trio/_socket.py
```diff
@@ -338,15 +339,13 @@ def _make_simple_sock_method_wrapper(methname, wait_fn, maybe_avail=False):
...
 
-            """.format(methname)
-    )
+            """
     if maybe_avail:
         wrapper.__doc__ += (
-            "Only available on platforms where :meth:`socket.socket.{}` "
-            "is available.".format(methname)
+            f"Only available on platforms where :meth:`socket.socket.{methname}`"
+            " is available."
```
oremanj, line 348: Can this space move to the end of the previous line now that we have more room?

trio/_socket.py
```diff
@@ -497,8 +497,7 @@ async def _resolve_address(self, address, flags):
         elif self._sock.family == _stdlib_socket.AF_INET6:
             if not isinstance(address, tuple) or not 2 <= len(address) <= 4:
                 raise ValueError(
-                    "address should be a (host, port, [flowinfo, [scopeid]]) "
-                    "tuple"
+                    "address should be a (host, port, [flowinfo, [scopeid]]) " "tuple"
```
oremanj, line 500: Merge strings

trio/_sync.py
```diff
@@ -321,8 +317,7 @@ def release_on_behalf_of(self, borrower):
         """
         if borrower not in self._borrowers:
             raise RuntimeError(
-                "this borrower isn't holding any of this CapacityLimiter's "
-                "tokens"
+                "this borrower isn't holding any of this CapacityLimiter's " "tokens"
```
oremanj, line 320: Merge strings

trio/_core/tests/test_ki.py
```diff
@@ -537,9 +535,9 @@ def test_ki_wakes_us_up():
     #   https://bugs.python.org/issue31119
     #   https://bitbucket.org/pypy/pypy/issues/2623
     import platform
+
     buggy_wakeup_fd = (
-        platform.python_implementation() == "CPython" and sys.version_info <
-        (3, 6, 2)
+        sys.version_info < (3, 6, 2) and platform.python_implementation() == "CPython"
```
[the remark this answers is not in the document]
pquentin: Unfortunately, with the other order Black produces the original ugly code with each number on its own line.

Issue #6, opened 2017-01-22
Design: thread pools


njsmith:
Right now, `run_in_worker_thread` just always spawns a new thread for the operation, and then kills it after. This might sound ridiculous, but it's not so obviously wrong as it looks! There's a [large comment](https://github.com/njsmith/trio/blob/6af8e23bf1b522aa2e7aedec8d0bd2fb2d2e4ffb/trio/_threads.py#L42-L133) in `trio._threads` talking about some of the issues.

Questions:

* Should there be *any* global control over worker thread spawning?
* If there is, should it be a strict limit on the number of threads, or something more subtle like a limiter on the rate at which they spawn?
* How do administrators configure this stuff? Or instrument it?
* What should the API to interact with it look like, e.g. do we need a way for a specific `run_in_worker_thread` to say that it shouldn't block waiting for a thread because it might unblock a thread?

Prior art: https://twistedmatrix.com/trac/ticket/5298
> Interacting with the products at Rackspace which use Twisted, I've seen problems caused by thread-pool maximum sizes with some annoying regularity. The basic problem is this: if you have a hard limit on the number of threads, *it is not possible to write a correct program which may require starting a new thread to un-block a blocked pool thread* - glyph

njsmith:
The Java SDK docs have surprisingly good notes on these trade-offs: [java.util.concurrent.ThreadPoolExecutor](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html)

njsmith:
[Why the CLR 2.0 SP1's threadpool default max thread count was increased to 250/CPU](http://joeduffyblog.com/2007/03/04/why-the-clr-20-sp1s-threadpool-default-max-thread-count-was-increased-to-250cpu/)

[Optimizing Concurrency Levels in the .NET ThreadPool: A Case Study of Controller Design and Implementation](http://controlofsystems.org/febid2008/1569117950.pdf)

[Applying Control Theory in the Real World: Experience With Building a Controller for the .NET Thread Pool](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.330.3350&rep=rep1&type=pdf)

(Not sure what the difference between the latter two papers is)

njsmith:
```
<benbangert> has anyone made anything to throttle/restrict the amount of tasks to defer to thread from a single protocol instance?
<runciter> benbangert: you might be interested in https://github.com/terrycojones/txrdq
[...]

PR #1545, opened 2020-05-23
Add thread cache

njsmith:
On my Linux laptop, this makes `await trio.to_thread.run_sync(lambda: None)` about twice as fast, from ~150 µs to ~75 µs.

Closes: #6

Test program:

```python
import trio
import time

COUNT = 10000

async def main():
    while True:
        start = time.monotonic()
        for _ in range(COUNT):
            await trio.to_thread.run_sync(lambda: None)
        end = time.monotonic()
        print("{:.2f} µs/job".format((end - start) / COUNT * 1e6))

trio.run(main)
```


njsmith:
Huh, failed test was a genuine, unrelated bug – see #1546

njsmith:
OK, maybe #1548 will help with the test failure...

I guess we're consistently seeing these test failures here, and only here, because `open_process` uses a thread, and this PR must be making such a speed difference on FreeBSD that it can trigger the race condition? It's odd.

njsmith:
Phew, tests finally passing, and I think this is ready to review.

Random question: do you think it's more intuitive to have `start_thread_soon(deliver, fn)`, or `start_thread_soon(fn, deliver)`?

oremanj:
> do you think it's more intuitive to have `start_thread_soon(deliver, fn)`, or `start_thread_soon(fn, deliver)`?

My intuition favors `(fn, deliver)` because you run `fn` first and then `deliver` ("data flows from left to right"), but I think either way is defensible.

trio/_core/_thread_cache.py
```diff
@@ -0,0 +1,168 @@
...
+                    # We successfully removed ourselves from the idle
+                    # worker queue, so no more jobs are incoming; it's safe to
+                    # exit.
+                    return
+
+
+class ThreadCache:
+    def __init__(self):
+        self._idle_workers = {}
+        self._cache_lock = Lock()
```
oremanj, line 94: You don't seem to use the `_cache_lock` anywhere. If you remove it, you could even make `THREAD_CACHE` be the dict directly, and move `ThreadCache.start_thread_soon()` into the global `start_thread_soon()`.

njsmith: Nice catch on `_cache_lock`!

We currently have a test that instantiates a private `ThreadCache` object (test_race_between_idle_exit_and_job_assignment`), and I don't see how to easily make it work otherwise, so I guess I'll leave the class there for now.

njsmith:
> My intuition favors (fn, deliver) because you run fn first and then deliver ("data flows from left to right"), but I think either way is defensible.

Yeah, having slept on it I think you're right – also `fn` is just the "more important" function from the perspective of the caller, so having it second feels weird. Swapped them.

trio/_core/_thread_cache.py
```diff
@@ -93,7 +93,7 @@ def __init__(self):
         self._idle_workers = {}
         self._cache_lock = Lock()
 
-    def start_thread_soon(self, deliver, fn):
+    def start_thread_soon(self, fn, deliver):
```
pquentin, line 95: For what it's worth, yes, this new order seems more intuitive to me, as deliver is called after fn

PR #1546, opened 2020-05-23
Fix flakiness in test_subprocess::test_basic

njsmith:
gh-1514 introduced a race condition in this test: now if the process
runs quickly enough, the returncode might have magically become
non-None before we check it.

Example failure: https://travis-ci.org/github/python-trio/trio/jobs/690291915

There are other tests that check the 'p.returncode is None' case (e.g.
test_auto_update_returncode), so we can just delete this.


njsmith:
...and according to some universal law of irony, github actions appears to have flaked out and decided not to test this PR at all. Maybe opening/closing will help?

PR #1551, opened 2020-05-25
"Guest mode", for cohabitation with Qt etc.

njsmith:
This adds `trio.lowlevel.start_guest_run`, which initiates a call to `trio.run` "on top of" some other "host" loop. It uses the regular Trio I/O backends, so all features are supported. It does *not* use
polling, so it's efficient and doesn't burn CPU. All you have to do is provide a threadsafe way to schedule a callback onto the host loop.

Since Trio code and the host loop are sharing the same main thread, it's safe for Trio code to call synchronous host loop functions, and for host code to call synchronous Trio functions.

~~This is an early draft. It has no docs or tests, and I've only tried the epoll mode.~~ This is ready for review/merge.

Fixes: #399

TODO/open questions:

- [x] Needs #1545 to be merged first

- [x] Check how much overhead this adds to regular `trio.run` mode

- [x] Double-check that it's actually reasonably efficient when combined with e.g. Qt

- [x] Tests

- [x] Docs

- [x] How on earth should lifetime management work between the two loops (maybe this is a problem for downstream libraries that use this? but it'd be nice if reasonable options were reasonable) [conclusion: we just need to document what's possible, and make a strong recommendation that when using this mode you start up the host loop, then immediately start up the trio loop, and treat the trio loop exiting as the signal to shut down the host loop]

- [x] What on earth should we do about signal handling in general? (share set_wakeup_fd with the host? use the host's run_sync_soon_threadsafe in our signal wakeup path?)

  - [x] and what about control-C in particular?

- [x] add a non-threadsafe `run_sync_soon` handler?

- [x] Should we support multiple `run` calls simultaneously in the same host loop? It has some logic, but would make accessing Trio from the host substantially more complicated... (conclusion: not in version 1)

-------

Simple demo of Trio and asyncio sharing a thread (also included as
`notes-to-self/aio-guest-test.py`):

<details>

```python
import asyncio
import trio

async def aio_main():
    loop = asyncio.get_running_loop()

    trio_done_fut = asyncio.Future()
    def trio_done_callback(main_outcome):
        print(f"trio_main finished: {main_outcome!r}")
        trio_done_fut.set_result(main_outcome)

    trio.lowlevel.start_guest_run(
        trio_main,
        run_sync_soon_threadsafe=loop.call_soon_threadsafe,
        done_callback=trio_done_callback,
    )

    (await trio_done_fut).unwrap()


async def trio_main():
    print("trio_main started!")

    to_trio, from_aio = trio.open_memory_channel(float("inf"))
    from_trio = asyncio.Queue()

    asyncio.create_task(aio_pingpong(from_trio, to_trio))

    from_trio.put_nowait(0)

    async for n in from_aio:
        print(f"trio got: {n}")
        await trio.sleep(1)
        from_trio.put_nowait(n + 1)
        if n >= 10:
            return

async def aio_pingpong(from_trio, to_trio):
    print("aio_pingpong started!")

    while True:
        n = await from_trio.get()
        print(f"aio got: {n}")
        await asyncio.sleep(1)
        to_trio.send_nowait(n + 1)


asyncio.run(aio_main())
```

Output:

```
trio_main started!
aio_pingpong started!
aio got: 0
trio got: 1
aio got: 2
trio got: 3
aio got: 4
trio got: 5
aio got: 6
trio got: 7
aio got: 8
trio got: 9
aio got: 10
trio got: 11
aio got: 12
trio_main finished: Value(None)
```

</details>


njsmith:
# Basic outline

## Changes to I/O backends

We split the old `handle_io(timeout)` into two pieces: `events = get_events(timeout); process_events(events)`. The idea is that `get_events` has the blocking part, and `process_events` has the part that touches global state, so we can shove `get_events` off into a thread.

I also had to add a `force_wakeup` method, which is largely redundant with the existing re-entry task system, but it seemed bad to depend on the re-entry task always running... maybe there's some way to merge these, I'm not sure

## Changes to core run loop

The core run loop stays the same, except that we replace:

```python
runner.io_manager.handle_io(timeout)
```

with

```python
events = yield timeout
```

So that means our core run loop is now a generator (!), and it requires some driver code to read out the `timeout`, call `io_manager.get_events(timeout)`, and then send the return value back in.

The normal `trio.run` has a trivial driver loop that just does that.

The "guest mode" driver is a bit cleverer: it arranges to push the `get_events` off into a thread, and then when it returns it schedules another iteration of our generator onto the host loop.

We keep a strict alternation: *either* the I/O thread is running looking for I/O, *or* the Trio scheduler is running (or waiting to run) on the host loop. You never have both running at once.

Originally I was imagining something a bit cleverer, where the I/O thread runs continuously with infinite timeout, and just streams back I/O events to Trio as they happen. The problem with this is that if we're not using the `io_manager` to handle timeouts, then we have to come up with some other mechanism for that. Now, we could require the host loop to provide some kind of `call_at(time, ...)` API. But it seemed to add a lot of complications: now you need some way to cancel and reschedule the `call_at` handlers, you need some way to make sure the time-triggered and I/O-triggered  and host-triggered scheduler ticks don't step on each other, the `wait_all_tasks_blocked` code gets more complicated (since it relies on knowing when a timeout has passed without any tasks being scheduled), etc.

Oh right, host-triggered scheduler ticks: we want to allow the host loop code to invoke synchronous Trio APIs, like `memory_channel.send_nowait`, `nursery.start_soon`, `cancel_scope.cancel`. These might causes tasks to be rescheduled, or global deadlines to change, *while* Trio is waiting for I/O! This is a new thing that can't happen in regular Trio.

The hack I used to make it work is: if a task becomes scheduled, or the next deadline changes, at a time when the I/O thread is running, then we force the I/O thread to wake up immediately, which triggers a scheduler tick, which will pick up whatever changes were made by the host. This is the most awkward part of this approach, and why I had to add the new `force_wakeup` method on IO managers.

(Note: I'm not 100% sure I got all the details of this state machine correct yet...)

However, even in the "cleverer" approach where timeouts are managed on the main thread, you *still* need the equivalent of a `force_wakeup` method, because you need to have some way to tell the background I/O thread to exit when shutting down!

Anyway, so this approach just seemed simpler all around, and hopefully not too inefficient. There are a handful of extra syscalls and thread transitions compared to the "fancy" approach, but AFAICT it ought to be unnoticeable in practice. Though we should confirm empirically :-)

njsmith:
On further thought, I guess signals are not a big deal. The tricky bit is `set_wakeup_fd`, but, the way we use the wakeup fd, all we need is that after a signal arrives then the Python VM should start interpretering bytecode again promptly. So there are two cases:

- The host loop has registered its own `set_wakeup_fd`. In this case we know it's a Python loop that's listening for signals and doing something with them, so we can assume that their `set_wakeup_fd` will wake up the Python VM and that's all we need.

- The host loop hasn't registered its own `set_wakeup_fd`. In this case we're free to register our own.

So all we have to do is check whether an fd is already registered, and if it is then skip registering ours.

Still no idea what to do about control-C though; I guess that's tightly tied into how people want to coordinate the lifetime of the two loops.

njsmith:
Ha, it turns out that like for everything involving async, Twisted also experimented with this kind of approach. They called it their `ThreadedSelectReactor`: https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/_threadedselect.py

Apparently they ran into some not-so-obvious problems with it – for example, downloads on Windows couldn't achieve line throughput, which was awkward since fancy download apps are one of the most obvious use cases for combining a GUI library + advanced networking library. There's also a lot different between our approach and theirs, though – different decade, use of `select` versus not, pretty different strategy for passing info back and forth between the main thread and the worker thread, etc. So it's not at all clear how much of their experience still applies. Still, good to get some warnings of what to watch out for.

Conversation with glyph:

```
<glyph> njs: ugh threadedselectreactor
<glyph> njs: it’s a testament to trio’s better factoring that the implementation is way less gross and I like the public api that’s being exposed
<njs> glyph: ha, I had no idea that threadedselectreactor existed
<njs> twisted minds think alike or something
<njs> though one of the main motivations is to completely eliminate the whole pluggable reactor concept, so I guess twisted didn't get the full benefit
<glyph> njs: oh, we did, it just sucks
<glyph> it performs terribly, and it can't take advantage of various native loop support things, and it's a bug magnet due to the inherently shared state
<glyph> not to mention it has nasty side effects, like popping up firewall access-control dialog boxes for its loopback sockets on windows
<glyph> granted, you could probably do a better job than we did
<njs> I don't think windows does that anymore... asyncio and trio both create loopback sockets unconditionally on windows
<glyph> there was definitely a brief period where we considered eliminating the pluggable reactor with such a thing but it was a definite "no" after a few years of experience with it
<glyph> perhaps I will expand in more detail in my memoir, "threads don't actually work and neither do signals"
<njs> I think I have a fair idea of what I'm getting into bug-magnet-wise, but I am curious about the terrible performance and the native loop support things you wanted to use but couldn't
<glyph> too bad tef already has the trademark on'lessons learned from a life wasted'
<njs> going in and out of a hot thread is ~10 µs on my laptop, which seems pretty acceptable for an idle GUI app (it only uses the thread when it knows it has to sleep for non-zero time)
<njs> oh, or by "native loop" do you mean "everything that's not 'select'"?
<glyph> njs: as I recall, Windows is a significant multiple of that, and there are strange thundering-herd problems with the scheduler when you start trying to benchmark straight line wire throughput that moves stuff between threads
<glyph> I know I'm talking a lot of Windows here, but "custom reactor" = "GUI", and the production environment for GUIs is either Windows 10 or iOS in the real world
<njs> yeah
<glyph> njs: my general advice here is both "maybe you will have entirely different problems, it's been a while" and also "beware microbenchmarks"
<njs> I guess I am assuming that most people do not care about whether their GUI app can handle 10k requests/second, so some throughput penalty is acceptable
<glyph> the microbenchmarks told us a REMARKABLY different story than actual whole-application profiles
<glyph> people are OK with a latency penalty on GUI apps! but your GUI app is probably a file downloader so people get real mad about throughput penalties :)
<njs> ha
<njs> so are you saying that microbenchmarks were over-optimistic?
<glyph> yes, or at least, they were
<glyph> "how long does it take to do this thing with a thread" was all pretty fast in isolation; "what does it actually look like when we shove bytes over this connection" ended up being implausibly slow based on those numbers
<njs> interesting
<njs> I'm actually having trouble figuring out where download throughput penalties would come from, if we only touch the thread after getting EWOULDBLOCK on recv
<njs> it looks like threadedselectreactor works a bit differently though
<glyph> njs: oh yeah the details are somewhat different
<glyph> njs: it's the thrashing back and forth to the thread; gigabit TCP, particularly with TLS, won't *reliably* keep your pipe full, though
<glyph> and over-the-internet transfers, you'll definitely see EWOULDBLOCK
<njs> sure
<glyph> (can you tell the application I tested this with the most did bulk data transfers a lot)
<njs> I guess the question is whether the stutter after each EWOULDBLOCK is long enough to overflow the transfer window and stall the underlying flow
<glyph> njs: again, it might just be different today, but a decade ago or whenever this was, yes, you would stall the underlying flow, and in being so stalled, you'd switch to a different thread, and that would tell the kernel scheduler that maybe a good guess is you're gonna be busy so to deprioritize you on the scheduler, which leads to smaller TCP window sizes, etc
<njs> I think the theory says that as long as going in and out of the thread has low latency compared to regular scheduling glitches and/or the latency of the underlying link, then it shouldn't affect throughput
<njs> but as we know, theory is not a very good predictor to how networks behave :-)
<njs> okay, yeah
<njs> thanks for the war stories, it's very helpful :-)
<njs> glyph: do you mind if I dump this log into the issue for reference?
<glyph> sure
```

trio/_core/_run.py
```diff
@@ -1767,7 +1862,7 @@ def run_impl(runner, async_fn, args):
             else:
                 break
 
-        if not runner.runq and idle_primed:
+        if not runner.runq and not events and idle_primed:
```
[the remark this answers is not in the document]
njsmith: Yeah...... this is so subtle that I originally couldn't figure out how to write a comment :-). I added one now though; see if it helps?

oremanj:
That comment is great, thank you!

trio/_core/_run.py
```diff
@@ -1656,28 +1750,7 @@ def run(
...
-    system_context.run(current_async_library_cvar.set, "trio")
-    runner = Runner(
-        clock=clock,
-        instruments=instruments,
-        io_manager=io_manager,
-        system_context=system_context,
-    )
-    GLOBAL_RUN_CONTEXT.runner = runner
+    runner = setup_runner(clock, instruments)
     locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
```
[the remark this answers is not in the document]
njsmith: But `run` is always on the stack in regular mode, and currently we only enable our control-C handling in regular mode :-). I'll worry about this later when we figure out what we want to do about `ki_manager` and all the associated miscellany...

trio/_core/_run.py
```diff
@@ -1118,6 +1130,62 @@ class Runner:
...
+            self.run_sync_soon_threadsafe(self.guest_tick)
+        else:
+            self.guest_tick_scheduled = False
+
+            def get_events():
+                return self.io_manager.get_events(timeout)
+
+            def deliver(events_outcome):
+                def in_main_thread():
+                    self.unrolled_run_next_send = events_outcome.unwrap()
```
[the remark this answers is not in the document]
njsmith: That would add a capture/unwrap cycle to regular mode, but maybe that doesn't matter.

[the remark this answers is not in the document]
oremanj: You're unwrapping in the driver, not inside the generator (so it will resume with `throw()` if get_events() raised an exception). Regular mode doesn't need to throw I/O exceptions into the generator, since it's just going to exit with a TrioInternalError anyway. But here the I/O exception would be raised out of the callback running in the host event loop, and the likelihood of getting it propagated properly is not good.

njsmith:
> If this turns out to be a blocker, note that you can use basically the same trick as greenback to write a "generator" in which you can yield from a callee. I think the greenlet library even includes this as one of their examples. That would probably make "guest mode" slower but imply less overhead for normal straight-through mode. (And "guest mode" would depend on greenlet.)

Interesting point. Actually, thinking about this, I don't think we'd even need greenlet... the greenlet code would look something like:

```python
timeout = ...
if runner.is_guest:
    events = greenlet_magic_yield(timeout)
else:
    events = io_manager.get_events(timeout)
io_manager.process_events(events)
...
```

...But if we decide all the yield/resumes are a problem, we could do that just as easily with a regular `yield`, since we only have one frame involved :-)

njsmith:
(Of course this would mean that the regular mode would have to do `unrolled_run(...).send(None)` instead of just `unrolled_run(...)`, but that's just odd-looking, not actually a problem.)

trio/_core/_run.py
```diff
@@ -1118,6 +1130,62 @@ class Runner:
...
+    unrolled_run_gen = attr.ib(default=None)
+    unrolled_run_next_send = attr.ib(default=None)
+    guest_tick_scheduled = attr.ib(default=False)
+
+    def guest_tick(self):
+        # XX no signal handling support at all currently
+        assert self.is_guest
+        try:
+            timeout = self.unrolled_run_gen.send(self.unrolled_run_next_send)
+        except StopIteration:
```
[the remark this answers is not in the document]
njsmith: Yeah, so this is a general question... in `trio.run`, if there's some unexpected crash, we convert it to a `TrioInternalError` and raise it, cool. In guest mode, I'm not actually sure what the desired behavior even is.

My cop-out in this first draft was to let any unexpected internal errors escape escape from the host loop callback, so the host loop could handle them... however it does that. (Probably dump them on the console and continue on.) That's not totally unreasonable, but maybe there's something better.

I guess ideally, we would call the completion callback with the `TrioInternalError`?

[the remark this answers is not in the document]
oremanj: Yep, "call the completion callback with the TrioInternalError" is what I was imagining. It seems closest to what normal Trio does. Relying on the host loop to do something reasonable with uncaught exceptions seems like a stretch to me based on standard practice in the field...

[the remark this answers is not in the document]
altendky: FWIW, I `sys.excepthook` in my PyQt programs and present a dialog.  Lots of people I suspect just lose the exceptions and painfully debug.

njsmith:
More wisdom from glyph:

```
<njs> glyph: so interestingly, Qt's native socket support on Windows is all based on WSAAsyncSelect, which AFAICT works by... running a thread in the background and then posting notifications back to the main thread
<glyph> Njs: oh everything on windows works like that :-). As I recall when i was measuring this the problems were largely with the implementation of the GIL. Maybe that’s gotten better in the intervening decade too
<njs> glyph: ah yeah plausible
<njs> glyph: the GIL did get replaced sometime in there, so at least it has a different set of perverse edge cases now...
<njs> glyph: worst comes to worst, I think I'd rather rewrite our GetQueuedCompletionEvents code in Rust so it can run in a thread without the GIL, than support pluggable backends
```

njsmith:
OK I think I rearranged enough stuff now that we should be converting any internals errors into `TrioInternalError` as appropriate and routing it into the `done_callback`.

We're also now calling `signal.set_wakeup_fd` by default in guest mode, so signals should hopefully work. And there's a new kwarg on `start_guest_run` to control whether we claim it or not, since this depends on the host loop.

Still need to figure out control-C handling.

njsmith:
Well, that's a fun new CI breakage: https://github.com/MagicStack/immutables/issues/46

I guess we might have to disable the `nightly` python build for a while.

altendky:
After making a working demo with Qt I tried to copy over some code I had for asyncio and twisted that allowed awaiting on Qt signals and got an error.  I boiled it down to a `trio.Event` and a `QTimer` to reproduce the issue.  At first I was getting a recursion error running with 90180663680b46438ffc34df1fc99228d440052c.  Now I'm getting the `UnboundLocalError` shown below.  Below was run with 5631de48d60add125ed969495a2c82bef613a636.

(I give up, I've already pulled updates three times while typing out this comment.  `:]` I'll try to add a test now that there's a test file)

<details> <summary>Failable Qt example</summary>

```python
import functools
import sys
import threading

import attr
import trio
import PyQt5.QtCore
import PyQt5.QtWidgets


class Runner(PyQt5.QtCore.QObject):
    signal = PyQt5.QtCore.pyqtSignal(object)

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

        self.signal.connect(self.slot)

    def run(self, f):
        print('Runner.run()', threading.get_ident())
        self.signal.emit(f)

    def slot(self, f):
        # TODO: no handling of context argument

        print('Runner.slot()', threading.get_ident())
        f()


@attr.s(auto_attribs=True)
class QTrio:
    application: PyQt5.QtWidgets.QApplication
    widget: PyQt5.QtWidgets.QTextEdit
    runner: Runner

    def run(self):
        PyQt5.QtCore.QTimer.singleShot(0, self.start_trio)
        return self.application.exec()

    async def main(self):
        print('QTrio.main()', threading.get_ident())
        self.widget.show()

        fail = True

        for i in range(10):
            if fail:
                print('a1')
                event = trio.Event()
                print('b1')
                timer = PyQt5.QtCore.QTimer.singleShot(1000, event.set)
                print('c1')
                await event.wait()
                print('d1')
            else:
                print('a2')
                await trio.sleep(1)
                print('b2')
            print('e')
            self.widget.append('{}'.format(i))
            print('f')

    def start_trio(self):
        # TODO: it feels a bit odd not getting anything back here.  is there
        #       really no object worth having access to?
        trio.lowlevel.start_guest_run(
            self.main,
            run_sync_soon_threadsafe=self.runner.run,
            done_callback=self.trio_done,
        )

    def trio_done(self, outcome):
        print('---', repr(outcome))
        self.application.quit()


def main():
    print('main()', threading.get_ident())
    qtrio = QTrio(
        application=PyQt5.QtWidgets.QApplication(sys.argv),
        widget=PyQt5.QtWidgets.QTextEdit(),
        runner=Runner(),
    )
    return qtrio.run()


main()
```
</details>

<details> <summary>From 90180663680b46438ffc34df1fc99228d440052c RecursionError: maximum recursion depth exceeded while calling a Python object</summary>

```
main() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
QTrio.main() 139969798338368
a1
b1
c1
Runner.run() 139969642370816
Runner.slot() 139969798338368
d1
e
f
a1
b1
c1
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
Runner.run() 139969798338368
Runner.slot() 139969798338368
  File "/home/altendky/repos/trio/y.py", line 27, in slot
    f()
  File "/home/altendky/repos/trio/trio/_core/_run.py", line 1146, in guest_tick
    timeout = self.unrolled_run_gen.send(self.unrolled_run_next_send)
  File "/home/altendky/repos/trio/trio/_core/_run.py", line 1845, in unrolled_run
  File "/home/altendky/repos/trio/trio/_core/_io_epoll.py", line 226, in process_events
    waiters = self._registered[fd]
RecursionError: maximum recursion depth exceeded while calling a Python object
```
</details>

<details> <summary>From b1319582b8e0d1ef9a89ec9b8559f5def61e08ce UnboundLocalError: local variable 'timeout' referenced before assignment</summary>

```
main() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
QTrio.main() 139804458121024
a1
b1
c1
Runner.run() 139804302153472
Runner.slot() 139804458121024
d1
e
f
a1
b1
c1
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
Runner.run() 139804458121024
Runner.slot() 139804458121024
--- Error(TrioInternalError('internal error in Trio - please file a bug!'))
  File "/home/altendky/repos/trio/y.py", line 27, in slot
    f()
  File "/home/altendky/repos/trio/trio/_core/_run.py", line 1156, in guest_tick
    if timeout <= 0 or type(events_outcome) is Error or events_outcome.value:
UnboundLocalError: local variable 'timeout' referenced before assignment
```
</details>

njsmith:
Ah, the recursion error is because you need to pass `QtCore.Qt.QueuedConnection` to your `connect` call. Otherwise, when you emit a signal from the main thread, Qt tries to run the callback immediately, rather than scheduling it to run on the next event loop tick.

And I think the unbound-local error is because I added some code to report unexpected errors like your `RecursionError`, but I forgot to make them abort the run loop after being reported :-)

altendky:
Quite so, thank you.  And I see you switched to an event in your gist anyways.

trio/_core/tests/test_guest_mode.py
```diff
@@ -0,0 +1,361 @@
...
+
+def test_host_altering_deadlines_wakes_trio_up():
+    def set_deadline(cscope, new_deadline):
+        cscope.deadline = new_deadline
+
+    async def trio_main(in_host):
+        with trio.CancelScope() as cscope:
+            in_host(lambda: set_deadline(cscope, -inf))
+            await trio.sleep_forever()
+            assert cscope.cancelled_caught
```
oremanj, line 105: I think these asserts should be dedented (AFAICT you're currently skipping them when the cancellation is raised on the previous line)

trio/_core/_run.py
```diff
@@ -1092,7 +1100,71 @@ class _RunStatistics:
...
+# So this object can reference Runner, but Runner can't reference it. The only
+# references to it are the "in flight" callback chain on the host loop /
+# worker thread.
+@attr.s(eq=False, hash=False, slots=True)
+class GuestState:
+    runner = attr.ib()
+    run_sync_soon_threadsafe = attr.ib()
+    done_callback = attr.ib()
+    unrolled_run_gen = attr.ib()
+    unrolled_run_next_send = attr.ib(factory=lambda: Value(None))
```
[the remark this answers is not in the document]
njsmith: `Outcome.send` and `Outcome.unwrap` are mutators – they consume the object. (To avoid accidentally rethrowing the same exception twice and getting garbled tracebacks.)

trio/_core/tests/test_guest_mode.py
njsmith: Nice catch!

trio/_core/_run.py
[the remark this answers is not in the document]
oremanj: Right you are, my bad!

trio/_core/_run.py
```diff
@@ -1092,7 +1100,71 @@ class _RunStatistics:
...
+            # XX if we had KI support, we'd have to do something with it here
+            self.done_callback(self.runner.main_task_outcome)
+            return
+        except TrioInternalError as exc:
+            self.done_callback(Error(exc))
+            return
+
+        # Optimization: try to skip going into the thread if we can avoid it
+        events_outcome = capture(self.runner.io_manager.get_events, 0)
+        if timeout <= 0 or type(events_outcome) is Error or events_outcome.value:
```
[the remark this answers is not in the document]
oremanj: Got it. I tend to prefer isinstance() because it's understood by type checkers, and it's marginally faster than type-is (although neither one is likely to be noticeable).

docs/source/reference-lowlevel.rst
```diff
@@ -528,10 +528,387 @@ Task API
...
+
+- **You must let the Trio program run to completion.** Many event
+  loops let you stop the event loop at any point, and any pending
+  callbacks/tasks/etc. just... don't run. Trio follows a more
+  structured system, where you can cancel things, but the code always
+  runs to completion, so ``finally`` blocks run, resources are cleaned
+  up, etc. If you stop your host loop early, before the
+  ``done_callback`` is invoked, then that cuts off the Trio run in the
+  middle without a chance to clean up, and can leave both your program
+  and Trio itself in an inconsistent state. (For example,
```
oremanj, line 814: dangling clause here?

trio/_core/_run.py
```diff
@@ -1805,10 +1805,57 @@ def start_guest_run(
...
+
+         When the Trio run has finished, Trio will invoke this callback to let
+         you know. The argument is an `outcome.Outcome`, reporting what would
+         have been returned or raised by `trio.run`. This function can do
+         anything you want, but commonly you'll want it to shut down the
+         host loop, unwrap the outcome, etc.
+
+      run_sync_soon_not_threadsafe: Optional. Like
+         ``run_sync_soon_threadsafe``, but will only be called from inside the
+         host loop's main thread.
```
oremanj, line 1850: Maybe explain why you'd want this? (if the host loop offers both, the non-threadsafe version is likely faster, and guest mode will have less overhead if Trio is able to use it.)

trio/_core/_run.py
```diff
@@ -1805,10 +1805,57 @@ def start_guest_run(
...
+    starts the host loop and then immediately starts the guest run, and then
+    shuts down the host when the guest run completes.
+
+    Args:
+
+      run_sync_soon_threadsafe: An arbitrary callable, which will be passed a
+         function as its sole argument::
+
+            def my_run_sync_soon_threadsafe(fn):
+                hi()
```
oremanj, line 1831: Maybe just `...` as the body? I'm not sure `hi()` adds any information

trio/_core/_run.py
```diff
@@ -1805,10 +1805,57 @@ def start_guest_run(
...
+
+         This callable should schedule ``fn`` to be run by the host on its
```
oremanj, line 1833: suggest `fn()` to emphasize it takes no arguments

docs/source/design.rst
```diff
@@ -539,6 +539,9 @@ supported systems.
 Guest mode
 ----------
 
+One of Trio's more unusual features is that it supports being run in
+"guest mode" on top of some other event loop (the "host").
```
pquentin, line 543: Should we link to the guest mode docs?

docs/source/reference-lowlevel.rst
```diff
@@ -528,10 +528,387 @@ Task API
...
+  making a GUI app with Qt as the host loop, then making a `cancel
+  button <https://doc.qt.io/qt-5/qpushbutton.html>`__ and connecting
+  it to a `trio.CancelScope` is as easy as writing::
+
+      # Trio code can create Qt objects without any special ceremony...
+      my_cancel_button = QPushButton("Cancel")
+      # ...and Qt can call back to Trio just as easily
+      my_cancel_button.clicked.connect(my_cancel_scope.cancel)
+
+- Consistent behavior: guest mode uses the code as regular Trio: the
```
pquentin, line 590:
```suggestion
- Consistent behavior: guest mode uses the same code as regular Trio: the
```

trio/_core/_run.py
njsmith: Ha, whoops! I had `...`, but then "temporarily" swapped it while trying to track down a weird sphinx error about not being able to tokenize the code bock... thank you, that would have been embarrassing :-)

docs/source/design.rst
njsmith: Oh right, I actually decided to put the high level "how does this even work" stuff into the main docs, and was going to delete this section. But then I forgot :-)

I guess the diagram is probably helpful... I'll move it to a comment by the functions.

trio/_core/_run.py
```diff
@@ -1577,6 +1666,92 @@ def remove_instrument(self, instrument):
...
+# | Run tasks                 |
+# | (unrolled_run)            |
+# +---------------------------+
+# | Block waiting for I/O     |
+# | (io_manager.get_events)   |
+# +---------------------------+
+# :
+#
+#
+# In guest mode using trio.start_guest_run, the scheduler runs on the main
```
oremanj, line 1699: trio.lowlevel.start_guest_run?

njsmith:
So after thinking about it a while, here's where I'm leaning on control-C handling (this is what's currently implemented in this PR):

We strongly recommend that Trio drive the lifetime of the surrounding loop, since... there's not really any better option. You can't have the surrounding loop directly drive Trio's lifetime, since Trio needs to run until the main task finishes. (You can have the surrounding loop nudge the main task to finish though, e.g. by cancelling stuff. But the actual shutdown sequence is still going to be Trio stops → host loop stops → program exits.)

Most loops have terrible stories for control-C, e.g. Python Qt apps tend to ignore it (you can find lots of discussions about this on the internet), asyncio tends to explode messily, etc.

Putting these two things together, it seems like most users will want Trio to take the `KeyboardInterrupt` safely, let it propagate out, and then that will trigger the host shutdown and exit the app. This does mean that you can't use control-C to break out of an infinite loop that's running in host mode, but realistically that doesn't work very well anyway, so... oh well?

There are a few loops that do do something sensible with control-C, e.g. twisted registers a signal handler that triggers the [loop shutdown procedure](https://twistedmatrix.com/documents/20.3.0/api/twisted.internet.interfaces.IReactorCore.html#fireSystemEvent). But I think this is fine:

- By default Trio will skip registering its control-C handler, because it will see that Twisted has already registered one.

- A twisted+trio integration is going to want to hook up Trio to the twisted shutdown procedure (i.e. adding a `"before"`, `"shutdown"` trigger that cancels the Trio main task and waits for it to exit), so using the Twisted control-C handling will be fine, no worse than what you get in a pure Twisted program

- A twisted+trio integration *might* want to override this so that Trio does register its control-C handler (e.g., hacky option: call `signal.signal(signal.SIGINT, signal.default_int_handler)` after starting the Twisted reactor but before calling `start_guest_run`). If so, then it'll give slightly nicer behavior for the Trio parts of your code (can break out of infinite loops), and effectively the same behavior for the Twisted parts of your code (after the control-C propagates out of Trio, it'll trigger a Twisted reactor shutdown, which is the same thing control-C does in vanilla Twisted).

So I think we don't even need any configuration knobs here... using our standard control-C logic in guest mode pretty much does the best thing possible in all cases.

njsmith:
To check for added overhead to regular `trio.run`, I wrote a tiny microbenchmark that just triggers scheduler ticks as fast as possible, figuring that this is the worst-possible-case for these changes:

```python
import time
import trio

LOOP_SIZE = 10_000_000

async def main():
    start = time.monotonic()
    for _ in range(LOOP_SIZE):
        await trio.lowlevel.cancel_shielded_checkpoint()
    end = time.monotonic()
    print(f"{LOOP_SIZE / (end - start):.2f} schedules/second")

trio.run(main)
```

On my laptop, with current master:

```
✦ ❯ PYTHONPATH=$HOME/trio perf stat python schedule-microbench.py
179727.56 schedules/second

 Performance counter stats for 'python schedule-microbench.py':

         55,762.59 msec task-clock                #    1.000 CPUs utilized          
               683      context-switches          #    0.012 K/sec                  
                 4      cpu-migrations            #    0.000 K/sec                  
             4,483      page-faults               #    0.080 K/sec                  
   178,323,685,701      cycles                    #    3.198 GHz                    
   348,488,306,514      instructions              #    1.95  insn per cycle         
    77,271,483,176      branches                  # 1385.723 M/sec                  
        93,989,483      branch-misses             #    0.12% of all branches        

      55.781963942 seconds time elapsed

      53.811304000 seconds user
       1.952215000 seconds sys
```

With current master + this PR:

```
✦ ❯ PYTHONPATH=$HOME/trio perf stat python schedule-microbench.py
177535.51 schedules/second

 Performance counter stats for 'python schedule-microbench.py':

         56,455.26 msec task-clock                #    1.000 CPUs utilized          
               223      context-switches          #    0.004 K/sec                  
                 2      cpu-migrations            #    0.000 K/sec                  
             4,042      page-faults               #    0.072 K/sec                  
   188,097,332,679      cycles                    #    3.332 GHz                    
   352,296,603,132      instructions              #    1.87  insn per cycle         
    78,393,967,951      branches                  # 1388.603 M/sec                  
       150,810,657      branch-misses             #    0.19% of all branches        

      56.456911871 seconds time elapsed

      54.687663000 seconds user
       1.768085000 seconds sys
```

So the walltime is ~1% slower, and it used ~1% more instructions. (I counted these mainly to cross-check against walltime, in case of stuff like weird thermal effects and CPU frequency changing.)

This is a *very* focused microbenchmark on a program that's really doing *nothing* except running through the scheduler as fast as it can. E.g. if you replace `cancel_shielded_checkpoint` with `checkpoint`, then it drops to ~25k schedules/second. (`checkpoint` is somewhat gratuitously slow, but still.) So my conclusion is that the existence of guest mode has extremely minimal impact on regular-mode programs.

docs/source/design.rst
```diff
@@ -534,3 +534,77 @@ there is no provision for "pluggable" backends. The intuition here is
 that we'd rather focus our energy on making one set of solid, official
 backends that provide a high-quality experience out-of-the-box on all
 supported systems.
+
+
+Guest mode
```
oremanj, line 539: I think this section can be removed now? If not, it should be fleshed out

docs/source/reference-lowlevel.rst
```diff
@@ -534,10 +534,413 @@ Task API
...
+``host_uses_signal_set_wakeup_fd=True`` to `start_guest_run`.
+
+If you aren't seeing any warnings with your initial prototype, you're
+*probably* fine. But the only way to be certain is to check your host
+loop's source. For example, asyncio may or may not use
+`signal.set_wakeup_fd` depending on the Python version and operating
+system.
+
+
+**Control-C handling** XX FIXME
```
oremanj, line 895: Update (can probably just remove since you discuss it above)

trio/_core/_io_kqueue.py
```diff
@@ -50,9 +61,19 @@ def handle_io(self, timeout):
...
                 # and loop back to the start
+        return events
+
+    def process_events(self, events):
         for event in events:
             key = (event.ident, event.filter)
-            receiver = self._registered[key]
+            try:
+                receiver = self._registered[key]
+            except KeyError:
```
oremanj, line 71: Maybe worth switching this to use the same idiom you used in the epoll manager? (ie, check whether it's the wakeup fd first, rather than catching the KeyError)

trio/_core/_ki.py
```diff
@@ -170,26 +171,31 @@ def wrapper(*args, **kwargs):
...
-            raise KeyboardInterrupt
-
-    signal.signal(signal.SIGINT, handler)
-    try:
-        yield
-    finally:
-        if signal.getsignal(signal.SIGINT) is handler:
-            signal.signal(signal.SIGINT, signal.default_int_handler)
+@attr.s
+class KIManager:
```
oremanj, line 175: Merge with master so this doesn't show up as part of the guest mode PR?

trio/_core/_run.py
```diff
@@ -509,10 +511,16 @@ def _might_change_registered_deadline(self):
...
                 runner = GLOBAL_RUN_CONTEXT.runner
+                if runner.is_guest:
+                    old_next_deadline = runner.next_deadline()
                 if old != inf:
                     del runner.deadlines[old, id(self)]
                 if new != inf:
                     runner.deadlines[new, id(self)] = self
+                if runner.is_guest:
+                    new_next_deadline = runner.next_deadline()
+                    if old_next_deadline != new_next_deadline:
```
oremanj, line 522: Technically I suppose this is only necessary if the new next deadline is sooner than the old one; otherwise it's harmless to wake up early and then go back to sleep. Doesn't much matter though.

trio/_core/_run.py
```diff
@@ -1118,6 +1130,62 @@ class Runner:
...
+                self.run_sync_soon_threadsafe(in_main_thread)
+
+            # XX temporary placeholder until #1545 is merged
+            def start_thread_soon(d, fn):
+                t = threading.Thread(daemon=True, target=lambda: deliver(capture(fn)))
+                t.start()
+
+            start_thread_soon(deliver, get_events)
+
+    def force_guest_tick_asap(self):
```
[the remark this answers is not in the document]
oremanj: Reminder on this?

docs/source/reference-lowlevel.rst
```diff
@@ -534,10 +534,413 @@ Task API
...
+  you're making a GUI app with Qt as the host loop, then making a
+  `cancel button <https://doc.qt.io/qt-5/qpushbutton.html>`__ and
+  connecting it to a `trio.CancelScope` is as easy as writing::
+
+      # Trio code can create Qt objects without any special ceremony...
+      my_cancel_button = QPushButton("Cancel")
+      # ...and Qt can call back to Trio just as easily
+      my_cancel_button.clicked.connect(my_cancel_scope.cancel)
+
+  (For async APIs, it's not that simple, but you can build on this
```
oremanj, line 596: build on this *to* make? (missing word)

trio/_core/_ki.py
njsmith: I did merge with master, and diff you're pointing at actually is new in the guest mode PR :-). Master has the change to flip the default control-C protection status; this change is to convert `ki_manager` from a context manager (which is awkward to wrap around guest mode) into a regular object with `install`/`close` methods.

njsmith: I'm not real happy with how the `KIManager` stuff is structured, so if you have any ideas on how to do it better than I'm interested, and it's probably worth scrutinizing a bit in review b/c everything around KI is tricky. But this does seem workable to me and we can always improve it later.

oremanj: Oh, whoops! I was going "I've seen this change before", but in fact I saw it in a previous commit on this diff. I'm happy with this approach.

trio/_core/_run.py
njsmith: I *think* you're right? But I couldn't quite convince myself that I understood all the timeout logic (esp. `idle_primed`) well enough to be confident of that, so I guess I'll leave it as-is. Like you say, it's hard to imagine this ever mattering at all.

Though I guess this is a hint that if we can simplify the timeout handling in `unrolled_run`, that would be good :-)

trio/_core/_run.py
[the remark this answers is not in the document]
njsmith: I think I'm going to leave it as "asap". Normally we use `soon` to mean "hey remember that this might not happen until some arbitrary time in the future", and here we kind of mean the opposite: "it was going to happen later, but we can't wait! make it happen now!".

njsmith:
Addressed all review comments.

Issue #455, opened 2018-02-22
Cancelled __aexit__ handlers eat exceptions


smurfix:
This is what happens when you don't wrap your async context's exit handler in a shielded cancel scope:

… trick question. Nothing happens – the error gets dropped on the floor. Since this is a fairly common mistake (at least IME) I wonder whether we can do something about it. If not (which is what I suspect) we need to document this pitfall more loudly.

```
import trio

class bar:
    async def __aenter__(self):
        return self
    async def __aexit__(self, *tb):
        await trio.sleep(0)

async def foo():
    with trio.open_cancel_scope() as s:
        async with bar():
            s.cancel()
            raise RuntimeError("Duh")

trio.run(foo)
```

njsmith:
Sorry this has been biting you! I'd like to understand better what exactly's happening. I guess it must be something like:

* Some code raises an exception

* While that's propagating, the surrounding scope gets cancelled

* While unwinding for the original exception, Python calls `__aexit__`, which calls some Trio primitive, which raises a `Cancelled` exception (probably this exception has the original exception stashed as its `__context__`, following Python's normal exception chaining rules)

* Then the `Cancelled` exception gets swallowed by the original cancel scope, so the original exception is lost entirely

Does that sound right?

It would be helpful to hear more concrete details about how you managed to trigger this combination of circumstances -- maybe it would give clues about how it could be handled better.

#285 and #416 touch on related/overlapping issues.

njsmith:
I should also say I'm very wary about making a blanket recommendation that people use shielding in any `__aexit__`/`finally` blocks... it might be the best solution for now in your case, I don't know, but in the general case shielding is dangerous as well (what if the blocking code in the `__aexit__` block gets stuck?), so hopefully there's a more specific solution we can come up with.

tgoodlet:
Ok I just got bit bad by something similar but incorporating async gens..
Makes sense but still was a nightmare to debug..

Check it:

```python
import trio

at_exit_entered = False

class bar:
    async def __aenter__(self):
        return self
    async def __aexit__(self, *tb):
        print(tb)
        global at_exit_entered
        at_exit_entered = True
        await trio.sleep(0)


async def nums(seed):
    async with bar():
        for i in range(seed):
            await trio.sleep(0)
            yield i


async def iterate_nums():
    with trio.open_cancel_scope() as cs:
        async for i in nums(10):
            print(i)
            # cs.shield = True
            await trio.sleep(0.1)
[...]

PR #1555, opened 2020-05-27
In run_process moved deliver_cancel exception handling to the caller code

### LINKED ISSUE #1532: What should happen if deliver_cancel raises an exception?
What should happen if deliver_cancel raises an exception? In the current implementation, the shielded cancel scope will prevent it from propagating until the process exits, which might take a while if the crash occurred before signaling the process in any way. Maybe on exception from a user-specified deliver_cancel we should call the default deliver_cancel to kill the process? Or just kill() since we don't care much about polish at that point, only about letting the user know what they need to fix.

_Originally posted by @oremanj in https://github.com/python-trio/trio/pull/1525_

guilledk:
Hey guys, so the test for this change is failing, I can't seem to catch the `OSError` on `trio/tests/test_subprocess.py:test_warn_on_failed_cancel_terminate`.

I guess it's being raised in another "context" (it context the right word)?

Also any comments on the exception logging? Is `trio.run_process` the correct logger? Also what about the message?

Closes #1532


trio/_subprocess.py
```diff
@@ -621,7 +614,14 @@ async def read_output(stream, chunks):
...
                     async def killer():
                         with killer_cscope:
-                            await deliver_cancel(proc)
+                            try:
+                                await deliver_cancel(proc)
+                            except BaseException as exc:
+                                LOGGER = logging.getLogger("trio.run_process")
+                                LOGGER.exception(
+                                    f"tried to kill process {proc!r}, but failed with: {exc!r}"
+                                    )
```
oremanj, line 623: Dedent this paren to satisfy black. (It's often easier to run `black setup.py trio` before you commit than it is to predict what black will like.)

This should have failed the formatting check, but it looks like we're not properly rejecting PRs that require black changes. I'll address that in a separate PR.

trio/tests/test_subprocess.py
```diff
@@ -464,7 +464,7 @@ def broken_terminate(self):
```
oremanj, line 464: This is no longer testing a warning, so you should change the name of the test (since it's currently "warn_on_failed_cancel_terminate")

oremanj:
> I'm not immediately sure why the test is failing. You'll need to do some digging.

I did some digging, and it's *super* subtle: #1559. Basically, the nursery in run_process() is surrounded by `async with process:`, and `Process.aclose()` raises `Cancelled` when it executed in a cancelled scope; this winds up with the error from the nursery getting stuck as the context of that `Cancelled` exception, which is swallowed by the cancel scope in the outermost nursery (in your test function).

I'm blanking on a fix right now. This is a much larger problem than just your diff -- see #455 for more discussion. @njsmith, any ideas?

njsmith:
Well, I guess there's an easy local fix... we already want to deprecate using `Process` objects as a context manager, so that will automatically take care of this :-) (#1104)

For the larger issue..... well, let's talk in #455/#1559.

njsmith:
To clarify: what we want to deprecate is the use of `async with <process object>`, which is the thing that's causing your test to fail.

That might also be the quickest way to get this working... we're already waiting for the process in all cases, so you could replace:

```python
async with await trio.open_process(...) as proc:
    blah
```

with

```python
proc = await trio.open_process(...)
blah
```

Issue #1523, opened 2020-05-15
Idea: "scoped" context variables that follow stack discipline


oremanj:
A context variable is a convenient way to track something about the scope in which a task executes. Sometimes that "something" effectively constitutes a promise to child tasks that some resource remains usable. For example, `trio_asyncio.open_loop()` provides the asyncio loop via a contextvar, but it's not much use if the `async with` block has exited already.

Unfortunately, contextvars are inherited from the environment of the task spawner, not the environment of the nursery in which the task runs. That means they can't reliably provide resources to a task, because the resources might be released before the task finishes. Imagine:
```
async with trio.open_nursery() as nursery:
    async with trio_asyncio.open_loop() as loop:
        nursery.start_soon(some_asyncio_task)
```

It would be nice to have a kind of contextvar that is inherited by child tasks based on the value it had when their nursery was opened, rather than the value it had when the child task was spawned. (In theory, cancellation status could use this if it existed. In practice, we probably want to continue to special-case cancellation status because it's so fundamental.)

Implementation idea: each scoped contextvar uses an underlying regular contextvar whose value is a tuple `(scoped contextvar's value, task that set it)`. Also, capture the contextvars context when creating a nursery (this is cheap, since contexts are meant to be shallow-copied all the time). When accessing the value of a scoped contextvar, check whether the underlying regular contextvar was set in the task we're currently in. If so, return the value from the underlying regular contextvar in the current context; otherwise, return its value in our parent nursery's captured context.

njsmith:
First reactions:

**Is this a problem?** Answer: Maybe?

I see your point about trio-asyncio. But in your example, I'm not sure it would be so bad if `some_asyncio_task` instead raised an exception when you exited the `loop` block... compare to:

```python
async with trio.open_nursery() as nursery:
    async with trio.open_tcp_stream(...) as stream:
        nursery.start_soon(some_stream_task, stream)
```

Seems like a pretty similar case, it does confuse new users sometimes, but... there's not really anything to do about it except teach folks what things mean.

I think the other way around might be potentially more compelling? If I do:

```python
global aio_supporting_nursery
async with trio_asyncio.open_loop() as loop:
    async with trio.open_nursery() as aio_supporting_nursery:
        await sleep_forever()

# Elsewhere:
aio_supporting_nursery.start_soon(aio_task)
```

Then currently that won't work, but maybe it would be convenient if it did? I'm not sure whether this example is realistic or contrived :-)

[...]

PR #1558, opened 2020-05-27
Add Task.eventual_parent_nursery introspection attribute

oremanj:
This tracks the nursery in which the task is going to run, even if it's not really running there yet because the `start()` that spawned it hasn't returned. Came up in discussion surrounding #1523 but seems independently useful.

### DIFF
--- a/newsfragments/1558.feature.rst
+++ b/newsfragments/1558.feature.rst
@@ -1,0 +1,11 @@
+Tasks spawned with `nursery.start() <trio.Nursery.start>` aren't treated as
+direct children of their nursery until they call ``task_status.started()``.
+This is visible through the task tree introspection attributes such as
+`Task.parent_nursery <trio.lowlevel.Task.parent_nursery>`. Sometimes, though,
+you want to know where the task is going to wind up, even if it hasn't finished
+initializing yet. To support this, we added a new attribute
+`Task.eventual_parent_nursery <trio.lowlevel.Task.eventual_parent_nursery>`.
+For a task spawned with :meth:`~trio.Nursery.start` that hasn't yet called
+``started()``, this is the nursery that the task was nominally started in,
+where it will be running once it finishes starting up. In all other cases,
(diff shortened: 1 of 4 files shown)


newsfragments/1558.feature.rst
```diff
@@ -0,0 +1,11 @@
...
+direct children of their nursery until they call ``task_status.started()``.
+This is visible through the task tree introspection attributes such as
+`Task.parent_nursery <trio.lowlevel.Task.parent_nursery>`. Sometimes, though,
+you want to know where the task is going to wind up, even if it hasn't finished
+initializing yet. To support this, we added a new attribute
+`Task.eventual_parent_nursery <trio.lowlevel.Task.eventual_parent_nursery>`.
+For a task spawned with :meth:`~trio.Nursery.start` that hasn't yet called
+``started()``, this is the nursery that the task was nominally started in,
+where it will be running once it finishes starting up. In all other cases,
+it hsa the same value as `~trio.lowlevel.Task.parent_nursery`.
```
njsmith, line 11: Typo: "hsa"

trio/_core/_run.py
```diff
@@ -1013,6 +1021,18 @@ def parent_nursery(self):
...
         return self._parent_nursery
 
+    @property
+    def eventual_parent_nursery(self):
+        """The nursery this task will be inside after it calls
+        ``task_status.started()``.
+
+        If this task has already called ``started()``, or if it was not
+        spawned using `nursery.start() <trio.Nursery.start>`, then
+        `eventual_parent_nursery` is the same as `parent_nursery`.
```
njsmith, line 1031: I think I have a mild preference for "if the task isn't in the middle of `start` shenanigans, then `eventual_parent_nursery` is `None`". What do you think?

oremanj: I'm indifferent between the two so will change it to suit your preference!

PR #1561, opened 2020-05-27
Update PyPI search link for Framework::Trio

Carreau:
It looks like PyPI now have a specific url for trove serach pattern.
This filter down the result to "only" 3 pages instead of 500.


pquentin:
The Travis build was successful, but for some reason GitHub hasn't seen it. Closing/reopening to trigger new builds.

PR #1568, opened 2020-06-02
process = await nursery.start(run_process, ...)

### LINKED ISSUE #1104: Should we have a conventional way to "spawn a process into a nursery"?

njsmith:
Fixes #1104

Todo:

- [ ] docs
- [ ] tests
- [ ] newsfragment
- [ ] deprecate Process.aclose

I think this gets the basic semantics right? Still need to do
everything around that though.


oremanj:
This looks like a good approach so far!

trio/_subprocess.py
```diff
@@ -55,22 +56,19 @@ def pidfd_open(fd, flags):
...
-    make sure your process doesn't end up getting abandoned by mistake or
-    after an exception, you can use ``async with``::
+    This way, `run_process` supervises the process, and makes sure that it is
+    cleaned up, while optionally checking the output, feeding it input, and so
+    on.
 
-       async with await trio.open_process(...) as process:
-           ...
+    If you need more control – for example, because you want to spawn a child
+    process that outlives your program – then you can use `open_process`::
```
goodboy, line 69: Wait is this true? 

Can `open_process()` outlive the "program" given the `subprocess` routine is launched using [.to_thread.run()`](https://github.com/python-trio/trio/blob/c0dbe792095b67655cb14deee42d4a050625846b/trio/_subprocess.py#L369)?

Maybe "program" isn't the ideal term here?

trio/_subprocess.py
```diff
@@ -592,44 +603,49 @@ async def my_deliver_cancel(process):
...
+                proc.stdout = None
+                proc.stdio = None
+            if capture_stderr:
                 nursery.start_soon(read_output, proc.stderr, stderr_chunks)
-            try:
+                proc.stderr = None
+            task_status.started(proc)
+            await proc.wait()
+        except BaseException:
+            with trio.CancelScope(shield=True):
```
goodboy, line 635: Yeah this is great; love how the spawning maps 1-to-1 with a task now.

This is exactly what we [ended up with in `tractor`](https://github.com/goodboy/tractor/blob/master/tractor/_spawn.py#L232) since we also needed a shielded `Process.wait()` before the call to `Process.__aexit__()` (internal IPC protocol takes care of inter-process cancellation) and the task mapping pattern was the only way to reliably collect sub-process errors and get cancellation logic right (even when using `multiprocessing`).

So the only motivation now for `open_process()` use directly is a non-SC process-lives-longer then "current nursery" type scenario yes? This `run_process()` call (presuming `.start()` use) is effectively the same but with guaranteed shielded supervision on teardown.

My only last question is **if** the async context mng interface is deprecated then is there now no longer a way to *manually cancel only a specific process*?

Afaict here, you can't force the teardown (634 onwards) without cancelling the task (but that usually requires cancelling a group scope since, nursery, and a new cancel scope isn't created internally here per process). So if cancellation is the new `__aexit__()`, then we somewhat lose granular control of per-process cancellation and/or its completion detection right? Unless of course you want to use `open_process()` directly (but then in that case shouldn't `__aexit__()` code get moved to the post-`yield` body of `open_process()`)?

The streams closing and process killing in [`.aclose()`](https://github.com/python-trio/trio/blob/c0dbe792095b67655cb14deee42d4a050625846b/trio/_subprocess.py#L182) is something we currently depend on in `tractor`.

With that last question, forgive me if I'm missing something.

shamrin:
@njsmith What is the current state of this PR? Now that I've stumbled upon `open_process()` weirdness in #1844, I'd rather try to help here rather than work-around `open_process()`.

trio/_subprocess.py
njsmith: We use a thread to spin up the process, but once it's started it's an independent entity.

trio/_subprocess.py
njsmith: > My only last question is if the async context mng interface is deprecated then is there now no longer a way to manually kill only a specific process?

You can still kill a process. `Process.terminate`/`Process.kill`/`Process.send_signal` are all still available, and you can still close the streams if you want. `Process.aclose` was always just a convenience wrapper around other existing public APIs, and that hasn't changed.

trio/_subprocess.py
goodboy: Right, my quiff is on the use of the term *program*.
I presume no thread outlives the `trio` program given the `deamon=True` flag is used iirc?

trio/_subprocess.py
goodboy: > Process.aclose was always just a convenience wrapper around other existing public APIs, and that hasn't changed.

Ok so it will henceforth follow that users *can* roll their own version of `.aclose()` but that won't be default behavior.

trio/_subprocess.py
njsmith: `open_process` doesn't spawn a thread, it spawns a process... there are no threads here (except very briefly as an implementation detail that isn't mentioned in the docs)

goodboy: Ahh i see. I didn't realize you could intentionally leak processes; gtk.

njsmith:
At long last, this is ready for review!

trio/_subprocess.py
```diff
@@ -568,12 +626,23 @@ async def my_deliver_cancel(process):
...
+        if options.get("stdout") is subprocess.PIPE:
+            raise ValueError(
+                "stdout=subprocess.PIPE doesn't make sense without "
+                "nursery.start, since there's no way to access the pipe"
+            )
+        if options.get("stderr") is subprocess.PIPE:
+            raise ValueError(
+                "stderr=subprocess.PIPE doesn't make sense without "
+                "nursery.start, since there's no way to access the pipe"
+            )
```
richardsheridan, line 645: I would say "is only valid with nursery.start, which provides access the pipe".  Fewer negations is more understandable!

trio/_subprocess_platform/__init__.py
```diff
@@ -4,6 +4,7 @@
 import sys
 from typing import Optional, Tuple, TYPE_CHECKING
 
+import trio
```
richardsheridan, line 7: I'm surprised this doesn't lead to a circular import problem. Also this kind of change seems out of place in this PR?

trio/_subprocess.py
```diff
@@ -482,24 +535,26 @@ async def run_process(
...
+      stdin (:obj:`bytes`, subprocess.PIPE, file descriptor, or None): The
+          bytes to provide to the subprocess on its standard input stream, or
+          ``None`` if the subprocess's standard input should come from the
+          same place as the parent Trio process's standard input. As is the
+          case with the :mod:`subprocess` module, you can also pass a file
+          descriptor or an object with a ``fileno()`` method, in which case
+          the subprocess's standard input will come from that file. And when
+          starting `run_process` as a background task, you can use
+          ``stdin=subprocess.PIPE``, in which case `Process.stdin` will be a
+          `~trio.abc.SendStream` that you can use to send data to the child.
```
richardsheridan, line 547: For most of these docs, there is a visual distinction (indentation, newlines) between info related to `await run_process` vs `await nursery.start(run_process)` but it is kind of lacking here. On the other hand, I don't know if sphinx will permit a newline here?

richardsheridan (commented):
Happy to say goodbye to `async with await`! My comments relate mainly to documentation; the code changes look good to me for whatever that's worth :wink:.

trio/_subprocess_platform/__init__.py
oremanj: It looks to me like it's potentially resolving a circular import, since previously we imported `lowlevel` directly (further below).

Also in general there has been a semi-recent trend towards having Trio use its own APIs under their public names where possible, because that makes it easier for others to understand the code.

oremanj (commented):
This looks great, thank you for finishing it up! Just a few small comments

trio/_subprocess_platform/__init__.py
njsmith: It's not a circular import because it just pulls in the `trio` module object – that module object isn't fully populated yet, so trying to access attributes on it is likely to fail. But we're not touching the module's attributes during the import, just the module object itself. We don't access attributes until later, when functions in this module are actually called, by which point the imports have all finished and the `trio` object is all filled in.

docs/source/reference-io.rst
```diff
@@ -664,22 +664,43 @@ Spawning subprocesses
...
+      normally for simple capturing, or write the loop yourself if you
+      have unusual needs. :meth:`~subprocess.Popen.communicate` has
+      quite unusual cancellation behavior in the standard library (on
+      some platforms it spawns a background thread which continues to
+      read from the child process even after the timeout has expired)
+      and we wanted to provide an interface with fewer surprises.
+
+If `trio.run_process` is too limiting, we also offer a low-level API,
+`trio.lowlevel.open_process`. For example, if you want to spawn a
+child process that will and outlive the parent process and be
```
richardsheridan, line 701: I think there is an extra "and" on this line.

njsmith:
The 3.10-dev failure appears to be some kind of incompatibility between IPython and 3.10-dev, nothing in our code.

PR #1569, opened 2020-06-02
Make test more specific to prevent flakiness

njsmith:
Sometimes ResourceWarnings decide to print during this test, and that
used to cause it to fail every once in a while.


njsmith:
I was hoping #1565 would fix this, but then: https://github.com/python-trio/trio/pull/1568/checks?check_run_id=729466074

So we should probably fix it correctly...

PR #1574, opened 2020-06-04
Remove deprecated BlockingTrioPortal

pquentin:
TODO
- [x] coverage
- [x] newsfragment

### DIFF
--- a/trio/_threads.py
+++ b/trio/_threads.py
@@ -21,20 +21,6 @@ from ._util import coroutine_or_error
 # Global due to Threading API, thread local storage for trio token
 TOKEN_LOCAL = threading.local()
 
-
-class BlockingTrioPortal:
-    def __init__(self, trio_token=None):
-        if trio_token is None:
-            trio_token = trio.lowlevel.current_trio_token()
-        self._trio_token = trio_token
-
-    def run(self, afn, *args):
-        return from_thread_run(afn, *args, trio_token=self._trio_token)
-
-    def run_sync(self, fn, *args):
-        return from_thread_run_sync(fn, *args, trio_token=self._trio_token)
-
-
 _limiter_local = RunVar("limiter")
 # I pulled this number out of the air; it isn't based on anything. Probably we
 # should make some kind of measurements to pick a good value.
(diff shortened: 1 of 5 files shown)


trio/_threads.py
```diff
@@ -227,17 +213,6 @@ def _run_fn_as_system_task(cb, fn, *args, trio_token=None):
                 "this thread wasn't created by Trio, pass kwarg trio_token=..."
             )
 
-    # TODO: This is only necessary for compatibility with BlockingTrioPortal.
```
oremanj, line 230: I think we might still need this for the case where the user passes the trio_token?

pquentin: I have no idea, I just grepped for BlockingTrioPortal, but I'm not familiar with those APIs. @njsmith Can you please confirm?

pquentin:
The coverage change happens in `public_modules`:

Screenshot 2020-06-04 at 10 55 11

I think we still want to ignore names starting with `_` for future deprecations, so should I add two nocover pragmas?

trio/_threads.py
njsmith: I think @oremanj is right. The test case would be something like:

```python
async def test():
    trio_token = trio.lowlevel.current_trio_token()
    with pytest.raises(RuntimeError):
        trio.from_thread.run_sync(lambda: None, trio_token=trio_token)
```

If we fail to detect we're calling `from_thread` inside the Trio thread, then this will deadlock.

njsmith:
> I think we still want to ignore names starting with _ for future deprecations, so should I add two nocover pragmas?

That seems reasonable, yeah.

njsmith:
Looks like there are a few more coverage changes still?

njsmith:
So close... now the last line of the new test is causing a partial coverage line, because the `lambda: None` never gets called. Good place for a `# pragma`, or I guess we could make it more specific by writing it like

```python
def not_called():  # pragma: no cover
    assert False

from_thread_run_sync(not_called, trio_token=trio_token)
```

pquentin:
Thanks for investigating! This was subtle. I found about it independently, and only because I generated an HTML report that explcitly mentioned the lambda. I think passing a None functiom is simpler, and it fixes the issue too.

njsmith:
We'll just have to switch back to passing a function when we eventually get type hints working :-) But, up to you. Feel free to merge when you're happy.

pquentin:
Thanks! Right, I forgot about mypy, your version is better, thanks.

PR #1579, opened 2020-06-07
Allow start_clock() to start the very first system task; make MockClock do so

oremanj:
Preparatory to #1521 in which the order will do anything. As discussed in #265, the autojumper should be the very last thing to get shut down, even after the run_sync_soon task.

This requires some shenanigans because start_clock() runs before the first step of init. We effectively execute the system nursery `__aenter__` outside of async context, so as to have a system nursery when start_clock() runs.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -753,6 +753,7 @@ class NurseryExiter:
     It is used only for the system nursery, which gets entered outside
     of async context.
     """
+
     _manager = attr.ib()
 
     async def __aenter__(self):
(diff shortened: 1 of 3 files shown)


njsmith:
What do you think of the idea discussed in the last few paragraphs of this comment, of making `MockClock` use an instrument instead of a task? https://github.com/python-trio/trio/issues/239#issuecomment-499756224

oremanj:
Interesting! I had rather glossed over that issue before, but have now gone back and read it in detail.

I think the instrument-based MockClock is going to be challenging to make work in all edge cases without more instrumentation hooks churn than I'm up for right now. (This issue is to unblock #1521, which together with #1543 unblocks #1564 and #1537... the yak stack depth is getting intimidating!) I commented on #239 with some hazards I saw when I sat down to try it.

I certainly don't mind removing the unusual system_nursery initialization in this PR if MockClock later changes to make it unnecessary.

njsmith:
I'm intrigued by the idea in this PR to make our bootstrap code start by creating a nursery, instead of starting by creating a task. It's not something I've really thought about before.

As far as fixing `MockClock` goes though, I think #1588 is probably easier to reason about than this PR, and also sets us up better for future features like chaos scheduling?

oremanj:
Closing this in favor of #1588. If we later need the system nursery to exist earlier for something else, that's what version control is for!

PR #1580, opened 2020-06-08
Don't shuffle each batch, instead reverse them half the time.

palkeo:
I consistently see batch shuffling taking 1% CPU on a real-world
application, and this is a simple fix to make this overhead disappear.
Suggested by @njsmith.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -2076,6 +2076,7 @@ def unrolled_run(runner, async_fn, args, host_uses_signal_set_wakeup_fd=False):
             # tie-breaker and the non-deterministic ordering of
             # task._notify_queues.)
             batch = list(runner.runq)
+            runner.runq.clear()
             if _ALLOW_DETERMINISTIC_SCHEDULING:
                 # We're running under Hypothesis, and pytest-trio has patched
                 # this in to make the scheduler deterministic and avoid flaky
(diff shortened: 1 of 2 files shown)


trio/_core/_run.py
```diff
@@ -2084,7 +2084,10 @@ def unrolled_run(runner, async_fn, args, host_uses_signal_set_wakeup_fd=False):
                 # seeded for tests.
                 batch.sort(key=lambda t: t._counter)
             runner.runq.clear()
-            _r.shuffle(batch)
+            # 50% chance of reversing the batch, this way each task
+            # can appear before/after any other task.
+            if _r.random() < 0.5:
+                batch = batch[::-1]
```
pquentin, line 2090: I guess `batch.reverse()` can only be more efficient since it reverses the list in place?

njsmith: Yeah:

```python
In [1]: l = list(range(100))                                                              

In [2]: %timeit l[::-1]                                                                   
480 ns ± 2.71 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit l.reverse()                                                               
120 ns ± 5.19 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

trio/_core/_run.py
```diff
@@ -2084,7 +2084,10 @@ def unrolled_run(runner, async_fn, args, host_uses_signal_set_wakeup_fd=False):
                 # seeded for tests.
                 batch.sort(key=lambda t: t._counter)
             runner.runq.clear()
-            _r.shuffle(batch)
```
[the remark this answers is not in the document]
njsmith: I guess that would be fine, but I don't think it's super important... I doubt hypothesis will actually find interesting orderings with simple random shuffling; I think we'll need more rigorous race-condition-provoking heuristics like discussed in #239.

trio/_core/_run.py
palkeo: Good point, I didn't remember it has a method to sort in-place.

trio/_core/_run.py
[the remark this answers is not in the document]
palkeo: Done.

PR #1593, opened 2020-06-09
Minor overhaul of our newsfragment categories

njsmith:
In particular, splits features into "regular features" and "headline
features", and split "removal" into "breaking changes without
deprecations" and "deprecations".


pyproject.toml
```diff
@@ -15,8 +15,15 @@ directory = "newsfragments"
...
 
-# Unfortunately there's no way to simply override
-# tool.towncrier.type.misc.showcontent
+[[tool.towncrier.type]]
+directory = "headline"
+name = "Headline features"
+showcontent = true
+
+[[tool.towncrier.type]]
+directory = "breakin"
```
smurfix, line 24: Could we perchance name this "breaking" so that it doesn't evoke associations with burglars et al.?

njsmith: ha, nice catch thanks

pyproject.toml
```diff
@@ -15,8 +15,15 @@ directory = "newsfragments"
 underlines = ["-", "~", "^"]
 issue_format = "`#{issue} <https://github.com/python-trio/trio/issues/{issue}>`__"
 
-# Unfortunately there's no way to simply override
-# tool.towncrier.type.misc.showcontent
+[[tool.towncrier.type]]
+directory = "headline"
+name = "Headline features"
```
oremanj, line 20: Let's be consistent about capitalization -- either this should be "Headline Features" or (e.g.) `deprecated` should be "Deprecations and removals". (I personally prefer leading-capital-only, but title-case is fine too.)

pquentin: Agreed. And #1574 was just merged, sorry about the "conflict"

njsmith:
Updated to address comments

I also marked the RTD build as required

(I guess we can get rid of the "Check docs" build?)

njsmith:
On further thought, I also rewrote some of the general instructions ins `newsfragments/README.rst`

pquentin:
> (I guess we can get rid of the "Check docs" build?)

I guess we can! However the Read the Docs build just failed mysteriously, and I can't merge this since that build is now required. :)

njsmith:
Huh, yeah, apparently the RTD build just.... stopped in the middle.

Well, all our CI systems sometimes fail due to random issues with the underlying infrastructure, so I guess this isn't a huge shock. I guess it's nice to know what that looks like on RTD :-)

PR #1596, opened 2020-06-09
Remove remaining aliases deprecated in 0.12.0

pquentin:
I still need a news fragment, but I'll wait for #1593 to go through. ~~And I guess I'll add the newsfragment for https://github.com/python-trio/trio/pull/1594 here too.~~


pquentin:
Please take another look.

PR #1598, opened 2020-06-10
Deprecate the tiebreaker= argument to wait_all_tasks_blocked

njsmith:
It was only added for the autojump clock, it wasn't very good at that,
and since gh-1588, the autojump clock doesn't even use it anyway.


oremanj:
test_highlevel_open_tcp_listeners check_backlog failure again

PR #1600, opened 2020-06-10
Deprecate support for multiple tasks in the private old_nursery used by start()

### LINKED ISSUE #1599: start() shouldn't support multiple tasks in the old_nursery
Currently, `Nursery.start()` is implemented internally by opening a private nursery and starting the new task there, and `TaskStatus.started()` moves all tasks from that private nursery (the "old_nursery") to the task's final home ("new_nursery", the nursery whose `start()` method was called). In theory, there should only be one task in the old nursery, since the nursery isn't exposed anywhere outside the body of `start()` for anyone to be able to start more. In practice, it's possible to access the old nursery via `current_task().parent_nursery` in the newly-started task, so we may have some users who are starting more tasks there, maybe not even realizing that it's not the same nursery on which start() was called. This currently works fine, but it's confusing and makes it hard to implement extensions to the nursery semantics such as #1521.

Since this works now, we probably want a deprecation period. The purpose of this issue is mostly to have a number to put in the deprecation message. If your code runs into this, ideally you would refactor it to pass a nursery around explicitly rather than assuming you can always start new tasks in your parent nursery. If that's not possible for some reason, you can use `task.eventual_parent_nursery or task.parent_nursery` to determine the parent nursery that will prevail once the `start()` call completes.

oremanj:
Fixes #1599; see there (or read the newsfragment) for rationale.

### DIFF
--- a/newsfragments/1599.deprecated.rst
+++ b/newsfragments/1599.deprecated.rst
@@ -1,0 +1,19 @@
+In the current implementation of `nursery.start()
+<trio.Nursery.start>`, the new task starts running in a private
+nursery created inside :meth:`trio.Nursery.start`, and only moves to
+its "real" nursery once it calls ``task_status.started()``. Since Trio
+doesn't hand out any references to that private temporary nursery, we
+would like to be able to assume it only contains the single task being
+started. However, the nursery object is still accessible using
+debugging APIs such as
+``trio.lowlevel.current_task().parent_nursery``, and it's possible
+that some users have been using those APIs to start tasks in it. (If
+it wasn't clear, we strongly recommend against this sort of thing,
+because it makes private implementation details affect the behavior of
+your code.)
+
+Until now, these sorts of shenanigans unintentionally *worked*: the
+new tasks would move alongside the task being ``start()``\ed, and one
+might not even realize they weren't getting the ``parent_nursery``
+they were expecting. Supporting this unintentional feature is becoming
+burdensome, though, so we're deprecating it with no replacement.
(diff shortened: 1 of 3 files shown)


njsmith:
Like I mentioned in chat: this might be a good idea, but I think it's hard to tell right now. The unknowns for me are:

- how much trouble does this actually cause for implementing #1521? That's still in the planning phase, and it's hard to tell this kind of thing without an actual implementation

- someone finding the nursery through introspection and spawning extra tasks into it is obviously a pretty obscure thing that we don't care that much about supporting. But still, if it happens, we have to do *something*. And in the current codebase, making it "Just Work" is pretty trivial. OTOH, if we want to raise an error or something, then... what do we do, exactly? We can't just abandon the tasks that aren't supposed to be there...

Hmm. I guess a less-intrusive way to handle this would be to make `start_soon` on the secret internal `start` nursery raise an error, similar to if the nursery was closed. Then the issue couldn't arise in the first place.

That feels a lot more "obvious good idea" to me.

njsmith:
So yeah having slept on this a bit I think I'd be fine with a reworked version of this PR that issued the deprecation at the point of spawning a child into that nursery (from `spawn_impl`), instead of later when it's discovered that there are multiple children running.

oremanj:
It's quite a lot of trouble for #1521 to support the case where new *service* tasks were spawned into the old_nursery. But it would work fine to forbid that case while continuing to allow non-service tasks to come along for the ride. I just figured it was better to be consistent.

Marking this as draft until we come back to #1521.

Issue #1595, opened 2020-06-09
Slower performance compared to curio


rthalley:
I've used trio in some projects lately, and also recently added it to dnspython.  As part of some random performance testing, I compared "minimal DNS server" performance between ordinary python sockets, curio, and trio.  I'll attach the trio program to the bottom of this report.  Trio performed significantly worse.  E.g. on a particular linux VM:

Regular python I/O:  9498 QPS
Curio: 8979 QPS
Trio: 5007 QPS

I've seem similar behavior on a Mac:

Regular python I/O: 8359 QPS
Curio: 8061 QPS
Trio: 3425 QPS

I'm using dnsperf to generate load, and it does a good job of keeping UDP input queues full, so the ideal expected behavior if you strace the program is to see a ton of recvfrom() and sendto() system calls, and nothing else.  In particular, you don't expect to see any epoll_wait().  Ordinary python I/O and curio behave as expected, but going through the loop in trio looks like:

recvfrom()
clock_gettime()
epoll_wait()  (not waiting on any fds if I'm reading strace output correctly)
clock_gettime()
clock_gettime()
epoll_wait()
clock_gettime()
sendto()
clock_gettime()
epoll_wait()
clock_gettime()

I don't understand trio's internals enough to debug this further at the moment, but I thought I would make a report, as this seems like excessive epolling.

I haven't dtrussed on the mac, but python tracing indicated a lot of time related to kqueue.

This was with trio 0.15.1, and Cpython 3.8.3 and 3.7.7.

```
import socket
import trio
import trio.socket

import dns.message
import dns.rcode

async def serve():
    with trio.socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
        await s.bind(('127.0.0.1', 5354))
        while True:
            (wire, where) = await s.recvfrom(65535)
            q = dns.message.from_wire(wire)
            r = dns.message.make_response(q)
            r.set_rcode(dns.rcode.REFUSED)
            await s.sendto(r.to_wire(), where)

def main():
    trio.run(serve)

if __name__ == '__main__':
    main()
```

rthalley:
P.S.  I enjoy trio, keep up the good work!

rthalley:
Also, if you want to generate the load, create a file called input with one line

www.example A

and then run

dnsperf -d input -p 5354 -l 10

rthalley:
[...]

PR #1608, opened 2020-06-11
Inject runner into TheIOManager

alexchamberlain:
Improve performance by injecting the runner, instead of using thread locals.

I've been following `trio` for a while, tinkering to understand how the `_core` works and https://github.com/python-trio/trio/issues/1595 inspired me to see whether a small change could speed things up. This change improved the dnsperf example given in that issue by around 10% on a couple of runs. Can anyone reproduce similar performance gains?

Is injecting the runner like this of interest, as I think there are a fair number of places where TLS is used to access the runner by an object in `_core`?

### DIFF
--- a/trio/_core/_io_common.py
+++ b/trio/_core/_io_common.py
@@ -4,7 +4,7 @@ from .. import _core
 
 
 # Utility function shared between _io_epoll and _io_windows
-def wake_all(waiters, exc):
+def wake_all(runner, waiters, exc):
     try:
         current_task = _core.current_task()
     except RuntimeError:
@@ -16,7 +16,7 @@ def wake_all(waiters, exc):
             if task is current_task:
                 raise_at_end = True
             else:
-                _core.reschedule(task, outcome.Error(copy.copy(exc)))
+                runner.reschedule(task, outcome.Error(copy.copy(exc)))
             setattr(waiters, attr_name, None)
     if raise_at_end:
         raise exc
--- a/trio/_core/_io_epoll.py
+++ b/trio/_core/_io_epoll.py
@@ -182,6 +182,7 @@ class EpollWaiters:
 
 @attr.s(slots=True, eq=False, hash=False)
 class EpollIOManager:
+    _runner = attr.ib()
     _epoll = attr.ib(factory=select.epoll)
     # {fd: EpollWaiters}
     _registered = attr.ib(factory=lambda: defaultdict(EpollWaiters))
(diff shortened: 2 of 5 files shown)


njsmith:
I'm excited that you're working on this :-)

10% seems too good to be true though... all the benchmarks I've done suggest that accessing thread-locals in Python is very fast, and I'm not sure that DNS benchmark even touches the IOManager. (We only enter the IOManager when we try to read from a socket that has no data, or try to write to a socket that the other side isn't draining, and that shouldn't happen much or at all with `dnsperf` blasting data as fast as it can.)

Can you give more details on your measurements?

alexchamberlain:
I'm going to have to withdraw the 10% claim, as in retrospect, my initial tests weren't sufficient. Once I ensured I had plenty of free memory, an idle CPU and give the server the highest priority, the performance differences are negligible. With this change on my Mac, I achieved 3251.4 queries per second over 5 runs, where as with master I achieved 3275.4 queries per second. (That's right, master was faster on this test.)

The client was run as `for i in {1..5}; do dnsperf -d input -p 5354 -l 10; sleep 2; done`. The server code was

```python
import socket
import trio
import trio.socket

import dns.message
import dns.rcode


async def serve():
    with trio.socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
        await s.bind(("127.0.0.1", 5354))
        while True:
            (wire, where) = await s.recvfrom(65535)
            q = dns.message.from_wire(wire)
            r = dns.message.make_response(q)
            r.set_rcode(dns.rcode.REFUSED)
            await s.sendto(r.to_wire(), where)


def main():
    trio.run(serve)


if __name__ == "__main__":
    main()
```

Note that the `dns` module has to be downloaded from http://www.dnspython.org/ AFAICT.

alexchamberlain:
`KqueueIOManager.process_events` is being called a lot, so this change is being "tested".

alexchamberlain:
> We only enter the IOManager when we try to read from a socket that has no data, or try to write to a socket that the other side isn't draining

@njsmith Where is this logic? [`_nonblocking_helper`](https://github.com/python-trio/trio/blob/master/trio/_socket.py#L554) looks like it always calls into the IOManager.

njsmith:
@alexchamberlain `process_events` is being called, but is `reschedule`? If there aren't any events then `process_events` just exits early.

alexchamberlain:
Yes, I added a very naive print in both functions an they were interleaved.

njsmith:
>  _nonblocking_helper looks like it always calls into the IOManager.

It's this bit:

https://github.com/python-trio/trio/blob/043def5a558908aae364e64281ea285449bf4451/trio/_socket.py#L576-L577

First we try calling the `send`/`recv`/whatever directly, and if that succeeds then we skip calling `wait_fn`.

alexchamberlain:
#fail I totally misread that!

njsmith:
> Yes, I added a very naive print in both functions an they were interleaved.

Sorry, I edited my comment slightly... `reschedule` is definitely being called, because that's the core function that's always used to schedule tasks :-). The question is whether `process_events` is the one calling `reschedule`, because if not.

njsmith:
> #fail I totally misread that!

`_nonblocking_helper` and `_try_sync` are super convoluted, and also slow... they're trying to reduce code duplication, but we should probably switch to doing some code gen instead, so at runtime it's straightforward code without all the indirections. (And if you look at the generated code, it will be more obvious what it's actually doing.)

alexchamberlain:
You're absolutely right; `len(events) == 0`. Back to my code reading...

oremanj:
Going to close this since the discussion suggests it's unlikely to be a performance win after all.

Issue #1604, opened 2020-06-10
test_race_between_idle_exit_and_job_assignment deadlocked


pquentin:
See https://github.com/python-trio/trio/runs/756580934?check_suite_focus=true. I was able to retrieve the test name thanks to the traceback. The test timed out on an acquire() call.

```
2020-06-10T05:22:22.9465010Z Timeout (0:01:00)!
2020-06-10T05:22:22.9465567Z Thread 0x00007feaed267700 (most recent call first):
2020-06-10T05:22:22.9466805Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/site-packages/jedi/inference/compiled/subprocess/__init__.py", line 37 in _enqueue_output
2020-06-10T05:22:22.9467371Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/threading.py", line 864 in run
2020-06-10T05:22:22.9467782Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/threading.py", line 916 in _bootstrap_inner
2020-06-10T05:22:22.9468195Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/threading.py", line 884 in _bootstrap
2020-06-10T05:22:22.9468437Z 
2020-06-10T05:22:22.9468695Z Thread 0x00007feaef1be700 (most recent call first):
2020-06-10T05:22:22.9469487Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/site-packages/trio/_core/tests/test_thread_cache.py", line 102 in acquire
2020-06-10T05:22:22.9470319Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/site-packages/trio/_core/_thread_cache.py", line 62 in _work
2020-06-10T05:22:22.9470793Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/threading.py", line 864 in run
2020-06-10T05:22:22.9471195Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/threading.py", line 916 in _bootstrap_inner
2020-06-10T05:22:22.9471607Z   File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/threading.py", line 884 in _bootstrap
2020-06-10T05:22:22.9471852Z 
[...]

PR #1610, opened 2020-06-12
Enforce mypy-cleanliness on trio._core

pquentin:
This is a less ambitious version of #1254, the idea being that we can introduce mypy cleanliness gradually. I did copy relevant code from #1254, and fixed other issues too.

The relevant message in CI can be found in the [check formatting check](https://github.com/python-trio/trio/pull/1610/checks?check_run_id=764547595):
```
++ mypy -p trio._core --platform linux
Success: no issues found in 45 source files
```

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -27,6 +27,8 @@ flake8 trio/ \
 # We specify Linux so that we get the same results on all platforms. Without
 # that, mypy would complain on macOS that epoll does not exist, and we can't
 # ignore it as it would cause a mypy error on Linux
+# In the future, we will run mypy on all platforms and use platform assert
+# guards to avoid typechecking the wrong IOManagers
 mypy -p trio._core --platform linux || EXIT_STATUS=$?
 
 # Finally, leave a really clear warning of any issues and exit
--- a/check.sh
+++ b/check.sh
@@ -23,6 +23,12 @@ flake8 trio/ \
     --ignore=D,E,W,F401,F403,F405,F821,F822\
     || EXIT_STATUS=$?
 
+# Run mypy
+# We specify Linux so that we get the same results on all platforms. Without
+# that, mypy would complain on macOS that epoll does not exist, and we can't
+# ignore it as it would cause a mypy error on Linux
+mypy -p trio._core --platform linux || EXIT_STATUS=$?
+
 # Finally, leave a really clear warning of any issues and exit
 if [ $EXIT_STATUS -ne 0 ]; then
     cat <<EOF
--- a/trio/_core/__init__.py
+++ b/trio/_core/__init__.py
@@ -4,7 +4,6 @@ and deal with private internal data structures. Things in this namespace
 are publicly available in either trio, trio.lowlevel, or trio.testing.
 """
 
-import typing as _t
 import sys
 
 from ._exceptions import (
@@ -76,13 +75,6 @@ from ._thread_cache import start_thread_soon
 
 from ._mock_clock import MockClock
 
-# Kqueue imports
-if not _t.TYPE_CHECKING:  # pragma: no branch
-    try:
-        from ._run import current_kqueue, monitor_kevent, wait_kevent
-    except ImportError:
-        pass
-
 # Windows imports
 if sys.platform == "win32":
     from ._run import (
@@ -93,3 +85,6 @@ if sys.platform == "win32":
         write_overlapped,
         readinto_overlapped,
     )
+# Kqueue imports
+elif sys.platform != "linux" and sys.platform != "win32":
+    from ._run import current_kqueue, monitor_kevent, wait_kevent
--- a/trio/_core/__init__.py
+++ b/trio/_core/__init__.py
@@ -88,3 +88,5 @@ if sys.platform == "win32":
 # Kqueue imports
 elif sys.platform != "linux" and sys.platform != "win32":
     from ._run import current_kqueue, monitor_kevent, wait_kevent
+
+del sys   # It would be better to import sys as _sys, but mypy does not understand it
--- a/trio/_core/__init__.py
+++ b/trio/_core/__init__.py
@@ -77,11 +77,7 @@ from ._thread_cache import start_thread_soon
 from ._mock_clock import MockClock
 
 # Kqueue imports
-if (
-    sys.platform.startswith("freebsd")
-    or sys.platform.startswith("darwin")
-    or not _t.TYPE_CHECKING
-):
+if not _t.TYPE_CHECKING:  # pragma: no branch
     try:
         from ._run import current_kqueue, monitor_kevent, wait_kevent
     except ImportError:
--- a/trio/_core/__init__.py
+++ b/trio/_core/__init__.py
@@ -4,6 +4,9 @@ and deal with private internal data structures. Things in this namespace
 are publicly available in either trio, trio.lowlevel, or trio.testing.
 """
 
+import typing as _t
+import sys
+
 from ._exceptions import (
     TrioInternalError,
     RunFinishedError,
@@ -74,13 +77,18 @@ from ._thread_cache import start_thread_soon
 from ._mock_clock import MockClock
 
 # Kqueue imports
-try:
-    from ._run import current_kqueue, monitor_kevent, wait_kevent
-except ImportError:
-    pass
+if (
+    sys.platform.startswith("freebsd")
+    or sys.platform.startswith("darwin")
+    or not _t.TYPE_CHECKING
+):
+    try:
+        from ._run import current_kqueue, monitor_kevent, wait_kevent
+    except ImportError:
+        pass
 
 # Windows imports
-try:
+if sys.platform == "win32":
     from ._run import (
         monitor_completion_key,
         current_iocp,
@@ -89,5 +97,3 @@ try:
         write_overlapped,
         readinto_overlapped,
     )
-except ImportError:
-    pass
--- a/trio/_core/__init__.py
+++ b/trio/_core/__init__.py
@@ -89,4 +89,4 @@ if sys.platform == "win32":
 elif sys.platform != "linux" and sys.platform != "win32":
     from ._run import current_kqueue, monitor_kevent, wait_kevent
 
-del sys   # It would be better to import sys as _sys, but mypy does not understand it
+del sys  # It would be better to import sys as _sys, but mypy does not understand it
--- a/trio/_core/_io_epoll.py
+++ b/trio/_core/_io_epoll.py
@@ -1,6 +1,7 @@
 import select
 import attr
 from collections import defaultdict
+from typing import DefaultDict
 
 from .. import _core
 from ._run import _public
@@ -184,8 +185,10 @@ class EpollWaiters:
 class EpollIOManager:
     _epoll = attr.ib(factory=select.epoll)
     # {fd: EpollWaiters}
-    _registered = attr.ib(factory=lambda: defaultdict(EpollWaiters))
-    _force_wakeup = attr.ib(factory=WakeupSocketpair)
+    _registered = attr.ib(
+        factory=lambda: defaultdict(EpollWaiters), type=DefaultDict[int, EpollWaiters]
+    )
+    _force_wakeup = attr.ib(factory=WakeupSocketpair, type=WakeupSocketpair)
     _force_wakeup_fd = attr.ib(default=None)
 
     def __attrs_post_init__(self):
--- a/trio/_core/_io_epoll.py
+++ b/trio/_core/_io_epoll.py
@@ -188,7 +188,7 @@ class EpollIOManager:
     _registered = attr.ib(
         factory=lambda: defaultdict(EpollWaiters), type=Dict[int, EpollWaiters]
     )
-    _force_wakeup = attr.ib(factory=WakeupSocketpair, type=WakeupSocketpair)
+    _force_wakeup = attr.ib(factory=WakeupSocketpair)
     _force_wakeup_fd = attr.ib(default=None)
 
     def __attrs_post_init__(self):
--- a/trio/_core/_io_epoll.py
+++ b/trio/_core/_io_epoll.py
@@ -1,7 +1,7 @@
 import select
 import attr
 from collections import defaultdict
-from typing import DefaultDict
+from typing import Dict
 
 from .. import _core
 from ._run import _public
@@ -186,7 +186,7 @@ class EpollIOManager:
     _epoll = attr.ib(factory=select.epoll)
     # {fd: EpollWaiters}
     _registered = attr.ib(
-        factory=lambda: defaultdict(EpollWaiters), type=DefaultDict[int, EpollWaiters]
+        factory=lambda: defaultdict(EpollWaiters), type=Dict[int, EpollWaiters]
     )
     _force_wakeup = attr.ib(factory=WakeupSocketpair, type=WakeupSocketpair)
     _force_wakeup_fd = attr.ib(default=None)
--- a/trio/_core/_io_kqueue.py
+++ b/trio/_core/_io_kqueue.py
@@ -19,7 +19,7 @@ class _KqueueStatistics:
 
 @attr.s(slots=True, eq=False)
 class KqueueIOManager:
-    _kqueue = attr.ib(factory=select.kqueue)
+    _kqueue = attr.ib(factory=select.kqueue)  # type: ignore
     # {(ident, filter): Task or UnboundedQueue}
     _registered = attr.ib(factory=dict)
     _force_wakeup = attr.ib(factory=WakeupSocketpair)
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -415,7 +415,7 @@ def traceback_exception_init(
         self.embedded = []
 
 
-traceback.TracebackException.__init__ = traceback_exception_init
+traceback.TracebackException.__init__ = traceback_exception_init  # type: ignore
 traceback_exception_original_format = traceback.TracebackException.format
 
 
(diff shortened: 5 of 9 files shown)


trio/_core/_io_kqueue.py
```diff
@@ -19,7 +19,7 @@ class _KqueueStatistics:
 
 @attr.s(slots=True, eq=False)
 class KqueueIOManager:
-    _kqueue = attr.ib(factory=select.kqueue)
+    _kqueue = attr.ib(factory=select.kqueue)  # type: ignore
```
pquentin, line 22: This works because we require Linux when running mypy. But if I don't specify the platform, I get the following errors on macOS:

```
trio/_core/_io_kqueue.py:22: error: unused 'type: ignore' comment
trio/_core/_io_epoll.py:186: error: Module has no attribute "epoll"; maybe "poll"?
```

I plan to fix that in a further pull request by adding `assert sys.platform.startswith("freebsd") or sys.platform.startswith("darwin")` at the top of the file and running mypy on the various platforms we support.

trio/_core/_run.py
```diff
@@ -1150,7 +1151,7 @@ class GuestState:
     run_sync_soon_not_threadsafe = attr.ib()
     done_callback = attr.ib()
     unrolled_run_gen = attr.ib()
-    unrolled_run_next_send = attr.ib(factory=lambda: Value(None))
+    unrolled_run_next_send = attr.ib(factory=lambda: Value(None), type=object)
```
pquentin, line 1154: I don't understand why I had to do this. 🤷

njsmith: I guess the type is actually `Outcome`?

pquentin: I get `trio/_core/_run.py:1154: error: Returning Any from function declared to return "object"` when trying anything else than object. I guess I need to understand how attrs typing support works.

pquentin:
https://travis-ci.org/github/python-trio/trio/jobs/697497032 confirms this happens in test_warn_on_failed_cancel_terminate as Travis is much better at showing output while the tests are still running.

The relevant stacktrace:

```
Thread 0x00007ff41382a740 (most recent call first):
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/trio/_core/_io_epoll.py", line 223 in get_events
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/trio/_core/_run.py", line 1889 in run
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/trio/testing/_trio_test.py", line 27 in wrapper
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/python.py", line 182 in pytest_pyfunc_call
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 87 in <lambda>
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/python.py", line 1477 in runtest
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/runner.py", line 135 in pytest_runtest_call
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 87 in <lambda>
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/runner.py", line 217 in <lambda>
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/runner.py", line 244 in from_call
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/runner.py", line 217 in call_runtest_hook
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/runner.py", line 186 in call_and_report
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/runner.py", line 100 in runtestprotocol
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/runner.py", line 85 in pytest_runtest_protocol
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 87 in <lambda>
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/main.py", line 272 in pytest_runtestloop
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 87 in <lambda>
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/main.py", line 247 in _main
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/main.py", line 191 in wrap_session
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/main.py", line 240 in pytest_cmdline_main
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 87 in <lambda>
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/site-packages/_pytest/config/__init__.py", line 125 in main
  File "/home/travis/virtualenv/python3.6-dev/bin/pytest", line 8 in <module>
```

pquentin:
Closing/reopening because the FreeBSD build failed with #1604

pquentin:
Coverage is OK, next up is understanding the unrolled_run_next_send issue.

pquentin:
`unrolled_run_next_send` fixed, please take another look.

trio/_core/_run.py
```diff
@@ -1151,7 +1151,8 @@ class GuestState:
     run_sync_soon_not_threadsafe = attr.ib()
     done_callback = attr.ib()
     unrolled_run_gen = attr.ib()
-    unrolled_run_next_send = attr.ib(factory=lambda: Value(None), type=object)
+    _value_factory: Callable[[], Value] = lambda: Value(None)
+    unrolled_run_next_send = attr.ib(factory=_value_factory, type=Value)
```
njsmith, line 1228: I think the type here should be `outcome.Outcome`, not `outcome.Value`? E.g. if `get_events` raises an exception, then `unrolled_run_next_send` will be an `outcome.Error`.

trio/_core/__init__.py
```diff
@@ -4,6 +4,9 @@
 are publicly available in either trio, trio.lowlevel, or trio.testing.
 """
 
+import typing as _t
+import sys
```
njsmith, line 8: I guess technically this should be `import sys as _sys`, though in practice it doesn't do much harm here...

trio/_core/_run.py
pquentin: Thanks! Fixed.

trio/_core/_multierror.py
```diff
@@ -415,7 +415,7 @@ def traceback_exception_init(
         self.embedded = []
 
 
-traceback.TracebackException.__init__ = traceback_exception_init
+traceback.TracebackException.__init__ = traceback_exception_init  # type: ignore
```
njsmith, line 418: I'm not too worried about `# type: ignore` comments inside this horrible blob of monkeypatching, but I am curious why you need them.

trio/_core/_run.py
```diff
@@ -2374,13 +2376,17 @@ async def checkpoint_if_cancelled():
...
-elif hasattr(select, "epoll"):
+elif sys.platform == "linux" or (not TYPE_CHECKING and hasattr(select, "epoll")):
     from ._io_epoll import EpollIOManager as TheIOManager
     from ._generated_io_epoll import *
-elif hasattr(select, "kqueue"):
+elif (
+    sys.platform == "darwin"
+    or sys.platform.startswith("freebsd")
+    or (not TYPE_CHECKING and hasattr(select, "epoll"))
+):
```
njsmith, line 2389: Well... this check is gross and we should probably complain about it to typing-sig.

But having written it, I guess we might as well copy-paste it into `trio/_core/__init__.py` too?

njsmith: Oh heh and there's a copy/paste error here – you accidentally replaced `hasattr(select, "kqueue")` with `hasattr(select, "epoll")`.

trio/_core/_multierror.py
pquentin: The error is `trio/_core/_multierror.py:418: error: Cannot assign to a method`. It appears to be a known mypy bug: https://github.com/python/mypy/issues/2427. It's unclear to me if `setattr` is a better workaround or not.

trio/_core/_run.py
pquentin: > Well... this check is gross and we should probably complain about it to typing-sig.

Do you mean that mypy should be able to tell wich platforms support epoll/kqueue?

> But having written it, I guess we might as well copy-paste it into `trio/_core/__init__.py` too?

Do you mean that this hack is better than the current `try/except ImportError` way in `trio/_core/__init__.py`?

> Oh heh and there's a copy/paste error here – you accidentally replaced `hasattr(select, "kqueue")` with `hasattr(select, "epoll")`.

Thanks, fixed.

trio/_core/__init__.py
pquentin: Unfortunately, importing `sys` as `_sys` breaks mypy...

trio/_core/_run.py
njsmith: > Do you mean that mypy should be able to tell wich platforms support epoll/kqueue?

Maybe? I mean that having to copy-paste this complicated and not-even-terribly-correct code into multiple places is suboptimal, and maybe it could be avoided somehow. But yeah, teaching mypy to evaluate `hasattr` statements seems like one good way to do that... it already has a model of which attributes the `select` module has!

And heh, on further investigation, it looks like the way mypy's understanding of `select` works is:

```python
if sys.platform != "linux" and sys.platform != "win32":
```

https://github.com/python/typeshed/blob/66a9a4b5ce72aaf022b24706bf58254602ef02f8/stdlib/2and3/select.pyi#L36-L38

That is probably nicer than trying to enumerate all the BSD variations... so I guess maybe for now we should do:

```python
if sys.platform == "win32":
    ...
elif sys.platform == "linux" or hasattr(select, "epoll"):
    ...
elif TYPE_CHECKING or hasattr(select, "kqueue"):
    ...
else:
    raise NotImplementedError
```

> Do you mean that this hack is better than the current try/except ImportError way in trio/_core/__init__.py?

Yeah, b/c the way that's written right now, if you try to run mypy on macOS then it'll complain that `current_kqueue` and friends are undefined.

pquentin: Thanks, done. (Btw, note that you can use "Quote reply" to preserve the markdown formatting of the text you're quoting.)

check.sh
```diff
@@ -23,6 +23,12 @@ flake8 trio/ \
     --ignore=D,E,W,F401,F403,F405,F821,F822\
     || EXIT_STATUS=$?
 
+# Run mypy
+# We specify Linux so that we get the same results on all platforms. Without
```
oremanj, line 27: Can you edit this comment to mention your future plans listed below? (run mypy on all platforms, use assert guards to avoid typechecking the wrong IOManagers)

trio/_core/__init__.py
oremanj: `del sys` at the end will clean up sufficiently.

Yeah, mypy's platform guards are rather brittle. It's strange since, for example, the magic `from typing import TYPE_CHECKING` seems to work under any name...

trio/_core/_io_epoll.py
```diff
@@ -1,6 +1,7 @@
 import select
 import attr
 from collections import defaultdict
+from typing import DefaultDict
```
oremanj, line 4: I generally just use `Dict` for defaultdicts, since you rarely wind up accessing the `.factory` member

trio/_core/_io_epoll.py
```diff
@@ -184,8 +185,10 @@ class EpollWaiters:
 class EpollIOManager:
     _epoll = attr.ib(factory=select.epoll)
     # {fd: EpollWaiters}
-    _registered = attr.ib(factory=lambda: defaultdict(EpollWaiters))
-    _force_wakeup = attr.ib(factory=WakeupSocketpair)
+    _registered = attr.ib(
+        factory=lambda: defaultdict(EpollWaiters), type=DefaultDict[int, EpollWaiters]
+    )
+    _force_wakeup = attr.ib(factory=WakeupSocketpair, type=WakeupSocketpair)
```
oremanj, line 191: Do you need the type= here? I would expect mypy to be able to infer it from the factory return type.

trio/_core/_run.py
```diff
@@ -1223,7 +1224,8 @@ class GuestState:
     run_sync_soon_not_threadsafe = attr.ib()
     done_callback = attr.ib()
     unrolled_run_gen = attr.ib()
-    unrolled_run_next_send = attr.ib(factory=lambda: Value(None))
+    _value_factory: Callable[[], Value] = lambda: Value(None)
+    unrolled_run_next_send = attr.ib(factory=_value_factory, type=Outcome)
```
oremanj, line 1228: You can use `cast()` to write this without introducing an additional class-level name. (`cast(Callable[[], Value], lambda: Value(None))`). I'm also curious what breaks if you write it on one line without the cast.

trio/_core/_run.py
oremanj: Any arm of the if that *might* be true (according to mypy's very limited static analysis) *must* type-check. That's the reason for all the `not TYPE_CHECKING and` in the original -- it basically is saying "I know you can't tell this hasattr() result when type-checking, so just ignore it".

njsmith: So I guess that means we should do something like this?

```python
if sys.platform == "win32":
    ...
elif sys.platform == "linux" or (not TYPE_CHECKING and hasattr(select, "epoll")):
    ...
elif TYPE_CHECKING or (not TYPE_CHECKING and hasattr(select, "kqueue")):
    ...
else:
    raise NotImplementedError
```

?

njsmith: Also raised this upstream here: https://github.com/python/mypy/issues/9042

oremanj: I don't think you need the `not TYPE_CHECKING` guard in the kqueue arm, because mypy can tell that `TYPE_CHECKING or ...` is true regardless of the `...`.

trio/_core/_run.py
pquentin: The error message without the lambda type hint is `Returning Any from function declared to return "object"`. Indeed according to `reveal_type` the type of `lambda: Value(None)` is `'def () -> Any'`.

So far so good, but if I switch to your inline cast which specifies the `Value` return type, I get `trio/_core/_run.py:1227: error: Redundant cast to "Callable[[], Any]"`! It's maybe a variance issue?

trio/_core/_io_epoll.py
pquentin: You're right, that was not needed! Thanks, fixed.

trio/_core/_io_epoll.py
pquentin: Switched to `Dict`

check.sh
pquentin: Thanks, done

trio/_core/_run.py
pquentin: Thanks, switched to this. Will probably help with multiplatform support even if right now I get the exact same errors

trio/_core/__init__.py
pquentin: Thanks, added `del sys`

trio/_core/_run.py
oremanj: Oh, this is probably because `outcome` is not typed, so mypy is replacing everything imported from it with `Any`.

We should add type hints to `outcome`. In the meantime, your workaround here seems fine, just note that we're not actually getting anything from declaring the type as `Outcome` as opposed to `Any`.

pquentin:
Thanks for the reviews! I adressed all comments, [lease take another look.

PR #1616, opened 2020-06-15
Update Google docstring format example link

altendky:


altendky:
I chose 3.x because it seemed closest to the existing 'stable' link's choice given that the next step was master instead of stable.

PR #1629, opened 2020-06-20
Switch to heap-based deadline tracking

njsmith:
Continuation of @Tronic's work in #1351

Main changes:

- Rebased on top of master, resolving a number of conflicts
- Reworked the heap management code in the `Deadlines` class, in particular to ignore defunct deadlines when computing the `next_deadline`.

Using the microbenchmark here: https://github.com/python-trio/trio/pull/1351#issuecomment-581390193

On my laptop I get:

**Current master**

```
❯ ./wrk http://localhost:5000 --latency
Running 10s test @ http://localhost:5000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   657.76us  284.24us  12.20ms   90.78%
    Req/Sec     7.74k     1.11k    8.86k    80.20%
  Latency Distribution
     50%  616.00us
     75%  734.00us
     90%    0.85ms
     99%    1.28ms
  155534 requests in 10.10s, 10.09MB read
Requests/sec:  15399.87
Transfer/sec:      1.00MB
```

**This PR**

```
❯ ./wrk http://localhost:5000 --latency
Running 10s test @ http://localhost:5000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   470.66us  116.54us   4.66ms   74.25%
    Req/Sec    10.64k     1.14k   21.23k    91.54%
  Latency Distribution
     50%  454.00us
     75%  524.00us
     90%  610.00us
     99%  792.00us
  212774 requests in 10.10s, 13.80MB read
Requests/sec:  21068.22
Transfer/sec:      1.37MB
```

**This PR + a stubbed-out `Deadlines` class** (i.e., this is what we'd
get if our deadline tracking heap somehow managed to perform all
operations in zero time)

<details>

```python
class Deadlines:
    def add(self, deadline, cancel_scope):
        pass

    def remove(self, deadline, cancel_scope):
        pass

    def next_deadline(self):
        return inf

    def expire(self, now):
        return False
```

</details>

```
❯ ./wrk http://localhost:5000 --latency
Running 10s test @ http://localhost:5000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   478.42us  274.25us   9.03ms   97.35%
    Req/Sec    10.72k     1.48k   21.79k    85.57%
  Latency Distribution
     50%  445.00us
     75%  527.00us
     90%  610.00us
     99%    0.86ms
  214433 requests in 10.10s, 13.91MB read
Requests/sec:  21231.28
Transfer/sec:      1.38MB
```

These benchmarks aren't super stable, because a laptop is a terrible benchmarking environment (thermal issues, etc.). But it seems fair to say that this gives roughly a 35% increase in throughput + improves latency, and that there's not a lot of headroom left for further algorithmic improvements. (At least until we make everything else faster!)

Co-authored-by: L. Kärkkäinen <tronic@users.noreply.github.com>

### DIFF
--- a/newsfragments/1629.feature.rst
+++ b/newsfragments/1629.feature.rst
@@ -1,0 +1,2 @@
+We switched to a new, lower-overhead data structure to track upcoming
+timeouts, which should make your programs faster.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -45,6 +45,8 @@ from .. import _core
 from .._deprecate import warn_deprecated
 from .._util import Final, NoPublicConstructor, coroutine_or_error
 
+DEADLINE_HEAP_MIN_PRUNE_THRESHOLD = 1000
+
 _NO_SEND = object()
 
 
@@ -155,6 +157,27 @@ class Deadlines:
                 heappop(self._heap)
         return inf
 
+    def _prune(self):
+        # In principle, it's possible for a cancel scope to toggle back and
+        # forth repeatedly between the same two deadlines, and end up with
+        # lots of stale entries that *look* like they're still active, because
+        # their deadline is correct, but in fact are redundant. So when
+        # pruning we have to eliminate entries with the wrong deadline, *and*
+        # eliminate duplicates.
+        seen = set()
+        pruned_heap = []
+        for deadline, tiebreaker, cancel_scope in self._heap:
+            if deadline == cancel_scope._registered_deadline:
+                if cancel_scope in seen:
+                    continue
+                seen.add(cancel_scope)
+                pruned_heap.append((deadline, tiebreaker, cancel_scope))
+        # See test_cancel_scope_deadline_duplicates for a test that exercises
+        # this assert:
+        assert len(pruned_heap) == self._active
+        heapify(pruned_heap)
+        self._heap = pruned_heap
+
     def expire(self, now):
         did_something = False
         while self._heap and self._heap[0][0] <= now:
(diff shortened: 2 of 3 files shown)


njsmith:
So there is one potential flaw I noticed in this; I'm not sure if we should care about it or not. Consider some pathological code that repeatedly switches a deadline back and forth between two distinct values:

```python
with CancelScope() as cscope:
    for _ in range(100000):
        cscope.deadline = 100
        cscope.deadline = 101
```

This creates tons and tons of entries in our heap, most of them stale. Eventually this will trigger a heap prune. But the prune will end up leaving a bunch of the stale entries in-place, because they all have a `deadline` that matches the current `cscope._registered_deadline`. And then we enter a pathological state, where we keep attempting to prune on every scheduler tick, while accomplishing nothing, creating a nasty slowdown.

Now that I write this, it occurs to me that there's a pretty trivial solution: have the prune operation keep a set of cancel scope objects it's seen, and drop duplicates. That adds a little bit of overhead to the prune operation, but it shouldn't be too bad.

njsmith:
> Now that I write this, it occurs to me that there's a pretty trivial solution: have the prune operation keep a set of cancel scope objects it's seen, and drop duplicates. That adds a little bit of overhead to the prune operation, but it shouldn't be too bad.

Done.

PR #1632, opened 2020-06-21
Start an RPC list

goodboy:
As discussed briefly with @njsmith in chat.

These are the only two projects I know of presently but I'm happy to dig for some more before this is merged.


njsmith:
> I'm happy to dig for some more before this is merged.

The list is a resource, not a definitive literature review; it's OK if it's incomplete :-) We can always add more later.

PR #1634, opened 2020-06-22
Be more confident when introducing ourselves

pquentin:
Those words were written in 2017 when Trio *was* experimental.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,8 +39,7 @@ Trio – a friendly Python library for async concurrency and I/O
    :width: 200px
    :align: right
 
-The Trio project's goal is to produce a production-quality,
-`permissively licensed
+Trio is a production-quality, `permissively licensed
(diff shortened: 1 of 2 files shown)


njsmith:
Hmm. You're absolutely right that this text doesn't do a great job of communicating our current status and needs to be revisited, but I'm not sure just removing the qualifiers is the right approach... we're still quite aggressive about deprecations, and we know we have some big breaking changes like #1208, #611, etc. still to come before we can declare any kind of 1.0 stability guarantee. I'll leave this open for now and try to do my own pass at it this week.

pquentin:
My reasoning was that the release notes + deprecation warnings do a good enough job of replacing #1. But yeah, we should also mention that we're not 1.0 yet and still agressive about deprecations, but that we work hard to make those deprecations easy to handle.

PR #1639, opened 2020-06-22
Don't hold a reference to the last  job in the thread cache

### LINKED ISSUE #1638: to_thread.run_sync() foils meta-class destructors
Invoking `to_thread.run_sync()` seems to prevent meta-class destructors of completely unrelated classes from being called on shutdown.

```python
import trio

class deltype(type):
    def __del__(self):
        print(f"{self.__name__} class deleted")

Test = deltype("Test", (object,), {})

async def async_main():
    await trio.to_thread.run_sync(lambda: print('hi'))
    await trio.sleep(0)

trio.run(async_main)
```

expected output:
```
hi
Test class deleted
```

actual output:
```
hi
```

if the `run_sync()` is commented out, the meta-class destructor is called as expected:
```
Test class deleted
```

smurfix:
Fixes #1638

### DIFF
--- a/trio/_core/tests/test_thread_cache.py
+++ b/trio/_core/tests/test_thread_cache.py
@@ -4,7 +4,7 @@ from queue import Queue
 import time
 import sys
 
-from .tutil import slow
+from .tutil import slow, gc_collect_harder
 from .. import _thread_cache
 from .._thread_cache import start_thread_soon, ThreadCache
 
@@ -23,6 +23,29 @@ def test_thread_cache_basics():
     outcome = q.get()
     with pytest.raises(RuntimeError, match="hi"):
         outcome.unwrap()
+
+
+def test_thread_cache_deref():
+    res = [False]
+
+    class del_me:
+        def __call__(self):
+            return 42
+
+        def __del__(self):
+            res[0] = True
+
+    q = Queue()
+
+    def deliver(outcome):
+        q.put(outcome)
+
+    start_thread_soon(del_me(), deliver)
+    outcome = q.get()
+    assert outcome.unwrap() == 42
+
+    gc_collect_harder()
+    assert res[0]
 
 
 @slow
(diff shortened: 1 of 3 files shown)


smurfix:
Yes. I adapted the `test_thread_cache_basics` test:
```
def test_thread_cache_deref():
    res = [False]

    def test_main():
        class deltype(type):
            def __del__(self):
                res[0] = True

        Test = deltype("Test", (object,), {})
        def fn():
            return 42

        q = Queue()
        def deliver(outcome):
            q.put(outcome)

        start_thread_soon(fn, deliver)

        outcome = q.get()
        assert outcome.unwrap() == 42
        assert not res[0]
    test_main()
    gc_collect_harder()
    assert res[0]
```
which unfortunately fails to trigger the problem in the unpatched case.

njsmith:
CI failure is the mysterious #1604, nothing related to this.

Nice catch, and we definitely want this. Can you add a test and newsfragment to the PR, though?

trio/_core/tests/test_thread_cache.py
```diff
@@ -25,6 +25,29 @@ def deliver(outcome):
...
 
+def test_thread_cache_deref():
+    res = [False]
+
+    class del_me:
+        def __call__(self):
+            return 42
+
+        def __del__(self):
+            res[0] = True
```
belm0, line 36: so this is the instance's destructor, where my original report was on the class destructor

but I see that the bug indeed affects both of these.  (I suppose that not destroying the instance is enough to keep a reference to the class?)

does this variant of the test fail before the patch?

belm0: > does this variant of the test fail before the patch?

I see it does 👍

smurfix: I couldn't easily reproduce the problem using a class destructor, so I opted for a minimal test that shows the reference to be gone / some destructor to be called.

PR #1641, opened 2020-06-23
Remove docs check

pquentin:
We're now testing docs via the ReadTheDocs pull request integration.


pquentin:
I used my superpowers to stop requiring the docs check.

PR #1644, opened 2020-06-24
Add local_address= kwarg to open_tcp_stream

njsmith:
Fixes gh-275

### DIFF
--- a/newsfragments/275.feature.rst
+++ b/newsfragments/275.feature.rst
@@ -1,0 +1,3 @@
+`trio.open_tcp_stream` has a new ``local_address=`` keyword argument
+that can be used on machines with multiple IP addresses to control
+which IP is used for the outgoing connection.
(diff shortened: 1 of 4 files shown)


oremanj:
Thanks for the detailed comment on `bind()` -- I learned something!

ntninja:
That was way too fast for me…! :open_mouth: 

Do I see it correctly that with this API limiting a connection to `socket.AF_INET`/`socket.AF_INET6` address families during name resolution is done by passing `"0.0.0.0"`/`"::"` respectively for the `local_address=…` arg?

njsmith:
@ntninja Currently, `open_tcp_stream` always does full name resolution, but if you pass `local_address="0.0.0.0"` then it will only skip trying to connect to the `AF_INET6` addresses and if you pass `local_address="::"` then it will skip trying to connect to the `AF_INET` addresses. Technically this is very slightly inefficient (`getaddrinfo` still has to look up both sets of addresses), but this is very minor, and in return, we get better error messages: e.g., if you accidentally try to connect to an IPv6-only host with `local_address="0.0.0.0"`, then it will fail, but the error message will tell you that there *were* IPv6 addresses it considered connecting to and if you changed your `local_address=` you might succeed.

ntninja:
Ok, thanks! I don't have any strong feelings on this, just wanted to know for sure!

PR #1645, opened 2020-06-25
Run mypy on all platforms

pquentin:


check.sh
```diff
@@ -23,13 +23,10 @@ flake8 trio/ \
...
-# Run mypy
-# We specify Linux so that we get the same results on all platforms. Without
-# that, mypy would complain on macOS that epoll does not exist, and we can't
-# ignore it as it would cause a mypy error on Linux
-# In the future, we will run mypy on all platforms and use platform assert
-# guards to avoid typechecking the wrong IOManagers
+# Run mypy on all supported platforms
 mypy -p trio._core --platform linux || EXIT_STATUS=$?
+mypy -p trio._core --platform macos || EXIT_STATUS=$?  # tests FreeBSD too
+mypy -p trio._core --platform windows || EXIT_STATUS=$?
```
njsmith, line 29: I think the `sys.platform` strings are `darwin` and `win32`?

trio/_core/_io_epoll.py
```diff
@@ -8,6 +9,8 @@
 from ._io_common import wake_all
 from ._wakeup_socketpair import WakeupSocketpair
 
+assert sys.platform == "linux"
```
njsmith, line 12: Maybe we should put these inside an `if TYPE_CHECKING:`? Both to make it clear that the purpose of this `assert` is  for the benefit of type-checkers, and because we've generally avoided doing actual `sys.platform` checks at runtime.

pquentin: If I put the asserts in `if TYPE_CHECKING:`, mypy no longer recognizes them!

check.sh
pquentin: Thanks, fixed!

oremanj:
I mentioned this on gitter, but putting it here for the record:

I'm leery of keeping the assertions if we can't guard them under `if TYPE_CHECKING:`. For example, illumos supports epoll, and this would be unnecessarily breaking it (although it's quite possibly already broken - not officially supported, at minimum).

I think the problem here is that `mypy -p` recursively checks all submodules of the package you specify, even if they're not reachable through any public names. For example, it will check `trio._core._io_epoll` on all platforms, even though it's only imported in `_run` on Linux.

Possible workaround: use `mypy -m trio._core` instead of `mypy -p trio._core`. (When we expand to cover all of Trio, we'll need `-m trio -m trio.testing` to cover all of Trio's publicly importable modules that don't automatically come along for the ride when you `import trio`.)

If you do this, are you able to remove the platform assertions?

trio/_core/_io_epoll.py
pquentin: Which is bad. We've talked about this with @oremanj in chat: https://gitter.im/python-trio/general?at=5ef98c54e0e5673398df291c

oremanj:
Turns out we can't do that because we have `follow_imports = silent` in our mypy.ini, and we can't remove that until we make the rest of Trio mypy-clean.

Let's go ahead with the platform assertions as a temporary stopgap. Once we have the rest of trio passing mypy, we'll switch to `mypy -m` and clean it up.

PR #1648, opened 2020-06-26
Organize instruments for reduced overhead

### LINKED ISSUE #1340: Efficient instrument implementation
[From this comment](https://trio.discourse.group/t/eliot-the-causal-logging-library-now-supports-trio/167/14): 
> It’s true that the current implementation still adds unnecessary overhead for operations you haven’t hooked, but that’s just because we’re lazy and were waiting for someone to complain.

I'm not sure what's involved here but I wouldn't want a potentially significant improvement to get lost in a comment. :-)

oremanj:
Fixes #1340.


trio/_core/tests/test_instrumentation.py
```diff
@@ -0,0 +1,266 @@
...
+    assert ("after", main_task) in r.record
+    assert ("after_run",) in r.record
+    # And we got a log message
+    exc_type, exc_value, exc_traceback = caplog.records[0].exc_info
+    assert exc_type is ValueError
+    assert str(exc_value) == "oops"
+    assert "Instrument has been disabled" in caplog.records[0].message
+
+
+def test_instruments_monkeypatch():
```
oremanj, line 204: Above this point are copied unchanged from test_run. Below this point are new tests.

njsmith:
This is really cool!

On a quick skim, it took me longer to figure out what the machinery was doing. I guess I would probably have gone for something more "elementary", like:

```python
# Invariant: all entries in this dict should be non-empty
instruments_by_method: Dict[str, Set[Instrument]]

def add_instrument(self, instrument):
    for method in ALL_INSTRUMENT_METHODS:
        if has_nontrivial_method(instrument, method):
            self.instruments_by_method.setdefault(method, set()).add(instrument)

def remove_instrument(self, instrument):
    for method, instruments in list(self.instruments_by_method.items()):
        instruments.discard(instrument)
        if not instruments:
            del self.instruments_by_method[method]

def call_instruments(self, method, *args, **kwargs):
    instruments = self.instruments_by_method.get(method)
    # Caller should check that key exists before calling,
    # and our invariant says that if the key exists, the value should be non-empty.
    assert instruments
    for instrument in list(instruments):
        try:
            getattr(instrument, method)(*args, **kwargs)
        except:
            ...

# Usage
if "task_scheduled" in self.instruments_by_method:
    self.call_instruments("task_scheduled", ...)
```

What do you think?

oremanj:
The biggest difference I see is that my approach allows type checking of instrument calls (because they look to the type checker like method calls, not strings). This is not *that* big an advantage, but I do prefer it somewhat.

Other differences:
* Your suggestion implements instrument removal as "try removing it from all the per-method collections"; the code in the PR remembers which methods each instrument implements. It's not possible to raise KeyError on a missing instrument when using your approach (because an instrument that implements no hooks gets recorded nowhere), which would be a backwards-compatibility break, albeit a tiny one that no one is likely to care about.
* You use sets for the per-method collections, while I use dicts. Using a dict makes instrument invocation order consistent, which seems worthwhile, and provides a convenient place to cache the method lookup, which might be worth a bit of performance (though I haven't tested).
* You iterate over a copy of the per-method collection, while I use a copy-on-write approach. It looks like the break-even point in performance between the two is at about two instruments per hook, so I'm happy to switch back to iterate-over-a-copy.
* I split the instrumentation logic out from `_run` into its own module. I think this is valuable since it's easy enough to do. `_run` is getting pretty hefty!

njsmith:
> The biggest difference I see is that my approach allows type checking of instrument calls (because they look to the type checker like method calls, not strings). This is not that big an advantage, but I do prefer it somewhat.

Hmm, I see. I was thinking that they were already getting type-checked by the `add_instrument(instrument: Instrument)` annotation, but I guess that just checks that the *user* methods have the correct types – what you're getting by subclassing `Instrument` is static checks of *Trio's* code that calls the instrumentation hooks. I guess I'm not terribly worried about us getting this wrong (we have actual tests, not just type checks :-)), and I've been thinking that most of the value for type-checking is for users.

> Your suggestion implements instrument removal as "try removing it from all the per-method collections"; the code in the PR remembers which methods each instrument implements. It's not possible to raise KeyError on a missing instrument when using your approach (because an instrument that implements no hooks gets recorded nowhere), which would be a backwards-compatibility break, albeit a tiny one that no one is likely to care about.

Fair. I guess fixing this would only require adding `self.instruments.add(instrument)` to `add_instrument` and `self.instruments.remove(instrument)` to `remove_instrument.

> You use sets for the per-method collections, while I use dicts. Using a dict makes instrument invocation order consistent, which seems worthwhile, and provides a convenient place to cache the method lookup, which might be worth a bit of performance (though I haven't tested).

Yeah, I was wondering about the consistent order thing. On the one hand, predictability is good. On the other, if we want to make guarantees about order, then "run all hooks from first-to-last" is kind of a weird guarantee to make, like... I'd expect that the newest instrument has its `before_task` hook called first and its `after_task` hook called last.

This is pretty orthogonal to the basic question though; we could use dicts in my approach too :-). The big thing I'm hesitating over is the magic with auto-generated slots and all that.

(I can't imagine the method lookup is a significant cost.)

> I split the instrumentation logic out from `_run` into its own module. I think this is valuable since it's easy enough to do. `_run` is getting pretty hefty!

Yeah, for sure. Unfortunately, it's hard to split up, because everything needs knowledge of everything else's internals :-). And that's true here too, since the whole point of this is that we want to inline cheap "is there an instrument registered for this method" checks everywhere...

I guess all the trade-offs here are about making the code easier to understand and get confidence that it's correct. And I think on net for me, this approach is harder to understand and be confident is correct, due to the dynamically generated `__slots__`, the tricky re-use of `trio.abc.Instrument` for something that's not semantically an `Instrument`, the extra mapping from instruments -> hooks, etc. Splitting into a separate file and extra type checks are good, but their benefits are pretty shallow compared to simpler logic.

TBH I could live with either approach though, and maybe this isn't worth overthinking.

oremanj:
I am convinced by your argument (we don't need type-checking internally that much, we have tests!) and have verified that doing these simplifications doesn't impact performance notably.

CPython 3.8.0:
```
(master) $ python t.py
        no instrument: 4.23 ± 0.044 usec per checkpoint
      noop instrument: 7.06 ± 0.12 usec per checkpoint
 task_step instrument: 6.81 ± 0.097 usec per checkpoint

(elide-instrument orig) $ python t.py
        no instrument: 4.34 ± 0.053 usec per checkpoint
      noop instrument: 4.36 ± 0.088 usec per checkpoint
 task_step instrument: 5.33 ± 0.057 usec per checkpoint

(elide-instrument simpler) $ python t.py
        no instrument: 4.24 ± 0.071 usec per checkpoint
      noop instrument: 4.2 ± 0.066 usec per checkpoint
 task_step instrument: 5.3 ± 0.11 usec per checkpoint
```
PyPy 7.3.1:
```
(master) $ python t.py
        no instrument: 0.713 ± 0.031 usec per checkpoint
      noop instrument: 1.03 ± 0.05 usec per checkpoint
 task_step instrument: 1.04 ± 0.028 usec per checkpoint

(elide-instrument simpler) $ python t.py
        no instrument: 0.746 ± 0.044 usec per checkpoint
      noop instrument: 0.695 ± 0.02 usec per checkpoint
 task_step instrument: 0.954 ± 0.026 usec per checkpoint
```
Performance measurement script:
```
import trio
import sys

class NullInstrument: pass

class DummyInstrument:
  def before_task_step(self, task): pass
  def after_task_step(self, task): pass

async def main():
  for _ in range(1000): await trio.lowlevel.cancel_shielded_checkpoint()
  t0 = trio.current_time()
  for _ in range(200000): await trio.lowlevel.cancel_shielded_checkpoint()
  t1 = trio.current_time()
  return (t1 - t0) * 5

times = {"no": [], "noop": [], "task_step": []}

for _ in range(10):
  times["no"].append(trio.run(main))
  times["noop"].append(trio.run(main, instruments=[NullInstrument()]))
  times["task_step"].append(trio.run(main, instruments=[DummyInstrument()]))
  print(".", end="")
  sys.stdout.flush()
print()

import numpy as np
import math

for key, vals in times.items():
  mean = np.mean(vals)
  stdev = np.std(vals)
  stderr = stdev / math.sqrt(len(vals))
  print(f"{key:>10} instrument: {mean:.3} ± {stderr*2.262:.2} usec per checkpoint")
```

pquentin:
> (we don't need type-checking internally that much, we have tests!)

Does that mean that we'll never be able to try mypyc?

njsmith:
I think, but am not certain, that mypyc is fine with explicitly annotated
dynamic casts and vague types like "Callable[..., Any]" – it just falls
back on runtime checking, like regular CPython. I'm more skeptical about it
being able to handle the dynamically generated slots in the initial PR :-)

PR #1650, opened 2020-06-26
Improved docs related to child cancellation

### LINKED ISSUE #1640: move_on_after inside nursery behaves surprisingly
Hello, I am looking into using trio since [this blogpost](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/) really has me convinced that there are some missing puzzle pieces in async code and trio might be a solution. While I was playing around with trio I ran into this very surprising behaviour:
```python
import time
import trio


async def heartbeat():
    while True:
        start = time.time()
        await trio.sleep(1)
        delay = time.time() - start - 1
        print(f'heartbeat delay = {delay:.3f}s')


async def main():
        with trio.move_on_after(5):
            nursery.start_soon(heartbeat)
        await trio.sleep(2.5)
        print('Start doing a thing')


if __name__ == '__main__':
    trio.run(main)
```
This does not behave as expected and after a little but of troubleshooting the [answer was right in the docs](https://trio.readthedocs.io/en/stable/reference-core.html#child-tasks-and-cancellation)! However, the documentation doesn't negate the surprise. Is there any reasonable way to have this behave as expected? Or is this gotcha unavoidable due to architecture of trio?

Just to be clear, the heartbeat never dies! The timeout is ignored since the timeout doesn't apply to the `nursery` and therefore it also doesn't apply to the `nursery.start_soon()` even though that function is called inside a timeout block.

RmStorm:
closes #1640

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -670,16 +670,14 @@ expires::
 
 Note that what matters here is the scopes that were active when
 :func:`open_nursery` was called, *not* the scopes active when
-``start_soon`` is called. The reason for this is that the ``start_soon`` call immediately exits and the scheduled task is awaited by the nursery. So for example, the timeout block below does
-nothing at all because the ``with move_on_after(TIMEOUT):`` block is immediately exited when ``start_soon`` returns and ``child`` is awaited by the nursery::
+``start_soon`` is called. So for example, the timeout block below does
+nothing at all::
 
    async with trio.open_nursery() as nursery:
        with move_on_after(TIMEOUT):  # don't do this!
            nursery.start_soon(child)
 
-
-This behaviour is also discussed here: `issue 1640
-<https://github.com/python-trio/trio/issues/1640>`_.
+The reason for this is that the ``start_soon`` call immediately returns and the scheduled task actually runs inside the nursery. So in the above snippet the contextmanager ``with move_on_after(TIMEOUT):`` block is immediately exited after ``start_soon`` returns. Since ``move_on_after`` only imposes a time limit on the body of its ``with`` block and that block is exited it cannot cancel the child task.
 
 Errors in multiple child tasks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -677,7 +677,11 @@ nothing at all::
        with move_on_after(TIMEOUT):  # don't do this!
            nursery.start_soon(child)
 
-The reason for this is that the ``start_soon`` call immediately returns and the scheduled task actually runs inside the nursery. So in the above snippet the contextmanager ``with move_on_after(TIMEOUT):`` block is immediately exited after ``start_soon`` returns. Since ``move_on_after`` only imposes a time limit on the body of its ``with`` block and that block is exited it cannot cancel the child task.
+Why is this so? Well, ``start_soon()`` returns as soon as it has scheduled the new task to start running. The flow of execution in the parent then continues on to exit the ``with move_on_after(TIMEOUT):`` block, at which point Trio forgets about the timeout entirely. In order for the timeout to apply to the child task, Trio must be able to tell that its associated cancel scope will stay open for at least as long as the child task is executing. And Trio can only know that for sure if the cancel scope block is outside the nursery block.
+
+You might wonder why Trio can't just remember "this task should be cancelled in ``TIMEOUT`` seconds", even after the ``with move_on_after(TIMEOUT):`` block is gone. The reason has to do with :ref:`how cancellation is implemented <cancellation>`. Recall that cancellation is represented by a `Cancelled` exception, which eventually needs to be caught by the cancel scope that caused it. (Otherwise, the exception would take down your whole program!) In order to be able to cancel the child tasks, the cancel scope has to be able to "see" the `Cancelled` exceptions that they raise -- and those exceptions come out of the ``async with open_nursery()`` block, not out of the call to ``start_soon()``.
+
+If you want a timeout to apply to one task but not another, then you need to put the cancel scope in that individual task's function -- ``child()``, in this example.
 
 Errors in multiple child tasks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -670,13 +670,16 @@ expires::
 
 Note that what matters here is the scopes that were active when
 :func:`open_nursery` was called, *not* the scopes active when
-``start_soon`` is called. So for example, the timeout block below does
-nothing at all::
+``start_soon`` is called. The reason for this is that the ``start_soon`` call immediately exits and the scheduled task is awaited by the nursery. So for example, the timeout block below does
+nothing at all because the ``with move_on_after(TIMEOUT):`` block is immediately exited when ``start_soon`` returns and ``child`` is awaited by the nursery::
 
    async with trio.open_nursery() as nursery:
        with move_on_after(TIMEOUT):  # don't do this!
            nursery.start_soon(child)
 
+
+This behaviour is also discussed here: `issue 1640
+<https://github.com/python-trio/trio/issues/1640>`_.
 
 Errors in multiple child tasks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


docs/source/reference-core.rst
```diff
@@ -670,14 +670,17 @@ expires::
 
 Note that what matters here is the scopes that were active when
 :func:`open_nursery` was called, *not* the scopes active when
-``start_soon`` is called. So for example, the timeout block below does
-nothing at all::
+``start_soon`` is called. The reason for this is that the ``start_soon`` call immediately exits and the scheduled task is awaited by the nursery. So for example, the timeout block below does
```
oremanj, line 673: I would recommend leaving this paragraph as it is, and then adding a more detailed explanation below the snippet (rather than linking to the place where the more detailed explanation can be found).

Also, this is a pretty foundational part of the docs, so I think it's important to be extra careful about consistent terminology: for example, the `start_soon` call *returns* (rather than "exits") and the scheduled task *runs inside* (rather than "is awaited by") the nursery.

RmStorm: Yeah, adding an extra paragraph or two below is probably a lot clearer than shoehorning it in like I did now. I will revise the PR sometime tomorrow!

Just a heads up: English is not my first language and I'm not fantastic at writing in general so this might take two or three tries to really get right. With that in mind, don't feel bad for having me revise it a couple times more, it's only to be expected. I also think I checked the box to let maintainers edit the PR directly😉

docs/source/reference-core.rst
```diff
@@ -677,6 +677,7 @@ nothing at all::
        with move_on_after(TIMEOUT):  # don't do this!
            nursery.start_soon(child)
 
+The reason for this is that the ``start_soon`` call immediately returns and the scheduled task actually runs inside the nursery. So in the above snippet the contextmanager ``with move_on_after(TIMEOUT):`` block is immediately exited after ``start_soon`` returns. Since ``move_on_after`` only imposes a time limit on the body of its ``with`` block and that block is exited it cannot cancel the child task.
```
oremanj, line 680: I expanded on this a bit further; feel free to apply my suggestion with or without changes if you think it makes all of this easier to understand! Also, please wrap lines in the documentation sources at approximately 80 chars -- we're not strict about this, but it's nice not to have things going way off the right hand side of the window when editing.

```suggestion
Why is this so? Well, ``start_soon()`` returns as soon as it has scheduled the new task to start running. The flow of execution in the parent then continues on to exit the ``with move_on_after(TIMEOUT):`` block, at which point Trio forgets about the timeout entirely. In order for the timeout to apply to the child task, Trio must be able to tell that its associated cancel scope will stay open for at least as long as the child task is executing. And Trio can only know that for sure if the cancel scope block is outside the nursery block.

You might wonder why Trio can't just remember "this task should be cancelled in ``TIMEOUT`` seconds", even after the ``with move_on_after(TIMEOUT):`` block is gone. The reason has to do with :ref:`how cancellation is implemented <cancellation>`. Recall that cancellation is represented by a `Cancelled` exception, which eventually needs to be caught by the cancel scope that caused it. (Otherwise, the exception would take down your whole program!) In order to be able to cancel the child tasks, the cancel scope has to be able to "see" the `Cancelled` exceptions that they raise -- and those exceptions come out of the ``async with open_nursery()`` block, not out of the call to ``start_soon()``.

If you want a timeout to apply to one task but not another, then you need to put the cancel scope in that individual task's function -- ``child()``, in this example.
```

RmStorm: That is much much clearer then I could have managed. Thank you! I'm sure this will help future people running into the issue.

pquentin:
We got a random failure in test_sleep on Windows.

https://github.com/python-trio/trio/pull/1650/checks?check_run_id=833555681

```

2020-07-03T07:53:57.4882456Z ================================== FAILURES ===================================
2020-07-03T07:53:57.4882748Z _________________________________ test_sleep __________________________________
2020-07-03T07:53:57.4882924Z 
2020-07-03T07:53:57.4883163Z     @slow
2020-07-03T07:53:57.4883409Z     async def test_sleep():
2020-07-03T07:53:57.4883678Z         async def sleep_1():
2020-07-03T07:53:57.4883942Z             await sleep_until(_core.current_time() + TARGET)
2020-07-03T07:53:57.4884171Z     
2020-07-03T07:53:57.4884431Z         await check_takes_about(sleep_1, TARGET)
2020-07-03T07:53:57.4884651Z     
2020-07-03T07:53:57.4884868Z         async def sleep_2():
2020-07-03T07:53:57.4885137Z             await sleep(TARGET)
2020-07-03T07:53:57.4885329Z     
2020-07-03T07:53:57.4885547Z >       await check_takes_about(sleep_2, TARGET)
2020-07-03T07:53:57.4885726Z 
2020-07-03T07:53:57.4885989Z C:\hostedtoolcache\windows\Python\3.8.3\x64\Lib\site-packages\trio\tests\test_timeouts.py:54: 
2020-07-03T07:53:57.4886274Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
2020-07-03T07:53:57.4886442Z 
2020-07-03T07:53:57.4886682Z f = <function test_sleep.<locals>.sleep_2 at 0x0000025FBFFB58B0>
2020-07-03T07:53:57.4886924Z expected_dur = 1.0
2020-07-03T07:53:57.4887402Z 
2020-07-03T07:53:57.4887624Z     async def check_takes_about(f, expected_dur):
2020-07-03T07:53:57.4887879Z         start = time.perf_counter()
2020-07-03T07:53:57.4888149Z         result = await outcome.acapture(f)
2020-07-03T07:53:57.4888383Z         dur = time.perf_counter() - start
2020-07-03T07:53:57.4888636Z         print(dur / expected_dur)
2020-07-03T07:53:57.4888873Z         # 1.5 is an arbitrary fudge factor because there's always some delay
2020-07-03T07:53:57.4889231Z         # between when we become eligible to wake up and when we actually do. We
2020-07-03T07:53:57.4889469Z         # used to sleep for 0.05, and regularly observed overruns of 1.6x on
2020-07-03T07:53:57.4889821Z         # Appveyor, and then started seeing overruns of 2.3x on Travis's macOS, so
2020-07-03T07:53:57.4890071Z         # now we bumped up the sleep to 1 second, marked the tests as slow, and
2020-07-03T07:53:57.4890336Z         # hopefully now the proportional error will be less huge.
2020-07-03T07:53:57.4890570Z         #
2020-07-03T07:53:57.4890777Z         # We also also for durations that are a hair shorter than expected. For
2020-07-03T07:53:57.4891359Z         # example, here's a run on Windows where a 1.0 second sleep was measured
2020-07-03T07:53:57.4891651Z         # to take 0.9999999999999858 seconds:
2020-07-03T07:53:57.4891940Z         #   https://ci.appveyor.com/project/njsmith/trio/build/1.0.768/job/3lbdyxl63q3h9s21
2020-07-03T07:53:57.4892245Z         # I believe that what happened here is that Windows's low clock resolution
2020-07-03T07:53:57.4892520Z         # meant that our calls to time.monotonic() returned exactly the same
2020-07-03T07:53:57.4892792Z         # values as the calls inside the actual run loop, but the two subtractions
2020-07-03T07:53:57.4893060Z         # returned slightly different values because the run loop's clock adds a
2020-07-03T07:53:57.4893331Z         # random floating point offset to both times, which should cancel out, but
2020-07-03T07:53:57.4893605Z         # lol floating point we got slightly different rounding errors. (That
2020-07-03T07:53:57.4893998Z         # value above is exactly 128 ULPs below 1.0, which would make sense if it
2020-07-03T07:53:57.4894261Z         # started as a 1 ULP error at a different dynamic range.)
2020-07-03T07:53:57.4894514Z >       assert (1 - 1e-8) <= (dur / expected_dur) < 1.5
2020-07-03T07:53:57.4894949Z E       assert (1.5743606000000057 / 1.0) < 1.5
2020-07-03T07:53:57.4895120Z 
2020-07-03T07:53:57.4895627Z C:\hostedtoolcache\windows\Python\3.8.3\x64\Lib\site-packages\trio\tests\test_timeouts.py:35: AssertionError
2020-07-03T07:53:57.4896253Z ---------------------------- Captured stdout call -----------------------------
2020-07-03T07:53:57.4896471Z 1.0007407000000015
2020-07-03T07:53:57.4896680Z 1.5743606000000057
```

Closing/reopening

pquentin:
And again in https://github.com/python-trio/trio/pull/1650/checks?check_run_id=833679472 with the following values:

```
---------------------------- Captured stdout call -----------------------------
1.0014539000000013
1.696614099999998
```

PR #1655, opened 2020-06-29
Only assert platform when type checking

pquentin:
We would like to write:

```
if TYPE_CHECKING:
    assert sys.platform == "linux"
```

But that hides the assert to mypy. This commits finds a way to avoid that issue.

We still use platform checks in `trio._core._run` and `trio._core.__init__`. Do we want to avoid them there too?

### DIFF
--- a/trio/_core/_io_kqueue.py
+++ b/trio/_core/_io_kqueue.py
@@ -11,7 +11,7 @@ from .. import _core
 from ._run import _public
 from ._wakeup_socketpair import WakeupSocketpair
 
-assert not TYPE_CHECKING or sys.platform != "linux" and sys.platform != "win32"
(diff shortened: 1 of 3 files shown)


trio/_core/_io_kqueue.py
```diff
@@ -10,7 +11,7 @@
 from ._run import _public
 from ._wakeup_socketpair import WakeupSocketpair
 
-assert sys.platform != "linux" and sys.platform != "win32"
+assert not TYPE_CHECKING or sys.platform != "linux" and sys.platform != "win32"
```
njsmith, line 14: Please put parens around the `and` clause here. I know `or` is lower precedence, but I had to look it up :-)

And feel free to merge once you've done that

pquentin: Fair enough :) Done, and merged

PR #1656, opened 2020-07-01
Make Trio pass 'mypy -m trio -m trio.testing'

oremanj:
This is enough to start adding type hints. We aren't yet type-checking our own tests though -- only the code that runs when someone says `import trio`.


oremanj:
Seems to have not picked up my .coveragerc changes; trying close/reopen.

oremanj:
That didn't work, so I'm replacing with a new PR.

PR #1657, opened 2020-07-01
Make Trio pass 'mypy -m trio -m trio.testing'

oremanj:
This is enough to start adding type hints. We aren't yet type-checking our own tests though -- only the code that runs when someone says `import trio`.

Identical copy of #1656, trying to get codecov to notice coveragerc changes.


njsmith:
Looks good to me. Not sure what's going on with the codecov results, and your post says you're fiddling with it, so I'll let you do the merge when you're ready :-)

I guess at some point we'll want to think about type coverage... I'm not sure if there's any way to get codecov to report two distinct kinds of coverage :-/

pquentin:
This is really cool, thanks!

PR #1661, opened 2020-07-04
Windows: Deal with Komodia-based LSPs that block SIO_BASE_HANDLE

### LINKED ISSUE #1659: IOCP polling doesn't work with Astrill VPN LSP
trio 0.16
win10 1803
```
  File "c:/Users/c/Desktop/test2.py", line 37, in <module>
    trio.run(parent)
  File "C:\Program Files\Python38\lib\site-packages\trio\_core\_run.py", line 1879, in run
    runner = setup_runner(
  File "C:\Program Files\Python38\lib\site-packages\trio\_core\_run.py", line 1781, in setup_runner
    io_manager = TheIOManager()
  File "C:\Program Files\Python38\lib\site-packages\trio\_core\_io_windows.py", line 349, in __init__
    base_handle = _get_base_socket(s, which=WSAIoctls.SIO_BASE_HANDLE)
  File "C:\Program Files\Python38\lib\site-packages\trio\_core\_io_windows.py", line 202, in _get_base_socket
    raise_winerror(code)
  File "C:\Program Files\Python38\lib\site-packages\trio\_core\_windows_cffi.py", line 321, in raise_winerror
    raise OSError(0, msg, filename, winerror, filename2)
OSError: [WinError 10014] The system detected an invalid pointer address in attempting to use a pointer argument in a call
```
both `echo-server.py` and `echo-client.py` in last of `https://trio.readthedocs.io/en/stable/tutorial.html` will cause this

oremanj:
Logic is ported from the analogous workaround in wepoll - thanks @piscisaureus for the pointer.

Closes #1659.


njsmith:
Overall looks good. I just pushed a change to test against a real Astrill install... hopefully that works. (I did verify that without your fix, it fails in the expected way: https://github.com/njsmith/trio/runs/836936836).

Given this, we should at least update the comment in the last test. Do you think we should keep that test? It's pretty a pretty complex set of monkeypatching. But I guess we aren't testing any real double-layering situations either, so it's adding some coverage there. (......maybe we *should* we be testing real double-layering situations?)

oremanj:
I think the most important part of the last test is testing what happens in the case where `SIO_BASE_HANDLE` fails and `SIO_BSP_HANDLE_POLL` returns the same socket. We don't know of any real LSP that does that, but if one does, it had better error rather than making an infinite loop. (Especially since this occurs in a KI-protected context such that the infinite loop would be unkillable.)

I think I can simplify the test to only test the error case. I'm not too worried about the double-layered case by comparison, especially since testing it for real would require installing two different *Komodia-based* LSPs.

PR #1671, opened 2020-07-21
CI: Fetch Komodia-based LSP from a source we control

### LINKED ISSUE #1667: CI: sketchy Komodia LSP link stopped working
See #1666. I wonder if they noticed an uptick in the frequency of it being downloaded...

I have the installer binary, but no good place to host it.

oremanj:
Fixes #1667.


oremanj:
cc @piscisaureus if you want to update wepoll's CI too

Issue #1691, opened 2020-08-24
`Nursery.start_soon()` does not return `True` nor `False` as specified in docstring/documentation


doomleika:
[Documentation ](https://trio.readthedocs.io/en/latest/reference-core.html#trio.Nursery.start_soon) and [docstring](https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L978) for the `Nursery.start_soon()` have stated

```
        Returns:
            True if successful, False otherwise.
```

However [this is not the case.](https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L986) 

```
        GLOBAL_RUN_CONTEXT.runner.spawn_impl(async_fn, args, self, name)
```

It doesn't return anything.

PR #1694, opened 2020-08-24
Remove "Returns" section from Nursery.start_soon docs.

joernheissler:
Closes: #1691


pquentin (approved):
Thanks! Also, nice catch, @doomleika.

Issue #1457, opened 2020-04-19
A nursery might inject a Cancelled even when none of the tasks received a Cancelled


vxgmichel:
I noticed that nurseries might inject `Cancelled` exception even when none of the tasks have received a Cancelled. I have two different examples reproducing this issue.

### A nursery with a single task raising an exception

The following code runs a nursery with a single task raising an exception while the outer scope has been cancelled:

```python
async def main():
    with trio.CancelScope() as scope:
        try:
            async with trio.open_nursery() as nursery:
                scope.cancel()
                raise RuntimeError
        except trio.MultiError as exc:
            print("Exceptions:", exc)
            print("Cancelled caught:", nursery.cancel_scope.cancelled_caught)
```

It prints:
```
Exceptions: RuntimeError(), Cancelled()
Cancelled caught: False
```
The `Cancelled()` exception is injected here:
https://github.com/python-trio/trio/blob/fbe3bd39ea23335bf39a3a034de3099e90dbc926/trio/_core/_run.py#L845-L851

A potential fix could to be to use [cancel_shielded_checkpoint()](https://trio.readthedocs.io/en/latest/reference-hazmat.html#trio.hazmat.cancel_shielded_checkpoint) instead of `checkpoint()`.

### A nursery with a two tasks, both raising an exception

The following code runs a nursery with two tasks both raising an exception while the outer scope has been cancelled:

```python
async def main():
    with trio.CancelScope() as scope:
        try:
            async with trio.open_nursery() as nursery:
                nursery.start_soon(_raise, RuntimeError(1))
                scope.cancel()
                raise RuntimeError(2)
        except trio.MultiError as exc:
            print("Exceptions:", exc)
            print("Cancelled caught:", nursery.cancel_scope.cancelled_caught)
```

It prints:
```
Exceptions: RuntimeError(2), Cancelled(), RuntimeError(1)
Cancelled caught: False
```
The `Cancelled()` exception is injected here:
https://github.com/python-trio/trio/blob/fbe3bd39ea23335bf39a3a034de3099e90dbc926/trio/_core/_run.py#L835-L840

I don't see the point of injecting a `Cancelled` here, since the cancellation is going to be ignored anyway: isn't it a rule of trio that a cancelled (in the sense of `cancel_called`) but completed operation should not raise a `Cancelled`?

> Raising Cancelled means that the operation did not happen. 
https://trio.readthedocs.io/en/latest/reference-core.html#cancellable-primitives

[...]

PR #1696, opened 2020-08-26
Nursery: don't act as a checkpoint when not running anything

catern:
Previously, a Nursery would checkpoint on exit, regardless of whether
it was running anything. This is a bit annoying, see #1457, so let's
change it to not checkpoint on exit. Now it won't be a checkpoint at
all if there's no tasks running inside.

However, this doesn't fully solve #1457, because we'll still inject a
cancel even if none of the child tasks raise Cancelled.

This seems to break
trio/_core/tests/test_asyncgen.py::test_last_minute_gc_edge_case
such that it no longer ever hits the edge case.

That seems like a pretty complicated test, so maybe @oremanj can say
better why exactly it's breaking with this change.


njsmith:
> However, this doesn't fully solve #1457, because we'll still inject a cancel even if none of the child tasks raise Cancelled.

Here's the code that does that, at line 923:

```python
            def aborted(raise_cancel):
                self._add_exc(capture(raise_cancel).error)
```

...I was going to say we could delete this line and simply ignore cancellation here, but that's not *quite* right. We do want to ignore `Cancelled` exceptions here, but if Trio is using cancellation to deliver a `KeyboardInterrupt` to the main task while it's blocked in `__aexit__`, then we still need that to be injected so that the nursery will cancel the other tasks and then raise `KeyboardInterrupt`. That should be fixed at some point Soon™, but in the mean time I guess we could work around this by adding a check for whether `capture(raise_cancel).error` is a `Cancelled`, and if so then ignore it.

oremanj:
I left a comment on #1457: I think we shouldn't drop the schedule point in either branch, only the cancellation point; and we shouldn't drop the cancellation point if the nursery didn't run any tasks, so that `async with open_nursery(): pass` still executes a full checkpoint.

The test is failing because it relies on nursery exit being a schedule point.

trio/_core/tests/test_run.py
```diff
@@ -1685,13 +1681,24 @@ async def nothing(task_status=_core.TASK_STATUS_IGNORED):
...
+    # but if the task does not execute any checkpoints, and exits, then start()
+    # doesn't raise Cancelled, since the task completed successfully.
+    async def started_with_no_checkpoint(task_status=_core.TASK_STATUS_IGNORED):
+        task_status.started("hi")
+
+    async with _core.open_nursery() as nursery:
+        with _core.CancelScope() as cs:
+            cs.cancel()
+            await nursery.start(started_with_no_checkpoint)
+
```
catern, line 1701: Some more implications here. This is because a Nursery is used in the implementation of start().

I'd argue that this is correct.

catern:
Also, what do you think about making assert_yields_or_not public? I just needed it for the test that a nursery is a schedule point but not a cancel point. Maybe that's something we should discourage?

njsmith:
> Also, what do you think about making assert_yields_or_not public? I just needed it for the test that a nursery is a schedule point but not a cancel point. Maybe that's something we should discourage?

I actually just ran into this issue again while working on #1805 on stream, so I did some tweaks to land this quicker. And one of those tweaks was to revert the `assert_yields_or_not` changes, because they seemed overly complex for a single weird internal test :-). So I guess that answers the question...

newsfragments/1457.feature.rst
```diff
@@ -0,0 +1,7 @@
+When exiting a nursery block, the parent task always waits for child
+tasks to exit. This wait cannot be cancelled. However, previously, if
+you tried to cancel it, it *would* inject a `Cancelled` exception,
+even though it wasn't cancelled. Most users probably never noticed
```
goodboy, line 4: > This wait cannot be cancelled. However, previously, if
you tried to cancel it, it would inject a Cancelled exception,
even though it wasn't cancelled.

So this means the `Cancelled` is still injected but now is ignored via the new [`isinstance` check](https://github.com/python-trio/trio/pull/1696/files#diff-568b66d9830d9516b524413eb041a1d4c8ccf0b76af4e10d96098888ce323235R924)?

Also if the following is true:

> This wait cannot be cancelled. However, previously, if
you tried to cancel it, it would inject a Cancelled exception,
even though it wasn't cancelled.

Then why [the change to ` cancel_shielded_checkpoint()`](https://github.com/python-trio/trio/pull/1696/files#diff-568b66d9830d9516b524413eb041a1d4c8ccf0b76af4e10d96098888ce323235R935)?

trio/_core/tests/test_run.py
```diff
@@ -433,12 +433,12 @@ async def crasher():
...
                 # nursery block continue propagating to reach the
                 # outer scope.
-                assert len(multi_exc.exceptions) == 5
+                assert len(multi_exc.exceptions) == 4
                 summary = {}
                 for exc in multi_exc.exceptions:
                     summary.setdefault(type(exc), 0)
                     summary[type(exc)] += 1
-                assert summary == {_core.Cancelled: 4, KeyError: 1}
+                assert summary == {_core.Cancelled: 3, KeyError: 1}
```
goodboy, line 441: Clarifying the cancelled count in a comment might be clearer here for those unfamiliar with the recent change?

Something like *we only receive cancelleds from subtasks not from the nursery itself*?

trio/_core/tests/test_run.py
```diff
@@ -1605,20 +1608,35 @@ async def test_trivial_yields():
...
+
+    before_schedule_points = task._schedule_points
+
+    async def noop_with_no_checkpoint():
+        pass
+
+    with _core.CancelScope() as cs:
+        cs.cancel()
+        async with _core.open_nursery() as nursery:
+            nursery.start_soon(noop_with_no_checkpoint)
```
goodboy, line 1630: I presume it's out of scope (and already covered in other tests) but normally doing this cancellation before opening the nursery results in cancellation at the first checkpoint right?

trio/_core/tests/test_run.py
```diff
@@ -1685,28 +1703,37 @@ async def nothing(task_status=_core.TASK_STATUS_IGNORED):
...
-    # and if after the no-op started(), the child crashes, the error comes out
-    # of start()
+    # but if the task does not execute any checkpoints, and exits, then start()
+    # doesn't raise Cancelled, since the task completed successfully.
+    async def started_with_no_checkpoint(task_status=_core.TASK_STATUS_IGNORED):
+        task_status.started("hi")
+
+    async with _core.open_nursery() as nursery:
+        with _core.CancelScope() as cs:
+            cs.cancel()
```
goodboy, line 1721: In the case where you cancel from inside the nursery after having started tasks with checkpoints, won't counts of expected cancels change as well or is that entirely covered by the one test change in `test_run.py`?

goodboy (commented):
Code all looks great (as usual) but some minor clarifications to an onlooker (newb) such as myself might be handy 😸

catern:
The changes look fine to me. Excited for this to be fixed!

richardsheridan:
The merge was not as painful as one might expect. The two new commits are nearly trivial fixes responsive to recent updates to the test suite.

PR #1698, opened 2020-08-27
Support alternatives to trio.run() for pytest-trio

altendky:
Draft for:
- [x] ~~Generally concluding the other PRs are going to work~~
  - Since moving `trio_test()` to pytest-trio, the other PRs no longer depend on this one.
- [ ] Maybe deprecate `trio_test()`, maybe just delete it?

Relates to:
- https://github.com/python-trio/pytest-trio/pull/105
- https://github.com/altendky/qtrio/pull/152

### DIFF
--- a/trio/testing/_trio_test.py
+++ b/trio/testing/_trio_test.py
@@ -12,18 +12,24 @@ from ..abc import Clock, Instrument
 #
 # Also: if a pytest fixture is passed in that subclasses the Clock abc, then
 # that clock is passed to trio.run().
-def trio_test(fn):
-    @wraps(fn)
-    def wrapper(**kwargs):
-        __tracebackhide__ = True
-        clocks = [c for c in kwargs.values() if isinstance(c, Clock)]
-        if not clocks:
-            clock = None
-        elif len(clocks) == 1:
-            clock = clocks[0]
-        else:
-            raise ValueError("too many clocks spoil the broth!")
-        instruments = [i for i in kwargs.values() if isinstance(i, Instrument)]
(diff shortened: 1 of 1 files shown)


njsmith:
What would you say to copying this code into `pytest-trio` instead, and fixing it there? It's sort of been on my todo list anyway, and doing it now seems like it would simplify your life by consolidating all the changes into one project...

altendky:
Whatever, I don't particularly care at this point so I'll plan on moving it.  Or, did you more literally mean copy and leave this copy alone?  It is accessible via the 'public path' of `trio.testing.trio_test()` and is also in the docs, albeit as a stub.  Should this be treated with a full deprecation process?

https://trio.readthedocs.io/en/stable/reference-testing.html#test-harness-integration
image

altendky:
https://github.com/search?q=%22pyfuncitem.obj+%3D+trio_test%28pyfuncitem.obj%29%22&type=Code
Looks like there's a `conftest.py` that's been copied around for a few years that does reference `trio_test()` (including trimeter...) so deprecation it is.  Unless it's just going to stick around?  (you did say 'copy')

altendky:
Or maybe I'm just jumping way ahead of myself with this because `trio_test()` is also used throughout Trio's own tests.  Maybe it's not worth even deprecating.  In the mean time trimeter's CI might be fixed.  `:]`

njsmith:
Yeah, idk, maybe we should deprecate and remove it, but we don't need to make a decision right now, now that this is off your critical path.

PR #1706, opened 2020-08-31
Fix setuptools 50 import in Python 3.6.1 pylint test

pquentin:
See https://github.com/pypa/setuptools/issues/2350 for details. pip, setuptools and wheel are the only packages that we don't pin, and indeed the pylint tests started failing with setuptools 50.0 on Python 3.6.1. (No other configuration is affected.)


pquentin:
Yeah. The good news is that this is specific to 3.6.1, for all other Pythons we can use the setuptools version! Thanks for the review

PR #1709, opened 2020-09-03
CI: Use older ipython/traitlets on 3.6 since latest versions have dropped 3.6 support

altendky:


altendky:
Note that this includes both #1704 and #1708.

PR #1715, opened 2020-09-05
Bump up the timeout in test_pipes to hopefully reduce spurious CI failures

njsmith:


altendky (approved):
Over in QTrio I had plenty of timeouts on subprocessed pytest tests including ones that did nearly nothing.  It seemed they would cluster a bit so there would be a few successful builds and then one job would fail three or four such subprocessed tests.  Bad luck getting slow workers?  Overloaded?  Bumping timeouts from 10 to 40 seconds seemed to end the flakiness so 30 sounds reasonable to me.

njsmith:
Yeah, I think these free CI systems are overprovisioned and have really noisy neighbors, so sometimes your whole VM will just like, freeze for a few seconds and then resume. So you need pretty generous timeouts.

PR #1720, opened 2020-09-11
docs: Fix simple typo, renegotation -> renegotiation

timgates42:
There is a small typo in trio/_ssl.py, trio/tests/test_ssl.py.

Should read `renegotiation` rather than `renegotation`.


tjstum:
Thanks!

PR #1737, opened 2020-09-29
clarified the 'Networking with Trio' section of the tutorial

Adrien-Luxey:
I misunderstood this section of the tutorial: I thought the "echo client" code provided both server and client. 

So I made it clear in the tutorial that the first code sample (the client) does not work without a server listening on the other end.

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -836,18 +836,24 @@ Networking with Trio
 Now let's take what we've learned and use it to do some I/O, which is
 where async/await really shines.
 
-
-An echo client
-~~~~~~~~~~~~~~
-
-The traditional application for demonstrating network APIs is an "echo
-server": a program that accepts arbitrary data from a client, and then
-sends that same data right back. (Probably a more relevant example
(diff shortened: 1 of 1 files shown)


docs/source/tutorial.rst
```diff
@@ -857,6 +863,9 @@ will send some data at our echo server and get responses back:
 .. literalinclude:: tutorial/echo-client.py
    :linenos:
 
+Note that this code will not work without a TCP server listening at 
+``127.0.0.1:12345``.
```
pquentin, line 867: I don't think the exact port number matters much, and it's useful to mention that we'll write a server below, so here's my suggestion:

```suggestion
Note that this code will not work without a TCP server such as the one
we'll implement below.
```

pquentin (changes requested):
Looking at the result on https://trio--1737.org.readthedocs.build/en/1737/tutorial.html, I only have one suggestion, please tell us what you think.

pquentin (approved):
Thanks!

PR #1742, opened 2020-10-01
add Python 3.9-rc to CI build

belm0:


belm0:
a naive attempt

belm0:
I guess Python versions are limited to what's available on the virtual-env pages like https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

Newer Python versions need to use "setup-python" action?

https://github.com/actions/python-versions
https://github.com/actions/setup-python

altendky:
Looks like I updated to `actions/setup-python` v2 over in https://github.com/altendky/qtrio/pull/113 when I was trying to get 3.9.  Though I can't say I remember the exact reason so it may not be relevant.  I'm using Docker for Linux, so that also isn't relevant, but I did some work to try to get anywhere from alpha through latest release without having to update it.

```
python-version: '${{ matrix.python }}.0-alpha - ${{ matrix.python }}.X'
```

belm0:
@altendky thanks, I'll try

it might be good enough to just hard-code to specific 3.9 releases, keeping it simple

belm0:
How do I make the 3.9 tests "required"?  Perhaps it needs admin access.

altendky:
That's part of why I add an 'all' job that depends on the things I want to require, then I require the 'all' job from GitHub.  Albeit with the cost of an extra job spinning up at the end of CI.

pquentin:
@belm0 Thanks! Please use your personal fork when submitting pull requests

pquentin:
By the way we were already testing 3.9-dev in Travis, so may want to remove the check in Travis.

We should also use the name "3.9-dev" in GitHub Actions, as it uses the latest dev 3.9, and does not stay at 3.9.0-rc2 forever. (Hopefully 3.9-0.dev continues to point to master when 3.9 is out, otherwise we'll have to keep using Travis for that.)

belm0:
> Please use your personal fork when submitting pull requests

Sorry, I thought I did.  It seems when I edit directly from github it doesn't give me an option.

belm0:
> We should also use the name "3.9-dev" in GitHub Actions, as it uses the latest dev 3.9, and does not stay at 3.9.0-rc2 forever.

@altendky mentioned a better solution if we really want to track latest rc:  https://github.com/python-trio/trio/pull/1742#issuecomment-701782845.  However 3.9 is expected to be released in a week or two, so it didn't seem bad to hard-code it (or at least, much better than having no coverage at all).

I don't think tracking "-dev" is appropriate, since it can change daily and perhaps doesn't go through any QA process.

altendky:
I thought we wanted both `-dev` (on Travis) and 'latest (pre)release' (on GitHub Actions).  And sure, using the alpha-through-whatever may not matter much for 3.9 but it's some general setup for future versions so they can be easily added as soon as there is any pre-release and not thought about again.  But...  I haven't even merged this in my own repo... Though I think that relates to my Linux/Docker usage.

PR #1760, opened 2020-10-14
Fix FreeBSD build by busting the cache

pquentin:
This reinstalls/rebuilds packages and stops depending on an old libffi version.

https://forums.freebsd.org/threads/certbot-python3-7-dependency-error.76808/

Failing build, for reference: https://travis-ci.org/github/python-trio/trio/jobs/735639080


altendky (approved):
I hit the failure over in #1761.  I can't say I made myself well aware of all the details of the caching but this seems to fix it and busting a cache shouldn't cost anything since the new cache has already been built.  Judge this review's value as you will.  `:]`

pquentin:
Thanks! :)

PR #1761, opened 2020-10-14
Use .png for favicon on RTD

altendky:


altendky:
Just cleaning up here after this was pointed out over at https://github.com/python-trio/pytest-trio/pull/112#pullrequestreview-508206073.

pquentin:
@altendky Thanks! Can you please rebase from master/merge master into your branch?

altendky:
@pquentin I thought I did for the FreeBSD cache bust and just double checked as well.  What do you see missing?

altendky:
Side note, I personally set repos to require PRs to be up to date to 1) point out when they are not, 2) make sure a real up-to-date build is present before merging, and 3) make the update a button click away.  But Trio usually has relevant reasons for being how it is...

pquentin:
Sorry, my bad. Also, I didn't realize you could configure all that stuff in GitHub

PR #1765, opened 2020-10-19
Fix a crash in pypy.

leojay:
Apparently, in pypy, "signal.set_wakeup_fd" doesn't have the keyword argument warn_on_full_buffer.

Here is a simple script to reproduce this problem:
```
import trio

async def handler(stream: trio.abc.Stream):
    await stream.receive_some(1)

async def trio_main():
    socket_listeners = await trio.serve_tcp(handler)

trio.run(trio_main)
```
The error message is:
```
  File "/home/leo/test/triotest/pypyvenv/site-packages/trio/_core/_run.py", line 2030, in unrolled_run
    runner.entry_queue.wakeup.wakeup_on_signals()
  File "/home/leo/test/triotest/pypyvenv/site-packages/trio/_core/_wakeup_socketpair.py", line 66, in wakeup_on_signals
    self.old_wakeup_fd = signal.set_wakeup_fd(fd, warn_on_full_buffer=False)
TypeError: set_wakeup_fd() got an unexpected keyword argument 'warn_on_full_buffer'

  File "t.py", line 9, in <module>
    trio.run(trio_main)
  File "/home/leo/test/triotest/pypyvenv/site-packages/trio/_core/_run.py", line 1919, in run
    timeout = gen.send(next_send)
  File "/home/leo/test/triotest/pypyvenv/site-packages/trio/_core/_run.py", line 2226, in unrolled_run

```

My environment is:
```
$ pypy3 --version
Python 3.7.4 (87875bf2dfd8, Sep 24 2020, 07:26:36)
[PyPy 7.3.2-alpha0 with GCC 7.3.1 20180303 (Red Hat 7.3.1-5)]
```
And 
```trio==0.17.0```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ jobs:
       dist: bionic
     - language: generic
       env: PYPY_NIGHTLY_BRANCH=py3.6
+    - language: generic
+      env: PYPY_NIGHTLY_BRANCH=py3.7
     # Qemu tests are also slow
     # FreeBSD:
     - language: generic
--- a/newsfragments/1765.bugfix.rst
+++ b/newsfragments/1765.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a crash in pypy-3.7
--- a/trio/_core/_wakeup_socketpair.py
+++ b/trio/_core/_wakeup_socketpair.py
@@ -18,7 +18,8 @@ def _has_warn_on_full_buffer():
 
     import inspect
 
-    return "warn_on_full_buffer" in inspect.getfullargspec(set_wakeup_fd).kwonlyargs
+    args_spec = inspect.getfullargspec(signal.set_wakeup_fd)
+    return "warn_on_full_buffer" in args_spec.kwonlyargs
 
 
 HAVE_WARN_ON_FULL_BUFFER = _has_warn_on_full_buffer()
--- a/trio/_core/_wakeup_socketpair.py
+++ b/trio/_core/_wakeup_socketpair.py
@@ -6,7 +6,7 @@ import warnings
 from .. import _core
 from .._util import is_main_thread
 
-if sys.version_info >= (3, 7) and '__pypy__' not in sys.builtin_module_names:
+if sys.version_info >= (3, 7) and "__pypy__" not in sys.builtin_module_names:
     HAVE_WARN_ON_FULL_BUFFER = True
 else:
     HAVE_WARN_ON_FULL_BUFFER = False
--- a/trio/_core/_wakeup_socketpair.py
+++ b/trio/_core/_wakeup_socketpair.py
@@ -6,10 +6,22 @@ import warnings
 from .. import _core
 from .._util import is_main_thread
 
-if sys.version_info >= (3, 7) and "__pypy__" not in sys.builtin_module_names:
-    HAVE_WARN_ON_FULL_BUFFER = True
-else:
-    HAVE_WARN_ON_FULL_BUFFER = False
+
+def _has_warn_on_full_buffer():
+    if sys.version_info < (3, 7): 
+        return False
+
+    if "__pypy__" not in sys.builtin_module_names:
+        # CPython has warn_on_full_buffer. Don't need to inspect.
+        # Also, CPython doesn't support inspecting built-in functions.
+        return True
+
+    import inspect
+
+    return "warn_on_full_buffer" in inspect.getfullargspec(set_wakeup_fd).kwonlyargs
+
+
+HAVE_WARN_ON_FULL_BUFFER = _has_warn_on_full_buffer()
 
 
 class WakeupSocketpair:
--- a/trio/_core/_wakeup_socketpair.py
+++ b/trio/_core/_wakeup_socketpair.py
@@ -8,7 +8,7 @@ from .._util import is_main_thread
 
 
 def _has_warn_on_full_buffer():
-    if sys.version_info < (3, 7): 
+    if sys.version_info < (3, 7):
         return False
 
     if "__pypy__" not in sys.builtin_module_names:
--- a/trio/_core/tests/test_guest_mode.py
+++ b/trio/_core/tests/test_guest_mode.py
@@ -502,6 +502,10 @@ def test_guest_mode_autojump_clock_threshold_changing():
 
 
 @pytest.mark.skipif(buggy_pypy_asyncgens, reason="PyPy 7.2 is buggy")
+@pytest.mark.xfail(
+    sys.implementation.name == "pypy" and sys.version_info >= (3, 7),
+    reason="async generator issue under investigation",
+)
 def test_guest_mode_asyncgens():
     import sniffio
 

### CODE REVIEW

**Comment by leojay** [comment]
I don't quite understand the failure. My change only affects pypy. It shouldn't change any behavior in CPython.

**Comment by pquentin** [comment]
> I don't quite understand the test failure. My change only affects pypy. It shouldn't change any behavior in CPython.

Yes, `test_signals` is known to be unreliable.


By the way, you'll find https://github.com/python-trio/trio/pull/1487 interesting too, even though it's Windows specific

**Comment by njsmith** [comment]
Don't worry about that test failure. It's a flaky test.

So to catch everyone up, the background here is: `warn_on_full_buffer=False` was a new feature that CPython added in 3.7. We currently test PyPy's 3.6 build, which isn't expected to have the feature, and it doesn't, so that's fine. But now PyPy is starting to promote their 3.7 build, which is supposed to be compatible with CPython 3.7, so we expect this feature to exist... but it doesn't.

The underlying problem is that PyPy's 3.7 branch *should* support this feature, but doesn't. That's a PyPy bug that we should file upstream with them.

A secondary problem is that we aren't testing against PyPy's 3.7 branch, which is why this wasn't caught sooner. So we should fix that.

And then finally, there's the question of what we should do to work around this issue until PyPy fixes it properly. I'm hesitant to just hard-code that we never use the feature on PyPy, because hopefully this is only a temporary issue, and when PyPy does fix it we want to start using it.

I guess one reasonable approach for now would be to optimistically attempt to use `warn_on_full_buffer_false`, and then backoff if that fails:

```python
try:
    self.old_wakeup_fd = signal.set_wakeup_fd(fd, warn_on_full_buffer=False):
except TypeError as exc:
    # For compatibility with Python 3.6, and some versions of PyPy 3.7 (see issue #1765)
    assert "unexpected keyword argument" in str(exc)
    self.old_wakeup_fd = signal.set_wakeup_fd(fd)
```

leojay:
I googled a little bit. It seems that someone has already reported this bug to pypy several months ago.
https://foss.heptapod.net/pypy/pypy/-/issues/3227

But unfortunately, it's not fixed yet.

@njsmith, you approach would change socket options. I think we shouldn't do that if warn_on_full_buffer is not supported.
https://github.com/python-trio/trio/blob/79ebffe51713466b969ff9270a2f8bf007ceb45e/trio/_core/_wakeup_socketpair.py#L33-L34

So I think HAVE_WARN_ON_FULL_BUFFER has to be set to false at the beginning of the file to make sure that, in pypy,  warn_on_full_buffer_false is not set and socket options are not changed.

leojay:
Hi, I change the code a little bit. This time, we inspect whether warn_on_full_buffer is supported.
Can you please take another look?
Thanks!

leojay:
Friendly ping.

njsmith:
Thanks for the ping!

Change looks reasonable. Can you also:

- add a newsfragment (see `newsfragments/README.rst`)
- update `.travis.yml` so that we're testing the pypy `py3.7` nightly branch, so that we'll know this change is getting tested going forward?

leojay:
Can you please suggest what python version string I should use for the pypy py3.7 nightly branch? I'm new to travis configurations. From this [travis doc](https://docs-staging.travis-ci.com/user/languages/python/#pypy-support), it seems that there is no pypy3.7 support.

pquentin:
@leojay You just need to change `.travis.yml` and add those two lines after the `PYPY_NIGHTLY_BRANCH=py3.6` line:

```
    - language: generic
      env: PYPY_NIGHTLY_BRANCH=py3.6
```

pquentin:
(Clarified my comment above. I'm posting a new one in case you're consuming them as email.)

leojay:
Done. Please take another look.
Thanks!

pquentin:
Unfortunately PyPy 3.7 fails on something else entirely:

```
__________________________ test_guest_mode_asyncgens ___________________________
    @pytest.mark.skipif(buggy_pypy_asyncgens, reason="PyPy 7.2 is buggy")
    def test_guest_mode_asyncgens():
        import sniffio
    
        record = set()
    
        async def agen(label):
            assert sniffio.current_async_library() == label
            try:
                yield 1
            finally:
                library = sniffio.current_async_library()
                try:
                    await sys.modules[library].sleep(0)
                except trio.Cancelled:
                    pass
                record.add((label, library))
    
        async def iterate_in_aio():
            # "trio" gets inherited from our Trio caller if we don't set this
            sniffio.current_async_library_cvar.set("asyncio")
            await agen("asyncio").asend(None)
    
        async def trio_main():
            task = asyncio.ensure_future(iterate_in_aio())
            done_evt = trio.Event()
            task.add_done_callback(lambda _: done_evt.set())
            with trio.fail_after(1):
                await done_evt.wait()
    
            await agen("trio").asend(None)
    
            gc_collect_harder()
    
        # Ensure we don't pollute the thread-level context if run under
        # an asyncio without contextvars support (3.6)
        context.run(aiotrio_run, trio_main, host_uses_signal_set_wakeup_fd=True)
    
>       assert record == {("asyncio", "asyncio"), ("trio", "trio")}
E       AssertionError: assert {('trio', 'trio')} == {('asyncio', ...rio', 'trio')}
E         Extra items in the right set:
E         ('asyncio', 'asyncio')
E         Full diff:
E         - {('asyncio', 'asyncio'), ('trio', 'trio')}
E         + {('trio', 'trio')}
../testenv/site-packages/trio/_core/tests/test_guest_mode.py:543: AssertionError
----------------------------- Captured stdout call -----------------------------
trio_fn finished: Value(None)
----------------------------- Captured stderr call -----------------------------
Exception ignored in: async generator finalizer
  File "/home/travis/build/python-trio/trio/testenv/site-packages/trio/_core/_asyncgens.py", line 86, in finalizer
    source=agen,
ResourceWarning: Async generator 'trio._core.tests.test_guest_mode.test_guest_mode_asyncgens.<locals>.agen' was garbage collected before it had been exhausted. Surround its use in 'async with aclosing(...):' to ensure that it gets cleaned up as soon as you're done using it.
Exception ignored in: <bound method Nursery.__del__ of <trio.Nursery object at 0x00000000083c03a0>>
  File "/home/travis/build/python-trio/trio/testenv/site-packages/trio/_core/_run.py", line 1050, in __del__
    assert not self._children
AssertionError
Exception ignored in: <bound method Nursery.__del__ of <trio.Nursery object at 0x00000000083c0410>>
  File "/home/travis/build/python-trio/trio/testenv/site-packages/trio/_core/_run.py", line 1050, in __del__
    assert not self._children
AssertionError
--- generated xml file: /home/travis/build/python-trio/trio/test-results.xml ---
```

@oremanj `buggy_pypy_asyncgens` seems to imply that the bug would be fixed in 7.3, but it's apparently not (yet?) the case. Do you have any information about that? Should we just bump the version number in `buggy_pypy_asyncgens`? If PyPy fails reliable, we can use `xfail(strict=True)` on all PyPy versions, since we will know when the bug is fixed.

@leojay Sorry about that :)

pquentin:
@oremanj Though I now realize that only one test is failing, so maybe we need a specific skip, since all the other tests work now

oremanj:
`buggy_pypy_asyncgens` is for a bug that existed in pypy 7.2 targeting Python 3.6 (the pypy release version that we test on), and was fixed in pypy 7.3 targeting Python 3.6 (the "pypy nightly 3.6" that we test on). This seems unrelated. Maybe mark the test as `@pytest.mark.xfail(sys.implementation.name == "pypy" and sys.version_info >= (3, 7), reason="async generator issue under investigation")` and I'll look at it when I have a chance? That might not be today, and I don't want to block @leojay for something that's not their fault. :-)

pquentin:
@oremanj That makes sense, thanks!

@leojay Would you be able to add that line to the failing test?

leojay:
Done.

pquentin:
Closing/reopening to rerun the Windows flaky test.

pquentin:
Hm, the IFS LSP build timeouted in the `Waiting for LSP to appear in Winsock catalog` step. Restarting again to see if that helps...

pquentin:
Okay so the Windows IFS LSP build appears to be broken. Sigh.

oremanj:
I've marked that check as not required, so you should be able to commit anyway.

pquentin:
Thanks @oremanj! On a related note, should we mark the 3.9 checks as required?

PR #1780, opened 2020-10-30
Try using an older version of Proxifier

njsmith:
Apparently the latest one no longer uses an LSP?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
             extra_name: ', with Komodia LSP'
           - python: '3.8'
             arch: 'x64'
-            lsp: 'http://www.proxifier.com/download/ProxifierSetup.exe'
+            lsp: 'https://www.proxifier.com/download/legacy/ProxifierSetup342.exe'
             lsp_extract_file: ''
             extra_name: ', with IFS LSP'
           - python: '3.8'


oremanj:
LGTM. Maybe worth adding the version we're using to https://github.com/python-trio/trio-ci-assets so it doesn't go away in the future?

njsmith:
Nice nice, this has fixed the one that was broken, but now Travis is stuck with a 5000-job-long backlog (https://www.traviscistatus.com/). I hear the solution is to [migrate to .com](https://docs.travis-ci.com/user/migrate/open-source-repository-migration)? Maybe I should try poking those buttons and see if anything happens.

njsmith:
> LGTM. Maybe worth adding the version we're using to https://github.com/python-trio/trio-ci-assets so it doesn't go away in the future?

Yeah, probably not a bad idea. (And ditto for the "non-IFS LSP" too while we're at it.)

But now that I look at the repo, I realize that I have no idea how to actually put things there :-). (The README says they're encrypted?)

Any chance you'd be up for uploading these, and also updating the README with notes on how to reproduce what you did?

njsmith:
OK, I think I migrated to travis-ci.com, so I'm going to try restarting the run here...

pquentin:
Thanks! I had to change to update Travis required check in settings to merge this.

njsmith:
@oremanj btw, @piscisaureus put all the LSP artifacts in a gist: https://gist.github.com/piscisaureus/03d8ddfb6d6929dd9beb3a3bd7354c69

PR #1781, opened 2020-10-31
Worker processes

### LINKED ISSUE #5: Missing piece: worker process pools, run_in_worker_process
It'd be nice to have a clean way to dispatch CPU-bound work. Possibly based on `multiprocessing`.

richardsheridan:
This PR provides the missing piece to close #5.  This solution is based on `multiprocessing` and mimics the modern Trio thread cache. (You will recognize a lot of re-used code!)

My motivation for writing this is that most of my projects end up cpu-bound. I iterated on a bunch of ways to interface with `multiprocessing` and `concurrent.futures`, and actually found some nearly trivial solutions that met my current needs. But, I could see some gaps that I thought really needed filling, and just couldn't let go. So, here we are!

The workers have the following features "requested" in #5:
- Bypass GIL for CPU bound work
- Cache processes LIFO
- Minimal internal complexity (modulo `multiprocessing`)

And some extras I feel are very important
- Cross platform
- Cancel seriously misbehaving code via SIGKILL/TerminateProcess
- Convert segfaults and other scary things to catchable errors

If those are the good, then the bad and ugly are:
- Uses threads to make `multiprocessing` async
- Has all the usual multiprocessing caveats (`freeze_support`, pickleable objects only)
- Daemon processes can't have children, but they can talk to each other with e.g. `multiprocessing.Manager`

In a sentence, the case for basing these workers on `multiprocessing` is that it keeps a lot of complexity outside of Trio while offering a set of quirks that users are likely already familiar with. 

From #5 I understand that enthusiasm for (1) including a `to_process.run_sync` API in Trio and (2) basing it on `multiprocessing` has dampened over the years, but I wanted to offer you the first bite at the apple (and use your CI!) before I go off and try to learn all the bits to make a trio-cookiecutter project work.

TODO:
- [x] Write tests
- [x] Expose public API
- [x] Write docs
- [x] Write newsfragment

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -1816,6 +1816,45 @@ to spawn a child thread, and then use a :ref:`memory channel
 
 .. literalinclude:: reference-core/from-thread-example.py
 
+.. _worker_processes:
+
+Worker Processes
+----------------
+
+Given that Python (and CPython in particular) has ongoing difficulties with
+CPU-bound work, Trio provides a method to dispatch synchronous function execution to
+special subprocesses known as "Worker Processes". By default, Trio will create as many
+workers as the system has CPUs (as reported by :func:`os.cpu_count`), allowing fair
+and truly parallel dispatch of CPU-bound work. As with Trio threads, these processes
+are cached in a process pool to minimize latency and resource usage. Despite this,
+executing a function in a process is at best an order of magnitude slower than in
+a thread, and possibly even slower when dealing with large arguments or a cold pool.
+Therefore, we recommend avoiding worker process dispatch for functions with a
+duration of less than about 10 ms.
+
+Unlike threads, subprocesses are strongly isolated from the parent process, which
+allows two important features that cannot be portably implemented in threads:
+
+  - Forceful cancellation: a deadlocked call or infinite loop can be cancelled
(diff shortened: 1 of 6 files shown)


richardsheridan:
I would really appreciate some advice on how to make the segfault test not flaky! :sweat_smile:

asvetlov (approved):

belm0:
interesting PR!

In our Trio experience, the use cases we've had for subprocess have been 1) some synchronous code like a heavy scipy call can't be divided, or otherwise we're playing games with "manual scheduling", i.e.  `trio.sleep()` calls to avoid large step sizes disrupting the responsiveness of other tasks; and 2) need a heterogeneous app to expand beyond one CPU core.

I once did make a primitive subprocess for the app, simply using the `subprocess` module and communicating pickled objects manually over a websocket.  It worked well, though was eventually retired due to broader changes in the system.  If the case comes up again I may give this a try.

Is there anything requiring `WorkerProc` to be in Trio proper, as opposed to a 3rd party module?

auvipy (approved):
this would be  great stuff

richardsheridan:
Closing in favor of [trio-parallel](https://github.com/richardsheridan/trio-parallel). Feel free to reopen someday though!

Issue #1773, opened 2020-10-27
Windows pipe can't receive more than DEFAULT_RECEIVE_SIZE bytes


richardsheridan:
Still working on my multiprocessing wrapper. I thought I was "done" and began stress testing it. I found that I cannot send large objects over the named pipe.  I believe this is a minimal reproducible example:

```python
import trio
from trio._windows_pipes import PipeReceiveStream, PipeSendStream, DEFAULT_RECEIVE_SIZE
from multiprocessing import Process, Pipe


def echo(recv_pipe, send_pipe):
    bytes = recv_pipe.recv_bytes()
    print('received', flush=True)
    send_pipe.send_bytes(bytes)
    print('sent', flush=True)
    send_pipe.send_bytes(b"")


async def main():
    child_recv_pipe, parent_send_pipe = Pipe(duplex=False)
    parent_recv_pipe, child_send_pipe = Pipe(duplex=False)
    proc = Process(target=echo, args=(child_recv_pipe, child_send_pipe), daemon=True)
    proc.start()
    parent_send_stream = PipeSendStream(parent_send_pipe.fileno())
    parent_recv_stream = PipeReceiveStream(parent_recv_pipe.fileno())
    try:
        await parent_send_stream.send_all(bytearray(DEFAULT_RECEIVE_SIZE+10**6))
        async for response in parent_recv_stream:
            print(len(response))
    finally:
        parent_recv_stream._handle_holder.handle = -1
        parent_send_stream._handle_holder.handle = -1
    print("done", flush=True)


if __name__ == "__main__":
    trio.run(main)
```
This raises `OSError: [WinError 234] More data is available` in `parent_send_stream.receive_some()`.  ERROR_MORE_DATA is apparently [not an error despite the name](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile#pipes) and just indicates you should keep reading. I guess this should be special-cased in `wait_overlapped`:

https://github.com/python-trio/trio/blob/02933b7f07df20d3dbb550bbfe2e59b332cdec37/trio/_core/_io_windows.py#L732-L746

Could the fix be as simple as `elif code == ErrorCodes.ERROR_MORE_DATA: pass`? That fixes my test case. It might also be possible to use [PeekNamedPipe](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe) at this point to provide a hint what the next receive size should be on this handle...

[...]

Issue #1767, opened 2020-10-21
TrioInternalError while using a PipeReceiveStream


richardsheridan:
Trying to convert a `multiprocessing.Pipe` to a `PipeSendStream`/`PipeReceiveStream` pair as I try to make my own `multiprocessing` wrapper. This is on windows 10, python 3.8.5, trio 0.17.

```python
import trio
from trio._windows_pipes import PipeReceiveStream, PipeSendStream
from multiprocessing import Process, Pipe, Event


def echo(recv_pipe, send_pipe, event):
    bytes = recv_pipe.recv_bytes()
    print('received', flush=True)
    event.set()
    send_pipe.send_bytes(bytes)
    print('sent', flush=True)


async def main():
    recv_pipe, send_pipe = Pipe()
    event = Event()
    proc = Process(target=echo, args=(recv_pipe, send_pipe, event), daemon=True)
    proc.start()
    send_stream = PipeSendStream(send_pipe.fileno())
    recv_stream = PipeReceiveStream(recv_pipe.fileno())
    await send_stream.send_all(b"Hello World")
    event.wait()
    response = await recv_stream.receive_some()
    print(response, flush=True)
    print("done", flush=True)


if __name__ == "__main__":
    trio.run(main)
```
Trio then crashes when trying to receive the data. The event, or some kind of work/sleep to delay the main process, is crucial to triggering the bug.
```
received
sent
Traceback (most recent call last):
  File "C:\redacted\site-packages\trio\_core\_run.py", line 2068, in unrolled_run
    runner.io_manager.process_events(events)
  File "C:\redacted\site-packages\trio\_core\_io_windows.py", line 516, in process_events
    waiter = self._overlapped_waiters.pop(entry.lpOverlapped)
KeyError: <cdata 'OVERLAPPED *' 0x000001EDFA33D220>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\redacted\site-packages\trio\_core\_run.py", line 1236, in guest_tick
    timeout = self.unrolled_run_next_send.send(self.unrolled_run_gen)
  File "C:\redacted\site-packages\outcome\_sync.py", line 91, in send
    return gen.send(self.value)
  File "C:\redacted\site-packages\trio\_core\_run.py", line 2226, in unrolled_run
    raise TrioInternalError("internal error in Trio - please file a bug!") from exc
trio.TrioInternalError: internal error in Trio - please file a bug!
Tk shutdown. Outcome: Value(None)

[...]

Issue #824, opened 2018-12-27
support for windows named pipes


njsmith:
The subprocess PR (#791) added the core functionality to work with Windows named pipes, but only exposing in them in the limited case of talking to subprocesses. People may also want to use named pipes directly in some cases; they're kind of like windows's equivalent to `AF_UNIX`. (Well, except now `AF_UNIX` sockets are the equivalent to `AF_UNIX` sockets, but only on recent versions of Windows 10, so that won't be universally available for some time yet.) Perhaps we should have client and server helpers.

I was going to try to cite an example of a case where you have to use named pipes to connect to some program, but the only example I know is Discord and [apparently they're using localhost sockets now](https://discordapp.com/developers/docs/topics/rpc)? I know they used to, but I can't find any docs that mention named pipes anymore. So perhaps this is not a great argument for named pipe support being important.

There is also some question to figure out about how much of the named pipe API to expose – the full thing is somewhat complicated, with unidirectional/bidirectional streams, bytes vs. packets, peekable reads, etc., but I suppose just exposing a regular `Stream`-based APi would be a reasonable start

njsmith:
One possible use case would be for setting up handles to talk to a child process, that *aren't* the standard 0, 1, 2 handles.

njsmith:
Well, here's evidence that at least one person does want to use these: https://stackoverflow.com/questions/61519940/async-named-pipes-in-windows-using-trio-and-python/

davidbrochart:
I think that this would be great to have. BTW why does this issue mention Windows only, since named pipes also exist on Unix?

Sxderp:
> BTW why does this issue mention Windows only

I think they're sufficiently different that they need to be handled separately.

Anyway, I ran into this as I wanted to port my MPV control program to Windows. MPV uses unix domain sockets on Linux and named pipes on Windows for IPC. I created a wrapper class that extends trio.abc.Stream for interacting with the pipe. I would not call it "production worthy" but it's enough of a toy example to work for me. Error handling is pretty poor. I wasn't sure how to determine if the other end finished so I just treat the broken pipe as "well that's the end."

I'm also not sure wtf `wait_send_all_might_not_block` is meant to do. I just left it unimplemented and things seem to work.

https://gist.github.com/Sxderp/cf797a1c1f510e0dfe1255c577fef74d

PR #1782, opened 2020-10-31
Use Windows message-oriented pipes to create Channels

richardsheridan:
Relevant to #1773, #1767, #824. Maybe even #959 and #1208?

This PR leverages Windows message-oriented pipes to create objects of type `Channel[Bytes]` (according to the current state of #1208), supplementing the existing `Stream`/`Channel[Byte]` classes. This came about as I was trying to find a way to asyncify `multiprocessing.connection.Pipe` for #1781, however the implementation stands totally separate from that. The internals of `Pipe` are *very* similar to `asyncio.windows_utils.pipe` except that the underlying objects are message-oriented rather than byte-oriented, meaning senders dump a whole message to the OS, and if receivers haven't allocated a big enough buffer, they get an "error", which is really just an invitation to get the rest with another, bigger buffer.

Working towards providing an API for #824, in addition to the class constructors it might make sense to provide `make_pipe_stream` and `make_pipe_channel` from `test_windows_pipes`, since it is quite easy to get a windows pipe handle that is in the wrong state. Or possibly add some checks into the class constructors to ensure the pipe is in the right configuration with `SetNamedPipeHandleState`? Also, users will need clear instructions as to how to get one end of the pipe into a subprocess. (I personally have no idea, I've been letting `multiprocessing` do its magic.)

TODO:
- [x] Write tests
- [ ] Expose public API?
- [ ] Write docs
- [ ] Write newsfragment

### DIFF
--- a/trio/_core/_windows_cffi.py
+++ b/trio/_core/_windows_cffi.py
@@ -357,3 +357,12 @@ def get_pipe_state(handle):
     ):
         raise_winerror()  # pragma: no cover
     return lpState[0]
+
+
+def peek_pipe_message_left(handle):
+    left = ffi.new("LPDWORD")
+    if not kernel32.PeekNamedPipe(
+        _handle(handle), ffi.NULL, 0, ffi.NULL, ffi.NULL, left,
+    ):
+        raise_winerror()  # pragma: no cover
+    return left[0]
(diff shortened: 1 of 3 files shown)


richardsheridan:
I think this PR is ready for review! not sure if CI needs to be refreshed somehow...

richardsheridan:
Woke up in a cold sweat having dreamed of an obscure cancellation bug. I think this is a safe use of the shielded cancel scope, otherwise I'll have to expose a sync function to call `ReadFile` in `_windows_cffi`

PR #1783, opened 2020-10-31
Use PipeChannels or FdStreams to make worker processes natively async

richardsheridan:
This PR uses #1782 to make the workers in #1781 more natively async. This is a separate PR because:

- #1781 is actually feature complete as far as I can tell even with the threads, which don't leak during cancellation.
- It adds substantial platform-dependent complexity into the new `WorkerProc` classes which would previously have been hidden within `multiprocessing.Pipe`.
- It needs to reach into the internals of the handle/fd holder objects to avoid the problems of #1767.

However, it provides two attractive opportunities:

- The pickler is exposed and available to be exchanged for something less picky to transfer objects between processes.
- Avoiding the use of worker threads as much as possible (This speeds up certain microbenchmarks, and ideally it could also use a thread-free replacement for `trio.lowlevel.WaitForSingleObject`)

These commits should be straightforward to rebase on top of #1781 and #1782, if those are accepted. Until then, this PR can just serve as a manifesto for the other PRs.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -1816,6 +1816,45 @@ to spawn a child thread, and then use a :ref:`memory channel
 
 .. literalinclude:: reference-core/from-thread-example.py
 
+.. _worker_processes:
+
+Worker Processes
+----------------
+
+Given that Python (and CPython in particular) has ongoing difficulties with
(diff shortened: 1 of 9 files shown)


richardsheridan:
Closing in favor of [trio-parallel](https://github.com/richardsheridan/trio-parallel). Feel free to reopen someday though!

PR #1797, opened 2020-11-09
Synchronous .close() method for memory channels

takluyver:
This is part of #719. @njsmith asked whether memory channels should have a sync `.close()` method. @oremanj , @lordmauve and myself have said that this would be useful, and no-one has argued against it (please correct me if I've overlooked it!).

My rationale: it's possible to create memory channels and send/receive using synchronous methods, so it seems odd that you need async code to close them. It also makes clean-up after cancellation more difficult, because `await chan.aclose()` will raise `Cancelled` again, so you either need to ensure nothing runs after it, or shield it as described in [Cancellation semantics](https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-semantics).

I've also included synchronous context manager (`with chan:`) support, as I think that's a natural counterpart to the synchronous `.close()` method.

### DIFF
--- a/newsfragments/1797.feature.rst
+++ b/newsfragments/1797.feature.rst
@@ -1,0 +1,2 @@
+Add synchronous ``.close()`` methods and context manager (``with x``) support
+for `.MemorySendChannel` and `.MemoryReceiveChannel`.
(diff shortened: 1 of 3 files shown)


pquentin (approved):
Thanks!

PR #1805, opened 2020-11-16
Avoid creating reference cycles

### LINKED ISSUE #1770: Trio run loop generates cyclic garbage, which can cause more frequent GC stalls on CPython
## symptom
**at a regular period (e.g. 10s or 20s), there are significant stalls (e.g. 30 ms) where it appears the main thread is pre-empted**

In terms of `trio.abc.Instrument`, the lost time appears between `before_task_step` and `after_task_step`, though tasks are doing nothing with threads or time.sleep().  In a single-threaded app with many heterogeneous tasks, the stall events appear to be distributed randomly among the tasks.

I've come up with a minimal program to reproduce the effect.  Note that it is CPU-intensive (a loop of very small `trio.sleep()`), as CPU use > 70% seems necessary to induce the problem.

```python3
import numba
import trio

async def _run():
    T = .000015
    t_start = trio.current_time()

    while True:
        t0 = trio.current_time()
        await trio.sleep(T)
        delta = trio.current_time() - t0
        if delta > .005:
            elapsed = trio.current_time() - t_start
            print(f'{(elapsed) // 60:02.0f}:{elapsed % 60:02.0f}  {delta * 1000:5.1f}')

trio.run(_run)
```

Note that the repro imports `numba`, and that is the spooky part.  With the program as written, observed output (timestamp, stall in ms):
```
$  python repro.py
00:00   29.1
00:11   35.5
00:21   33.4
00:32   33.8
```

However, removing the `numba` import, the stall durations are significantly reduced (6 ms) and at a different period (3 ms):

```
$  ./.venv/bin/python repro.py
00:00    5.2
00:03    6.5
00:05    6.7
00:08    6.4
00:10    6.1
```

And here is importing `numpy` instead of `numba` (10 ms stalls, 4ms period):
```
$  ./.venv/bin/python repro.py
00:00    8.9
00:04   11.8
00:08   10.5
00:11   10.1
00:15   11.2
```

## conditions

seen in Trio 0.17.0

it's observed by multiple people running the repro, across multiple architectures (at least Linux/amd64 and and OS X)

**So far it's specific to Trio.  I've tried equivalent loops using `asyncio` and `time.sleep()`, which do not exhibit the problem.**

original bug filed to numba:  https://github.com/numba/numba/issues/6347

## theories

  * T1. **kernel scheduling artifact?** - is this just an expected effect as a process approaches 100% CPU, as the OS/kernel need cycles to do internal work?  But why doesn't the problem show up with asyncio or time.sleep() implementations?  Why does the stall period depend on unused package imports?
  * T2. **weird GIL interaction among main thread and one of Trio's background threads?** - The GIL is essentially a scheduler, I've read.
  * ...?

## next steps

  * [x] try earlier versions of Trio - result: no change (tested back to 0.11.0)
  * [ ] ...

njsmith:
When put together, this change +
https://github.com/python-trio/outcome/pull/29 should fix #1770

Note that the tests on this PR are expected to fail until that fix to
`outcome` lands and has been released, since this PR includes a full
end-to-end test.

### DIFF
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -114,6 +114,9 @@ def _filter_impl(handler, root_exc):
     preserved = set()
     new_root_exc = filter_tree(root_exc, preserved)
     push_tb_down(None, root_exc, preserved)
+    # Delete the local functions avoid a reference cycle (see
+    # test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage)
+    del filter_tree, push_tb_down
     return new_root_exc
 
 
@@ -132,6 +135,7 @@ class MultiErrorCatcher:
     def __exit__(self, etype, exc, tb):
         if exc is not None:
             filtered_exc = MultiError.filter(self._handler, exc)
+
             if filtered_exc is exc:
                 # Let the interpreter re-raise it
                 return False
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -114,7 +114,7 @@ def _filter_impl(handler, root_exc):
     preserved = set()
     new_root_exc = filter_tree(root_exc, preserved)
     push_tb_down(None, root_exc, preserved)
-    # Delete the local functions avoid a reference cycle (see
+    # Delete the local functions to avoid a reference cycle (see
     # test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage)
     del filter_tree, push_tb_down
     return new_root_exc
--- a/trio/_core/tests/test_run.py
+++ b/trio/_core/tests/test_run.py
@@ -9,6 +9,7 @@ import warnings
 from contextlib import contextmanager, ExitStack
 from math import inf
 from textwrap import dedent
+import gc
 
 import attr
 import outcome
@@ -2195,3 +2196,33 @@ async def test_cancel_scope_deadline_duplicates():
             cscope.deadline = now + 9998
             cscope.deadline = now + 9999
         await sleep(0.01)
+
+
+@pytest.mark.skipif(
+    sys.implementation.name != "cpython", reason="Only makes sense with refcounting GC"
+)
+async def test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage():
+    # https://github.com/python-trio/trio/issues/1770
+    gc.collect()
+
+    async def do_a_cancel():
+        with _core.CancelScope() as cscope:
+            cscope.cancel()
+            await sleep_forever()
+
+    old_flags = gc.get_debug()
+    try:
+        gc.collect()
+        gc.set_debug(gc.DEBUG_SAVEALL)
+
+        await do_a_cancel()
+        await do_a_cancel()
+
+        async with _core.open_nursery() as nursery:
+            nursery.start_soon(do_a_cancel)
+
+        gc.collect()
+        assert not gc.garbage
+    finally:
+        gc.set_debug(old_flags)
+        gc.garbage.clear()
--- a/trio/_core/tests/test_run.py
+++ b/trio/_core/tests/test_run.py
@@ -2226,3 +2226,26 @@ async def test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage():
     finally:
         gc.set_debug(old_flags)
         gc.garbage.clear()
+
+
+@pytest.mark.skipif(
+    sys.implementation.name != "cpython", reason="Only makes sense with refcounting GC"
+)
+async def test_nursery_cancel_doesnt_create_cyclic_garbage():
+    # https://github.com/python-trio/trio/issues/1770#issuecomment-730229423
+    gc.collect()
+
+    old_flags = gc.get_debug()
+    try:
+        for i in range(3):
+            async with _core.open_nursery() as nursery:
+                gc.collect()
+                gc.set_debug(gc.DEBUG_LEAK)
+                nursery.cancel_scope.cancel()
+
+            gc.collect()
+            gc.set_debug(0)
+            assert not gc.garbage
+    finally:
+        gc.set_debug(old_flags)
+        gc.garbage.clear()
--- a/newsfragments/1770.bugfix.rst
+++ b/newsfragments/1770.bugfix.rst
@@ -1,0 +1,3 @@
+Trio now avoids creating cyclic garbage as often. This should have a
+minimal impact on most programs, but can slightly reduce how often the
+cycle collector GC runs on CPython, which can reduce latency spikes.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -937,7 +937,12 @@ class Nursery(metaclass=NoPublicConstructor):
         popped = self._parent_task._child_nurseries.pop()
         assert popped is self
         if self._pending_excs:
-            return MultiError(self._pending_excs)
+            try:
+                return MultiError(self._pending_excs)
+            finally:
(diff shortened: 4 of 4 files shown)


trio/_core/_multierror.py
```diff
@@ -114,6 +114,9 @@ def push_tb_down(tb, exc, preserved):
     preserved = set()
     new_root_exc = filter_tree(root_exc, preserved)
     push_tb_down(None, root_exc, preserved)
+    # Delete the local functions avoid a reference cycle (see
```
belm0, line 117: "to avoid a..."

trio/_core/_multierror.py
```diff
@@ -114,6 +114,9 @@ def push_tb_down(tb, exc, preserved):
...
+    # test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage)
```
belm0, line 118: How important is it to reference the test?  Seems like this comment could get stale easily.

trio/_core/tests/test_run.py
```diff
@@ -2195,3 +2196,33 @@ async def test_cancel_scope_deadline_duplicates():
             cscope.deadline = now + 9998
             cscope.deadline = now + 9999
         await sleep(0.01)
+
+
+@pytest.mark.skipif(
+    sys.implementation.name != "cpython", reason="Only makes sense with refcounting GC"
+)
+async def test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage():
```
belm0, line 2204: 👍 

Would it be worth adding a test to catch garbage regressions at a higher level, say `sleep(0)`?  Even if that case isn't at 0 now, we could assert on the count being below what's observed now.

belm0 (approved):

njsmith:
Note: in Trio there are several places that have the same pattern as `Error.unwrap`, where they stash an exception in a local variable and then raise it, which should create a reference cycle. Specifically, the `__exit__` handler for `MultiError.catch`, `CancelScope`, and `open_nursery`. I *think* this means that they should probably get a similar fix. But I tried extending the test in this PR to confirm this, and I couldn't get it to fail. So something funky is going on that I don't totally understand yet. (This PR should be enough to fix the original issue in #1770 though. @belm0 if you have a chance to try the modified `outcome` + this PR and can confirm that it fixes the actual production issue you were seeing that would be great.)

belm0:
> Note: in Trio there are several places that have the same pattern as `Error.unwrap`, where they stash an exception in a local variable and then raise it, which should create a reference cycle. Specifically, the `__exit__` handler for `MultiError.catch`, `CancelScope`, and `open_nursery`. I _think_ this means that they should probably get a similar fix. But I tried extending the test in this PR to confirm this, and I couldn't get it to fail. So something funky is going on that I don't totally understand yet. (This PR should be enough to fix the original issue in #1770 though. @belm0 if you have a chance to try the modified `outcome` + this PR and can confirm that it fixes the actual production issue you were seeing that would be great.)

In https://github.com/python-trio/trio/issues/1770#issuecomment-717882202 there is a self-contained example that counts garbage across sleep().  I'd expect it to work within a test 🤔 

I'll try to test staged trio + outcome changes against that test and our app in a day or so.

njsmith:
> In [#1770 (comment)](https://github.com/python-trio/trio/issues/1770#issuecomment-717882202) there is a self-contained example that counts garbage across sleep(). I'd expect it to work within a test thinking

Yeah, the test included in this PR was derived from that example, and the test does indeed fail without the fixes in these two PRs.

The weird thing is that I tried locally adding some extra versions of the test to catch cases that I thought should *also* fail, like a nursery where one task is cancelled and another raises an exception, and in *those* cases I couldn't get it to fail or figure out why it wasn't failing.

I guess if anyone is desperately curious you could check out the video at https://www.twitch.tv/videos/804632095 since I did this all on stream (NB that link will probably stop working in a few weeks, following twitch's garbage collection algorithm). But I'll probably get back to it later, just ran out of steam for today.

trio/_core/_multierror.py
pquentin: Good question! It's possibly still worth the risk, as stale comments can be fixed, but losed context is harder to fix!

pquentin:
`outcome` was released, and this pull request now tests successfully against it.

@njsmith Should we request `outcome>=1.1.0` in setup.py? It's not mandatory, but without that version the newsfragment would no longer be true, I guess.

belm0:
> @njsmith Should we request `outcome>=1.1.0` in setup.py? It's not mandatory, but without that version the newsfragment would no longer be true, I guess.

I'd opt for just noting in the newsfragment that Trio is doing it's part, and together with the recent `outcome` package addresses the worst of the GC issue.

belm0:
From the standalone test at https://github.com/python-trio/trio/issues/1770#issuecomment-717882202, here are the garbage items per sleep count for various cases:

  * no fixes - 36
  * both fixes - 0
  * without outcome fix - 28
  * without trio fix - 8

However, even with both fixes I can create significant garbage (about 40 items) by way of a cancelled background task:
```python3
    nursery.start_soon(trio.sleep_forever)
    await trio.sleep(T/N)
    nursery.cancel_scope.cancel()
```

richardsheridan:
@belm0 does this test code reproduce what you are talking about here?
```python
import gc
import trio

import pprint
import objgraph


async def amain():
    for _ in range(3):
            gc.collect()
            gc.set_debug(gc.DEBUG_LEAK)
            n.cancel_scope.cancel()
        gc.collect()
        gc.set_debug(0)
        print(len(gc.garbage))
        if _ < 9:
            gc.garbage.clear()


if __name__ == "__main__":
    trio.run(amain)
    gc.collect()
    pprint.pprint(gc.garbage)
    stuff = [x for x in gc.garbage if "Cancelled" in repr(x)]
    objgraph.show_backrefs(stuff,extra_ignore=(id(gc.garbage),),max_depth=10,filename="nursery_cycle.png")
```

This is the reference cycle I observe with that code:
nursery_cycle
(how do you scale images on github? this is way too big)
The problem seems to originate in this bit of code.
https://github.com/python-trio/trio/blob/e113e561ab6af788bb15947f147d69e8d04fcf2d/trio/_core/_run.py#L932-L940

I've been able to break the cycle by doing:
```python

        if self._pending_excs:
            try:
                return MultiError(self._pending_excs)
            finally:
                del self._pending_excs
```
or
```python

        if self._pending_excs:
            try:
                return MultiError(self._pending_excs)
            finally:
                del popped, self
```

belm0:
@richardsheridan thank you for your contributions on the GC issue

I think you'll find some other cycle cases by adding these inside the cancelled nursery:
```python3
nursery.start_soon(trio.sleep_forever)
await trio.sleep(.01)
```

richardsheridan:
Adding those doesn't seem to make any more garbage for me with the fixes I outlined above. I guess the "more garbage" you see is more objects kept alive by connection to the same cycle. (they all merge up in aexit)

belm0:
> Adding those doesn't seem to make any more garbage for me with the fixes I outlined above.

confirmed, thank you

> ```python3
>        if self._pending_excs:
>           try:
>                return MultiError(self._pending_excs)
>            finally:
>               del self._pending_excs
>```

For what it's worth, the try/finally may not be needed.  I think `self._pending_excs` can be stored in a local and deleted upfront.

belm0:
Shall we merge this PR as is, or add the additional fix by @richardsheridan with corresponding test?  I'm happy to commit the latter to this PR.

pquentin (approved):
Thanks!

PR #1811, opened 2020-12-06
Migrate Alpine, Fedora, FreeBSD CIs to SourceHut

McSinyx:
This moves:

- FreeBSD 12.1 on Travis to latest FreeBSD (currently 12.x) on builds.sr.ht (now takes 4 min)
- Fedora 32 on Travis to Fedora Rawhide on builds.sr.ht (now takes 2 min)
- Alpine (unspecified version) on GitHub Action to latest Alpine release on builds.sr.ht (now takes 3 min)

The Fedora image maintainer is a bit busy lately, so Rawhide is still 32 at the moment, but probably we should track `latest` instead.

There's a rule set on the GitHub repo (I think) requiring the Alpine job so I think an admin needs to turn it off manually, since I can't find it in the source tree.

/cc GH-1785


njsmith:
Huh, that's pretty impressive!

I assume you need to switch the `sources:` line now?

It would also be nice if we could get useful names on the codecov reports – it looks like the sr.ht builds are the three that are listed as "-unknown" here: https://codecov.io/gh/python-trio/trio/commit/1fe87396178140f9348a21f1595a9301e3ade1aa/build

(This is mostly only relevant for debugging codecov issues, but unfortunately those happen a lot. I think it just requires adding `JOB_NAME: <some sort of name>` to the environment blocks?

njsmith:
Also at the top of `ci.sh`, right before the `env | sort` line, can you add a call to `uname -a`, so the logs will show what kernel version etc. we're running on?

McSinyx:
> I assume you need to switch the `sources:` line now?

Yes I do :smile:  I've also added `JOB_NAME` to each YAML and `uname -a` to the script, let's see if it reports the name correctly.

pquentin:
Another discussion that needs to happen is that SourceHut isn't free for OSS, and will require at least [$20/year](https://sourcehut.org/pricing/) sometime [in the future](https://sourcehut.org/alpha-details/). It's not a huge amount of money for the service they provide, but we should still pay that, right?

McSinyx:
@pquentin, per the original discussion on Gitter, I'm willing to donate that money (mainly because I want SourceHut to grow), at least in the next 5 years (i.e. until my card expires) so it wouldn't be a problem.  After that, to quote @njsmith,

> eventually we should figure out how to collect donations etc instead of just paying for things out of pocket

pquentin:
I've removed "Alpine" from the required checks, will wait a bit before requiring the SourceHut checks to avoid failing the other pull requests.

Thanks a lot for this! It's awesome

McSinyx:
Thanks for the merge!  BTW I assume that with my newly gained privilege, it'd be safe for me to merge dependabot bumps that look safe to me?

pquentin:
@McSinyx Sure! You can merge anything you want. But dependabot is set to merge automatically. It currently takes a long time because of Travis delays

McSinyx:
Thanks, I figure that some tests are flaky and I only need to rerun the jobs.  I'm a bit sad that there's only an option to rerun all GHA jobs which feels kind of wasteful :smile:

pquentin:
See https://github.com/python-trio/trio/issues/200 for our approach to flaky tests, in short we prefer to close/reopen anyway :)

njsmith:
Also we really need to fix those flaky tests :-)

PR #1827, opened 2020-12-18
Test Python development versions in GitHub Actions

pquentin:
This is actually different from Travis as we're not testing nightly builds, but alpha versions. Should we use the deadsnakes nightly builds instead?

Also, 3.10 does not work yet, due to https://github.com/benjaminp/six/issues/341.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.6', '3.7', '3.8', '3.9.0-rc.2']
+        python: ['3.6', '3.7', '3.8', '3.9']
         arch: ['x86', 'x64']
         lsp: ['']
         lsp_extract_file: ['']
@@ -59,7 +59,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.6', '3.7', '3.8', '3.9.0-rc.2']
+        python: ['3.6', '3.7', '3.8', '3.9']
         check_formatting: ['0']
         extra_name: ['']
         include:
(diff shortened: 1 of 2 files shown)


njsmith:
I think for 3.x versions that have already been released, we want the nightlies? I don't think there are alphas/etc. for 3.x.y point releases, after 3.x.0 comes out.

For the next release, 3.10 currently, I guess alphas/betas/rcs do have some argument in their favor -- in particular, they might reduce the frequency of weird breakage. But if we have to edit the .yml every time a new alpha comes out, we aren't going to do it, so we'll end up not actually testing the new versions... not sure if there's any way to avoid that.

altendky:
I don't know anything about nightlies but if I did my testing right (I don't recall in detail) then this will progress from alphas through release anyways.

https://github.com/pytest-dev/pytest-twisted/blob/8a955b66bb22c0e2f09fc5cdaed241dd93e53a75/.github/workflows/ci.yml#L146
```yaml
        uses: actions/setup-python@v2
        with:
          python-version: '${{ matrix.python.action }}.0-alpha - ${{ matrix.python.action }}.X'
          architecture: x64
```

njsmith:
Regardless of what we end up doing for 3.10, this looks like a definite improvement over what we have now. I would hit merge now, but it looks like merging is currently blocked though because some rc.2 builds are expected but missing? I could go disable them but I'm not quite sure what's going on so I'd be more comfortable delegating that to @pquentin, assuming your the one who was in the middle of changing them and knows what the current status is :-)

pquentin:
Right I also have a unrelated commit where I change 3.9.0-rc.2 to 3.9. Sorry I'm currently on holidays and have limited time to fix this. Feel free to take over.

njsmith:
Right then, let's do it. (And @pquentin, hope you had/are having a great holiday :-))

PR #1832, opened 2020-12-24
avoid cubic AFD cancellation scaling with AFDGroups

### LINKED ISSUE #1280: Work around Windows kernel bad asymptotic behavior when there are too many operations outstanding on the same handle
So [as discussed over here](https://github.com/python-trio/trio/issues/52#issuecomment-548215128), apparently the `CancelIoEx` operation on Windows becomes super-slow when you have a lot of operations outstanding on the same handle. And #1269 makes it so that every `wait_readable`/`wait_writable` call issues an operation on the same handle. So cancelling a bunch of these calls becomes like O(n^3) or something.

And we also issue `CancelIoEx` whenever we transition from 1->2 or 2->1 waiters on the same socket, so this can also be hit during normal usage, without Trio-level cancellation. (I think a typical case would be when you have a socket where one task is blocked for a long time in `wait_readable`, while another task is sending as fast as it can, so it keeps going in and out of `wait_writable`.)

The overhead isn't really noticeable up to like 1000 sockets or so, but eventually it dominates: at 10k sockets, cancelling a `wait_readable` is like 10x slower, and at 30k sockets, it's like 100x slower, getting into multiple ms per call.

@piscisaureus is the one who pointed this out, and he says that he's found that the cost is based on the number of AFD_POLL operations issued *per AFD helper handle*. So as a workaround, we could open multiple AFD helper handles, and spread our AFD_POLL operations around over them. This will involve some annoying bookkeeping, but it shouldn't be *too* bad – basically we'll want to keep a pool of AFD handles, and for each one set a limit on how many total simultaneous operations we'll allow. We'll keep a pool of handles that are below their limit, and pull from there if we can; and if we run out, we'll allocate a new handle.

There's a script in `notes-to-self/socket-scaling.py` to measure the problem, and that can be used to validate our fix.

richardsheridan:
Fixes #1280. By placing the AFD handle within an `AFDGroup` class that keeps track of the number of associated `AFDPollOp`s and giving each op a reference to its group, we can use a `SortedList` to grab the largest not-full group and therefore AFD handle with a simple pop. The bookkeeping is a little unencapsulated, but the bulk of the changes are inside of `WindowsIOManager._refresh_afd` anyway. 

My own benchmarking showed negligible overhead and an optimal `MAX_AFD_GROUP_SIZE` of 500 to avoid scaling issues but YMMV.

### DIFF
--- a/newsfragments/1832.bugfix.rst
+++ b/newsfragments/1832.bugfix.rst
@@ -1,1 +1,0 @@
-Avoid O(n**3) scaling of windows socket cancellation
--- a/newsfragments/1832.bugfix.rst
+++ b/newsfragments/1832.bugfix.rst
@@ -1,0 +1,1 @@
+Avoid O(n**3) scaling of windows socket cancellation
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -628,6 +628,7 @@ class WindowsIOManager:
                 afd_group = AFDGroup(0, _afd_helper_handle())
                 self._register_with_iocp(afd_group.handle, CKeys.AFD_POLL)
                 self._all_afd_handles.append(afd_group.handle)
+            self._vacant_afd_groups.add(afd_group)
 
             lpOverlapped = ffi.new("LPOVERLAPPED")
 
@@ -666,8 +667,8 @@ class WindowsIOManager:
             waiters.current_op = op
             self._afd_ops[lpOverlapped] = op
             afd_group.size += 1
-            if afd_group.size < MAX_AFD_GROUP_SIZE:
-                self._vacant_afd_groups.add(afd_group)
+            if afd_group.size >= MAX_AFD_GROUP_SIZE:
+                self._vacant_afd_groups.remove(afd_group)
 
     async def _afd_poll(self, sock, mode):
         base_handle = _get_base_socket(sock)
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -6,6 +6,7 @@ import sys
 from typing import TYPE_CHECKING
 
 import attr
+from sortedcontainers import SortedList
 
 from .. import _core
 from ._run import _public
@@ -343,6 +344,24 @@ class AFDPollOp:
     lpOverlapped = attr.ib()
     poll_info = attr.ib()
     waiters = attr.ib()
+    afd_group = attr.ib()
+
+
+# With the new IOCP-based loop, the cost of scheduling is constant, which is
+# good. But, we find that the cost of cancelling a single wait_readable
+# appears to grow like O(n**2) or so in the number of outstanding
+# wait_readables. This is bad -- it means that cancelling all of the
+# outstanding operations here is something like O(n**3)! To avoid this, we
+# multiple AFD helper handles for each set of MAX_AFD_GROUP_SIZE and distribute
+# AFD_POLL operations across them.
+MAX_AFD_GROUP_SIZE = 500  # at 1000, the cubic scaling is just starting to bite
+
+
+@attr.s(slots=True)
+class AFDGroup:
+    size = attr.ib()
+    afd_id = attr.ib()
+    afd_handle = attr.ib()
 
 
 @attr.s(slots=True, eq=False, frozen=True)
@@ -374,15 +393,14 @@ class WindowsIOManager:
         # touches to safe values up front, before we do anything that can
         # fail.
         self._iocp = None
-        self._afd = None
+        self._sorted_afd_groups = None
 
         self._iocp = _check(
             kernel32.CreateIoCompletionPort(INVALID_HANDLE_VALUE, ffi.NULL, 0, 0)
         )
         self._events = ffi.new("OVERLAPPED_ENTRY[]", MAX_EVENTS)
 
-        self._afd = _afd_helper_handle()
-        self._register_with_iocp(self._afd, CKeys.AFD_POLL)
+        self._sorted_afd_groups = SortedList()
         # {lpOverlapped: AFDPollOp}
         self._afd_ops = {}
         # {socket handle: AFDWaiters}
@@ -439,10 +457,11 @@ class WindowsIOManager:
                 self._iocp = None
                 _check(kernel32.CloseHandle(iocp))
         finally:
-            if self._afd is not None:
-                afd = self._afd
-                self._afd = None
-                _check(kernel32.CloseHandle(afd))
+            if self._sorted_afd_groups is not None:
+                afd_groups = self._sorted_afd_groups
+                self._sorted_afd_groups = None
+                for afd_group in afd_groups:
+                    _check(kernel32.CloseHandle(afd_group.afd_handle))
 
     def __del__(self):
         self.close()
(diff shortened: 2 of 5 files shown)


njsmith:
Some high-level comments after a quick read:

We don't really need to care about picking the exact right handle. All we really care about is finding *some* handle that has available capacity. So maybe it would be simpler to keep two `set`s: one of all handles (so we can close them at the end), and one of the handles that have at least one slot available? I doubt `SortedList` adds major overhead in practice compared to all the syscalls, but it is substantially heavier than a `set`.

Speaking of closing them at the end: the current code doesn't track groups that are at capacity. In practice I guess this *mostly* doesn't matter because by the time trio exits, all the groups should be empty, but I think it'd be simpler to keep a set of *all* groups to use at the end, so we don't have to think hard about this.

I feel like we shouldn't unconditionally close AFD handles as soon as they have zero users, because if an app is rapidly cycling between 0 and 1 socket in use then they'll end up constantly opening/closing AFD handles. I kind of wonder if we should just... never close AFD handles at all, until trio exits? AFAIK having a few unused handles hanging around shouldn't cause any problems, and the worst case would still be bounded by the maximum number of sockets the program has used simultaneously divided by the max group size.

We should make sure we have a test that actually forces the use of multiple groups :-)

njsmith:
Oh and don't forget your newsfragment :-)

richardsheridan:
I toyed with using a `dict` or `Counter` instead of a set and the extra attr.s class, but leaving it this way makes it easy to refactor back to a `SortedList` if it becomes necessary somehow.

trio/_core/_io_windows.py
```diff
@@ -632,9 +667,12 @@ def _refresh_afd(self, base_handle):
                     # Do this last, because it could raise.
                     wake_all(waiters, exc)
                     return
-            op = AFDPollOp(lpOverlapped, poll_info, waiters)
+            op = AFDPollOp(lpOverlapped, poll_info, waiters, afd_group)
             waiters.current_op = op
             self._afd_ops[lpOverlapped] = op
+            afd_group.size += 1
+            if afd_group.size < MAX_AFD_GROUP_SIZE:
+                self._vacant_afd_groups.add(afd_group)
```
njsmith, line 675: Hmm, I think there's a real bug here, not just a style nitpick like my other comments :-). If the `DeviceIoControl` call fails and the function exits early, then we can "lose" the `afd_group` -- it still has vacancies, but we won't put it back in `self._vacant_afd_groups`. I think the clearest way would be to leave the group in `_vacant_afd_groups` until we're ready to increment `.size`.

It's annoying there's no equivalent of `pop` that *doesn't* remove the item. I guess the choices are `afd_group = next(iter(self._vacant_afd_groups))` or `afd_group = self._vacant_afd_groups.pop(); self._vacant_afd_groups.add(afd_group)`.

trio/tests/test_socket.py
```diff
@@ -990,3 +990,20 @@ async def receiver():
...
+    sockets = []
+    for _ in range(total // 2):
+        a, b = stdlib_socket.socketpair()
+        sockets += [a, b]
+    async with _core.open_nursery() as nursery:
+        for s in sockets:
+            nursery.start_soon(_core.wait_readable, s)
+        await _core.wait_all_tasks_blocked()
+        for _ in range(1000):
+            await _core.cancel_shielded_checkpoint()
```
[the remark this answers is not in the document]
richardsheridan: This was copied verbatim from the benchmark script, but it appears to truly do nothing, so I removed it.

njsmith:
Test failure is #1838, I'm pretty sure.

trio/tests/test_socket.py
```diff
@@ -990,3 +990,26 @@ async def receiver():
...
+
+
+async def test_many_sockets():
+    total = 5000  # Must be more than MAX_AFD_GROUP_SIZE
+    sockets = []
+    for x in range(total // 2):
+        try:
+            a, b = stdlib_socket.socketpair()
+        except OSError as e:
+            if e.errno != 24:
```
njsmith, line 1002: Is this an error you hit on Linux? It looks like on Linux this is `errno.EMFILE`, which would make sense, since `EMFILE` means "you hit the limit on the number of file descriptors that a single process is allowed to have open". (There's also the confusingly-similar `ENFILE`, which means "you hit some kind of systemwide limit on the number of file descriptors".) So at the least, we should change this to `errno.EMFILE` instead of having a magic integer in the source code :-). And probably add `ENFILE` as an option too. Or maybe simplify to

```
except OSError as exc:
    assert exc.errno in (errno.EMFILE, errno.ENFILE)
    break
```

BTW, a handy trick: if you have some errno number and you want to figure out what the error code is, you can do:

```python
>>> import errno
>>> errno.errorcode[24]
'EMFILE'
```

newsfragments/1832.bugfix.rst
```diff
@@ -0,0 +1 @@
+Previously, on Windows, Trio programs using thousands of sockets at the same time could trigger extreme slowdowns in the Windows kernel. Now, Trio works around this issue, so you should be able to use as many sockets as you want.
```
njsmith, line 1: I rewrote this, lmk what you think

trio/tests/test_socket.py
```diff
@@ -990,3 +990,26 @@ async def receiver():
...
+        sockets += [a, b]
+    async with _core.open_nursery() as nursery:
+        for s in sockets:
+            nursery.start_soon(_core.wait_readable, s)
+        await _core.wait_all_tasks_blocked()
+        nursery.cancel_scope.cancel()
+    for sock in sockets:
+        sock.close()
+    if x != total // 2 - 1:
+        pytest.xfail(f"Unable to open more than {(x-1)*2} sockets.")
```
njsmith, line 1015: Let's not use `xfail` here -- to me `xfail` means "there's a known bug in the package", and this isn't a bug in trio at all, it's just a limitation of the test environment. We could do a `print` just for info and debugging?

newsfragments/1832.bugfix.rst
richardsheridan: Looks good to me!

njsmith:
Closing/reopening to re-run CI, now that #1838 is merged.

njsmith:
Huh. Apparently the close/reopen trick doesn't work for sr.ht builds. That's annoying.

njsmith:
Huh! And apparently sr.ht tests the PR branch directly, instead of PR-branch-merged-with-latest-master like every other CI system does. That's super weird. I guess we need to manually merge in master here to pick up the fixes?

richardsheridan:
Do you prefer merge over rebase? I might go back and squash a few lame commits otherwise.

njsmith:
@richardsheridan if I were doing it I would probably just merge from master because it would be the least work for me, but if you want to do it then feel free to do whatever you like best :-)

richardsheridan:
Hmm that doesn't seem to have worked @njsmith

njsmith:
@richardsheridan It looks like your rebased commits still branch off from master at https://github.com/python-trio/trio/commit/a34ee840ba37de5c966f8903c6fd8328838a3481, which is before the fix sr.ht fix. Maybe you need to fetch the latest `master` and then rebase onto it?

richardsheridan:
forgot to fetch before rebase before...

njsmith:
Easy mistake to make :-)

And phew, that was a bit of a journey, but you did it! Thank you so much :-)

PR #1834, opened 2020-12-24
Pool WaitForSingleObject calls into minimal threads

### LINKED ISSUE #1271: Race condition in trio.hazmat.WaitForSingleObject
https://github.com/python-trio/trio/pull/575/files#diff-ed34813c30595dc83bb2a66e0973e22dR54

This appears to have a race condition. The new thread might start waiting only after the handle has been closed. This could be fixed by making the thread close the `cancel_handle`.

_Originally posted by @GSPP in https://github.com/python-trio/trio/issues/233#issuecomment-546681264_

richardsheridan:
Fixes #1271, but also optimizes `WaitForSingleObject` calls. This fulfills a wish I had in #1783 to reduce the number of threads needed during async subprocess usage on Windows. 

UPDATE: The following was true until 783d093, but removed in a3a4f10. The current implementation is equivalent to "Pool per process". I preserved the benchmarking code [here](https://github.com/richardsheridan/trio/tree/wait_pool_with_benchmark).

Right now it contains some awkward monkeypatching routines to select from multiple implementations on the fly for benchmarking.  The new implementation options are:

1.  Win32: a simple wrapping of `RegisterWaitForSingleObject` and `UnregisterWait` as you suggested in #233. This is the lightest weight and most performant option, and used by [asyncio](https://github.com/python/cpython/blob/32bd68c839adb7b42af12366ab0892303115d1d1/Modules/overlapped.c#L350), [libuv](https://github.com/libuv/libuv/blob/30ff5bf2161257921f3a3ce5655804f7cb282aa9/src/win/process.c#L1155) and [tokio](https://github.com/tokio-rs/tokio/blob/1eefbc250ad49d9086bdb2b8be4404bd445f63fb/tokio/src/process/windows.rs#L119) (with the same wait flags, too). The objections to it are `ffi.callback()` breaks on some platforms (none of which are Windows), `UnregisterWait` is difficult to understand (by reimplementing it in Trio, I am pretty comfortable answering any questions), and we don't control the threads or thread pool (valid, and the pool has a default limit of 500 threads which could bite at any time).
2.  Pool per run: a `WaitPool` chops up wait requests into `WaitGroup`s which implicitly have-a (unstructured) background Trio task that jumps in and out of the Trio thread cache to alternately wait and drain signaled handles from the group. The largest not-full `WaitGroup` is identified using a `SortedList` so that we minimize the number of `WaitGroup`s under normal circumstances. `WaitPool` does not attempt to merge `WaitGroup`s at any point, the handles simply drain as completed. Benchmarking shows this has significantly higher high-water-mark memory usage and lower throughput compared to...
3.  Pool per process: the same `WaitPool` logic makes `WaitGroup`s which correspond to background threads from the Trio cache but otherwise behave identically to the above tasks. Compared to Win32, this method is about 50% slower and heavier in terms of throughput and high water mark memory respectively, but compared to pooling per run with tasks, it is better by a factor of 2 or 3. Everything has to be carefully locked, but the lock contention shouldn't be much worse than the event-loop-contention of pool per run, unless there are many many separate threads with Trio runs all waiting on handles.

Also included is a benchmark script in the notes-to-self folder. Even if the pooling options turn out to be not desirable for Trio, I would propose rolling the implementations into the benchmark script directly so that if someone someday does turn out to be memory-limited by the current thread-per-handle approach, there will be a quick place to go with drop-in replacements available.


trio/_wait_for_object.py
```diff
@@ -1,62 +1,200 @@
...
+        # This with block prevents a race with the spawning thread on init
+        with WAIT_POOL.lock:
+            n_handles = len(wait_group) - 1
+        while n_handles:  # quit thread when only wake handle remains
+            signaled_handle = WaitForMultipleObjects_sync(*wait_group)
+            with WAIT_POOL.lock:
+                if signaled_handle is wait_group[0]:
+                    # Reset is OK here because only this thread waits on the event
+                    kernel32.ResetEvent(wait_group[0])
+                elif signaled_handle in WAIT_POOL:
```
pquentin, line 125: For what it's worth, codecov complains that we never evaluate this condition to False

richardsheridan: Yeah, I think it needs a pragma: no branch, unless I can figure out how to trigger the race mentioned in the comment in the next block

richardsheridan:
This PR is dramatically streamlined as of a3a4f10, so if @pquentin, @njsmith or @oremanj ever wanted to do a deep read of a "finished product", it is available now!

honglei:
Any news?

richardsheridan:
To be honest, I didn't think anyone was watching this one. Do you have a specific use-case? It might add some motivation for someone to (do an easy rebase and) merge the feature.

honglei:
@richardsheridan  I was directed to this issue from https://github.com/python-trio/trio/issues/1271 , so the problom was solved in #1271 right now?

richardsheridan:
Yes, this PR surely fixes #1271, as each wait group thread cleans up its own wakeup event (equivalent to cancel handle from the current implementation). I think the hesitation to merge comes from the possibility that some other subtle bug may be introduced from the extra complexity.

pquentin:
I merged the master branch. But I'm only the janitor, I don't know the internals of Trio well enough to help here, sorry!

PR #1846, opened 2020-12-29
Replaced .format() with f-strings in the tutorial examples

rednafi:
Since the lowest version we're now supporting is Python3.6, I took the liberty to replace the .format() style strings in the tutorial with f-strings for enhanced readability.


njsmith:
Hey, good idea, thanks!

PR #1848, opened 2020-12-30
Fstring reference core

rednafi:
F-stringified reference-core doc examples.


pquentin (approved):
Thanks! I wonder if pyupgrade can handle those automatically.

PR #1852, opened 2021-01-04
Mention Nursery.cancel_scope in CancelScope docs

### LINKED ISSUE #1831: Mention Nursery.cancel_scope in CancelScope API doc
A year after first learning about Trio I finally did an experiment with it. It worked great! One thing that I haven't realized is the existence of `nursery.cancel_scope`. For some time I was using explicit `CancelScope` around my `open_nursery` block.

I think it would have helped me if `CancelScope` doc would mention `Nursery.cancel_scope`. I did learn about it after I rewatched @njsmith PyCon talk.

https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope

shamrin:
fixes #1831

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -524,6 +524,10 @@ objects.
 
    .. autoattribute:: cancel_called
 
+Often there is no need to create :class:`CancelScope` object. Trio
+already includes :attr:`~trio.Nursery.cancel_scope` attribute in a
+task-related :class:`Nursery` object. We will cover nurseries later in
+the manual.
 
 Trio also provides several convenience functions for the common
 situation of just wanting to impose a timeout on some code:


andersea:
Have you considered putting it in the cheat sheet items further below? Looks like it would be a good fit there.

shamrin:
Hmm, interesting idea! However, the cheatsheet looks like the continuation of "timeouts" section. And it wouldn't help me-in-the-past. I had a vague idea of:

1. I need to cancel these tasks
2. Let's google for "python trio cancel"
3. Ok, here's cancellation section (Cmd-F for "cancel").
4. Ok, I need to create the `CancelScope` explicitly.

My goal is to help people avoid doing 4 and point them to `Nursery.cancel_scope` right away. The cheatsheet is further below after it switches to timeout topic. I didn't need timeouts.

Your suggestion is logical, but it likely won't help with quick search-around-for-solution.

pquentin (approved):
That sounds like a nice little improvement to me! While there are other improvements, perfect is the enemy of good and we can always add them in other pull requests.

Issue #1860, opened 2021-01-10
Clarify that the deadline in `trio.sleep_until` is according to `trio.current_time()`


cubranic:
From the [docs](https://trio.readthedocs.io/en/stable/reference-core.html#trio.sleep_until):
> Pause execution of the current task until the given time.
> 
> The difference between `sleep()` and `sleep_until()` is that the former takes a relative time and the latter takes an absolute time.

I take it from the intro to the "Time and Clocks" section that this deadline is not according to any computer clock, but Trio's own internal one? If so, it might be useful to say that more explicitly in the function docstring. (I'm happy to contribute a PR once you confirm this is the case.)

pquentin:
> I take it from the intro to the "Time and Clocks" section that this deadline is not according to any computer clock, but Trio's own internal one?

Yes, that's right. Maybe writing "... absolute time according to Trio's internal clock" is enough?

PR #1863, opened 2021-01-12
Clarify that deadline in `sleep_until()` uses internal clock units

cubranic:
Fixes: #1860


pquentin:
You ran into an issue with deadsnakes, it's already fixed. Closing/reopening to get the fix.

pquentin (approved):
Thank you!

PR #1864, opened 2021-01-12
make sure coroutine locals are promptly destroyed

richardsheridan:
This PR fixes a "bug" where `unrolled_run` hangs on to `Cancelled` tracebacks too long. They essentially persist until the _next_ exception is thrown out of a task, which may be an indefinite period of time. This was not cleaned up with the earlier cyclic garbage bugfix #1770 because it is not cyclic garbage; it's actually a strong reference that lives unnecessarily long.

(Thinking out loud here) This is just the minimum possible fix. There are actually a number of locals akin to `final_outcome` in `unrolled_run` that may live up to `_MAX_TIMEOUT` seconds and contain references to tasks and potentially to hefty objects, but I haven't actually observed any issues along those lines yet. Scattering `del` statements everywhere is probably untenable, but factoring chunks of `unrolled_run` into function calls would "automate"  that kind of cleanup, at the cost of some overhead.

### DIFF
--- a/newsfragments/1864.bugfix.rst
+++ b/newsfragments/1864.bugfix.rst
@@ -1,0 +1,2 @@
+The event loop now holds on to references of coroutine frames for only
+the minimum necessary period of time.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -2219,12 +2219,15 @@ def unrolled_run(runner, async_fn, args, host_uses_signal_set_wakeup_fd=False):
                         # which works for at least asyncio and curio.
                         runner.reschedule(task, exc)
                         task._next_send_fn = task.coro.throw
+                    # prevent long-lived reference
+                    # TODO: develop test for this deletion
                     del msg
 
                 if "after_task_step" in runner.instruments:
                     runner.instruments.call("after_task_step", task)
                 del GLOBAL_RUN_CONTEXT.task
-                # prevent long-lived task reference
+                # prevent long-lived references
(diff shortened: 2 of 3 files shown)


richardsheridan:
Hmm, the test passes on my machine... I'll look into it this weekend.

oremanj:
FWIW, I do think scattering `del` statements is entirely tenable. I'd be less excited about adding function call overhead on every task step.

belm0 (commented):
worth a newsfragment?

richardsheridan:
I think I got em all, but I would appreciate a double-check!

belm0:
> I think I got em all, but I would appreciate a double-check!

would it be hard to expand tests to cover the additional explicit deletes?

richardsheridan:
Well I haven't been able to come up with a good test so far, let me stew on it another week. Also TIL `except ... as exc` blocks delete `exc` at the end of the block, so two of those are technically unnecessary!

belm0:
> Well I haven't been able to come up with a good test so far, let me stew on it another week. Also TIL `except ... as exc` blocks delete `exc` at the end of the block, so two of those are technically unnecessary!

So just keeping track, the deletions of `msg`, `task`, `next_send`, `next_send_fn` don't have test coverage?  Consider adding a TODO in the code.

belm0 (approved):

belm0:
not sure how to trigger rebuild on sourcehut, will try close & re-open of PR

PR #1866, opened 2021-01-13
Add three random fixes noticed when releasing

pquentin:


pquentin:
I remove one line of code that was covered, which is why the coverage is "dropping".

oremanj:
Thanks!

PR #1873, opened 2021-01-22
Merge type stubs

altendky:
Draft for:
- [ ] Continue enabling more mypy options...
- [ ] Fix existing complaints
- [ ] Compare against trio-typing
- [x] https://github.com/python/typeshed/pull/5013
- [ ] Review `nitpick_ignore`s
- [ ] Remove type descriptions in docstrings
  - Need to check what the documentation setup does presently
- [ ] Organize type aliases
- [ ] Consider `stubtest` to help with verifying dynamic stuff
- [ ] Suggestions
  - [ ] Make `CapacityLimiter` generic https://github.com/python-trio/trio/pull/1873#discussion_r584281535
  - [ ] More specific `.__aexit__()` hints https://github.com/python-trio/trio/pull/1873#pullrequestreview-601634148


trio/_sync.py
```diff
@@ -219,7 +236,7 @@ def acquire_nowait(self):
         self.acquire_on_behalf_of_nowait(trio.lowlevel.current_task())
 
     @enable_ki_protection
-    def acquire_on_behalf_of_nowait(self, borrower):
+    def acquire_on_behalf_of_nowait(self, borrower: Union[object, Task]) -> None:
```
graingert, line 239: `Union[object, Task]` simplifies to `object`

graingert: @altendky you probably want CapacityLimiter to be generic so you can do:

```suggestion
    def acquire_on_behalf_of_nowait(self, borrower: T) -> None:
```

so you can do:

```
class MyResource:
    def __init__(self):
        self._limit: trio.CapacityLimiter[MyResource] = trio.CapacityLimiter(40)

    def acquire(self):
        self._limit.acquire_on_behalf_of_nowait(self)
    def release(self):
        self._limit.release_on_behalf_of(self)
```

graingert: the methods that use a task would have to refine the self type:

```python

    @enable_ki_protection
    def acquire_nowait(self: _CapacityLimiter[abc.Task[object]]) -> None:
        """Borrow a token from the sack, without blocking.
        Raises:
          WouldBlock: if no tokens are available.
          RuntimeError: if the current task already holds one of this sack's
              tokens.
        """
        self.acquire_on_behalf_of_nowait(trio.lowlevel.current_task())
```

altendky: Hints like `Union[object, Task]` were pretty intentional, but open to change.  In this case it mirrored the documentation describing a `Task` as a common particular type of object used here.  In other cases, code actually treated the specific type differently.  Given the suggestion of a more complicated change that I'm not going to get to right now, I'll leave it as is for now.

With the scale of this PR I expect tasks like this to get lost.  I'll make an explicit list of them in the OP with links.  Maybe I'll get lucky and find it is pointlessly redundant.

If there's any interest in bulk contributing to the PR...  I'd welcome it.  `:]`  There's lots more hinting to add, even as just a 'first pass' to satisfy not-so-loose Mypy settings.

trio/_abc.py
```diff
@@ -254,10 +284,10 @@ async def aclose(self):
 
         """
 
-    async def __aenter__(self):
+    async def __aenter__(self: _T) -> _T:
         return self
 
-    async def __aexit__(self, *args):
+    async def __aexit__(self, *args: object) -> None:
```
nimaje, line 290: shouldn't that be
```python
async def __aexit__(self, exc_type: Optional[Type[BaseException]],
        exc_value: Optional[BaseException],
        exc_trackback: Optional[TracebackType]) -> Optional[bool]:
```
As that is the type the contextmanager protocol uses
(Maybe with added `= None`s, not sure about that)

altendky: I don't think this helps anything, but it does seem more proper.  Though I think `-> None` is an accurate and good expression of the behavior.  Thanks.

Added to the list in the OP.

nimaje: using `-> None` would disallow subclasses from returning `True` if they have some reason for that
(returning `None` and `False` have the same effect)

altendky: Good point.  Thanks for following up.

Zac-HD:
Closing this as stale; it looks like we're taking a more incremental approach - so far in #2477, #2671, #2682, and generally with https://github.com/python-trio/trio/pulls?q=is%3Apr+label%3Atyping.  I really appreciate all the work that went into this though!

PR #1881, opened 2021-01-30
WindowsIOManager.wait_overlapped returns CompletionKeyEventInfo

lojack5:
Ok, been a long time since I looked at this (#937), but had another go at it recently.  Here's a working example of using IOCP with 
[ReadDirectoryChanges](/lojack5/ReadDirectoryChanges).  Let me know if this needs a test or more documentation.

### DIFF
--- a/trio/_core/_io_windows.py
+++ b/trio/_core/_io_windows.py
@@ -6,6 +6,7 @@ import sys
 from typing import TYPE_CHECKING
 
 import attr
+from outcome import Value
 
 from .. import _core
 from ._run import _public
@@ -407,7 +408,6 @@ class WindowsIOManager:
 
         # {lpOverlapped: task}
         self._overlapped_waiters = {}
-        self._overlapped_entries = {}
         self._posted_too_late_to_cancel = set()
 
         self._completion_key_queues = {}
@@ -536,8 +536,7 @@ class WindowsIOManager:
                 info = CompletionKeyEventInfo(
                     lpOverlapped=overlapped, dwNumberOfBytesTransferred=transferred
                 )
-                self._overlapped_entries[overlapped] = info
-                _core.reschedule(waiter)
+                _core.reschedule(waiter, Value(info))
             elif entry.lpCompletionKey == CKeys.LATE_CANCEL:
                 # Post made by a regular I/O event's abort_fn
                 # after it failed to cancel the I/O. If we still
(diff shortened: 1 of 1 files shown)


trio/_core/_io_windows.py
```diff
@@ -530,6 +531,12 @@ def process_events(self, received):
             elif entry.lpCompletionKey == CKeys.WAIT_OVERLAPPED:
                 # Regular I/O event, dispatch on lpOverlapped
                 waiter = self._overlapped_waiters.pop(entry.lpOverlapped)
+                overlapped = entry.lpOverlapped
+                transferred = entry.dwNumberOfBytesTransferred
+                info = CompletionKeyEventInfo(
+                    lpOverlapped=overlapped, dwNumberOfBytesTransferred=transferred
+                )
+                self._overlapped_entries[overlapped] = info
```
oremanj, line 539: You can pass the structure into the task as you reschedule it: `_core.reschedule(waiter, Value(info))` (`from outcome import Value`). Then you don't need the sideband `_overlapped_entries` dict.

trio/_core/_io_windows.py
```diff
@@ -763,6 +770,7 @@ def abort(raise_cancel_):
             return _core.Abort.FAILED
 
         await _core.wait_task_rescheduled(abort)
+        info = self._overlapped_entries.pop(lpOverlapped)
```
oremanj, line 773: `wait_task_rescheduled` returns the (unwrapped) thing you pass to `reschedule`, so this would be `info = await _core.wait_task_rescheduled(abort)`

oremanj (commented):
Overall looks good; one suggestion for tightening it up a bit.

This should get documented eventually, but `wait_overlapped` is already undocumented, so I guess this doesn't make it any worse!

trio/_core/_io_windows.py
lojack5: Took awhile to dig into `wait_task_rescheduled` to grok what was going on there, but that's way easier than what I had originally done, thanks!

lojack5:
Ok, redid this with your suggestions, much cleaner that way.  Quick quesetion: does trio normally do a squashed merge? If not I can squash those two commits since it's just noise the way it is.

oremanj:
We do not normally squash merges nor do we worry about rewriting history to be cleaner. :-)

lojack5:
Ok, good to know, thanks!

Issue #1868, opened 2021-01-20
0.18.0: testy suite is failing


kloczek:
```
+ cd trio-0.18.0
+ /usr/bin/python3 -m pytest -v
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.8.3, pytest-6.2.1, py-1.10.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/tkloczko/rpmbuild/BUILD/trio-0.18.0/.hypothesis/examples')
rootdir: /home/tkloczko/rpmbuild/BUILD/trio-0.18.0, configfile: setup.cfg
plugins: flaky-3.6.1, forked-1.3.0, shutil-1.7.0, virtualenv-1.7.0, cov-2.10.1, xdist-2.2.0, asyncio-0.14.0, hypothesis-6.0.2, expect-1.1.0
collected 608 items

[..]

================================================================================= FAILURES =================================================================================
___________________________________________________________________ test_fallback_when_no_hook_claims_it ___________________________________________________________________

capsys = <_pytest.capture.CaptureFixture object at 0x7f7b86f53340>

    @pytest.mark.skipif(buggy_pypy_asyncgens, reason="pypy 7.2.0 is buggy")
    async def test_fallback_when_no_hook_claims_it(capsys):
        async def well_behaved():
            yield 42

        async def yields_after_yield():
            with pytest.raises(GeneratorExit):
                yield 42
            yield 100

        async def awaits_after_yield():
            with pytest.raises(GeneratorExit):
                yield 42
            await _core.cancel_shielded_checkpoint()

        await step_outside_async_context(well_behaved())
        gc_collect_harder()
        assert capsys.readouterr().err == ""

        await step_outside_async_context(yields_after_yield())
        gc_collect_harder()
>       assert "ignored GeneratorExit" in capsys.readouterr().err
E       AssertionError: assert 'ignored GeneratorExit' in ''
E        +  where '' = CaptureResult(out='', err='').err
E        +    where CaptureResult(out='', err='') = <bound method CaptureFixture.readouterr of <_pytest.capture.CaptureFixture object at 0x7f7b86f53340>>()
E        +      where <bound method CaptureFixture.readouterr of <_pytest.capture.CaptureFixture object at 0x7f7b86f53340>> = <_pytest.capture.CaptureFixture object at 0x7f7b86f53340>.readouterr

[...]

PR #1893, opened 2021-02-09
Bump to Pytest 6.2.2 with -p no:unraisableexception -p no:threadexception

altendky:
See #1878 for the initial dependabot PR showing the build issues.

https://docs.pytest.org/en/stable/changelog.html (See 6.2.0 features)
https://docs.pytest.org/en/stable/usage.html#unraisable

I think that disabling these new plugins for now is reasonable since it allows us to keep moving forward with pytest versions while retaining the same lack of these checks as if we held the version back.  If there is concern about globally disabling these checks then I think a follow up ticket and PR would be appropriate.

I did make a cursory effort to use `@pytest.mark.filterwarnings()` but failed to get it to successfully avoid the new errors.  I suspect something related to the combination with `-W error` and other pytest features but I did not dig into it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,3 +6,4 @@ markers =
 junit_family = xunit2
 addopts =
     -p no:unraisableexception
+    -p no:threadexception
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,3 +4,5 @@ faulthandler_timeout=60
 markers =
     redistributors_should_skip: tests that should be skipped by downstream redistributors
 junit_family = xunit2
+addopts =
+    -p no:unraisableexception
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -102,7 +102,7 @@ pyparsing==2.4.7
     # via packaging
 pytest-cov==2.11.1
     # via -r test-requirements.in
-pytest==6.1.2
+pytest==6.2.2
     # via
     #   -r test-requirements.in
     #   pytest-cov


sbraz:
Hi @altendky, I have never used dependabot before so I wouldn't know how to help. I have not looked too hard at this issue, I just found the failure with newer pytest versions quite weird.

As for #1868, it did have the one failure that the pytest bump created (but the rest was unrelated, I agree):
```
FAILED trio/_core/tests/test_asyncgen.py::test_fallback_when_no_hook_claims_it - AssertionError: assert 'ignored GeneratorExit' in ''
```

setup.cfg
```diff
@@ -4,3 +4,6 @@ faulthandler_timeout=60
 markers =
     redistributors_should_skip: tests that should be skipped by downstream redistributors
 junit_family = xunit2
+addopts =
```
graingert, line 7: @altendky if you set

```
filterwarnings =
    error
    ignore::pytest.PytestUnraisableExceptionWarning
    ignore::pytest.PytestUnhandledThreadExceptionWarning
```

and remove the `-W error` does that work?

graingert:
@altendky I've had some success running the suite with stepwise and wrapping the places that currently raise in:


```python
@contextlib.contextmanager
def restore_unraisable_hook():
    sys.unraisablehook, prev = sys.__unraisablehook__, sys.unraisablehook
    try:
        yield
    finally:
        sys.unraisablehook = prev
```

```python
    with restore_unraisable_hook(), pytest.warns(RuntimeWarning, match="Trio guest run got abandoned"):
        do_abandoned_guest_run()
        gc_collect_harder()
```

graingert:
eg some of the code is relying on the default unraisablehook printing to stderr:

https://github.com/python-trio/trio/blob/58e9b80c283caf5dd95a9bb1aaf12e49a47e9bcf/trio/_core/tests
/test_asyncgen.py#L270-L281

graingert (approved):

PR #1898, opened 2021-02-11
reenable pytest -p unraisableexception and -p threadexception

graingert:
by restoring the default unraisablehook, and disabling the threading.excepthook in tests that need it

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,3 +50,10 @@ directory = "misc"
 name = "Miscellaneous internal changes"
 showcontent = true
 
+[tool.pytest.ini_options]
+addopts = ["--strict-markers", "--strict-config"]
+xfail_strict = true
+faulthandler_timeout = 60
+markers = ["redistributors_should_skip: tests that should be skipped by downstream redistributors"]
+junit_family = "xunit2"
+filterwarnings = ["error"]
--- a/trio/_core/tests/tutil.py
+++ b/trio/_core/tests/tutil.py
@@ -107,11 +107,11 @@ if sys.version_info >= (3, 8):
 else:
 
     @contextmanager
-    def restore_unraisablehook():
+    def restore_unraisablehook():  # pragma: no cover
         yield
 
     @contextmanager
-    def disable_threading_excepthook():
+    def disable_threading_excepthook():  # pragma: no cover
         yield
 
 
--- a/trio/_core/tests/tutil.py
+++ b/trio/_core/tests/tutil.py
@@ -1,5 +1,6 @@
 # Utilities for testing
 import socket as stdlib_socket
+import threading
 import os
 import sys
 from typing import TYPE_CHECKING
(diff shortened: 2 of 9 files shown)


trio/_core/tests/tutil.py
```diff
@@ -80,6 +81,40 @@ def ignore_coroutine_never_awaited_warnings():
...
+    def restore_unraisablehook():
+        sys.unraisablehook, prev = sys.__unraisablehook__, sys.unraisablehook
+        try:
+            yield
+        finally:
+            sys.unraisablehook = prev
+
+    @contextmanager
+    def disable_threading_excepthook():
+        threading.excepthook, prev = _noop, threading.excepthook
```
graingert, line 100: sadly `threading.__excepthook__` is only in 3.10

trio/_core/tests/tutil.py
```diff
@@ -80,6 +81,40 @@ def ignore_coroutine_never_awaited_warnings():
...
+        try:
+            yield
+        finally:
+            threading.excephtook = prev
+
+
+else:
+
+    @contextmanager
+    def restore_unraisablehook():
```
graingert, line 110: I could do this if it were prefered:

```python
def _noop(*args, **kwargs):
    pass

@contextmanager
def nullcontext():
    yield

assert nullcontext()  # force coverage

if sys.version_info >= (3, 8):

    @contextmanager
    def restore_unraisablehook():
        sys.unraisablehook, prev = sys.__unraisablehook__, sys.unraisablehook
        try:
            yield
        finally:
            sys.unraisablehook = prev

    @contextmanager
    def disable_threading_excepthook():
        threading.excepthook, prev = _noop, threading.excepthook
        try:
            yield
        finally:
            threading.excephtook = prev


else:
    restore_unraisablehook = disable_threading_excepthook = nullcontext
```

altendky: That does seem maybe generally better, though I don't super like the `assert nullcontext()`.  I wouldn't hold up here for Trio to change to my quickly-thought-up `coverage unneeded`/`coverage wanted` strategy so maybe keep moving with `no cover` on the `nullcontext()` or a 'real' test that uses `nullcontext()` with a `with`.


```python
def test_nullcontext():
    with nullcontext() as null:
        assert null is None
```

Or somesuch...  easy to add right adjacent to the definition, not complicated, not too noisy I think, and it actually checks 'interesting' properties of the code (as opposed to the truthiness of the context manager as the `assert nullcontext()` did).

graingert: @altendky I'm happy to go with any of these options

altendky: Your proposal here plus the actual test seems good?

pquentin:
See https://gitter.im/python-trio/general?at=602b8e1293539e23437b2daa for the Fedora failure

altendky (changes requested):

trio/_core/tests/tutil.py
```diff
@@ -80,6 +81,40 @@ def ignore_coroutine_never_awaited_warnings():
...
 
 
+def _noop(*args, **kwargs):
+    pass
+
+
+if sys.version_info >= (3, 8):
+
+    @contextmanager
+    def restore_unraisablehook():
```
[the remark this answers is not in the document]
graingert: They do different things because I can't restore the threading excepthook on py37-39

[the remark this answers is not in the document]
altendky: Ok, I guess that's not how I read 'restore' in this context.  I read it as restoring at the end of the context manager to what it was before entering.  I see your point though.  I guess it's internal so it's not a huge deal anyways.

pyproject.toml
```diff
@@ -50,3 +50,10 @@ directory = "misc"
 name = "Miscellaneous internal changes"
 showcontent = true
 
+[tool.pytest.ini_options]
+addopts = ["--strict-markers", "--strict-config"]
```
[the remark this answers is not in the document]
graingert: yeah tangentially - I wanted to add the `filterwarnings = ["error"]` filter in the config directly, which meant I figured increasing the strictness elsewhere was relevant enough for a drive-by fix

pyproject.toml
```diff
@@ -50,3 +50,10 @@ directory = "misc"
...
+xfail_strict = true
+faulthandler_timeout = 60
+markers = ["redistributors_should_skip: tests that should be skipped by downstream redistributors"]
+junit_family = "xunit2"
+filterwarnings = ["error"]
```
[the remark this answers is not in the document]
graingert: it's redundant but does no harm there

trio/_core/tests/tutil.py
altendky: Probably shouldn't hold this up.  Maybe I'll add something in https://github.com/python-trio/trio/pull/1921.

altendky (approved):

PR #1920, opened 2021-03-06
Test against 3.10 alpha

altendky:


altendky:
I always forget I can't use GitHub's edit feature due to desires for PRs to be from forks.  Maybe due to the double build so people can get non-PR builds?  Anyways...

PR #1921, opened 2021-03-06
Test against 3.10 alpha

altendky:
Draft for:
- [x] Maybe follow up on https://github.com/python-trio/trio/pull/1898#discussion_r579725430
- [x] Something with https://github.com/python-trio/trio/pull/1914 (merged here)
- [x] There's some risk with adding the `compact=` keyword to `TracebackException.__init__()` in Python <3.10.  Is this acceptable?
- [x] Whatever else is broken
- [x] Document the `python-version:` insanity in the GHA workflow, just in case it doesn't get rejected...
- [x] `typed-ast` not building on 3.10-dev ~~Check out the 3.10-dev compilation failure~~
  - Worked around using their `master` branch but likely don't want to leave that laying around post-merge.
  - https://github.com/python/typed_ast/issues/156
  - Fixed by either the next mypy release or the next typed_ast release
- [ ] Update required checks (since we don't have an `All` job `:]`)


altendky:
I'm not really sure why `typed-ast` only fails on 3.10-dev and not 3.10 but https://github.com/python/typed_ast/issues/156 seems related.  Supposedly fixed on master, so, let's see.

trio/_core/tests/tutil.py
```diff
@@ -97,7 +97,11 @@ def restore_unraisablehook():
 
     @contextmanager
     def disable_threading_excepthook():
```
altendky, line 99: @graingert, I think in 3.10+ this could be called `restore_threading_excepthook()`, yes?  I'm lazy about refactoring it now but maybe it'd be correct?  Correct in the future looking sense...

graingert: Yep

trio/_core/_multierror.py
```diff
@@ -381,10 +381,13 @@ def traceback_exception_init(
     limit=None,
     lookup_lines=True,
     capture_locals=False,
+    compact=False,
```
altendky, line 384: I'm a bit concerned about this introducing some confusion for other projects that are a bit to sensitive about stuff they shouldn't be or who test with trio around and are then confused when something seemingly stdlib is suddenly different without trio.

oremanj: I wouldn't worry about it. It won't affect static analysis tools because they don't understand monkeypatching. The only case I can think of that it might affect is people using `help()` interactively in the REPL, and I don't think that's substantial enough to be worth worrying about.

trio/_core/tests/tutil.py
altendky: Meh, no need to expand this PR.  If I do rename I'll make another PR for it.

ci.sh
```diff
@@ -69,6 +69,8 @@ python -c "import sys, struct, ssl; print('#' * 70); print('python:', sys.versio
 python -m pip install -U pip setuptools wheel
 python -m pip --version
 
+python -m pip install 'typed-ast @ git+https://github.com/python/typed_ast; python_version >= "3.10"'
+
```
graingert, line 73:
```suggestion
```

trio/_core/tests/tutil.py
```diff
@@ -97,7 +98,11 @@ def restore_unraisablehook():
...
-        threading.excepthook, prev = _noop, threading.excepthook
+        if sys.version_info >= (3, 10):
+            threading.excepthook, prev = threading.__excepthook__, threading.excepthook
+        else:
+            threading.excepthook, prev = _noop, threading.excepthook
+
```
njsmith, line 105: This is mysterious :-). What's going on here?

njsmith (approved):
One question but it's not really important, I'm just curious. Feel free to merge when you're happy.

trio/_core/tests/tutil.py
altendky: I think this was for disabling the thread hooks that pytest started using in 6.2 or so to catch and complain about uncaught exceptions in threads.  In certain cases, IIRC, we literally have explanations in the code, or even the error messages, saying that yes, these will fall out un-handled in the thread or whatever but that that's the way it is and we know.  But @graingert did this bit.  The added bit here is to use the 'original' excepthook instead of a noop when it is available (in 3.10).  This additional feature in 3.10 let's this code be more consistent with the `restore_unraisablehook()` above.

Hopefully that's in the direction you were wondering about...

agronholm (approved):
LGTM. I don't fully understand the logic in the CI workflow where the Python version is selected, so hopefully someone else can review that :)

altendky:
@agronholm, yeah, coding `if` statements via `JSON` is...  not the most obvious.  Not one of GitHub Action's strong points.

PR #1924, opened 2021-03-07
Made TrioToken weak referable

agronholm:


altendky:
Worth a quick test?  Maybe the `isinstance()` is 'overkill' but if we aren't directly calling the class maybe it is appropriate.

```python
async def test_trio_token_weak_referenceable():
	token = current_trio_token()
	assert isinstance(token, TrioToken)
    weak_reference = weakref.ref(token)
    assert token is weak_reference()
```

agronholm:
I added that to `test_threads.py`.

newsfragments/1924.bugfix.rst
```diff
@@ -0,0 +1 @@
+The ``TrioToken`` class can now be used as a target of a weak reference.
```
altendky, line 1: Sorry I didn't catch this on the first round...  Note that this doesn't produce a link.

https://trio--1924.org.readthedocs.build/en/1924/history.html#bugfixes

I think this would cut it.

```suggestion
The :class:`TrioToken` class can now be used as a target of a weak reference.
```

altendky (changes requested):
I'll approve and merge after the newsfragment link works.

newsfragments/1924.bugfix.rst
agronholm: That wasn't quite it, but I've fixed the link now.

altendky (approved):

PR #1926, opened 2021-03-09
use attr.s for slotted classes

graingert:
to avoid accidentally forgetting the `__weakref__` slot

### DIFF
--- a/trio/_core/_local.py
+++ b/trio/_core/_local.py
@@ -1,24 +1,25 @@
 # Runvar implementations
+import attr
+
 from . import _run
 
 from .._util import Final
 
 
+@attr.s(eq=False, hash=False, slots=True)
 class _RunVarToken:
     _no_value = object()
 
-    __slots__ = ("_var", "previous_value", "redeemed")
+    _var = attr.ib()
+    previous_value = attr.ib(default=_no_value)
+    redeemed = attr.ib(default=False, init=False)
 
     @classmethod
     def empty(cls, var):
-        return cls(var, value=cls._no_value)
-
-    def __init__(self, var, value):
-        self._var = var
-        self.previous_value = value
-        self.redeemed = False
+        return cls(var)
 
 
+@attr.s(eq=False, hash=False, slots=True)
 class RunVar(metaclass=Final):
     """The run-local variant of a context variable.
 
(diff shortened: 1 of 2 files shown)


newsfragments/1926.misc.rst
```diff
@@ -0,0 +1 @@
+Made classes :class:`~.lowlevel.TrioToken`, :class:`~.lowlevel.RunVar`, and :class:`~._core._local._RunVarToken` attrs classes.
```
graingert, line 1:
```suggestion
Made classes :class:`~.lowlevel.TrioToken`, :class:`~.lowlevel.RunVar` attrs classes.
```

agronholm (approved):
LGTM.

njsmith:
I don't think we need a newsfragment? There's nothing user-visible in this
change, is there?

agronholm:
True that.

newsfragments/1926.misc.rst
```diff
@@ -0,0 +1 @@
+Made classes :class:`~.lowlevel.TrioToken`, :class:`~.lowlevel.RunVar` attrs classes.
```
graingert, line 1:
```suggestion
Made classes :class:`~.lowlevel.TrioToken`, :class:`~.lowlevel.RunVar` attrs classes.
```

graingert (commented):
I added a suggestion, does an empty newsfragment work here?

agronholm:
Just delete the file?

trio/_core/_local.py
```diff
@@ -1,24 +1,25 @@
...
 from .._util import Final
 
 
+@attr.s(eq=False, hash=False, slots=True)
 class _RunVarToken:
     _no_value = object()
 
-    __slots__ = ("_var", "previous_value", "redeemed")
+    _var = attr.ib()
+    previous_value = attr.ib(default=_no_value)
```
altendky, line 14: We are ok with the `.__init__()` parameter name changing from `value` to `previous_value`?  This might be private enough that passing tests suffice.  I thought [`attr.ib()`](https://www.attrs.org/en/20.3.0/api.html#attr.ib) had something for this but I don't see it now.

This isn't a requested change, just making sure it's considered.

agronholm: I considered it and came to the same conclusions you did: it's private API and the tests pass so it's okay.

graingert: Yeah I also considered it a better internal API:

`_RunVarToken(var=value, value=old_value)`

PR #1946, opened 2021-04-05
CI: fetch Proxifier and Spyware Doctor AV LSP's from a source we control

### LINKED ISSUE #1938: Windows LSP host missing
I'm getting a CI failure from missing download.pctools.com, see https://github.com/python-trio/trio/runs/2197656935

I've already restarted the run once, we might want to make sure the artifact we need is still available.

belm0:
Closes #1938


njsmith:
Weirdly, this seems to be consistently failing. I don't know why, since it looks like it should be doing the same thing as our current system...

PR #1947, opened 2021-04-05
Disable Windows "with IFS LSP" test based on Spyware Doctor AV

### LINKED ISSUE #1938: Windows LSP host missing

belm0:
since we lost the installer binary

Closes #1938


belm0:
the corresponding requirement in the repo branch merge rule needs to be disabled

PR #1948, opened 2021-04-05
implement Event directly with wait_task_rescheduled rather than ParkingLot

belm0:
raised in #1944

### DIFF
--- a/trio/_sync.py
+++ b/trio/_sync.py
@@ -1,16 +1,20 @@
 import math
 
 import attr
-import outcome
 
 import trio
 
+from . import _core
 from ._core import enable_ki_protection, ParkingLot
-from ._deprecate import deprecated
 from ._util import Final
 
 
-@attr.s(repr=False, eq=False, hash=False)
+@attr.s(frozen=True)
+class _EventStatistics:
+    tasks_waiting = attr.ib()
+
+
+@attr.s(repr=False, eq=False, hash=False, slots=True)
 class Event(metaclass=Final):
     """A waitable boolean value useful for inter-task synchronization,
     inspired by :class:`threading.Event`.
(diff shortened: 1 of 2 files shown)


trio/_sync.py
```diff
@@ -59,7 +65,15 @@ async def wait(self):
         if self._flag:
             await trio.lowlevel.checkpoint()
         else:
-            await self._lot.park()
+            task = _core.current_task()
+            self._tasks.add(task)
+            task.custom_sleep_data = self
```
richardsheridan, line 70: Why use custom sleep data here? Can't you just access `self._tasks` directly in the closure?

trio/_sync.py
```diff
@@ -47,8 +51,10 @@ def is_set(self):
     @enable_ki_protection
     def set(self):
         """Set the internal flag value to True, and wake any waiting tasks."""
-        self._flag = True
-        self._lot.unpark_all()
+        if not self._flag:
+            self._flag = True
+            for task in self._tasks:
+                _core.reschedule(task)
```
njsmith, line 57:
```suggestion
                _core.reschedule(task)
            self._tasks.clear()
```

trio/_sync.py
belm0: good point

I've copied this code from ParkingLot, so I guess I'll apply the same simplification there

njsmith: `ParkingLot` needs `custom_sleep_data` b/c of the "requeueing" functionality that condition variables need -- see `ParkingLot.repark`

belm0: I see, ParkingLot uses `custom_sleep_data` to support `repark()`

belm0:
ready to go (unless it needs a newsfragment)

suggest "squash and merge" option for merge

PR #1950, opened 2021-04-06
DOC: Demphasize `asks` from awesome libraries (broken)

Carreau:
The project seem to have no activity since October, and does not work
out of the box when installing:

```
 >>> ImportError: cannot import name 'aopen' from 'anyio'
```
It requires anyio ~2.x but is actually failing with it.

As it is the first of the list; it is likely to be tried first and users
wondering if they are doing something wrong (guilty of this myself), or
if trio is broken.

### DIFF
--- a/docs/source/awesome-trio-libraries.rst
+++ b/docs/source/awesome-trio-libraries.rst
@@ -27,20 +27,13 @@ Getting Started
 
 Web and HTML
 ------------
+* `httpx <https://www.python-httpx.org/>`__ - HTTPX is a fully featured HTTP client for Python 3, which provides sync and async APIs, and support for both HTTP/1.1 and HTTP/2.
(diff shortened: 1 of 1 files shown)


njsmith:
Maybe we should just drop `asks` from the list for now? This doesn't have to be a comprehensive list of every project, and we can always add it again later if it becomes more competitive.

Maybe also move httpx up to the first position in the list, since I guess an http client is one of the most common things people are looking for?

Carreau:
> Maybe we should just drop `asks` from the list for now? This doesn't have to be a comprehensive list of every project, and we can always add it again later if it becomes more competitive.

That was my first though but I tried to be conservative.  Done.

> Maybe also move httpx up to the first position in the list, since I guess an http client is one of the most common things people are looking for?

And done, seeing that there are already 3 :+1: on your comment.

Carreau:
Pleasure to do business with you :-)
I haven't had a pull request merged that fast after submission in a long time, and that is a great feeling !

PR #1956, opened 2021-04-12
update typed-ast to 1.4.3 for testing

altendky:


altendky:
Hopefully this helps with https://github.com/python-trio/trio/pull/1921

Issue #1958, opened 2021-04-13
In worker threads, don't hold onto `return` object


njsmith:
This is the core part of the loop that our worker threads currently run:

https://github.com/python-trio/trio/blob/6b0a094c8e5860ff5e26e825ff867229bf9707bb/trio/_core/_thread_cache.py#L60-L73

You can see that we currently explicitly `del fn` and `del deliver`. That's because these are two arbitrary user objects, and if we didn't delete them, then they would stick around until the worker thread is assigned some other job, which might be a long time and could pin an arbitrary amount of objects in memory until then.

However, we don't `del result`, even though that is also an arbitrary user object and has the exact same problem... whoops.

We could add `del result`, but I think there's a better option. Obviously, trying to manually `del` each object is error prone -- we messed it up :-).

The better option is to take the chunk of code that handles the job, and move it into its own function or method, that the main loop calls. That way, any temporary state we create while handling a job is automatically released as soon as the job is finished. Basically let the interpreter worry about putting the `del`s in for us, so we can't mess it up.

PR #1964, opened 2021-04-16
Addresses #1958: Move code that handles job into its own method

cslecrone:
Attempting to address #1958 by moving the code that handles jobs into its own method. This is my first contribution, so let me know if I need to do anything differently. Thanks!

### DIFF
--- a/trio/_core/_thread_cache.py
+++ b/trio/_core/_thread_cache.py
@@ -57,6 +57,8 @@ class WorkerThread:
         thread.start()
 
     def _handle_job(self):
+        # Handle job in a separate method to ensure user-created
+        # objects are cleaned up in a consistent manner. 
         fn, deliver = self._job
         self._job = None
         result = outcome.capture(fn)
--- a/trio/_core/_thread_cache.py
+++ b/trio/_core/_thread_cache.py
@@ -58,7 +58,7 @@ class WorkerThread:
 
     def _handle_job(self):
         # Handle job in a separate method to ensure user-created
-        # objects are cleaned up in a consistent manner. 
+        # objects are cleaned up in a consistent manner.
         fn, deliver = self._job
         self._job = None
         result = outcome.capture(fn)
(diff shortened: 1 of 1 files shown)


njsmith:
That's weird... it looks like Travis just... doesn't know about this PR, for some reason: https://travis-ci.com/github/python-trio/trio/pull_requests

Might be related to their limited OSS credits, but OTOH they are currently building the latest commit to the master branch, so... idk. We should get off of Travis (see #1785). For right now, I'll try closing/reopening and see if that's enough to tickle Travis into doing something.

njsmith:
https://travis-ci.com/github/python-trio/trio/pull_requests

njsmith:
Looks like travis is running now, so fingers crossed it works out :crossed_fingers: 

Anyway, the code change looks good! Could you add a comment explaining why `_handle_job` is separated into its own method?

cslecrone:
Of course! Added one just now.

richardsheridan:
In https://github.com/python-trio/trio/issues/1864#issuecomment-759113248 @oremanj expressed concerns about injecting extra function frames into frequently-called codepaths, preferring `del` everywhere in that case. I wonder if he would raise the some objection here given the opportunity? He ran into the same issue while implementing #1652 (https://github.com/python-trio/trio/pull/1652/files#r486806508) and bugfixed it by using `del result`, @njsmith noted in #1958.

njsmith:
> In #1864 (comment) @oremanj expressed concerns about injecting extra function frames into frequently-called codepaths, preferring del everywhere in that case. I wonder if he would raise the some objection here given the opportunity?

Ah, interesting point, I didn't think of that! Thanks for making the connection.

Looking at it, I think this case is different: there he was talking about the core scheduling loop, which runs on every checkpoint, so it's like... the hottest code in trio by a long way. Here, we're talking about one extra frame per job assigned to a worker thread. Passing a job into a worker thread is pretty cheap, but it's still much *much* more expensive than pushing a call frame. Concretely, the measurements in https://github.com/python-trio/trio/pull/1545 are in tens of µs, while on my same laptop a trivial Python function call is in tens of ns, so ~3 orders of magnitude faster. So in this case I don't think it's a concern.

oremanj:
I’m not worried about the overhead in this case, though I agree it’s good to think about in general!

PR #1976, opened 2021-04-23
fix some typos in the code comments

sophoi:


sophoi:
I looked at the coverage reports, it doesn't seem that the comment changes would make those changes in coverages, can the coverage change be because of something else?

pquentin:
> I looked at the coverage reports, it doesn't seem that the comment changes would make those changes in coverages, can the coverage change be because of something else?

This is happening due to a [recent GitHub Actions change](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/): jobs won't run for first-time contributors, annoyingly. This does not apply to the build.sr.ht builds, so we have some coverage, but it's less than expected. I was able to "approve and run", let's see if the builds and coverage pass.

PR #1978, opened 2021-04-23
updated race example to use variable assignment to track winner instead of memory channel

kbiggers:
When I was first reading through the docs for trio I was a bit thrown off by the use of a memory channel in the race example. Since the memory channel itself is not what is meant to be demonstrated here, I think it might be better to remove it in favor of simple variable assignment. Another plus of this is it demonstrates the ability to use nurseries themselves for waiting. 

I may be alone in being confused with the memory channel, so please let me know if you prefer it the original way, and also let me know if you have any questions about this.

Thanks!

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -794,23 +794,23 @@ finishes first::
        if not async_fns:
            raise ValueError("must pass at least one argument")
 
-       send_channel, receive_channel = trio.open_memory_channel(0)
+       winner = None
 
-       async def jockey(async_fn):
-           await send_channel.send(await async_fn())
+       async def jockey(async_fn, cancel_scope):
+           nonlocal winner
+           await winner = async_fn()
+           cancel_scope.cancel()
 
        async with trio.open_nursery() as nursery:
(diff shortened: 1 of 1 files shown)


twosigmajab (changes requested):
Thanks for the PR, @kbiggers! (Caught a minor issue in the code and left a suggested fix, but) the main change here makes sense to me: Folks new to Trio seem to reach for memory channels more often than more experienced users do, and it's interesting to learn that this particular section of the documentation was partly responsible for this for you and others, especially since it's currently the top result for e.g. https://trio.readthedocs.io/en/stable/search.html?q=race

Looks like these docs were last touched 3 years ago, so I wonder if we have a better understanding of Trionic style now, or if @njsmith thinks these docs should actually stay as-written.

kbiggers:
ah good catch about the await, I'll fix that!

njsmith:
Failed test is spurious, caused by https://github.com/python-trio/trio/issues/1785#issuecomment-827132869

Anyway, looks good, so thanks! Hopefully this will make the next reader's life just a little bit easier.

PR #1983, opened 2021-04-26
Reduce coverage flapping

njsmith:
Add `# pragma: no cover` to a few places where our tests have
fallbacks for OS flakiness, and that are causing coverage to randomly
go up and down depending on CI host load.


pquentin (approved):
Nice!

PR #2030, opened 2021-06-04
Make TracebackException work with Python 3.10.

nascheme:
Pass '_seen' argument as None if empty (on first call). Pass 'compact'
kwarg as needed.

### DIFF
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -381,10 +381,9 @@ def traceback_exception_init(
     limit=None,
     lookup_lines=True,
     capture_locals=False,
+    compact=False,
     _seen=None,
 ):
-    if _seen is None:
-        _seen = set()
 
     # Capture the original exception and its cause and context as TracebackExceptions
     traceback_exception_original_init(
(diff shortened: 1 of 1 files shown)


nascheme:
This might fix #1899.

altendky:
Is this covered in https://github.com/python-trio/trio/pull/1921?

nascheme:
> Is this covered in #1921?

I don't think so.  It fixes the issue with the keyword argument.  However this is a more subtle problem with Python 3.10.  I believe the key commit is 6dfd1734f5b230bb8fbd2a9df806c1333b6652a8 in the cpython repo:
bpo-42848: remove recursion from TracebackException (GH-24158)

The meaning of `_seen = None` is now different than from `_seen = set()`.  If you pass `_seen=set()`, the `__init__` method will not correctly initialize the TracebackException object.

altendky:
I believe we have just implemented different means of passing `None` or `set()`, but that they are the same result?

https://github.com/altendky/trio/blob/c079d3cfdf4b91c76cd9aa376d1ddb76db4d40d7/trio/_core/_multierror.py#L405-L423

I tried a GitHub compare but it isn't showing what I expect. 
 https://github.com/altendky/trio/compare/c079d3cfdf4b91c76cd9aa376d1ddb76db4d40d7...5b3378629c492e5cc458c760d5d3b4df57f75218  So, here is a screenshot of my local diff.  This PR on the left and https://github.com/python-trio/trio/pull/1921 on the right.

image

Aside from the `_seen` bit, it seems like the `kwargs` in this PR are unused.

graingert:
fixed in https://github.com/python-trio/trio/pull/1921 ?

graingert:
ah no this is different

Zac-HD:
I think this has been fixed by the recent change to `exceptiongroup`, @agronholm can reopen if I'm wrong.

PR #2035, opened 2021-06-11
collect telemetry on bpo-32219

graingert:


trio/tests/test_ssl.py
```diff
@@ -98,14 +100,20 @@ def ssl_echo_serve_sync(sock, *, expect_fail=False):
...
-                    # it's not our bug, so ignore it.
-                    exceptions += (ssl.SSLWantWriteError,)
                     try:
                         wrapped.unwrap()
                     except exceptions:
                         pass
+                    except ssl.SSLWantWriteError:  # pragma: no cover
+                        # Under unclear conditions, CPython sometimes raises
+                        # SSLWantWriteError here. This is a bug (bpo-32219),
+                        # but it's not our bug, so ignore it.
```
altendky, line 110: Maybe mention why it is only being complained about for 3.8+ since "ignore it" is no longer true.

graingert, line 110:
```suggestion
                        # but it's not our bug, so ignore it on 3.7
                        # Christian Heimes thinks it's fixed in recent
                        # python versions so we fail the test instead
```

altendky:
```suggestion
                        # but it's not our bug.  Christian Heimes thinks
                        # it's fixed in 'recent' CPython versions so we fail
                        # the test for those and ignore it for earlier
                        # versions.
```

I guess this should also be checking `sys.implementation` since we're ignoring a known CPython bug.

graingert: There aren't any other 3.8 implementations yet

altendky: Doesn't sway my opinion, sorry.  `:]`

trio/tests/test_ssl.py
```diff
@@ -98,14 +100,20 @@ def ssl_echo_serve_sync(sock, *, expect_fail=False):
...
+                        if sys.version_info >= (3, 8):
```
altendky, line 111:
```suggestion
                        if not (sys.implementation == "cpython" and sys.version < (3, 8)):
```

graingert: Right we expect pypy not to have this bug either

altendky (approved):
thanks `:]`

PR #2038, opened 2021-06-14
✏️  Fix a couple of typos in tutorial

tiangolo:
✏️  Fix a couple of typos in tutorial.

Also change the word "trace" that is more technical and probably not necessary (or could deserve an explanation) to the word "point", which is simpler to understand for newcomers and in the sentence is equivalent.


pquentin (approved):
Thanks! codecov is a little flaky, which is is why we don't require it.

If you're interested in improving the documentation to better understand Trio, we have a number of open issues about the docs in general and the tutorial in particular.

tiangolo:
Awesome, thanks!

PR #2044, opened 2021-06-22
remove unused global trio._core._parking_lot._counter

graingert:
according to git -S the last occurance of next(_counter) was removed in 9cc347779ff520299f9b6639bc75e5cae2d898f7


altendky (approved):
Tests pass, two unused lines removed and coverage reported accordingly lower by two lines.  coo'

Issue #2010, opened 2021-05-15
DTLS support


njsmith:
DTLS is a variant of TLS that's used for encrypting communication over UDP. It's used for applications like VPNs and VoIP, where you don't want TLS retransmits and scheduler messing with your packets, but still need confidentiality and integrity. Like I mentioned in gitter the other day, a company that would prefer to remain anonymous has hired me to implement DTLS for Trio. It might end up part of Trio itself (like `trio.SSLStream` already is), or a separate project – that's to be determined. But I figured I'd start making notes on it here either way.

# DTLS info

DTLS is defined by a diff against the TLS spec.

- Latest version is DTLS 1.2, defined in [RFC 6347](https://datatracker.ietf.org/doc/html/rfc6347)
- It's based on TLS 1.2, defined in [RFC 5246](https://datatracker.ietf.org/doc/html/rfc5246)
- There's a recent followup RFC about connection tracking that's pretty relevant to the code we'll need to handle on our end, that's finished but hasn't yet been assigned an official number: ([Draft](https://tlswg.org/dtls-conn-id/draft-ietf-tls-dtls-connection-id.html), [Draft status tracker](https://datatracker.ietf.org/doc/draft-ietf-tls-dtls-connection-id/))
- And the RFC for DTLS 1.3 is also finished and waiting for an RFC number, and AFAICT there aren't any implementations yet, so we'll be ignoring that for now. ([Draft](https://tlswg.org/dtls13-spec/draft-ietf-tls-dtls13.html), [Draft status tracker](https://datatracker.ietf.org/doc/draft-ietf-tls-dtls13/), [OpenSSL implementation tracking issue](https://github.com/openssl/openssl/issues/13900))

OpenSSL supports DTLS, but really as an afterthought, basically by wedging it into a TLS-shaped box. There's a ton of complicated glue required to hook it up to something like a Trio socket.

## Multiplexing

A DTLS socket, like all UDP sockets, can handle lots of peers simultaneously, and act as both a client and a server to different peers. OpenSSL assumes that each transport has a single peer. So it's the user's job to figure out which packet belongs to which OpenSSL connection, and route them appropriately.

Solution: handle the actual socket I/O ourselves. When a packet comes in, use the source address to look up the appropriate OpenSSL connection object, and pass it in.

[...]

PR #2047, opened 2021-06-26
DTLS support

njsmith:
Here's an initial draft of #2010. It's not done -- in particular, needs a ton more tests and docs, as well as some strategy to avoid adding a hard dependency on pyopenssl -- but the core logic is all there in some form.

It depends on upstream changes https://github.com/pyca/cryptography/pull/6138 and https://github.com/pyca/pyopenssl/pull/1026, so the tests are all going to fail for now.


trio/_dtls.py
```diff
@@ -0,0 +1,959 @@
...
+            # ...or, well, OpenSSL didn't like it, so I guess we didn't.
+            return
+        old_stream = dtls._streams.get(address)
+        if old_stream is not None:
+            if old_stream._client_hello == packet:
+                # ...but it's just a duplicate of a packet we got before, so never mind.
+                return
+            else:
+                # Ok, this *really is* a new handshake; the old stream should go away.
+                old_stream._replaced()
```
gesslerpd, line 628: Got an error in this code path about bool not being callable, line 709 potentially a method/attribute name collision

njsmith:
This is still blocked waiting on cryptography+pyopenssl to release, but I believe everything on our end is done, so if anyone is excited about reviewing, go ahead! Docs are viewable here: https://trio--2047.org.readthedocs.build/en/2047/reference-io.html#datagram-tls-support

Aside from the DTLS support itself, this also adds a MVP of a mock network layer for testing (#170). It's not exposed publicly yet, because so far it only supports UDP, the mechanism for manipulating packets in flight is awkward, and I just generally didn't want to make finalizing that API a blocker for this PR. But the DTLS tests use it extensively, and using it felt *fabulous*, so I'm excited.

This is a bit of a weird situation as far as reviewing goes, because this is a huge complex piece of work, and I got paid for it but y'all didn't. So like... any review is much appreciated, but please don't feel obliged to spend days on it or anything! It is not the end of the world if some of this code doesn't get intense scrutiny – it has an extremely thorough test suite and we can always fix stuff later.

Thanks again to [redacted] for sponsoring this work.

njsmith:
Update: pyopenssl has pushed their release, so this no longer needs hacky workarounds to test against the dev version. And I think I tracked down the last few test failures that were happening on pypy? Including finding a fun pypy bug: https://foss.heptapod.net/pypy/pypy/-/issues/3656

Let's see what CI says.

gesslerpd:
Seems like test coverage has gone down by a small amount causing the CI check failures.

@njsmith do you think this PR is ready for review/merge?

pquentin:
The coverage does go down but it's not blocking in my opinion.

gesslerpd (approved):

PR #2055, opened 2021-07-05
add starlette and aiometer to awesome-trio

graingert:


florimondmanca:
Wondering if adding to https://github.com/timofurrer/awesome-asyncio would be worth it as well?

pquentin:
@florimondmanca `aiometer` supports Trio, so it makes sense in the list! I'll be happy to point people to `aiometer` from now on, I wasn't aware of it

PR #2063, opened 2021-07-14
test and eliminate additional sources of cyclic garbage

richardsheridan:
from `MultiErrorCatcher.__exit__`, `_multierror.copy_tb`, `MultiError.filter`, `CancelScope.__exit__`, and `NurseryManager.__aexit__` methods.  This was nearly impossible to catch until #1864 landed so it wasn't cleaned up in #1805 (or during the live stream: https://www.youtube.com/watch?v=E_jvJVYXUAk).

@belm0: curious if this shows up in your application benchmarking.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -528,12 +528,14 @@ class CancelScope(metaclass=Final):
             self._cancel_status = None
         return exc
 
-    @enable_ki_protection
     def __exit__(self, etype, exc, tb):
-        # locals()["@TRIO_KI_PROTECTION_ENABLED"]=True
         # NB: NurseryManager calls _close() directly rather than __exit__(),
         # so __exit__() must be just _close() plus this logic for adapting
         # the exception-filtering result to the context manager API.
+
+        # This inlines the enable_ki_protection decorator so we can fix
+        # f_locals *locally* below to avoid reference cycles
+        locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
 
         # Tracebacks show the 'raise' line below out of context, so let's give
         # this variable a name that makes sense out of context.
@@ -555,6 +557,10 @@ class CancelScope(metaclass=Final):
                 # delete references from locals to avoid creating cycles
                 # see test_cancel_scope_exit_doesnt_create_cyclic_garbage
                 del remaining_error_after_cancel_scope, value, _, exc
+                # deep magic to remove refs via f_locals
+                locals()
+                # TODO: check if PEP558 changes the need for this call
+                # https://github.com/python/cpython/pull/3640
 
     def __repr__(self):
         if self._cancel_status is not None:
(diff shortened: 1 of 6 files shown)


trio/_core/tests/test_run.py
```diff
@@ -2222,8 +2225,10 @@ async def do_a_cancel():
         await do_a_cancel()
         await do_a_cancel()
 
-        async with _core.open_nursery() as nursery:
-            nursery.start_soon(do_a_cancel)
+        with pytest.raises(ValueError):
+            async with _core.open_nursery() as nursery:
+                nursery.start_soon(do_a_cancel)
+                nursery.start_soon(raise_ValueError)
```
belm0, line 2231: I think a comment is warranted here to note that we're covering leaks in MultiError by raising this parallel exceptions

belm0:
> @belm0: curious if this shows up in your application benchmarking.

I haven't been keeping an eye on this lately, but I don't think it came up.  Generally we try to avoid internal async API's that raise exceptions because multi-errors are so hard to deal with.

richardsheridan:
Just noticed that `CancelScope.__exit__` was mentioned as another possible source of cycles but it doesn't show up in the test yet. Happy to hear suggestions, or else I'll poke at it this weekend.

trio/_core/tests/test_run.py
richardsheridan: Additionally, I tried to fully cross-reference all the `del`s and tests.

richardsheridan:
I found a gross refcycle in the KI system.  Here is the output of `objgraph.show_backrefs`:

graph

Note that the `args` tuple is referenced both by the frame locals as well as the `f_locals` dict, leading to the spooky code at 

https://github.com/python-trio/trio/blob/f9d0e2a927c75eae028d2be8fcb7ce6d40703cd8/trio/_core/_ki.py#L162-L168

an alternative solution might be to *inline* the `locals()["@TRIO_KI_PROTECTION_ENABLED"]=True` in `CancelScope.__exit__`, as that is a special case; that way we don't have to run extra bytecode every time KI protection is toggled.

trio/_core/_run.py
```diff
@@ -530,6 +530,7 @@ def _close(self, exc):
 
     @enable_ki_protection
     def __exit__(self, etype, exc, tb):
+        # locals()["@TRIO_KI_PROTECTION_ENABLED"]=True
```
belm0, line 533: need to resolve this (remove, uncomment, or document)

belm0 (commented):
> an alternative solution might be to inline the locals()["@TRIO_KI_PROTECTION_ENABLED"]=True in CancelScope.__exit__, as that is a special case; that way we don't have to run extra bytecode every time KI protection is toggled.

this is appealing-- perhaps add another commit implementing it and let njs weigh in

trio/_core/_run.py
richardsheridan: Oops! 😬 I'll tidy that up when I push the inlined ki protection.

belm0:
LGTM.  Add a newsfragment and check that PR description matches latest changes?

belm0 (approved):

PR #2091, opened 2021-08-04
Update awesome-trio-libraries.rst

andersea:
Add new section for larger libraries and frameworks with new entry for Slurry.


pquentin:
Do you think Tractor and Slurry should be in the same section somehow?

pquentin (approved):

andersea:
@pquentin I kind of wanted to put it along with tractor, in multiprocessing, but slurry doesn't really do multiprocessing. Not as a core feature at least. Wasn't really sure where to put it. It is intended as a data processessing library, so it kind of fits under IOT, but then again not really.

andersea:
@pquentin @goodboy we agreed on "Stream Processing" instead. :)

pquentin (approved):

PR #2118, opened 2021-09-10
Update awesome trio libraries

klen:
Added info about [aio-databases](https://github.com/klen/aio-databases) and [peewee-aio](https://github.com/klen/peewee-aio)


pquentin:
Thanks!

PR #2136, opened 2021-10-10
fix stale function name in example

richardsheridan:
I'm not sure if it was renamed or just a typo originally but I just stumbled upon this function referenced in a docstring that doesn't exist but is pretty obviously meant to be `current_effective_deadline`.


pquentin:
Thanks! Would merge but build is broken on master (and here to)

PR #2139, opened 2021-10-18
Added a macos/linux httpserver command

Sengolda:
this pull-request adds a command for Linux/macos to serve a web server in docs/source/contributing.rst

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -388,7 +388,7 @@ docs directory. The whole process might look something like this::
     make html
 
 You can then browse the docs using Python's builtin http server:
-  - Windows - ``python -m http.server 8000 --bind 127.0.0.1 --directory build/html``
+  - Windows - ``py -m http.server 8000 --bind 127.0.0.1 --directory build/html``
   - Mac/Linux - ``python3 -m http.server 8000 --bind 127.0.0.1 --directory build/html``
 
 and then opening ``http://127.0.0.1:8000/`` in your web browser.


tjstum:
Hey there! Thanks for this submission.

Can you explain a little more of the context behind this change? The command, as originally written, works fine on all platforms (Linux, Windows and macOS). Did you run into some difficulty that led to this proposal? What was it?

I would be more inclined to change this to `python3` unconditionally, or something, but I'm not convinced that there should be different commands for the different platforms.
Understanding what the confusion was will help us craft the best version of the docs!

Sengolda:
> Hey there! Thanks for this submission.
> 
> Can you explain a little more of the context behind this change? The command, as originally written, works fine on all platforms (Linux, Windows and macOS). Did you run into some difficulty that led to this proposal? What was it?
> 
> I would be more inclined to change this to `python3` unconditionally, or something, but I'm not convinced that there should be different commands for the different platforms. Understanding what the confusion was will help us craft the best version of the docs!

Yes, on some macos systems, they have python 2 installed which means `python` will run python 2 and not 3.

Also on windows the command is actually `py -m` I just found that out, I have commited that as well.

Sengolda:
> Thanks for that additional context.
> 
> I'm still not convinced the change here is worthwhile. The rest of this document uses the unqualified `pip` and `pytest`, which have the same set of issues that you describe here (and potentially more).
> 
> I suppose the background assumption of this entire doc is that a contributor would be using something like `venv` or `virtualenv` to create and activate an isolated environment in which to install packages. In an activated virtual environment (that was created with some version Python 3), running the unqualified `python` will be the correct version of Python. Even on Windows!
> 
> My personal opinion is that, to keep the documentation succinct, information about how to use tools like `venv` doesn't belong in the contributing doc, except as maybe a suggestion to use it (and a link to its own docs). If you would like to improve this contributing doc, maybe it would be best to try that direction.

Yes, now that I think about it, I think it would be best to put something in the contributing guide about using venvs and things that may happen if you don't, but should these me in the contributing guide or a different file?

PR #2140, opened 2021-10-21
Add PyPy 3.8 (and nightly) to the build matrix

A5rocks:
Additionally, start testing PyPy on Windows, as that has broken before. Adding to that, also drop PyPy 3.7 nightly tests, as PyPy 3.8 nightly covers that case for the most part.

I'll see if I can perhaps PR dependency packages to package wheels for PyPy 3.8, as installing is currently a non-significant chunk of testing time.

Here are the CI results (as GitHub prevents workflows from running for first-time contributors): https://github.com/A5rocks/trio/actions/runs/1365911667

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,7 +115,7 @@ jobs:
       fail-fast: false
       matrix:
         python: ['3.6', '3.7', '3.8', '3.9', '3.10']
-      include:
+        include:
           - python: '3.8'
             arch: 'x64'
             pypy_nightly_branch: 'py3.8'


A5rocks:
Is the coverage drop because PyPy 3.7 nightly? But 7% is... Well, I'll re-add that then.

A5rocks:
Tracked down the issue with CI to https://github.com/ipython/ipython/issues/13208, perhaps it'll be fixed soon?

pquentin:
Wow, thanks! The next step is to find out the last good build and the first bad one in Trio, both for 3.9 and 3.10. This will give the iPython developers a list of CPython commits that could have caused the regression.

PR #2144, opened 2021-10-26
Clarify PyPy CI setup

pquentin:
@A5rocks I didn't want to do another round just to fix that. Can you please review this, and merge it if you've accepted our invitation?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
           #  lsp: 'http://download.pctools.com/mirror/updates/9.0.0.2308-SDavfree-lite_en.exe'
           #  lsp_extract_file: ''
           #  extra_name: ', with non-IFS LSP'
-          - python: '3.8'
+          - python: '3.8'  # <- not actually used
             arch: 'x64'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       matrix:
         python: ['3.6', '3.7', '3.8', '3.9', '3.10']
         include:
-          - python: '3.8'
+          - python: '3.8'  # <- not actually used
```
webknjaz, line 119: @pquentin it's not just that the `python` value is not used — this inclusion does not generate a macOS job at all. This is because it references non-existing factors (which is solved in other matrixes with an empty string).

pquentin: Oh, good catch. We should fix that.

webknjaz: Yeah, I noticed it while preparing https://github.com/python-trio/trio/pull/2443

webknjaz: I also noticed that those non-existing factors aren't even referenced in steps.

PR #2172, opened 2021-11-05
Ignore unexpected shutdown of TLS connections in OpenSSL 3

pquentin:
Let's see if that fixes the Fedora 36 failures.


pquentin:
No, that did not help.

PR #2182, opened 2021-11-15
Fix test_close_at_bad_time_for_send_all on ppc64el

basak:
On ppc64el, PIPE_BUF is 8192 but make_clogged_pipe() ends up writing a
total of 1048576 bytes before the pipe is full, and then a subsequent
receive_some(10000) isn't sufficient for orig_wait_writable() to return
for our subsequent aclose() call. It's necessary to empty the pipe
further before this happens. So we must loop until the pipe is empty to
make sure that the sender wakes up even in this case. Otherwise
patched_wait_writable() never gets to the aclose(), so
expect_closedresourceerror() never returns, the nursery never finishes
all tasks and this test hangs.


njsmith:
tickle CI

njsmith:
CI is being cranky. Can you rebase on latest head?

basak:
Done, thanks

Issue #851, opened 2019-01-11
Intermittent test failure in test_subprocess.test_signals


njsmith:
It looks like #849 failed to automerge because of an intermittent failure in `test_subprocess.test_signals`: https://ci.cryptography.io/blue/organizations/jenkins/python-trio%2Ftrio/detail/PR-849/1/pipeline

The traceback is:

```
_________________________________ test_signals _________________________________

    async def test_signals():
        async def test_one_signal(send_it, signum):
            with move_on_after(1.0) as scope:
                async with subprocess.Process(SLEEP(3600)) as proc:
                    send_it(proc)
            assert not scope.cancelled_caught
            if posix:
                assert proc.returncode == -signum
            else:
                assert proc.returncode != 0
    
        await test_one_signal(subprocess.Process.kill, SIGKILL)
>       await test_one_signal(subprocess.Process.terminate, SIGTERM)
../.venv/lib/python3.5/site-packages/trio/tests/test_subprocess.py:240: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
send_it = <function Process.terminate at 0x102bc6d90>
signum = <Signals.SIGTERM: 15>

    async def test_one_signal(send_it, signum):
        with move_on_after(1.0) as scope:
            async with subprocess.Process(SLEEP(3600)) as proc:
                send_it(proc)
        assert not scope.cancelled_caught
        if posix:
>           assert proc.returncode == -signum
E           assert -9 == -<Signals.SIGTERM: 15>
E            +  where -9 = <trio._subprocess.Process object at 0x105d6c908>.returncode

../.venv/lib/python3.5/site-packages/trio/tests/test_subprocess.py:235: AssertionError
```

CC: @oremanj

njsmith:
(This failure is in jenkins which we're phasing out, and indeed I have no idea why it's even running here. But I assume that if it failed here it can probably fail on other systems too.)

oremanj:
I'm thoroughly confused at this one -- we're calling `Popen.terminate()` which should always send SIGTERM, we're waiting for the process to exit after sending the signal, the cancel scope wasn't cancelled so we wouldn't have sent kill()...

I can nerf the test so that it permits exited-on-SIGKILL in response to sending SIGTERM, but I'm a little reluctant to do that unless we see this failure more than once - thoughts?

[...]

PR #2190, opened 2021-12-06
Improve stability of tests_subprocess

tjstum:
On macOS, there appears to be some sort of race where sending SIGTERM
to a process "too early" in its lifetime causes the exit status to
appear as if SIGKILL was sent.
I can reproduce this with the stdlib subprocess. Additionally, this
problem doesn't occur if there is a tiny sleep between spawning a
child subprocess and terminating it.

My proposal is to use `/bin/sleep` on posix (which reliably does not
have this issue on macOS) and continue using the Python form elsewhere.

Related to #851

### DIFF
--- a/trio/tests/test_subprocess.py
+++ b/trio/tests/test_subprocess.py
@@ -15,7 +15,6 @@ from .. import (
     sleep_forever,
     Process,
     run_process,
-    TrioDeprecationWarning,
     ClosedResourceError,
 )
 from ..lowlevel import open_process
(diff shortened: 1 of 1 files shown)


pquentin (approved):
While this sounds good, we can't merge until we fix/silence the Fedora build which fails due to OpenSSL 3.0.

tjstum:
Should I delete those build plans, then?

pquentin:
Ideally we should use pytest.mark.xfail when detecting OpenSSL 3.0: https://docs.pytest.org/en/latest/how-to/skipping.html#condition-parameter

(Well ideally we should support OpenSSL 3.0, but that hasn't happened yet)

PR #2193, opened 2021-12-09
Fix leaking file descriptors when subprocess creation fails

tjstum:
If subprocess spawning fails (i.e. unsuccessful call to `exec`), the
created trio resources (i.e. the trio ends of the pipes that will be
used to send/receive data to the child) don't get cleaned up until the
next garbage collection run (which will call their `__del__`).

It would be nicer if this was deterministic, and that resources
created during the run of `open_process` were either completely
transferred to the `Process` object (which can take over the
management of them) or cleaned up (on failure).

To implement this, two ExitStacks are used: One will manage the
resources that need cleaned up unconditionally (the child ends of the
pipe); the other will manage the resources that need cleaned up on
failure.

Trio currently supports versions of python that lack
`contextlib.AsyncExitStack`. While there is a backport for 3.6
available on PyPI, the objects that represent the pipes (_FdHolder and
_HandleHolder) don't really need to have an async close. They are
internal implementation details anyway, so this refactors their
implementation to have the Stream/AsyncResource methods on the actual
Stream objects themselves. This allows us to use the regular ExitStack
and avoid an extra dep

### DIFF
--- a/newsfragments/2193.bugfix.rst
+++ b/newsfragments/2193.bugfix.rst
@@ -1,0 +1,3 @@
+Trio now deterministically cleans up file descriptors that were opened before
+subprocess creation fails. Previously, they would remain open until the next run of
+the garbage collector.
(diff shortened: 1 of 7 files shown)


Fuyukai (approved):
Realistically I don't see any issue with just using ``/dev/fds`` because I don't see any actual possibility for the closing behaviour to change on Windows.

PR #2206, opened 2021-12-24
Update CI matrix.

Fuyukai:
3.6 is EOL so there's no reason to test on it.

3.8-dev and 3.9-dev are released versions, and are unnecessary.

I see no reason to test on both 3.7 and 3.8 PyPy nightly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.6', '3.7', '3.8', '3.9', '3.10']
+        python: ['3.7', '3.8', '3.9', '3.10']
         arch: ['x86', 'x64']
         lsp: ['']
         lsp_extract_file: ['']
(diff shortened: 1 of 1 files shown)


oremanj:
Thank you for cleaning this up! Note we now have some uncovered lines for features that were only being exercised on 3.6. I think it would make the most sense to keep at least some tests for 3.6 until we actually remove 3.6 support, but I'm also OK with committing this as-is if you plan to then remove 3.6 support and remove the 3.6-supporting code that is no longer covered.

oremanj:
I've also updated the branch protection rules so that this is now mergeable.

Fuyukai:
>  if you plan to then remove 3.6 support and remove the 3.6-supporting code that is no longer covered.

Sorry for the delay, new years. Yes, this is what I plan to do ASAP.

pquentin:
Python 3.6 support was dropped in #2210.

Testing on dev versions allows catching issues before releases, but I agree that for already released versions it's less likely to be useful.
I'm not sure why PyPy would not be worthy of tests on both 3.7 and 3.8?

Fuyukai:
> I'm not sure why PyPy would not be worthy of tests on both 3.7 and 3.8?

This would drop for 3.7 nightly and 3.8 nightly, not regular 3.7 or 3.8 with the same justification as dropping -dev for older releases.

pquentin:
Right. In the past Nathaniel often raised PyPy bugs and the PyPy developers were (and still are) extremely reactive: fixes often show up in the nightly the next day. That happens less these days, though.

Fuyukai:
pretty sure this got changed like 10000 days ago so this is no longer relevant

Issue #470, opened 2018-03-16
trio.run() should take **kwargs in addition to *args


kennethreitz:
this would greatly enhance usability (especially for function declarations like `async def request(method, url, *, pool=None, preload_content=False):`

njsmith:
Hmm, this has come up a few times, and it's a totally reasonable question – I think we need some sort of FAQ entry for this... maybe in the tutorial? Will think about it.

Anyway, here's the problem: a function like `run` has to take the thing to run, and it also has to take its own arguments to configure how it runs it. So we need some way to distinguish which arguments are for which, but all the possible solutions involve some kind of awkwardness.

BTW, in ordinary usage you wouldn't use `trio.run` to call `request`, that wouldn't make much sense – you normally call `trio.run` once at the very top of your program, so it's usually something like `trio.run(main)`. (I guess you're looking at making synchronous shim wrappers or something?) But there are a bunch of functions in trio that have this same problem: `nursery.start_soon`, `run_sync_in_worker_thread`, etc. etc., so we do need a general solution.

What we do is recommend people do `from functools import partial` and then `trio.run(partial(request, method, url, pool=my_pool))` or whatever. I'm not a *huge* fan, but all the alternatives seem worse. Some advantages of this are: (a) it works uniformly for async and sync callables, (b) it's ordinary standard Python, so once you learn it  you can use it everywhere, with or without trio, (c) it's introspectable, so e.g. if you start a task like this and then later use trio's debugging interfaces to visualize the task tree, then you'll see that trio has figured out that this task is running a function named `request`, not one named `partial` or `<lambda>` or something, (d) it allows extra arguments to be added, which is kind of a niche thing but important for [`nursery.start`](https://trio.readthedocs.io/en/latest/reference-core.html#trio.The%20nursery%20interface.start) (note: different from `nursery.start_soon`), (e) the syntax really isn't that bad, and in particular supports all of Python's normal funcall syntax like `partial(request, *args, extra_arg, something=whatever, **my_kwargs)`. The left edge looks a little funny, but otherwise it's using function call syntax to represent a function call.

Rejected alternatives include:

* YOLO it up: `run` takes its arguments, any unrecognized arguments are passed on to the called function. But if we did this, then we could never add any new kwargs to `run` (or similar functions) without breaking backcompat, which isn't workable in a foundational library like trio that needs to (eventually) be super careful about backcompat.

[...]

PR #2208, opened 2021-12-26
✨ Add alternative APIs that improve typing and tooling support (e.g. autocompletion)

tiangolo:
✨ Add alternative APIs that improve typing and tooling support (e.g. autocompletion in editors)

## Description

I want to propose adding a new API style for three functions (additional to the current API) that would take advantage of new type annotations ([PEP 612 - Parameter Specification Variables](https://www.python.org/dev/peps/pep-0612/)), to get better tooling and editor support.

In particular, this would allow/support **mypy** checks, **inline editor errors**, and ✨ **autocompletion** ✨ for positional and **keyword arguments** for:

* Sending async tasks in a nursery (equivalent to `nursery.start_soon()`)
* Sending blocking tasks to a worker thread (equivalent to `to_thread.run_sync()`)
* Sending async tasks to the main async loop from a worker thread (equivalent to `from_thread.run()`)

This would also allow mypy and editors to provide type checks, autocompletion, etc. for the **return values** received when sending tasks to a worker thread and when receiving them from a worker thread:

* Equivalent to `to_thread.run_sync()`
* Equivalent to `from_thread.run()`

## Previous Discussion

I proposed this in AnyIO here: https://github.com/agronholm/anyio/pull/403

I was pointed out at the discussion here: https://github.com/python-trio/trio/issues/470

That discussion revolves mainly around accepting keyword arguments for functions called, and this is also related to that. But one of the main things I want from this is better tooling and editor support: **autocompletion** and **inline errors**... not only supporting kwargs. So I thought it was worth sharing this additional perspective.

## Usage

The main difference is that instead of having a function that takes the function to call plus:

* positional arguments (and that requires using partials for keyword arguments)
* configs (like task name)

```Python
trio_function(worker_function, arg1, arg2)
```

...this new API only takes the function to call plus Trio-specific configs (like the task name), and then *returns* another function that takes the positional and keyword arguments. And when that function is called, it does the actual work. So it's used like this:

```Python
trio_functionify(worker_function)(arg1, arg2)
```

And now it also supports keyword arguments, not only positional ones, so it's not necessary to use (and understand) partials to use keyword arguments:

```Python
trio_functionify(worker_function)(arg1, arg2, kwarg1="a", kwarg2="b")
```

...but the main benefit, from my point of view, is actually better typing/tooling support. Autocompletion and inline type errors in the editor and mypy.

### Nursery

```Python
import trio


async def program() -> None:
    async def worker(i: int, message: str = "Default worker"):
        print(f"{message} - worker {i}")

        for i in range(3):
            nursery.soonify(worker)(i=i, message="Hello ")


trio.run(program)
```

* Autocompletion for a nursery:

Selection_032

* Inline errors for a nursery:

Selection_033

* Mypy error detection:

```
main.py:10: error: Too few arguments
Found 1 error in 1 file (checked 1 source file)
```

### To Thread (asyncify)

```Python
import time
import trio


def run_slow(i: int, message: str = "Hello ") -> str:
    time.sleep(i)
    return message + str(i)


async def program() -> None:
    for i in range(3):
        result = await trio.to_thread.asyncify(run_slow)(i=i)
        result + 3
        print(result)


trio.run(program)
```

* Autocompletion for sending to worker thread:

Selection_034

* Inline errors for sending to worker thread:

Selection_035

* Autocompletion for result value:

Selection_036

* Inline errors for result value:

Selection_037

* Mypy error detection:

```
main.py:13: error: Unsupported operand types for + ("str" and "int")
Found 1 error in 1 file (checked 1 source file)
```

### From Thread (syncify)

```Python
import trio


async def aio_run_slow(i: int, message: str = "Async Hello ") -> str:
    await trio.sleep(i)
    return message + str(i)


def run_slow_embed(i: int) -> str:
    result = trio.from_thread.syncify(aio_run_slow)(i=i, message="Hello sincify ")
    return result


async def program() -> None:
    for i in range(3):
        result = await trio.to_thread.asyncify(run_slow_embed)(i=i)
        print(result)


trio.run(program)
```

* Autocompletion for sending from worker thread:

Selection_038

* Inline errors for sending from worker thread:

Selection_039

* Autocompletion for result value:

Selection_040

* Inline errors for result value:

Selection_041

* Mypy error detection:

```
main.py:11: error: Unsupported operand types for + ("str" and "int")
Found 1 error in 1 file (checked 1 source file)
```

## Function names Bikeshedding

I wanted to have short function names that were descriptive enough, that's why the "somethingify".

I thought any more descriptive variant like `generate_async_function_that_calls_run_sync` would be too long. :sweat_smile: 

My rationale is that if a user wants to call a blocking function while inside async stuff, the user might want to have a way to "asyncify" that blocking function. And also the "somethingify" doesn't imply it is being executed/called right away, but that something is "converted" in some way. I would hope that would help with the intuition that it just returns another function that is then called, so it's necessary to add the extra chained parenthesis with any possible needed arguments.

But of course, I can change the names if there's another preference.

## Tests and Docs

I want to gather feedback first before writing tests and docs, but I'll add them if this or something like this is acceptable.

## Type Annotations

I'm also adding the bare minimum type annotation stuff to make it all work with mypy.

## Note on AnyIO

I want to have this on AnyIO: https://github.com/agronholm/anyio/pull/403, but AnyIO would want any decision to be taken here first, so I'm making the PR here as well, even if just as a conversation starter.

## Note on FastAPI

I want to have this type of interface for users and myself, I use **autocompletion** and tooling/types a lot. I thought of adding something like this to FastAPI itself, but I think this would really belong here in Trio and AnyIO, not in FastAPI.

I also intend to add it to the FastAPI docs to explain "advanced" use cases with concurrency and blocking code in threads run from async code, etc. I would like to **document** all those things **in FastAPI** referring to and explaining these libraries directly (AnyIO, Trio), instead of writing wrappers in FastAPI or anything else. :nerd_face:

As I see this (in particular taking kwargs) is quite debated, ~I'm considering building~ *I built* a small package *([Asyncer](https://github.com/tiangolo/asyncer))* just with those wrappers, documenting it as highly experimental and subjective (to my point of view). Maybe that could help gauge how useful that API style is for Trio and AnyIO before making any commitments. I don't like the feel of "yet another library", but it might help separate what is more stable (Trio and AnyIO) and what is a Frankenstein temporary experiment (this idea I'm proposing).

**2022-01-08 Edit**: I built [Asyncer](https://github.com/tiangolo/asyncer) to try out this API design. :nerd_face:

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ include LICENSE LICENSE.MIT LICENSE.APACHE2
 include README.rst
 include CODE_OF_CONDUCT.md CONTRIBUTING.md
 include test-requirements.txt
+include trio/py.typed
 recursive-include trio/tests/test_ssl_certs *.pem
 recursive-include docs *
 prune docs/build
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -973,7 +973,7 @@ class Nursery(metaclass=NoPublicConstructor):
     ) -> Callable[T_ParamSpec, None]:
         def wrapper(*args: T_ParamSpec.args, **kwargs: T_ParamSpec.kwargs) -> None:
             partial_f = functools.partial(async_fn, *args, **kwargs)
-            return self.start_soon(partial_f, name=name)
+            self.start_soon(partial_f, name=name)
 
         return wrapper
 
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -18,7 +18,12 @@ import enum
 from contextvars import copy_context
 from math import inf
 from time import perf_counter
-from typing import Callable, TYPE_CHECKING
+from typing import Any, Awaitable, Callable, TYPE_CHECKING, TypeVar
+
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
 
 from sniffio import current_async_library_cvar
 
@@ -851,6 +856,10 @@ def open_nursery():
     return NurseryManager()
 
 
+T_Retval = TypeVar("T_Retval")
+T_ParamSpec = ParamSpec("T_ParamSpec")
+
+
 class Nursery(metaclass=NoPublicConstructor):
     """A context which may be used to spawn (or cancel) child tasks.
 
@@ -956,6 +965,17 @@ class Nursery(metaclass=NoPublicConstructor):
                 # avoid a garbage cycle
                 # (see test_nursery_cancel_doesnt_create_cyclic_garbage)
                 del self._pending_excs
+
+    def soonify(
+        self,
+        async_fn: Callable[T_ParamSpec, Awaitable[Any]],
+        name: str = None,
+    ) -> Callable[T_ParamSpec, None]:
+        def wrapper(*args: T_ParamSpec.args, **kwargs: T_ParamSpec.kwargs) -> None:
+            partial_f = functools.partial(async_fn, *args, **kwargs)
+            return self.start_soon(partial_f, name=name)
+
+        return wrapper
 
     def start_soon(self, async_fn, *args, name=None):
         """Creates a child task, scheduling ``await async_fn(*args)``.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -806,7 +806,7 @@ class NurseryManager:
     """
 
     @enable_ki_protection
-    async def __aenter__(self):
+    async def __aenter__(self) -> "Nursery":
         self._scope = CancelScope()
         self._scope.__enter__()
         self._nursery = Nursery._create(current_task(), self._scope)
@@ -845,7 +845,7 @@ class NurseryManager:
         assert False, """Never called, but should be defined"""
 
 
-def open_nursery():
+def open_nursery() -> NurseryManager:
     """Returns an async context manager which must be used to create a
     new `Nursery`.
 
--- a/trio/_threads.py
+++ b/trio/_threads.py
@@ -1,14 +1,22 @@
 # coding: utf-8
 
+import functools
 import threading
 import queue as stdlib_queue
 from itertools import count
+from typing import Awaitable, TypeVar, Callable, Optional
+import sys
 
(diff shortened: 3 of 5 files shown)


Fuyukai:
I have no real comments about the implementation of this but I am absolutely, vehemently against the concept of making libraries work around Python's anemic type system. 

This should be fixed upstream, by making ``functools.partial`` properly representable, not by adding new APIs to every library that takes a callable with args.

agronholm:
It should also be noted that trio, for reasons beyond my understanding, does not want type annotations in the actual codebase, but instead in [trio-typing](https://pypi.org/project/trio-typing/).

pquentin:
> It should also be noted that trio, for reasons beyond my understanding, does not want type annotations in the actual codebase, but instead in [trio-typing](https://pypi.org/project/trio-typing/).

This isn't true, for what it's worth. It's just that adding types after the facts is hard and the efforts to merge type stubs stalled, see https://github.com/python-trio/trio/pull/1873. I'd like to add that trying to do everything in one pull request is highly likely to fail. If anyone wants to work on adding types, they should do it gradually, like urllib3 did: https://sethmlarson.dev/blog/tests-arent-enough-case-study-after-adding-types-to-urllib3

TeamSpen210:
Once [PEP 646](https://peps.python.org/pep-0646/) (Variadic Generics) is implemented in checkers, the current APIs can be typed directly so this API change isn't as necessary.

Zac-HD:
Closing this since subsequent PEPs mean we'll be able to precisely type-annotate the current signatures 🙂

rafalkrupinski:
I would love to see this merged!

jakkdl:
We have improved the type annotations to the code now, so all of the effects of this PR should be live afaik.

rafalkrupinski:
Can you elaborate? Documentation `reads to_thread.run_sync` doesn't support kwargs.

agronholm:
You can pass either a `functools.partial()` or a lambda to it.

rafalkrupinski:
This interface would be much more convenient.

PR #2210, opened 2022-01-15
Drop Python 3.6 support

agronholm:
This is the (required) first step towards adopting PEP 654 (exception groups).


pquentin (changes requested):
There are many other things to remove:

 * the ipython hack in ci.sh
 * the mention of "Python 3.6" in docs/source/reference-io.rst
 * setup.py that still says "3.6-or-better"
 * trio/_core/_entry_queue: we should say dict is always ordered
 * trio/_core/_run.py: remove _count_context_run_tb_frames and CONTEXT_RUN_TB_FRAMES
 * trio/_highlevel_ssl_helpers.py: remove lines that says py36
 * trio/_util.py: remove "and you're using Python 3.6"

My strategy for finding those is to grep for 3.6, py36 and "3, 6" and do the same for 3.5 and 3.7, excluding notes-to-self and docs/source/history.rst. I then ask myself: "if written today, would we still mention the version number".

agronholm:
I've updated all the code and comments I could find.

pquentin (approved):
Thanks! I added three more commits, one of them found when searching for "3.7" (in requirements.txt) and the other found by looking at the coverage changes. There are two coverage issues left that I don't know how to fix properly:

 * `if hasattr(ssl, "OP_NO_TLSv1_3")` is now always True, so we don't test the other configuration as hoped. I don't think there's a good way to use an old OpenSSL version, so we should probably bite the bullet and remove that branch
 * In trio/_util.py, BaseMeta is always ABCMeta in tests, does that mean we can remove the GenericMeta branch? The comment suggests otherwise

See those issues for yourself in https://app.codecov.io/gh/python-trio/trio/compare/2210/changes/. Those two issues seem minor enough that I've approved this PR, but please review my commits before merging.

agronholm:
I looked through your changes and they seem fine to me.

agronholm:
Is it customary to squash commits on a merge in this projects? That would be my preference here.

agronholm:
Thanks for the review and the assistance!

pquentin:
Thank you for working on this! BTW I forgot, but we need a deprecation newsfragment. It can be as simple as https://github.com/python-trio/trio/pull/1481/files#diff-01d318a61ac955b73555728cb95e87911f78ed8cca35f236214f2403c5df81a3

PR #2212, opened 2022-01-16
Drop dependency on async_generator

agronholm:
This removes all uses of `async_generator`, replacing it with contextlib2 in the cases where `aclosing()` is needed, and then only on Python < 3.10.

This depends on #2210.


pquentin:
While it makes sense to stop depending on `async_generator` at runtime, surely we should continue making sure in tests that we support it correctly for projects that are still using it?

agronholm:
I suppose so. I will restore the tests.

agronholm:
> While it makes sense to stop depending on `async_generator` at runtime, surely we should continue making sure in tests that we support it correctly for projects that are still using it?

We cannot drop async_generator and continue to support it at the same time. That is the inconvenient truth. [Here](https://github.com/python-trio/trio/blob/master/trio/_core/_ki.py#L144) is the line that makes all the difference.

agronholm:
I think we need a change in strategy: Deprecate `async_generator` itself first by emitting a deprecation warning on import. After a sufficient amount of time, we can drop the support in trio.

pquentin:
Agreed. In any case this means we can close this pull request.

PR #2213, opened 2022-01-16
Replace MultiError with BaseExceptionGroup

### LINKED ISSUE #2211: Replace MultiError with (Base)ExceptionGroup
Now that PEP 654 has been accepted, and is implemented in the upcoming Python 3.11, and with the [backport](https://pypi.org/project/exceptiongroup/) out, we should retire `MultiError` in favor of the standard `(Base)ExceptionGroup` exception(s).

### LINKED ISSUE #2427: Except hook incompatibility with exceptiongroup backport
Since version 6.54.0, hypothesis uses exception notes (PEP 678) to show falsifying examples underneath the raised exception. With python < 3.11, this is done using a custom exception hook provided by [exceptiongroup](https://github.com/agronholm/exceptiongroup), a backport of PEP 654. The problem is that trio also uses a custom excepthook to manage multi-errors. In practice, this leads to the falsifying example not being displayed depending on the import order. Even worse, simply having `pytest-trio` installed causes the falsifying examples to disappear. Consider the following test:
```python

from hypothesis import given
from hypothesis import strategies as st

@given(x=st.integers())
def test_notes(x):
    1/x
```
Without `pytest-trio` installed you would get:
```python
    @given(x=st.integers())
    def test_notes(x):
>       1/x
E       ZeroDivisionError: division by zero
E       Falsifying example: test_notes(
E           x=0,
E       )

test_notes.py:7: ZeroDivisionError
```
And with `pytest-trio` installed:
```python
    @given(x=st.integers())
    def test_notes(x):
>       1/x
E       ZeroDivisionError: division by zero

test_notes.py:7: ZeroDivisionError
```

agronholm:
This PR has two dependencies:

- ~#2210 (required)~
- ~#2212 (recommended but not strictly necessary)~ (skipped)

Closes #2211.
Closes #2427.


agronholm:
This needs discussion on whether to drop `MultiError` cold turkey or to have it inherit from `BaseExceptionGroup`. It's not yet certain if inheriting it is possible in the first place.

newsfragments/2211.removal.rst
```diff
@@ -0,0 +1,4 @@
+``trio.MultiError`` has been removed in favor of the built-in ``BaseExceptionGroup``
```
pquentin, line 1: We usually don't outright remove things, but deprecate them, and give users a few months or a few release to use the new thing. Is it possible in that case?

pquentin: Oh, just noticed this was discussed in chat: https://gitter.im/python-trio/general?at=61f0909d7842bd3ca96472f5

agronholm:
Here are my thoughts about deprecating `MultiError`. If we were to deprecate it and have it inherit from `BaseExceptionGroup` instead, what would be the migration path? There are two cases how it might be used by downstream users:

1. Raising them
2. Catching them (or their sub-exceptions)

To maintain backwards compatibility, both use cases would have to be covered. Trio itself would have to raise `MultiError` with warning suppression turned on to counter the deprecation warning we would have have its constructor emit. Then we would have to have the extra class methods (`catch()` and `filter()`) raise deprecation warnings as well. This only leaves one hole: cases where users catch `MultiError` directly. We can't emit a deprecation warning from the `exceptions` property because that would also happen for the users doing the right thing (catching `BaseExceptionGroup` instead).

Either way, a migration guide would probably be in order.

smurfix:
Meh. Frankly I don't think we can achieve 100% backwards compatibility, which is especially bad since the affected code is on code paths that are typically not tested very well.

Thus I'd advise to spend the effort on a good migration guide that covers all the bases, preferably from multiple angles. Code that wants to be able to handle both `BaseExceptionGroup` and `MultiError` will just have to duplicate the affected error handlers for a year or two.

agronholm:
> Meh. Frankly I don't think we can achieve 100% backwards compatibility, which is especially bad since the affected code is on code paths that are typically not tested very well.
> 
> Thus I'd advise to spend the effort on a good migration guide that covers all the bases, preferably from multiple angles. Code that wants to be able to handle both `BaseExceptionGroup` and `MultiError` will just have to duplicate the affected error handlers for a year or two.

This seems to be the consensus (on Gitter as well). I will attempt to write a migration guide somewhere in the docs.

agronholm:
I amended the documentation and the news fragment to point users in the right direction. While this is not strictly a "How to migrate from MultiError to BaseExceptionGroup", trio users tend to be more tech savvy than the average Python user. I hope that is enough.

agronholm:
I managed to make the tests pass on py3.11, but the changes broke the style check job, possibly due to upgraded dependencies. Is it okay to include those fixes in this PR?

richardsheridan:
You could fix it in #2241 and then merge or rebase master, what about that?

agronholm:
Hm, I think you need to elaborate. Say I merge that PR, then what? Another PR to fix the changes required by the new Black version?

richardsheridan:
I was thinking push commits to that dependabot branch until CI comes back green. That should fix the style change in not-beta black regarding the `**` operator while keeping this PR is "clean" of black style changes.

On the other hand I have no problem personally with mixing purposes in PRs! 🤷 Also now that I think about it, I'm not sure you (or I) would have permissions to do that push...

agronholm:
It's in the same repository so we should have push access. I'll try doing that.

agronholm:
Ok, done. Any idea why I can't merge it? All the required checks pass.

oremanj:
I think we should try harder to make this backwards-compatible. The big benefit of ExceptionGroups (`except*` syntax) will not be available to basically any of our users for ~10 more months, and asking them to take a breaking change with neither forward- nor backward-compatibility is a big deal. It becomes necessary for them to update their code and Trio in lockstep. In a lot of environments this is a *major* hassle.

I would propose a scheme like the following sketch:
```
class _MultiError(BaseException):
    @classmethod
    @deprecated(...)
    def catch(...):
        # implementation as before

    @classmethod
    @deprecated(...)
    def filter(...):
        # implementation modified to work on both MultiError and BaseExceptionGroup

    # remaining methods as before

enable_attribute_deprecations(__name__)
__deprecated_attributes__ = {"MultiError": DeprecatedAttribute(_MultiError, ...)}

class _TrioBaseExceptionGroup(BaseExceptionGroup, _MultiError):
    __new__ = BaseExceptionGroup.__new__
    __init__ = BaseExceptionGroup.__init__
    __repr__ = BaseExceptionGroup.__repr__
    def derive(self, excs): return _TrioBaseExceptionGroup(self.message, excs)

class _TrioExceptionGroup(ExceptionGroup, _MultiError):
    __new__ = ExceptionGroup.__new__
    __init__ = ExceptionGroup.__init__
    __repr__ = BaseExceptionGroup.__repr__
    def derive(self, excs): return _TrioExceptionGroup(self.message, excs)
```
And make Trio raise `_TrioBaseExceptionGroup` / `_TrioExceptionGroup`. Note these are internal names; we expect user code to catch them as either `[Base]ExceptionGroup` or `MultiError`.

> Trio itself would have to raise MultiError with warning suppression turned on to counter the deprecation warning we would have have its constructor emit.

In the above sketch, the deprecation is attached to the `MultiError` name, not its constructor. Trio uses an alternate name `_MultiError` which does not produce the warning.

> Then we would have to have the extra class methods (catch() and filter()) raise deprecation warnings as well.

I don't see the problem? These are in fact deprecated APIs, and users can upgrade these to use `exceptiongroup.catch()`. Note these need to be deprecated specifically, and can't ride the MultiError deprecation, because they'll be inherited in `_TrioBaseExceptionGroup` so we need to catch people using `caught_exc.filter()` -- not that I really expect anyone to do that...

> This only leaves one hole: cases where users catch MultiError directly. We can't emit a deprecation warning from the exceptions property because that would also happen for the users doing the right thing (catching BaseExceptionGroup instead).

Luckily BaseExceptionGroup also has an `exceptions` property so there's no need to deprecate the property! We can just deprecate the MultiError name.

----

Does anyone see anything I'm missing in the above? If not, I'm happy to work on fleshing it out and updating this PR to be backwards-compatible if @agronholm isn't up for that.

agronholm:
Was your plan for to not raise any deprecation warnings in the `MultiError` constructor?

oremanj:
It's pretty hard to construct a MultiError without writing the name `trio.MultiError`. If we raise a deprecation warning when the name is accessed, I don't think we need another one on the constructor call.

agronholm:
What do you mean by "accessing" the `MultiError` name? What does that mean?

oremanj:
I mean that this construct
```
enable_attribute_deprecations(__name__)
__deprecated_attributes__ = {"MultiError": DeprecatedAttribute(_MultiError, ...)}
```
causes a deprecation warning to be raised at the point where someone writes `trio.MultiError`, regardless of whether they then do anything with it. It basically works by defining a `__getattr__` on the module.

agronholm:
Ok, it was by no means obvious to me what those functions did. I found `enable_attribute_deprecations()` in `_deprecate.py`, but now that we support only py3.7+ and up, we could just define that `__getattr__` directly in `trio/__init__.py` I suppose?

oremanj:
I figure since we already have the explicit facility we may as well use it, but I guess a module-level `__getattr__` works too.

agronholm:
If the existing facility works for this then I will defer changing that to another PR.

agronholm:
I hope this is enough. I'm quite mentally exhausted now.

trio/_core/_run.py
```diff
@@ -116,6 +119,28 @@ class IdlePrimedTypes(enum.Enum):
...
 
+def collapse_exception_group(excgroup):
+    """Recursively collapse any single-exception groups into that single contained
+    exception.
+
+    """
+    exceptions = list(excgroup.exceptions)
+    modified = False
+    for i, exc in enumerate(exceptions):
+        if isinstance(exc, BaseExceptionGroup):
```
oremanj, line 130: I think this should probably only modify trio MultiErrors? It's plausible that someone could raise an exception group from a different library, and we shouldn't mess with it (unlike ours, it might carry important data in its message or other attributes).

trio/_core/_run.py
```diff
@@ -510,7 +529,17 @@ def _close(self, exc):
             and self._cancel_status.effectively_cancelled
             and not self._cancel_status.parent_cancellation_is_visible_to_us
         ):
-            exc = MultiError.filter(self._exc_filter, exc)
+            if isinstance(exc, Cancelled):
```
oremanj, line 532: The new behavior here has a number of differences from the old, which I'm not sure were intended and which I've pointed out individually. Do you think it might be wise to just use the old `MultiError.filter()` instead of trying to reimplement its semantics imprecisely? You could expose it under an internal non-deprecated name. And then when everyone has ported over to ExceptionGroups we can remove MultiError as well as the collapsing of single-error MultiErrors into their constituent.

trio/_core/_multierror.py
```diff
@@ -174,21 +176,22 @@ class MultiError(BaseException):
 
     """
 
-    def __init__(self, exceptions):
+    def __init__(self, exceptions, *, _collapse=True):
         # Avoid recursion when exceptions[0] returned by __new__() happens
         # to be a MultiError and subsequently __init__() is called.
-        if hasattr(self, "exceptions"):
+        if _collapse and hasattr(self, "_exceptions"):
```
oremanj, line 182: Using the underscore-prefixed `_exceptions` name in this test feels like a backport-ism; does 3.11 have that name? I think you can probably use `hasattr(self, "exceptions")` since presumably the property will raise AttributeError if the underlying attribute doesn't exist.

trio/_core/tests/test_multierror.py
```diff
@@ -150,21 +141,10 @@ def handle_ValueError(exc):
         else:
             return exc
 
-    filtered_excs = MultiError.filter(handle_ValueError, excs)
+    filtered_excs = pytest.deprecated_call(MultiError.filter, handle_ValueError, excs)
     assert isinstance(filtered_excs, NotHashableException)
 
 
-def test_traceback_recursion():
```
oremanj, line 157: I think this is still a good test for wherever you're doing your traceback monkeypatching; I wouldn't want to remove it here unless it's been added in `exceptiongroup` and in CPython. Same comment applies to the big chunk of removed code further below in this file.

trio/_core/tests/test_run.py
```diff
@@ -1861,7 +1866,7 @@ def handle(exc):
                 else:  # pragma: no cover
                     return exc
```
oremanj, line 1867: Is the `catch` handler supposed to reraise the exception if it's not handled? You documented that it's equivalent to an `except*` block, but here you're returning `exc` rather than raising it.

trio/_highlevel_open_tcp_stream.py
```diff
@@ -114,8 +116,12 @@ def close_all():
                 sock.close()
             except BaseException as exc:
                 errs.append(exc)
-        if errs:
-            raise trio.MultiError(errs)
+        if len(errs) == 1:
+            raise errs[0]
+        elif errs:
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", DeprecationWarning)
```
oremanj, line 123: You don't need this warnings filter anymore. Also it looks like you might need a `# pragma: no cover` on this block.

trio/tests/test_util.py
```diff
@@ -93,6 +95,9 @@ def not_main_thread():
 
 # @coroutine is deprecated since python 3.8, which is fine with us.
 @pytest.mark.filterwarnings("ignore:.*@coroutine.*:DeprecationWarning")
+@pytest.mark.skipif(
```
oremanj, line 98: Can you guard just the actual use of `@asyncio.coroutine` with this condition? The test covers many different errors and most of them should still work on 3.11.

newsfragments/2211.deprecated.rst
```diff
@@ -0,0 +1,12 @@
+``MultiError`` has been deprecated in favor of the standard :exc:`BaseExceptionGroup`
+(introduced in :pep:`654`). On Python < 3.11, this exception and its derivative
+:exc:`ExceptionGroup` are provided by the backport_. Trio still raises ``MultiError``,
+but it has been refactored into a subclass of :exc:`BaseExceptionGroup` which users
+should catch instead of ``MultiError``. Uses of the ``filter()`` class method should be
+replaced with :meth:`BaseExceptionGroup.split`. Uses of the ``catch()`` class method
+should be replaced with either ``except*`` clauses (on Python 3.11+) or the ``catch()``
+context manager provided by the backport_.
```
oremanj, line 8: Please document the differences between the replacements and the old functions.

oremanj:
Thank you for working on this!

We should have some tests that actually use `except*` on 3.11. Ideally we could rewrite them (maybe automatically?) to use `with catch()` on <3.11 (and on 3.11 too, for good measure) but getting that right is probably annoying so I will accept a 3.11-only version for now (using exec() or similar to avoid failing at test collection time on older versions). But we can't in good conscience advertise a feature that we're not testing.

BTW I think this is a headline feature, not just a deprecation/removal. Probably it deserves entries in both sections.

It would be nice to raise something that can be caught as ExceptionGroup if it doesn't contain any BaseExceptions. You could probably do this using two subclasses of MultiError and some cleverness in `__new__`. That doesn't need to be in this PR though, since our current users are used to all MultiErrors being BaseExceptions.

We need a plan for how we're going to eventually get rid of MultiError. It would be good to message that alongside the discussion of this transition so that people know what to expect. It doesn't have to be part of this PR but I'd like to have something before we release, and I'm happy to work on it. There should also be documentation published for the `exceptiongroup` backport package before a release occurs. If that's not feasible we can make up for it in our own documentation, but the current level of documentation isn't sufficient when core functionality like `catch()` isn't fully documented.

oremanj (changes requested):

docs/source/reference-core.rst
```diff
@@ -709,18 +711,46 @@ limitation. Consider code like::
...
+
+    from exceptiongroup import catch
+
+    def handle_keyerror(exc):
+        ...  # handle each KeyError
+
+    def handle_indexerror(exc):
+        ...  # handle each IndexError
+
+    with catch({
```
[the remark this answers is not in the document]
agronholm: It's in the [README](https://pypi.org/project/exceptiongroup/). Let me know if that's not enough.

docs/source/reference-core.rst
```diff
@@ -641,7 +641,7 @@ crucial things to keep in mind:
 
   * Any unhandled exceptions are re-raised inside the parent task. If
     there are multiple exceptions, then they're collected up into a
-    single :exc:`MultiError` exception.
+    single ``BaseExceptionGroup`` or ``ExceptionGroup`` exception.
```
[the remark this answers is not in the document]
agronholm: Yes, in fact they can be now that I've pointed the intersphinx link explicitly to Python 3.11.
```suggestion
    single :exc:`BaseExceptionGroup` or :exc:`ExceptionGroup` exception.
```

agronholm:
> We should have some tests that actually use except* on 3.11. Ideally we could rewrite them (maybe automatically?) to use with catch() on <3.11 (and on 3.11 too, for good measure) but getting that right is probably annoying so I will accept a 3.11-only version for now (using exec() or similar to avoid failing at test collection time on older versions). But we can't in good conscience advertise a feature that we're not testing.

Why do we need to duplicate the tests already performed in CPython and the exceptiongroup backport?
As a side note, such tests can be left uncollected if they are isolated in their own specially named test file when we have an appropriate collection hook in place.

> We need a plan for how we're going to eventually get rid of MultiError. It would be good to message that alongside the discussion of this transition so that people know what to expect.

It's been deprecated as it says in the release notes – I thought that was obvious enough? Users will see deprecation warnings which tell them to switch to (Base)ExceptionGroup. So, wait maybe 1-2 releases and then just replace it with (Base)ExceptionGroup. Done?

newsfragments/2211.deprecated.rst
```diff
@@ -0,0 +1,12 @@
+``MultiError`` has been deprecated in favor of the standard :exc:`BaseExceptionGroup`
+(introduced in :pep:`654`). On Python < 3.11, this exception and its derivative
```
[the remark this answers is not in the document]
agronholm:
```suggestion
(introduced in :pep:`654`). On Python versions below 3.11, this exception and its derivative
```

docs/source/tutorial/echo-server.py
```diff
@@ -11,6 +11,7 @@
 
 CONNECTION_COUNTER = count()
 
+
```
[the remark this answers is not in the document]
agronholm:
```suggestion
```

trio/_highlevel_open_tcp_stream.py
```diff
@@ -114,8 +116,12 @@ def close_all():
...
+                raise MultiError(errs)
```
agronholm, line 124:
```suggestion
            raise MultiError(errs)
```

trio/_core/tests/test_run.py
agronholm: This was a leftover from the original test code which did return `None` or `exc`. I've removed that now.

trio/tests/test_util.py
agronholm: Done.

trio/_core/tests/test_multierror.py
agronholm: I'll look into adopting this in `exceptiongroup`.

newsfragments/2211.deprecated.rst
agronholm: In the release notes, or in the actual documentation?

docs/source/reference-core.rst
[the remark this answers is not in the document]
agronholm: What exact semantics have been left undocumented there? It would be easier for me to fill in the holes if I knew what they are.

pquentin:
> Do I create a new news fragment with the same issue ID, i.e. 2211.headline.rst?

Yes, please!

njsmith:
From a quick skim of the doc changes here, it looks like with this PR, nurseries only raise an `ExceptionGroup` if multiple tasks raised an exception, and if only one task raises an exception then it is *not* wrapped. Is that correct?

Unfortunately one of the core design decisions for `ExceptionGroup` was to not try to support this behavior; we have to wrap unconditionally (and that's probably what's going to break users the most).

agronholm:
> From a quick skim of the doc changes here, it looks like with this PR, nurseries only raise an `ExceptionGroup` if multiple tasks raised an exception, and if only one task raises an exception then it is _not_ wrapped. Is that correct?
> 
> Unfortunately one of the core design decisions for `ExceptionGroup` was to not try to support this behavior; we have to wrap unconditionally (and that's probably what's going to break users the most).

I'm confused. This was done explicitly for backwards compatibility with `MultiError` – are you saying that I shouldn't have preserved that?

agronholm:
I'm going to need my last question answered before this PR can go forward. Anyone?

agronholm:
@oremanj @njsmith I've made all the changes you requested in #2301 and Gitter.

docs/source/reference-core.rst
```diff
@@ -709,18 +711,60 @@ limitation. Consider code like::
...
+exceptions themselves, and treat them as if they were standard `BaseExceptionGroup` or
+`ExceptionGroup` instances instead.
+
+The compatibility exception classes have one additional quirk: when only a single
+exception is passed to them, they will spit out that single exception from the
+constructor instead of the appropriate ``MultiError`` instance. This means that a
+nursery can pass through any arbitrary exception raised by either a spawned task or the
+host task. This behavior can be controlled by the ``strict_exception_groups=True``
+argument passed to either :func:`open_nursery` or :func:`run`. If enabled, a nursery
+will always wrap even single exception raised in it in an exception group.
```
oremanj, line 765: This paragraph is full of useful information but I don't think it's presented in a way that would be very clear to someone who hasn't been intimately involved with the process of developing the ExceptionGroup change. I'd suggest:

```
"Strict" versus "loose" ExceptionGroup semantics
++++++++++++++++++++++++++++++++++++++++++++++++

Ideally, in some abstract sense we'd want everything that *can* raise an
`ExceptionGroup` to *always* raise an `ExceptionGroup` (rather than, say, a single
`ValueError`). Otherwise, it would be easy to accidentally write something like ``except
ValueError:`` (not ``except*``), which works if a single exception is raised but fails to
catch _anything_ in the case of multiple simultaneous exceptions (even if one of them is
a ValueError). However, this is not how Trio worked in the past: as a concession to
practicality when the ``except*`` syntax hadn't been dreamed up yet, the old
``trio.MultiError`` was raised only when at least two exceptions occurred
simultaneously. Adding a layer of `ExceptionGroup` around every nursery, while
theoretically appealing, would probably break a lot of existing code in practice.

Therefore, we've chosen to gate the newer, "stricter" behavior behind a parameter
called ``strict_exception_groups``. This is accepted as a parameter to
:func:`open_nursery`, to set the behavior for that nursery, and to :func:`trio.run`,
to set the default behavior for any nursery in your program that doesn't override it.

* With ``strict_exception_groups=True``, the exception(s) coming out of a nursery will
  always be wrapped in an `ExceptionGroup`, so you'll know that if you're handling
  single errors correctly, multiple simultaneous errors will work as well.

* With ``strict_exception_groups=False``, a nursery in which only one task has failed
  will raise that task's exception without an additional layer of `ExceptionGroup`
  wrapping, so you'll get maximum compatibility with code that was written to
  support older versions of Trio.

To maintain backwards compatibility, the default is ``strict_exception_groups=False``.
The default will eventually change to True in a future version of Trio, once
Python 3.11 and later versions are in wide use.
```

docs/source/reference-core.rst
```diff
@@ -837,104 +867,6 @@ The nursery API
    See :meth:`~Nursery.start`.
 
 
-Working with :exc:`MultiError`\s
```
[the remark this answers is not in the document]
oremanj: There's nothing wrong with it, but it does feel like a downgrade when the previous set of docs had multiple examples, describing more edge cases, exception chaining, output, etc.

trio/__init__.py
```diff
@@ -113,6 +114,12 @@
         issue=1104,
         instead="trio.lowlevel.open_process",
     ),
+    "MultiError": _deprecate.DeprecatedAttribute(
+        value=_MultiError,
+        version="0.20.0",
+        issue=2211,
+        instead="BaseExceptionGroup (py3.11+) or exceptiongroup.BaseExceptionGroup",
```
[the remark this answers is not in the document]
oremanj: this comment is still relevant

trio/_core/_multierror.py
oremanj: This comment is still relevant. I double-checked; here's 3.11:
```
>>> eg = ExceptionGroup("hi", [Exception("one"), Exception("two")])
>>> hasattr(eg, "exceptions")
True
>>> hasattr(eg, "_exceptions")
False
>>>
```

trio/_core/_multierror.py
```diff
@@ -200,14 +203,17 @@ def __new__(cls, exceptions):
...
+            return super().__new__(cls, "multiple tasks failed", exceptions)
 
     def __str__(self):
         return ", ".join(repr(exc) for exc in self.exceptions)
 
     def __repr__(self):
         return "<MultiError: {}>".format(self)
 
+    def derive(self, __excs):
+        return MultiError(__excs, _collapse=False)
```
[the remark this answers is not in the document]
oremanj: Still relevant

trio/_core/_multierror.py
```diff
@@ -224,6 +230,11 @@ def filter(cls, handler, root_exc):
           ``handler`` returned None for all the inputs, returns None.
 
         """
+        warnings.warn(
```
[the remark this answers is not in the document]
oremanj: Still relevant

trio/_core/_multierror.py
```diff
@@ -359,158 +375,3 @@ def concat_tb(head, tail):
     for head_tb in reversed(head_tbs):
         current_head = copy_tb(head_tb, tb_next=current_head)
     return current_head
-
-
-################################################################
-# MultiError traceback formatting
```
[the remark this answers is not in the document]
oremanj: I checked -- it looks like you do monkeypatch `traceback` and `excepthook` but you're missing the handling for IPython and apport. So there is some regression here, which should be fixed.

trio/_core/_run.py
oremanj: I don't think you addressed any of these comments (this one or the ones in `collapse_exception_group()` that it's referring to).

trio/_core/_run.py
```diff
@@ -447,13 +472,7 @@ def __enter__(self):
...
         return self
 
-    def _exc_filter(self, exc):
-        if isinstance(exc, Cancelled):
-            self.cancelled_caught = True
-            return None
-        return exc
-
-    def _close(self, exc):
+    def _close(self, exc, collapse=True):
```
oremanj, line 475: The default for collapse should be false, so that passing through a normal CancelScope (that is not attached to a nursery) doesn't strip ExceptionGroup layers.

trio/_core/tests/test_multierror.py
oremanj: I'm all in favor of moving tests of `exceptiongroup` functionality to `exceptiongroup`, but it's important to me that we not throw tests away without moving them.

trio/_core/tests/test_multierror.py
```diff
@@ -414,359 +397,22 @@ def test_assert_match_in_seq():
...
-            # override it
-            "custom running!",
-            # The MultiError
-            "MultiError:",
-        ],
-        completed.stdout.decode("utf-8"),
-    )
+def test_base_multierror():
+    """
+    Test that MultiError() witho at least one base exception will return a MultiError
```
oremanj, line 402: s/witho/with/

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -138,7 +139,7 @@ async def open_tcp_listeners(port, *, host=None, backlog=None):
             errno.EAFNOSUPPORT,
             "This system doesn't support any of the kinds of "
             "socket that that address could use",
-        ) from trio.MultiError(unsupported_address_families)
+        ) from MultiError(unsupported_address_families)
```
oremanj, line 142: I think you may as well just change this to ExceptionGroup, since as a `__cause__` it's not directly exposed to users' `except` clauses

trio/_highlevel_open_tcp_stream.py
```diff
@@ -364,7 +368,7 @@ async def attempt_connect(socket_args, sockaddr, attempt_failed):
             msg = "all attempts to connect to {} failed".format(
                 format_host_port(host, port)
             )
-            raise OSError(msg) from trio.MultiError(oserrors)
+            raise OSError(msg) from MultiError(oserrors)
```
oremanj, line 371: I think this one can become ExceptionGroup too, for the same reason

oremanj (commented):
Thanks for the updates. In general the direction here seems good, but there are still a number of things I pointed out in the initial review that you don't seem to have responded to, and I think some changes will need to be made to `exceptiongroup` also to maintain feature parity with current Trio. I've noted those inline where applicable.

docs/source/reference-core.rst
```diff
@@ -709,18 +711,60 @@ limitation. Consider code like::
...
+
+To catch individual exceptions encapsulated in an exception group, the ``except*``
+clause was introduced in Python 3.11 (:pep:`654`). Here's how it works::
+
+    try:
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(broken1)
+            nursery.start_soon(broken2)
+    except* KeyError:
+        ...  # handle each KeyError
```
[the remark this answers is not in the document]
agronholm: It's hard to believe no one else brought this up in the backport. I will change its behavior to match.

trio/_highlevel_open_tcp_stream.py
agronholm: Done.

trio/_highlevel_open_tcp_listeners.py
agronholm: Done.

trio/_core/tests/test_multierror.py
agronholm: Done.

docs/source/reference-core.rst
agronholm: Updated.

docs/source/reference-core.rst
```diff
@@ -709,18 +711,60 @@ limitation. Consider code like::
...
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(broken1)
+            nursery.start_soon(broken2)
+
+.. hint:: If your code, written using ``except*``, would set local variables, you can do
+    the same with handler callbacks as long as you declare those variables ``nonlocal``.
+
+For reasons of backwards compatibility, nurseries raise ``MultiError`` and
+``NonBaseMultiError`` which inherit from `BaseExceptionGroup` and `ExceptionGroup`,
+respectively. Users should refrain from attempting to raise or catch the trio specific
```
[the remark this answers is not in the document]
agronholm: Done.

docs/source/reference-core.rst
```diff
@@ -709,18 +711,60 @@ limitation. Consider code like::
...
+        IndexError: handle_indexerror
+    }):
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(broken1)
+            nursery.start_soon(broken2)
+
+.. hint:: If your code, written using ``except*``, would set local variables, you can do
+    the same with handler callbacks as long as you declare those variables ``nonlocal``.
+
+For reasons of backwards compatibility, nurseries raise ``MultiError`` and
```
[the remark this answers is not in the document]
agronholm: Done.

docs/source/reference-core.rst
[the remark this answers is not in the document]
agronholm: https://github.com/agronholm/exceptiongroup/pull/10

docs/source/reference-core.rst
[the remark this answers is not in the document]
agronholm: Ok, I'll restore and update the docs.

newsfragments/2211.deprecated.rst
oremanj: I think an overview in the release notes would be helpful. The documentation explains the new behavior but doesn't particularly contrast it with the old behavior.

docs/source/reference-core.rst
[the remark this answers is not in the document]
oremanj: That's definitely better, thanks. I still think there's an important behavioral difference from py3.11, which I noted in a comment on that PR.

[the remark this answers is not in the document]
agronholm: Fixed in 1.0.0rc7.

agronholm:
I cannot figure out how to respond to some comments properly as they do not have a "reply" field, but regarding the IPython and apport traceback patches: I don't think trio should be monkey patching these. Given that this is a language feature now, IPython ought to pick this up upstream, using the backport for earlier than Python 3.11. As for apport, the same probably applies. How many trio users would care about this anyway?

agronholm:
I've collected the remaining issues to be resolved here:

- Restoring examples for multierrors
- `MultiError._exceptions` (line 182)
- Document why _collapse=False in `MultiError.derive()`
- Use `trio.warn_deprecated()` in `MultiError.filter()`
- Explain that `ExceptionGroup` does backport traceback formatting
- IPython/apport support missing (should be upstreamed)
- `collapse_exception_group()` should only modify `MultiErrors`?
- Don't throw away frames added to the exception group after if was raised, in `collapse_exception_group()`
- Should `MultiError._close()` have `collapse=False` by default?
- Should we use `MultiError.filter()` internally?

trio/_core/_run.py
agronholm: > The new behavior here has a number of differences from the old

What differences are you referring to? We now have the `strict_exception_groups` option which should be respected, and using `MultiError.filter()` would unconditionally flatten the group.

trio/_core/_run.py
agronholm: It only affects MultiErrors, and if I did this, `test_cancel_scope_multierror_filtering()` would start failing, as it **expects** the `MultiError` passing through the cancel scope to be collapsed into the contained `KeyError`.

docs/source/reference-core.rst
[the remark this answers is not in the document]
agronholm: I took the time to evaluate the examples from the removed section, but since the catching API in the backport works in such a different way, I'm not sure what there is to document. The backport doesn't provide any functionality that the `except*` syntax doesn't. There's no exception conversion, and to reraise exceptions, you just use `raise` as usual. That said, I've amended the documentation to at least show how to iterate over the exceptions in the group.

agronholm:
I've added a tracking issue to get exceptiongroup support to IPython and Apport here: https://github.com/agronholm/exceptiongroup/issues/23

oremanj:
Can we make it so deprecated names don't show up in tracebacks? Right now tracebacks show `trio.NonBaseMultiError` or `trio.MultiError`, but you get a warning if you type those names. I think ideally we would have `trio.BaseExceptionGroup` and `trio.ExceptionGroup` as the things Trio actually throws, which inherit from `trio.MultiError` and `trio.NonBaseMultiError` respectively (so that old code `except trio.MultiError:` can still catch them), but are not deprecated. If we go this route I actually think we could get rid of `NonBaseMultiError` entirely, which is nice because that's an awkward name.

oremanj:
> regarding the IPython and apport traceback patches: I don't think trio should be monkey patching these. Given that this is a language feature now, IPython ought to pick this up upstream, using the backport for earlier than Python 3.11.

In terms of the ideal long-run state, I agree with you. In terms of the actual experience of our users when this is released, dropping the IPython monkeypatch is a major regression: exception group tracebacks will silently show only some of the frames. I think we need to carry that monkeypatch until IPython actually upstreams proper exception group support.

> As for apport, the same probably applies. How many trio users would care about this anyway?

Every single one of our users on Ubuntu will get a warning when they load Trio if we don't continue to carry the apport workaround. It was a very frequently reported issue before we added the workaround. I am not comfortable with removing it.

trio/_core/_run.py
oremanj: The differences/issues I'm still tracking are the traceback concatenation (which you're now doing, but in the wrong order) and the extra collapsing added by a cancel scope around a strict_exception_groups nursery. I commented on both of those above. I think I originally had a third comment but removed it because I realized it was mistaken; sorry for the confusion.

trio/_core/_run.py
```diff
@@ -910,7 +958,9 @@ def _child_finished(self, task, outcome):
         self._check_nursery_closed()
 
     async def _nested_child_finished(self, nested_child_exc):
-        """Returns MultiError instance if there are pending exceptions."""
+        """
```
oremanj, line 961: This isn't full documentation for the function (nor does it need such, it's not public) so should probably be an ordinary `# comment`. Also the result might not be a MultiError instance if we're in loose mode for this nursery.

oremanj, line 961: This isn't full documentation for the function (nor does it need such, it's not public) so should probably be an ordinary `# comment`. Also the result might not be a MultiError instance if we're in loose mode for this nursery.

trio/_core/tests/test_run.py
```diff
@@ -1853,18 +1857,12 @@ async def __anext__(self):
...
-                else:  # pragma: no cover
-                    return exc
-
-            with _core.MultiError.catch(handle):
+            try:
                 async with _core.open_nursery() as nursery:
                     for i, f in enumerate(nexts):
                         nursery.start_soon(self._accumulate, f, items, i)
+            except ExceptionGroup as excgroup:
+                got_stop = bool(excgroup.split(StopAsyncIteration))
```
oremanj, line 1865: split() returns a 2-tuple so this will always be true...

oremanj (commented):
This is looking good. I think you've addressed all my previous comments, so it's just the ones I left here that remain. I'm sorry for the lengthy delay in reviewing; I will endeavor to respond within one week to your remaining updates on this PR.

agronholm:
> > regarding the IPython and apport traceback patches: I don't think trio should be monkey patching these. Given that this is a language feature now, IPython ought to pick this up upstream, using the backport for earlier than Python 3.11.
> 
> In terms of the ideal long-run state, I agree with you. In terms of the actual experience of our users when this is released, dropping the IPython monkeypatch is a major regression: exception group tracebacks will silently show only some of the frames. I think we need to carry that monkeypatch until IPython actually upstreams proper exception group support.
> 
> > As for apport, the same probably applies. How many trio users would care about this anyway?
> 
> Every single one of our users on Ubuntu will get a warning when they load Trio if we don't continue to carry the apport workaround. It was a very frequently reported issue before we added the workaround. I am not comfortable with removing it.

Even though the Ubuntu wiki states that Apport is [disabled by default](https://wiki.ubuntu.com/Apport)? And what warning would they get?

trio/_core/_multierror.py
```diff
@@ -224,7 +236,9 @@ def filter(cls, handler, root_exc):
           ``handler`` returned None for all the inputs, returns None.
 
         """
-
+        warn_deprecated(
+            "MultiError.filter()", "0.22.0", instead="MultiError.split()", issue=2211
```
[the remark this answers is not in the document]
agronholm: Fixed.

newsfragments/2211.headline.rst
```diff
@@ -0,0 +1,12 @@
+``MultiError`` has been deprecated in favor of the standard :exc:`BaseExceptionGroup`
+(introduced in :pep:`654`). On Python versions below 3.11, this exception and its derivative
+:exc:`ExceptionGroup` are provided by the backport_. Trio still raises ``MultiError``,
+but it has been refactored into a subclass of :exc:`BaseExceptionGroup` which users
+should catch instead of ``MultiError``. Uses of the ``filter()`` class method should be
```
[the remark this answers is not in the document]
agronholm: Fixed.

trio/_core/_multierror.py
```diff
@@ -236,12 +250,22 @@ def catch(cls, handler):
           handler: as for :meth:`filter`
 
         """
+        warnings.warn(
```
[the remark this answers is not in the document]
agronholm: Fixed.

trio/_core/tests/test_run.py
agronholm: It now turns the first item of the tuple into a boolean.

trio/_core/_run.py
agronholm: Turned the docstring into a comment and made the wording more precise.

trio/_core/_run.py
```diff
@@ -116,6 +119,31 @@ class IdlePrimedTypes(enum.Enum):
...
+    for i, exc in enumerate(exceptions):
+        if isinstance(exc, MultiError):
+            new_exc = collapse_exception_group(exc)
+            if new_exc is not exc:
+                modified = True
+                exceptions[i] = new_exc
+
+    if len(exceptions) == 1:
+        exceptions[0].__traceback__ = concat_tb(
+            exceptions[0].__traceback__, excgroup.__traceback__
```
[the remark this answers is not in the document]
agronholm: I've reversed the order now.

agronholm:
> In terms of the ideal long-run state, I agree with you. In terms of the actual experience of our users when this is released, dropping the IPython monkeypatch is a major regression: exception group tracebacks will silently show only some of the frames. I think we need to carry that monkeypatch until IPython actually upstreams proper exception group support.

I've restored the IPython custom error handler, even though I still feel this is the wrong place for it. It also now supports (Base)ExceptionGroup.

trio/_core/_multierror.py
```diff
@@ -174,21 +180,21 @@ class MultiError(BaseException):
...
+    def __init__(self, exceptions, *, _collapse=True):
+        # Avoid double initialization when _collapse is True and exceptions[0] returned
+        # by __new__() happens to be a MultiError and subsequently __init__() is called.
+        if _collapse and getattr(self, "exceptions", None) is not None:
+            # This exception was already initialized.
             return
-        self.exceptions = exceptions
 
-    def __new__(cls, exceptions):
+        super().__init__("multiple tasks failed", exceptions)
```
Zac-HD, line 190: `"multiple tasks failed"` might not be true, if `strict_exception_groups=True` - do we want to change the message?

trio/_core/_multierror.py
```diff
@@ -361,95 +391,7 @@ def concat_tb(head, tail):
...
-
-
-def trio_excepthook(etype, value, tb):
-    for chunk in traceback.format_exception(etype, value, tb):
-        sys.stderr.write(chunk)
-
-
-monkeypatched_or_warned = False
-
+# Remove when IPython gains support for exception groups
```
Zac-HD, line 394: Could link to https://github.com/ipython/ipython/issues/13753

trio/_core/_multierror.py
agronholm: Then we'd have to check the entire exception hierarchy to see if they collapse to a single exception, wouldn't we?

Zac-HD: Hmm, yeah, that's a good argument against.

trio/_core/_multierror.py
agronholm: Done.

docs/source/design.rst
```diff
@@ -465,9 +465,6 @@ There are two notable sub-modules that are largely independent of
 the rest of Trio, and could (possibly should?) be extracted into their
 own independent packages:
```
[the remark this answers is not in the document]
agronholm: Updated.

oremanj:
> I've restored the IPython custom error handler, even though I still feel this is the wrong place for it.

Thanks. Hopefully IPython will implement proper EG support (including the backport) and we can remove Trio's patching of it soon.

> Even though the Ubuntu wiki states that Apport is [disabled by default](https://wiki.ubuntu.com/Apport)? And what warning would they get?

I don't have easy access to an Ubuntu machine right now, but IIRC ubuntu installs a custom `sys.excepthook` unconditionally, which then calls the default one unless Apport is enabled. But it still looks like a custom `excepthook` so Trio won't install its own, so you don't get MultiError tracebacks.

I guess the exceptiongroup backport doesn't print a warning like Trio previously would if it fails to patch `sys.excepthook`. So users won't get a confusing warning... instead, they'll silently fail to see the "multi" parts of the traceback, which is even worse UX. I think that warning is fairly important in helping people understand what's going on.

agronholm:
> I guess the exceptiongroup backport doesn't print a warning like Trio previously would if it fails to patch `sys.excepthook`. So users won't get a confusing warning... instead, they'll silently fail to see the "multi" parts of the traceback, which is even worse UX. I think that warning is fairly important in helping people understand what's going on.

I'm running Ubuntu on my home server and I can confirm that outside of a virtualenv, the apport hook is indeed present and will prevent exceptiongroup from installing its own hook. So, to not degrade the trio UX, I will fix this by adapting the old trio apport exceptionhook.

trio/_core/_run.py
```diff
@@ -116,6 +119,31 @@ class IdlePrimedTypes(enum.Enum):
...
+        )
+        return exceptions[0]
+    elif modified:
+        return excgroup.derive(exceptions)
```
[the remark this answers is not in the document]
agronholm: Added a test and ensured with a breakpoint that this branch is getting tested.

agronholm:
Code-wise this PR should be fine now.

trio/_core/_multierror.py
```diff
@@ -414,3 +418,35 @@ def trio_show_traceback(self, etype, value, tb, tb_offset=None):
...
+    and getattr(sys.excepthook, "__name__", None) == "apport_excepthook"
+):
+    from types import ModuleType
+
+    import apport_python_hook
+    from exceptiongroup import format_exception
+
+    assert sys.excepthook is apport_python_hook.apport_excepthook
+
+    def replacement_webhook(etype, value, tb):
```
oremanj, line 446: this isn't a webhook, can you choose a different name?

trio/_core/_multierror.py
```diff
@@ -414,3 +418,35 @@ def trio_show_traceback(self, etype, value, tb, tb_offset=None):
...
+        sys.stderr.write("".join(format_exception(etype, value, tb)))
```
oremanj, line 447: The original Trio hook did `for chunk in format_exception(...): sys.stderr.write(chunk)`. I sort of prefer that approach, so that if something in format_exception() raises an exception, you can at least see the traceback before the bad part. (format_exception() is a generator)

trio/_core/_run.py
```diff
@@ -475,7 +475,7 @@ def __enter__(self):
             task._activate_cancel_status(self._cancel_status)
         return self
 
-    def _close(self, exc, collapse=True):
+    def _close(self, exc, collapse=False):
```
oremanj, line 478: It doesn't look like this `collapse` parameter is used anymore, can you remove it?

trio/_core/_multierror.py
agronholm: Fixed. That was a brain fart, sorry.

trio/_core/_multierror.py
agronholm: But `format_exception()` returns a list, not a generator, so I don't see how that would help.

trio/_core/_run.py
agronholm: Done.

trio/_core/_multierror.py
oremanj: Ah you're right -- the Trio hook used a generator but the exceptiongroup one doesn't. I don't think it matters enough to change it.

oremanj (approved):

pquentin:
Thanks for your hard work, @agronholm and @oremanj! We're lucky to have you. I'm really excited!

excited

Will release this tomorrow.

PR #2220, opened 2022-01-26
Bump ipython to 7.31.1

pquentin:
GitHub keeps annoying me about https://github.com/advisories/GHSA-pq7m-3gw7-gq5x so I figured I would cave in and update to 7.x (since 8.x does not support Python 3.7).

I also made sure to run pip-compile using Python 3.7, which restores a few dependencies that are probably not needed with later versions.

### DIFF
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -1,7 +1,8 @@
 # For tests
 pytest >= 5.0         # for faulthandler in core
 pytest-cov >= 2.6.0
-ipython               # for the IPython traceback integration tests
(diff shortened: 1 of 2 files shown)


agronholm (approved):
LGTM.

agronholm:
About that typing-extensions conflict: does ipython 7 require typing-extensions < 4?

pquentin:
No, it's just that I merged the typing-extensions bump and it caused a conflict. I've solved that conflict, will merge if tests pass

PR #2256, opened 2022-02-21
Merge two competing newsfragments

pquentin:
It turns out towncrier will ignore the removal and only consider the
feature, which means we have to merge them.


njsmith:
Well, that's annoying. Seems like a clear towncrier bug. This PR LGTM though. (Or if you wanted to just fix up the changelog by hand in the release branch that'd be fine too.)

pquentin:
Thanks, I've reported the issue here: https://github.com/twisted/towncrier/issues/372

PR #2289, opened 2022-04-13
unawesome unmaintained libraries

belm0:
* purerpc - not maintained, depends on ancient anyio version
  * trio-jsonrpc - not maintained

### DIFF
--- a/docs/source/awesome-trio-libraries.rst
+++ b/docs/source/awesome-trio-libraries.rst
@@ -83,12 +83,6 @@ Stream Processing
 * `Slurry <https://github.com/andersea/slurry>`__ - Slurry is a microframework for building reactive, data processing applications with Trio.
 
 
-RPC
----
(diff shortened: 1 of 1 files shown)


pquentin:
I'm inclined to merge this, but... do they not work?

belm0:
> I'm inclined to merge this, but... do they not work?

I think if someone attempts to use them, there will only be frustration.

belm0:
(I'm intending to fix the bitrot of purerpc for private use by our project.  If python-trio org wants to own the fork, I'd be happy to share.)

pquentin:
It's true the purepc annoyingly puts upper bound on all its dependencies. Of course, the python-trio org will own the fork since you're a member, but that does not mean you will get contributors :)

belm0:
> Of course, the python-trio org will own the fork since you're a member, but that does not mean you will get contributors :)

The point is not necessary to get contributors, but for the code to be able to be fixed or maintained when someone in the community is motivated to do it (and after I've lost interest).

Actually, I was able to contact the purerpc author and he's open to transferring the repo & pypi to python-trio.

PR #2319, opened 2022-05-15
Fixed remaining Python 3.11 test failures

agronholm:
This also removes 3.8-dev, 3.9-dev and 3.10-dev from the test matrix. Let me know if those were needed for something and I will restore them.


trio/tests/test_exports.py
```diff
@@ -81,6 +81,9 @@ def public_modules(module):
     + re.escape("The distutils.sysconfig module is deprecated, use sysconfig instead")
     + ":DeprecationWarning",
 )
+@pytest.mark.filterwarnings(
+    "ignore:module 'sre_constants' is deprecated:DeprecationWarning",
+)
```
pquentin, line 86: What is responsible for this warning? It would be nice to quickly open an issue for the relevant project, as we did above for Astroid. That way we can track the issue and remove the clutter when the issue is fixed: https://github.com/python-trio/trio/pull/2320

pquentin (approved):
Thanks! I'm happy to remove the other 3.x-dev jobs. They were added as a courtesy to the CPython developers as they allow catching bugs before releases - they don't really help us, no.

I've added a non blocking comment.

trio/tests/test_exports.py
agronholm: It's caused by an old vendored version of pyparsing from `pkg_resources`.

agronholm: I checked and it looks like the other warning filters on that test can be removed.

agronholm: I found the relevant, existing issue in setuptools and I've added a link to that.

PR #2344, opened 2022-06-07
replace `async_cm` decorator:

lojack5:
Changes usages of `@async_cm` to using a mixin class for the same purpose.  No mechanical reason for it, but both pylance and pylint (possibly others, I didn't test past those and pycharm) complain about wrapped classes (ex: `trio.Lock`) not implementing `__aenter__` / `__aexit__`.  Really just a small change for my sanity.

### DIFF
--- a/trio/_sync.py
+++ b/trio/_sync.py
@@ -87,21 +87,14 @@ class Event(metaclass=Final):
         return _EventStatistics(tasks_waiting=len(self._tasks))
 
 
-def async_cm(cls):
+class AsyncContextManagerMixin:
     @enable_ki_protection
     async def __aenter__(self):
         await self.acquire()
 
-    __aenter__.__qualname__ = cls.__qualname__ + ".__aenter__"
-    cls.__aenter__ = __aenter__
-
     @enable_ki_protection
     async def __aexit__(self, *args):
         self.release()
-
-    __aexit__.__qualname__ = cls.__qualname__ + ".__aexit__"
-    cls.__aexit__ = __aexit__
-    return cls
 
 
 @attr.s(frozen=True)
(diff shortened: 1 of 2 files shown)


pquentin (commented):
I'm happy to merge this, but note that in the past when we made changes specifically for linters we tested them: https://github.com/python-trio/trio/blob/09fcd83e98d453d4039909902a2a9f8313629cff/trio/tests/test_exports.py

Do you want to do that here too or are you OK with the total lack of guarantee going forward?

lojack5:
I'm OK with going forward, but if you (or anyone else) have ideas on *how* I would write the test, I'll be happy to add them.  Specifically I'd need a way to test that linters don't flag on code like this:
```python
lock = trio.Lock()
with lock:
    ...
```
For the types affected (`CapacityLimiter`, `Semaphore`, `Lock`, `Condition`).  I'm not sure how to test linters (and I'm assuming it's different for each one) doesn't flag the `with lock:`.  I know for pylint it's `E1701`, and pylance reports it as a `reportGeneralTypeIssues` error.  Now pylance uses pyright under the hood, but I'm having issues getting pyright working on its own, some sort of nodedev issue).

Fuyukai:
I see literally no possible reason to ever have a linter test for this. That's not testing the change, that's testing the linter.

lojack5:
Sounds good to me, ready to merge then.

pquentin:
> I see literally no possible reason to ever have a linter test for this. That's not testing the change, that's testing the linter.

Well it's testing both? Various people worked hard to make Trio less magic so that tools could provide better lint, type hints, completion, etc. And those improvements were tested, see my comment above. This is very similar in my opinion. While linters can always improve, they can never fully understand all magic without solving the halting problem, so helping them and making sure we don't get back to the old behavior can be worth it.

PR #2354, opened 2022-06-19
Guard rail cancellable argument of thread dispatch

richardsheridan:
The cancellable argument is typed bool but at runtime will work with any object with truthiness. However, if someone passes something like an numpy ndarray, with no notion of truthiness, the error is raised in an abort function and causes a TrioInternalError.

There are other unguarded uses of user arguments in abort functions, but they are in the lowlevel namespace, and aren't expected to have robust guardrails.

### DIFF
--- a/trio/_threads.py
+++ b/trio/_threads.py
@@ -138,6 +138,7 @@ async def to_thread_run_sync(sync_fn, *args, cancellable=False, limiter=None):
 
     """
     await trio.lowlevel.checkpoint_if_cancelled()
+    cancellable = bool(cancellable)  # raise early if cancellable.__bool__ raises
     if limiter is None:
         limiter = current_default_thread_limiter()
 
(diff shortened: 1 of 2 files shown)


trio/_threads.py
```diff
@@ -138,6 +138,7 @@ async def to_thread_run_sync(sync_fn, *args, cancellable=False, limiter=None):
 
     """
     await trio.lowlevel.checkpoint_if_cancelled()
+    cancellable = bool(cancellable)  # Sanitize for use in abort function
```
pquentin, line 141: I'm not sure the comment adds much, and I'd prefer if we could avoid ableist language here

pquentin (approved):
Looks good, thanks!

trio/_threads.py
richardsheridan: I do think a comment is needed to associate it with the abort function, otherwise it could look like an unnecessary step (even to my future self!) Maybe "normalize" or "canonicalize"?

pquentin: "Ensure is bool even if _bool_ not implemented"?

richardsheridan: Good suggestion, I went slightly broader in the amended and rebased commit

pquentin: Oh I understand what this does now, completely missed it with the previous comment

pquentin (approved):
Thanks!

Issue #892, opened 2019-01-29
Awaiting futures versus starting coroutines (How opinionated is trio?)


abetkin:
Hi, @njsmith and trio folks!

As I see in trio docs, the main API is letting trio run my async functions for me.
Like this:

```python
async def hobby():
    # long-running
    dream = await trio.sleep(a_lot)
    return dream

async def someday():
    result = await hobby()
    with trio.open_nursery() as ns:
        counter = 0
        for b in everybody:
            ns.start_soon(boast, b, result, counter)
    print('score', counter)

async def boast(someone, something, counter):
    '''
    if `someone` has envy for `something`, please increment the `counter`.
    '''
```

This is a valid approach, and probably a good style, since it passes the arguments explicitly
to each coroutine. There is another aproach though.

Please, say if I'm wrong, but a task in trio docs usually equals to coroutine, i.e., not yet started or even created. In asyncio, a task is a Future. This aproach tries to use them:

```python
counter = 0

async def boast(somebody):
    result = await hobby_task
    if result in somebody.envy_list:
        counter += 1


async def main():
    hobby_task = asyncio.create_task(hobby())
    tasks = [
        asyncio.create_task(boast(b))
        for b in everybody
    ]
    await asyncio.gather(hobby_task, *tasks)
    print('score', counter)
```

The difference is that we are using global variables: `hobby_task` and `boast` tasks. Probably, an antipattern. But imagine we are handling a web request, and a web request can serve as a natural scope, so globally-scoped tasks are actually scoped by a web request.

In many cases, it makes sense to execute async tasks once, and let other async tasks use the result.
In our example, that task is `hobby()`. Can't do `await hobby()` in any coroutine because it's a lengthy task.

Also, you can't cancel the hobby task from another coroutine: not without the additional knowledge, which other coroutines are using it.

My question is: can trio be used with this application design? Or does it have parts that can be reused for this?

Basically, tasks (started coroutines) can be viewed as a functional components that form some kind of dependency tree.
Theoretically, the nodes in this tree can do supervision/nursery for the dependent nodes.
And the tree knows all the dependencies, so it knows which tasks can be cancelled safely.

Also, do you see any issues with this design? Maybe, you know async frameworks that are more suited for it?

P.S.
This issue was born from me trying to adapt [apistar](https://github.com/encode/apistar/tree/version-0.5.x) by @tomchristie to the needs of a project at work.
[...]

PR #2374, opened 2022-07-07
avoid cancel_handle close race with weakref.finalize

### LINKED ISSUE #1271: Race condition in trio.hazmat.WaitForSingleObject

richardsheridan:
Fixes #1271. This is essentially the reference counting scheme from that issue, but python is doing the refcounting. I've also made sure the "nonblocking" case is a full checkpoint, win32 errors don't pass silently, and the logic is protected from `KeyboardInterrupt`.

I wrote this in response to https://github.com/python-trio/trio/pull/1834#issuecomment-1174512119 where @honglei seemed more interested in the bugfix than the features of #1834. (Hopefully you can test this branch in your application?)

Ideally, if someone reviews this they would also have a look at #1834 so I can either close or merge that one.

### DIFF
--- a/trio/_wait_for_object.py
+++ b/trio/_wait_for_object.py
@@ -1,15 +1,18 @@
-import math
-from . import _timeouts
+import weakref
+
 import trio
+from ._core import enable_ki_protection
 from ._core._windows_cffi import (
     ffi,
     kernel32,
     ErrorCodes,
     raise_winerror,
     _handle,
+    _check,
 )
 
 
+@enable_ki_protection
 async def WaitForSingleObject(obj):
     """Async and cancellable variant of WaitForSingleObject. Windows only.
 
@@ -20,6 +23,7 @@ async def WaitForSingleObject(obj):
       OSError: If the handle is invalid, e.g. when it is already closed.
 
     """
+    await trio.lowlevel.checkpoint_if_cancelled()
     # Allow ints or whatever we can convert to a win handle
     handle = _handle(obj)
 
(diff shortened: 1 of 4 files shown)


honglei:
at the beginning I wanna  make pynorm https://github.com/USNavalResearchLaboratory/norm  work with `asyncio` and `fastapi`,  so I need an async version of `WaitForSingleObject ` , then I found `trio`. 
But trio is not `asyncio`, https://github.com/python-trio/trio/issues/892 , 
so I need some time to learn/understand how the `trio` makes things work.

```python
import pynorm
instance:pynorm.Instance = pynorm.Instance()
async def watch_norm_events():
    '''
    NormGetNextEvent docs : non-blocking operation may be achieved by using the NormGetDescriptor() function
    to obtain a descriptor(int for Unix or HANDLE for WIN32) suitable for 
    asynchronous input/output (I/O) notification
    using such system calls the Unix select() or Win32 WaitForMultipleObjects() calls. 
    The descriptor is signaled when a notification event is pending and a call to 
    NormGetNextEvent() will not block
    '''
    global instance
    #instance.setDebugLevel(level=12)

    handle = instance.getDescriptor() #handler opened by norm lib
    while True:
        await WaitForSingleObject(handle)  # wait for event 
        event: pynorm.event.Event = instance.getNextEvent( )
        print(event) 
```

richardsheridan:
This PR has 100% diff branch coverage on my machine.

trio/_wait_for_object.py
```diff
@@ -30,24 +34,36 @@ async def WaitForSingleObject(obj):
...
         )
-    finally:
-        # Clean up our cancel handle. In case we get here because this task was
-        # cancelled, we also want to set the cancel_handle to stop the thread.
-        kernel32.SetEvent(cancel_handle)
-        kernel32.CloseHandle(cancel_handle)
+    except trio.Cancelled:
+        # Clean up our thread. We get here because this task was cancelled,
+        # so we also want to set the cancel handle to stop the thread.
+        _check(kernel32.SetEvent(cancel_handle))
```
njsmith, line 66: The failure to re-`raise` here is definitely a bug. I would just go back to using `finally:` ... setting the `cancel_handle` on success is harmless. (In general, `except trio.Cancelled` is a code smell.)

trio/_wait_for_object.py
```diff
@@ -30,24 +34,36 @@ async def WaitForSingleObject(obj):
     if retcode == ErrorCodes.WAIT_FAILED:
         raise_winerror()
     elif retcode != ErrorCodes.WAIT_TIMEOUT:
+        await trio.lowlevel.cancel_shielded_checkpoint()
```
njsmith, line 37: I think we can make this a plain `trio.lowlevel.checkpoint()` and remove the `checkpoint_if_cancelled` at the top of the function?

trio/_wait_for_object.py
```diff
@@ -30,24 +34,36 @@ async def WaitForSingleObject(obj):
...
+    weakref.finalize(cancel_handle, cleanup)
+
     try:
         await trio.to_thread.run_sync(
             WaitForMultipleObjects_sync,
             handle,
             cancel_handle,
             cancellable=True,
-            limiter=trio.CapacityLimiter(math.inf),
+            limiter=trio.CapacityLimiter(1),
```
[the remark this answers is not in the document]
richardsheridan: Nothing super important, but it avoids an import and an attribute lookup. Do you remember why the default thread limiter is not used?

trio/_wait_for_object.py
richardsheridan: Gosh, I guess I'm getting used to flake8-trio holding my hand! I'm happy to switch to finally though.

trio/_wait_for_object.py
richardsheridan: I thought `checkpoint_if_cancelled` was a nice micro-optimization for avoiding system calls if this function is already cancelled before entering, but a plain checkpoint is easier to understand for sure.

trio/_wait_for_object.py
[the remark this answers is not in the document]
richardsheridan: Just answering my own question here: https://github.com/python-trio/trio/issues/4#issuecomment-398967572

PR #2381, opened 2022-07-20
fix typo and extra sleep in test_unsafe_cancellable_kwarg

richardsheridan:
removes a subtle early execution of a time.sleep call in the main thread. All the time and deadline stuff is not actually necessary in the happy path.


pquentin (approved):
Thanks!

PR #2389, opened 2022-08-06
fix deadlock when deliver raises

richardsheridan:
I guess I'm building a reputation for finding trio bugs by writing embarrassingly bad code! Raising an exception in `deliver` argument to `start_thread_soon` will cause the next usage of the thread cache to deadlock, since the thread has stopped running but the cache thinks it can still assign jobs to it.

Printing out the traceback and carrying on is the obvious thing to do, but I'm open to suggestions on alternatives!

### DIFF
--- a/trio/_core/_thread_cache.py
+++ b/trio/_core/_thread_cache.py
@@ -1,3 +1,5 @@
+import sys
+import traceback
 from threading import Thread, Lock
 import outcome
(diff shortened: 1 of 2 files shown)


Fuyukai:
print()s in lib code is not really the best imo but it's better than silent exceptions and also better than deadlocks.

PR #2390, opened 2022-08-07
Clarify docs on checkpoints in async iterables

### LINKED ISSUE #2388: Clarify docs on checkpoints in async iterators
https://github.com/python-trio/trio/blob/077e8fc3f1634b42ef5024814898439d8a8430d4/docs/source/reference-core.rst#L92-L95

Does this mean *exactly* one checkpoint after the last iteration, or *at least* one?  I'm pretty sure it's the latter but that's not entirely obvious from reading, and adding one of those specifiers would remove any doubt🙂

Zac-HD:
Closes #2388

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -90,9 +90,9 @@ them. Here are the rules:
   exception, it might act as a checkpoint or might not.)
 
   * This includes async iterators: If you write ``async for ... in <a
-    trio object>``, then there will be at least one checkpoint before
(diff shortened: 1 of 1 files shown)


Sxderp:
nit: I think that comma is unnecessary.

Zac-HD:
Comma might not be *necessary*, but I like it 🤷‍♂️   Merging because there's no danger of breaking APIs for a docs change, there are no major objections, and I'd like to move on.

Issue #961, opened 2019-03-01
Proposal: make checkpoint_if_cancelled() sync-colored and rename it accordingly


oremanj:
I'd like to propose that we deprecate `checkpoint_if_cancelled()` in favor of a synchronous function that I'm thinking of as `trio.check_cancelled()` (happy to take suggestions on the name, or move into hazmat). This would raise `Cancelled` if the current scope is cancelled, and do nothing otherwise. i.e., it is a pure cancel point, not a schedule point, not even a schedule point if we're cancelled.

This becomes possible due to some recent or pending changes:
* #474, which permits the sync-color by saying we don't need to checkpoint before raising an exception (I'm assuming Cancelled counts as an exception for the purposes of that rule)
* #901, which means the Cancelled-raising path is just `raise Cancelled._init()`
* #958, which makes it fast to see if we're cancelled or not

And I think would have benefits for:
* #70, by implementing efficient `checkpoint()` as `await cancel_shielded_checkpoint(); check_cancelled()`
* #606, allowing `run_sync_in_worker_thread` functions to call check_cancelled() occasionally without messing with portals (because it's synchronous! and if/when #958 lands I think we can make the check_cancelled() operation sufficiently thread-safe)
* iirc #658 indicated some user interest in a sync-colored raise-cancelled operation?

Possible downside: a sync-colored function called from an async function can now call check_cancelled() and wind up raising Cancelled. Previously that couldn't happen. My feeling is that "functions that do something off-the-wall might raise an exception their caller wasn't expecting" is hardly a new problem, but maybe there's some aspect of what we'd be giving up here that I'm missing.

njsmith:
I think the first step here is to implement #474 – update docs and tests, and clean up some of the current clutter for checkpointing error paths. (Of course this last part may take a while and it doesn't have to be perfect, but I think doing at least some cleanup will help make it clearer what we want here.)

> implementing efficient checkpoint() as await cancel_shielded_checkpoint(); check_cancelled()

We could do this now. It's probably also easy to implement `checkpoint` as a dedicated trap, like `cancel_shielded_checkpoint`.

njsmith:
> #901, which means the Cancelled-raising path is just raise Cancelled._init()
> #958, which makes it fast to see if we're cancelled or not

[...]

PR #2392, opened 2022-08-08
Expand cancellation usability from native trio threads

### LINKED ISSUE #606: Propagating cancellation through threads

richardsheridan:
Fixes #606. I used types as suggested there to distinguish the different messages from the worker thread, but it also IMHO nicely refactored the logic of the `from_thread` functions; they are now thin wrappers with docstrings. The host task containing a call to `to_thread.run_sync` is reused to handle calls to `from_thread.run*` if `trio_token is None`. The context of the thread is copied and used back in the main thread to propagate any contextvar changes that may have happened in the thread. I'd be happy to hear suggestions for applying the context in `from_thread.run` that could avoid two extra checkpoints!

This PR reuses the thread local object in `_threads.py` to stash the `raise_cancel` provided to abort, which basically implements a suggestion by @njsmith in https://github.com/python-trio/trio/issues/961#issuecomment-501597755:

>Another idea: in `run_sync_in_thread`, start the thread, then block in `wait_task_rescheduled`. In our blocked state, if our >`abort_fn` is called, stash the `raise_fn` somewhere where the thread can see it. Then `trio.from_thread.check_cancel()` is just >something like `if THREAD_STATE.raise_fn is not None: raise_fn()`.

>This might be simpler, but I bet it will also have some complications by the time we figure out how to forward arbitrary >cancellations to the child thread, and also let the thread re-enter trio and deliver cancellations to *those* tasks.

This is used in the existing `from_thread` functions to avoid race conditions, but I would like to use this to allow a sqlite query to poll for cancellation. Currently, I would have to dig into the internals and use this recipe from @oremanj:

```
def make_cancel_poller():
    _cancel_status = trio.lowlevel.current_task()._cancel_status

    def poll_for_cancel(*args, **kwargs):
        if _cancel_status.effectively_cancelled:
            raise trio.Cancelled._create()

    return poll_for_cancel
```

Furthermore, that method is not thread safe if shielding is being toggled:

https://github.com/python-trio/trio/blob/2d62ff037f73ffd129485809ac0f14001e1eab6d/trio/_core/_run.py#L223-L231

So I added a new function `from_thread.check_cancelled` which is a very simple wrapper to grab the appropriate `raise_cancel` function if it is there and call it. A question though: Would it be better to call it or just return it?

```
if trio.from_thread.check_cancelled():
   ... # do cleanup and don't raise
```

might be more along the lines of what people are expecting compared to 

```
try:
    trio.from_thread.check_cancelled()
except:
    ... # do cleanup and don't raise
```

But I feel like people should be nudged to actually raise `Cancelled` if they are in fact responding to cancellation!


graingert:
I'd like to see threads be able to run code directly in the nursery that started it, this should fix this issue more robustly.

I'll edit this comment to link to what I mean shortly. Remind me if I don't

richardsheridan:
Is this similar to what you are suggesting: https://github.com/python-trio/trio/issues/606#issuecomment-515667078

richardsheridan:
@graingert I have implemented your request (finally!). Would you have a look at it?  Also, @agronholm this might have implications for AnyIO so it would be better to have your input sooner than later!

agronholm:
This looks very interesting, and I can use this. I'm looking forward to adding this to AnyIO once it's released for Trio.

richardsheridan:
it should be type-safe now, but i'm a typing newbie so maybe I could get a review on that aspect as well?

trio/_threads.py
```diff
@@ -240,37 +315,59 @@ def deliver_worker_fn_result(result: outcome.Outcome[RetT]) -> None:
...
+     Only applicable to threads spawned by `trio.to_thread.run_sync`. Poll to allow
+     ``cancellable=False`` threads to raise :exc:`trio.Cancelled` at a suitable
+     place, or to end abandoned ``cancellable=True`` threads sooner than they may
+     otherwise.
+
+    Raises:
+        Cancelled: If the corresponding call to `trio.to_thread.run_sync` has had a
+            delivery of cancellation attempted against it, regardless of the value of
+            ``cancellable`` supplied as an argument to it.
+        AttributeError: If this thread is not spawned from `trio.to_thread.run_sync`.
```
njsmith, line 359: Can you make this a `RuntimeError`?

njsmith (commented):
What does this do if the thread call gets "uncancelled"? Ie this sequence:
1. Some outer enclosing scope enters the cancelled state, so `run_sync` becomes cancelled
2. A more-inner enclosing scope gets mutated as `cscope.shield = True`, so `run_sync` is no longer cancelled
3. Then the user calls `check_cancelled`

trio/_threads.py
richardsheridan: yes!

richardsheridan:
> What does this do if the thread call gets "uncancelled"? Ie this sequence:
> 
> 1. Some outer enclosing scope enters the cancelled state, so `run_sync` becomes cancelled
> 2. A more-inner enclosing scope gets mutated as `cscope.shield = True`, so `run_sync` is no longer cancelled
> 3. Then the user calls `check_cancelled`

The straightforward implementation was to make this a level change in cancellation for the thread, so `check_cancelled` forever raises `Cancelled` IFF the cancellation delivery was attempted at least once. I believe the docs reflect this, though maybe that needs to be clarified.

I think this is also the "best" behavior because it is an approximation of what would happen if the thread were blocked in `trio.from_thread.run(afn)`. In that case, even in your scenario, a `Cancelled` would come out of `afn` and take an indeterminate amount of time to propagate back to the host task, meaning that it could be delivered after raising shields.

richardsheridan:
also if the user has selected `cancellable=True`, the previous behavior would be to wake up and abandon the thread. I wanted `check_cancelled` to be a tool for abandoned threads to end early and should not depend on the host task's cancel status.

TeamSpen210 (approved):
Had a look through the code in regards to the typing aspects, looks all good there aside from one question and a potential improvement.

trio/_threads.py
```diff
@@ -170,6 +236,9 @@ async def to_thread_run_sync(  # type: ignore[misc]
     # for the result – or None if this function was cancelled and we should
     # discard the result.
     task_register: list[trio.lowlevel.Task | None] = [trio.lowlevel.current_task()]
+    # Holds a reference to the raise_cancel function provided if a cancellation
+    # is attempted against this task - or None if no such delivery has happened.
+    cancel_register: list[RaiseCancelT | None] = [None]  # type: ignore[assignment]
```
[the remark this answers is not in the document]
richardsheridan: Again, I'm new at this and my local mypy just told me that I was a bad person for trying to assign this. On it's face, and in practical execution, it seems fine. I didn't try sending it to CI though, so maybe it could be a local machine issue. If you remove the ignore, does it work locally for you?

trio/_threads.py
```diff
@@ -240,37 +315,64 @@ def deliver_worker_fn_result(result: outcome.Outcome[RetT]) -> None:
...
+            RetT
+        ] = await trio.lowlevel.wait_task_rescheduled(abort)
+        if type(msg_from_thread) is ThreadDone:
+            return msg_from_thread.result.unwrap()  # type: ignore[no-any-return]
+        elif type(msg_from_thread) is Run:
+            await msg_from_thread.run()
+        elif type(msg_from_thread) is RunSync:
+            msg_from_thread.run_sync()
+        else:  # pragma: no cover, internal debugging guard
+            raise TypeError(
```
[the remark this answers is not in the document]
richardsheridan: I like this, but are we doing `assert_never` anywhere else in the library? otherwise I think we could hold off here and do it in multiple places as a separate PR.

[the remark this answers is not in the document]
TeamSpen210: We can’t be no, since we’d need to copy it in as I mentioned. We can’t directly use the `typing_extensions` back port since we’re avoiding a runtime dependency.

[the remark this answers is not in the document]
richardsheridan: > If you add `@typing.final` to all three classes (disallowing subclassing), we’d then be able to use `typing.assert_never()` here to have the checker prove it’s impossible. We can just copy the definition for earlier Python versions into utils:

Side note, i'm using `type(x) is Type` rather than `isinstance`, doesn't that also disallow subclasses?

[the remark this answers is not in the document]
TeamSpen210: Not quite, because if someone did pass a subclass of one of the types in, it’d fail all the if checks since it isn’t one of those specific types.

agronholm:
> What does this do if the thread call gets "uncancelled"? Ie this sequence:
> 
> 1. Some outer enclosing scope enters the cancelled state, so `run_sync` becomes cancelled
> 2. A more-inner enclosing scope gets mutated as `cscope.shield = True`, so `run_sync` is no longer cancelled
> 3. Then the user calls `check_cancelled`

Could you explain how the host task would get un-cancelled when it's still waiting on `run_sync()`? And if it gets cancelled, then this is pretty clear: the thread gets the cancelled flag set too, no matter what happens to the host task afterwards.

smurfix:
> What does this do if the thread call gets "uncancelled"?

I'd regard support for un-cancellation as strictly optional, doubly so regarding threads.

IMHO a cancel scope either shields, or it doesn't. Changing its mode after the fact smells like a race condition waiting to happen.

newsfragments/2392.feature.rst
```diff
@@ -0,0 +1 @@
+If called from a thread spawned by `trio.to_thread.run_sync`, `trio.from_thread.run` and `trio.from_thread.run_sync` now reuse the task and cancellation status of the host task and have `trio.from_thread.check_cancelled` to efficiently poll for cancellation.
```
oremanj, line 1: "now reuse the task and cancellation status of the host task; this means that context variables and cancel scopes naturally propagate 'through' threads spawned by Trio. You can also use `trio.from_thread.check_cancelled` to efficiently check for cancellation without reentering the Trio thread."

trio/_tests/test_threads.py
```diff
@@ -810,25 +809,28 @@ def test_from_thread_run_during_shutdown():
...
-    async def main():
-        save.append(agen())
+    async def main(use_system_task):
+        save.append(agen(_core.current_trio_token() if use_system_task else None))
         await save[-1].asend(None)
 
-    _core.run(main)
+    _core.run(main, True)  # System nursery will be closed and raise RunFinishedError
+    _core.run(main, False)  # host task will not be rescheduled
     assert record == ["ok"]
```
oremanj, line 828: Can you explain why this only produces a single `"ok"` rather than the two that I would expect from looking at it?

trio/_tests/test_threads.py
```diff
@@ -842,3 +844,166 @@ def __bool__(self):
...
+        tid = threading.get_ident()
+        return from_thread_run(to_thread_run_sync, threading.get_ident)
+
+    assert tid != await to_thread_run_sync(get_tid_then_reenter)
+
+
+async def test_from_thread_host_cancelled():
+    queue = stdlib_queue.Queue()
+
+    def sync_check():
```
oremanj, line 873: I had a bit of a hard time parsing this test; can you add a comment like "`run_sync` doesn't raise Cancelled because synchronous functions don't contain any checkpoints"?

trio/_tests/test_threads.py
```diff
@@ -842,3 +844,166 @@ def __bool__(self):
...
+        await to_thread_run_sync(sync_check)
+
+    assert not cancel_scope.cancelled_caught
+    assert not queue.get_nowait()
+
+    with _core.CancelScope() as cancel_scope:
+        await to_thread_run_sync(sync_check, cancellable=True)
+
+    assert cancel_scope.cancelled_caught
+    assert await to_thread_run_sync(partial(queue.get, timeout=1))
```
oremanj, line 892: Hm, this suggests that `from_thread.run_sync` *does* raise Cancelled if the thread was abandoned. Where is that coming from? It can't be from the cancel scope, because the cancel scope doesn't exist anymore (in the general case). If it's a general "this thread was abandoned" marker, then that's mildly backcompat-breaking and I think I disprefer it regardless - "sync functions have no checkpoints and therefore can't be the cause of a Cancelled" is a nice and easy rule to remember and works better if we don't have corner-case exceptions to it.

I'm not sure how _async_ `from_thread.run` from an abandoned thread should work. Options I see are:
* run it in a system task in a cancelled cancel scope
* run it in a system task that isn't cancelled, like we did before this PR
* raise Cancelled immediately

I would prefer one of the first two, since the third is unlike anything done elsewhere in Trio (Cancelled is raised at checkpoints, entry to an async function is not a checkpoint) and makes it hard to hand off resources safely (if you 'own' a resource and you pass it to a task that begins `[async] with resource:`, then that's normally safe but wouldn't be with the new semantics because Cancelled could be raised before the recipient enters the context manager).

Regardless of what we choose, we need to document it carefully, because it's a subtle consideration.

trio/_threads.py
```diff
@@ -27,13 +27,14 @@
...
 
 class _TokenLocal(threading.local):
     """Global due to Threading API, thread local storage for trio token."""
 
     token: TrioToken
+    cancel_register: list[RaiseCancelT | None]
+    task_register: list[trio.lowlevel.Task | None]
 
 
 TOKEN_LOCAL = _TokenLocal()
```
oremanj, line 40: it's not just the TrioToken anymore, maybe worth a new name? first that comes to mind is `ThreadParentData`

trio/_threads.py
```diff
@@ -70,6 +71,71 @@ class ThreadPlaceholder:
...
+    async def unprotected_afn(self) -> RetT:
+        coro = coroutine_or_error(self.afn, *self.args)
+        return await coro
+
+    async def run(self) -> None:
+        task = trio.lowlevel.current_task()
+        old_context = task.context
+        task.context = self.context.copy()
+        try:
+            await trio.lowlevel.cancel_shielded_checkpoint()
```
oremanj, line 99: Comment that this is to pick up the change in `context`. It's a shame that there's not a fast "these contexts definitely are equal, yes or no" operation to be able to bypass the extra checkpoint in the common case where the thread doesn't touch its context -- it would be easy to implement given the underlying representation of Context.

trio/_threads.py
```diff
@@ -70,6 +71,71 @@ class ThreadPlaceholder:
...
+
+    @disable_ki_protection
+    def unprotected_fn(self) -> RetT:
+        ret = self.fn(*self.args)
+
+        if inspect.iscoroutine(ret):
+            # Manually close coroutine to avoid RuntimeWarnings
+            ret.close()
+            raise TypeError(
+                "Trio expected a sync function, but {!r} appears to be "
```
oremanj, line 128: nit: I think our usual documentation style is to spell out "synchronous function" but abbreviate "async function" for parity with the syntax

trio/_threads.py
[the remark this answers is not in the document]
oremanj: Seems like a mypy bug to me; you can't assign `list[None]` to `list[T | None]` (because you could use the more general type to append a T, violating the type of the original) but you should be able to assign an actual literal `[None]` to the latter because it doesn't have any other references. And it worked three lines ago in an analogous situation. mypy will warn us if the type-ignore is no longer needed (we are using strict mode, right?) so fine by me to leave it for now.

trio/_threads.py
```diff
@@ -240,37 +315,64 @@ def deliver_worker_fn_result(result: outcome.Outcome[RetT]) -> None:
...
             return trio.lowlevel.Abort.FAILED
 
-    # wait_task_rescheduled return value cannot be typed
-    return await trio.lowlevel.wait_task_rescheduled(abort)  # type: ignore[no-any-return]
+    while True:
+        # wait_task_rescheduled return value cannot be typed
+        msg_from_thread: ThreadDone[RetT] | Run[object] | RunSync[
+            object
+        ] = await trio.lowlevel.wait_task_rescheduled(abort)
+        if type(msg_from_thread) is ThreadDone:
```
oremanj, line 333: Use isinstance(). It's more idiomatic _and_ faster:
```
% python3.8 -m timeit -s "A = type('A', (), {}); a = A()" "isinstance(a, A)"
5000000 loops, best of 5: 44 nsec per loop
% python3.8 -m timeit -s "A = type('A', (), {}); a = A()" "type(a) is A"
5000000 loops, best of 5: 63.4 nsec per loop
```
(the lead is more marginal on more recent Pythons, but still present in my testing)

trio/_threads.py
```diff
@@ -240,37 +315,64 @@ def deliver_worker_fn_result(result: outcome.Outcome[RetT]) -> None:
...
-    context: contextvars.Context,
-    trio_token: TrioToken | None = None,
-    # Outcome isn't typed, so Ret2T is used only in the return type.
-) -> Ret2T:  # type: ignore[type-var]
-    """Helper function for from_thread.run and from_thread.run_sync.
+def from_thread_check_cancelled() -> None:
+    """Raise trio.Cancelled if the associated Trio task entered a cancelled status.
+
+     Only applicable to threads spawned by `trio.to_thread.run_sync`. Poll to allow
+     ``cancellable=False`` threads to raise :exc:`trio.Cancelled` at a suitable
```
oremanj, line 351: man, we really need a better name for this `cancellable` thing now. Not something that needs to be changed in this PR, but maybe we should mention the history in the docs for `to_thread.run_sync`.

oremanj (commented):
Thank you, this generally looks excellent!

trio/_threads.py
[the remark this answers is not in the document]
jakkdl: > mypy will warn us if the type-ignore is no longer needed (we are using strict mode, right?)

`strict=True` is not literally set (though I realize now it probably should be to automatically enable any future checks added by mypy), but `warn_unused_ignores` is, so yeah it will give a warning.

trio/_tests/test_threads.py
richardsheridan: I think I could on board for "run in cancelled system task" (maybe catching and dropping any exceptions?) but documenting the nuances of "if you abandon your thread, your async functions are going to run in a weird different task" seems really hard. 

What do you think of re-using `RunFinishedError` or making a `TaskFinishedError` to raise immediately? This would at least not trample on the semantics of `Cancelled`, but it wouldn't help with the resource handoff issue (on the other hand, users must already handle `RunFinishedError` in the present case).

trio/_tests/test_threads.py
richardsheridan: Past me did not do a great job of documenting this, and possibly didn't understand what was going on at all.  Present me notices that `pytest.raises` is not finding the error it expected, but that is not propagating out to the test runner.

I'll change it so `RunFinishedError` appends one string and no exception appends a different one.

trio/_threads.py
richardsheridan: This is really old text, and protected by a test, but on the other hand I'm happy to update them both!

trio/_tests/test_threads.py
richardsheridan: I implemented the semantics I think you expected, and there were enough degrees of freedom in the test suite that it didn't need much adjustment. That makes me think that we should think of a way to (a) document the surprising abandon-to-system-task semantics and (b) anchor them in the test suite. I can't yet think of which behaviors really should be represented that way though.

oremanj: Yeah, on further consideration I agree my previous proposal is fairly wonky and we can probably do better. I think my preference would actually be to have `cancellable=True` threads use the old semantics: don't pass cancellation, don't run `from_thread.run` reentrant calls in the `to_thread.run_sync` task, just use a system task (that is not cancelled) like we did before this PR. Rationale:
* Making stuff fail that worked before isn't great for backcompat. If someone is using the explicit "abandon thread when I'm cancelled" flag, then they might specifically want the abandoned thread to continue operating; our attempt to interrupt it as soon as we regain control might be counterproductive.
* Since the `cancellable=True` thread explicitly might outlive the original task, any attempt it makes to call back into Trio is probably for something unrelated to the original task's context, so it's strange for that attempt to reflect the cancellation status of the original task.
* I imagine a conceptual distinction between `cancellable=False` threads being an "extension of the current task", vs `cancellable=True` being "fire-and-forget + retrieve the result if you're still around when it becomes available". It just feels wrong to propagate cancellation in the latter case.

oremanj:
> I'd be happy to hear suggestions for applying the context in from_thread.run that could avoid two extra checkpoints!

That's the only available way to change a task's context without creating a new task, and it's probably lower-overhead than the new task approach (but I didn't measure). The new task approach would be to push the reentrant call into a system task, which can have whatever context you want, and wrap it in a cancel scope that gets cancelled when the `to_thread.run_sync` call gets cancelled. I probably prefer the approach here where the reentrant functions run underneath the original task, but they're both workable.

> What does this do if the thread call gets "uncancelled"?

`from_thread.check_cancelled()`: The initial cancellation will call to_thread.run_sync's abort_fn, which will save the raise_cancel and use it the next time the thread checks for cancellation; the "uncancellation" will not be noticed. I think this is fine; it's well-defined, deterministic modulo the question of whether the thread checks for cancellation at all in between when cancellation is asserted and when the thread completes (which is a source of nondeterminism even without shielding), and the Cancelled exception will propagate fine through the shield (we have and test a similar edge case in the non-threads world).

`from_thread.run()`: The function that runs upon reentry is executed as a normal async call in the `to_thread.run_sync` task, so it does see the uncancellation. I'm realizing that this results in a discrepancy between `from_thread.run(trio.sleep, 0)` and `from_thread.check_cancelled()`, which is a bit unfortunate, and probably we should document it, but I don't think it's a problem in practice. `from_thread.check_cancelled()` can't observe that the 'uncancellation' has occurred at all unless it pokes into internals (the CancelStatus object). There's no way I know of to be woken up if 'uncancellation' occurs, so the main `to_thread.run_sync` task can't help it out. Maybe inspecting CancelStatus is fine/worthwhile? Doesn't quite seem worth it to me (it breaks our rules about how non-`_core` code is supposed to consume `_core` code), but it's a near thing. But it would be easy to add later if we decide to; I'm comfortable taking the simpler road for now.

There was a comment upthread that said inspecting CancelStatus wouldn't be thread-safe, but I don't think that's actually a problem: if the check is racing with the uncancellation then AFAICT it doesn't really matter whether the check sees the uncancellation or not. However my earlier snipept (which would `raise trio.Cancelled._create()` if the CancelStatus showed a cancelled status) was incorrect because it didn't handle the KeyboardInterrupt possibility. So we need the current `raise_cancel` logic regardless; any CancelStatus check for uncancellation would be in addition to that.

trio/_tests/test_threads.py
richardsheridan: I was going to suggest a new, overriding kwarg `abandon_on_cancel`. That tells you what happens and when, and my multiprocessing library could use `kill_on_cancel` instead. `cancellable` would then hang around indefinitely, deprecated but with unchanged semantics.

richardsheridan:
> `from_thread.check_cancelled()`: The initial cancellation will call to_thread.run_sync's abort_fn, which will save the raise_cancel and use it the next time the thread checks for cancellation; the "uncancellation" will not be noticed. I think this is fine; it's well-defined, deterministic [...]

I think there may be another nondeterministic edge case: if the cancellation arrives while the task is busy in `await message.run()` AND that work happens to be shielded, AND a shield gets raised uptree, then the main "have we ever been cancelled" check will not be triggered. I think a nursery with a task with the sole purpose of waiting for cancellation might be needed to get that exactly right.

As long as we need a nursery to get those semantics right, we might as well also consider an alternative implementation where the each `from_thread` call spawns a new task in that nursery. After all, it's not really the fact that we are reusing the task which is so useful, but rather that it exists in the right cancellation, contextvar, and treevar contexts. Child tasks would have all these properties, although they might be marginally less efficient.

trio/_tests/test_threads.py
oremanj: I'm equally happy with `abandon_on_cancel` as a name.

oremanj:
> I think there may be another nondeterministic edge case: if the cancellation arrives while the task is busy in await message.run() AND that work happens to be shielded, AND a shield gets raised uptree, then the main "have we ever been cancelled" check will not be triggered. I think a nursery with a task with the sole purpose of waiting for cancellation might be needed to get that exactly right.

I think it depends what you mean by "exactly right". :-) The analogous situation without the thread would not raise a cancellation in this case, so I don't think the thread needs to either. In any event I don't think it would be worth the trouble and overhead of a whole separate task to wait for cancellation.

In general the way Trio behaves is that you need to execute a checkpoint _while_ you are effectively cancelled (have a cancelled cancel scope closer to you than the closest shielded cancel scope) in order to raise Cancelled. If you perform a cancellation and then a shielding in the way that causes the effectively-cancelled status to toggle back and forth, but no checkpoints observe the "cancelled" state, then for your purposes it didn't happen. I think this analogizes nicely to threads if we take the "checkpoints" to be calls to `check_cancelled()` + (unshielded) checkpoints in reentrant calls to `from_thread.run()` functions. We can only get that exact behavior if we peek inside the internal CancelStatus object on `check_cancelled`; otherwise we will get a more conservative / "cancel-happy" behavior where `check_cancelled` will raise Cancelled if the `to_thread.run_sync()` call has _ever_ been effectively cancelled, even if it's not anymore. Which is also reasonable! We just need to document the subtleties here.

docs/source/reference-core.rst
```diff
@@ -1823,8 +1823,22 @@ to spawn a child thread, and then use a :ref:`memory channel
 
 .. literalinclude:: reference-core/from-thread-example.py
 
-You can also perform a non-blocking check for cancellation from threads spawned
-by `trio.to_thread.run_sync`.
+.. note::
+
+   The ``from_thread.run*`` functions reuse the host task that called
+   :func:`trio.to_thread.run_sync` to run your provided function in the typical case,
+   namely when ``cancellable=False`` so Trio can be sure that the task will always be
```
oremanj, line 1830:
```suggestion
   :func:`trio.to_thread.run_sync` to run your provided function, as long as you're
   using the default ``cancellable=False`` so Trio can be sure that the task will remain
```

trio/_threads.py
```diff
@@ -357,9 +391,16 @@ def from_thread_check_cancelled() -> None:
             delivery of cancellation attempted against it, regardless of the value of
             ``cancellable`` supplied as an argument to it.
         RuntimeError: If this thread is not spawned from `trio.to_thread.run_sync`.
+
+    .. note::
+
+       The check for cancellation attempts of ``cancellable=False`` threads is
+       interrupted while executing ``from_thread.run*`` functions, which can lead to
+       edge cases where this function may raise or not depending on the timing of
+       :class:`~trio.CancelScope` shields being raised or lowered in the Trio threads.
```
oremanj, line 400:
```suggestion
       To be precise, :func:`~trio.from_thread.check_cancelled` checks whether the task
       running :func:`trio.to_thread.run_sync` has ever been cancelled since the last
       time it was running a :func:`trio.from_thread.run` or :func:`trio.from_thread.run_sync`
       function. It may raise `trio.Cancelled` even if a cancellation occurred that was
       later hidden by a modification to `trio.CancelScope.shield` between the cancelled
       `~trio.CancelScope` and :func:`trio.to_thread.run_sync`. This differs from the
       behavior of normal Trio checkpoints, which raise `~trio.Cancelled` only if the
       cancellation is still active when the checkpoint executes. The distinction here is
       *exceedingly* unlikely to be relevant to your application, but we mention it
       for completeness.
```
In general it's better to say what the edge cases are than to just say there can be edge cases.

trio/_threads.py
```diff
@@ -457,17 +456,15 @@ def from_thread_run(
         RunFinishedError: if the corresponding call to :func:`trio.run` has
             already completed, or if the run has started its final cleanup phase
             and can no longer spawn new system tasks.
-        Cancelled: if the corresponding `trio.to_thread.run_sync` task is
-            cancellable and exits before this function is called, or
-            if the task enters cancelled status or call to :func:`trio.run` completes
-            while ``afn(*args)`` is running, then ``afn`` is likely to raise
+        Cancelled: if the task enters cancelled status or call to :func:`trio.run`
```
oremanj, line 459:
```suggestion
        Cancelled: If the original call to :func:`trio.to_thread.run_sync` is cancelled
            (if *trio_token* is None) or the call to :func:`trio.run` completes
            (if *trio_token* is not None) while ``afn(*args)`` is running,
            then *afn* is likely to raise
```

oremanj (approved):
Looking good, thank you for seeing this through! I think the main remaining thing is documentation wording, for which I've left some suggestions. It looks like there are also a few new lines that aren't covered by the tests. You're welcome to merge once you've resolved the remaining comments.

trio/_threads.py
```diff
@@ -70,6 +73,103 @@ class ThreadPlaceholder:
...
+            assert task is not None, "guaranteed by abandon_on_cancel semantics"
+            trio.lowlevel.reschedule(task, outcome.Value(self))
+
+        token.run_sync_soon(in_trio_thread)
+
+    def run_in_system_nursery(self, token: TrioToken) -> None:
+        def in_trio_thread() -> None:
+            try:
+                trio.lowlevel.spawn_system_task(
+                    self.run, name=self.afn, context=self.context
```
richardsheridan, line 122:
```suggestion
                    self.run_system, name=self.afn, context=self.context
```
This is the likely source of the missed coverage.

PR #2407, opened 2022-08-26
Add `__main__` module with `await`-compatible console

wbadart:
I recently learned that `python -m asyncio` drops you into an
interpreter session that knows how to deal with `await` expressions. For
example,

```console
$ python -m asyncio
asyncio REPL 3.10.6 [...] on darwin
Use "await" directly instead of "asyncio.run()".
>>> import asyncio
>>> await asyncio.sleep(1); print("hi")  # prints after one second
hi
```

This quickly became an important part of my async debugging toolbox.

I've also been learning Trio recently, but I've been missing my async
console.

This PR takes the script from [`asyncio.__main__`][1] and adapts it to
use a Trio run loop. With this patch applied, users can run `python -m
trio` to drop into an interactive interpreter session that lets them
directly `await` Trio-based async expressions.

```console
$ python -m trio
Trio 0.21.0+dev, Python 3.10.6 [...] on darwin
Use "await" directly instead of "trio.run()".
>>> import trio
>>> await trio.sleep(1); print("hi")  # prints after one second
hi
```

[1]: https://github.com/python/cpython/blob/master/Lib/asyncio/__main__.py

I'm very new to Trio, so I'm sure my approach is suboptimal. In
particular, I'd really appreciate feedback from the team/ community on:

- not using `threading` with a brand new `trio.run` to evaluate
  awaitable expressions. I found that if I simply used
  `self.nursery.start_soon` in the same thread as `runcode`, `await`
  expressions wouldn't get the chance to be evaluated until the session
  was shutting down (might have been blocked by reading keyboard input,
  if I had to guess). Could there be a solution in `trio`'s threading
  tools?
- what documentation updates should accompany this PR
- what sort of tests I should add (since this change implements a brand
  new code path that won't be touched by any existing or future usage of
  `trio` itself, I wasn't sure if it would even be appropriate to _try_
  and test it)
- anything else to align this addition to the Trio way!

Many thanks for your consideration.


efiring:
Given that #2972 has been merged, this PR should be closed, shouldn't it? (With
thanks to @wbadart for initiating it, and to @CoolCat467 for pushing it forward.)

A5rocks:
Yeah it should have been. Thanks for the contribution nonetheless!

wbadart:
Glad to see this feature has been added, and I'm happy to have played a (small) role in the solution!

PR #2410, opened 2022-08-31
Add a method to get the traceback of a task

VincentVanlaer:
This PR adds a method to trio.Task to get a stack trace of a task. It is based on the example that was in the docs, but I added support for async generators. This was non-trivial as cpython doesn't expose the underlying generator object. I originally used a standalone function like this in a modified version of pytest-trio to support timeouts (I might upstream that later once I clean it up a bit). However, I felt that such a function would belong better in the main trio repo, where it can easily be used by others.


Fuyukai:
I'm in favour of this, not having to deal with the async generator introspection black hole is essentially useful enough to put it in trio itself, and it was clearly useful to put in the docs. I'll merge if/when the tests pass

PR #2415, opened 2022-09-05
Revert "docs: fix bug in NotVeryGoodLock example"

pquentin:
Reverts python-trio/trio#2414 per https://gitter.im/python-trio/general?at=631490ca11a6a83d049a8631

also, clarify the code with a comment and by minimizing the scope of acquire's "retry"

### DIFF
--- a/docs/source/reference-lowlevel.rst
+++ b/docs/source/reference-lowlevel.rst
@@ -464,10 +464,10 @@ this does serve to illustrate the basic structure of the
            while self._held:
                # Someone else has the lock, so we have to wait.
                task = trio.lowlevel.current_task()
+               self._blocked_tasks.append(task)
                def abort_fn(_):
                    self._blocked_tasks.remove(task)
                    return trio.lowlevel.Abort.SUCCEEDED
-               self._blocked_tasks.append(task)
                await trio.lowlevel.wait_task_rescheduled(abort_fn)
                # At this point the lock was released -- but someone else
                # might have swooped in and taken it again before we
--- a/docs/source/reference-lowlevel.rst
+++ b/docs/source/reference-lowlevel.rst
@@ -460,7 +460,7 @@ this does serve to illustrate the basic structure of the
            self._held = False
 
        async def acquire(self):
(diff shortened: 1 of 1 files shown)


docs/source/reference-lowlevel.rst
```diff
@@ -460,7 +460,7 @@ this does serve to illustrate the basic structure of the
            self._held = False
 
        async def acquire(self):
-           if self._held:
+           while self._held:
                task = trio.lowlevel.current_task()
                self._blocked_tasks.append(task)
                def abort_fn(_):
```
belm0, line 466: I think there's a different bug in the code then.

If this loop can run more than once, and without abort being called, then the task will be appended to `_blocked_tasks` multiple times.

This avoids the bug and makes it clear what things aren't changing on each iteration:
```python
async def acquire(self):
    if self._held:
        task = trio.lowlevel.current_task()
        self._blocked_tasks.append(task)
        def abort_fn(_):
            self._blocked_tasks.remove(task)
            return trio.lowlevel.Abort.SUCCEEDED
        # may need multiple attempts since reschedule() is not instant
        while self._held:
            await trio.lowlevel.wait_task_rescheduled(abort_fn)
    self._held = True
```

belm0:
Updated PR and description, please take a look.

> also, clarify the code with a comment and by minimizing the scope of acquire's "retry"

njsmith:
For me, I think it's easier to read with the whole retry logic together in one block, instead of broken up into the part that's idempotent and the part that needs to be retried. I rewrote a bit to add more detailed comments at the critical points though; see what y'all think.

docs/source/reference-lowlevel.rst
```diff
@@ -460,13 +460,22 @@ this does serve to illustrate the basic structure of the
            self._held = False
 
        async def acquire(self):
-           if self._held:
+           # We might have to try several times to acquire the lock.
```
belm0, line 463: I think this comment can be removed, retry is explained below

belm0 (approved):
thank you

PR #2431, opened 2022-09-28
clarify RunVar test

belm0:


trio/_core/tests/test_local.py
```diff
@@ -92,7 +92,7 @@ async def task2(tok):
 
             n.start_soon(task2, token)
             await _core.wait_all_tasks_blocked()
-            assert t1.get() == "cod"
+            assert t1.get() == "swordfish"
```
belm0, line 95: it wasn't clear that this expectation corresponds to the `set()` in task2, rather than the `set()` at line 86

pquentin (changes requested):
This makes sense, thanks!

So I'm going to be very pedantic, but I just learned at PyCon UK 2022 that cod and plaice are the two traditional choices for fish and chips, when swordfish is not I think? Maybe use haddock: https://en.wikipedia.org/wiki/Fish_and_chips#Choice_of_fish

pquentin (approved):
Thank you!

PR #2443, opened 2022-10-13
Introduce a gate/check GHA job

webknjaz:
This adds a GHA job that reliably determines if all the required dependencies have succeeded or not.

It also allows to reduce the list of required branch protection CI statuses to just one — `check`. This reduces the maintenance burden by a lot and have been battle-tested across a small bunch of projects in its action form and in-house implementations of other people.

It is now in use in aiohttp (and other aio-libs projects), CherryPy, attrs, coveragepy, some of the Ansible repositories, pip-tools, pydantic, spaceship-prompt, Towncrier, all of the jaraco's projects (like `setuptools`, `importlib_metadata`), some PyCQA, PyCA, PyPA and pytest projects, a few AWS Labs projects.

The story behind this is explained in more detail at https://github.com/marketplace/actions/alls-green#why.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,16 @@ jobs:
           - python: '3.8'  # <- not actually used
             pypy_nightly_branch: 'py3.8'
             extra_name: ', pypy 3.8 nightly'
+    continue-on-error: >-
+      ${{
+        (
+          matrix.check_formatting == '1'
+          || matrix.pypy_nightly_branch == 'py3.7'
+          || endsWith(matrix.python, '-dev')
+        )
+        && true
+        || false
+      }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -138,3 +148,21 @@ jobs:
         env:
           # Should match 'name:' up above
           JOB_NAME: 'macOS (${{ matrix.python }})'
+
+  # https://github.com/marketplace/actions/alls-green#why
+  check:  # This job does nothing and is only used for the branch protection
+
+    if: always()
+
+    needs:
+      - Windows
+      - Ubuntu
+      - macOS
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}


.github/workflows/ci.yml
```diff
@@ -88,6 +88,16 @@ jobs:
           - python: '3.8'  # <- not actually used
             pypy_nightly_branch: 'py3.8'
             extra_name: ', pypy 3.8 nightly'
+    continue-on-error: >-
```
webknjaz, line 91: ^ this is needed to match the currently deselected jobs in branch protection.

.github/workflows/ci.yml
```diff
@@ -138,3 +148,21 @@ jobs:
         env:
           # Should match 'name:' up above
           JOB_NAME: 'macOS (${{ matrix.python }})'
+
+  # https://github.com/marketplace/actions/alls-green#why
+  check:  # This job does nothing and is only used for the branch protection
```
webknjaz, line 153: And this ^ needs to go to the branch protection

.github/workflows/ci.yml
altendky: How about adding a `not_required` attribute to the matrix?  I guess that would require dev Python versions get an `include:` entry.  Seems to me like a worthwhile trade for simplicity and explicitness here.

.github/workflows/ci.yml
```diff
@@ -138,3 +148,21 @@ jobs:
...
+    needs:
+      - Windows
+      - Ubuntu
+      - macOS
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
```
altendky, line 166: People generally use mutable tags so there's no security around this, in case you care.  Could pin to the commit hash instead after a review if anyone is concerned about it.

altendky (commented):
I'll leave it to others to actually review this in terms of whether they want this behavior and via the action or not.  But, I do like this form and have used it for quite awhile.  It basically shifts the actions jobs 'required checks' project configuration into the source so it can be kept in sync as new Python versions are added and old removed.  No transitional period nor extra admin actions needed.

It has become a bit more complicated "recently" due to changes in GitHub actions and how they handle dependent jobs.  That change was accounted for in PRs such as https://github.com/twisted/towncrier/pull/392.  I also looked at creating an action and a reusable workflow and while they are useful, the way all this data works does force you to define the `if:` here at the usage site as well as passing that along through a `toJSON()` call.  So it's a bit less tidy than I'd like, but I also don't have a recommendation for a cleaner form than provided here in terms of the job itself.

webknjaz:
> nor extra admin actions needed.

One admin action is needed — updating the branch protection for the last time. But it is non-blocking and may happen later. The current setup will keep being usable until that's done.

.github/workflows/ci.yml
webknjaz: I guess. I just didn't want to include too much changes in the PR.
FWIW GitHub's own docs feature the use of an `experimental` flag for this purpose.

altendky:
Sorry, yes, I meant after the final transition.  `:]`  Good to clarify though.

.github/workflows/ci.yml
webknjaz: Here's a convenient `I care` button if anybody needs this:
```suggestion
        uses: re-actors/alls-green@13b4244b312e8a314951e03958a2f91519a6a3c9
```

webknjaz:
@pquentin your approval is needed here

pquentin:
I don't have any opinions here, sorry. :) Any other member can approve, not sure why @altendky felt he could not.

webknjaz:
@pquentin technically, I have a green button too. Branch protection doesn't require approvals. I just wanted visibility/feedback from more regular contributors. Plus, somebody with admin access (@njsmith?) would need to eventually update the branch protection to make use of this.

altendky (approved):
I do feel I have the technical knowledge around this topic to be ok with this, but at least recently I have seen opinions around various aspects of this.  I am not involved enough at this point to feel that my opinion would override any other actually active maintainers.  I also don't have the authority on GitHub to make the related required check changes and figured whoever would do that part may as well be aparty to the approval here.

Anyways, here's my approval for what it's worth and with all the caveats laid out in the prior comments.

pquentin:
Thanks, I updated the required checks as follows:


image

webknjaz:
@pquentin could you also tick the checkbox on the main page of the repo settings for always showing the "update branch" button? Old PRs have CI runs reported without the new required check. This means that they are unmergable until there's a new CI run. This usually requires a branch update to trigger the check. With that button always shown, it's easy to rebase it from the GH web UI.

pquentin:
Thanks, done.

PR #2448, opened 2022-10-18
Moved the metadata into machine-readable `setup.cfg`.

KOLANICH:
Version is now fetched from git tags automatically.


pquentin:
Closing in favor of #2449  because if we're going to do that it will be with pyproject.toml.

PR #2449, opened 2022-10-18
Moved the metadata from setup.cfg into PEP 621-compliant `pyproject.toml`

KOLANICH:


KOLANICH:
This PR is built on top of an another PR #2448. It is expected that that PR is landed first, and than this one onto that one. The first commit moves the metadata into `setup.cfg`, that is available for ancient versions of `setuptools`, but is still declarative and machine-parseable. The second commit moves the metadata into `pyproject.toml`, it has a drawback of a more narrow range of `setuptools` versions supported. So if someone still has to use the ancient version, he can `git revert` the second commit.

pyproject.toml
```diff
@@ -1,3 +1,114 @@
...
+    "sniffio",
+    # cffi 1.12 adds from_buffer(require_writable=True) and ffi.release()
+    # cffi 1.14 fixes memory leak inside ffi.getwinerror()
+    # cffi is required on Windows, except on PyPy where it is built-in
+    "cffi>=1.14; os_name == 'nt' and implementation_name != 'pypy'",
+    "exceptiongroup >= 1.0.0rc9; python_version < '3.11'",
+]
+dynamic = ["version"]
+
+[project.readme]
```
pquentin, line 48: I realize you just copied what setup.py did but if we're going to switch to PEP 621 we should use the opportunity to use README.md directly. But this requires making sure it renders well both in GitHub and PyPI.

trio/_version.py
```diff
@@ -1,3 +0,0 @@
-# This file is imported from __init__.py and exec'd from setup.py
-
-__version__ = "0.22.0+dev"
```
pquentin, line 3: We want to keep this here because we need to inspect `trio.__version__` at runtime. Hatchling supports this.

smurfix: Can't you use `pkg_resources.get_distribution("trio").version`for that?

pquentin: I've had bad experiences in the past with the overhead of pkg_resources but maybe I didn't use it correctly. It's also semi deprecated in favor of importlib.

Most importantly, this seems like a complex solution for a problem we don't currently have.

KOLANICH:
>We want to keep this here because we need to inspect trio.__version__ at runtime. Hatchling supports this.

`_version.py` file is generated automatically by `setuptools_scm`

>Can't you use pkg_resources.get_distribution("trio").versionfor that?

We can, but it is damn slow.

>But this requires making sure it renders well both in GitHub and PyPI.

Checked using https://github.com/pypa/readme_renderer, LGTM.

pquentin (changes requested):
There are still various references to setup.py left.

Also we want to have a setup.py file to give a clear error because someone will inevitably use it, including GitHub itself: https://github.com/urllib3/urllib3/blob/main/setup.py

And I'm still not convinced setuptools_scm is better than hatchling :shrug: but it's not a huge deal. https://github.com/python-trio/trio/blob/master/docs/source/releasing.rst needs to change though if we're going to use that.

ofek:
Hatchling maintainer here! I do agree with @pquentin that Hatchling would make things simpler here (and generally speaking too, which is why it's now the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/)).

For example, the following:

```toml
[tool.setuptools]
include-package-data = true

[tool.setuptools.packages]
find = {namespaces = false}
```

is unnecessary with Hatchling because:

1. There is no concept of package data; files are just files, which makes [selection](https://hatch.pypa.io/latest/config/build/#file-selection) easier to reason about.
2. The `wheel` target [defaults](https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection) would do the appropriate thing here.

Additionally, Hatchling:

- Supports static analysis tools for editable installations by default whereas the new setuptools [does not](https://github.com/microsoft/pylance-release/issues/3407) i.e. it requires enabling an option for IDEs to work.
- Uses [Git-style glob patterns](https://git-scm.com/docs/gitignore#_pattern_format) rather than stdlib globs to better match user expectations.
- Allows easy and explicit control of what gets shipped in the `sdist` [target](https://hatch.pypa.io/latest/config/build/) whereas with setuptools it's an objectively confusing mix of conditional wheel options and a [`MANIFEST.in`](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html) file.
- Hatchling produces a less verbose license section on PyPI because until they support [PEP 639](https://peps.python.org/pep-0639/) the rendering is not what one may think. It's actually `"License: {license_classifier.split(' :: ')[-1]} ({license['text']})"` so what you have currently would become `License: Apache Software License, MIT License (MIT OR Apache-2.0)`.

KOLANICH:
`hatch_vcs` has no `write_to`.

ofek:
- https://github.com/ofek/hatch-vcs#build-hook
- https://github.com/psf/black/blob/f22273a72b3f1c15085f2d4a43e8d785bf48c822/pyproject.toml#L105-L106

KOLANICH:
thanks for the info, probably the docs about `write_to` should have the build hook being mentioned.

ofek:
`write_to` is a `setuptools-scm` option

KOLANICH:
I mean that `hatch-vcs`' ReadMe says

>The `write_to` and `write_to_template` parameters are ignored.

so I propose to write 

>Use `build.hooks.vcs.{version-file,template}` instead.

right after that text.

pquentin:
I'd rather continue writing to trio/_version.py manually.

KOLANICH:
The setuptools-specific code is for the ones who prefer using `setuptools` over `hatchling`. It is activated only after patching `pyproject.toml`.

ofek:
Oh [found it](https://github.com/tartley/colorama/issues/349#issuecomment-1211583184), I didn't realize it was you 😄

ci.sh
```diff
@@ -66,11 +66,11 @@ fi
 
 python -c "import sys, struct, ssl; print('#' * 70); print('python:', sys.version); print('version_info:', sys.version_info); print('bits:', struct.calcsize('P') * 8); print('openssl:', ssl.OPENSSL_VERSION, ssl.OPENSSL_VERSION_INFO); print('#' * 70)"
 
-python -m pip install -U pip setuptools wheel
+python -m pip install -U pip hatchling build
```
webknjaz, line 69: Backends are never needed in outer envs — the frontends automatically provision them in the build venvs.

docs/source/releasing.rst
```diff
@@ -44,7 +48,7 @@ Things to do for releasing:
 * push to PyPI::
 
     git clean -xdf   # maybe run 'git clean -xdn' first to see what it will delete
-    python3 setup.py sdist bdist_wheel
+    python3 -m build -ws
```
webknjaz, line 51: Do not pass both `--sdist` and `--wheel`, because this causes wheels building from the Git checkout, as opposed to building wheels from sdist (as `pip wheel` does).
```suggestion
    python3 -m build
```

ci.sh
```diff
@@ -66,11 +66,11 @@ fi
...
 python -m pip --version
 
-python setup.py sdist --formats=zip
-python -m pip install dist/*.zip
+python -m build -nsx .
```
webknjaz, line 72: It's highly discouraged to use short options with unknown meanings in shell scripts. The typical convention is to use long names so that the readers could know what's happening without having to google or misinterpret things. Also, no need to pass the current dir as it's the default. And two of those options are probably harmful in this case too.

docs-requirements.txt
```diff
@@ -59,6 +59,8 @@ pytz==2022.4
     # via babel
 requests==2.28.1
     # via sphinx
+setuptools-scm==7.0.5
```
webknjaz, line 62: This doesn't seem needed/related. Also, it's a bad idea to change behavior together with refactoring. If this plugin is needed, it should be integrated separately.

ci.sh
graingert:
```suggestion
python -m pip install -U pip build
```

docs/source/.conf.py.kate-swp
graingert, line 1: this seems like an accidental commit, you should probably add `.kate-swp` to your global git ignore

graingert (changes requested):

graingert:
closing in favour of https://github.com/python-trio/trio/pull/2860 which was merged

PR #2470, opened 2022-11-08
Add name parameter to to_thread.run_sync [SQUASH]

### LINKED ISSUE #1148: Optionally allow to set the thread's name where possible
This request is kind of specific to Celery's needs but I'm sure other's may agree this is worthwhile doing.

On platforms which support this behaviour (Linux, MacOS, *BSD, New versions of Windows) it is possible to set the thread's name (or description on Windows).

It requires some C/C++ or an optional dependency but shouldn't be that hard to implement.

jakkdl:
Fixes #1148

This doesn't work on windows, requiring an entirely separate mechanism (https://stackoverflow.com/a/16199587). It should be possible to get working on Mac (https://stackoverflow.com/a/7989973), but not having a mac machine I need the CI to test that for me. BSD variants also might need slightly different code, but I won't bother with that not having a way to test it. I haven't added a decorator on the test, so will probably see failed tests on other OS's for now.

The default in #1148 was suggested to be "Thread for {trio.lowlevel.current_task().name" by @njsmith - but pthread limits thread names to 15 characters so that's kind of a bad default. Wasn't sure what else to have though so that's the current default, but we probably want something else.

I currently set the thread name from within the thread, but it's possible to set it from the parent thread (although maybe not on mac) but the first way I found out of getting a thread id was with `threading.get_ident()` so I did it in the child.

Options for setting the name was `pthread`, `sys/prctl.h` and `libcap.so`. The first should be the one most commonly found on systems, being installed with `glibc` (on my Arch system), https://stackoverflow.com/a/51668422 says `prctl` depends on `libcap-dev` (though it's also bundled with `glibc` on mine), and `libcap.so` is for me dependent on `libcap` which itself depends on `glibc`.

Actually seeing the thread name in `ps` turned out to be flaky with everybody on SO saying different things, only `ps -eT` worked for me.

Making it a draft until it's been tested on other systems.

@Zac-HD

### DIFF
--- a/trio/_core/_thread_cache.py
+++ b/trio/_core/_thread_cache.py
@@ -16,6 +16,12 @@ def get_os_thread_name_func() -> Optional[Callable[[Optional[int], str], None]]:
         if ident is not None:
             setname(ident, bytes(name[:15], "ascii", "replace"))
 
+    def darwin_namefunc(
+        setname: Callable[[bytes], int], ident: Optional[int], name: str
+    ):
+        if ident is not None:
+            setname(bytes(name[:15], "ascii", "replace"))
+
     libpthread_path = ctypes.util.find_library("pthread")
     if not libpthread_path:
         return None
@@ -26,9 +32,14 @@ def get_os_thread_name_func() -> Optional[Callable[[Optional[int], str], None]]:
         return None
 
     # specify function prototype
-    pthread_setname_np.argtypes = [ctypes.c_void_p, ctypes.c_char_p]
     pthread_setname_np.restype = ctypes.c_int
 
+    # on mac OSX pthread_setname_np does not take a thread id
+    if sys.platform == "darwin":
+        pthread_setname_np.argtypes = [ctypes.c_char_p]
+        return partial(darwin_namefunc, pthread_setname_np)
+
+    pthread_setname_np.argtypes = [ctypes.c_void_p, ctypes.c_char_p]
     return partial(namefunc, pthread_setname_np)
 
 
--- a/trio/_core/_thread_cache.py
+++ b/trio/_core/_thread_cache.py
@@ -2,7 +2,46 @@ import sys
 import traceback
 from threading import Thread, Lock
 import outcome
+import sys
+import ctypes
+import ctypes.util
 from itertools import count
+
+from typing import Callable, Optional, Tuple
+from functools import partial
+
+
+def get_os_thread_name_func() -> Optional[Callable[[Optional[int], str], None]]:
+    def namefunc(setname: Callable[[int, bytes], int], ident: Optional[int], name: str):
+        if ident is not None:
+            setname(ident, bytes(name[:15], "ascii", "replace"))
+
+    libpthread_path = ctypes.util.find_library("pthread")
+    if not libpthread_path:
+        return None
+    libpthread = ctypes.CDLL(libpthread_path)
+
+    pthread_setname_np = getattr(libpthread, "pthread_setname_np", None)
+    if pthread_setname_np is None:
+        return None
+
+    # specify function prototype
+    pthread_setname_np.argtypes = [ctypes.c_void_p, ctypes.c_char_p]
+    pthread_setname_np.restype = ctypes.c_int
+
+    return partial(namefunc, pthread_setname_np)
+    ## set the name
+    # try:
+    #    bname = name.encode("ascii", "replace")
+    #    thread = threading.current_thread()
+    #    # if thread is not None:
+    #    thread.name = name
+    #    pthread_setname_np(thread.ident, bname[:15])
+    # except Exception:
+    #    return nofunc
+
+
+set_os_thread_name = get_os_thread_name_func()
 
 # The "thread cache" is a simple unbounded thread pool, i.e., it automatically
 # spawns as many threads as needed to handle all the requests its given. Its
@@ -44,7 +83,7 @@ name_counter = count()
 
 class WorkerThread:
     def __init__(self, thread_cache):
-        self._job = None
+        self._job: Optional[Tuple[Callable, Callable, str]] = None
         self._thread_cache = thread_cache
         # This Lock is used in an unconventional way.
         #
@@ -54,16 +93,35 @@ class WorkerThread:
         # Initially we have no job, so it starts out in locked state.
         self._worker_lock = Lock()
         self._worker_lock.acquire()
-        thread = Thread(target=self._work, daemon=True)
-        thread.name = f"Trio worker thread {next(name_counter)}"
-        thread.start()
+        self._default_name = f"Trio thread {next(name_counter)}"
+
+        self._thread = Thread(target=self._work, daemon=True)
+        self._thread.name = self._default_name
+
+        if set_os_thread_name:
+            set_os_thread_name(self._thread.ident, self._default_name)
+        self._thread.start()
 
     def _handle_job(self):
         # Handle job in a separate method to ensure user-created
         # objects are cleaned up in a consistent manner.
-        fn, deliver = self._job
+        assert self._job is not None
+        fn, deliver, name = self._job
         self._job = None
+
+        # set name
+        if name is not None:
+            self._thread.name = name
+            if set_os_thread_name:
+                set_os_thread_name(self._thread.ident, name)
         result = outcome.capture(fn)
+
+        # reset name if it was changed
+        if name is not None:
+            self._thread.name = self._default_name
+            if set_os_thread_name:
+                set_os_thread_name(self._thread.ident, self._default_name)
+
         # Tell the cache that we're available to be assigned a new
         # job. We do this *before* calling 'deliver', so that if
         # 'deliver' triggers a new job, it can be assigned to us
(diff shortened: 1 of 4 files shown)


jakkdl:
Oh and if the thread name should be possible to set (or get) in any other functions or exposed in any way that shouldn't be too hard to implement.

trio/_threads.py
```diff
@@ -59,7 +59,9 @@ class ThreadPlaceholder:
 
 
 @enable_ki_protection
-async def to_thread_run_sync(sync_fn, *args, cancellable=False, limiter=None):
+async def to_thread_run_sync(
+    sync_fn, *args, name: str = "", cancellable=False, limiter=None
```
Zac-HD, line 63: I prefer `thread_name=None` here, to make it clear what the name is for, and that we'll construct a default if not given a name.

trio/_threads.py
```diff
@@ -215,6 +227,34 @@ def abort(_):
     return await trio.lowlevel.wait_task_rescheduled(abort)
 
 
+# inspired by code from Kovid Goyal https://bugs.python.org/issue15500#msg230736
+def _set_thread_name(name: str):
```
Zac-HD, line 231: This will be a little fiddly to plumb through, but we can do better: because Trio ultimately uses `threading.Thread` under the hood, we can assign to the `.name` attribute which should work on all platforms.  Following `start_thread_soon` through will get you to `trio/_core/_thread_cache.py`.

Additional complication: we actually have a cached pool of threads, so we'll want to set the name when we starting working, and reset to a generic e.g. `trio-idlethread` when the task is over.

jakkdl: could avoid having to plumb through several levels of call by using `threading.current_thread()` instead of `.get_ident()`.

That doesn't sound to bad, just unset the name after `sync_fn(*args)` is called in `worker_fn`? Though I guess that might fit better in `_core/_thread_cache` and is an argument for doing proper plumbing and placing all the logic in there.

jakkdl:
Pushed what a solution might look like, but I'm quite unhappy with it - ugly, cumbersome and probably slows down thread operations a bit even if you're not using names (since it will try to find the pthread library anew in each new thread). And given that `_thread_cache` talks about memory cache behavior I presume the performance here is somewhat important. Working on moving it into `_thread_cache` and speeding up the default case of not naming your threads.

Though there's a question there of if you should bother re-naming idle and active threads (probably not?) as well as how worth it is keeping the name counter in the name (means the thread needs to remember the name/number when renaming so it can restore it)

trio/_threads.py
Zac-HD: Yeah, I'd definitely put it in https://github.com/python-trio/trio/blob/ce4c99e696a516db40b55fb076deca2e586626cf/trio/_core/_thread_cache.py#L59-L70

Then we can just do `self.thread.name = thread_name` before `result = outcome.capture(fn)` and reset afterwards.  It also doesn't look like this interface shares the 15-character limit?

jakkdl: yeah the 15-char limit is just in pthreads / the OS, thread.name is only used internally in python so that's a normal unicode string.

jakkdl:
Okay this is starting to look like a decent solution. Question remains on how to handle naming idle threads, and default names in general. But I think otherwise it's in a decent spot.

Would it be possible to give me access as if I'm not a first-time contributor so I can iterate on getting it to work / not crash horribly on other OS's? Gonna take ages to finish this if I have to wait for approval on each workflow run.

jakkdl:
Okay this should be pretty decent.

`ctypes.util.find_library("pthread")` fails on the CI mac machines, which I didn't expect, so I'm going to leave getting it to work on mac to somebody else.

Zac-HD (commented):
With a few comments below, this looks good to me - thanks @jakkdl!

I'm also unsure whether changing OS thread names per-task is going to be a net gain for debuggability; I lean yes but will want to check in with some colleagues and @oremanj before we merge.

trio/_core/_thread_cache.py
```diff
@@ -1,7 +1,46 @@
...
+    return partial(namefunc, pthread_setname_np)
+    ## set the name
+    # try:
+    #    bname = name.encode("ascii", "replace")
+    #    thread = threading.current_thread()
+    #    # if thread is not None:
+    #    thread.name = name
+    #    pthread_setname_np(thread.ident, bname[:15])
+    # except Exception:
+    #    return nofunc
```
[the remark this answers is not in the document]
jakkdl: Oops, yep!

trio/_threads.py
```diff
@@ -168,6 +179,10 @@ def do_release_then_return_result():
 
     current_trio_token = trio.lowlevel.current_trio_token()
 
+    if thread_name is None:
+        # TODO, better default since it caps at 15 chars
+        thread_name = f"Thread for {trio.lowlevel.current_task().name}"
```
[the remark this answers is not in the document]
jakkdl: Yeah I'm not sure how to default it, 15 chars really isn't a lot and can easily get hit just with the function name. Could do `f'{trio.lowlevel.current_task().name.split(".")[-1]'[:14-len("thread_number") {thread number}]`

I think the last part of the name is usually the most interesting part? Or maybe `.__name__` on the passed function. And I think you might want the thread number in case you're launching several identical ones and want to disambiguate them.

jakkdl:
It seems to work fully on Mac now! The macOS 3.7 doesn't have pthreads apparently, but the later ones do and they pass the tests with flying colors. And windows doesn't crash (and sets Thread.name).
The codecov fail is due to differing platform behaviour, and one weird edge case of pthread but no pthread_setname_np that def should be pragma'd. But I'm not sure what the praxis is on the other platform stuff, I can litter everything non-linux with `#pragma: no cover` if you want.

Fuyukai (approved):
looks fine to me, idrc about the default thread name that much as there's no easy way to decide it. it's up to you if you want to change it or not.

trio/_threads.py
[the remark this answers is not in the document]
Zac-HD: I find the thread-ID point pretty convincing, so how about we just `thread_name.removeprefix("Thread for ")` before truncating and assigning the OS thread name?  See other comment for details.

trio/_core/_thread_cache.py
```diff
@@ -1,7 +1,48 @@
...
 from itertools import count
 
+from typing import Callable, Optional, Tuple
+from functools import partial
+
+
+def get_os_thread_name_func() -> Optional[Callable[[Optional[int], str], None]]:
+    def namefunc(setname: Callable[[int, bytes], int], ident: Optional[int], name: str):
+        if ident is not None:
+            setname(ident, bytes(name[:15], "ascii", "replace"))
```
Zac-HD, line 15: Let's add a helper function:

```python
def _to_os_thread_name(name: str) -> bytes:
    if name.startswith("Thread for "):  # str.removeprefix is new in 3.9
        name = name.replace("Thread for ", "")
    return name.encode("ascii", errors="replace")[:16]  # 15 chars + implicit trailing \x00
```

trio/_core/_thread_cache.py
```diff
@@ -1,7 +1,48 @@
...
+
+    def darwin_namefunc(
+        setname: Callable[[bytes], int], ident: Optional[int], name: str
+    ):
+        if ident is not None:
+            setname(bytes(name[:15], "ascii", "replace"))
```
Zac-HD, line 21: You comment below that `on mac OSX pthread_setname_np does not take a thread id`, but the function does actually take that - bug?  If they *can* have the same function signatures then there's some duplicate code instead.

Zac-HD (commented):
The coverage looks fine to me.  See comments below; after that I think we're ready to merge!

trio/_core/_thread_cache.py
jakkdl: The code calling namefunc will always pass a thread id, so it doesn't need to differ between platforms. But `setname` doesn't get `ident` passed to it.
I wasn't sure what to do if `ident is None` in `darwin_namefunc`. Documentation says that happens if the thread isn't started, but I don't really know what that means wrt naming the thread, but I went with the safe option of not doing anything in that case. (and idk if it *can* happen given Trio's usage of it)
Will add some comment to it.

trio/_threads.py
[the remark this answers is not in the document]
jakkdl: Oh actually, I think removing the prefix could trip up the end user. Somebody might manually assign "Thread for [a/b/c/d/e]" as their thread name. I'm ditching "Thread for" entirely, will push up a new version soon.

trio/_core/_thread_cache.py
jakkdl: btw `[:16]` doesn't work - it seems ctypes adds the null byte on it's own so it fails on anything more than `[:15]`

jakkdl:
Ubuntu3.7 seems to consistently fail for not-my-fault reasons.
I pragma'd all branches which I don't think ever gets entered, but codecov doesn't seem to include the macOS CI runs so 57-58 is the only failing part atm.

jakkdl:
So yeah afaik this should be mergeable now! @Zac-HD

trio/tests/test_threads.py
```diff
@@ -164,6 +167,119 @@ async def main():
...
+        print(f"no pthread on {sys.platform})")
+        return None
+    libpthread = ctypes.CDLL(libpthread_path)
+
+    pthread_getname_np = getattr(libpthread, "pthread_getname_np", None)
+    # this should never fail afaik, but if so just switch the below lines
+    assert pthread_getname_np
+    # if not pthread_getname_np: # pragma: no cover
+    #    print(f"no pthread_getname_np on {sys.platform})")
+    #    return None
```
Zac-HD, line 214: Uncomment or delete?

Zac-HD (approved):
The code itself looks good to me, but I'm not comfortable merging while we have CI failing with an SSL-related error.

trio/tests/test_threads.py
jakkdl: I guess line spacing didn't make it clear, but the comment on line 210 refers to 211 vs 212-214

jakkdl:
The SSL errors looks to be unrelated #2488

trio/tests/test_threads.py
jakkdl: but probably unnecessary, updated the comment and removed the commented out lines.

Zac-HD:
🎉 Thanks again to everyone involved in this (un?) surprisingly long pull request!

PR #2477, opened 2022-11-16
Add some low-effort type annotations

harahu:
This PR adds some uncontroversial annotations to the central `trio.run` method. From my perspective, it is mainly intended as an experiment into whether a gradual, opportunistic approach to adding types to this project is feasible.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,4 @@
 [mypy]
-show_error_codes = True
-
 # Be flexible about dependencies that don't have stubs yet (like pytest)
 ignore_missing_imports = True
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,5 @@
 [mypy]
-# TODO: run mypy against several OS/version combos in CI
-# https://mypy.readthedocs.io/en/latest/command_line.html#platform-configuration
+show_error_codes = True
 
 # Be flexible about dependencies that don't have stubs yet (like pytest)
 ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,4 +1,7 @@
 [mypy]
+# TODO: run mypy against several OS/version combos in CI
+# https://mypy.readthedocs.io/en/latest/command_line.html#platform-configuration
+
 # Be flexible about dependencies that don't have stubs yet (like pytest)
 ignore_missing_imports = True
 
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -76,7 +76,7 @@ def _count_context_run_tb_frames() -> int:
     try:
         ctx.run(function_with_unique_name_xyzzy)
     except ZeroDivisionError as exc:
-        tb = cast(TracebackType, exc.__traceback__)
+        tb = exc.__traceback__
         # Skip the frame where we caught it
         tb = tb.tb_next
         count = 0
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -12,7 +12,7 @@ import enum
 from contextvars import copy_context
 from math import inf
 from time import perf_counter
-from typing import Callable, TYPE_CHECKING
+from typing import Callable, Final, NoReturn, TYPE_CHECKING
 
 from sniffio import current_async_library_cvar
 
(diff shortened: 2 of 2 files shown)


pquentin:
Thank you for reviving this effort! Previous attempts are https://github.com/python-trio/trio-typing, https://github.com/python-trio/trio/pull/1873 and https://github.com/python-trio/trio/pull/2208.

A gradual approach is certainly possible: https://sethmlarson.dev/blog/tests-arent-enough-case-study-after-adding-types-to-urllib3. But I think you have to make it work file by file using an allowlist, as done for urllib3. Case in point, the mypy check is failing for this pull request.

harahu:
@pquentin We can probably figure out the exact mypy config as we go.

The reason it failed was due to the `check_untyped_defs = False` option being set. The moment I started adding types to a few previously untyped functions, I uncovered issues hidden in said functions. This is probably fine, though, as the alternative would be pretty noisy at the moment.

I've made mypy succeed now, but I ended up adding some uncovered lines in the process.

mypy.ini
```diff
@@ -1,6 +1,5 @@
 [mypy]
-# TODO: run mypy against several OS/version combos in CI
-# https://mypy.readthedocs.io/en/latest/command_line.html#platform-configuration
+show_error_codes = True
```
agronholm, line 2: This is the default setting as of mypy 0.990.

trio/_core/_run.py
```diff
@@ -13,7 +13,11 @@
 from contextvars import copy_context
 from math import inf
 from time import perf_counter
-from typing import Callable, TYPE_CHECKING
+from typing import Callable, NoReturn, TypeVar, TYPE_CHECKING
+from typing import Deque
```
A5rocks, line 17: This should be a single import

trio/_core/_run.py
```diff
@@ -47,14 +51,16 @@
 if sys.version_info < (3, 11):
     from exceptiongroup import BaseExceptionGroup
 
-DEADLINE_HEAP_MIN_PRUNE_THRESHOLD = 1000
+T = TypeVar("T")
```
A5rocks, line 54: Should we try going for a more descriptive `TypeVar`? `Fn = TypeVar("Fn", bound=Callable[..., Any])`

Completely optional though, what we have now works.

trio/_core/_run.py
A5rocks: Also quick note that `typing.Callable` is deprecated, even though the alternative is 3.9+ which is awkward.

harahu: Sorry for taking a long time to get back to you. Thank you for reviewing this. These issues should be addressed now.

trio/_core/_run.py
harahu: I went with a more descriptive type var.

trio/_core/_run.py
agronholm: `typing.Deque` is deprecated too, as `collections.deque` is generic now.

harahu: In the deque case, I'd rather not do anything until python 3.8 EOL since the type identifiers aren't exactly the same. I'd have to either do `from typing import Deque as deque` or `from collections.abc import deque as Deque`, both of which look kinda ugly tbh.

harahu: While trio still promises support for python versions  < 3.9, using `deque` as a generic type will lead to runtime errors, unless we also do `from __future__ import annotations`.

Typing with `Deque` allows for backwards compatibility until support for python 3.8 and older is discontinued.

agronholm: Well, what's the problem with `from __future__ import annotations`? AnyIO uses it everywhere and so far nobody has complained.

harahu: It makes some runtime uses of annotations more difficult. But I doubt that'll be an issue here. I can go with that approach.

A5rocks (approved):
The reordering of imports is unfortunate, but this looks good overall.

Unfortunately compatibility with `typing` is a bit cursed: `from __future__ import annotations` is in [limbo](https://docs.python.org/3.11/library/__future__.html#id2) (note the lack of resolution for 3.12). If that ever gets removed, we need to make sure we use that version's types at runtime -- of course, it shouldn't be. It's likely that we'll have to change it to [`co_annotations`](https://peps.python.org/pep-0649/) or similar soon. We'll just have to take this as we can. A good next step would be adding `mypy` to `ci.sh` to run it on every platform we test (other than PyPy ones, of course).

PR #2478, opened 2022-11-16
remove async_generator as dependency

### LINKED ISSUE #2458: Drop `async_generator` dependency
The `async_generator` backport has served us well, but now that we've dropped Python 3.5 async generators exist, and having dropped Python 3.6 all the introspection functions we need are now in the standard library too.

I expect this to be a fairly small cleanup, but it's always nice to delete code and dependencies you don't need anymore 😄

jakkdl:
Moved to the section of test dependencies in `test-requirenments.in` but all tests that require it are also guarded with skips. I didn't regenerate `test-requirements.txt` as that led to a couple changes irrelevant to this PR.

I considered giving a deprecation warning if it detects usage of `@async_generator`, but since aclosing() is only available on >=3.10 and trio supports >=3.7 there are still legitimate uses of it.

Thanks to @belm0 for giving the suggestion of handling checking for `@async_generator` that I ultimately went with.

Fixes #2458

### DIFF
--- a/trio/_core/_ki.py
+++ b/trio/_core/_ki.py
@@ -2,9 +2,8 @@ import inspect
 import signal
 import sys
 from functools import wraps
-import attr
 
-import async_generator
+import attr
 
 from .._util import is_main_thread
 
@@ -109,6 +108,14 @@ def currently_ki_protected():
     return ki_protection_enabled(sys._getframe())
 
 
+# This is to support the async_generator package necessary for aclosing on <3.10
+# functions decorated @async_generator are given this magic property that's a
+# reference to the object itself
+# see python-trio/async_generator/async_generator/_impl.py
+def legacy_isasyncgenfunction(obj):
+    return getattr(obj, "_async_gen_function", None) == id(obj)
+
+
 def _ki_protection_decorator(enabled):
     def decorator(fn):
         # In some version of Python, isgeneratorfunction returns true for
@@ -141,7 +148,7 @@ def _ki_protection_decorator(enabled):
                 return gen
 
             return wrapper
-        elif async_generator.isasyncgenfunction(fn):
+        elif inspect.isasyncgenfunction(fn) or legacy_isasyncgenfunction(fn):
 
             @wraps(fn)
             def wrapper(*args, **kwargs):
--- a/trio/_core/tests/test_asyncgen.py
+++ b/trio/_core/tests/test_asyncgen.py
@@ -1,13 +1,15 @@
 import sys
 import weakref
 import pytest
+import contextlib
 from math import inf
 from functools import partial
-from async_generator import aclosing
+
 from ... import _core
 from .tutil import gc_collect_harder, buggy_pypy_asyncgens, restore_unraisablehook
 
 
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="no aclosing() in stdlib<3.10")
 def test_asyncgen_basics():
     collected = []
 
@@ -46,7 +48,7 @@ def test_asyncgen_basics():
         assert collected.pop() == "abandoned"
 
         # aclosing() ensures it's cleaned up at point of use
-        async with aclosing(example("exhausted 1")) as aiter:
+        async with contextlib.aclosing(example("exhausted 1")) as aiter:
             assert 42 == await aiter.asend(None)
         assert collected.pop() == "exhausted 1"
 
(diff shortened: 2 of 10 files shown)


Zac-HD (approved):
Looks good to me - thanks @jakkdl 🙂 

I'm a little concerned about the coverage though, so will leave merging to someone with more detailed knowledge of internals.

trio/_core/_ki.py
```diff
@@ -109,6 +108,14 @@ def currently_ki_protected():
     return ki_protection_enabled(sys._getframe())
 
 
+# functions decorated @async_generator are given this magic property that's a
+# reference to the object itself
+# see python-trio/async_generator/async_generator/_impl.py
+# don't give deprecationwarning since aclosing() is not in stdlib <3.10
+def async_generator_isasyncgenfunction(obj):
```
belm0, line 115: perhaps naming it `legacy_isasyncgenfunction()` would be more clear

comment should clearly state this is supporting the async_generator backfill package

trio/_core/_ki.py
```diff
@@ -109,6 +108,14 @@ def currently_ki_protected():
...
+    return getattr(obj, "_async_gen_function", -1) == id(obj)
+
+
```
belm0, line 118: I know it's copying existing code, but `None` would be a more natural default

trio/_core/tests/test_asyncgen.py
```diff
@@ -3,10 +3,15 @@
...
-from async_generator import aclosing
+
 from ... import _core
 from .tutil import gc_collect_harder, buggy_pypy_asyncgens, restore_unraisablehook
 
+try:
+    from asyncgenerator import aclosing
+except ImportError:
+    aclosing = None
+
```
belm0, line 14: My opinion is that this is not worth it (dealing with more conditionals in the tests, coverage, etc.).

It's reasonable to just skip tests needing aclosing for older python versions (that will be dropped in the coming years anyway).  pytest makes this easy.

trio/_core/tests/test_ki.py
```diff
@@ -6,13 +6,12 @@
...
-from async_generator import (
-    async_generator,
-    yield_,
-    isasyncgenfunction,
-    asynccontextmanager,
-)
+try:
+    from async_generator import yield_, async_generator
+except ImportError:
+    async_generator = yield_ = None
```
belm0, line 14: I'm confused-- if `async_generator` is still declared in test dependencies, why would this import fail?

In any case, I don't understand why we're continuing to use the async_generator API in tests.

jakkdl: I did it to make sure that the code works both with and without `async_generator` installed and manually uninstalled it. But the risk of a backslide in the future should be pretty slim so yeah can get rid of the conditional.

But keeping tests that check that `@async_generator` works, i.e. `test_async_generator_agen_protection` I think is quite reasonable - if getting rid of those then kinda might as well get rid of async_generator support entirely.

I suppose I can recreate https://github.com/python-trio/async_generator/blob/master/async_generator/_impl.py#L435-L441 if we want to get rid of the dependency in testing too, feels a bit sketch but if we can assume that the implementation will never change then I guess it's okay.

Fuyukai:
I think removing async_generator support entirely is the better idea, but that's for another release imo as it can be considered by some a breaking change.

Fuyukai (approved):

trio/_core/_ki.py
```diff
@@ -109,6 +108,14 @@ def currently_ki_protected():
     return ki_protection_enabled(sys._getframe())
 
 
+# This is to support the async_generator package necessary for aclosing on <3.10
+# functions decorated @async_generator are given this magic property that's a
+# reference to the object itself
```
belm0, line 113: I think there are multiple scenarios of continued async_generator dependency by apps and libraries, so this comment may be confusing.

1. not using aclosing, but haven't bothered to migrate off of async_generator.  This is fair, and it may not be under an application's control-- the dependency may be coming in by transitive dependency, like some generator util library.
2. using async_generator for aclosing only (not using the module's generators) - it's perfectly fine, and has no bearing on this legacy detector.

trio/_core/tests/test_ki.py
belm0: > I did it to make sure that the code works both with and without async_generator

That's fair enough, but then the unit tests should explicitly test both those cases, rather than relying on the grace of some developer to manually uninstall async_generator and test it.

jakkdl: right, perfect solution would be to fiddle with `sys.modules` in that case. But feels like it's not worth putting a ton of effort into a perfect solution for something that's on the way out soonish. But this feels good enough for now to get the dependency out for the 99% that don't care for it, and cleanup can be put into a separate PR.

trio/_core/_ki.py
jakkdl: Sure, feel free to update it in a separate PR.

PR #2480, opened 2022-11-16
Modernize codebase to python 3.7+ standard

harahu:
This PR is the result of a one-off run of `pyupgrade` on the trio code base.

For future reference, this was done by running the following command from the project root:

```sh
find . -name '*.py' -exec pyupgrade --py3-only --py37-plus {} +
```

I also ran black in preview mode (`black --preview`), which mainly results in string cleanup.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 #
 # Trio documentation build configuration file, created by
 # sphinx-quickstart on Sat Jan 21 19:11:14 2017.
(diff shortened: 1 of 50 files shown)


harahu:
I hope this can be merged. I'm realizing that  it might seem daunting to review this, given the diff size, but the tools I've used are conservative w.r.t modifying behavior, so the changes should be relatively safe.

pquentin:
@Fuyukai as a member, you get to approve and merge this if you like it!

harahu:
I hope it can be merged. I want to make further changes on top of this.

pquentin (changes requested):
Thanks! Can you please exclude notes-to-self?

harahu:
> Thanks! Can you please exclude notes-to-self?

@pquentin I've done so now. I get one failing test suite, but I think it might be a case of test flakiness? What is your take?

pquentin:
Not sure what's going on sorry. I've relaunched the Ubuntu 3.7 tests a few times and it appears to fail consistently which is concerning.

harahu:
> Not sure what's going on sorry. I've relaunched the Ubuntu 3.7 tests a few times and it appears to fail consistently which is concerning.

I'll see if I can figure out what is going on. Perhaps later today.

pquentin:
I can confirm that the Ubuntu 3.7 failures are not specific to this pull request.

harahu:
> I can confirm that the Ubuntu 3.7 failures are not specific to this pull request.

Thanks @pquentin! Was scratching my head over this one.

pquentin:
Would you mind opening a pull request that skips this test?  I can only imagine that a fix in more recent versions was not backported there. I don't want to invest time in debugging this, especially when Python 3.7 won't be supported anymore in 7 months, and this is blocking us.

pquentin (approved):
LGTM! THanks for your patience.

pquentin:
(I'll let you click on the green merge button yourself!)

PR #2484, opened 2022-11-22
Update old style type hints

harahu:
Came across a few type hints embedded in comments. I figured it was a good idea to update these.

I leverage postponed evaluation of annotations in a few places. This allows me to get away with fewer imports from `Typing`, since I can use a more modern annotation style. Look for:

```py
```


pquentin (approved):
Thanks! LGTM.

PR #2502, opened 2022-12-13
Add types to `_unix_pipes`

harahu:
Complete type annotations for the `_unix_pipes` module.

### DIFF
--- a/trio/_unix_pipes.py
+++ b/trio/_unix_pipes.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
+
 import os
 import errno
+from typing_extensions import Final as FinalType
 
 from ._abc import Stream
 from ._util import ConflictDetector, Final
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! This looks good, but how did you test it? Is running mypy on this file working? If so we should do it in CI too.

harahu:
> This looks good, but how did you test it? Is running mypy on this file working?

@pquentin I tested it by running mypy on the file. It is working.

> If so we should do it in CI too.

We already are, for python 3.8 at least: https://github.com/python-trio/trio/actions/runs/3683469595/jobs/6232645740

pquentin:
This makes sure that the annotations are valid, but not that they are complete. If tomorrow I add a new function to this file without annotations mypy won't complain about it.

It's not blocking for this pull request but we should strive to have complete annotations for each file and enforce that in CI using mypy --strict. This approach was used successfully by urllib3.

harahu:
> It's not blocking for this pull request but we should strive to have complete annotations for each file and enforce that in CI using mypy --strict. This approach was used successfully by urllib3.

I'm aware that this technique has been used with success, and I'm a fan of introducing it. Doing so should be a separate PR. I might look into it soon.

For now, though, annotation coverage is so low that regressing coverage is the least of my worries.

PR #2507, opened 2022-12-20
Fix crash due to runtime import of `typing_extensions`

gschaffner:
hi!

#2502 added a runtime dependency on `typing_extensions` but did not update the packaging config to specify this dependency.

presently no runtime features of `typing_extensions` are being used, so it does not need to be a dependency at runtime.

(I have not included a changelog entry here since #2502 did not either and the bug this fixes was never user-facing.)

### DIFF
--- a/trio/_unix_pipes.py
+++ b/trio/_unix_pipes.py
@@ -2,12 +2,15 @@ from __future__ import annotations
 
 import os
 import errno
-from typing_extensions import Final as FinalType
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks!

A5rocks:
Looks good! Why didn't our test suit catch this? Do we need to separate out typing dependencies (mypy depends on `typing_extensions`) and test dependencies?

PR #2514, opened 2022-12-29
New Python versions for CI

A5rocks:
Python 3.12 released, PyPy 3.9 released.

Questions:
 - PyPy 3.7 is no longer in the support window for PyPy. Should it be removed?


A5rocks:
3.12 segfault is cause:
```py
import sys

def test_reproduction():
    sys.unraisablehook = sys.__unraisablehook__

    def finalizer(j):
        raise Exception("h")

    sys.set_asyncgen_hooks(finalizer=finalizer)

    async def agen_fn():
        yield

    agen = agen_fn()
    anext(agen)

test_reproduction(**{})
```

I will raise an issue upstream (and maybe fix it myself)

YES. EVERY BIT OF THAT IS NECESSARY. From `**{}` to `sys.unraisablehook = sys.__unraisablehook__`. Wow am I frustrated at this.

There's another GC thing but that's minor.

A5rocks:
The segfault was fixed in https://github.com/python/cpython/pull/100613

Waiting on 3.12a4 for next look. That will be ... today.

A5rocks:
GC is different.

```py
import weakref
import trio
import gc

async def f():
        nursery.cancel_scope.cancel()

    h = weakref.ref(nursery)
    del nursery
    print(h(), gc.get_referrers(h()) if h() is not None else [])

trio.run(f)
```

Under 3.10 this prints
```
None []
```

Under 3.12a3+ this prints
```
<trio.Nursery object at 0x7f3f5585d4f0> [NurseryManager(strict_exception_groups=False), <cell at 0x7f3f5585d540: Nursery object at 0x7f3f5585d4f0>, <frame at 0x7f3f55e220c0, file '/workspaces/trio/trio/_core/_run.py', line 1005, code _nested_child_finished>]
```

I will try minimizing this more.

richardsheridan (changes requested):
Some requests:

- rebase on master!
- remove pypy nightlies, including in ci.sh
- add `pypy-3.9-dev`
- add (unsupported) pypy-3.7 to `continue-on-error` (we should remove entirely when cpython drops 3.7 support)

richardsheridan:
I thing it's great you're fixing bugs off of this but I don't think that's any reason to keep the PR as draft forever!

A5rocks:
> remove pypy nightlies, including in ci.sh

See https://github.com/python-trio/trio/pull/2536 maybe? (that'll have to be updated after this is merged)

> add `pypy-3.9-dev`

See also that PR :P

> add (unsupported) `pypy-3.7` to continue-on-error (we should remove entirely when cpython drops 3.7 support)

Hmm... I think requiring stable PyPy 3.7 to pass while we support Python 3.7 should be fine. I removed PyPy 3.7 nightly (again, above PR) because PyPy 3.7 is no longer supported as you mention.

---

I hope that is all? (I added a merge commit instead of a rebase cause... idk, this should be squash merged instead of merged with commit anyways)

richardsheridan (approved):
I was wondering the relationship between these two PRs!

Issue #2534, opened 2023-01-18
`from_thread.run(_sync)?` fails in any thread running Trio


gschaffner:
hi! `from_thread.run(_sync)?` uses a heuristic to prevent the user from accidentally deadlocking their event loop by writing code like https://github.com/python-trio/trio/pull/1574/files#r435039316.

the heuristic is currently much more aggressive than is required to prevent this deadlock, however. it has the side-effect[^1] of preventing one from calling `from_thread.run(_sync)?` in _any_ thread running trio, _even though such calls are almost always safe and deadlock-free_.

[^1]: citing the `RuntimeError`s reported in `from_thread.run(_sync)?`'s docstrings, this would appear to be an unintended side-effect of the current heuristic.

it would be useful to be able to call `from_thread.run(_sync)?` from trio threads[^2], but the current heuristic blindly prevents this. i'd like to propose relaxing `from_thread.run(_sync)?` to allow calling `from_thread.run(_sync)?` from trio threads. (one use-case of this is https://github.com/agronholm/anyio/issues/525.)

[^2]: for example in order to call `Nursery.start_soon` in a thread-safe manner.

oremanj:
I agree that the check is more zealous than it strictly needs to be. I don't think the idea of multiple Trio threads was contemplated when `trio.from_thread` was written, and even if you loosen the check, `from_thread` is still not a good fit for Trio-to-Trio calls (because it will block the originating thread until the target thread finishes doing whatever you asked it to do). I think a different interface that is designed for cross-Trio-thread calls (with the entrypoint functions being async instead of synchronous) would be clearer; what do you think?

I'm curious what you're doing that requires multiple Trio threads anyway -- we've considered the possibility but hadn't come across anyone with a concrete use case!

smurfix:
Isn't it fairly trivial to cross call from one Trio thread to the other?
`await trio.to_thread(functools.partial(trio.from_thread.run, fn, *args, trio_token=other_task.token))`

If you don't want the overhead, well, creating something like `trio._threads.cross_thread.run(fn, *args)` should be fairly simple.

However IMHO relaxing the rules about not calling any `trio.from_thread.*` function from within Trio isn't the right way to do this.

oremanj:
@gschnaffer thank you for the extremely detailed response! I'm convinced :-)

[...]

PR #2535, opened 2023-01-18
Fix `from_thread.run(_sync)?` failing in any thread running Trio

### LINKED ISSUE #2191: run_fn_as_system_task does not account for nested Trio
I'm new to Trio, so forgive me if I get the terminology wrong. [This check](https://github.com/python-trio/trio/blob/141bd6e8e53c51ce8fcd8f830b4ec2160ae2af1d/trio/_threads.py#L234) for deadlock only checks if there is a Trio task at all, not that the task is the same as the caller. So, if we have something like this:

```py
from concurrent.futures import as_completed
from anyio.from_thread import start_blocking_portal

async def endpoint():
  with start_blocking_portal() as portal:
    futures = [portal.start_task_soon(long_running_task, i) for i in range(1, 5)]
    for future in as_completed(futures):
        print(future.result())
```

It will fail and exit the portal prematurely because `endpoint` is running in Trio, but the portal is a different Trio task. The check sees if there's a task at all, rather than checking if the call would actually deadlock the task in the case of nesting.

If I remove the current task check, the portal runs successfully with no deadlock.

gschaffner:
resolves #2191. resolves its [duplicate](https://github.com/python-trio/trio/issues/2534#issuecomment-1403376261) #2534.

with this patch, trio would only prevent users from running `from_thread.run(_sync)?` when trio _knows_ that such a call would deadlock. so, with this patch, trio still prevents "first-order" `from_thread.run(_sync)?` deadlocks like https://github.com/python-trio/trio/pull/1574/files#r435039316. i call this first-order because it's the smallest possible cyclic graph: `trio_threadA` is blocking waiting for `trio_threadA`.

there is a side-effect of this, however. the deadlock heuristic's current aggressiveness means that trio also currently prevents higher-order `from_thread.run(_sync)?` deadlocks, such as a second-order deadlock where the graph cycle is `trio_threadA -> trio_threadB -> trio_threadA`. higher-order deadlocks can be created through contrived examples:

<details>
<summary>(click to expand) a higher order `trio.from_thread.run(_sync)?` cyclic deadlock</summary>

```python
from collections.abc import Awaitable
from collections.abc import Callable
from concurrent.futures import Future
from typing import Any
from typing import TypeVar

import trio
from trio import Nursery
from trio.lowlevel import TrioToken

T = TypeVar("T")


async def cyclic_deadlock(order: int) -> None:
    # create a cyclic deadlock where `order` threads form a cycle and each thread blocks
    # itself by calling `from_thread.run(_sync)?` to wait on the next thread in the
    # cycle.
    #
    # for reference, the `order=1` case reduces to
    # https://github.com/python-trio/trio/pull/1574/files#r435039316.
    init_token = trio.lowlevel.current_trio_token()

    if order < 1:
        raise ValueError("order must be positive")

    async def cyclic_deadlock(order: int) -> None:
        print(f"blocking thread #{order} now")
        remaining = order - 1
        if remaining:
            # in concept, this ACM block is like
            # `anyio.from_thread.start_blocking_portal`.
                next_token_future = Future[TrioToken]()

                async def next_thread_worker() -> None:
                    next_token_future.set_result(trio.lowlevel.current_trio_token())
                    await trio.sleep_forever()

                nursery.start_soon(to_thread_run, next_thread_worker)
                try:
                    next_token = await resolve_future(next_token_future)
                    # in concept, this call is like `BlockingPortal.call`.
                    trio.from_thread.run(
                        cyclic_deadlock, remaining, trio_token=next_token
                    )
                finally:
                    nursery.cancel_scope.cancel()
        else:
            print("entering the cyclic deadlock now!")
            trio.from_thread.run_sync(lambda: None, trio_token=init_token)

    await cyclic_deadlock(order)


async def to_thread_run(afn: Callable[..., Awaitable[T]], *args: Any) -> T:
    # like `to_thread.run_sync`, but runs an async function instead of a sync function.
    #
    # this differs from `to_thread.run_sync(trio.run, partial(afn, *args),
    # cancellable=True)` in an important way: it obeys normal cancellation semantics,
    # i.e. it doesn't unstructure the concurrency, i.e. it can't leave behind a dangling
    # thread!
    #
    # this function can be used to create a "daemonic-like" thread: a thread that exits
    # if it's still alive when the nursery that started it cancels. unlike Python's
    # normal daemonic threads, a daemonic-like thread exits when its "parent" nursery
    # exits rather than when all non-daemonic threads exit. also unlike Python's normal
    # daemonic threads, daemonic-like threads do not skip cleanup code (such as `except
    # Cancelled`, `finally`, `__exit__`, `__aexit__`, etc.) when they exit.
    #
    # ...
    #
    # this seems pretty obscure and unlikely to be needed in real code, though. if
    # you're in a trio thread and need to run a async function, you just...run the async
    # function. conceivably, i suppose, you might want to spin out a second thread
    # running trio to run an async function that is not well-behaved (i.e. it goes a
    # long time without checkpointing, interfering with other tasks on the event loop?).
    # doing so can obviously be buggy when used with interthread communication if not
    # done carefully since (a) there may not be a GIL and (b) if there is a GIL, it's
    # free to switch at places that are not checkpoints.
    nursery_future = Future[Nursery]()

    async def worker() -> T:
            nursery_future.set_result(nursery)
            return await afn(*args)

    try:
        return await trio.to_thread.run_sync(trio.run, worker, cancellable=True)
    finally:
        try:
            nursery = nursery_future.result(timeout=0)
        except TimeoutError:
            # `to_thread.run_sync` raised either because it failed to start `worker` or
            # because `worker` failed while opening the nursery or setting
            # `nursery_future`. thus the nursery either never opened or has already died
            # on its own, so it's okay that we can't cancel it.
            pass
        else:
            nursery.cancel_scope.cancel()


async def resolve_future(future: Future[T], /) -> T:
    # async variant of `Future.result`.
    #
    # note: this implementation requires Python >= 3.8 (as it does not support `Future`s
    # that have a result/exception set multiple times).

    # note: it probably would be better to do done-signaling with sockets instead so
    # that trio can handle the polling and we wouldn't have to hardcode a sleep interval
    # to prevent spinlocks.
    while not future.done():
        await trio.sleep(0.05)
    return future.result()


if __name__ == "__main__":
    trio.run(cyclic_deadlock, 2)
```
</details>

but while a higher-order deadlock can be created through a contrived example, i don't think they would ever arise naturally; they seem to require having:

1.  multiple threads running trio

1.  at least two trio threads sharing their tokens with another trio thread (in a manner that creates a graph cycle of length >= 2)

1.  a trio thread (say it has token `tokenA`) calling `trio.from_thread.run(_sync)?(foo, trio_token=tokenB)`

1.  `foo` being a function that will result in a trio thread calling `trio.from_thread.run(_sync)?(..., trio_token=tokenA)`.

it seems like writing code that does all three of these things is not something that someone could actually do without realizing that they've probably written a deadlock. (1) is probably uncommon already, (2) requires pretty heavy sharing of `lowlevel.TrioToken`s in a way that raises red flags and screams "potential deadlock!", and (3) requires choosing to call a blocking function (that is not short-lived) in a trio thread.

it seems to me that the motivation for the current deadlock heuristic was to prevent first-order `from_thread.run(_sync)?` deadlocks (their implementation is only one line of code, after all), but i don't think that `from_thread.run(_sync)?` should be artificially restricted to prevent users from shooting themselves with a footgun that (unless i have missed something important...) appears to be rather difficult to manage to shoot oneself with.

### DIFF
--- a/trio/_threads.py
+++ b/trio/_threads.py
@@ -245,13 +245,18 @@ def _run_fn_as_system_task(cb, fn, *args, context, trio_token=None):
                 "this thread wasn't created by Trio, pass kwarg trio_token=..."
             )
 
-    # Avoid deadlock by making sure we're not called from Trio thread
+    # Avoid deadlock by making sure we're not called from the same Trio thread we're
+    # trying to enter
     try:
-        trio.lowlevel.current_task()
+        current_trio_token = trio.lowlevel.current_trio_token()
     except RuntimeError:
         pass
     else:
-        raise RuntimeError("this is a blocking function; call it from a thread")
+        if trio_token == current_trio_token:
+            raise RuntimeError(
+                "this is a blocking function; using it to re-enter the current Trio "
+                "thread would cause a deadlock"
+            )
 
     q = stdlib_queue.SimpleQueue()
     trio_token.run_sync_soon(context.run, cb, q, fn, args)
--- a/newsfragments/2191.bugfix.rst
+++ b/newsfragments/2191.bugfix.rst
@@ -1,0 +1,1 @@
+`trio.from_thread.run` and `trio.from_thread.run_sync` no longer raise `RuntimeError` when used to enter a thread running Trio from a different thread running Trio.


trio/_threads.py
```diff
@@ -245,13 +245,17 @@ def _run_fn_as_system_task(cb, fn, *args, context, trio_token=None):
                 "this thread wasn't created by Trio, pass kwarg trio_token=..."
             )
 
-    # Avoid deadlock by making sure we're not called from Trio thread
+    # Avoid deadlock by making sure we're not called from the token's Trio thread
```
oremanj, line 248: maybe "from the same thread we're trying to enter"

newsfragments/2534.bugfix.rst
```diff
@@ -0,0 +1 @@
+`trio.from_thread.run` and `trio.from_thread.run_sync` no longer raise `RuntimeError` when called in a deadlock-free way from a thread running Trio.
```
oremanj, line 1: instead of "from a thread running Trio", maybe elaborate "to enter one thread running Trio from another"? I also wonder if it would be good to add a disclaimer here that using them this way is still suboptimal (will block the originating thread's event loop) but we support it because it can be useful in some cases.

oremanj:
Coverage is failing because the `else` case of the new condition you added is not exercised -- could you add a test that exercises it?

njsmith:
.....I'm not super keen. If you want to call between Trio threads, then it would be better to have an async way to do that? Intentionally doing blocking calls from a Trio thread can work under the right circumstances, but it's Sketchy. Also, while running multiple Trio threads simultaneously is definitely supported, it's a bit odd and I wonder what you're doing, and if there's a better way to accomplish it?

PR #2536, opened 2023-01-18
Attempt to fix PyPy CI

A5rocks:
This should in theory enable the already defined PyPy runs on MacOS and Windows. They never worked!!

This also bumps the `@actions/setup-python` version we use (the old one errored for nightly PyPy versions & gave us a warning in action runs due to using node 12 vs 16)


webknjaz:
> 3332 Segmentation fault      (core dumped)

I wonder if this is happening consistently.. Does faulthandler work under PyPy? In some cases, I've found that pytest was crashing due to an old version of its `py` dependency and `pytest-cov` was hiding that (the problem only became apparent after disabling it).

webknjaz:
> I wonder if this is happening consistently..

retriggered to verify: https://github.com/python-trio/trio/actions/runs/3945507679/jobs/6766090599.

A5rocks:
> Oh, and it's not clear why not use PyPy that GHA ships for testing...

I was simply updating `ci.sh` -- now that I look at it, I see that `@actions/python-setup` can install a nightly version of pypy! I'll change that out soon.

webknjaz:
> > 3332 Segmentation fault      (core dumped)
> 
> I wonder if this is happening consistently..

I just saw it outside this PR.

webknjaz:
> On a related note, I've recently made an action that allows running tests from sdist instead of relying on the Git checkout which may miss including important bits needed for testing: [twitter.com/webknjaz/status/1609252309749440512](https://twitter.com/webknjaz/status/1609252309749440512) — maybe somebody will get to integrating it before I do.

I've taken a stab at it https://github.com/python-trio/trio/pull/2541

A5rocks:
It looks like the failure in `trio/_core/tests/test_windows.py::test_readinto_overlapped` *should* happen. We pass an immutable bytes object to something that eventually calls `ffi.from_buffer(buffer, require_writable=True)` (which certainly *sounds* like it won't accept an immutable bytes object). Why hasn't this errored before?

I will look at the rest of the errors at a later time.

A5rocks:
OK I think this is ready for another review. The workflows *will* error but they shouldn't block required checks. We can figure those out at a later date.

A5rocks:
Looks like all that needs to happen is macos needs its timeout bumped to like 20 minutes (just like windows) https://github.com/python-trio/trio/actions/runs/4169791936/jobs/7218120369

A5rocks:
Both MacOS tests timed out on `test_handshake_over_terrible_network`. I don't really have a really accessible macos device but I might be able to try to reproduce this...

A5rocks:
I filed an issue about our GC behavior over at <https://foss.heptapod.net/pypy/pypy/-/issues/3900>; hopefully they can fix our problem or tell us what we're missing.

PR #2537, opened 2023-01-18
Fix crash due to runtime import of `typing_extensions`

gschaffner:
hi! the issue fixed in #2507 was re-introduced when #2477 was merged.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -23,9 +23,6 @@ from outcome import Error, Outcome, Value, capture
 from sniffio import current_async_library_cvar
 from sortedcontainers import SortedDict
 
-# An unfortunate name collision here with trio._util.Final
-from typing_extensions import Final as FinalT
-
 from .. import _core
 from .._util import Final, NoPublicConstructor, coroutine_or_error
 from ._asyncgens import AsyncGenerators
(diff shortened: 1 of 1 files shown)


pquentin:
@harahu Do you think we could add a test to avoid introducing this a third time?

harahu:
Sorry for not picking up on this!

@pquentin My suggestion would be to, rather than adding a test, I'd make sure that dynamic analysis in CI (testing, that is) is run in an environment that is as minimal as possible. That could be achieved by separating static and dynamic analysis.

To be a bit more concrete: In my understanding, this problem was camouflaged because `typing-extensions` is listed in `test-requirements.*`. `typing-extensions` is a test requirement because, in this project, "test" refers both to dynamic tests, but also static analysis, like running mypy.

If we break this up into a new `static-analysis-requirements.*` alongside a trimmed down `test-requirements.*`, and we make sure pytest is run with only `test-requirements.*` installed, we are more likely to detect problems like this.

I'm still getting to know this project, so I don't know if my understanding is correct, so please share your thoughts.

harahu:
Alternatively, if `flake8` is run in CI, there's a neat plugin that maybe would have caught this: https://pypi.org/project/flake8-requirements/

PR #2538, opened 2023-01-18
Sort imports

harahu:
Using `isort` to sort imports. In my opinion, this improves readability.


jakkdl:
is there any reason this wasn't followed up on / isort isn't used in general? I'm a big fan of isort

A5rocks:
I mean, other than isort's order breaking trio?

Fuyukai:
isort also doesn't support the geriatric versions of Python that trio mandates support for anymore so you can't even install it unless you pin an old version.

jakkdl:
> isort also doesn't support the geriatric versions of Python that trio mandates support for anymore so you can't even install it unless you pin an old version.

that doesn't stop us from running isort in CI - which runs on 3.8 currently.

jakkdl:
> I mean, other than isort's order breaking trio?

you can `isort: skip` files where you have to avoid circular imports - but I like using it to enforce import formatting, sorting, pruning unused imports, etc to improve git diffs and increase readability.

jakkdl:
Implemented in #2686

PR #2540, opened 2023-01-18
ConflictDetector: kill both tasks

smurfix:
It's not always obvious which task is the "bad guy" when a conflict occurs. If the "legitimate" task crashes and the buggy one, i.e. the one that forgot to take the lock, does not, the backtrace is not helpful.

Thus, kill them both.

### DIFF
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -180,15 +180,20 @@ class ConflictDetector:
     def __init__(self, msg):
         self._msg = msg
         self._held = False
-
+        self._conflicted = False
+    
     def __enter__(self):
         if self._held:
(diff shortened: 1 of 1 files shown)


smurfix:
These codecov results seem rather spurious.

oremanj:
This is a breaking change. Previously, if you wanted to see if some other task was waiting on an FD, you could try to wait on it and catch BusyResourceError. I have used this pattern to implement "make sure someone is waiting" on multiplexed interfaces (such as inotify) without a system task. I would be frustrated if we broke that.

thomld20 (approved):

PR #2549, opened 2023-01-28
Add generic typing support for Memory[Send/Receive]Channel

### LINKED ISSUE #1327: `MemoryXXXChannel` doesn't support generic typing
When using `mypy` and type hints for `trio.abc.SendChannel` and `trio.abc.ReceiveChannel`, `mypy` complains about calling things like:

- `channel.clone()`
- `channel.send_nowait()`

This is because `trio.abc.XXXChannel` don't have these interfaces, only the `MemoryXXXChannel` does.  The problem is that the `trio.abc.SendChannel` allow for specifying the type of object that passes through the channel via `trio.abc.SendChannel[SomeType]`, however, the `MemoryXXXChannel` objects do not.

Both of the solutions for this are not ideal.

- `cast(MemoryXXXChannel, my_channel)`: this loses the object type that passes through the channel.
- `# type: ignore`: this just silences the error....

jakkdl:
Fixes #1327

I didn't figure out a good way to add tests for it, but I reproduced the error from the original PR, and it now works as expected (in a somewhat pared down variant of it):
```python
from trio import open_memory_channel, MemorySendChannel, MemoryReceiveChannel
from typing import cast, Tuple

OutboundBroadcastChannelPair = Tuple[
    MemorySendChannel[int],
    MemoryReceiveChannel[int],
]

(outbound_send_channel, outbound_receive_channel) = cast(
    OutboundBroadcastChannelPair, open_memory_channel(100)
)

outbound_send_channel.send_nowait(5)
outbound_send_channel.send_nowait(None) # argument 1 has incompatible type

reveal_type(outbound_receive_channel.receive_nowait()) # revealed type is builtins.int
```
One way would be to add type hints to the tests in `trio/tests/test_channel.py`, and also check the `trio.tests` submodule with mypy in `check.sh`

As noted in https://github.com/python-trio/trio/blob/master/trio/_util.py#L251-L254 `@generic_function` won't type-check, and I didn't figure out what mypy plugin / clever stub would enable that.

I didn't manage to figure out what `trio.lowlevel.wait_task_rescheduled` actually returns, so it's possible that the return type on `MemoryReceiveChannel.receive` is incorrect.

I assumed that the co/contravariantness of the `TypeVar`s in `trio/_abc.py` for `SendChannel` and `MemoryChannel` would apply for Memory[Send/Receive]Channel as well.

While on it I also typed ~everything in `trio/_channel.py`. (though the `__exit__` ones look kinda crazy :sweat_smile: )

### DIFF
--- a/trio/_channel.py
+++ b/trio/_channel.py
@@ -1,5 +1,18 @@
+from __future__ import annotations
+
 from collections import deque, OrderedDict
+from collections.abc import Callable
 from math import inf
+
+from types import TracebackType
+from typing import (
+    Any,
+    Generic,
+    NoReturn,
+    TypeVar,
+    TYPE_CHECKING,
+    Tuple,  # only needed for typechecking on <3.9
+)
 
 import attr
 from outcome import Error, Value
@@ -8,11 +21,28 @@ from .abc import SendChannel, ReceiveChannel, Channel
 from ._util import generic_function, NoPublicConstructor
 
 import trio
-from ._core import enable_ki_protection
+from ._core import enable_ki_protection, Task, Abort, RaiseCancelT
+
+# A regular invariant generic type
+T = TypeVar("T")
+
+# The type of object produced by a ReceiveChannel (covariant because
+# ReceiveChannel[Derived] can be passed to someone expecting
+# ReceiveChannel[Base])
+ReceiveType = TypeVar("ReceiveType", covariant=True)
+
+# The type of object accepted by a SendChannel (contravariant because
+# SendChannel[Base] can be passed to someone expecting
+# SendChannel[Derived])
+SendType = TypeVar("SendType", contravariant=True)
+
+# Temporary TypeVar needed until mypy release supports Self as a type
+SelfT = TypeVar("SelfT")
 
 
-@generic_function
-def open_memory_channel(max_buffer_size):
+def _open_memory_channel(
+    max_buffer_size: int,
+) -> tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]:
     """Open a channel for passing objects between tasks within a process.
 
     Memory channels are lightweight, cheap to allocate, and entirely
@@ -68,36 +98,57 @@ def open_memory_channel(max_buffer_size):
         raise TypeError("max_buffer_size must be an integer or math.inf")
     if max_buffer_size < 0:
         raise ValueError("max_buffer_size must be >= 0")
-    state = MemoryChannelState(max_buffer_size)
+    state: MemoryChannelState[T] = MemoryChannelState(max_buffer_size)
     return (
-        MemorySendChannel._create(state),
-        MemoryReceiveChannel._create(state),
+        MemorySendChannel[T]._create(state),
+        MemoryReceiveChannel[T]._create(state),
     )
+
+
+# This workaround requires python3.9+, once older python versions are not supported
+# or there's a better way of achieving type-checking on a generic factory function,
+# it could replace the normal function header
+if TYPE_CHECKING:
+    # written as a class so you can say open_memory_channel[int](5)
+    # Need to use Tuple instead of tuple due to CI check running on 3.8
+    class open_memory_channel(Tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]):
+        def __new__(  # type: ignore[misc]  # "must return a subtype"
+            cls, max_buffer_size: int
+        ) -> tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]:
+            return _open_memory_channel(max_buffer_size)
+
+        def __init__(self, max_buffer_size: int):
+            ...
+
+else:
+    # apply the generic_function decorator to make open_memory_channel indexable
+    # so it's valid to say e.g. ``open_memory_channel[bytes](5)`` at runtime
+    open_memory_channel = generic_function(_open_memory_channel)
 
 
 @attr.s(frozen=True, slots=True)
 class MemoryChannelStats:
-    current_buffer_used = attr.ib()
-    max_buffer_size = attr.ib()
-    open_send_channels = attr.ib()
-    open_receive_channels = attr.ib()
-    tasks_waiting_send = attr.ib()
-    tasks_waiting_receive = attr.ib()
+    current_buffer_used: int = attr.ib()
+    max_buffer_size: int = attr.ib()
+    open_send_channels: int = attr.ib()
+    open_receive_channels: int = attr.ib()
+    tasks_waiting_send: int = attr.ib()
+    tasks_waiting_receive: int = attr.ib()
 
 
 @attr.s(slots=True)
-class MemoryChannelState:
-    max_buffer_size = attr.ib()
-    data = attr.ib(factory=deque)
+class MemoryChannelState(Generic[T]):
+    max_buffer_size: int = attr.ib()
+    data: deque[T] = attr.ib(factory=deque)
     # Counts of open endpoints using this state
-    open_send_channels = attr.ib(default=0)
-    open_receive_channels = attr.ib(default=0)
+    open_send_channels: int = attr.ib(default=0)
+    open_receive_channels: int = attr.ib(default=0)
     # {task: value}
-    send_tasks = attr.ib(factory=OrderedDict)
+    send_tasks: OrderedDict[Task, T] = attr.ib(factory=OrderedDict)
     # {task: None}
-    receive_tasks = attr.ib(factory=OrderedDict)
+    receive_tasks: OrderedDict[Task, None] = attr.ib(factory=OrderedDict)
 
-    def statistics(self):
+    def statistics(self) -> MemoryChannelStats:
         return MemoryChannelStats(
             current_buffer_used=len(self.data),
             max_buffer_size=self.max_buffer_size,
@@ -109,28 +160,28 @@ class MemoryChannelState:
 
 
 @attr.s(eq=False, repr=False)
-class MemorySendChannel(SendChannel, metaclass=NoPublicConstructor):
-    _state = attr.ib()
-    _closed = attr.ib(default=False)
+class MemorySendChannel(SendChannel[SendType], metaclass=NoPublicConstructor):
+    _state: MemoryChannelState[SendType] = attr.ib()
+    _closed: bool = attr.ib(default=False)
     # This is just the tasks waiting on *this* object. As compared to
     # self._state.send_tasks, which includes tasks from this object and
     # all clones.
-    _tasks = attr.ib(factory=set)
+    _tasks: set[Task] = attr.ib(factory=set)
 
-    def __attrs_post_init__(self):
+    def __attrs_post_init__(self) -> None:
         self._state.open_send_channels += 1
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "<send channel at {:#x}, using buffer at {:#x}>".format(
             id(self), id(self._state)
         )
 
-    def statistics(self):
+    def statistics(self) -> MemoryChannelStats:
         # XX should we also report statistics specific to this object?
         return self._state.statistics()
 
     @enable_ki_protection
-    def send_nowait(self, value):
+    def send_nowait(self, value: SendType) -> None:
         """Like `~trio.abc.SendChannel.send`, but if the channel's buffer is
         full, raises `WouldBlock` instead of blocking.
 
@@ -150,7 +201,7 @@ class MemorySendChannel(SendChannel, metaclass=NoPublicConstructor):
             raise trio.WouldBlock
 
     @enable_ki_protection
-    async def send(self, value):
+    async def send(self, value: SendType) -> None:
         """See `SendChannel.send <trio.abc.SendChannel.send>`.
 
         Memory channels allow multiple tasks to call `send` at the same time.
@@ -170,15 +221,16 @@ class MemorySendChannel(SendChannel, metaclass=NoPublicConstructor):
         self._state.send_tasks[task] = value
         task.custom_sleep_data = self
 
-        def abort_fn(_):
+        def abort_fn(_: RaiseCancelT) -> Abort:
             self._tasks.remove(task)
             del self._state.send_tasks[task]
             return trio.lowlevel.Abort.SUCCEEDED
 
         await trio.lowlevel.wait_task_rescheduled(abort_fn)
 
+    # Return type must be stringified or use a TypeVar
     @enable_ki_protection
-    def clone(self):
+    def clone(self) -> "MemorySendChannel[SendType]":
         """Clone this send channel object.
 
         This returns a new `MemorySendChannel` object, which acts as a
@@ -206,14 +258,19 @@ class MemorySendChannel(SendChannel, metaclass=NoPublicConstructor):
             raise trio.ClosedResourceError
         return MemorySendChannel._create(self._state)
 
-    def __enter__(self):
+    def __enter__(self: SelfT) -> SelfT:
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
         self.close()
 
     @enable_ki_protection
-    def close(self):
+    def close(self) -> None:
         """Close this send channel object synchronously.
 
         All channel objects have an asynchronous `~.AsyncResource.aclose` method.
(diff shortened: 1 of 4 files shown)


trio/_channel.py
```diff
@@ -109,28 +137,28 @@ def statistics(self):
 
 
 @attr.s(eq=False, repr=False)
-class MemorySendChannel(SendChannel, metaclass=NoPublicConstructor):
-    _state = attr.ib()
-    _closed = attr.ib(default=False)
+class MemorySendChannel(SendChannel, Generic[SendType], metaclass=NoPublicConstructor):
```
A5rocks, line 140: Should `SendChannel` be generic?

trio/_channel.py
```diff
@@ -241,30 +275,32 @@ def close(self):
...
 
 @attr.s(eq=False, repr=False)
-class MemoryReceiveChannel(ReceiveChannel, metaclass=NoPublicConstructor):
-    _state = attr.ib()
-    _closed = attr.ib(default=False)
-    _tasks = attr.ib(factory=set)
-
-    def __attrs_post_init__(self):
+class MemoryReceiveChannel(
+    ReceiveChannel, Generic[ReceiveType], metaclass=NoPublicConstructor
```
A5rocks, line 285: Should `ReceiveChannel` be generic?

trio/_channel.py
```diff
@@ -1,18 +1,46 @@
...
+
+# The type of object produced by a ReceiveChannel (covariant because
+# ReceiveChannel[Derived] can be passed to someone expecting
+# ReceiveChannel[Base])
+ReceiveType = TypeVar("ReceiveType", covariant=True)
+
+# The type of object accepted by a SendChannel (contravariant because
+# SendChannel[Base] can be passed to someone expecting
+# SendChannel[Derived])
+SendType = TypeVar("SendType", contravariant=True)
```
A5rocks, line 37: This feels a bit wrong but thinking about it, it looks right. (I made some code examples to prove it to myself, this variance stuff is really tricky).

trio/_channel.py
```diff
@@ -170,15 +198,16 @@ async def send(self, value):
         self._state.send_tasks[task] = value
         task.custom_sleep_data = self
 
-        def abort_fn(_):
+        def abort_fn(_) -> Abort:
```
A5rocks, line 201: Should the argument be typed here?

trio/_channel.py
```diff
@@ -206,14 +235,19 @@ def clone(self):
             raise trio.ClosedResourceError
         return MemorySendChannel._create(self._state)
 
-    def __enter__(self):
+    def __enter__(self) -> "MemorySendChannel[SendType]":
```
A5rocks, line 238: This should return `Self`. (from `typing_extensions`, that is). (unlike `clone` above, this doesn't always return `MemorySendChannel`.)

If we don't want to use that, we can do `def __enter__(self: SelfT) -> SelfT` where `SelfT = TypeVar("SelfT")`.

trio/_channel.py
```diff
@@ -306,15 +342,15 @@ async def receive(self):
...
         task.custom_sleep_data = self
 
-        def abort_fn(_):
+        def abort_fn(_) -> Abort:
             self._tasks.remove(task)
             del self._state.receive_tasks[task]
             return trio.lowlevel.Abort.SUCCEEDED
 
-        return await trio.lowlevel.wait_task_rescheduled(abort_fn)
+        return await trio.lowlevel.wait_task_rescheduled(abort_fn)  # type: ignore
```
A5rocks, line 350: Why does this need a `# type: ignore` but not the other similar pattern in `MemorySendChannel`.

trio/_channel.py
```diff
@@ -345,14 +381,19 @@ def clone(self):
             raise trio.ClosedResourceError
         return MemoryReceiveChannel._create(self._state)
 
-    def __enter__(self):
+    def __enter__(self) -> "MemoryReceiveChannel[ReceiveType]":
```
A5rocks, line 384: Should return `Self`, I think.

A5rocks (changes requested):
I've got a couple nitpicks re: types

trio/_channel.py
jakkdl: the abc has generic, so I think you're right that the type should be passed through to it as well here.

trio/_channel.py
jakkdl: Yes, but I gave up on figuring out what it should be :sweat_smile:

I started trying to figure out `trio.lowlevel.wait_task_rescheduled` but I had a lot of trouble with it and ultimately decided to postpone tackling lowlevel - since at least in this case it doesn't matter when the parameter is unused.
[looking at trio-typing](https://github.com/python-trio/trio-typing/blob/d5835f7065f58ebdd95ca27b09d1d5cbe686b8b8/trio-stubs/lowlevel.pyi#L131-L133) i think it should be `Callable[[], NoReturn]`, though don't ask me why.

Also see discussion further down on the `type: ignore`

trio/_channel.py
jakkdl: `MemorySendChannel.send()` doesn't return anything, which this method does. The problem is that `wait_task_rescheduled` isn't typed so we get an `Any` as return type, which breaks us having specified the method to return `ReceiveType`.

When I looked into `trio-typing` it explicitly says that `wait_task_rescheduled` returns `Any`
```python
async def wait_task_rescheduled(
    abort_func: Callable[[Callable[[], NoReturn]], Abort]
) -> Any: ...
```
https://github.com/python-trio/trio-typing/blob/d5835f7065f58ebdd95ca27b09d1d5cbe686b8b8/trio-stubs/lowlevel.pyi#L131-L133

but I didn't manage to figure out whether that's a limitation in the typing, or what's up with that. (as I admitted in my first post).

trio/_channel.py
jakkdl: As I said in my first post I just assumed it'd be the same as the abc, so thanks for double checking it :)

trio/_channel.py
A5rocks: We could always just do `_: object` :^)

(in all seriousness I'll take a look and see eventually:tm:)

trio/_channel.py
A5rocks: We could probably just move the `return` to a new line. Let's verify with `wait_task_rescheduled` first though, maybe? Perhaps this is old code.

A5rocks:
cc @agronholm as you've had some opinions on types before

agronholm:
The previous typing PR used `from __future__ import annotations`, so I see no reason to not do that here too. That lets you modernize the annotations too by not using deprecated types from the `typing` module.

trio/_channel.py
agronholm: Annoyingly the latest release of mypy does not support `Self` :frowning_face: 
Support is in, it just hasn't been included in a release yet.

A5rocks: Unfortunate! mypy v1.0 is right around the corner so maybe this PR could be blocked on that? Actually I don't know how soon that will be, maybe the `SelfT` hack would be best for now and we can make a PR cleaning it up soon :P

trio/_channel.py
jakkdl: Ye, if the type doesn't get passed through then this code raises no errors on the assignment to bar:
```python
foo: MemorySendChannel[int] = ...
bar: SendChannel[str] = foo
```

trio/_channel.py
jakkdl: Typed it as `SelfT` and added a comment saying it's temporary until mypy releases support for it.

trio/_channel.py
jakkdl: Moved it to a new line, but yeah `wait_task_rescheduled` need to be figured out.

jakkdl:
Pushed a commit with updates from the reviews, though I got stuck trying to resolve `open_memory_channel`. It currently complains

```python
trio/_channel.py:100: error: Incompatible return value type
(got "Tuple[MemorySendChannel[SendType], MemoryReceiveChannel[ReceiveType]]",
expected "Tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]")  [return-value]
```

and I'm in general quite confused how to write that the function, with the `@generic_function` decorator is expected to take one `type`, which is passed to `MemoryChannelState`, and then we're to return the classes typed with two other typevars, presumably bound to the `type` parameter passed, but with different `covariant`/`contravariant`ness. So if anybody has an idea how to cram all that into the function header (or class header, see below), I would love to be enlightened.

I looked at trio-typing, and they [resort to making `open_memory_channel` a class:](
https://github.com/python-trio/trio-typing/blob/2bc84bab1d7f07e85e7d6f9fc95237c347cdcf55/trio-stubs/__init__.pyi#L283-L288)
```python
# written as a class so you can say open_memory_channel[int](5)
class open_memory_channel(Tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]):
    def __new__(  # type: ignore[misc]  # "must return a subtype"
        cls, max_buffer_size: float
    ) -> Tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]: ...
    def __init__(self, max_buffer_size: float): ...
```
which looks kinda hackish, but maybe is fine? And they don't appear to use `@generic_function` anywhere in `trio-typing`.

trio/_channel.py
```diff
@@ -109,28 +138,30 @@ def statistics(self):
 
 
 @attr.s(eq=False, repr=False)
-class MemorySendChannel(SendChannel, metaclass=NoPublicConstructor):
-    _state = attr.ib()
-    _closed = attr.ib(default=False)
+class MemorySendChannel(
+    SendChannel[SendType], Generic[SendType], metaclass=NoPublicConstructor
```
A5rocks, line 142: I don't think the `Generic` here is necessary, FWIW (same for `MemoryReceiveChannel`)

A5rocks:
> Pushed a commit with updates from the reviews, though I got stuck trying to resolve `open_memory_channel`. It currently complains
> 
> ```python
> trio/_channel.py:100: error: Incompatible return value type
> (got "Tuple[MemorySendChannel[SendType], MemoryReceiveChannel[ReceiveType]]",
> expected "Tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]")  [return-value]
> ```
> 
> and I'm in general quite confused how to write that the function, with the `@generic_function` decorator it's expected to take one `type`, which is passed to `MemoryChannelState`, and then we're to return the channel classes typed with two other typevars, presumably bound to the `type` parameter passed, but with different `covariant`/`contravariant`ness. So if anybody has an idea how to cram all that into the function header (or class header, see below), I would love to be enlightened.

I think this is because of `NoPublicConstructor`'s `_create` function. Not sure how to fix it, but hopefully that makes some stuff clearer.

jakkdl:
Ah, that made me think of the super simple solution!
```python
    return (
        MemorySendChannel[T]._create(state),
        MemoryReceiveChannel[T]._create(state),
    )
```

I should definitely write some test cases to check that it actually works though.

agronholm:
> Ah, that made me think of the super simple solution!
> 
> ```python
>     return (
>         MemorySendChannel[T]._create(state),
>         MemoryReceiveChannel[T]._create(state),
>     )
> ```
> 
> I should definitely write some test cases to check that stuff actually works though.

This requires Python 3.8 at minimum, just FYI.

trio/_channel.py
jakkdl: ah of course, thanks!

jakkdl:
> > Ah, that made me think of the super simple solution!
> > ```python
> >     return (
> >         MemorySendChannel[T]._create(state),
> >         MemoryReceiveChannel[T]._create(state),
> >     )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I should definitely write some test cases to check that stuff actually works though.
> 
> This requires Python 3.8 at minimum, just FYI.

ah, shoot

agronholm:
> > > Ah, that made me think of the super simple solution!
> > > ```python
> > >     return (
> > >         MemorySendChannel[T]._create(state),
> > >         MemoryReceiveChannel[T]._create(state),
> > >     )
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I should definitely write some test cases to check that stuff actually works though.
> > 
> > 
> > This requires Python 3.8 at minimum, just FYI.
> 
> ah, shoot

Actually, no. The class level `__getitem__()` method was in fact introduced in Python 3.7. It's just the `collections.abc` module that gained such support in 3.8.

jakkdl:
Yay!

Replacing the `open_memory_channel` definition with the class-based one won't work until 3.9+, but I realized I could get the best of both worlds with an `if TYPE_CHECKING` guard. (I first thought I could write a `.pyi` only for that function, but that'd override the whole file).

I also wrote some tests using https://github.com/davidfritzsche/pytest-mypy-testing to check for expected mypy error outputs, but it's kinda broken and unmaintained, and I failed to get it working with skipping on earlier versions, so not gonna try and get that working in the CI. But I did get it passing locally at least. It might be worth trying to get trio-typings testing infrastructure replicated here.

A5rocks:
According to the docs, `wait_task_rescheduled` returns whatever `trio.lowlevel.reschedule` is passed. I think `Any` (or `object` but pragmatism > type safety) would be best.

Maybe we should brainstorm a nicer interface for this? (we could call it a one-shot channel /s :^) But we should probably just make an issue instead of making on the spot.

trio/_channel.py
```diff
@@ -68,36 +97,56 @@ def open_memory_channel(max_buffer_size):
...
+        MemoryReceiveChannel[T]._create(state),
     )
 
 
+# This workaround requires python3.9+, once older python versions are not supported
+# or there's a better way of achieving type-checking on a generic factory function,
+# it could replace the normal function header
+if TYPE_CHECKING:
+    # written as a class so you can say open_memory_channel[int](5)
+    class open_memory_channel(tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]):
```
A5rocks, line 112: CI is complaining about this, heh. We can just `type: ignore` it because we know CI is wrong here (it's not run at runtime).

trio/_channel.py
```diff
@@ -306,15 +361,19 @@ async def receive(self):
...
             self._tasks.remove(task)
             del self._state.receive_tasks[task]
             return trio.lowlevel.Abort.SUCCEEDED
 
-        return await trio.lowlevel.wait_task_rescheduled(abort_fn)
+        result = await trio.lowlevel.wait_task_rescheduled(abort_fn)
+
+        # wait_task_rescheduled is set to return Any currently, but maybe it's
+        # guaranteed to return ReceiveType (or get aborted)?
+        return result  # type: ignore
```
A5rocks, line 373: Looking at the code: it's not *guaranteed* to return `ReceiveType` (you could always call `reschedule` yourself and ruin trio's day...) but it's close enough: . I think the comment should be updated the reflect this, but having the `type: ignore` is aok. (Also I misunderstood earlier -- we don't need to move the `return` to a new line, I had thought `wait_task_rescheduled` just wouldn't return)

https://github.com/python-trio/trio/blob/ca850f67c84c06e79a8d5ae6ed7b24a000e8d024/trio/_channel.py#L146

trio/_core/_traps.py
```diff
@@ -64,7 +65,10 @@ class WaitTaskRescheduled:
     abort_func = attr.ib()
 
 
-async def wait_task_rescheduled(abort_func):
+RaiseCancelT = Callable[[], NoReturn]  # TypeAlias
```
A5rocks, line 68: I'm not sure about this naming or having a type alias here. But it's your call! The type alias does seem useful, even if I'm not the biggest fan of the name.

trio/_core/_traps.py
```diff
@@ -64,7 +65,10 @@ class WaitTaskRescheduled:
...
+
+# Can this function be retyped to return something other than Any?
```
A5rocks, line 70: Discussed: if you agree with my reasoning, then the answer seems to be no, not really.

trio/lowlevel.py
```diff
@@ -16,6 +16,7 @@
 from ._core import (
     cancel_shielded_checkpoint,
     Abort,
+    RaiseCancelT,
```
A5rocks, line 19: This import seems extraneous?

A5rocks (changes requested):
A couple last comments, I think everything else is done.

trio/_channel.py
jakkdl: That's because the check is run on 3.8 ... which seems kinda old by this point.
It's easily fixed by using `Tuple`, so I'll just do that until the check is run on newer stuff.

trio/_core/_traps.py
jakkdl: Feels like it might be possible to make `Task` generic and send around enough types that it all works out, but that's definitely out of scope for this PR.

trio/lowlevel.py
jakkdl: Just removing it from there makes `test_core_is_properly_reexported` angry, so it's mostly a question of whether an internal `TypeVar` should be exposed in the API or not. It's not incredibly complex or anything, but might be of use to users? Unsure, I made it internal now but easy to change back.

jakkdl:
> According to the docs, `wait_task_rescheduled` returns whatever `trio.lowlevel.reschedule` is passed. I think `Any` (or `object` but pragmatism > type safety) would be best.
> 
> Maybe we should brainstorm a nicer interface for this? (we could call it a one-shot channel /s :^) But we should probably just make an issue instead of making one on the spot.

This is definitely above my paygrade in terms of experience or opinions on trio, so a separate issue to evaluate it sounds good to me if you want to tackle it.

trio/lowlevel.py
A5rocks: Ah didn't notice it would make a test angry, nvm then. If we're going to make a type alias we might as well go all the way and make it public

A5rocks (approved):
Looks good, someone else should take a look and merge if it looks good to them too.

Zac-HD (approved):

jakkdl:
:rocket: 
A big thank you to @A5rocks for the very helpful and detailed reviews! :rocket:

PR #2574, opened 2023-02-21
Ensure that every new Trio task returns 'trio' from sniffio.current_async_library()

oremanj:
Previously the context in which `start_soon()` was called would be inherited, which is a problem when doing cross-library calls.


richardsheridan:
#2462 contains these changes!

PR #2587, opened 2023-03-05
Fix CI failures on Ubuntu

richardsheridan:


.github/workflows/ci.yml
```diff
@@ -106,9 +106,7 @@ jobs:
       }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
-        with:
-          ref: ${{ github.event.pull_request.head.ref  }}
```
A5rocks, line 111: When I check why this happened, it seems I added it (sorry!) in order to fix something that made committing not work in response to dependabot PRs... does `actions/checkout@v3` fix this?

A5rocks (approved):
I'm feeling lucky!

PR #2590, opened 2023-03-05
split highly permissioned dependabot autoformatter into separate job

richardsheridan:
We don't need every ubuntu job to have high permissions or checkout the head ref, only those issued by dependabot.

Unfortunately, we can't know if it fixes dependabot until merging!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,12 +71,6 @@ jobs:
     name: 'Ubuntu (${{ matrix.python }}${{ matrix.extra_name }})'
     timeout-minutes: 10
     runs-on: 'ubuntu-latest'
-    # https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#changing-github_token-permissions
-    permissions:
-      pull-requests: write
-      issues: write
-      repository-projects: write
-      contents: write
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -126,9 +120,33 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Check formatting
```
A5rocks, line 144: We can probably combine this with the next check now that it's not the normal check formatting job

A5rocks (changes requested):
~~We still want normal people's PRs to have formatting checked (afaict this pr removes that?) -- just the autoformat thing is only for dependabot.~~

Ignore

.github/workflows/ci.yml
```diff
@@ -126,9 +120,33 @@ jobs:
...
+        run: |
+          python -m pip install -r test-requirements.txt
+          source check.sh
```
A5rocks, line 147: Don't forget to add the `CHECK_FORMATTING` env var

A5rocks, line 147:
```suggestion
          ./check.sh
```

I believe?

.github/workflows/ci.yml
A5rocks: Ah I see. This is fine then.

A5rocks (approved):
Feeling lucky (again!)

PR #2592, opened 2023-03-05
Drop dependabot in favor of automated pip-tools

### LINKED ISSUE #2594: Dependency management
The current setup of dependabot is intensively manual and has some negative interactions with pip-tools. I've created 2 PRs with opposite solutions to this problem: #2593 maxes out dependabot usage and automates the PR merges, but sacrifices the precise output of pip-compile, whereas #2592 drops dependabot entirely and does a periodic (and also manually triggerable) pip-compile bump.

Personally I would favor dropping dependabot mainly because of the fewer commits, but I recognize that monthly mass updates are marginally more likely to have difficult-to-diagnose breakage from bad interactions.

richardsheridan:
Closes #2594.


richardsheridan:
One subtle quirk of this is that we have to manually delete branches after they've been merged.

richardsheridan:
@A5rocks this primarily affects you at the moment so please review and merge when you can, as this PR seems to be the preferred option in #2594

ci.sh
```diff
@@ -115,7 +115,7 @@ else
         # when installing, and then running 'certmgr.msc' and exporting the
         # certificate. See:
         #    http://www.migee.com/2010/09/24/solution-for-unattendedsilent-installs-and-would-you-like-to-install-this-device-software/
-        certutil -addstore "TrustedPublisher" .github/workflows/astrill-codesigning-cert.cer
+        certutil -addstore "TrustedPublisher" trio/tests/astrill-codesigning-cert.cer
```
A5rocks, line 118: Hm, I'm not too sure of this rename. It isn't getting out into the sdist right?

A5rocks (commented):
I think I'm ok with this, but just to confirm: when a PR fails to automerge, if we don't merge it that day (? I can't read cron, I presume this runs once daily?) CI will remake it?

ci.sh
richardsheridan: I suppose we could prune it from MANIFEST.in, but at least [one person thinks it aught to go in there](https://github.com/python-trio/trio/pull/2541#issuecomment-1400712472)! Do you have a specific suggestion for a better home for it if we are not going to include it in the sdist?

richardsheridan:
> I think I'm ok with this, but just to confirm: when a PR fails to automerge, if we don't merge it that day (? I can't read cron, I presume this runs once daily?) CI will remake it?

It's once a month ([best webtool ever](https://crontab.guru/#0_0_1_*_*)). The branch name is distinguished by the [base commit SHA](https://github.com/python-trio/trio/blob/74cef56b9c5955f337ce39f613868820981768e1/.github/workflows/autodeps.yml#L49) so if it so happens that the automerge fails AND no other commits are pushed over the month, then the PR branch will be overwritten by force-push. Otherwise, the bot will make a new branch and PR.

ci.sh
A5rocks: I don't have much experience with these so maybe it would be best to include it in the sdist...

But that feels weird. I would rather a certificate like this be in a top-level `assets` directory (maybe rename the top-level `logo` directory?).

A5rocks: Just to be clear: in my mind this certificate is just like a pytest plugin -- it may be required to run (some) tests but it certainly shouldn't weigh everyone who installs the package down.

A5rocks:
Once a month sounds reasonable! Alright then.

ci.sh
richardsheridan: I just built a wheel and sdist on my machine and they do not pick up the cert, so I think this question can be deferred to #2541

A5rocks: Alright. I now realize I could have checked CI logs but... whatever.

A5rocks (approved):
Looks alright to me! I'd like one other person skim through this before they press merge but I think this will work nicely.

njsmith:
I am excited about getting rid of dependabot spam!

I think that we'll need to switch to using a personal access token to create the PR, because otherwise Github will refuse to run CI on the PR: https://github.com/peter-evans/create-pull-request/issues/48

> One subtle quirk of this is that we have to manually delete branches after they've been merged.

I just flipped this setting on, which I think should take care of it?

image

richardsheridan:
> I am excited about getting rid of dependabot spam!
> 
> I think that we'll need to switch to using a personal access token to create the PR, because otherwise Github will refuse to run CI on the PR: [peter-evans/create-pull-request#48](https://github.com/peter-evans/create-pull-request/issues/48)

I think I did know this at some point... can you help me make that token? I think it requires admin access

> > One subtle quirk of this is that we have to manually delete branches after they've been merged.
> 
> I just flipped this setting on, which I think should take care of it?

Nice!

A5rocks:
> > I am excited about getting rid of dependabot spam!
> 
> > I think that we'll need to switch to using a personal access token to create the PR, because otherwise Github will refuse to run CI on the PR: https://github.com/peter-evans/create-pull-request/issues/48
> 
> I think I did know this at some point... can you help me make that token? I think it requires admin access

I found this just now and it seems useful; here's a list of what we could do: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#workarounds-to-trigger-further-workflow-runs

Choice options:
 - close/reopen the PRs ;-)
 - SSH "deploy keys" (we run the actions on push assuming the branch is on this repository, I believe?)
 - Personal Access Tokens (on a machine account?)
 - ... It seems we can just use the existing trio bot to make a token for maximum coolness? But that might leak secrets somehow which would suck.

A5rocks:
Hi! Just coming back here to bump this PR again. I wouldn't mind having to close/reopen our automated PRs if necessary, if we can't decide on what to do.

richardsheridan:
Right, you can merge it today and it should work just like that. In the future, someone with admin privileges can create an appropriately scoped personal token as in https://github.com/python-trio/trio/issues/2594#issuecomment-1480262890 and simply make a PR to change the `GH_TOKEN` argument in the workflow to point to it and it should become fully automatic.

njsmith:
Unfortunately, I think if we put a PAT in as a secret on this repo, then
that can be trivially extracted by anyone with commit privileges and then
used to impersonate the original person, right?

A5rocks:
I haven't wrapped my head around how secrets function in GitHub Actions (yet) but that certainly sounds right. I'd suggest some sort of machine account? The account making the PRs needs no permissions at all, after all. It could always make a PR from a fork (though that would complicate things).

I'll merge this now cause there's a fix no matter how cludgey and we can figure out how to do this at a future point.

PR #2593, opened 2023-03-05
Go all in dependabot and only manually run pip-tools

### LINKED ISSUE #2594: Dependency management

richardsheridan:
Closes #2594.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
 - package-ecosystem: pip
   directory: "/"
   schedule:
-    interval: daily
+    interval: "weekly"
   open-pull-requests-limit: 10
   allow:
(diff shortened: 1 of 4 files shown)


A5rocks:
Closed in favor of https://github.com/python-trio/trio/pull/2592.

PR #2595, opened 2023-03-05
Rework packaging to use Poetry instead.

Fuyukai:
This does several things:

1. Kill ``setup.py`` and replace it with a Poetry-based ``pyproject.toml``.
2. Move the code into the src-layout instead.
3. Separate the tests out into their own directory and package.

Questions I invisige might be asked:

Q: Why Poetry and not <insert my personal preference here>?
A: I only know Poetry and this is my PR.

Q: Why ``src`` layout?
A: It has less shenanigans with PYTHONPATH. Also, it's better to do it now and break history (well, git can see the renames but github, in its infinite usability, cannot?) than do a bunch of work on the project and eventually move it over in the future.

Q: Why separate out tests?
A: I couldn't find an answer to "how to make pytest use conftest for installed package tests" in 5 minutes so I just spent the effort to separate them out instead. Also, it annoys me on some level that every trio download includes a few extra kilobytes of tests for no reason.

TODO:

- [ ] Make sure CI works again (I tried, but CI is CI)
- [ ] Quash dissent


Fuyukai:
Likely not going to happen, and this PR predates several major organisational changes.

PR #2596, opened 2023-03-05
Document that new tasks from `nursery.start_soon()` may start in any order

### LINKED ISSUE #970: Document that Nursery.start_soon starts tasks in random order

Zac-HD:
Closes #970, rebased and refined from #972.

Also replaces two unrelated `’` 'smart' singlequotes with `'` ascii singlequotes, since there was one in the docstring I edited.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1022,18 +1022,17 @@ class Nursery(metaclass=NoPublicConstructor):
     def start_soon(self, async_fn, *args, name=None):
         """Creates a child task, scheduling ``await async_fn(*args)``.
 
-        This and :meth:`start` are the two fundamental methods for
+        If you want to run a function and immediately wait for its result,
+        then you don't need a nursery; just use ``await async_fn(*args)``.
+        If you want to wait for the task to initialize itself before
+        continuing, see :meth:`start`, the other fundamental method for
         creating concurrent tasks in Trio.
 
         Note that this is *not* an async function and you don't use await
         when calling it. It sets up the new task, but then returns
-        immediately, *before* it has a chance to run. The new task won’t
-        actually get a chance to do anything until some later point when
-        you execute a checkpoint and the scheduler decides to run it.
-        If you want to run a function and immediately wait for its result,
-        then you don't need a nursery; just use ``await async_fn(*args)``.
-        If you want to wait for the task to initialize itself before
-        continuing, see :meth:`start`.
+        immediately, *before* the new task has a chance to do anything.
+        New tasks may start running in any order, and at any checkpoint the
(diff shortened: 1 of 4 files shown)


trio/_core/_run.py
```diff
@@ -1022,18 +1022,17 @@ def aborted(raise_cancel):
...
-        immediately, *before* it has a chance to run. The new task won’t
-        actually get a chance to do anything until some later point when
-        you execute a checkpoint and the scheduler decides to run it.
-        If you want to run a function and immediately wait for its result,
-        then you don't need a nursery; just use ``await async_fn(*args)``.
-        If you want to wait for the task to initialize itself before
-        continuing, see :meth:`start`.
+        immediately, *before* the new task has a chance to do anything.
+        New tasks may start running in any order, and at any checkpoint the
+        scheduler chooses - at latest when the nursery is waiting to exit.
```
Zac-HD, line 1035: I believe that in practice we currently start running `start_soon()`'d tasks at the _next_ checkpoint in random order, but prefer not to guarantee that behavior.

A5rocks: I'm not too sure about this decision: it's both a quite useful property to rely on, I'd expect, and I can't quite imagine an implementation that would require losing that property. I suppose it's up to you and we could introduce that guarantee at any time in the future if we need to.

A5rocks (approved):
Wording looks fine. I have a comment but it's quite extraneous. I'd prefer someone else read through this PR before merging however!

trio/_core/_run.py
njsmith: We might want to implement some strategy for skipping no-op checkpoints as an optimization, which would mean that you couldn't necessarily predict where the next "real" checkpoint is. And in general, any code that relies on counting checkpoint is inherently very fragile, so we want to discourage people from going down that road in general. So I'm fine with the lack-of-guarantee here.

PR #2602, opened 2023-03-11
add trio-level timeout to test_dtls.test_handshake_over_terrible_network

richardsheridan:
Hopefully will reveal insight into intermittent timeout failures in CI, which are not isolated to pypy:

https://github.com/python-trio/trio/actions/runs/4392870655/jobs/7692906224#step:4:985


richardsheridan:
insight: autojump clock makes this not work

PR #2603, opened 2023-03-11
use autojump_threshold in test_handshake_over_terrible_network

richardsheridan:
observation: any autojump threshold over 0 takes MUCH longer than 0
hypothesis: 10 second timeout cscope at the end of test is spuriously cancelling attempts on slow machines
experiment: set a minimal threshold and scale back # of handshakes to finish in a reasonable time, check if CI timeouts are avoided

### DIFF
--- a/trio/tests/test_dtls.py
+++ b/trio/tests/test_dtls.py
@@ -103,10 +103,12 @@ async def test_smoke(ipv6):
 @slow
 async def test_handshake_over_terrible_network(autojump_clock):
     # PyPy is not fast enough
-    HANDSHAKES = 500 if sys.implementation.name == "pypy" else 1000
+    HANDSHAKES = 50 if sys.implementation.name == "pypy" else 100
     r = random.Random(0)
     fn = FakeNet()
     fn.enable()
+    # avoid spurious timeouts on slow machines
+    autojump_clock.autojump_threshold = 0.001
 
     async with dtls_echo_server() as (_, address):
         async with trio.open_nursery() as nursery:
--- a/trio/tests/test_dtls.py
+++ b/trio/tests/test_dtls.py
@@ -3,7 +3,6 @@ import trio
 import trio.testing
 from trio import DTLSEndpoint
 import random
-import sys
 import attr
 from contextlib import asynccontextmanager
 from itertools import count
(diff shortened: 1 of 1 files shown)


richardsheridan:
@A5rocks I think this puts the cherry on top of #2536 what do you think?  I might even try a version that makes `HANDSHAKES` consistent across platforms again.

A5rocks:
I'm not sure I agree with the reasoning in the description of the PR (the 10 second cancel scope should solely be skipping sends?). I think I agree with that changes though, just the description needs to be rethought!

I would make `HANDSHAKES` consistent across platforms if it works.

richardsheridan:
I suppose I should put a fuller explanation here in the PR. I just learned about this stuff today as well after all!

The `autojump_clock` pytest fixture installs a `MockClock` that transforms any timeouts/fail_afters/deadlines into immediate wakeups. It is one step removed from actually calling `cancel_scope.cancel()` right before your task goes to sleep. The interaction of this tool with this test has the effect of making magnitude of a timeout argument irrelevant. 

My thinking is that the send is effectively "nonblocking" and immediately reschedules itself, whereas the receive has to wait long enough for the system (kernel? python?) to get around to handling the sent payload. This handling process may be just slow enough on an overloaded CI machine to outlast one cycle of the run loop that applies for the default setting, but not the 1 ms `autojump_threshold` (or for that matter a 1 microsecond threshold that I tested locally).

Finally, it does seem to work, I've rerun CI on [55abf37](https://github.com/python-trio/trio/pull/2603/commits/55abf377a88b2dc49b613635b547d5147257a2ac) 4 times and [never had a pypy timeout failure](https://github.com/python-trio/trio/actions/runs/4393554310). Take another look after I unify `HANDSHAKES`, so can merge this and stabilize CI! we can always fix any wording or explanatory comments in a separate issue.

A5rocks (approved):
That does seem like a pretty important issue! Alright, I'm on board. I'd rather wait for confirmation on whether this autojump + cancel scope issue might impact elsewhere (from maybe njs?), though that's not too important.

A5rocks:
I shouldn't have left this out to dry for so long, sorry!

Issue #2625, opened 2023-03-31
Add `pyright --verifytypes` to CI & checks


jakkdl:
In order to be able to eventually retire trio-typing (see https://github.com/python-trio/trio-typing), track progress on #543 (and once type complete, make sure that we stay that way) and as a minor bonus give a check on docstrings, I suggest adding
`pyright --verifytypes` to the checks in CI, that at a beginning is required only to go up from current state, and when we've achieved 100% type completeness will then require not dropping below that.
See https://github.com/python-trio/trio/pull/2623 where I reinvented the wheel in a quite clunky way and some discussion around it, https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness for the definition of type completeness, and https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#verifying-type-completeness for documentation on `--verifytypes`.

There's unfortunately a couple blockers before we're able to use it.
1. [x] Imports in `__init_.py`, `socket.py`, `abc.py`, `from_thread.py`, `lowlevel.py`, `socket.py` and `to_thread.py` are not visible to `--verifytypes`. Also related to https://github.com/python-trio/trio/issues/542
   * See https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface for official docs on why this is the case, and https://github.com/microsoft/pyright/issues/4869 for a quick explainer
   * I can come up with the following alternatives:
     1. Define `__all__`, see discussion in https://github.com/python-trio/trio/pull/594#issuecomment-411313454
     2. Change the imports to `from X import A as A` - a bit verbose, but simple
     3. Change the imports to `from X import *` - I'm not sure if we're currently exporting everything, so this might expose more symbols unless they are/get named with a leading underscore.
     4. Overhaul the current file structure, creating subdirectories with their own `__init__`'s for the different submodules, and remove leading underscores from file names. (this has likely been discussed in #542 and other places and there's presumably some decent logic for the current layout?).
2. [x] trio.tests is a public module and is included as a submodule that needs verifying. See https://github.com/python-trio/trio/issues/274, suggestions there include renaming `trio/tests` to `trio/_tests`, moving it out to a separate top-level `tests/`, or removing `trio/tests/__init__.py`.
3. [x] How to actually check that the type completeness score doesn't go down.
[...]

PR #2628, opened 2023-04-05
move **/tests/ to **/_tests/

jakkdl:
Simply renames the directories trio/tests and trio/_core/tests, and updates references to them.

Addresses some (or all?) problems in #274 
Fixes the second task in #2625 

@njsmith - especially wrt to considerations for #274


jakkdl:
ah, `trio/_tests/conftest.py` is not parsed for CLI flags (i.e. `--runslow`) before running. Could e.g. solve that by moving it into `trio/`, but that'd make it public.
There's probably other ways of achieving it ... but it's surprisingly tricky, and I've yet to manage something else decent

A5rocks:
I poked at this a bit and adding `-p "trio._tests.conftest"` seems to get further -- until it complains about adding the plugin twice (second time as normal part of test discovery).

Like the comment in the file says, we can essentially split `trio._tests.conftest` (which contains all of `trio._core._tests.conftest`) into a plugin and that'll work fine.

jakkdl:
> I poked at this a bit and adding `-p "trio._tests.conftest"` seems to get further -- until it complains about adding the plugin twice (second time as normal part of test discovery).
> 
> Like the comment in the file says, we can essentially put `trio._tests.conftest` (which contains all of `trio._core._tests.conftest`) into a plugin and that'll work fine.

huh, I don't get that far:
```python
  File "/home/h/Git/trio/trio/_tests/conftest.py", line 10, in <module>
    from ..testing import trio_test, MockClock
  File "/home/h/Git/trio/trio/testing/__init__.py", line 7, in <module>
    from ._sequencer import Sequencer
  File "/home/h/Git/trio/trio/testing/_sequencer.py", line 11, in <module>
    from .. import Event
ImportError: Error importing plugin "trio._tests.conftest": cannot import name 'Event' from 'trio' (unknown location)
```
though maybe I've messed up my environment somehow. Feel free to push a commit that makes it a plugin if you can get it to work.

jakkdl:
I'm still having some env issues with running the tests myself sometimes, but they seem to pass CI - so it should be fine on a clean install.

Added `trio/tests.py` and guessed how it was supposed to look :)

A5rocks (commented):
Presumably doing `import trio.tests` will work as a test but I don't think we need one.

trio/tests.py
```diff
@@ -0,0 +1,10 @@
+from ._deprecate import warn_deprecated
+import trio
+from ._tests import *
```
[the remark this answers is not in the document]
jakkdl: I think the best way of handling that is by seeing that they're not listed in `pyright --verifytypes` in #2629 - and not do anything about it in this PR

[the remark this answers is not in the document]
jakkdl: a test for `import trio.tests` would increase code coverage I suppose

jakkdl:
Cleaned up and added an import+deprecation test.
I'm a bit confused about codecov wrt `trio.testing`, but I don't think it's a problem?

Formatting seems to have hit a snag with ... something unrelated to this PR.
and 3.12 continues to fail like usual.

jakkdl:
@A5rocks this one is also ready for review

A5rocks (commented):
Cursory review comments, I might find something else in a second review!

A5rocks:
FYI master fixes the CI runs, so just merge that in!

trio/tests.py
```diff
@@ -0,0 +1,36 @@
...
+
+# This won't give deprecation warning on import, but will give a warning on use of any
+# attribute in tests, and static analysis tools will also not see any content inside.
+class TestsDeprecationWrapper:
+    __name__ = "trio.tests"
+
+    def __getattr__(self, attr: str) -> Any:
+        warn_deprecated(
+            f"trio.tests.{attr}",
+            "0.23.0",
```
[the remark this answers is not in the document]
jakkdl: Sounds good, I just copy-pasted another deprecation warning without considering the versioning :)

jakkdl:
rebased on top of master, and fixed review comments

A5rocks:
Sounds good, we want to redo codecov a bit too since it doesn't work with like Windows and whatever ATM. So feel free to change things slightly if necessary!

jakkdl:
uh, looks like codecov isn't turning up as a check at all now. I'll just revert it and we can put a pin into running with `codecov -m` for the upcoming codecov changes in a separate PR. I'd like this merged so it stops blocking #2629

trio/tests.py
```diff
@@ -0,0 +1,36 @@
+import importlib
+import sys
+from typing import Any
+
+# don't ask which of the _tests imports is necessary
+import trio._tests
+
+from . import _tests
```
A5rocks, line 8: I'm a bit confused by this, won't the namespaces shadow and so the first is useless?

trio/tests.py
```diff
@@ -0,0 +1,36 @@
...
+from ._deprecate import warn_deprecated
+from ._tests import *
```
A5rocks, line 10: ... I don't think this is necessary anymore?

A5rocks (commented):
Looks mostly good, I'd like @njsmith to weigh in cause I'm not too used to changes like this!

jakkdl:
Fixed review comments, removing unnecessary imports. And reading the PR after having forgotten what the code does unsurprisingly lead to some improved comments and tests :innocent:

trio/tests.py
```diff
@@ -0,0 +1,31 @@
...
+        # needed to access e.g. trio._tests.tools, although pytest doesn't need it
+        if not hasattr(_tests, attr):  # pragma: no cover
+            importlib.import_module(f"trio._tests.{attr}", "trio._tests")
+            return attr
+
+        return getattr(_tests, attr)
+
+
+# https://stackoverflow.com/questions/2447353/getattr-on-a-module
+sys.modules[__name__] = TestsDeprecationWrapper()  # type: ignore[assignment]
```
njsmith, line 31: I'm not sure we really need to be this careful about backcompat here -- *surely* people aren't *actually* using `trio.tests`, right? -- but I guess it's harmless.

Though... I guess we actually want to get rid of `trio.tests` entirely, so maybe we should issue the warning on import? If someone is importing it but not using it (for some reason), then they won't get a warning, but they'll still get broken in the future when we delete this file.

njsmith (approved):

trio/tests.py
jakkdl: ah, took me a minute but figured out how to get it to warn on import as well. Only thing it doesn't warn on now that would break is `import trio; assert trio.tests` - i.e. access of the module, but without importing it, and not accessing any of it's attributes. (I tried changing from using `__getattr__` to `__getattribute__` but that made no difference in this case). There's probably some way of addressing that, but now we're getting into very weird cases.

jakkdl:
removed the test import from `__init__.py`, so it's now also invisible to static analysis. Added a corresponding exception for it being a missing symbols in the static analysis tests.

jakkdl:
@A5rocks any reservations about merging this?

A5rocks (approved):

PR #2629, opened 2023-04-05
change imports to `from Y import X as X` for compatibility with `pyright --verifytypes`

jakkdl:
Completes the first task in #2625 

Should probably not be merged before a test for `test_exports.py` is written that checks that the exports seen by verifytypes is the same as the other static analyzers, and/or the CI check for type completeness.

remaining TODO:
* [x] make sure `py.typed` exists in `tests.test_exports.test_static_tool...`
* [x] save the new results from verifytypes in git in CI
* [x] get pyright to play along by setting `PYRIGHT_FORCE_PYTHON_VERSION` in all the places.


A5rocks:
Funnily enough I also did this yesterday, though I see I did `trio/socket` and you didn't. (I'm not too sure if the `as` is necessary there TBF).

Also we re-export some symbols via `from A import X as Y` and I'm not sure how to make that work here.

URL: https://github.com/A5rocks/trio/commit/37bb8ede6374d9d5370c14c76fa80ac03d4cda04

jakkdl:
> Funnily enough I also did this yesterday, though I see I did `trio/socket` and you didn't. (I'm not too sure if the `as` is necessary there TBF).
> 
> Also we re-export some symbols via `from A import X as Y` and I'm not sure how to make that work here.
> 
> URL: https://github.com/A5rocks/trio/commit/37bb8ede6374d9d5370c14c76fa80ac03d4cda04

haha, good that it's mostly just some regex search&replace.

The `as` is only needed when you're re-exporting from a private module, so I intentionally skipped all the ones in socket - and one or two in the other files iirc. But a `test_export` test should settle it anyhow

trio/tests/verify_types.json
A5rocks, line 1: Perhaps mark this as autogenerated? In `.gitattributes`, that is.

Also, we should *probably* not bundle such a huge file in our final package, is there a better place for this?

trio/tests/test_exports.py
```diff
@@ -96,6 +97,31 @@ def no_underscores(symbols):
         script = jedi.Script(f"import {modname}; {modname}.")
         completions = script.complete()
         static_names = no_underscores(c.name for c in completions)
+
+    elif tool == "pyright_verifytypes":
+        # it's maybe possible to run this through pyright internals
```
A5rocks, line 102: FYI, pyright is written in JS so we won't really be able to poke at its internals. The pyright python package is just a transparent wrapper around setting up a tiny node env and using it.

trio/tests/verify_types.json
jakkdl: After #2628 the size of the file should drop quite a bit, but one can also just not dump the individual errors for each symbol and just get the numbers - although

trio/tests/verify_types.json
```diff
@@ -0,0 +1,32 @@
...
+      "withAmbiguousType": 46,
+      "withKnownType": 174,
+      "withUnknownType": 651
+    },
+    "ignoreUnknownTypesFromImports": true,
+    "missingClassDocStringCount": 20,
+    "missingDefaultParamCount": 0,
+    "missingFunctionDocStringCount": 436,
+    "moduleName": "trio",
+    "moduleRootDirectory": "/home/h/Git/trio/.tox/py38-checksh/lib/python3.8/site-packages/trio",
```
A5rocks, line 21: Probably should be removed

trio/tests/verify_types.json
```diff
@@ -0,0 +1,32 @@
...
+    "otherSymbolCounts": {
+      "withAmbiguousType": 13,
+      "withKnownType": 222,
+      "withUnknownType": 293
+    },
+    "packageName": "trio",
+    "packageRootDirectory": "/home/h/Git/trio/.tox/py38-checksh/lib/python3.8/site-packages/trio",
+    "pyTypedPath": "/home/h/Git/trio/.tox/py38-checksh/lib/python3.8/site-packages/trio/py.typed"
```
A5rocks, line 29: Probably should be removed

jakkdl:
whee, finally managed to wrangle CI to pass
.
Only thing left now is getting CI to save the changes in `verify_types.json` after running `check_type_completeness` in in this repo, or another repo, or somewhere else.

[It seems there's a bunch of ways of doing that](https://github.com/orgs/community/discussions/25234) but idk what the setup with secrets and stuff is, and seems related to the discussion that went on in #2592 that stalled out when it got into complications around that.

jakkdl:
@A5rocks there's maybe some codecov fighting I can do, but shenanigans from windows and pytest plugin+testing make it somewhat of a lost cause. Otherwise this should be ready for review/merge

check.sh
```diff
@@ -37,6 +37,13 @@ if git status --porcelain | grep -q "requirements.txt"; then
     EXIT_STATUS=1
 fi
 
+python trio/_tests/check_type_completeness.py --overwrite-file || EXIT_STATUS=$?
```
A5rocks, line 40: Just a note that we might want to check that type completeness won't use the platform it currently runs on for e.g. filtering out `if sys.version_info < ...` stuff. Not required though, this just popped into mind while seeing this.

trio/__init__.py
```diff
@@ -13,86 +13,101 @@
 # This file pulls together the friendly public API, by re-exporting the more
 # innocuous bits of the _core API + the higher-level tools from trio/*.py.
 
+# Uses `from x import y as y` for compatibility with `pyright --verifytypes` (#2625)
+
 from ._version import __version__
```
A5rocks, line 18: `as __version__`?

trio/__init__.py
```diff
@@ -13,86 +13,101 @@
...
+    serve_ssl_over_tcp as serve_ssl_over_tcp,
 )
 
 from ._core._multierror import MultiError as _MultiError
 from ._core._multierror import NonBaseMultiError as _NonBaseMultiError
 
-from ._deprecate import TrioDeprecationWarning
+from ._deprecate import TrioDeprecationWarning as TrioDeprecationWarning
 
 # Submodules imported by default
```
A5rocks, line 112: Do submodules not need the `as ...` dance?

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python3
+# this file is not run as part of the tests, instead it's run standalone from check.sh
+import subprocess
+import json
+from pathlib import Path
+from typing import Any
+import sys
+import argparse
+
+RES_FILE = Path(__file__).parent / "verify_types.json"
```
A5rocks, line 10: res -> result?

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,138 @@
...
+        else:
+            strcurrent = str(current)
+            strlast = str(last)
+        print(
+            f"{key} has gone {'down' if current<last else 'up'} from {strlast} to {strcurrent}"
+        )
+
+    def check_zero(key):
+        if current_dict[key] != 0:
+            global failed
```
A5rocks, line 49: My personal style involves putting these right under the function definition.

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,138 @@
...
+        del current_result["summary"]["timeInSec"]
+
+        # don't fail on version diff so pyright updates can be automerged
+        del current_result["version"]
+
+        for key in (
+            # don't save huge file for now
+            "symbols",
+            "modules",
+            # don't save path
```
A5rocks, line 116: I know why this is here but it might be nice to include a short "because that varies from machine to machine" to lessen thinking about this later.

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,138 @@
...
+    current_dict = current_dict["exportedSymbolCounts"]
+    last_dict = last_dict["exportedSymbolCounts"]
+
+    for key in "withUnknownType", "withAmbiguousType":
+        check_less_than(key)
+    check_less_than("withKnownType", invert=True)
+
+    assert (
+        res.returncode != 0
+    ), "Fully type complete! Replace this script with running pyright --verifytypes directly in CI and checking exit code."
```
A5rocks, line 100: When you say replace this script, you do mean straight up just deleting this script right?

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,138 @@
...
+    for key in (
+        "missingFunctionDocStringCount",
+        "missingClassDocStringCount",
+        "missingDefaultParamCount",
+    ):
+        check_less_than(key)
+    check_less_than("completenessScore", invert=True)
+
+    current_dict = current_dict["exportedSymbolCounts"]
+    last_dict = last_dict["exportedSymbolCounts"]
```
A5rocks, line 92: nit: I think it might be more obvious to index off `last_result`/`current_result` vs off `last_dict`/`current_dict`.

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,138 @@
...
+import sys
+import argparse
+
+RES_FILE = Path(__file__).parent / "verify_types.json"
+failed = False
+
+
+def run_pyright():
+    return subprocess.run(
+        ["pyright", "--verifytypes=trio", "--outputjson", "--ignoreexternal"],
```
A5rocks, line 16: This looks cool, but why `--ignoreexternal`?

trio/_tests/check_type_completeness.py
A5rocks: Also maybe put the command in backticks and add the `--verifytypes=trio --ignoreexternal` thing? (if that's necessary)

trio/_tests/test_exports.py
```diff
@@ -118,18 +116,45 @@ def no_underscores(symbols):
...
+            # unclear why this one is required
+            if sys.platform == "win32" and sys.version_info[:2] == (3, 7):
+                static_names.add("IPPROTO_IPV6")
+
     else:  # pragma: no cover
         assert False
 
+    # remove py.typed file
+    if tool in ("mypy", "pyright_verifytypes") and not py_typed_exists:
+        py_typed_path.unlink()
```
A5rocks, line 156: Do you know why codecov is saying this line isn't covered by tests when the tests should be doing this?

trio/_tests/test_exports.py
```diff
@@ -118,18 +116,45 @@ def no_underscores(symbols):
...
+        end_index = res.stdout.find(b"Other referenced symbols: ", start_index)
+        assert start_index != -1, (res.stdout, res.stderr)
+        assert end_index != -1, res.stdout
+        static_names = {
+            x.strip().decode()[len(modname) + 1 :]
+            for x in res.stdout[start_index:end_index].split(b"\n")
+            if modname + "." in x.decode()
+        }
+
+        # don't require verifytypes to see reexported constants from socket
```
A5rocks, line 141: I disagree with this comment but if it's annoying we can bump this off

trio/_tests/test_exports.py
```diff
@@ -118,18 +116,45 @@ def no_underscores(symbols):
...
 
         # check that there were no errors (exit code 0), otherwise print the errors
-        assert res[2] == 0, res[0]
+        assert mypy_res[2] == 0, mypy_res[0]
         return
+    elif tool == "pyright_verifytypes":
+        import subprocess
+
+        res = subprocess.run(
+            ["pyright", f"--verifytypes={modname}", "--verbose"],
```
A5rocks, line 128: Does `--outputjson` not require we parse the output like this?

trio/_timeouts.py
A5rocks, line 1: Why the changes to this file?

trio/socket.py
```diff
@@ -6,6 +6,10 @@
 # here.
 # We still have some underscore names though but only a few.
 
+
+# Uses `from x import y as y` for compatibility with `pyright --verifytypes` (#2625)
+# It ignores CONSTANTS From socket though.
```
A5rocks, line 11: Hmm what does this mean? Also `From` -> `from`.

A5rocks (changes requested):
Minor comments, will be happy to merge once you fix. I'll try to be much more responsive with types-specific PRs!

check.sh
jakkdl: hmm. Maybe a reason to update the `check` environment to something newer than 3.8
I suppose a thorough check would run this on all supported versions, but that feels a bit overkill.

trio/__init__.py
jakkdl: huh, `pyright --verifytypes` seems to entirely ignore `__version__`. No matter how it's imported, or even if it's statically defined in the file. (but if I create a `__verison__` variable that one shows up with `--verbose`).

trio/__init__.py
jakkdl: They don't have a leading underscore so they're considered public by default https://microsoft.github.io/pyright/#/typed-libraries?id=library-interface

trio/_tests/check_type_completeness.py
jakkdl: > If the --verifytypes option is combined with --ignoreexternal, any incomplete types that are imported from other external packages are ignored. This allows library authors to focus on adding type annotations for the code that is directly under their control.

I'll add a comment (I didn't remember either :innocent:)

jakkdl: I'm not fully sure if it makes a difference, but could try removing it later on and potentially open PR's to any external libraries that affects us.

trio/_tests/test_exports.py
jakkdl: oops, yep. `mypy` has a `return` at the end of its `if`-branch so it never cleans up after it.

trio/_tests/check_type_completeness.py
jakkdl: yes, I'll clarify.

and will do

trio/_tests/check_type_completeness.py
jakkdl: even when they're accessing shared-state variables? Can def agree with putting the `print` below it, but makes sense to me to define the variables they'll use before defining the functions.

trio/_tests/check_type_completeness.py
jakkdl: yeah re-reading this now it's kinda unreadable. Should be better now

trio/_tests/check_type_completeness.py
jakkdl: Moved them out and got rid of shared-state variables

trio/_tests/test_exports.py
jakkdl: you can't combine `--outputjson` and `--verbose` (for some reason), and `--outputjson` won't include symbols without an error. So we have to do the dirty output parsing.
Adding a comment

trio/_tests/test_exports.py
jakkdl: wait no. hmm

trio/socket.py
jakkdl: see discussion in previous conversation.

fixed~

trio/__init__.py
jakkdl: see later conversation on trio.socket constants

trio/_timeouts.py
jakkdl: reset to state in main

trio/tests/verify_types.json
jakkdl: Actually yeah, not being able to gitdiff I think is a pretty big problem. When it gives an error that *something* has changed that's extremely unhelpful.
But it's not getting bundled in the package as it's not listed in `MANIFEST.in` - so it's only the git repo that grows by ~400kB

trio/_tests/test_exports.py
jakkdl: wait no, I think `--outputjson` does include symbols without error and I just got confused by the `trio.socket` shenanigans

check.sh
jakkdl: but I don't fully know how pytest handles that stuff internally, which is what matters.

jakkdl:
I have no clue why the `omit` isn't working in `.coveragerc`

trio/_tests/check_type_completeness.py
A5rocks: Ah sorry, I meant putting it above the if statement for this comment in particular! All's good though.

trio/__init__.py
jakkdl: Aha, found the documentation for it:
https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness
> Type annotations can be omitted in a few specific cases where the type is obvious from the context:
> ...
> The following module-level symbols do not require type annotations: `__all__`,`__author__`, `__copyright__`, `__email__`, `__license__`, `__title__`, `__uri__`, `__version__`.

trio/_tests/test_exports.py
jakkdl: > Type annotations can be omitted in a few specific cases where the type is obvious from the context:
> * ...
> * Constants that are assigned simple literal values (e.g. RED = '#F00' or MAX_TIMEOUT = 50 or room_temperature: Final = 20). A constant is a symbol that is assigned only once and is either annotated with Final or is named in all-caps. A constant that is not assigned a simple literal value requires explicit annotations, preferably with a Final annotation (e.g. WOODWINDS: Final[list[str]] = ['Oboe', 'Bassoon']).
https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness

So it's plausibly a bug that they don't show up in `--verbose` or `--outputjson`, but it's intended that they're not covered by `--verifytypes`.

jakkdl: Actually no, it's the usual
> Imported symbols are considered private by default. If they use the “import A as A” (a redundant module alias), “from X import A as A” (a redundant symbol alias), or “from . import A” forms, symbol “A” is not private unless the name begins with an underscore. 

that hits here. So what'd need to be changed here is changing the huge list of imports from `socket` to be `from socket import (X as X, Y as Y, Z as Z)`. 

I don't think we care for this in terms of `--verifytypes`, *but* if somebody uses pyright as their type checker it will give an error of `reportPrivateImportUsage` when they try accessing it. That seems important enough to warrant making that huge bulk of an import statement even messier

jakkdl:
Okay, I think that's all comments addressed. The drop in codecov I think is solely due to #2645

trio/tests/verify_types.json
A5rocks: FWIW you can still see the diffs for autogenerated files, just have to click through one "load autogenerated" text button thing. Up to you though, this isn't too important.

jakkdl: ah, and maybe there's a way to see the diff in CLI as well. But regardless I think I prefer it being treated as a normal file in terms of diffing

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,160 @@
...
+                del diag["file"]
+                diag.pop("range", None)
+            new_symbols.append(symbol)
+
+        current_result["typeCompleteness"]["symbols"] = new_symbols
+
+        with open(RESULT_FILE, "w") as file:
+            json.dump(current_result, file, sort_keys=True, indent=2)
+            # add newline at end of file so it's easier to manually modify
+            file.write("\n")
```
A5rocks, line 144: OK I reread things and I kinda don't get why we keep `symbols` at all. I don't see them being used elsewhere, am I missing something? They're a *gigantic* list and probably most of the reason the file is as big as it is.

trio/_tests/test_exports.py
A5rocks: Hm yeah, well it's a gigantic change so let's maybe sit on it a bit! Maybe an ad-hoc codemod would help out a ton, in combination with a run from `black`?

check.sh
A5rocks: I'm going to guess it handles this right and if it doesn't, we can fix then.

trio/_tests/check_type_completeness.py
jakkdl: We initially removed it, but I realized when I was trying out stuff that without it you won't get any feedback if/when the check fails. The numbers will change, but unless you `checkout` previous revisions, manually run `pyright --verifytypes`, pipe into a file, and manually diff against a manual `--verifytypes` run on your PR - you're left guessing what symbol you broke where. And if your local environment differs from CI, then you're absolutely toast. (I have recurring problems with CI not giving same results as running stuff in my virtualenv and/or tox envs)

I could shorten the file further, removing more info that's probably redundant, not pretty-print it - or like dump `symbols` into a compressed pickle that the script reads to say what symbols have changed, but:
1. the json file is not included when packaging
2. the json file will likely drop down in size a lot early on as I do a first pass and introduce a lot of trivial typing on functions.

If you still think it's a problem, I'll make a pass at sizing it down. Or I can toss symbols out for now, run the script modified locally, and after (2) when the file is smaller add it to the repo.

Other esoteric options include commiting the file to a different branch, git-annex, have it in a submodule...

A5rocks: Hmm, my concern is that we're making the repository larger for anyone who wants a (full) clone. Do you have measurements of the .git folder before and after this? (You may want to rebase first to get rid of the old much bigger json file)

jakkdl: Looks like going from something like 7.9M -> 8.1M
But it's a good point, I hadn't realized that big files will stick around forever in the history. I'll test pruning it down to being just a list of symbols with no extra info

jakkdl:
decided to go back to running with `--ignoreexternal` - while some were stuff that should be fixed, a lot of it was it complaining about the hacky imports in `trio.socket` and similar low prio stuff. #FutureProblem

trio/_tests/test_exports.py
jakkdl: vim macros FTW :D

A5rocks (commented):
I *think* this looks good? I'll review in slightly more detail once this is rebased!

trio/__init__.py
```diff
@@ -12,87 +12,104 @@
 #
 # This file pulls together the friendly public API, by re-exporting the more
 # innocuous bits of the _core API + the higher-level tools from trio/*.py.
+#
+# Uses `from x import y as y` for compatibility with `pyright --verifytypes` (#2625)
 
-from ._version import __version__
+# pyright explicitly does not care about `__version__`
+# see https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness
+from ._version import __version__ as __version__
```
A5rocks, line 20: The comment yet having `import x as x` is kinda confusing me. Did you just forget to revert the `import x as x` change here?

ci.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -ex -o pipefail
+export PYRIGHT_PYTHON_IGNORE_WARNINGS=1
```
A5rocks, line 4: TBF this probably is better off in the `check.sh` file, probably?

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,159 @@
...
+from pathlib import Path
+import sys
+import argparse
+
+# the result file is not marked in MANIFEST.in so it's not included in the package
+RESULT_FILE = Path(__file__).parent / "verify_types.json"
+failed = False
+
+
+# consider checking manually without `--ignoreexternal`, and/or
```
A5rocks, line 14: Maybe stick a `TODO` here, just to mark it explicitly?

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,159 @@
...
+        strcurrent = str(current)
+        strlast = str(last)
+    print(
+        f"{key} has gone {'down' if current<last else 'up'} from {strlast} to {strcurrent}"
+    )
+
+
+def check_zero(key, current_dict):
+    if current_dict[key] != 0:
+        global failed
```
A5rocks, line 47:
```suggestion
def check_zero(key, current_dict):
    global failed
    if current_dict[key] != 0:
```

trio/_tests/check_type_completeness.py
```diff
@@ -0,0 +1,159 @@
...
+            # add newline at end of file so it's easier to manually modify
+            file.write("\n")
+
+    if py_typed_file is not None:
+        print("deleting py.typed")
+        py_typed_file.unlink()
+
+    print("*" * 20)
+
+    return int(failed)
```
A5rocks, line 151: `bool` is a subclass of `int` so this isn't necessary. But maybe nice for clarity? Dunno.

trio/socket.py
```diff
@@ -210,3 +92,481 @@
...
+# re-export them. Since the exact set of constants varies depending on Python
+# version, platform, the libc installed on the system where Python was built,
+# etc., we figure out which constants to re-export dynamically at runtime (see
+# below). But that confuses static analysis tools like jedi and mypy. So this
+# import statement statically lists every constant that *could* be
+# exported. It always fails at runtime, since no single Python build exports
+# all these constants, but it lets static analysis tools understand what's
+# going on. There's a test in test_exports.py to make sure that the list is
+# kept up to date.
+try:
```
A5rocks, line 107: Maybe we should put this in an `if TYPE_CHECKING` block? But eh, I guess it does literally say "It always fails at runtime" and I don't know how this might change tools other than type checkers. Ig we can push this change off if you want!

A5rocks (approved):
Looks good overall, small little nitpicks.

ci.sh
jakkdl: added a comment

trio/socket.py
jakkdl: looks like `jedi` doesn't care, so switching it to be a `TYPE_CHECKING` block instead.

trio/_tests/check_type_completeness.py
jakkdl: haha, I guess it's my c++ programming leaking. But yeah I don't mind the int conversion

A5rocks (approved):
Looks good, quickly glanced through and didn't see anything! If tests pass, feel free to merge this after making commit history more succinct.

jakkdl:
:tada: :tada: :rocket: :rocket: :rocket:

Issue #2630, opened 2023-04-10
Static tools doesn't see class members (at least for trio.Path)


jakkdl:
Jedi:
```python
> import jedi
> script = jedi.Script("import trio; trio.Path.")
> comp = script.complete()
> [cc.name for cc in comp if not cc.name.startswith('_')]
Out: ['mro', 'open']
```
mypy also doesn't see them, which is probably the bigger problem
```sh
$ mypy trio/tests/test_path.py
...
trio/tests/test_path.py:189: error: "Type[Path]" has no attribute "joinpath"  [attr-defined]
```
which means downstream users will get typing errors if they use `trio.Path` in their code.

I haven't figured out how to check pylint, but otherwise it should be doable to modify `test_static_tool_sees_all_symbols` to also check class members.

For `trio.Path`, there's presumably some reason it's doing lots of magic instead of inheriting from `Path` - so if that's not on the table one would either have to define all the methods, or if that's what `_forward_magic` is for it's maybe possible to list them there, or some other constant.

Fuyukai:
The magic with ``Path`` is one of the less egregious dynamic-style things in the lib because you actually cannot inherit from Path:

```
>>> from pathlib import Path
>>> class MyPath(Path):
...     pass
...
>>> p = MyPath("/tmp")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/pathlib.py", line 960, in __new__
    self = cls._from_parts(args)
  File "/usr/lib/python3.10/pathlib.py", line 594, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/usr/lib/python3.10/pathlib.py", line 587, in _parse_args
    return cls._flavour.parse_parts(parts)
AttributeError: type object 'MyPath' has no attribute '_flavour'
```

(that being said, I still do not like ``trio.Path``)

TeamSpen210:
Pathlib will actually be subclassable, as of [3.12](https://github.com/python/cpython/pull/31691). Before then you'd have to set the `_flavour` attribute to the private `_PosixFlavour()` or `_WindowsFlavour()` classes.

For statically defining the methods, that could be done for mypy by using `if typing.TYPE_CHECKING:`, then defining stub methods for them. Or use a `.pyi` file. Looks like jedi will support that too.

jakkdl:
[...]

PR #2631, opened 2023-04-12
Add stubs for wrapped methods to trio.Path, add tests for mypy seeing exported symbols, and test for seeing class members

jakkdl:
For #2630


trio/_path.py
```diff
@@ -182,6 +182,70 @@ async def open(self, *args, **kwargs):
         value = await trio.to_thread.run_sync(func)
         return trio.wrap_file(value)
 
+    if TYPE_CHECKING:
```
A5rocks, line 185: We don't need this `if TYPE_CHECKING` FYI, as the metaclass will overwrite these methods.

Probably OK though.

TeamSpen210: It does make it more clear that these are for checkers only, and also improves import performance a tiny bit.

trio/_path.py
```diff
@@ -182,6 +183,78 @@ async def open(self, *args, **kwargs):
...
+        __truediv__ = pathlib.Path.__truediv__
+        __rtruediv__ = pathlib.Path.__rtruediv__
+
+        # Note: these are not parsed properly by jedi.
+        # It might be superior to just manually implement all the methods and get rid
+        # of all the magic wrapping stuff.
+        ArgsT = ParamSpec("ArgsT")
+        RetT = TypeVar("RetT")
+
+        # mypy complains about missing `self` argument
```
TeamSpen210, line 198: Might want to move this to module scope and underscore-prefix it, since right now tools will also think `trio.Path.make_aync()` is something you can do. Also `ParamSpec` needs to be imported from `typing_extensions` at runtime for 3.9 and below.

jakkdl: yeah good call

and thanks!

TeamSpen210:
Actually, does trio want a dependency on `typing_extensions`? #2507 suggests not, so it'd be better to move the import into the `if TYPE_CHECKING` guard, then just unconditionally import from `typing`.

jakkdl:
> Actually, does trio want a dependency on `typing_extensions`? #2507 suggests not, so it'd be better to move the import into the `if TYPE_CHECKING` guard, then just unconditionally import from `typing`.

ah yes, good catch

jakkdl:
I should maybe spin up a virtual machine at some point so I can test windows stuff better lol

jakkdl:
Huh, looks like pypy has different dunders? I'm guessing they can just be ignored.
>  'trio.EndOfChannel': {'extra': {'__basicsize__',
                                 '__dictoffset__',
                                 '__itemsize__',
                                 '__sizeof__',
                                 '__weakrefoffset__'},
                       'missing': {'__unicode__'}},

Windows ones are failing due to path having a colon in `C:` which is easily worked around.

TeamSpen210:
Yeah those aren't relevant. `__unicode__` would be Py2 remnants, PyPy mustn't have removed that yet. The rest are how the memory layout of the object is defined/exposed for introspection, which isn't something Python code should care about unless it's using `ctypes` or something.

Fuyukai:
Truly incredible. Now we have the worst of both worlds: you have to type out all the wrapper declarations manually *and* you're testing that you even wrote out the wrapper declarations - and if we ever want to have real asynchronous file I/O all of this has to go anyway.

A5rocks:
I think it's worth the effort to provide better autocomplete! (Sorry if I'm misinterpreting your tone!)

The test is just a continuation of what we already test -- just that we are checking that class members match and not just module members.

jakkdl:
> Truly incredible. Now we have the worst of both worlds: you have to type out all the wrapper declarations manually and you're testing that you even wrote out the wrapper declarations - and if we ever want to have real asynchronous file I/O all of this has to go anyway.

I could've implemented the class member check as making the export test more thorough, but was easier to just write a new one than modifying the old one to also do class members. So yes the tests are very much what's currently done already, except with one more tool and testing the public class methods that weren't.

`_path.py` is messy and ugly, I don't disagree with that, but it's the only class that requires this much of a mess, and it's not like the interface of `pathlib.Path` is gonna change anytime soon, so this shouldn't introduce any maintenance overhead or anything.

jakkdl:
These last errors are confusing me (and then there's other problems currently untested that'll need adressing), but I *think* it boils down to `dir(pathlib.Path)` includes 'group', 'owner' and 'is_mount' on both `win32` and not - but if you'd try calling them on `win32` you'd get an error. And `mypy` solves that by saying they're not attributes of `pathlib.Path` if run with `--platform=win32`.
I thought that maybe `dir(trio.Path)` was bugged on win32 in listing them, but if `pathlib.Path` also does it then I guess it isn't really a bug - it's just mypy special-casing it and so the correct thing is to expect the current status quo with `dir()` and `mypy` not being the same, and special-case them in the tests.

jakkdl:
Okay since getting the stubs correct for the path methods wasn't trivial, I'll postpone that to a different PR so this one can get merged.

jakkdl:
would somebody mind reviewing this? It's ready for merge unless there's any more comments.

A5rocks:
Sorry, haven't gotten to this! Eventually I will, don't worrk.

A5rocks:
(just updating this branch based on master as that (should) fix the macos pypy-3.8-nightly cancellation and the formatting check)

trio/_path.py
```diff
@@ -182,6 +182,73 @@ async def open(self, *args, **kwargs):
         value = await trio.to_thread.run_sync(func)
         return trio.wrap_file(value)
 
+    if TYPE_CHECKING:
+        # the dunders listed in _forward_magic that aren't seen otherwise
+        __bytes__ = pathlib.Path.__bytes__
+        __truediv__ = pathlib.Path.__truediv__
+        __rtruediv__ = pathlib.Path.__rtruediv__
+
+        # Note: these are not parsed properly by jedi.
```
A5rocks, line 191: I think this comment is outdated?

trio/_path.py
```diff
@@ -182,6 +182,73 @@ async def open(self, *args, **kwargs):
...
+        stat: Any
+        symlink_to: Any
+        touch: Any
+        unlink: Any
+        with_name: Any
+        with_suffix: Any
+        write_bytes: Any
+        write_text: Any
+
+        if sys.platform != "win32":
```
A5rocks, line 238: It would be nice to be able to auto-generate this (read from relevant file in https://github.com/python/typeshed and change the types... like just stick an `async` in the right places and call it a day). I suppose that's all part of "how will we do this" that is the followup, so ignore this :^)

trio/_path.py
```diff
@@ -182,6 +182,73 @@ async def open(self, *args, **kwargs):
...
+
+        if sys.version_info >= (3, 8):
+            link_to: Any
+        if sys.version_info >= (3, 9):
+            is_relative_to: Any
+            with_stem: Any
+            readlink: Any
+        if sys.version_info >= (3, 10):
+            hardlink_to: Any
+
```
A5rocks, line 251: This line isn't the one I wanted to comment on, but I can't comment on it cause it's out of the diff github shows :(

Namely, the line that is `os.PathLike.register(Path)`. Presumably we could *subclass* `os.PathLike` and that would help some typing stuff? I don't know though.

trio/tests/test_exports.py
```diff
@@ -96,6 +101,30 @@ def no_underscores(symbols):
         script = jedi.Script(f"import {modname}; {modname}.")
         completions = script.complete()
         static_names = no_underscores(c.name for c in completions)
+    elif tool == "mypy":
+        if not RUN_SLOW:
+            pytest.skip("use --run-slow to check against mypy")
```
A5rocks, line 106: Does `pytest.skip` not need a return? I just want to be sure!

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
         assert False
 
 
+# this could be sped up by only invoking mypy once per module, or even once for all
+# modules, instead of once per class.
+@slow
+# see commend on test_static_tool_sees_all_symbols
```
A5rocks, line 149:
```suggestion
# see comment on test_static_tool_sees_all_symbols
```

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+@pytest.mark.redistributors_should_skip
+# pylint/jedi often have trouble with alpha releases, where Python's internals
+# are in flux, grammar may not have settled down, etc.
+@pytest.mark.skipif(
+    sys.version_info.releaselevel == "alpha",
+    reason="skip static introspection tools on Python dev/alpha releases",
+)
```
A5rocks, line 156: I'm not sure this is actually necessary (we allow alpha releases to fail iirc!) but given `test_static_tool_sees_all_symbols` has this too, this is probably fine.

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+            symbol
+            for symbol in symbols
+            if (not symbol.startswith("_")) or symbol.startswith("__")
+        }
+
+    if tool == "mypy":
+        if sys.implementation.name != "cpython":
+            pytest.skip("mypy not installed in tests on pypy")
+        # create py.typed file
+        (Path(trio.__file__).parent / "py.typed").write_text("")
```
A5rocks, line 174: I'm not *thrilled* about this, given... could `trio.__file__` point to a user's development checkout of trio? If there's a very simple alternative I would prefer that, though this is very much a transient issue and easy to work around on a user's side.

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+
+    errors: dict[str, Any] = {}
```
A5rocks, line 176:
```suggestion
    errors: dict[str, object] = {}
```

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+    if module_name == "trio.tests":
+        return
+    for class_name, class_ in module.__dict__.items():
+        if not isinstance(class_, type):
+            continue
+        if module_name == "trio.socket" and class_name in dir(stdlib_socket):
+            continue
+        # Deprecated classes are exported with a leading underscore
+        if class_name.startswith("_"):  # pragma: no cover
```
A5rocks, line 185: Given the `pragma: no cover`, am I right to assume this doesn't happen at the moment? I presume if we ever do export a deprecated class (?), it'll have member annotations so this exception might be unnecessary.

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+
+            missing = runtime_names - static_names
+            extra = static_names - runtime_names
+            if BaseException in class_.__mro__ and sys.version_info > (3, 11):
+                missing.remove("add_note")
+
+            # TODO: why is this? Is it a problem?
+            if class_ == trio.StapledStream:
+                extra.remove("receive_stream")
+                extra.remove("send_stream")
```
A5rocks, line 240: If we're so confused about this, is this worth raising an issue with jedi?

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+
+            if missing | extra:
```
A5rocks, line 242:
```suggestion
            if missing or extra:
```

might be more approachable!

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+            res = run(
+                [
+                    "--config-file=",
+                    "--follow-imports=silent",
+                    "--disable-error-code=operator",
+                    str(tmpfile),
+                ]
+            )
+            if res[2] != 0:
+                print(res)
```
A5rocks, line 265: Print what `module.class` is being tested, maybe?

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+                it = iter(res[0].split("\n")[:-2])
```
A5rocks, line 266: There's an option to silence the ending thing. Uhh, here's two flags I think are useful, now that I have the docs up:
 - `--soft-error-limit=-1`
 - `--no-error-summary`

Both can be viewed at https://mypy.readthedocs.io/en/stable/command_line.html#advanced-options

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+                for output_line in it:
+                    # hack to get around windows path starting with <drive>:<path>
+                    output_line = output_line[2:]
+
+                    _, line, error_type, message = output_line.split(":")
```
A5rocks, line 271:
```suggestion
                    _, line, error_type, message = output_line.rsplit(":", 4)
```
I think this works?

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+                    if (
+                        symbol
+                        in ("trio.Path.group", "trio.Path.owner", "trio.Path.is_mount")
+                        and sys.platform == "win32"
+                        and sys.implementation.name == "cpython"
+                    ):
+                        continue
+
+                    # a bunch of symbols have this error, e.g. trio.lowlevel.Task.context
+                    # but I don't think that's a problem - and if it is it would likely
```
A5rocks, line 291: It's not a problem: it's just complaining we're accessing instance-only attributes on a class! See this test for a minimized version that causes this error: https://github.com/python/mypy/blob/c517b86b9ba7487e7758f187cf31478e7aeaad47/test-data/unit/check-slots.test#L515-L523.

trio/tests/test_exports.py
```diff
@@ -114,6 +143,169 @@ def no_underscores(symbols):
...
+
+                    errors[symbol] = error_type + ":" + message
+
+        else:  # pragma: no cover
+            assert False
+
+    if errors:  # pragma: no cover
+        from pprint import pprint
+
+        print(f"\n{tool} can't see the following symbols in {module_name}:")
```
A5rocks, line 304: Do we want a bigger separator? Like, `\n\n--------\n{tool} can't see the following symbols in {module_name}:`? IDK

A5rocks: On second thought, why do we need to pprint `errors`? Shouldn't `pytest` do that for us with `assert not errors`?

A5rocks (changes requested):
I was telling myself not to nitpick too much but... Oops. If something feels like a nitpick, feel free to ignore it! (or if it's missing something too obvious, I guess)

trio/_path.py
TeamSpen210: Definitely, `register` isn't going to be fully supported everywhere. It'd also be better to inherit to indicate that we provide string paths, not bytes. A little tricky though since it's only as of 3.9 that you could subscript `PathLike`:
```py
if TYPE_CHECKING:
    _StrPathLike = os.PathLike[str]
else:
    _StrPathLike = os.PathLike
...
class Path(_StrPathLike):
```

trio/_path.py
TeamSpen210, line 1: Something else we'll want to do is guard `__getattr__` with `if not TYPE_CHECKING`, because if a class has that method type checkers will allow accessing any attribute.

trio/_path.py
jakkdl: I at least don't remember why I wrote it, so removing it for now.

trio/tests/test_exports.py
jakkdl: Looks like it's marked `NoReturn` and works by raising a special exception: https://docs.pytest.org/en/stable/_modules/_pytest/outcomes.html#skip
and definitely worked on my machine!

Though as the doc says should maybe use `pytest.mark.skipif` instead

trio/tests/test_exports.py
jakkdl: yeah I didn't really look into why it's there and just copy-pasted it

trio/tests/test_exports.py
jakkdl: I can clean it up afterwards if it didn't exist before running

trio/tests/test_exports.py
jakkdl: it's faster too! I'm actually a bit surprised no linter pointed it out, altho I guess you need type info to check it and mypy is not in the business of pointing out minor things like this.

trio/tests/test_exports.py
jakkdl: ah, this was a debug print when I was trying to figure out the splits and stuff. Removed

trio/tests/test_exports.py
jakkdl: updated the comment

trio/tests/test_exports.py
jakkdl: oh shit, `--soft-error-limit` is maybe why I had problems when I was running `mypy` and passing the code to check as a command-line parameter instead of printing to a file

trio/tests/test_exports.py
jakkdl: ooh, I didn't know about rsplit - very handy in this case! It should only be a 3 though, we want to parse 3 `:` splitting into 4 variables.

trio/_path.py
jakkdl: That does not seem correct, if I comment out an attribute in the hardcoded list (e.g. `mkdir`) then mypy will complain that it doesn't exist when I attempt to access it and the test will fail

trio/tests/test_exports.py
jakkdl: Maybe some weird interaction with `attr.ib()`?
https://github.com/python-trio/trio/blob/9676408f2d41e4968ac97076fabdb68e4c14d01c/trio/_highlevel_generic.py#L39-L40
https://github.com/python-trio/trio/blob/9676408f2d41e4968ac97076fabdb68e4c14d01c/trio/_highlevel_generic.py#L75-L76

trio/tests/test_exports.py
jakkdl: MultiError is exported as `_MultiError`, so it is happening at the moment. But I didn't wanna care about whether code that's on it's way out is typed correctly or not, hence skipping it.
The reason for the `pragma` is so that when `_MultiError` *does* get removed it doesn't decrease coverage in this random test.

I think that was my reasoning at least, I'll make the comment clearer

jakkdl: Oh, except `MultiError` has no problems anyway (now, it got fixed with all the other exception fixes) - so I'll just remove the if/continue.

trio/tests/test_exports.py
jakkdl: oh nvm, cant use `pytest.mark.skipif` on a parametrized variable.

trio/_path.py
jakkdl: let's leave that to a different PR that fully types `trio.Path`, can either continue with that in #2630 or open a different issue

jakkdl:
Fixed all changes requested, except where noted. Good nitpicks @A5rocks! :heart:

trio/tests/test_exports.py
jakkdl: nvm, it does have problems https://github.com/jakkdl/trio/actions/runs/4893133337/jobs/8735733831#step:5:1779
reintroduced the if/continue and clarified the comment

trio/_path.py
TeamSpen210: Ahh, I think I know why. In the checks, you're looking them up on the class scope, so `__getattr__` wouldn't apply. If the tests were changed to do `Path().mkdir` then I think it'll take effect.

jakkdl: Ah yes, it did! Although changing to be an instance check gives a bunch of false alarms in jedi (and a couple typing errors in mypy) that'd need cleaning up, since the `dir()` call is on the class and not a class instance. And instantiating the class during runtime sounds quite bothersome. So I'll just typeguard `__getattr_`, ignore it missing in tests, and leave it at that.

A5rocks (approved):
Seems great from a cursory glance, we can always make a bunch of followups!

jakkdl:
Okay I'm done fighting with codecov, merging!

PR #2634, opened 2023-05-01
Update pip-tools

A5rocks:
https://github.com/python-trio/trio/actions/runs/4847059956/jobs/8636964320 :(

Rerun the workflow once this is merged


richardsheridan:
Looks like pip-tools has new code for pip 23 on main but not released yet https://github.com/jazzband/pip-tools/blob/main/piptools/_compat/pip_compat.py#L79

richardsheridan:
Potential real fix in #2636

PR #2636, opened 2023-05-01
Invert no changes check in autodeps

richardsheridan:
Changes in the working tree report 1 with this command, and 1 is falsy in bash.


A5rocks (approved):

A5rocks:
Still doesn't work, now onto https://github.com/python-trio/trio/pull/2637 :S

PR #2637, opened 2023-05-01
Ensure our autodeps workflow adds the new files

A5rocks:
Currently it's failing because https://github.com/python-trio/trio/actions/runs/4853254871/jobs/8649206439#step:6:56

I think this fixes that? Not sure.

### DIFF
--- a/.github/workflows/autodeps.yml
+++ b/.github/workflows/autodeps.yml
@@ -49,6 +49,7 @@ jobs:
           git switch --force-create autodeps/bump_from_${GITHUB_SHA:0:6}
           git config user.name 'github-actions[bot]'
           git config user.email '41898282+github-actions[bot]@users.noreply.github.com'
+          git add .
           git commit -am "Dependency updates"
           git push --force --set-upstream origin autodeps/bump_from_${GITHUB_SHA:0:6}
 


richardsheridan (approved):

richardsheridan:
Why did it allow me to merge that without waiting for checks? disturbing

A5rocks:
Nope it didn't work. I... don't know what's up.

richardsheridan:
I tested the sequence of commands in a test repo before I implemented it here. 

https://github.com/richardsheridan/workflow-sandbox/actions/runs/4345469352/jobs/7590222158

The only difference I can make out is the changes are coming from pip-compile.

A5rocks:
I think there just aren't any changes to the file. I ran through the steps myself and `git status` is empty.

A5rocks:
Ah, maybe we need to use `pip-compile -U` (`-U` -> `--upgrade`).

A5rocks:
Hm, I don't know. I checked and that exit code should only mean that there's changes (`git diff-index`: <https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code>). I... don't really get why this doesn't work.

richardsheridan:
Maybe instead of running `--quiet` we should use `--exit-code` so we can see exactly what the changes are.

A5rocks:
> If `pip-compile` finds an existing `requirements.txt` file that fulfils the dependencies then no changes will be made, even if updates are available.

<https://pip-tools.readthedocs.io/en/latest/#updating-requirements> 🤔 not sure if this applies in the way we are using it. I'll add `-U` anyways!

PR #2638, opened 2023-05-04
Some small improvements before debugging autodeps

A5rocks:
Hopefully this fixes this. Even if it doesn't, this should print out more detailed information for debugging (such as local changes).


A5rocks:
https://github.com/python-trio/trio/pull/2639 !!

PR #2640, opened 2023-05-05
add flake8-trio to list of Tools and Utilities in Awesome Trio Libraries

jakkdl:
Adding the linter created by me and @Zac-HD to the list of Tools and Utilities.


Zac-HD (approved):
LGTM, thankfully 😜 

Merging now, since the py312 job is also failing on main.

PR #2641, opened 2023-05-06
Add "concurrency" to CI so old unused CI runs automatically get cancelled

A5rocks:
See https://docs.github.com/en/actions/using-jobs/using-concurrency


A5rocks:
cc @webknjaz since you looked at some of our CI stuff!

altendky:
Will this cancel runs on `master` as well if a new run is started before the previous finishes?  This is what I've been using recently to only cancel PR triggered runs.  This lets pushes to master and tag triggers avoid cancellation.

https://github.com/gorakhargosh/watchdog/blob/48c49a1fe0538be455d52aa9a0eb57519a6ce2eb/.github/workflows/tests.yml#L11-L13
```yaml
concurrency:
  group: ${{ github.ref }}-${{ github.workflow }}-${{ github.event_name }}-${{ github.event_name != 'pull_request' && github.sha || '' }}
  cancel-in-progress: true
```

Though I notice that here we're running both `push` and `pull_request` for the PRs so that might need tweaked a bit to actually check if it's building on the default branch instead more like this.

https://github.com/pytest-dev/pytest-twisted/blob/98b40e9919d4e6884600fed9449bc88d69a36a41/.github/workflows/ci.yml#L16-L18
```yaml
concurrency:
  group: ${{ github.ref }}-${{ github.workflow }}-${{ github.event_name }}-${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && github.sha || '' }}
  cancel-in-progress: true
```

A5rocks:
> Will this cancel runs on master as well if a new run is started before the previous finishes? 
I believe so. I have this on a private repository and it's been cancelling runs as necessary:
```

**Comment by altendky** [comment]
I have become used to this behavior and like it, but I'm also not very active here.  I'll leave this for someone else that's been working on Trio recently to chime in with an approval of the behavior change.  Thanks for the work here!

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -6,6 +6,10 @@ on:
       - "dependabot/**"
   pull_request:
 
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}-${{ github.event_name }}-${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && github.sha || '' }}
```
webknjaz, line 10:
```suggestion
  group: ${{ github.ref }}-${{ github.workflow }}-${{ github.event_name }}${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && format('-{0}', github.sha) || '' }}
```

webknjaz (approved):
I like Kyle's suggestion you implemented. I've been seeing different approaches, more complex than your original one. This one looks a bit more involved but seems to be better than what I've been using so far.

I've added a tiny suggested change below — it is meant to get rid of an ugly trailing dash in some cases. Though, it's up to you whether you want to accept it or not...

webknjaz:
@A5rocks look what I've just found: https://github.com/neovim/neovim/blob/5a08b5b/.github/workflows/test.yml#LL12C37-L12C133 — that's in case you'd want to borrow some ideas from them. But the current implementation seems good to me even without that. I'll probably inspect why neovim does what it does, though, for my future project improvements...

altendky:
https://github.com/neovim/neovim/commit/e71c7898ca3cf3af1243227ff3cba526d48897e8
> This will allow contributors to test changes in their own fork when pushing without needing to make a pull request.

I guess.  I've just accepted in my setups that  a 'pull request' doesn't mean 'please merge', it's just a way to track a set of changes and associated comments and ci runs.  But, maybe this is nice.  If anyone actually runs a fork then it makes their fork work differently.

A5rocks:
If I'm not reading the actions conditions wrong (I probably am), that seems to essentially overlap with what we do except that *also* never cancels old runs in the entire un-forked repository. I'm not that big of a fan of that, especially not for this repository in particular.

I'll wait for someone to confirm my logic above (or tell me what I'm missing!) before merging!

webknjaz:
> But, maybe this is nice. If anyone actually runs a fork then it makes their fork work differently.

Exactly, I personally like CI runs on pushes. Though, in some projects, like Ansible, we are forbidden to push branches to the upstream repositories — everybody has to work in forks, even core. This has a nice effect of not polluting the tree for people making forks, as they'd get all the garbage branches in their outdated states frozen in their repository copies, which could confuse many newbies later on...

> in the entire un-forked repository

Only for the commit.

webknjaz:
Merging since this can be improved later on, no need to block this.

PR #2642, opened 2023-05-06
Make sure the index is synced before diff-index

A5rocks:
This should fix autodeps once and for all. Hopefully!

Seems to work here: https://github.com/python-trio/trio/actions/runs/4899533479/jobs/8749560235#step:6:57

### DIFF
--- a/.github/workflows/autodeps.yml
+++ b/.github/workflows/autodeps.yml
@@ -39,6 +39,9 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          # git diff-index needs a git update-index, which can be done more concisely with `git add`!
+          git add .
+
           # https://stackoverflow.com/a/3879077/4504950
           if git diff-index --exit-code HEAD; then
             echo "No changes to commit!"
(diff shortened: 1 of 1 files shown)


richardsheridan (commented):
Good sleuthing! Does this mean that the `-a` of the `git commit` is redundant now? Also, what do you think of just checking the return code of `git commit -a`?

A5rocks:
Done and it still works: https://github.com/python-trio/trio/actions/runs/4936843958/jobs/8824875611#step:6:55

richardsheridan:
Amazing, thanks!  It's much simpler to follow this way.

Sxderp:
https://github.com/python-trio/trio/blob/d0d8d618157c5b9270f7f012ba809118cfc6ce9f/.github/workflows/autodeps.yml#L47

Should be "git commit" no?

richardsheridan:
Yes, could you make a fix PR that we could merge?

A5rocks:
Ahh oops, well no wonder it errored -> no changes :S

PR #2648, opened 2023-05-24
fix pickle support for MultiError

### LINKED ISSUE #2543: MultiError is un-unpicklable
```python
import trio
import pickle
me = trio.MultiError([ValueError()])
pickle.loads(pickle.dumps(eg))
```
throws a
```shell
----> 1 pickle.loads(pickle.dumps(me))

TypeError: MultiError.__new__() takes 2 positional arguments but 3 were given
```
error, I think because of some conflict with how ExceptionGroups - which expect a message as well as an error list - are constructed.

trio version 0.22.0, python 3.10.6

jakkdl:
fixes #2543 

I think the fundamental cause of the problem is that `MultiError` does not take a `message` parameter in `__new__` despite it being listed in `__dict__`. But `__collapse__` being a keyword parameter might also have contributed to the problem. I tried using [`__getnewargs_ex__`](https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__) and friends, but I failed to get anything but [`__reduce__`](https://docs.python.org/3/library/pickle.html#object.__reduce__) to do anything (they didn't seem to get called at all... :thinking:). Could also have rewritten `__new__`, but suspect that might maybe break some code out there.


trio/_core/_tests/test_multierror.py
```diff
@@ -555,3 +557,36 @@ def test_apport_excepthook_monkeypatch_interaction():
...
+        my_except = e
+
+    # MultiError will collapse into different classes depending on the errors
+    for cls, errors in (
+        (ZeroDivisionError, [my_except]),
+        (NonBaseMultiError, [my_except, ValueError()]),
+        (MultiError, [BaseException(), my_except]),
+    ):
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
A5rocks, line 581: What specifically is this here for?

Also, can we add a test against strict MultiErrors (from `strict_exception_groups=True`) while we're at it? (it should pass, I think?)

jakkdl: For ignoring the deprecation warnings on MultiError. I'll specify it in the simplefilter.
I explicitly wanted to use `trio.MultiError` (instead of the `from .._multierror import MultiError` that the rest of the tests in the file uses) to make sure pickling works through the deprecation layer.

I'm not totally sure what you mean, and all the `strict_exception_groups=True` tests in `trio/_core/_tests/test_run.py` are using MultiError - so maybe it's handled over there? Otherwise feel free to push a commit / open a separate PR with the test you're thinking of, I haven't touched `strict_exception_groups` much at all.

A5rocks: Sorry, what I mean is just that maybe we should test that the errors that come with `strict_exception_groups=True` (which are still `MultiErrors`, I think? Just non-collapsed, I believe?) are also pickle-able.

jakkdl: Ah, thank you!
It fails due to traceback's not being pickle-able, but if I install and enable https://pypi.org/project/tblib/ it works. Although it might be because I'm running the test slightly different and not related to `strict_exception_groups` that the behaviour differs, idk.
So could leave the test out, add it but skip it if `tblib` isn't installed, add `tblib` to test requirements, or play around more with tests to figure out exactly how/where it differs. But regardless I think it works both with&without `strict_exception_groups`.

A5rocks: Makes sense. Didn't realize it would require a dependency, all's good.

Issue #2646, opened 2023-05-22
Fix Python 3.12 errors


A5rocks:
Now that Python 3.12 is entering its beta releases (as of today, according to PEP 693), there should no longer be any feature additions. This means it's going to be more stable, I hope. So, it's time to start poking at test failures!!

https://github.com/python-trio/trio/actions/runs/5013776086/jobs/8987273168 shows that currently pytest is... broken on Python 3.12. That's inconvenient. We are going to be waiting on https://github.com/pytest-dev/pytest/pull/10981 to merge, I guess! I'll keep this issue updated.

jakkdl:
We can for the moment just suppress that particular DeprecationWarning by adding `'ignore:ast.Str is deprecated:DeprecationWarning',` to `filterwarnings` in `pyproject.toml` https://docs.pytest.org/en/6.2.x/warnings.html

It also looks like `test_nursery_cancel_doesnt_create_cyclic_garbage` is failing, I looked into it a little bit ago and it looked like `toggle_collected()` was called, but `collected` didn't get set to `True`, so I have no clue what's going on there.
https://github.com/python-trio/trio/blob/855f5fd72e9e126b8dae2348a918427d7a1f38cd/trio/_core/tests/test_run.py#L2285-L2294

A5rocks:
For that one specifically, I've already raised a CPython issue (https://github.com/python/cpython/issues/100964) but I need to investigate and respond, heh.

I remember seeing others, though. I suspect it's not worth the effort to ignore the warning as the pytest PR should merge whenever the beta comes out.

jakkdl:
I would love to have a quick patch that silences the 3.12 errors. When actively developing I find it's nice to be able to peek at the PR list and see if tests are failing or not - without having to check the long test list every time. And I also have it set up to get notifications on failing CI tests, which becomes ... not useful.

(codecov not handling windows is also at fault here, but that one's at least somewhat possible to work around)

webknjaz:
> (codecov not handling windows is also at fault here, but that one's at least somewhat possible to work around)

AFAIU, the current test runner uses a codecov uploader version that's been deprecated 3 years ago: https://github.com/python-trio/trio/issues/2650.

webknjaz:
> We can for the moment just suppress that particular DeprecationWarning by adding `'ignore:ast.Str is deprecated:DeprecationWarning',` to `filterwarnings` in `pyproject.toml`

[...]

PR #2651, opened 2023-05-26
Ignore Pytest's warning

A5rocks:
This is a temporary change until https://github.com/pytest-dev/pytest/pull/10894 is merged. cc @jakkdl


jakkdl:
Oh boi, with it exiting alpha now all the static tool tests are let loose.

How about marking all the currently failings tests with xfail on 3.12 so the CI passes? We can still track them more properly in #2646 but I imagine it'll be a while until all tests are passing on 3.12

```python
@pytest.mark.xfail(sys.version_info[:2] == (3, 12), reason="blabla")
```

A5rocks:
I'll do that in a separate PR. I suspect the largest diff will be fixed in the next upgrade of typing extensions (on June 1st we'll have our autodeps thing run) and for the rest, dunno.

A5rocks:
OK I got a bit impatient, https://github.com/python-trio/trio/pull/2652 should fix some stuff here

A5rocks:
To be blunt: I would like this merged before working on xfail-ing, fixing things, etc.

More generally I am of the opinion that we shouldn't xfail stuff that we plan to fix soon-ish but I want to investigate the errors a bit. Static stuff should be quick to fix, changes in cpython impl details not so much.

Issue #2650, opened 2023-05-25
[TODO] Replace codecov uploader with a modern version in GHA


webknjaz:
Currently, codecov is called from `ci.sh` but it uses a long-deprecated version of the uploader. Codecov deprecated their “legacy” uploaders after the breach in 2020 and have since migrated to a unified one. They also have an official GitHub Action that makes use of it. Several months ago, they've started actively deleting the uploading capabilities (and removed the PyPI package, for example).

I think, having custom logic in a shell script brings more maintenance burden than having it in the CI definition directly. Let's move towards GHA, especially since it's not really necessary to make uploads from the local runs outside the CI...

PR #2653, opened 2023-05-29
Replace codecov with their GHA thing

A5rocks:
Fixes https://github.com/python-trio/trio/issues/2650

I'm not sure this works. We'll see!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,8 +73,6 @@ jobs:
         env:
           LSP: '${{ matrix.lsp }}'
           LSP_EXTRACT_FILE: '${{ matrix.lsp_extract_file }}'
-          # Should match 'name:' up above
-          JOB_NAME: 'Windows (${{ matrix.python }}, ${{ matrix.arch }}${{ matrix.extra_name }})'
       - uses: codecov/codecov-action@v3
         with:
           directory: empty
@@ -122,8 +120,6 @@ jobs:
         run: ./ci.sh
         env:
           CHECK_FORMATTING: '${{ matrix.check_formatting }}'
-          # Should match 'name:' up above
-          JOB_NAME: 'Ubuntu (${{ matrix.python }}${{ matrix.extra_name }})'
       - uses: codecov/codecov-action@v3
         with:
           directory: empty
@@ -156,9 +152,6 @@ jobs:
           cache-dependency-path: test-requirements.txt
       - name: Run tests
         run: ./ci.sh
-        env:
-          # Should match 'name:' up above
-          JOB_NAME: 'macOS (${{ matrix.python }})'
       - uses: codecov/codecov-action@v3
         with:
           directory: empty
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,7 @@ jobs:
       - uses: codecov/codecov-action@v3
         with:
           directory: empty
+          name: 'Windows (${{ matrix.python }}, ${{ matrix.arch }}${{ matrix.extra_name }})'
 
   Ubuntu:
     name: 'Ubuntu (${{ matrix.python }}${{ matrix.extra_name }})'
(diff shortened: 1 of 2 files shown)


A5rocks:
Oh this is gonna wreck builds.sr.ht things. I don't think they work, anyways?

A5rocks:
Quick note to self, if this works then:
 - remove `JOB_NAME` cruft
 - remove the old copying-over-coveragerc hack

A5rocks:
See https://app.codecov.io/gh/python-trio/trio/pull/2653

Looks like it works.

.github/workflows/ci.yml
```diff
@@ -73,8 +73,9 @@ jobs:
         env:
           LSP: '${{ matrix.lsp }}'
           LSP_EXTRACT_FILE: '${{ matrix.lsp_extract_file }}'
-          # Should match 'name:' up above
-          JOB_NAME: 'Windows (${{ matrix.python }}, ${{ matrix.arch }}${{ matrix.extra_name }})'
+      - uses: codecov/codecov-action@v3
+        with:
+          directory: empty
```
webknjaz, line 78: Could you add flags here? Look into sphinxcontrib-towncrier, cheroot, ansible/pylibssh and aiomysql for examples. Make sure the flag names don't have special symbols — they may break uploads.

.github/workflows/ci.yml
```diff
@@ -73,8 +73,9 @@ jobs:
         env:
           LSP: '${{ matrix.lsp }}'
           LSP_EXTRACT_FILE: '${{ matrix.lsp_extract_file }}'
-          # Should match 'name:' up above
-          JOB_NAME: 'Windows (${{ matrix.python }}, ${{ matrix.arch }}${{ matrix.extra_name }})'
```
webknjaz, line 77: I wonder if this could've been reported to codecov.

.github/workflows/ci.yml
A5rocks: Surprisingly never heard of those, they sound neat! But... do we need them? The [docs](https://docs.codecov.com/docs/flags) say that they are for use with either different kinds of tests (which, meh, I don't think trio really shows?) or a monorepo.

Do you have a specific idea for how to use them here?

.github/workflows/ci.yml
A5rocks: It could be, the codecov action takes a name parameter. I'm not too sure what it does, so I might add it back in out of caution. Do you know what leaving this out does?

webknjaz:
> See https://app.codecov.io/gh/python-trio/trio/pull/2653\n\nLooks like it works.

It's 3 uploads short it seems..

> Coverage data is based on HEAD ca40b5e
(28 uploads)
 compared to BASE b21d4b9
(25 uploads)
>
> > Commits have different number of coverage report uploads learn more

A5rocks:
I can certainly see why we had the name, previously. Let me add that back, quickly.

.github/workflows/ci.yml
webknjaz: I think it is displayed on the UI but maybe what the action autodetects is fine too.

.github/workflows/ci.yml
webknjaz: Besides the test types (like with unit vs. integration running separately), I usually put the Python version and the OS there (hence the examples I mentioned). This allows seeing which code paths are tested under which runtimes.

.github/workflows/ci.yml
A5rocks: I checked and it just showed some arbitrary number. I added back the names (this should help debug mismatching upload counts!)

.github/workflows/ci.yml
webknjaz: The UI also has a way of toggling the flags and you'll see which flags cover the lines or don't.

A5rocks:
Oh wow I panicked too much, looks like this PR (as of last complete snapshot) makes 3 *more* uploads, not 3 less. Weird, but I'll take it!

webknjaz (approved):

PR #2654, opened 2023-05-30
specify --cov-config=.coveragerc in ci.sh

jakkdl:
seeing settings defined in .coveragerc not being respected, testing if this fixes it

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -99,7 +99,7 @@ else
     INSTALLDIR=$(python -c "import os, trio; print(os.path.dirname(trio.__file__))")
     cp ../pyproject.toml $INSTALLDIR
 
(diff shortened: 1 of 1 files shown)


jakkdl:
hmmnope, didn't help.

What I'm seeing is `if TYPE_CHECKING` not ignoring branches: https://app.codecov.io/gh/python-trio/trio/blob/master/trio/_path.py#L190

And in #2629 I tried omitting `trio/_tests/check_type_completeness.py` which didn't do shit.

A5rocks:
Could be because I removed a thing that copies .coveragerc into the empty directory? You might need to --cov-config=../.coveragerc

jakkdl:
okay that did fix `omit` (although not sure if that changes anything in this branch) - but I fail to get the `exclude_lines` regex to work / the content of `if TYPE_CHECKING:` ignored when running locally. I tried modifying it, I tried `exclude_also` instead, but no go.

jakkdl:
Oh, it looks like it worked in CI .. I think? `trio/_path.py` and `trio/_channel.py` are at 100% at least. The load of negative percentages I think is because we're back to running with `branch=True`?

PR #2656, opened 2023-05-31
fix static analysis tests on 3.12

jakkdl:
See #2646, fixes all of the errors in `test_exports.py`

Opened https://github.com/python/mypy/issues/15330, didn't seem like there was a tracking issue for ast being deprecated.

Did ~zero research on why jedi is missing additional dunders, so somebody might want to check that out and/or open an issue upstream.


A5rocks:
Codecov seems like it's being skipped on 3.12 cause the pytest step is failing. We should put an if: always() or something on it.

A5rocks (approved):

jakkdl:
Nice!

Future problem to figure out / prevent the timeouts

PR #2661, opened 2023-06-02
run mypy with --no-site-packages in export tests

jakkdl:
the mypy export tests are extremely slow, and while noodling around with them found this flag that reduces the runtime quite a bit - while not having any effect on the test (since we only care if the symbols exist, not what the types are, and that's not affected by other site packages)


jakkdl:
oh, nvm. Does not work since `trio` itself is installed as a "site package".

Issue #2662, opened 2023-06-02
mypy tests in test_exports.py are *extremely* slow.


jakkdl:
With the timeout on the test run on main https://github.com/python-trio/trio/actions/runs/5148864245

I ran pytest with `--durations=10` and realized *just* how slow the mypy tests are. On my machine they are 70% of the runtime.

running on 3.12
```
65.34s call     trio/_tests/test_exports.py::test_static_tool_sees_class_members[mypy-trio]
40.21s call     trio/_tests/test_exports.py::test_static_tool_sees_all_symbols[mypy-trio]
13.44s call     trio/_tests/test_exports.py::test_static_tool_sees_class_members[mypy-trio.abc]
6.77s call     trio/_tests/test_exports.py::test_static_tool_sees_class_members[mypy-trio.lowlevel]
4.67s call     trio/_tests/test_exports.py::test_static_tool_sees_class_members[mypy-trio.testing]
3.78s call     trio/_tests/test_unix_pipes.py::test_del
2.01s call     trio/_tests/test_timeouts.py::test_fail
2.01s call     trio/_tests/test_timeouts.py::test_sleep
1.92s call     trio/_tests/test_util.py::test_coroutine_or_error
1.52s call     trio/_tests/test_dtls.py::test_openssl_retransmit_doesnt_break_stuff
```
total time 174.26, the mypy tests sum up to 130.43, or 75%

Running on python3.11, I got 80.84/128.92=63%

Alternatives:
1. create a `@superslow` + `--run-super-slow` decorator, (or `@mypy-tests` + `--run-mypy-tests`) and only run the mypy tests on a single machine in CI.
2. optimize/rewrite the tests, so they make fewer / a single mypy call. Though even with a single call it's probably going to be somewhat slow.
3. set up a shared mypy cache across machines
4. Don't run the tests in CI at all, or like run them weekly/monthly instead of on-push. pyright + jedi export tests should catch most/all errors anyway, and are both much faster.
5. #2661 was a no-go. It's *maybe* possible to do some magic with skipping import-following, but it quickly gets quite messy because you still want to import all the trio submodules.

Fuyukai:
Alternative 6: Remove all the really stupid dynamicism, and then you won't even *have* to run this test!

A5rocks:
I believe we can snoop mypy's cache (which should be pretty stable version to version, but less guarantees). I set some time out for that earlier but got a bit distracted. I'll take a look tomorrow probably if nobody beats me to that.

[...]

PR #2664, opened 2023-06-20
Make the mypy-based tests faster!

A5rocks:
This makes *some* of the tests slower cause it stops using an (implicit) cache in CWD. This makes them generally faster, however.

Fixes https://github.com/python-trio/trio/issues/2662

Edit: huh, did not realize I pressed the "assignees" button!

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -103,6 +103,12 @@ else
     INSTALLDIR=$(python -c "import os, trio; print(os.path.dirname(trio.__file__))")
     cp ../pyproject.toml $INSTALLDIR
 
+    # TODO: remove this once we have a py.typed file
+    touch "$INSTALLDIR/py.typed"
+
+    # get mypy tests a nice cache
+    MYPYPATH=".." mypy --config-file= --cache-dir=./.mypy_cache -c "import trio" >/dev/null 2>/dev/null || true
+
     if pytest -r a -p trio._tests.pytest_plugin --junitxml=../test-results.xml --run-slow ${INSTALLDIR} --cov="$INSTALLDIR" --cov-report=xml --cov-config=../.coveragerc --verbose; then
         PASSED=true
     else
--- a/trio/_tests/test_exports.py
+++ b/trio/_tests/test_exports.py
@@ -18,6 +18,8 @@ from .. import _core, _util
 from .._core._tests.tutil import slow
 from .pytest_plugin import RUN_SLOW
 
+mypy_cache_updated = False
+
 
 def test_core_is_properly_reexported():
     # Each export from _core should be re-exported by exactly one of these
@@ -59,8 +61,8 @@ PUBLIC_MODULE_NAMES = [m.__name__ for m in PUBLIC_MODULES]
 # won't be reflected in trio.socket, and this shouldn't cause downstream test
 # runs to start failing.
 @pytest.mark.redistributors_should_skip
-# pylint/jedi often have trouble with alpha releases, where Python's internals
-# are in flux, grammar may not have settled down, etc.
+# Static analysis tools often have trouble with alpha releases, where Python's
+# internals are in flux, grammar may not have settled down, etc.
 @pytest.mark.skipif(
     sys.version_info.releaselevel == "alpha",
     reason="skip static introspection tools on Python dev/alpha releases",
@@ -72,6 +74,7 @@ PUBLIC_MODULE_NAMES = [m.__name__ for m in PUBLIC_MODULES]
     "ignore:module 'sre_constants' is deprecated:DeprecationWarning",
 )
 def test_static_tool_sees_all_symbols(tool, modname, tmpdir):
+    global mypy_cache_updated
     module = importlib.import_module(modname)
 
     def no_underscores(symbols):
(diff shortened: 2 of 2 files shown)


A5rocks:
Timeouts are annoying, I think fixing those (probably just making mypy run once without pytest timeout and share that cache whereever) should be another PR however.

trio/_tests/test_exports.py
```diff
@@ -264,111 +314,116 @@ def no_hidden(symbols):
...
+
+        # TODO: why is this? Is it a problem?
+        # see https://github.com/python-trio/trio/pull/2631#discussion_r1185615916
+        if class_ == trio.StapledStream:
+            extra.remove("receive_stream")
+            extra.remove("send_stream")
+
+        # I have not researched why these are missing, should maybe create an issue
+        # upstream with jedi
+        if tool == "jedi" and sys.version_info >= (3, 11):
```
A5rocks, line 403: In case it isn't obvious, this changed from `sys.version_info >= (3, 12)` to `sys.version_info >= (3, 11)` cause of a backport, probably.

trio/_tests/test_exports.py
```diff
@@ -106,28 +109,34 @@ def no_underscores(symbols):
...
-        )
+        cache = Path.cwd() / ".mypy_cache"
         from mypy.api import run
 
-        mypy_res = run(["--config-file=", "--follow-imports=silent", str(tmpfile)])
-
-        # check that there were no errors (exit code 0), otherwise print the errors
-        assert mypy_res[2] == 0, mypy_res[0]
+        # This pollutes the `empty` dir. Should this be changed?
+        run(["--config-file=", "--cache-dir=./.mypy_cache", "-c", f"import {modname}"])
```
jakkdl, line 116: Do you really need to run mypy for each submodule? I tested it quickly and it seems to populate cache for the whole package when just running `import trio`.

So could e.g. do something like having a global variable for having generated the cache this test run, and use that both here and in `test_static_tool_sees_class_members`.

ci.sh
```diff
@@ -100,6 +100,10 @@ else
     mkdir empty || true
     cd empty
 
+    # get mypy tests a nice cache
+    # (this ugly env var is necessary until we have a py.typed file)
+    MYPYPATH=".." mypy --config-file= --cache-dir=./.mypy_cache -c "import trio" >/dev/null 2>/dev/null || true
+
```
jakkdl, line 106: Since the long-term plan is to have one anyway, we could just temporarily create one here in `ci.sh` (and remove it at the end if it was created).

I had that previously but switched to creating it in the tests themselves so they could be run directly through pytest w/o having to manually create a `py.typed`, but not much reason not to do both.

trio/_tests/test_exports.py
```diff
@@ -179,7 +188,7 @@ def no_underscores(symbols):
 @slow
 # see comment on test_static_tool_sees_all_symbols
 @pytest.mark.redistributors_should_skip
-# pylint/jedi often have trouble with alpha releases, where Python's internals
+# jedi/mypy often have trouble with alpha releases, where Python's internals
```
jakkdl, line 191:
```python
# Static analysis tools often have [...]
```

and same on `test_static_tool_sees_all_symbols`.

trio/_tests/test_exports.py
jakkdl:
```bash
$ pwd
~/Git/trio
$ touch trio/py.typed
$ pip install .
$ mkdir empty; cd empty
$ mypy --config-file= --cache-dir=.mypy_cache -c "import trio"
$ l .mypy_cache/3.11/trio | wc -l
61
```

jakkdl (changes requested):
We could also set up a shared cache dir across the jobs - I think we can assume that mypy will generate the same cache on different OS's, and could then also use that in `check.sh`.

Not sure if https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows or https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts would be the way to go.

jakkdl:
> Timeouts are annoying, I think fixing those (probably just making mypy run once without pytest timeout and share that cache whereever) ~should be another PR however.~ NVM I need to fix that to merge this...
> 
> Edit: also I see this ended up being slower by 2 minutes... huh??

Is it still slower? Maybe mypy's cache parsing is sufficiently optimized that (only) sidestepping it doesn't actually improve it.

trio/_tests/test_exports.py
jakkdl: A global variable that tracks that the cache is updated once per pytest invocation is plenty, and on my machine validating the cache takes 0.5s - so saving the extra mypy calls means a total of ~14 seconds of time saved, which makes quite the difference imo.

trio/_tests/test_exports.py
```diff
@@ -222,14 +235,16 @@ def no_hidden(symbols):
...
         # This pollutes the `empty` dir. Should this be changed?
-        run(
-            [
-                "--config-file=",
-                "--cache-dir=./.mypy_cache",
-                "-c",
-                f"import {module_name}",
-            ]
-        )
+        if not mypy_cache_updated:
```
jakkdl, line 238: Throw in a `# pragma: no cover` here, as this will only be covered if `test_static_tool_sees_all_symbols` isn't run.

trio/_tests/test_exports.py
```diff
@@ -222,14 +235,16 @@ def no_hidden(symbols):
...
+            run(
+                [
+                    "--config-file=",
+                    "--cache-dir=./.mypy_cache",
+                    "-c",
+                    f"import {module_name}",
+                ]
+            )
+            mypy_cache_updated = False
```
jakkdl, line 247: This should be `True`, not `False`!

jakkdl (approved):
2 small changes, feel free to merge after addressing them!

trio/_tests/test_exports.py
A5rocks: Oh geh, of course!

A5rocks:
Rewrote history a bit to make the merge look neater, feel free to merge via merge commit (do the honors!)!

Issue #2645, opened 2023-05-17
use coverage instead of pytest-cov


jakkdl:
#2628 replaces `trio/_tests/conftest.py` and `trio/_core/tests/conftest.py` with a pytest plugin `trio/_tests/pytest_plugin.py` that's parsed on startup with `-p`. This means that the plugin, or anything it imports (i.e. `trio.testing`) does not get "seen" by pytest-cov, as that's loaded later.

https://stackoverflow.com/a/62224494 suggest using `coverage -m pytest [...]` instead of `pytest --cov [...]` to launch coverage before running pytest, which solves this. I started trying to fix it in https://github.com/python-trio/trio/commit/00fe706e7ae73c47097a165f04c5b3c77045dfa6 but decided to postpone it for a separate PR.

jakkdl:
@A5rocks since you played around with the coverage, could you implement this as well?

PR #2665, opened 2023-06-20
Try out coverage instead of pytest-cov

A5rocks:
Not sure this will work. Fixes https://github.com/python-trio/trio/issues/2645

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,9 @@ branch=True
 source=trio
 omit=
   setup.py
+# These are run in subprocesses, but still don't work. We follow
+# coverage's documentation to no avail.
+  trio/_core/_tests/test_multierror_scripts/*
 # Omit the generated files in trio/_core starting with _generated_
   */trio/_core/_generated_*
 # Script used to check type completeness that isn't run in tests
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,7 +5,7 @@ omit=
   setup.py
 # These are run in subprocesses, but still don't work. We follow
 # coverage's documentation to no avail.
-  trio/_core/_tests/test_multierror_scripts/*
+  */trio/_core/_tests/test_multierror_scripts/*
 # Omit the generated files in trio/_core starting with _generated_
   */trio/_core/_generated_*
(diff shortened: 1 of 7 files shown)


A5rocks:
https://github.com/python-trio/trio/pull/2664 has the fix for the Ubuntu 3.11 failure.

jakkdl (changes requested):
Nice, the code in `testing/` is showing up as covered as expected!

Seems like we've lost coverage of `test_multierror_scripts` since they're run with `subprocess.run`. But looks like there's documentation on how to fix that: https://coverage.readthedocs.io/en/7.2.7/subprocess.html
Is the drop in [trio/_core/_multierror.py](https://app.codecov.io/gh/python-trio/trio/pull/2665/blob/trio/_core/_multierror.py) also because of that?

The drop in [trio/_tests/test_exports.py](https://app.codecov.io/gh/python-trio/trio/pull/2665/blob/trio/_tests/test_exports.py) seems to be the test timeout on `3.12` so that's not an issue.

A5rocks:
Wait a second, `trio/_core/_tests/test_multierror_scripts/_common.py` literally does the thing coverage.py recommends doing. It still doesn't work. What??

A5rocks:
of *course* the thing that literally reads the file is not going to work. why did i miss that...

A5rocks:
welp no clue why the coverage for a few more files isn't working. i give up for now.

jakkdl:
> welp no clue why the coverage for a few more files isn't working. i give up for now.

I suppose we can just exclude `trio/_core/_tests/test_multierror_scripts/*` from coverage and leave fighting with this for some future crazy person.

jakkdl:
Huh, there's *four* codecov checks? :o 

Anyway, it looks good now - merging~~

A5rocks:
> Huh, there's _four_ codecov checks? :o

... weird. Eh, we can fix that eventually!

A5rocks:
Ok, the four codecov thing was just cause this was an old PR: https://github.com/python-trio/trio/pull/2674 only has 2 checks! (as it should be)

PR #2666, opened 2023-06-23
xfail test_nursery_cancel_doesnt_create_cyclic_garbage on 3.12

jakkdl:
Marks the remaining fail in #2646 as xfail so CI tests can fully "succeed" and status goes back to indicating whether tests passed or not.

Also removes a couple unused imports in the file

### DIFF
--- a/trio/_core/_tests/test_run.py
+++ b/trio/_core/_tests/test_run.py
@@ -2279,6 +2279,10 @@ async def test_cancel_scope_exit_doesnt_create_cyclic_garbage():
         gc.garbage.clear()
 
 
+@pytest.mark.xfail(
+    sys.version_info >= (3, 12),
+    reason="Waiting on https://github.com/python/cpython/issues/100964",
+)
 @pytest.mark.skipif(
     sys.implementation.name != "cpython", reason="Only makes sense with refcounting GC"
 )


jakkdl:
3.11 fail is fixed in #2664 
3.12 timeout fail is fixed in #2665 
codecov fail is due to the 3.12 timeout segfault
macOS 3.7 fail is a new one to me, but looks like an intermittent error?

trio/_core/_tests/test_run.py
```diff
@@ -1,18 +1,14 @@
...
 import types
-import warnings
 import weakref
-from contextlib import contextmanager, ExitStack
+from contextlib import ExitStack
 from math import inf
-from textwrap import dedent
 import gc
 
-import attr
```
A5rocks, line 15: All the removed imports makes me wonder if this happens for any other test files! Would be neat to clean those up in another PR, IMO.

trio/_core/_tests/test_run.py
```diff
@@ -2279,6 +2275,10 @@ async def crasher():
         gc.garbage.clear()
 
 
+@pytest.mark.xfail(
+    sys.version_info >= (3, 12),
+    reason="https://github.com/python/cpython/issues/100964",
```
A5rocks, line 2280: I'd love a tiny bit more elaboration: "waiting on <issue link>" maybe.

A5rocks (approved):
Looks good, slight feedback re: elaborating on the xfail reason.

trio/_core/_tests/test_run.py
jakkdl: Oh for sure, imports are kind of a mess in this project - which is kind of bound to happen without `isort` or any linter checking for it. See https://github.com/python-trio/trio/pull/2538

```bash
$ flake8 trio/_tests trio/_core/_tests --select F401 | wc -l
45
```
Some of them are false alarms, but vast majority aren't. There's a bunch in non-test files as well, but you'd need rules to exclude the ones used for exporting public symbols.

jakkdl:
Undid changes to imports, and elaborated reason and commit message.

jakkdl:
Needs #2664 so the check passes, otherwise I'd merge this now.

trio/_core/_tests/test_run.py
jakkdl: Oh, `F401` is explicitly disabled in check.sh
Well, I'll just make a PR that does per-file disables instead.

PR #2667, opened 2023-06-23
timeout functions now raise ValueError on NaN inputs

### LINKED ISSUE #2493: trio.sleep(float('nan')) is valid and sleeps forever
`trio.sleep(float('nan'))` currently sleeps forever. Fixing it is pretty trivial by reordering the if statements but there's maybe other stuff that would break with NaN, and it might not be worth fixing a theoretical issue that idk if it might ever happen in reality - though I could imagine it biting somebody somewhere.

for reference: `time.sleep()` gives a ValueError, `asyncio.sleep()` sleeps for 0

jakkdl:
Fixes #2493

Consolidated all `ValueError` tests into a single test.

### DIFF
--- a/newsfragments/2493.breaking.rst
+++ b/newsfragments/2493.breaking.rst
@@ -1,0 +1,1 @@
+Timeout functions now raise `ValueError` if passed `math.nan`. This includes `trio.sleep`, `trio.sleep_until`, `trio.move_on_at`, `trio.move_on_after`, `trio.fail_at` and `trio.fail_after`.
--- a/trio/_timeouts.py
+++ b/trio/_timeouts.py
@@ -1,3 +1,4 @@
+import math
 from contextlib import contextmanager
 
 import trio
@@ -10,7 +11,12 @@ def move_on_at(deadline):
     Args:
       deadline (float): The deadline.
 
+    Raises:
+      ValueError: if deadline is NaN.
+
     """
+    if math.isnan(deadline):
+        raise ValueError("deadline must not be NaN")
     return trio.CancelScope(deadline=deadline)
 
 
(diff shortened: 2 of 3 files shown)


trio/_timeouts.py
```diff
@@ -10,7 +11,12 @@ def move_on_at(deadline):
     Args:
       deadline (float): The deadline.
 
+    Raises:
+      ValueError: if deadline is NaN.
+
     """
+    if math.isnan(deadline):
+        raise ValueError("deadline must not be NaN")
```
A5rocks, line 19: Alternatively, "deadline must be a number" but that's ambiguous w/r/t infinity. I think this error message works best, just leaving this here cause... I don't know why.

A5rocks (approved):
Looks good. 1 genuine comment and 2 filler comments.

A5rocks (changes requested):
Actually, I just remembered changes like this should *probably* get a newsfragment!

trio/_timeouts.py
```diff
@@ -119,10 +131,13 @@ def fail_after(seconds):
     it's caught and discarded. When it reaches :func:`fail_after`, then it's
     caught and :exc:`TooSlowError` is raised in its place.
 
+    Args:
+      seconds (float): The timeout.
+
```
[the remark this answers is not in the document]
jakkdl: This sounds like something that can be auto-generated / have the signature displayed in another manner in the doc, once the public interface is fully typed.

jakkdl:
> Actually, I just remembered changes like this should _probably_ get a newsfragment!

Hah! I've been working so much solely on internal stuff I completely forgot this was a thing, good catch!

Feel free to merge if it looks good otherwise :rocket:

newsfragments/2493.breaking.rst
```diff
@@ -0,0 +1 @@
+Timeout functions now raise `ValueError` if passed `nan`. This includes `trio.sleep`, `trio.sleep_until`, `trio.move_on_at`, `trio.move_on_after`, `trio.fail_at` and `trio.fail_after`.
```
A5rocks, line 1: I'd rather "NaN" over "nan"

A5rocks (approved):
Small nitpick. `breaking` *does* feel a bit far out but honestly that is probably the best category.

I'll merge (or you can) after blocking CI is fixed.

PR #2669, opened 2023-06-25
Enable F401 in flake8, configure per-file disables.

jakkdl:
See discussion in https://github.com/python-trio/trio/pull/2666#discussion_r1240448036


trio/_core/_tests/test_run.py
```diff
@@ -1598,7 +1594,7 @@ async def test_asyncio_function_inside_nursery_does_not_explode():
     # Regression test for https://github.com/python-trio/trio/issues/552
     with pytest.raises(TypeError) as excinfo:
         async with _core.open_nursery() as nursery:
-            import asyncio
+            # import asyncio
```
jakkdl, line 1597: I'm not sure if this `import` is to ensure `asyncio` is available or something

A5rocks: I think this is just missed in PR review. cc @oremanj as author of https://github.com/python-trio/trio/pull/1178 even though that was.... 4 years ago wow.

jakkdl:
Any objection to just merging this? It'll quickly lead to a lot of merge conflicts if I start opening type-checking PRs in parallel to it.

.flake8
A5rocks, line 1: How about using `flake8-pyproject` to consolidate all our configuration into 1 file?

trio/_core/_tests/test_util.py
```diff
@@ -1 +1 @@
-import pytest
+import pytest  # noqa: F401
```
A5rocks, line 1: Hm, why is this staying?

A5rocks (commented):
Looks mostly good. I don't quite get why anything needs to have this warning ignored for (unless it's doing some runtime stuff? which kinda sucks but ig that's a different issue to fix)

trio/_core/_tests/test_util.py
jakkdl: I assumed that importing pytest was just what the test *did* (for some strange reason), but looking at the history of the file it seems like it should probably have been removed a long time ago.

jakkdl:
> Looks mostly good. I don't quite get why anything needs to have this warning ignored for (unless it's doing some runtime stuff? which kinda sucks but ig that's a different issue to fix)

added comments to most of the noqa's.

check.sh
```diff
@@ -18,10 +18,8 @@ if ! black --check setup.py trio; then
     black --diff setup.py trio
 fi
 
-# Run flake8 without pycodestyle and import-related errors
-flake8 trio/ \
-    --ignore=D,E,W,F401,F403,F405,F821,F822\
-    || EXIT_STATUS=$?
+# Run flake8, configured in .flake8
```
A5rocks, line 21: Forgot to update the comment!

pyproject.toml
```diff
@@ -1,6 +1,18 @@
 [tool.black]
 target-version = ['py37']
 
+[tool.flake8]
+extend-ignore = ['D','E','W','F403','F405','F821','F822']
```
A5rocks, line 5: Eh, mind adding spacing?

A5rocks (approved):
Other than these 2 + fixing the formatting, this looks good!

pyproject.toml
jakkdl: Ran `pyproject.toml` through https://github.com/macisamuele/language-formatters-pre-commit-hooks

A5rocks:
Formatting CI is still failing :(

I would be a +1 on adding a nicer error message (e.g. putting a `git diff *-requirements.txt` or something) in the failure case for pip-compile.

jakkdl:
I'm inclined to ignore the failing pypy-3.7 export symbol test since support will be dropped soon anyway.

A5rocks:
Could you just toss an xfail at it I guess, it's required for merging :(

A5rocks:
https://github.com/python/typing_extensions/issues/258 heh. I'll make a PR.

A5rocks (approved):
Rereviewing, seems good to me.

PR #2671, opened 2023-06-26
Type completeness improvement (timeouts, CancelScope, and more)

jakkdl:
Just going down the list of untyped symbols, don't think there's much anything tricky. Will run trio-typing/stubtest against the changes


jakkdl:
Didn't see anything of interest running `stubtest` - although `trio-stubs` are plenty out of date and messy anyway so idk if it says much.

@A5rocks what is the ideal size for pure typing PRs like this, in your opinion? I could easily continue and chug through tons of symbols, but don't want to saddle you with huge code reviews.

trio/_tests/verify_types.json
```diff
@@ -45,54 +45,32 @@
...
-      "trio.move_on_after",
-      "trio.sleep_forever",
-      "trio.sleep_until",
-      "trio.sleep",
-      "trio.fail_after",
-      "trio._sync.Event",
       "trio._sync.Event.is_set",
       "trio._sync.Event.wait",
       "trio._sync.Event.statistics",
-      "trio._sync.CapacityLimiter",
```
jakkdl, line 88: Most/all of these random classes being type complete is due to `Final.__new__` getting typed.

A5rocks:
I don't really have a preferred size given I'm not *too* used to typing PRs. I'd kinda like smaller ones, I think. Maybe an exported function and everything it depends on or about that much length in diffs?

(Basically I don't really think I care about length, just about having the context to check types!)

jakkdl:
huh, what's going on with codecov?

trio/_timeouts.py
```diff
@@ -1,10 +1,11 @@
 import math
 from contextlib import contextmanager
+from typing import Iterator, ContextManager
```
A5rocks, line 3: Hm. You're probably aware, but as of [Python 3.9](https://peps.python.org/pep-0585/), `collections.abc` is to be preferred. I'm decently sure it's fine to stick to `typing` versions for now and use [pyupgrade](https://github.com/asottile/pyupgrade) when dropping 3.8 so that we get the new non-deprecated syntax?

trio/_util.py
```diff
@@ -280,7 +281,9 @@ class SomeClass(metaclass=Final):
     - TypeError if a subclass is created
     """
 
-    def __new__(cls, name, bases, cls_namespace):
+    def __new__(
+        cls, name: str, bases: tuple[type, ...], cls_namespace: dict[str, t.Any]
```
A5rocks, line 285: Personally I dislike `Any` except when totally necessary: can we use `object` here?

trio/_util.py
```diff
@@ -312,7 +315,7 @@ class SomeClass(metaclass=NoPublicConstructor):
     - TypeError if a subclass or an instance is created.
     """
 
-    def __call__(cls, *args, **kwargs):
+    def __call__(cls, *args: t.Any, **kwargs: t.Any) -> None:
```
A5rocks, line 318: Same as above, does `object` work here?

trio/_util.py
```diff
@@ -1,4 +1,5 @@
 # Little utilities we use internally
+from __future__ import annotations
```
A5rocks, line 2: Should we preemptively stick this at the top of all files so that these kinds of change is already done?

trio/_timeouts.py
A5rocks: Speaking of which can you run `pyupgrade` on your PR dropping 3.7? It might find some stuff!

trio/_util.py
A5rocks: Actually IDK how I feel about using the cooler syntax early: we could use `typing.List` etc. and already-mentioned `pyupgrade` will drop that for newer syntax the moment we drop 3.8.

On the other hand it's more readable and more understandable. I guess `List[int]` vs `list[int]` (using `from typing import ...`) isn't too big a difference.

jakkdl: The biggest downside to using `typing.List` is the additional overhead in import management - which is fully manual given our current lack of `isort`. This can be quite frustrating IME.

idc much whether to preemptively adding it

trio/_timeouts.py
jakkdl: I changed them out - and might stick to collections.abc/contextlib mostly for the sake of getting used to the habit, but yeah don't think it really matters.

trio/_util.py
jakkdl: I've liked using https://pypi.org/project/flake8-type-checking/ (or https://pypi.org/project/ruff/ ) for managing this.

trio/_timeouts.py
TeamSpen210: There doesn’t need to be any rush to switch over - `typing` isn’t planning on removing these for a long while, given they’re just aliases.

A5rocks: Oh yeah maybe I wasn't completely clear: I'm aok on using `typing` versions especially since there's a tool that will automatically switch them out, just as long as we're aware and we run that tool when we drop 3.8.

trio/_core/_run.py
```diff
@@ -890,7 +898,9 @@ def __exit__(self):  # pragma: no cover
         assert False, """Never called, but should be defined"""
 
 
-def open_nursery(strict_exception_groups=None):
+def open_nursery(
+    strict_exception_groups: bool | None = None,
+) -> AbstractAsyncContextManager[Nursery]:
```
Fuyukai, line 903: why not ``typing.AsyncContextManager``?

trio/_timeouts.py
```diff
@@ -92,7 +94,7 @@ class TooSlowError(Exception):
 
 
 @contextmanager
-def fail_at(deadline):
+def fail_at(deadline: float) -> Iterator[trio.CancelScope]:
```
Fuyukai, line 97: iterattor typehint makes this useless for autocomplete.

trio/_timeouts.py
```diff
@@ -120,7 +122,7 @@ def fail_at(deadline):
         raise TooSlowError
 
 
-def fail_after(seconds):
+def fail_after(seconds: float) -> AbstractContextManager[trio.CancelScope]:
```
Fuyukai, line 125: same thing here wrt asynccontextmanager

trio/_util.py
Fuyukai: there's never a reason to not use ``from __future__ import annotations`` so you might as well use 3.11+ syntax wherever possible.

Fuyukai (changes requested):

trio/_util.py
Fuyukai: fucking github. anyway there's never a reason not to use ``from __future__ import annotations`` so you might as well just use 3.11+ typing syntax always.

trio/_timeouts.py
TeamSpen210: `Iterator` is the correct type here, or alternatively `typing.Generator[trio.CancelScope, None, None]`. The function itself doesn't return a CM, it returns an iterator. `contextmanager` then transforms the function into one returning a CM. If the autocomplete tool doesn't have a special case for `contextmanager`, it's more a problem there...

Fuyukai: I don't care what tthe type checker says. I use type hints for autocomplete.

trio/_core/_run.py
smurfix: NB: `var: bool | None = None`  is redundant in that you can skip the `| None` part. Type checkers infer that from the `None` default.

TeamSpen210: The `typing` version is simply an alias of the `contextlib` one after 3.9 / with `__future__.annotations` enabled.

TeamSpen210: The `None` thing was originally the case, but more recently checkers have preferred that you do it explicitly, so there's less special cases in the rules.

trio/_timeouts.py
jakkdl: What program are you using for autocomplete?
Could possibly define a different signature within `if TYPE_CHECKING` if it doesn't support decorators/`contextmanager` and is used widely enough to warrant special casing.

trio/_core/_run.py
jakkdl: yeah it confused me a bit that `typing.ContextManager` is not `contextlib.contextmanager`, but rather `contextlib.AbstractContextManager`.
https://docs.python.org/3/library/typing.html#typing.ContextManager

jakkdl: @Fuyukai Is the usage of this alias problematic for your tooling, or just a matter of curiosity?

trio/_util.py
jakkdl: https://github.com/Zac-HD/flake8-trio/pull/78#pullrequestreview-1208679518
https://github.com/Instagram/LibCST/issues/870
There's some scenarios where it matters, but I think we're fine.

trio/_core/_run.py
Fuyukai: Nope, just curiosity.

trio/_timeouts.py
Fuyukai: I use PyCharm (relevant [bug](https://youtrack.jetbrains.com/issue/PY-36444/PyCharm-doesnt-infer-types-when-using-contextlib.contextmanager-decorator). Doing an ``if TYPE_CHECKING:`` and then having an empty function that returns ``AsyncContextManager`` instead would work out great.

trio/_timeouts.py
```diff
@@ -120,7 +124,18 @@ def fail_at(deadline):
...
-def fail_after(seconds):
+# workaround for PyCharm not being able to infer return type from @contextmanager
+# see https://youtrack.jetbrains.com/issue/PY-36444/PyCharm-doesnt-infer-types-when-using-contextlib.contextmanager-decorator
+if TYPE_CHECKING:
+
+    def fail_at(deadline: float) -> AbstractContextManager[trio.CancelScope]:
+        return _fail_at(deadline)
+
+else:
+    fail_at = _fail_at
```
jakkdl, line 135: The suggested workaround in https://youtrack.jetbrains.com/issue/PY-36444/PyCharm-doesnt-infer-types-when-using-contextlib.contextmanager-decorator is ugly as shit:

```python
def successful_workaround_using_wrapper_function_annotated_returning_context_manager():
    """PyCharm autocompletion works and mypy is also happy, however this is an ugly workaround"""

    def wrapper() -> ContextManager[str]:
        @contextmanager
        def generator_function():
            yield "some value"

        return generator_function()

    with wrapper() as value:
        print(value.upper())  # PyCharm autocompletion works
```

I think this one is much better, it allows both pycharm autocompletion as well as typechecking of the original function, doesn't incur any runtime penalt, and is relatively small and unobtrusive

TeamSpen210: I'd suggest setting `__name__` to `fail_at`, or just move the function under the `else:` block. The type checking version can just have a body of `...`.

jakkdl: ah, we're losing the docstring

TeamSpen210: Another way to do it perhaps:
```py
def fail_at(...) -> ContextManager[trio.CancelScope]: # type: ignore
    ...

if not TYPE_CHECKING:
    fail_at = contextmanager(fail_at)
```
Just hide the decorator part from the type checker, though we need an ignore comment since Mypy would rightly complain that the return type is wrong.

jakkdl: > I'd suggest setting `__name__` to `fail_at`, or just move the function under the `else:` block. The type checking version can just have a body of `...`.

I was thinking the docstring need to be accessible both during runtime and during static analysis

jakkdl: I also came up with:
```python
# workaround for PyCharm not being able to infer return type from @contextmanager
# see https://youtrack.jetbrains.com/issue/PY-36444/PyCharm-doesnt-infer-types-when-using-contextlib.contextmanager-decorator
@contextmanager
def _fail_at(deadline: float) -> Iterator[trio.CancelScope]:
    with move_on_at(deadline) as scope:
        yield scope
    if scope.cancelled_caught:
        raise TooSlowError

def fail_at(deadline: float) -> AbstractContextManager[trio.CancelScope]:
    """Creates a cancel scope with the given deadline, and raises an error if it
    ...

    """
    return _fail_at(deadline)
```

jakkdl: > Another way to do it perhaps:
> 
> ```python
> def fail_at(...) -> ContextManager[trio.CancelScope]: # type: ignore
>     ...
> 
> if not TYPE_CHECKING:
>     fail_at = contextmanager(fail_at)
> ```
> 
> Just hide the decorator part from the type checker, though we need an ignore comment since Mypy would rightly complain that the return type is wrong.

this one seems pretty clean though, and doesn't introduce another frame like the above

Fuyukai (approved):

Fuyukai:
That's my comments addressed fine.

TeamSpen210 (approved):

PR #2673, opened 2023-06-28
Fix typo

karosis88:


A5rocks (changes requested):
Mind changing this to be an f-string? Not sure why it wasn't initially.

jakkdl:
Thanks @karosis88! :rocket:

PR #2676, opened 2023-06-30
Prevent typing_extensions 4.7.0

A5rocks:
This is making our static tool tests fail.

### DIFF
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -23,7 +23,7 @@ pip-tools >= 6.13.0
 # typed_ast is deprecated as of 3.8, and straight up doesn't compile on 3.10-dev as of 2021-12-13
 typed_ast; implementation_name == "cpython" and python_version < "3.8"
 mypy-extensions; implementation_name == "cpython"
-typing-extensions; implementation_name == "cpython"
+typing-extensions < 4.7.0
 
 # Trio's own dependencies
 cffi; os_name == "nt"
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -39,3 +39,6 @@ isort < 5.12.0
 
 # cryptography 40.0.2 (and presumably prior) segfaults on PyPy 3.7
 cryptography < 40.0.0
+
+# typing_extensions 4.7.0 breaks pylint on pypy 3.7
+typing_extensions < 4.7.0
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -39,6 +39,3 @@ isort < 5.12.0
 
 # cryptography 40.0.2 (and presumably prior) segfaults on PyPy 3.7
 cryptography < 40.0.0
-
-# typing_extensions 4.7.0 breaks pylint on pypy 3.7
-typing_extensions < 4.7.0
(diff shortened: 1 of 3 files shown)


A5rocks:
Oh duh, CI doesn't read the actual requirements file, only the locked one. And for the locked one, pip-tools doesn't realize that it should drop the cpython-only bit?

Anyways looks like typing_extensions might make a bugfix release. Otherwise I'll uh, idk, xfail the test? We'll see in a day.

jakkdl (approved):

jakkdl (commented):
Actually, just mark it as `skip` and we can merge it and go ahead with merging other PRs. We don't need to care about whether this gets fixed upstream or not

jakkdl:
> And for the locked one, pip-tools doesn't realize that it should drop the cpython-only bit?

if you remove/comment out the first time `typing_extensions` is listed on line 26 it works. Playing around with it, it does not seem to be able to generate multiple different requirements in `requirements.txt` for the same package even if different lines in `test-requirements.in` are explicitly mutually exclusive.

Fuyukai:
Why are we even testing PyPy 3.7? As far as I can tell they're not even releasing 3.7 or 3.8 builds anymore.

jakkdl:
> Why are we even testing PyPy 3.7? As far as I can tell they're not even releasing 3.7 or 3.8 builds anymore.

we're dropping 3.7 support soon #2668 - so it's just a matter of it not having been dropped yet.

Fuyukai:
But PyPy support is different to CPython support. PyPy dropped 3.8 already, and it dropped 3.7 quite a while ago.

jakkdl:
Was surprisingly tricky to find what versions are supported, but yeah looks like it's [3.9 and 3.10](https://www.pypy.org/download_advanced.html). I'll drop pypy-3.8 as well in #2668 and add pypy-3.10 as well, or do that in a different PR in case it's messy.

A5rocks:
I'm just going to wait a little before merging this to see if typing extensions releases a bugfix to fix this (they have a pr open on their side and I *think* they will?)

A5rocks:
> Why are we even testing PyPy 3.7? As far as I can tell they're not even releasing 3.7 or 3.8 builds anymore.

Let's see what the PyPy maintainers recommend testing against! Here's where the question was asked: https://foss.heptapod.net/pypy/pypy/-/issues/3958#note_299941

A5rocks:
Well, they merged in the bugfix 6 hours ago... but haven't made a release. Time window's up!

A5rocks:
@jakkdl let's revert this in https://github.com/python-trio/trio/pull/2668

Additionally, could you update the README in that to say we're only going to support versions of PyPy PyPy itself still supports? (Namely 3.9 and 3.10 atm) See: https://foss.heptapod.net/pypy/pypy/-/issues/3958#note_299944

jakkdl:
> @jakkdl let's revert this in #2668
> 
> Additionally, could you update the README in that to say we're only going to support versions of PyPy PyPy itself still supports? (Namely 3.9 and 3.10 atm) See: https://foss.heptapod.net/pypy/pypy/-/issues/3958#note_299944

It does currently say:
> **Cool, but will it work on my system?** Probably! As long as you have
some kind of Python 3.8-or-better (CPython or currently maintaned PyPy3 are
both fine), and are using Linux, macOS, Windows, or FreeBSD, then Trio
will work.

but I reworded it slightly

Issue #2655, opened 2023-05-30
Please make a new release


mgorny:
Could you please make a new release? I'm specifically looking for the `async_generator` dependency being removed, as that should be good enough to enable Python 3.12 support in the Gentoo package.

mgorny:
I've seen only 1 new test fail on Gentoo compared to Python 3.11  (`trio/_core/_tests/test_run.py::test_nursery_cancel_doesnt_create_cyclic_garbage`). I'm yet to see in greater detail but it seems good enough for httpx at a first glance.

I'm currently playing with a snapshot from current git, so hopefully that'll be good enough before the next release.

bastimeyer:
Hi, would it be possible to publish a new release before waiting for the support end of Python 3.7? I don't care about 3.7, but the last release of trio was in September of last year, which is quite a while.

I'm specifically looking for support for generic typing of `MemorySendChannel`/`MemoryReceiveChannel` which was added in #2549 in February.

The most recent release of `trio-typing` already defines

```py
class MemorySendChannel(trio.abc.SendChannel[T_contra]): ...
class MemoryReceiveChannel(trio.abc.ReceiveChannel[T_co]): ...
```

This just bit me a bit as a new user of trio, because I thought that this was supported. Right now I get this error when trying to implement generic memory channels:

> TypeError: <class 'trio.MemorySendChannel'> is not a generic class

```py
class EventListener(Generic[TEvent]):
    sender: trio.MemorySendChannel[TEvent]
    receiver: trio.MemoryReceiveChannel[TEvent]
    ...
```

Installing trio from git is not an option for me, so I'm kindly asking for a new release. Thanks!

oremanj:
3.7 EOL is in two weeks. I won't personally be able to roll a new release before that, but maybe someone else will.

You can use `from __future__ import annotations` to resolve the MemoryXyzChannel-not-generic error.

A5rocks:
Only https://github.com/python-trio/trio/pull/2674!

A5rocks:
Oh yeah, and pylint failing on pypy-3.7 isn't great, since the required checks need to pass before releasing. I've left a comment on https://github.com/python-trio/trio/pull/2669 cause I think you could xfail it there or maybe you (or I) could make a seperate PR.

jakkdl:
:rocket: :rocket: :rocket:

mgorny:
Thanks!

Issue #2485, opened 2022-11-26
Incorrect version in trio/_version.py on v0.22.0 tag


mgorny:
When checking out trio from the `v0.22.0` tag, https://github.com/python-trio/trio/blob/v0.22.0/trio/_version.py incorrectly states the version as:

```
__version__ = "0.21.0+dev"
```

Noticed this because we use GitHub archives for packaging on Gentoo, and trio ended up being installed as `0.21.0+dev`, therefore breaking version constraints in other packages.

pquentin:
Ouch, sorry about that. I'll make an emergency v0.22.1 release.

mgorny:
Well, it's not a big deal for us (I've just `sed`-ed it) but I don't know if anybody else's impacted.

Issue #2583, opened 2023-03-01
0.22.0: version not updated


kloczek:
Looks like `trio.__vesion__` has not been updated before tagging
https://github.com/python-trio/trio/blob/v0.22.0/trio/_version.py#L1-L3

A5rocks:
Just a note that the published release has the correct `trio.__version__`, just that tag does not have it.

Do you have a use case that requires this? I'm sure we can find a way but at first glance I'm not quite sure how we can update a tag like that -- maybe make a `v0.22.0` branch and add a fixing commit?

kloczek:
Yes however I'm using VCS tar ball becaue I want to have guarantee that on top of release tree it would be possible to apply any patches from git commmits 😋 
Just please remeber about update version before tagging (best woild be if you will be using Ci target to make release (chechout tree, update version and than tag and push that tag to github automatically). 😃

pquentin:
This is a duplicate of https://github.com/python-trio/trio/issues/2485. I did not get to make a release, would you like to give it a try @A5rocks?

A5rocks:
I'd love to make a release after a few of @richardsheridan's PRs land (and I suppose that would be a minor version bump?)

PR #2680, opened 2023-07-02
Release 0.22.1

A5rocks:
- Fixes https://github.com/python-trio/trio/issues/2655
- Fixes https://github.com/python-trio/trio/issues/2485 (and duplicate https://github.com/python-trio/trio/issues/2583)

### DIFF
--- a/trio/tests.py
+++ b/trio/tests.py
@@ -7,7 +7,7 @@ from ._deprecate import warn_deprecated
 
 warn_deprecated(
     "trio.tests",
-    "0.24.0",
+    "0.22.1",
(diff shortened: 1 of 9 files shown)


trio/tests.py
```diff
@@ -7,7 +7,7 @@
 
 warn_deprecated(
     "trio.tests",
-    "0.24.0",
+    "0.22.1",
```
A5rocks, line 10: I accidentally told @jakkdl to do this, not realizing this is for the version the function is deprecated in! (same for the other occurance).

PR #2682, opened 2023-07-03
lots of typing improvements

jakkdl:
Lots and lots of trivial typehints, but a couple tricky ones hidden inside the huge diff >.>
I'll point some of them out in a review of my own.


trio/_core/_run.py
```diff
@@ -889,18 +898,18 @@ async def __aexit__(self, etype, exc, tb):
...
+    def __exit__(self) -> None:  # pragma: no cover
         assert False, """Never called, but should be defined"""
 
 
 def open_nursery(
     strict_exception_groups: bool | None = None,
-) -> AbstractAsyncContextManager[Nursery]:
+) -> NurseryManager:
     """Returns an async context manager which must be used to create a
     new `Nursery`.
```
jakkdl, line 914: Not entirely sure about this one.

trio/_core/_run.py
```diff
@@ -963,17 +977,17 @@ def __init__(self, parent_task, cancel_scope, strict_exception_groups):
...
     @property
-    def parent_task(self):
+    def parent_task(self) -> Task:
         "(`~trio.lowlevel.Task`):  The Task that opened this nursery."
         return self._parent_task
 
-    def _add_exc(self, exc):
+    def _add_exc(self, exc: BaseException) -> None:
         self._pending_excs.append(exc)
         self.cancel_scope.cancel()
```
jakkdl, line 992: qualified guess that this was supposed to be `BaseException`

trio/_core/_run.py
```diff
@@ -1146,12 +1160,11 @@ def __del__(self):
 
 @attr.s(eq=False, hash=False, repr=False, slots=True)
 class Task(metaclass=NoPublicConstructor):
-    _parent_nursery = attr.ib()
-    coro = attr.ib()
+    _parent_nursery: Nursery | None = attr.ib()
+    coro: Coroutine[Any, outcome.Outcome[object], Any] = attr.ib()
```
jakkdl, line 1164: from `trio-typing`

trio/_core/_run.py
```diff
@@ -1167,24 +1180,27 @@ class Task(metaclass=NoPublicConstructor):
     # Tasks start out unscheduled.
     _next_send_fn = attr.ib(default=None)
     _next_send = attr.ib(default=None)
-    _abort_func = attr.ib(default=None)
-    custom_sleep_data = attr.ib(default=None)
+    _abort_func: Callable[[Callable[[], NoReturn]], Abort] | None = attr.ib(
+        default=None
```
jakkdl, line 1184: from `trio-typing`

trio/_core/_run.py
```diff
@@ -1167,24 +1180,27 @@ class Task(metaclass=NoPublicConstructor):
...
+    )
+    # could possible be set with a TypeVar
+    custom_sleep_data: Any = attr.ib(default=None)
```
jakkdl, line 1187: see comment

trio/_core/_run.py
```diff
@@ -1240,11 +1256,11 @@ def print_stack_for_task(task):
             if hasattr(coro, "cr_frame"):
                 # A real coroutine
                 yield coro.cr_frame, coro.cr_frame.f_lineno
-                coro = coro.cr_await
+                coro = coro.cr_await  # type: ignore # TODO
             elif hasattr(coro, "gi_frame"):
                 # A generator decorated with @types.coroutine
                 yield coro.gi_frame, coro.gi_frame.f_lineno
-                coro = coro.gi_yieldfrom
+                coro = coro.gi_yieldfrom  # type: ignore # TODO
```
jakkdl, line 1263: idk if these are an issue, but since the signature is now typed the type checker complains about them.

trio/_core/_run.py
```diff
@@ -1279,12 +1295,14 @@ def _activate_cancel_status(self, cancel_status):
...
-    def _attempt_abort(self, raise_cancel):
+    def _attempt_abort(self, raise_cancel: Callable[[], NoReturn]) -> None:
         # Either the abort succeeds, in which case we will reschedule the
         # task, or else it fails, in which case it will worry about
         # rescheduling itself (hopefully eventually calling reraise to raise
         # the given exception, but not necessarily).
-        success = self._abort_func(raise_cancel)
+
+        # TODO: what happens if _abort_func is None?
+        success = self._abort_func(raise_cancel)  # type: ignore
```
jakkdl, line 1305: `_abort_func` can clearly be None - since it's explicitly set to be `None` in a few lines. But this seems like a possibly nasty thing to `# type: ignore`.

trio/_core/_run.py
```diff
@@ -2435,17 +2453,17 @@ def unrolled_run(
 
 
 class _TaskStatusIgnored:
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "TASK_STATUS_IGNORED"
 
-    def started(self, value=None):
+    def started(self, value: None = None):
         pass
```
jakkdl, line 2460: what's this parameter for anyway?

trio/_highlevel_generic.py
```diff
@@ -72,34 +77,34 @@ class StapledStream(HalfCloseableStream, metaclass=Final):
 
     """
 
-    send_stream = attr.ib()
-    receive_stream = attr.ib()
+    send_stream: SendStream = attr.ib()
+    receive_stream: ReceiveStream = attr.ib()
 
-    async def send_all(self, data):
+    async def send_all(self, data: bytes | bytearray | memoryview) -> None:
```
jakkdl, line 83: docstring of `SendStream.send_all` says `bytes|bytearray|memoryview`. `trio-typing` says `bytes|memoryview`.

trio/_highlevel_generic.py
```diff
@@ -72,34 +77,34 @@ class StapledStream(HalfCloseableStream, metaclass=Final):
...
 
         """
         if hasattr(self.send_stream, "send_eof"):
-            return await self.send_stream.send_eof()
+            return await self.send_stream.send_eof()  # type: ignore
         else:
             return await self.send_stream.aclose()
 
-    async def receive_some(self, max_bytes=None):
+    async def receive_some(self, max_bytes: int | None = None) -> bytes | bytearray:
```
jakkdl, line 103: `trio-typing` says *only* bytes. But given that it didn't include `bytearray` above, I guessed that should maybe be here as well?
And what about `memoryview`?

trio/_socket.py
```diff
@@ -472,44 +478,45 @@ def __getattr__(self, name):
             return getattr(self._sock, name)
         raise AttributeError(name)
 
-    def __dir__(self):
-        return super().__dir__() + list(self._forward)
+    def __dir__(self) -> Iterable[str]:
+        yield from super().__dir__()
+        yield from self._forward
```
jakkdl, line 483: `dir()` isn't guaranteed to return a list - so seems bad to rely on that.

trio/_socket.py
```diff
@@ -472,44 +478,45 @@ def __getattr__(self, name):
...
         return _SocketType(self._sock.dup())
 
-    def close(self):
+    def close(self) -> None:
         if self._sock.fileno() != -1:
             trio.lowlevel.notify_closing(self._sock)
             self._sock.close()
 
-    async def bind(self, address):
+    async def bind(self, address: tuple[Any, ...] | str | bytes) -> None:
```
jakkdl, line 519: from `trio-typing`

trio/_sync.py
```diff
@@ -89,20 +96,20 @@ def statistics(self):
...
-    async def __aenter__(self):
-        await self.acquire()
+    async def __aenter__(self) -> None:
+        await self.acquire()  # type: ignore[attr-defined]
 
     @enable_ki_protection
-    async def __aexit__(self, *args):
-        self.release()
+    async def __aexit__(self, *args: object) -> None:
+        self.release()  # type: ignore[attr-defined]
```
jakkdl, line 104: slightly illegal `type: ignore`s, but would probably require some significant refactoring to get rid of.

jakkdl:
whoa, nasty CI fails. Will look at that tomorrow or so

trio/_abc.py
```diff
@@ -412,10 +414,10 @@ async def receive_some(self, max_bytes=None):
 
         """
 
-    def __aiter__(self):
+    def __aiter__(self) -> ReceiveStream:
```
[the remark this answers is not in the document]
Fuyukai: ... so why doesn't it return Self?

trio/_core/_parking_lot.py
```diff
@@ -98,13 +106,13 @@ class ParkingLot(metaclass=Final):
 
     # {task: None}, we just want a deque where we can quickly delete random
     # items
-    _parked = attr.ib(factory=OrderedDict, init=False)
+    _parked: OrderedDict[Task, None] = attr.ib(factory=OrderedDict, init=False)
```
Fuyukai, line 109: why not just change this to dict now?

trio/_core/_parking_lot.py
```diff
@@ -203,7 +213,7 @@ def repark_all(self, new_lot):
         """
         return self.repark(new_lot, count=len(self))
 
-    def statistics(self):
+    def statistics(self) -> _ParkingLotStatistics:
```
Fuyukai, line 216: if you're exposing it via types, it should be renamed without the underscore

trio/_core/_run.py
Fuyukai: nurserymanager is an internal class. it should just be asynccontextmanager.

trio/_highlevel_socket.py
```diff
@@ -125,15 +130,15 @@ async def send_eof(self):
             with _translate_socket_errors_to_stream_errors():
                 self.socket.shutdown(tsocket.SHUT_WR)
 
-    async def receive_some(self, max_bytes=None):
+    async def receive_some(self, max_bytes: int | None = None):
```
Fuyukai, line 133: return type?

trio/_socket.py
```diff
@@ -472,44 +478,45 @@ def __getattr__(self, name):
...
+    def __dir__(self) -> Iterable[str]:
+        yield from super().__dir__()
+        yield from self._forward
 
-    def __enter__(self):
+    def __enter__(self) -> Self:
         return self
 
-    def __exit__(self, *exc_info):
+    def __exit__(self, *exc_info: object) -> None:
```
Fuyukai, line 488: why not fix this signature instead of just hacking onto the hack?

trio/_sync.py
```diff
@@ -319,7 +327,7 @@ def release_on_behalf_of(self, borrower):
         self._borrowers.remove(borrower)
         self._wake_waiters()
 
-    def statistics(self):
+    def statistics(self) -> _CapacityLimiterStatistics:
```
Fuyukai, line 330: same here, if it's hinted expose the class properly instead of this underscore nonsense.

trio/_sync.py
```diff
@@ -548,7 +556,7 @@ def release(self):
         else:
             self._owner = None
 
-    def statistics(self):
+    def statistics(self) -> _LockStatistics:
```
Fuyukai, line 559: same here

Fuyukai (changes requested):
too many individual ones to comment on, but ``__exit__`` and ``__aexit__`` returns [bool, not None.](https://docs.python.org/3/reference/datamodel.html?highlight=__exit__#object.__exit__) (or, well, it returns either True or anything else).

trio/_abc.py
```diff
@@ -252,10 +254,10 @@ async def aclose(self):
 
         """
 
-    async def __aenter__(self):
+    async def __aenter__(self) -> AsyncResource:
         return self
 
-    async def __aexit__(self, *args):
+    async def __aexit__(self, *args: object) -> None:
```
A5rocks, line 260: I think fixing this signature would be nice!

trio/_abc.py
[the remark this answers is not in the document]
A5rocks: New typing feature, I'm going to guess. It should be in `typing_extensions`, however. And worst comes to worst, `-> Self` is essentially shorthand for `def __aiter__(self: T) -> T`.

trio/_highlevel_socket.py
```diff
@@ -1,13 +1,18 @@
...
 from contextlib import contextmanager
+from typing import TYPE_CHECKING
 
 import trio
 from . import socket as tsocket
 from ._util import ConflictDetector, Final
 from .abc import HalfCloseableStream, Listener
 
+if TYPE_CHECKING:
+    from ._socket import _SocketType as SocketType
```
A5rocks, line 14: I think we should come up with a policy around runtime introspection. This can stay for now while we figure it out, but I don't know if the current direction this is all heading (you cannot inspect trio's types at runtime for some functions) is great. I don't know if there's any solid reasons to introspect at runtime either, so that's a thing.

trio/_highlevel_socket.py
```diff
@@ -330,7 +335,7 @@ class SocketListener(Listener[SocketStream], metaclass=Final):
 
     """
 
-    def __init__(self, socket):
+    def __init__(self, socket: SocketType):
```
A5rocks, line 338: Huh, TIL that mypy (strict, too) is fine without the explicit `-> None` bit here.

trio/_socket.py
A5rocks: Mind trying `tuple[object, ...]` and seeing if that works?

A5rocks (changes requested):
> too many individual ones to comment on, but __exit__ and __aexit__ returns [bool, not None.](https://docs.python.org/3/reference/datamodel.html?highlight=__exit__#object.__exit__) (or, well, it returns either True or anything else).

FWIW, in typeshed these are typed as `Optional[bool]`. This is really context-dependent: if one of these is typed as returning `None` or `Literal[False]`, then mypy at least can infer some special stuff about the context managers as it knows they won't swallow errors. (refs: they reverted https://github.com/python/mypy/issues/7577 but reachability is changed by this: https://github.com/python/mypy/issues/7666)

This was just a quick review though!

trio/_abc.py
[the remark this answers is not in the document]
jakkdl: because I didn't want to delve into pushing another commit midway through my review, sorry ^^;

trio/_socket.py
jakkdl: `str` certainly isn't a `tuple[object, ...]`, but it would match against `Iterable[object]`.

trio/_socket.py
jakkdl: Ah right, it should have the normal trio of exception values. Will do

trio/_highlevel_socket.py
jakkdl: Yeah `__init__` has a special case exception.

trio/_abc.py
```diff
@@ -278,7 +280,7 @@ class SendStream(AsyncResource):
     __slots__ = ()
 
     @abstractmethod
-    async def send_all(self, data):
+    async def send_all(self, data: bytes | bytearray | memoryview) -> None:
```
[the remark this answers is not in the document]
jakkdl: You're very welcome to open it for discussion! I'm not aware of the nuances here at all, just copy-pasting types.

trio/_core/_parking_lot.py
jakkdl: Why was it `'OrderedDict' in the first place? And why not just use `set`?

trio/_highlevel_socket.py
jakkdl: oops, missed that. Is it just `Any`?

trio/_highlevel_socket.py
jakkdl: yeah, and add isort+flake8-annotations to CI cause imports are a mess.

trio/_abc.py
jakkdl: I'll do a big search&replace on all `__exit__` and `__aexit__` in all files, so I won't have to juggle the trio of types and names more. I'll also standardize them to be the same.

trio/_socket.py
A5rocks: Sorry I meant instead of tuple[Any, ...] :^)

jakkdl: oh derp, yeah that works.

jakkdl:
> > too many individual ones to comment on, but **exit** and **aexit** returns [bool, not None.](https://docs.python.org/3/reference/datamodel.html?highlight=__exit__#object.__exit__) (or, well, it returns either True or anything else).
> 
> FWIW, in typeshed these are sometimes typed as `None` also, just as precedent. These return types matter though: if one of these is typed as returning `None` or `Literal[False]`, then mypy at least can infer some special stuff about the context managers as it knows they won't swallow errors. (refs: they reverted [python/mypy#7577](https://github.com/python/mypy/issues/7577) but reachability is changed by this: [python/mypy#7666](https://github.com/python/mypy/issues/7666))

typing them to be returning `bool` gives type issues unless you add explicit `return` statements - and those can't just `return False`. Adding extraneous return statements does not seem great, so I left a bunch of them as `-> None` - but I updated several to do `-> bool|None` or `-> bool`.

trio/_abc.py
jakkdl: > This signature is fine, given users shouldn’t be calling directly. It accepts the parameters, and means we don’t need to import `types`.

I'm hiding the `types` import behind `if TYPE_CHECKING`, so if anything it saves importing `typing` - which pretty much all of the files will be importing anyway.

jakkdl:
ugh, ofc the docs build now requires docstrings for all `.*Statistics`

trio/_core/_run.py
```diff
@@ -903,13 +903,18 @@ def __enter__(self) -> None:
             "use 'async with open_nursery(...)', not 'with open_nursery(...)'"
         )
 
-    def __exit__(self) -> None:  # pragma: no cover
+    def __exit__(
+        self,  # pragma: no cover
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> bool:
```
TeamSpen210, line 911: This along with `__enter__` should probably be hidden behind `if not TYPE_CHECKING`, so type checkers will correctly just give an error if used in a regular sync `with` statement. Potentially also the return types could be changed to `typing.NoReturn`, for runtime checkers perhaps.`__exit__()` will need to be swapped to `raise AssertionError(...)`, which is probably better anyway (since `-O` would cause it to just return `None`).

trio/_highlevel_socket.py
```diff
@@ -130,7 +130,7 @@ async def send_eof(self) -> None:
             with _translate_socket_errors_to_stream_errors():
                 self.socket.shutdown(tsocket.SHUT_WR)
 
-    async def receive_some(self, max_bytes: int | None = None):
+    async def receive_some(self, max_bytes: int | None = None) -> Any:
```
TeamSpen210, line 133: This should just be `-> bytes` I think, since that's what `recv()` ultimately [should return](https://docs.python.org/3/library/socket.html#socket.socket.recv)?

jakkdl: ah yes, thanks!

trio/_highlevel_socket.py
jakkdl: typed as `-> bytes`, see https://github.com/python-trio/trio/pull/2682#discussion_r1252424862

trio/_core/_run.py
```diff
@@ -854,7 +858,7 @@ class NurseryManager:
 
     """
 
-    strict_exception_groups = attr.ib(default=False)
+    strict_exception_groups: bool = attr.ib(default=False)
 
     @enable_ki_protection
     async def __aenter__(self):
```
TeamSpen210, line 864:
```suggestion
    async def __aenter__(self) -> Nursery:
```

trio/_core/_run.py
TeamSpen210: Be better to make this `object`, forcing users to do an `isinstance` check each time. Since anything touching the task could have set this, it's not really going to be safe to assume that this had the value you saw it with last.

trio/_core/_run.py
TeamSpen210: Maybe just type `coro: Any` to effectively disable type-checking, since this is doing a bunch of dynamic introspection, and would accept objects that duck-type as coroutines/generators (Cython)...

trio/_dtls.py
```diff
@@ -833,8 +838,13 @@ def close(self):
     def __enter__(self):
```
TeamSpen210, line 838:
```suggestion
    def __enter__(self) -> DTLSChannel:
```

trio/_dtls.py
```diff
@@ -1182,8 +1196,13 @@ def close(self):
...
```
TeamSpen210, line 1196:
```suggestion
    def __enter__(self) -> DTLSEndpoint:
```

TeamSpen210 (approved):
Few minor things, otherwise looks good.

trio/_highlevel_socket.py
A5rocks: opened https://github.com/python-trio/trio/issues/2687

trio/_core/_run.py
jakkdl: Sure, could be made stricter later on if somebody wants.

trio/_core/_run.py
jakkdl: Sounds reasonable, added a comment copying most of your explanation.

trio/_core/_parking_lot.py
jakkdl: That feels like it can be trivially be replaced with `set.pop()`? Or is there any difference in execution speed or something.

trio/_dtls.py
jakkdl: leaving for future PR.

trio/_core/_run.py
jakkdl: Hm, getting a bunch of errors in `trio._channel`:
```
trio/_channel.py:181: error: "object" has no attribute "_tasks"  [attr-defined]
trio/_channel.py:281: error: "object" has no attribute "_tasks"  [attr-defined]
trio/_channel.py:318: error: "object" has no attribute "_tasks"  [attr-defined]
trio/_channel.py:427: error: "object" has no attribute "_tasks"  [attr-defined]
```

so internal trio functions are certainly making assumptions about the type.

jakkdl: `type: ignore` them, or add `isinstance` checks? (Or put it back to `Any` and leave properly typing it with a `TypeVar` or something for the future)

jakkdl:
ugh, messing with symbols and trying to understand readthedocs build errors is a pain. If anybody is in the mood for looking at that it'd be quite welcome, otherwise I'll procrastinate on it for another day.

trio/_core/_run.py
A5rocks: Fwiw in my opinion this and the rescheduling thing are probably the only places I actually think exposing `Any` is a good thing :P 

There's not much you can do if you suddenly find the data has changed, after all. But idk up to you ig

TeamSpen210:
I think the issue there is that they need to have a definition in the documentation, so there's somewhere for them to link to.

trio/_core/_parking_lot.py
jakkdl: Ah, so exactly as the comment says (duh) they want a FIFO/deque that also supports removing random items. Though given that items seem to only ever be removed randomly (in `_pop_several`) I'm a bit confused why it needs to be FIFO order.

trio/_core/_run.py
jakkdl: I'll set it to `Any` for now, and let future PR make it stricter if anybody cares.

trio/_core/_parking_lot.py
A5rocks: Err, I think you're confused a bit: the pop several method is ordered. The only reason we want fast random removal is because of cancellation!

jakkdl: oooh. Thanks!

trio/_tests/test_exports.py
```diff
@@ -492,4 +492,8 @@ def test_classes_are_final():
                 continue
             # ... insert other special cases here ...
 
+            # don't care about the *Statistics classes
```
TeamSpen210, line 495: Instead of opting these out, wouldn't it be better to make them final? People shouldn't be inheriting from them.

docs/source/reference-core.rst
```diff
@@ -1096,6 +1096,8 @@ Broadcasting an event with :class:`Event`
 .. autoclass:: Event
    :members:
 
+.. autoclass:: EventStatistics
```
TeamSpen210, line 1099: Instead of documenting the class itself here, it might be possible to just put an [`index`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup) directive in the `statistics()` method, so the name links back there. That way we don't have the same information repeated twice.

trio/_core/_run.py
```diff
@@ -1167,7 +1167,10 @@ def __del__(self) -> None:
 @attr.s(eq=False, hash=False, repr=False, slots=True)
 class Task(metaclass=NoPublicConstructor):
     _parent_nursery: Nursery | None = attr.ib()
-    coro: Coroutine[Any, outcome.Outcome[object], Any] = attr.ib()
+    # could be typed as `Coroutine[Any, outcome.Outcome[object], Any]` but
+    # the code does a lot of dynamic introspection on it and as long as it passes
+    # those it's fine.
+    coro: Any = attr.ib()
```
TeamSpen210, line 1173: I was thinking that this could stay typed as `Coroutine` for the public API of the class, but cast the `coro` local variable in `iter_await_frames()`. That way type checking is still done.

TeamSpen210 (approved):
A few minor things, but otherwise seems good.

docs/source/conf.py
```diff
@@ -62,10 +62,22 @@
...
 autodoc_inherit_docstrings = False
 default_role = "obj"
 
+# These have incorrect __module__ set in stdlib and give the error
+# `py:class reference target not found`
+# Some of the nitpick_ignore's above can probably be fixed with this.
+# See https://github.com/sphinx-doc/sphinx/issues/8315#issuecomment-751335798
+autodoc_type_aliases = {
+    # aliasing doesn't actually fix the warning for types.FrameType, but displaying
+    # "types.FrameType" is more helpfun than just "frame"
```
Zac-HD, line 76:
```suggestion
    # "types.FrameType" is more helpful than just "frame"
```

idea for a future PR, we might find `codespell` useful?

trio/_core/_run.py
```diff
@@ -1266,9 +1295,9 @@ def print_stack_for_task(task):
 
     # The CancelStatus object that is currently active for this task.
     # Don't change this directly; instead, use _activate_cancel_status().
-    _cancel_status = attr.ib(default=None, repr=False)
+    _cancel_status: CancelStatus = attr.ib(default=None, repr=False)
 
-    def _activate_cancel_status(self, cancel_status):
+    def _activate_cancel_status(self, cancel_status: CancelStatus):
```
Zac-HD, line 1300:
```suggestion
    def _activate_cancel_status(self, cancel_status: CancelStatus) -> None:
```
?

trio/_core/_run.py
Zac-HD: Searching for `_attempt_abort(`, we only ever call it in the methods immediately below, both of which check that `self._abort_func is not None` first.  I think we'd therefore be safe with a comment along those lines, an `assert self._abort_func is not None, "FATAL INTERNAL ERROR"`, and no type ignore.

trio/_highlevel_generic.py
Zac-HD: No, I think in this case we accept sending `bytearray` from the caller, but only provide `bytes` to the caller (_ala_ Postel's Principle).  That's also consistent with the `_highlevel_socket` implementation.

trio/_socket.py
Zac-HD: https://docs.python.org/3/reference/datamodel.html#object.__dir__ says "a sequence must be returned", and the builtin `dir()` function converts it to a list.  Let's stick with the status quo here.

trio/_socket.py
```diff
@@ -518,7 +525,7 @@ async def bind(self, address):
         ):
             # Use a thread for the filesystem traversal (unless it's an
             # abstract domain socket)
-            return await trio.to_thread.run_sync(self._sock.bind, address)
+            return await trio.to_thread.run_sync(self._sock.bind, address)  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
Zac-HD: Let's leave a comment to that effect, so we know when it's safe to remove the ignore?

Zac-HD (approved):
Small notes below, but this looks fantastic overall and I'm really looking forward to using Trio with annotations.

trio/_socket.py
TeamSpen210: Probably better to do `[*super().__dir__(), *self._forward]` then so we ourselves don't depend on it being a sequence.

docs/source/conf.py
A5rocks: I'm a +1 on codespell, I've heard good things about it!

trio/_tests/test_exports.py
jakkdl: I don't really see any strong reason to *forbid* people from inheriting them either? I could plausibly see somebody writing a wrapper around one of the locks and them deeming it useful to inherit from the Statistics class.

docs/source/conf.py
jakkdl: no fun allowed.

Yeah `codespell` has caught a bunch of typos of mine in flake8-trio, I'm all for adding it.

trio/_socket.py
jakkdl: The status quo yields a typing error, since `super().__dir__()` is `Iterable[str]`. But sticking with returning a list to avoid unnecessary changes to the API is probably wise.

trio/_socket.py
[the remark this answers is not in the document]
jakkdl: `mypy` should raise an `unnecessary type: ignore`, esp as I specified the precise error to be ignored. But no harm in adding a comment.

docs/source/reference-core.rst
jakkdl: After trying a bunch of different variants I can't get `index` to create a `class`-type reference. I can create an index named `trio.EventStatistics` that I can link to with [`:ref:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations) - but I failed to create one I can link to with `:class:` which is what autodoc generates the docstring to do.

jakkdl:
I'll go ahead and merge this unless @A5rocks or @Fuyukai wants to re-review

trio/_core/_run.py
```diff
@@ -2433,17 +2465,17 @@ def unrolled_run(
 
 
 class _TaskStatusIgnored:
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "TASK_STATUS_IGNORED"
 
-    def started(self, value=None):
+    def started(self, value: Any = None):
```
A5rocks, line 2471: Pedantic note: `object` here? (I realize it doesn't change anything at all...)

Also return type

trio/_dtls.py
```diff
@@ -1167,12 +1178,16 @@ def __del__(self):
...
-    def close(self):
+    def close(self) -> None:
         """Close this socket, and all associated DTLS connections.
 
         This object can also be used as a context manager.
 
         """
+        # Do nothing if this object was never fully constructed
+        if self.socket is None:  # pragma: no cover
+            return
```
A5rocks, line 1189: ... I don't think this case is possible? Well not unless someone's doing something _really_ funky like overriding our `__del__` to call `.close()`. Which is like, no, I don't think people will be doing that.

I think an `assert` is good enough, though it would be nice to have `socket` typed as non-optional. Not exactly sure how to do that though.

trio/_highlevel_generic.py
```diff
@@ -72,34 +79,35 @@ class StapledStream(HalfCloseableStream, metaclass=Final):
...
         """Calls ``self.send_stream.send_all``."""
         return await self.send_stream.send_all(data)
 
-    async def wait_send_all_might_not_block(self):
+    async def wait_send_all_might_not_block(self) -> None:
         """Calls ``self.send_stream.wait_send_all_might_not_block``."""
         return await self.send_stream.wait_send_all_might_not_block()
 
-    async def send_eof(self):
+    async def send_eof(self) -> None:
```
A5rocks, line 93: Why did you type this as returning `None`? That led to a `# type: ignore`. I think that's totally avoidable!

trio/lowlevel.py
```diff
@@ -7,10 +7,12 @@
 import sys
 import typing as _t
 
+# Generally available symbols
 # Generally available symbols
```
A5rocks, line 11: Duplicated comment?

trio/_sync.py
```diff
@@ -8,10 +11,16 @@
...
+
+    from ._core import Task
+    from ._core._parking_lot import ParkingLotStatistics
 
-@attr.s(frozen=True)
-class _EventStatistics:
-    tasks_waiting = attr.ib()
+
+@attr.s(frozen=True, slots=True)
+class EventStatistics:
```
A5rocks, line 22: This might want a docstring too?

trio/_sync.py
```diff
@@ -84,25 +93,30 @@ def statistics(self):
...
-        return _EventStatistics(tasks_waiting=len(self._tasks))
+        return EventStatistics(tasks_waiting=len(self._tasks))
 
 
 class AsyncContextManagerMixin:
     @enable_ki_protection
-    async def __aenter__(self):
-        await self.acquire()
+    async def __aenter__(self) -> None:
+        await self.acquire()  # type: ignore[attr-defined]
```
A5rocks, line 102: We can drop this `attr-defined` through fancy self-type schenanigans:

```py
import typing

class ACM(typing.Protocol):
    async def acquire(self) -> None:
        pass

class AsyncContextManagerMixinWithError:
    async def __aenter__(self) -> None:
        await self.acquire()

class AsyncContextManagerMixinWithoutError:
    async def __aenter__(self: ACM) -> None:
        await self.acquire()
```

Feel free to check (I find <https://mypy-play.net> nice for testing small things like this) :D

trio/_sync.py
```diff
@@ -84,25 +93,30 @@ def statistics(self):
...
     @enable_ki_protection
-    async def __aexit__(self, *args):
-        self.release()
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        self.release()  # type: ignore[attr-defined]
```
A5rocks, line 111: Comment from above applies here too.

A5rocks (changes requested):
Looks mostly fine. I've been assuming that the types are valid and from a couple checks early on it looks like they are.

I held back extraneous commentary!!

trio/_dtls.py
jakkdl: Adding a comment so it can be addressed in an upcoming typing PR. I don't want to gradually add more typing to this PR, requiring more reviews.

trio/_highlevel_generic.py
jakkdl: the alternative is `-> Any`, since `self.send_stream.send_eof()` is per definition unknown as it's not defined in `SendStream`. Restricted the `type: ignore` further and added a comment. If `send_eof` is supposed to return something that should be documented somewhere.

trio/lowlevel.py
jakkdl: `isort` probably merged two sets of imports from `_core`

trio/_sync.py
jakkdl: Added for CapacityLimiter too

trio/_sync.py
jakkdl: Ooooh, fancy! Leaving it for a future PR, adding a comment with a `#TODO`, as I'm unused to Protocols.

jakkdl:
I have started on a follow-up typing PR, so I'll merge this one tomorrow unless you have any final objections @A5rocks (or you can go ahead and merge it)

A5rocks (approved):
Looks good now

PR #2685, opened 2023-07-05
don't open git diff in pager, mypy export tests now error on any mypy output

jakkdl:
various small fixes for issues that arose when running tests locally / in tox

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -32,14 +32,14 @@ pip-compile docs-requirements.in
 
 if git status --porcelain | grep -q "requirements.txt"; then
     git status --porcelain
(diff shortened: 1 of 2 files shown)


A5rocks (approved):
This looks good, can you make `git diff` always output in color? I don't think it does that already for GH actions but the logs here support it! It would be quite nice IMO.

jakkdl:
> This looks good, can you make `git diff` always output in color? I don't think it does that already for GH actions but the logs here support it! It would be quite nice IMO.

looks great!
https://github.com/python-trio/trio/actions/runs/5474969037/jobs/9970426746?pr=2685#step:5:1069

PR #2686, opened 2023-07-05
isort take 2

jakkdl:
#2538 got abandoned, but that felt in large part because it was just a one-off random big reorg. This adds `isort` to CI, and it only needed a single `isort: skip` for it to work for the whole project.


A5rocks:
Huh, CI failing seems weird. It doesn't seem to be any sort of environment change but it's hard to imagine import re-ordering affecting static tools like that- actually, it isn't that hard to imagine.

I'll guess there's maybe duplicate somethings somewhere...? Or I don't know, maybe just revert certain files and `# isort: skip` them. Not sure.

I'm also not too sure about how I feel about such a large diff that will probably lead to a ton of merge conflicts! It's probably fine though since I think most open PRs are dead (as far as I know).

trio/_tests/test_exports.py
```diff
@@ -422,19 +444,19 @@ def lookup_symbol(symbol):
...
+        #    trio.MemoryReceiveChannel,
+        #    trio.MemorySendChannel,
+        #    trio.SSLListener,
+        #    trio.SocketListener,
+        # ):
+        #    missing.remove("__aenter__")
+        #    missing.remove("__aexit__")
+        # if class_ in (trio.DTLSChannel, trio.MemoryReceiveChannel):
+        #    missing.remove("__aiter__")
+        #    missing.remove("__anext__")
```
jakkdl, line 459: the 3.11&3.12 failures are because I accidentally pushed this - I think my local environment updated `jedi` or something.

Will look at the windows fails

trio/socket.py
```diff
@@ -56,25 +56,25 @@
...
         from socket import (
-            sethostname as sethostname,
+            if_indextoname as if_indextoname,
             if_nameindex as if_nameindex,
             if_nametoindex as if_nametoindex,
-            if_indextoname as if_indextoname,
+            sethostname as sethostname,
         )
     except ImportError:
         pass
```
jakkdl, line 80: The windows CI fails is because `pyright` now doesn't see `if_indextoname`, `if_nameindex` or `if_nametoindex` on windows ... but it sees `sethostname`? And why did it "work" in the first place, since `sys.platform == "win32"` and `_t.TYPE_CHECKING == TRUE` it shouldn't have ever seen these symbols if it's parsing the `if` statement.

Regardless, I don't think it's an issue, and I don't care to dig deeper into it atm, so I'll just `Remove` those in the test.

jakkdl:
> I'm also not too sure about how I feel about such a large diff that will probably lead to a ton of merge conflicts! It's probably fine though since I think most open PRs are dead (as far as I know).

They're either dead, or by me :innocent: 
In the long term this should cut down on merge conflicts from imports though, and hopefully any dead PR's that get resurrected doesn't play around *too* much with imports, and resolving import conflicts is usually much easier than in normal code (esp when you can just add both sets fully, then run isort+flake8 w/ f401)

jakkdl:
The only failing check is an erroneously duplicated `codecov/project`

trio/_core/_generated_instrumentation.py
A5rocks, line 1: Do we need to run `isort` after it's regenerated?

A5rocks (commented):
Passes a spot check. I didn't look too hard but I assume this is all good cause tests pass!

I'm going to see if people on Gitter have Opinions about this but I think this is a good enough change to merge despite merge conflicts it will cause.

pyproject.toml
```diff
@@ -14,6 +14,12 @@ per-file-ignores = [
   'trio/testing/__init__.py: F401'
 ]
 
+[tool.isort]
+combine_as_imports = true
+profile = "black"
+py_version = "311"
```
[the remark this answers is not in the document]
TeamSpen210: Looks like it'd be better to just use the default. This [controls what modules are considered standard library](https://pycqa.github.io/isort/docs/configuration/options.html#python-version), the default is to compare against all Python versions.

Kludex:
I didn't read the PR or issue, but [this is useful](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) for this kind of refactor.

trio/_core/_generated_instrumentation.py
jakkdl: I manually sorted the imports in `gen_exports.py` so they pass - though upon reflection I should probably just `# isort: skip` as that's how the black formatting is handled, and shoehorning in `isort` to work with `gen_exports.py --check` or forcing future people to manually sort the imports does not sound fun.

jakkdl:
> I didn't read the PR or issue, but [this is useful](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) for this kind of refactor.

oooh, this is great. Thanks a lot!

jakkdl:
> I don't think you can add the `.git-blame-ignore-revs` on this PR, it needs to be on the next one, cause the commits are going to be squashed (I guess).

I rebased and split the PR among three commits, one to add `isort` to CI and configuring it (that's not ignored by blame, so people can see when it was introduced), one that runs `isort` on all the files, and a third one that adds `.git-blame-ignore-revs`. I can then merge/rebase it without squashing and keep all three commits

Kludex:
> [...] without squashing and keep all three commits

Great! 👍

jakkdl:
Since nobody's complaining, I'll go ahead and merge :rocket:

Kludex:
I think the commit is actually 933f77b96f0092e1baab4474a9208fc2e379aa32 ?

PR #2691, opened 2023-07-07
Make type completeness results consistent for different platforms

TeamSpen210:
If the type completeness test is run on Windows, the slightly different imports produce different counts, and also cause the JSON file to end up with symbols in different orders. This sorts the JSON file to prevent that from affecting diff results, and configures `pyright` to always behave like the platform is Linux. Perhaps it would be better to test both platforms and merge the results (so OS-specific code gets checked for missing types), but that would require a bigger change to the script.


TeamSpen210:
Maybe this might not work, there's still 4 symbols that don't match...

jakkdl:
Ooh, this is great.

If you want to hunt down what's the cause of the difference you can [temporarily] print the ones with known type to the json and not just the ones without, but I think we can just remove `withKnownType` from the results and ignore it in the check if you don't manage to figure it out. We don't *actually* care about the raw number of known types after all, and the actual goal is to bring `withUnknownType` to 0.

TeamSpen210:
The difference are these symbols:
* `trio._path.Path.hardlink_to()`
* `trio._path.Path.is_relative_to()`
* `trio._path.Path.readlink()`
* `trio._path.Path.with_stem()`

... Which are caused by me using Python 3.11 not 3.8.

A5rocks (approved):
This looks good to me.

jakkdl (approved):
Looks awesome, thanks for the fix.

Issue #2688, opened 2023-07-06
PermissionError: [WinError 5] Access is Denied when importing trio


chrisbeardy:
Trio version 0.22.1
Python version 3.11.4 and 3.10.12 tried
Windows 10

After creating a venv, ttrying to import trio in a python console gives a permission error.

Tthis only happens with 0.22.1, reverting to 0.22.0 works just fine

ttraceback:
```
Traceback (most recent call last):
  File "C:\Users\chrisb\AppData\Local\Programs\Python\Python310\lib\code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2022.1.3\plugins\python-ce\helpers\pydev\_pydev_bundle\pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
  File "C:\Users\chrisb\GitWS\async_and_thread_test\venv2\lib\site-packages\trio\__init__.py", line 22, in <module>
    from ._core import (
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2022.1.3\plugins\python-ce\helpers\pydev\_pydev_bundle\pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
  File "C:\Users\chrisb\GitWS\async_and_thread_test\venv2\lib\site-packages\trio\_core\__init__.py", line 27, in <module>
    from ._run import (
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2022.1.3\plugins\python-ce\helpers\pydev\_pydev_bundle\pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
  File "C:\Users\chrisb\GitWS\async_and_thread_test\venv2\lib\site-packages\trio\_core\_run.py", line 34, in <module>
    from ._thread_cache import start_thread_soon
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2022.1.3\plugins\python-ce\helpers\pydev\_pydev_bundle\pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
  File "C:\Users\chrisb\GitWS\async_and_thread_test\venv2\lib\site-packages\trio\_core\_thread_cache.py", line 66, in <module>
    set_os_thread_name = get_os_thread_name_func()
[...]

Issue #2689, opened 2023-07-06
Is it possible to make codecov action not "fail" until all other runs are completed?


jakkdl:
As I'm currently juggling several PRs and since the added xfail's on 3.12 it's not possible to see the :heavy_check_mark: vs :x: on a PR to see at a glance if it passes test. But, as soon as the first result is pushed to codecov it will fail and display :x:, until however longer time later when enough results have come in that it passes. I'd personally much prefer if it was possible to mark it as not giving a result until all results are in, so I can rely on the status icon of :heavy_check_mark: vs :x: vs :yellow_circle:.

One solution would be:
> One thought I have on this, though, is that what Hynek advocates for in his article https://hynek.me/articles/ditch-codecov-python/ could be useful here too. Maybe, if we merged all the reports on the GHA side and made just a single upload to Codecov, it could help with the rate limit just enough for this CI to have the best of two worlds: stability + ability to view coverage on Codecov's web UI.

as mentioned in https://github.com/python-trio/trio/pull/2672#pullrequestreview-1498778904

jakkdl:
oh, and somewhat related - I'm getting duplicate codecov checks, e.g. in https://github.com/python-trio/trio/pull/2686

jakkdl:
Gonna bump this after codecov flakiness in https://github.com/python-trio/trio/pull/2812#issuecomment-1763328467

richardsheridan:
I implemented something inspired by Hynek's post in my repo fwiw: https://github.com/richardsheridan/trio-parallel/pull/114

PR #2693, opened 2023-07-07
fix PermissionError due to accessing pthreads upon importing trio

jakkdl:
quickfix for #2688


jakkdl:
CI fail is due to https://github.com/python-trio/trio/issues/2689#issuecomment-1623572585

A5rocks (changes requested):
I don't really know if this works cause not my kinda thing. Would be happy to merge and get a quick fix out though, just please add a news fragment!

A5rocks:
Actually we can probably ask the original issue reporter to test this. I forgot the syntax to make pip install from a specific branch :(

jakkdl:
ah oops, I wrote a newsfragment but looks like I forgot to add the file

jakkdl:
@chrisbeardy would you mind installing trio from this branch and seeing if it works? https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch

chrisbeardy:
Yep will do, give me a day or so as away for the weekend

chrisbeardy:
Can confirm this imports ok now "on my machine"

jakkdl:
> Can confirm this imports ok now "on my machine"

Awesome! It's the only data point we have about it not working, so that's good enough for me. Feel free to merge and maybe push out a quick bugfix release if the newsfragment looks good (otherwise just edit it directly) @A5rocks

A5rocks (approved):

PR #2697, opened 2023-07-12
move mypy config to pyproject.toml

jakkdl:
Continuing to unify config location (as with flake8). Also sets mypy to always use `py38`


jakkdl:
I don't see this being controversial or problematic in any way, and I'd like it for PRs I'm working on, so I'll just quickly merge this.

PR #2700, opened 2023-07-12
Modernize sniffio integration

oremanj:
Use `sniffio.thread_local` to indicate which async library is running, instead of the deprecated `sniffio.current_async_library_cvar`. This allows us to remove a number of workarounds for the contextvar being inherited where it shouldn't be.


CoolCat467:
Why was this merged with failing mypy run?

A5rocks:
Because mypy was already failing on master, I think.

PR #2703, opened 2023-07-13
typecheck trio/_abc.py

jakkdl:


TeamSpen210 (approved):
LGTM, though I'm not very experienced with sockets so someone else might want to check that bit.

A5rocks (approved):
This seems good! Discussion about what to do with `_SocketType` goes into https://github.com/python-trio/trio/pull/2705 so I think that's fine.

I'm also not too familiar with sockets so... not completely sure.

trio/_abc.py
```diff
@@ -186,7 +209,12 @@ class SocketFactory(metaclass=ABCMeta):
     """
 
     @abstractmethod
-    def socket(self, family=None, type=None, proto=None):
+    def socket(
+        self,
+        family: socket.AddressFamily | int = socket.AF_INET,
+        type: socket.SocketKind | int = socket.SOCK_STREAM,
+        proto: int = 0,
+    ) -> _SocketType:
```
A5rocks, line 217: Meh, I don't really feel comfortable changing away from the `=None` without knowing why exactly they're here in the first place. Additonally, in typeshed this is how a similar function is defined:

```py
if sys.platform == "win32":
    def socketpair(family: int = ..., type: int = ..., proto: int = 0) -> tuple[socket, socket]: ...

else:
    def socketpair(
        family: int | AddressFamily | None = None, type: SocketType | int = ..., proto: int = 0
    ) -> tuple[socket, socket]: ...
```

A5rocks (changes requested):
I went ahead and checked against typeshed just to make sure the socket types are correct!

My only concern is related to going from `=None` in the `def socket(...)` classmethod. Why was it done this way to begin with @njsmith?

trio/_abc.py
jakkdl: Actually figuring out the value of the defaults required a bit of digging, and it took me a bit to be certain enough that they were static and not dynamic or something, given that pretty much all default values in typeshed is `= ...`. I wouldn't have set the default values to anything but `...` if not for `typeright --verifytypes` complaining when default values aren't explicit.

https://github.com/python-trio/trio/blob/315dbed5abf0663b11522022e9c7abbc1df54dc8/trio/_socket.py#L270-L289
does have explicit logic for `None` though, so reverting the default value to `None` seems eminently plausible.

jakkdl:
> I'm also not too familiar with sockets so... not completely sure.

Given that we're (to the extent possible) only changing typing, and not any behaviour, I don't think we need to be 100% certain that all the types are correct and I expect some incorrect signatures to slip through, and/or existing weird behavior that surfaces because of previously untyped code now being typechecked, will require followup PRs/releases.

A5rocks (approved):
Seems good to me.

PR #2705, opened 2023-07-14
typecheck _socket and _core._local

jakkdl:
This one's a bit messier, and I failed to resolve what to do with `_SocketType`. The stubs in typeshed for `socket` aren't the cleanest either, but I managed to mess around enough to satisfy `pyright --verifytypes` stringent requirements of unambigious and specify the actual default value - and I think/hope everything's correct, though no big harm if they aren't.

Typing `RunVar` with `no_value = object()` seemed not very possible, but I think this change shouldn't affect behaviour.

I'll do another push to figure out if there's any decent way of resolving autodoc, otherwise I'll just spam `nitpick_ignore` in these three PRs. `_SocketType` is the main one here that might merit documenting.


Fuyukai:
Typing ``_socket`` makes it look legitimate and not something that needs to be put on a road and stomped in the head repeatedly.

trio/_socket.py
```diff
@@ -244,7 +294,7 @@ async def getprotobyname(name):
 ################################################################
 
 
-def from_stdlib_socket(sock):
+def from_stdlib_socket(sock: _stdlib_socket.socket) -> _SocketType:
```
TeamSpen210, line 297: Ideally, public functions shouldn't return private types, since users can't then annotate their own stuff with that. `SocketType` is kinda too empty to be useful though...

trio/_socket.py
```diff
@@ -424,18 +521,24 @@ async def _resolve_address_nocp(type, family, proto, *, ipv6_v6only, address, lo
     # The above ignored any flowid and scopeid in the passed-in address,
     # so restore them if present:
     if family == _stdlib_socket.AF_INET6:
-        normed = list(normed)
+        list_normed = list(normed)
         assert len(normed) == 4
+        # typechecking certainly doesn't like this logic, but given just how broad
+        # Address is, it's kind of impossible to write the below without type: ignore
```
TeamSpen210, line 527: This might be possible to type, if you define `TypeGuard` functions for the two len checks here, then change each branch to just build a whole tuple instead of converting to list (so we preserve the heterogeneity). Or unpack it into four variables.

trio/_socket.py
```diff
@@ -481,7 +584,7 @@ def __init__(self, sock: _stdlib_socket.socket):
         "share",
     }
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str) -> Any:
```
TeamSpen210, line 587: Should probably hide this behind `TYPE_CHECKING`,  then add stubs for each of those methods/properties.

trio/_socket.py
```diff
@@ -338,19 +418,21 @@ def _sniff_sockopts_for_fileno(family, type, proto, fileno):
 )
 
 
-def _make_simple_sock_method_wrapper(methname, wait_fn, maybe_avail=False):
-    fn = getattr(_stdlib_socket.socket, methname)
-
+def _make_simple_sock_method_wrapper(
+    fn: Callable[Concatenate[_stdlib_socket.socket, P], T],
+    wait_fn: Callable,
```
TeamSpen210, line 423: Any reason `wait_fn` isn't fully typed?

TeamSpen210 (changes requested):
Found a bunch of minor things. `_SocketType` seems tricky though, we wouldn't want to require users to use that private name for their annotations, but `SocketType` is empty so you can't do anything with it...

oremanj:
I think the issue with `SocketType` vs `_SocketType` is an excellent reason to use a stub file `_socket.pyi` here. These interfaces don't change often (almost at all, really) so I'm not as concerned about shear between the interface and the implementation.

The other option if you don't like that would be to make `SocketType` be an ABC with abstractmethods for all the public-facing socket methods. `_SocketType` has the implementations, but we only use the interface name `SocketType` in public-facing types.

trio/_socket.py
```diff
@@ -30,10 +29,16 @@
...
+    from typing_extensions import Buffer, Self, TypeAlias, Concatenate, ParamSpec, SupportsIndex
 
     from ._abc import HostnameResolver, SocketFactory
 
+    # Duplicated from _socket type stubs
+    if sys.version_info >= (3, 8):
+        FileDescriptor: TypeAlias = SupportsIndex
+    else:
+        FileDescriptor: TypeAlias = SupportsInt
+
```
jakkdl, line 41: Ugh, let's just get #2668 merged

jakkdl:
> I think the issue with `SocketType` vs `_SocketType` is an excellent reason to use a stub file `_socket.pyi` here. These interfaces don't change often (almost at all, really) so I'm not as concerned about shear between the interface and the implementation.
> 
> The other option if you don't like that would be to make `SocketType` be an ABC with abstractmethods for all the public-facing socket methods. `_SocketType` has the implementations, but we only use the interface name `SocketType` in public-facing types.

I think I'll just leave _SocketType as is for this PR, and open an issue discussing possible solutions and then a PR dedicated for that. Making `SocketType` abstract with a bunch of abstract methods will probably break downstream code that inherits from it and just happens not to implement one of the methods made abstract.

trio/_socket.py
jakkdl: nope, especially when it is in `_nonblocking_helper` - I just forgot. I should enable https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_generics

trio/_socket.py
jakkdl: Ooh, I didn't try with `TypeGuard`. Though I don't think it's terribly important to fix these lines, unless those typeguard functions can be reused elsewhere.

trio/_threads.py
```diff
@@ -59,8 +64,12 @@ class ThreadPlaceholder:
 
 @enable_ki_protection
 async def to_thread_run_sync(
-    sync_fn, *args, thread_name: Optional[str] = None, cancellable=False, limiter=None
-):
+    sync_fn: Callable[..., T],
+    *args: Any,
```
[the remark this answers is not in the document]
jakkdl: We likely want https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_expr enabled down-the-line anyway though, and for the few cases where we give up add explicit `type: ignore`s

trio/_socket.py
```diff
@@ -697,77 +824,144 @@ async def connect(self, address):
...
+    send = _make_simple_sock_method_wrapper(  # noqa: F811
+        _stdlib_socket.socket.send, _core.wait_writable
+    )
 
     ################################################################
     # sendto
     ################################################################
 
+    @overload
```
jakkdl, line 944: note to self: add `@overload` to list of markers that `Coverage` should ignore the content of

trio/_core/_local.py
```diff
@@ -28,14 +39,23 @@ class RunVar(metaclass=Final):
...
+    @overload
+    def get(self, default: T) -> T:
+        ...
+
+    @overload
+    def get(self, default: NoValue = _NO_DEFAULT) -> T | NoValue:
+        ...
 
-    def get(self, default=_NO_DEFAULT):
+    def get(self, default: T | NoValue = _NO_DEFAULT) -> T | NoValue:
```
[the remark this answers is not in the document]
jakkdl: If it can't return `NoValue` there's no point in doing the `@overload`, so removing those.

Not totally in love with using features whose PR's are yet to be merged, and not supported in pyright either. But pretty minor and seems likely to be supported soon, so might as well.

trio/_sync.py
```diff
@@ -268,7 +268,7 @@ def acquire_nowait(self) -> None:
         self.acquire_on_behalf_of_nowait(trio.lowlevel.current_task())
 
     @enable_ki_protection
-    def acquire_on_behalf_of_nowait(self, borrower: Task) -> None:
+    def acquire_on_behalf_of_nowait(self, borrower: object) -> None:
```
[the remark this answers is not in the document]
jakkdl: Thought about adding a generic, but that seems somewhat overkill - and would have to be `T|Task` since class methods do stuff like `self.acquire_on_behalf_of(trio.lowlevel.current_task()`. Could maybe have some other type/class that `ThreadPlaceHolder` and user functions inherit from / have a `Protocol`.

trio/_threads.py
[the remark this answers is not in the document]
jakkdl: oops, meant https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_generics
any-expr is further down the line

trio/_socket.py
jakkdl: ooor just get rid of `forward` altogether.

trio/_core/_local.py
```diff
@@ -28,44 +39,44 @@ class RunVar(metaclass=Final):
...
+    def set(self, value: T) -> RunVarToken[T]:
         """Sets the value of this :class:`RunVar` for this current run
         call.
 
         """
         try:
             old_value = self.get()
         except LookupError:
-            token = _RunVarToken.empty(self)
+            token = RunVarToken[T]._empty(self)
```
A5rocks, line 70: Huh, I feel like mypy should infer `T`.

trio/_socket.py
```diff
@@ -17,7 +30,19 @@
...
+    from typing_extensions import Buffer, Concatenate, ParamSpec, Self, TypeAlias
+
+    from ._abc import HostnameResolver, SocketFactory
+
+    P = ParamSpec("P")
+
+
+T = TypeVar("T")
+
+# must use old-style typing for TypeAlias
```
A5rocks, line 42: (more specifically, this is because it's at runtime, correct?)

trio/_socket.py
```diff
@@ -178,9 +225,10 @@ def numeric_only_failure(exc):
             # idna.encode will error out if the hostname has Capital Letters
             # in it; with uts46=True it will lowercase them instead.
             host = _idna.encode(host, uts46=True)
-    hr = _resolver.get(None)
+    hr: HostnameResolver | None = _resolver.get(None)
```
A5rocks, line 228: I think these type hints should also be inferred?

trio/_socket.py
```diff
@@ -241,9 +292,14 @@ def from_stdlib_socket(sock):
 
 
 @_wraps(_stdlib_socket.fromfd, assigned=(), updated=())
-def fromfd(fd, family, type, proto=0):
+def fromfd(
+    fd: SupportsIndex,
+    family: AddressFamily | int = _stdlib_socket.AF_INET,
+    type: SocketKind | int = _stdlib_socket.SOCK_STREAM,
```
A5rocks, line 298: Were the defaults missing before?

trio/_socket.py
```diff
@@ -241,9 +292,14 @@ def from_stdlib_socket(sock):
...
+    proto: int = 0,
+) -> _SocketType:
     """Like :func:`socket.fromfd`, but returns a Trio socket object."""
-    family, type, proto = _sniff_sockopts_for_fileno(family, type, proto, fd)
+    family, type, proto = _sniff_sockopts_for_fileno(family, type, proto, int(fd))
```
A5rocks, line 302: I don't think `int(fd)` is guaranteed to be the same behavior as `fd.__index__()` but it's probably ... fine?

trio/_socket.py
```diff
@@ -446,36 +550,72 @@ def __init__(self, sock: _stdlib_socket.socket):
...
+        return self._sock.fileno()
+
+    def getpeername(self) -> Any:
+        return self._sock.getpeername()
+
+    def getsockname(self) -> Any:
+        return self._sock.getsockname()
+
+    @overload
+    def getsockopt(self, __level: int, __optname: int) -> int:
```
A5rocks, line 567: We dropped support for 3.7

trio/_socket.py
```diff
@@ -685,77 +843,144 @@ async def connect(self, address):
...
+        _stdlib_socket.socket.send, _core.wait_writable
+    )
 
     ################################################################
     # sendto
     ################################################################
 
+    @overload
+    async def sendto(
+        self, __data: Buffer, __address: tuple[Any, ...] | str | Buffer
```
A5rocks, line 965: `tuple[Any, ...]`...?

A5rocks (changes requested):
I didn't look *too* much but this should be most of my comments.

trio/_socket.py
TeamSpen210: Yep. Could quote the whole thing and use `: TypeAlias`, but that then wouldn't work at runtime if there's unbound typevars present.

trio/_core/_local.py
TeamSpen210: I guess it doesn't backtrack from `return token` to determine `token: RunVarToken[T]`. Might be better to just add a variable annotation before the `try` block for it, then it should be able to infer.

trio/_socket.py
TeamSpen210: We'd want `operator.index()` here. `int()` would allow floats/strings which probably isn't a good thing...

trio/_core/_local.py
jakkdl: Something something inheritance seems to break the typevar (or maybe because one is `trio._socket.T` and one is `trio._util.T`), so line 72 needs an explicit `T` (otherwise I get `Incompatible types in assignment (expression has type "RunVarToken[T]", variable has type "RunVarToken[T]")`) and a variable annotation before the `try` doesn't help. But I could remove the one at line 70.

trio/_socket.py
jakkdl: yep, removing it has no effect. Though idk if having redundant explicit type hints is a bad thing.

trio/_socket.py
jakkdl: yep, they were forwarded with `getattr()` and didn't have it's parameters typed out at all.

trio/_socket.py
jakkdl: I'm not sure what you're referring to here, I'm not finding what's changed and this is what's in typeshed.

jakkdl: I'm not sure what you're referring to here, I'm not finding what's changed and this is what's in typeshed.

trio/_socket.py
jakkdl: taken from typeshed, I've been thinking I should improve them upstream.

trio/_socket.py
jakkdl: clarified the comment

trio/_socket.py
TeamSpen210: 3.8 introduced `/` pos arguments, so we can use that instead of the `__arg` hack.

jakkdl: Changed it in a bunch of places, but I failed to remove them from `sendto`, and from the ones where `_make_simple_sock_method_wrapper` is used to verify the signature.

trio/_socket.py
```diff
@@ -451,36 +548,69 @@ def __init__(self, sock: _stdlib_socket.socket):
...
+        self,
+        /,
+        level: int,
+        optname: int,
+        value: int | Buffer | None,
+        optlen: int | None = None,
+    ) -> None:
+        if optlen is None:
+            return self._sock.setsockopt(level, optname, cast("int|Buffer", value))
+        return self._sock.setsockopt(level, optname, cast(None, value), optlen)
```
TeamSpen210, line 597: Looks like we need a test using `value = None`. Instead of the casts here, perhaps it’d better to add an a check in each side of the branch to verify that value is or is not `None`, so that can be eliminated from the union. Better to be a `raise` instead of `assert` since it’s user code that would be wrong.

jakkdl: Implemented :+1:

TeamSpen210: The first exception message is wrong, `optlen` isn’t set there?

TeamSpen210 (approved):
Aside from the sockopt message, everything looks good here.

jakkdl:
I skimmed through socket documentation on windows and failed to find something that seemed interesting - might even be none, so I'll just skip testing that.

It looks like pypy doesn't support optlen at all. I feel like a comment at the call site that will pop up in the traceback should be good enough if it ever pops up that somebody runs pypy (uncommon) and tries to use the optlen argument (very uncommon - it wasn't even tested for in the trio tests previously). But could also wrap the call in a try/except. Typeshed doesn't make a distinction for pypy, nor do type checkers support guards using `sys.implementation.name` afaik

trio/_core/_local.py
```diff
@@ -1,25 +1,34 @@
...
 
-from .._util import Final
+from .._util import Final, NoPublicConstructor
 from . import _run
 
+T = TypeVar("T")
+
+
+@final
+class _NoValue:
```
A5rocks, line 15:
```suggestion
class _NoValue(metaclass=Final):
```

trio/_socket.py
```diff
@@ -451,36 +547,80 @@ def __init__(self, sock: _stdlib_socket.socket):
...
+            )
+
+        # Note: PyPy may crash here due to setsockopt only supporting
+        # four parameters.
+        return self._sock.setsockopt(level, optname, value, optlen)
+
+    def listen(self, /, backlog: int = min(_stdlib_socket.SOMAXCONN, 128)) -> None:
+        return self._sock.listen(backlog)
+
+    def get_inheritable(self) -> bool:
```
A5rocks, line 612: I don't understand why some methods (e.g. `listen`, `setsockopt`) have pos-only `self` params, whereas some (e.g. `get_inheritable`, `set_inheritable`) don't.

trio/_socket.py
```diff
@@ -575,7 +719,19 @@ async def _resolve_address_nocp(self, address, *, local):
             local=local,
         )
 
-    async def _nonblocking_helper(self, fn, args, kwargs, wait_fn):
+    # args and kwargs must be starred, otherwise pyright complains:
+    # '"args" member of ParamSpec is valid only when used with *args parameter'
+    # '"kwargs" member of ParamSpec is valid only when used with **kwargs parameter'
+    # wait_fn and fn must also be first in the signature
+    # 'Keyword parameter cannot appear in signature after ParamSpec args parameter'
+
```
A5rocks, line 727: FWIW this isn't just a pyright thing as both these requirements come directly from the Parameter Specification PEP.

trio/_socket.py
```diff
@@ -690,77 +848,146 @@ async def connect(self, address):
...
     ################################################################
 
+    # Not possible to typecheck with a Callable (due to DefaultArg), nor with a
+    # callback Protocol (https://github.com/python/typing/discussions/1040)
+    # but this seems to work. If not explicitly defined then pyright --verifytypes will
+    # complain about AmbiguousType
     if TYPE_CHECKING:
 
-        async def recv(self, buffersize: int, flags: int = 0) -> bytes:
+        def recv(__self, __buflen: int, __flags: int = 0) -> Awaitable[bytes]:
```
A5rocks, line 863: Don't forget `/`!

trio/_socket.py
```diff
@@ -690,77 +848,146 @@ async def connect(self, address):
...
 
     ################################################################
     # recv_into
     ################################################################
 
-    recv_into = _make_simple_sock_method_wrapper("recv_into", _core.wait_readable)
+    if TYPE_CHECKING:
+
+        def recv_into(
+            __self, buffer: Buffer, nbytes: int = 0, flags: int = 0
```
A5rocks, line 880: Another pos-only parameter than can use `/`: I won't comment on any others, as you can just ctrl+f "`__`"

trio/_socket.py
A5rocks: Go ahead, typeshed loves contributions! (If you haven't submitted a PR there, it's an experience: they merge them so quickly!)

trio/_sync.py
```diff
@@ -143,10 +143,15 @@ class CapacityLimiterStatistics:
...
-    borrowers: list[Task] = attr.ib()
+    borrowers: list[Task | object] = attr.ib()
     tasks_waiting: int = attr.ib()
 
 
+# Can be a generic type with a default of Task if/when PEP 696 is released
+# and implemented in type checkers. Making it fully generic would currently
+# introduce a lot of unnecessary hassle.
+
+
```
A5rocks, line 154: Extra lines!

A5rocks (changes requested):
Looks good to me other than these. I don't think I need to rereview again!

trio/_socket.py
jakkdl: As I mentioned above (but gosh, navigating discussion threads in long PR's are impossible):
> Changed it in a bunch of places, but I failed to remove them from sendto, and from the ones where _make_simple_sock_method_wrapper is used to verify the signature.

As it checks if the signature is identical to typeshed it will require updating those. I'll add a comment in the code

trio/_socket.py
jakkdl: Huh, weird that mypy didn't complain about it then.

trio/_socket.py
jakkdl: Oh, I do have a comment for that already.
https://github.com/python-trio/trio/blob/8a8b0b7964e870a0d22963805677065827b65a39/trio/_socket.py#L866-L868

trio/_socket.py
jakkdl: Uh, neither `listen` nor `setsockopt` have pos-only `self` params? The only ones that do are ones that use `__self` to match typeshed.

I guess there's some inconsistency with whether methods use pos-only at all, but that inconsistency seems to be there in typeshed as well (`set_inheritable` doesn't have pos-only in typeshed) - and there were some mismatches previously (`bind` has pos-only `address` in typeshed, but previous versions of trio didn't, so I don't think I should change that and possibly break existing code).

jakkdl:
and even more relevant: https://github.com/python/typeshed/issues/10113

PR #2706, opened 2023-07-16
Define types for `AsyncIOWrapper` and `trio.Path`

TeamSpen210:
This fully types both these wrapper classes. Dealing with the complicated hierarchy of IO methods was tricky, until I realised self-type-protocols can be used to make `AsyncIOWrapper` only have the methods that the underlying file does. It does unfortunately require a rather verbose set of definitions. It could be simplified a little by grouping methods that always are defined together, but that's less accurate.

Due to the length of the stub definitions, I did turn off Black formatting there so they could be formatted compactly like type stubs, otherwise the line count would get much longer and harder to read. Not sure what's desirable either way?


trio/_file_io.py
```diff
@@ -48,7 +74,132 @@
...
+FileT = TypeVar("FileT")
+FileT_co = TypeVar("FileT_co", covariant=True)
+T = TypeVar("T")
+T_co = TypeVar("T_co", covariant=True)
+T_contra = TypeVar("T_contra", contravariant=True)
+AnyStr_co = TypeVar("AnyStr_co", str, bytes, covariant=True)
+AnyStr_contra = TypeVar("AnyStr_contra", str, bytes, contravariant=True)
+
+# Define a protocol for every method/property we expose. Each is effectively a predicate checking
+# whether a class defines this method/property.
```
jakkdl, line 86: This .. is kinda magic to me, and probably to lots of other people as well. Could you add a couple more lines explaining how/why this works?

trio/_file_io.py
```diff
@@ -48,7 +74,132 @@
...
 
 
-class AsyncIOWrapper(AsyncResource):
+FileT = TypeVar("FileT")
+FileT_co = TypeVar("FileT_co", covariant=True)
+T = TypeVar("T")
+T_co = TypeVar("T_co", covariant=True)
+T_contra = TypeVar("T_contra", contravariant=True)
+AnyStr_co = TypeVar("AnyStr_co", str, bytes, covariant=True)
+AnyStr_contra = TypeVar("AnyStr_contra", str, bytes, contravariant=True)
```
jakkdl, line 83: These types might also warrant some comments. One day I'll get comfortable with co- & contravariant - but some brief comments here saying which ones are used where and why will probably help readers and any future people trying to modify it.

trio/_file_io.py
```diff
@@ -125,18 +278,169 @@ async def aclose(self):
...
+    file: _OpenFile,
+    mode: OpenTextMode = "r",
+    buffering: int = -1,
+    encoding: str | None = None,
+    errors: str | None = None,
+    newline: str | None = None,
+    closefd: bool = True,
+    opener: _Opener | None = None,
+) -> AsyncIOWrapper[io.TextIOWrapper]:
+    ...
```
jakkdl, line 346: Let's get a separate single-line PR that quickly adds `@overload` to `exclude_lines` in `.coveragerc`.

trio/_path.py
```diff
@@ -189,75 +294,106 @@ async def open(self, *args, **kwargs):
 
     if TYPE_CHECKING:
         # the dunders listed in _forward_magic that aren't seen otherwise
```
jakkdl, line 296: It would be nice to get some auto-checking of these signatures - partly to ensure the current signatures are correct, but even more to ensure they're kept up to date with new python versions. Most crude way would be to write a test, another way would be something akin to what I wrote in `_socket`
https://github.com/python-trio/trio/blob/aa32d87dd620511d2f225fe2f7a2f607511636d5/trio/_socket.py#L881-L890
but maybe not if that means this section has to get *even* longer.

jakkdl (changes requested):
Looks ... about as good as it could look :sweat_smile:

I would like to see you enable stricter checks for these files in `pyproject.toml` as I've done in my PR's: https://github.com/python-trio/trio/pull/2705/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711
At least `disallow_untyped_defs` - but `disallow_any_expr` would be nice as well. My thinking was that we'll gradually allow those for all files (at some point, transitioning to it being an ignore-list instead of an enable-list).

Are there really no changes to `_tests/verify_types.json`? Looks like there's 14 lines referencing `_path` atm:
https://github.com/python-trio/trio/blob/6f187fb37795829773acedbb5c0ee9e146bffb6d/trio/_tests/verify_types.json#L106-L119

and `open_file` & `wrap_file` seems to be in there as well.

trio/_file_io.py
TeamSpen210: I mainly picked/needed them because protocols complain if you don't have the right ones :)

trio/_path.py
TeamSpen210: Unfortunately it's a little tricky, since only some methods are sync vs async, `pathlib.Path` is rewrapped to `trio.Path` in return values, version-only variants, etc.

jakkdl: yeah it'd need some wrapping and modifying of the types. But it's not a critical wrapper or anything so we're probably fine even if some mistakes would slip through.

jakkdl (approved):
Looks great!

trio/_util.py
```diff
@@ -199,10 +202,14 @@ def __exit__(
         self._held = False
 
 
-def async_wraps(cls, wrapped_cls, attr_name):
+def async_wraps(
+    cls: type[t.Any],
+    wrapped_cls: type[t.Any],
```
A5rocks, line 207:
```suggestion
    cls: type[object],
    wrapped_cls: type[object],
```

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
+# This is a little complicated. IO objects have a lot of methods, and which are available on
+# different types varies wildly. We want to match the interface of whatever file we're wrapping.
+# This pile of protocols each has one sync method/property, meaning they're going to be compatible
+# with a file class that supports that method/property. The ones parameterized with AnyStr take
+# either str or bytes depending.
+
+# The wrapper is then a generic class, where the typevar is set to the type of the sync file we're
+# wrapping. For generics, adding a type to self has a special meaning - properties/methods can be
+# conditional - it's only valid to call them if the object you're accessing them on is compatible
+# with that type hint. By using the protocols, the type checker will be checking to see if the
```
A5rocks, line 94: Maybe link to mypy's docs on self-types? They're an amazingly powerful construct and more explanation might be nice for some people!

Namely this: https://mypy.readthedocs.io/en/stable/more_types.html#advanced-uses-of-self-types

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
+
+    class _HasFileNo(Protocol):
+        def fileno(self) -> int: ...
+
+    class _HasIsATTY(Protocol):
+        def isatty(self) -> bool: ...
+
+    class _HasNewlines(Protocol):
+        @property
+        def newlines(self) -> Any: ...  # None, str or tuple
```
A5rocks, line 122: I don't really understand this comment.

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
+        def getvalue(self) -> AnyStr_co: ...
+
+    class _CanGetBuffer(Protocol):
+        def getbuffer(self) -> memoryview: ...
+
+    class _CanFlush(Protocol):
+        def flush(self) -> None: ...
+
+    class _CanRead(Protocol[AnyStr_co]):
+        def read(self, __size: int | None = ...) -> AnyStr_co: ...
```
A5rocks, line 167: We dropped 3.7 support!

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
+
+    class _CanRead1(Protocol):
+        def read1(self, __size: int | None = ...) -> bytes: ...
```
A5rocks, line 170: Same here, this can be a pos-only parameter. (I won't comment on any others like this, just note same comment applies to them!)

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
+        def tell(self) -> int: ...
+
+    class _CanTruncate(Protocol):
+        def truncate(self, __size: int | None = ...) -> int: ...
+
+    class _CanWrite(Protocol[AnyStr_contra]):
+        def write(self, __data: AnyStr_contra) -> int: ...
+
+    class _CanWriteLines(Protocol[T_contra]):
+        """The lines parameter varies for bytes/str, so use a typevar to make the async match."""
```
A5rocks, line 200: This is probably supposed to be a comment?

It would be nice if `T_contra` was specialized to be `bound=bytes | str` but that's enforced anyways. I'm fine without that change.

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
+        def writelines(self, __lines: Iterable[T_contra]) -> None: ...
+
+    class _CanPeek(Protocol[AnyStr_co]):
+        def peek(self, __size: int = 0) -> AnyStr_co: ...
+
+    class _CanDetach(Protocol[T_co]):
+        """The T typevar will be the unbuffered/binary file this file wraps."""
```
A5rocks, line 207: This was probably also supposed to be a comment? I guess describing typevars works for a docstring, but it would at least be standardized how you do that.

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
 
-    All properties and methods defined in in :mod:`~io` are exposed by this
+    All properties and methods defined in :mod:`~io` are exposed by this
     wrapper, if they exist in the wrapped file object.
-
     """
 
-    def __init__(self, file):
+    # FileT needs to be covariant for the protocol trick to work - the real IO types are a subtype
+    # of the protocols.
```
A5rocks, line 224: I think this belongs above the `class` definition line (where you first subclass from `Generic[FileT_co]`.

trio/_file_io.py
```diff
@@ -48,59 +74,199 @@
...
+# The wrapper is then a generic class, where the typevar is set to the type of the sync file we're
+# wrapping. For generics, adding a type to self has a special meaning - properties/methods can be
+# conditional - it's only valid to call them if the object you're accessing them on is compatible
+# with that type hint. By using the protocols, the type checker will be checking to see if the
+# wrapped type has that method, and only allow the methods that do to be called. We can then alter
+# the signature however it needs to match runtime behaviour.
+if TYPE_CHECKING:
+    from typing_extensions import Buffer, Protocol
+
+    # fmt: off
```
A5rocks, line 100: ... I don't think you ever reenabled black formatting?

trio/_file_io.py
```diff
@@ -125,18 +291,167 @@ async def aclose(self):
 
         await trio.lowlevel.checkpoint_if_cancelled()
 
+    if TYPE_CHECKING:
+        # fmt: off
```
A5rocks, line 295: I don't think you ever reenable black formatting?

trio/_file_io.py
```diff
@@ -125,18 +291,167 @@ async def aclose(self):
...
+        async def readinto(self: AsyncIOWrapper[_CanReadInto], __buf: Buffer) -> int | None: ...
+        async def readline(self: AsyncIOWrapper[_CanReadLine[AnyStr]], __size: int = -1) -> AnyStr: ...
+        async def readlines(self: AsyncIOWrapper[_CanReadLines[AnyStr]]) -> list[AnyStr]: ...
+        async def seek(self: AsyncIOWrapper[_CanSeek], __target: int, __whence: int = 0) -> int: ...
+        async def tell(self: AsyncIOWrapper[_CanTell]) -> int: ...
+        async def truncate(self: AsyncIOWrapper[_CanTruncate], __size: int | None = None) -> int: ...
+        async def write(self: AsyncIOWrapper[_CanWrite[AnyStr]], __data: AnyStr) -> int: ...
+        async def writelines(self: AsyncIOWrapper[_CanWriteLines[T]], __lines: Iterable[T]) -> None: ...
+        async def readinto1(self: AsyncIOWrapper[_CanReadInto1], __buffer: Buffer) -> int: ...
+        async def peek(self: AsyncIOWrapper[_CanPeek[AnyStr]], __size: int = 0) -> AnyStr: ...
```
A5rocks, line 338: Can you write a short test that verifies the `_FILE_SYNC_ATTRS`/... are present? Could be as simple as seeing if `\n\s+def {name}` matches (basic regex check) if we want to make sure async/non-async is correct. Alternatively, just to check presence, we could just check they pop up in some pyright output? (idk how I feel about that...)

trio/_file_io.py
A5rocks: This was done in this PR.

trio/_file_io.py
```diff
@@ -125,18 +291,167 @@ async def aclose(self):
...
+async def open_file(
+    file: _OpenFile,
+    mode: str,
+    buffering: int = -1,
+    encoding: str | None = None,
+    errors: str | None = None,
+    newline: str | None = None,
+    closefd: bool = True,
+    opener: _Opener | None = None,
+) -> AsyncIOWrapper[IO[Any]]:
```
A5rocks, line 440: I don't really like this `Any` but I suppose practicality over purity...

trio/_path.py
```diff
@@ -1,49 +1,86 @@
...
 
-def _forward_factory(cls, attr_name, attr):
+def _forward_factory(
+    cls: AsyncAutoWrapperType,
+    attr_name: str,
+    attr: Callable[Concatenate[pathlib.Path, P], T],
+) -> Callable[Concatenate[Path, P], T | Path]:
     @wraps(attr)
-    def wrapper(self, *args, **kwargs):
+    def wrapper(self: Path, *args: P.args, **kwargs: P.kwargs) -> Path | Any:
```
A5rocks, line 47: You can use `T` instead of `Any` here. Unless there's a mypy bug (I've experienced some in this kinda place before).

trio/_path.py
```diff
@@ -65,20 +104,31 @@ async def wrapper(self, *args, **kwargs):
...
 
 
-def classmethod_wrapper_factory(cls, meth_name):
-    @classmethod
+def classmethod_wrapper_factory(
+    cls: AsyncAutoWrapperType, meth_name: str
+) -> classmethod:
     @async_wraps(cls, cls._wraps, meth_name)
-    async def wrapper(cls, *args, **kwargs):
+    async def wrapper(cls: type[Path], *args: Any, **kwargs: Any) -> Path:
```
A5rocks, line 111: I don't really like the `Any`s here, I presume paramspecs are hard to put in place? If so then eh, not a priority, no need to fix.

A5rocks (changes requested):
My eyes glazed over when looking over `trio/_path.py`. I'll take a look later.

trio/_file_io.py
TeamSpen210: It corresponds to what's in [typeshed for regular `open`](https://github.com/python/typeshed/blob/754473a30d50b918ac6f1a824f41600806defa94/stdlib/builtins.pyi#L1624). That overload's only going to be taken for a non-literal mode, where we don't really have any idea what IO object we're getting...

trio/_file_io.py
TeamSpen210: I thought I read that it'd just re-enable after the dedent, but perhaps not.

trio/_file_io.py
TeamSpen210: That's a very good idea yes.

trio/_file_io.py
A5rocks: Oh it does. My bad.

trio/_path.py
TeamSpen210: In several of these functions the method isn't passed in at all, so there's nothing to bind the paramspec to. I didn't think it would be particularly useful to try defining a bunch of protocols/vars and rearranging this code to make it type safe. It's too dynamic for that to really validate anything, and these are private anyway.

trio/_file_io.py
TeamSpen210: I preserved it from the `io` type hints, looks like the docs say this returns those but typeshed chose to make it `Any`. Actually we could just use a typevar here to do whatever the file object says it returns.

trio/_file_io.py
```diff
@@ -117,9 +118,10 @@ def fileno(self) -> int: ...
     class _HasIsATTY(Protocol):
         def isatty(self) -> bool: ...
 
-    class _HasNewlines(Protocol):
+    class _HasNewlines(Protocol[T_co]):
+        # Type varies here - documented to be None, tuple of strings, strings. Typeshed uses Any.
         @property
-        def newlines(self) -> Any: ...  # None, str or tuple
+        def newlines(self) -> T: ...
```
A5rocks:
```suggestion
        def newlines(self) -> T_co: ...
```

Fuyukai:
Posting in this issue too that type hinting ``trio.Path`` implies it is a legitimate class and not a bastard hack that needs to be prosecuted with prejudice.

trio/_path.py
```diff
@@ -1,49 +1,86 @@
...
+        OpenBinaryModeUpdating,
+        OpenBinaryModeWriting,
+        OpenTextMode,
+    )
+    from typing_extensions import Concatenate, Literal, ParamSpec, TypeAlias
+
+    P = ParamSpec("P")
+
+T = TypeVar("T")
+StrPath: TypeAlias = "str | os.PathLike[str]"
```
A5rocks, line 30: Could you make this a `typing.Union` and not have to keep it as a string? I think that's the approach we've taken elsewhere. It would be nice to be consistent!

trio/_path.py
```diff
@@ -1,49 +1,86 @@
...
+
     return wrapper
 
 
-def _forward_magic(cls, attr):
-    sentinel = object()
+def _forward_magic(
+    cls: AsyncAutoWrapperType, attr: Callable[..., object]
+) -> Callable[..., Path | Any]:
+    sentinel: Any = object()
```
A5rocks, line 63: Does this have to be `Any`?

A5rocks (approved):
I'm just going to trust this is all good!

trio/_path.py
TeamSpen210: Actually no, since `other` is also just `object`.

jakkdl:
:rocket:

PR #2713, opened 2023-07-21
set merge strategy for verify_types.json in .gitattributes

jakkdl:
Quickly merging this small one to ease merging my typechecking branches into eachother.


TeamSpen210 (approved):
Forcing a conflict makes sense, just regenerate it in that situation.

PR #2717, opened 2023-07-28
typecheck trio/_core/_unbounded_queue

jakkdl:


TeamSpen210 (approved):
Looks good to me, a straightforward one.

trio/_core/_unbounded_queue.py
```diff
@@ -41,26 +60,27 @@ class UnboundedQueue(metaclass=Final):
 
     """
 
+    # deprecated is not typed
```
A5rocks, line 63: This isn't true anymore

trio/_core/_unbounded_queue.py
```diff
@@ -128,22 +148,14 @@ async def get_batch(self):
...
-        Currently the following fields are defined:
-
-        * ``qsize``: The number of items currently in the queue.
-        * ``tasks_waiting``: The number of tasks blocked on this queue's
-          :meth:`get_batch` method.
-
-        """
-        return _UnboundedQueueStats(
+    def statistics(self) -> UnboundedQueueStatistics:
+        """Return an UnboundedQueueStatistics object containing debugging information."""
```
A5rocks, line 152:
```suggestion
        """Return an `UnboundedQueueStatistics` object containing debugging information."""
```

... maybe that's nicer?

A5rocks (changes requested):
Probably just have to merge in master

trio/_core/_unbounded_queue.py
jakkdl: Or ```""":class:`UnboundedQueueStatistics`"""```, though as it's deprecated and not documented anywhere it doesn't matter all that much.

PR #2718, opened 2023-07-28
typecheck trio/_core/_entry_queue.py

jakkdl:
Somewhat sad signatures, but not sure if it's possible to do more without `TypeVarTuple`.


A5rocks (approved):
Seems trivial enough. I agree with the assessment that this isn't possible w/o `TypeVarTuple`s.

trio/_core/_entry_queue.py
```diff
@@ -63,14 +72,16 @@ async def kill_everything(exc):
                     # system nursery is already closed.
                     # TODO(2020-06): this is a gross hack and should
                     # be fixed soon when we address #1607.
-                    _core.current_task().parent_nursery.start_soon(kill_everything, exc)
+                    parent_nursery = _core.current_task().parent_nursery
+                    assert parent_nursery is not None
```
TeamSpen210, line 76: Instead of an assert, it might be better to manually raise an `AssertionError`, so we can chain with `exc` not the `RuntimeError`. Shouldn't actually happen though.

trio/_core/_entry_queue.py
```diff
@@ -43,7 +50,9 @@ async def task(self):
         #     https://bugs.python.org/issue13697#msg237140
         assert self.lock.__class__.__module__ == "_thread"
 
-        def run_cb(job):
+        def run_cb(
+            job: tuple[Callable[..., object], Iterable[object]]
+        ) -> Literal[True]:
```
TeamSpen210, line 55: Nothing ever uses this return value, might as well just remove it while we're here.

TeamSpen210 (approved):
Looks fine overall. Actually even `TypeVarTuple` can't fully help type this, since there isn't a way to make a typevar with the scope of an individual item in a collection. It'd make the public interface safe though.

Since it's repeated a lot, perhaps a type alias for the callable/args tuple might help readability?

PR #2719, opened 2023-07-29
Add types to some internal modules (`_util`, `_deprecate`, `_ki`)

TeamSpen210:
Noticed these weren't typed, which will cause issues when typing the decorated functions. Fairly straightforward, with only two functional changes: fixing an invalid use of the deprecation logic which would include the URL twice, and making `ki_protection_enabled` coerce `LOCALS_KEY_KI_PROTECTION_ENABLED` to a bool. I think that should be more correct anyway, in case someone assigns a non-bool for some reason.


trio/_util.py
```diff
@@ -14,10 +14,23 @@
...
+
+if t.TYPE_CHECKING:
+
+    def signal_raise(signum: int) -> None:
+        ...  # Don't check the implementation below, pthread_kill does not exist on Windows.
 
 
 # Equivalent to the C function raise(), which Python doesn't wrap
-if os.name == "nt":
+elif os.name == "nt":
```
jakkdl, line 33: unrelated to PR, but I've mostly seen this as `sys.platform == "win32"` in other files. Any reason to prefer one over the other?

trio/_util.py
```diff
@@ -338,7 +354,7 @@ def _create(cls: t.Type[T], *args: object, **kwargs: object) -> T:
         return super().__call__(*args, **kwargs)  # type: ignore
 
 
-def name_asyncgen(agen):
+def name_asyncgen(agen: t.Any) -> str:
```
jakkdl, line 357: any reason this is left as `Any`?

jakkdl (changes requested):
Should update `pyproject.toml` to turn up the strictness on these files.

trio/_util.py
TeamSpen210: Hmm, looking at the docs, `os.name` is more coarse, and would just give `posix` I think for Linux, MacOS, etc. But actually [type checkers](https://typing.readthedocs.io/en/latest/source/stubs.html#version-and-platform-checks) are only required to support `platform`.

trio/_util.py
TeamSpen210: The `hasattr` check made me expect that this was intended to be callable on any object, with a graceful fallback. But looking at the call site, it's only going to be called via `sys.set_asyncgen_hooks()`. So it should always be a real async gen object, or something that's duck typing very well (Cython).

jakkdl: the docstring looks to imply it should only be called with AsyncGen iterators.

trio/_util.py
jakkdl: `git grep` gives `os.name` 15 times across 11 files, `sys.platform` 57 times across 29 files. Sounds like something for a dedicated PR

jakkdl: Looking at SO, seems like `os.name`s coarseness has it's advantages when we don't care about mac/linux distinction if we want users to be able to attempt running `trio` on weird:tm: systems, as `sys.platform` can return values outside of win32/linux/darwin.

trio/__init__.py
```diff
@@ -117,7 +117,7 @@
 
 _deprecate.enable_attribute_deprecations(__name__)
 
-__deprecated_attributes__ = {
+__deprecated_attributes__: "dict[str, _deprecate.DeprecatedAttribute]" = {
```
jakkdl, line 120: can do `from __future__ import annotations` to not require quoting.

trio/_core/_ki.py
```diff
@@ -173,26 +183,34 @@ def wrapper(*args, **kwargs):
...
 
 
 @attr.s
 class KIManager:
     handler = attr.ib(default=None)
 
-    def install(self, deliver_cb, restrict_keyboard_interrupt_to_checkpoints):
+    def install(
+        self,
+        deliver_cb: Callable[[], object],
```
jakkdl, line 203: I had typed `deliver_cb` as `Callable[[], None]` to indicate that the return value is unused. Ofc that means in exchange that you get a type error if you pass in a function that incidentally returns a value/is used elsewhere for it's return value/is not directly from the users code. Not sure which is superior/standard practice.

trio/_util.py
```diff
@@ -224,10 +240,10 @@ def decorator(func: CallT) -> CallT:
     return decorator
 
 
-def fixup_module_metadata(module_name, namespace):
-    seen_ids = set()
+def fixup_module_metadata(module_name: str, namespace: dict[str, t.Any]) -> None:
+    seen_ids: set[int] = set()
 
-    def fix_one(qualname, name, obj):
+    def fix_one(qualname: str, name: str, obj: t.Any) -> None:
```
jakkdl, line 246: The function body assumes nothing about the values in the dict, so might as well use `object` instead of `t.Any` to keep it that way.

jakkdl (changes requested):
attempted a merge with #2721 and only noticed some minor noteworthy things, as well as Generics without types you'll notice when you up strictness.

trio/__init__.py
TeamSpen210: Indeed, but since there's only one type hint in the whole module this seemed a little tidier.

trio/_core/_ki.py
TeamSpen210: [Best practice](https://typing.readthedocs.io/en/latest/source/best_practices.html#using-any-and-object) is to use `object` if you're ignoring the return type, since a function that does return something would be valid here.

trio/_util.py
```diff
@@ -183,11 +199,11 @@ class ConflictDetector:
 
     """
 
-    def __init__(self, msg):
+    def __init__(self, msg: str):
```
CoolCat467, line 202: Add ` -> None`

TeamSpen210: It's not actually required. Type checkers know that `__init__` will always return `None`, so it's fine to leave it unannotated as long as at least one parameter is annotated to prevent the gradual typing logic from just skipping the method.

trio/__init__.py
```diff
@@ -1,5 +1,6 @@
 """Trio - A friendly Python library for async concurrency and I/O
 """
+from __future__ import annotations  # isort: skip
```
CoolCat467, line 3: I don't think you should have to tell isort to skip a `__future__` import

trio/_core/_ki.py
```diff
@@ -3,17 +3,21 @@
 import inspect
 import signal
 import sys
+import types
 from functools import wraps
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Callable, Final, TypeVar
```
CoolCat467, line 8: I feel like it is better practice to `import Callable from collections.abc` instead of from typing, as the typing version is just an alias

trio/_core/_ki.py
```diff
@@ -115,29 +119,35 @@ def currently_ki_protected():
...
     return getattr(obj, "_async_gen_function", None) == id(obj)
 
 
-def _ki_protection_decorator(enabled):
-    def decorator(fn):
+def _ki_protection_decorator(
+    enabled: bool,
+) -> Callable[[Callable[ArgsT, RetT]], Callable[ArgsT, RetT]]:
+    # The "ignore[return-value]" below is because the inspect functions cast away the
+    # original return type of fn, making it just CoroutineType[Any, Any, Any] etc.
```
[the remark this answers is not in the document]
CoolCat467: You could cast it back yourself instead of ignoring it

trio/_util.py
```diff
@@ -338,7 +358,7 @@ def _create(cls: t.Type[T], *args: object, **kwargs: object) -> T:
         return super().__call__(*args, **kwargs)  # type: ignore
 
 
-def name_asyncgen(agen):
+def name_asyncgen(agen: collections.abc.AsyncGenerator[object, t.NoReturn]) -> str:
```
CoolCat467, line 361: In #2726 I ended up using `types.AsyncGeneratorType` here instead because it has the attributes this function uses defined unlike `collections.abc.AsyncGenerator`, and then you don't have errors that `__qualname__` doesn't exist

CoolCat467 (changes requested):

trio/_core/_ki.py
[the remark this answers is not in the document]
A5rocks: I really dislike how typeguards don't intersect when narrowing :(

trio/_deprecate.py
```diff
@@ -87,10 +107,16 @@ def wrapper(*args, **kwargs):
...
+    new_fn: Callable[ArgsT, RetT],
+    version: str,
+    *,
+    issue: int | None,
+) -> Callable[ArgsT, RetT]:
     @deprecated(version, issue=issue, instead=new_fn)
     @wraps(new_fn, assigned=("__module__", "__annotations__"))
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: ArgsT.args, **kwargs: ArgsT.kwargs) -> RetT:
         "Deprecated alias."
```
A5rocks, line 120: Should this use 3 quotes?

trio/_util.py
```diff
@@ -224,10 +240,10 @@ def decorator(func: CallT) -> CallT:
     return decorator
 
 
-def fixup_module_metadata(module_name, namespace):
-    seen_ids = set()
+def fixup_module_metadata(module_name: str, namespace: dict[str, t.Any]) -> None:
```
A5rocks, line 243:
```suggestion
def fixup_module_metadata(module_name: str, namespace: dict[str, object]) -> None:
```

You end up passing it as an `object` to `fix_one` anyways.

trio/__init__.py
TeamSpen210: You shouldn't no, it's because of the `skip` directive on the next import. But using `split` instead behaves more correctly.

trio/_util.py
```diff
@@ -252,7 +269,10 @@ def fix_one(qualname, name, obj):
             fix_one(objname, objname, obj)
 
 
-class generic_function:
+# We need ParamSpec to type this "properly", but that requires a runtime typing_extensions import
+# to use as a class base. This is only used at runtime and isn't correct for type checkers anyway,
+# so don't bother.
+class generic_function(t.Generic[RetT]):
```
A5rocks, line 275: I think a runtime import isn't too bad, but now that I think of it I forgot we don't depend on `typing_extensions` iirc...

A5rocks (approved):
Looks good

trio/_util.py
TeamSpen210: The main reason I didn't change this one is to ensure we don't have problems with variance issues on the caller side. `vars(module)` probably gives `Any` anyway though, so that wouldn't matter. Swapping to `Mapping` would also work too.

A5rocks: Oh yeah mapping's probably a nicer signature.

trio/_util.py
CoolCat467: Mypy complains if you up the strictness I'm pretty sure

TeamSpen210: Took a bit of searching, but found this [Mypy code](https://github.com/python/mypy/blob/002502a0111852c360f2255830951473bcfec4a7/mypy/semanal.py#L850-L853) which converts `-> Any` annotations to `-> None` for `__init__`. There's also a specific [error](https://github.com/python/mypy/blob/002502a0111852c360f2255830951473bcfec4a7/mypy/message_registry.py#L84) for not returning `None`.

.coveragerc
```diff
@@ -21,11 +21,14 @@ exclude_lines =
...
   @overload
 
 partial_branches =
   pragma: no branch
   if not TYPE_CHECKING:
   if not _t.TYPE_CHECKING:
+  if not t.TYPE_CHECKING:
   if .* or not TYPE_CHECKING:
   if .* or not _t.TYPE_CHECKING:
+  if .* or not t.TYPE_CHECKING:
```
jakkdl, line 34: nit, but could be made cleaner by properly utilizing regexes.

trio/_core/_ki.py
jakkdl: Good to know, thanks!
I don't think it's been in any of my PR's other than #2721

trio/_util.py
jakkdl: I don't have strong opinions either way, but if we want to enforce that it should be enabled - otherwise it's covered by `no_untyped_defs` only enforcing `-> None` on `__init__` without typed params.

trio/_core/_ki.py
jakkdl: We've had this discussion in other PRs, and mostly settled on that it doesn't really matter and once we drop older versions and run `pyupgrade` it'll get autofixed.

trio/_util.py
jakkdl: yeah, we only depend on it for tests.

jakkdl (approved):

CoolCat467 (approved):
Looks good to me if everyone agrees about the `pyproject.toml` changes

PR #2723, opened 2023-07-29
Add slots and -> None to __init__ in classes

CoolCat467:
This PR adds slots to classes and missing return None type annotations to `__init__` functions in `_dtls.py`.


TeamSpen210:
Slots definitely would be a slight improvement in a lot of cases, but we might need to be careful - they technically are backwards incompatible, since arbitrary attributes can't be added anymore.

A5rocks:
IMO I'm +1 on only adding `-> None` on `__init__`s that need it (ie where there are no other typed parameters). But this is just a styling thing which isn't too big a deal.

jakkdl:
> IMO we should only add `-> None` on `__init__`s that need it (ie where there are no other typed parameters). But this is just a styling thing which isn't too big a deal.

yeah `no_untyped_defs` will catch `__init__`s that require it, so unless we configure mypy (if that's possible? I tried looking for the config option for that) to always require it I don't think PRs should go out of their way to add them.

docs-requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.11
```
jakkdl, line 2: I could send you my (messy and inofficial) `tox.ini` if you want

jakkdl:
What's the case for adding a static `__slots__`? If it's Good:tm: should they be added everywhere? (where attr doesn't already generate them)

TeamSpen210:
Well the main purpose is that it improves memory usage, since the `__dict__` for attributes don't need to be created. And as a side effect which can also be useful, additional attribute names can't be used. Though, in 3.11+ optimisations to regular objects means that the memory improvement is basically negligible. The other issue is that they break weakrefs unless you add a slot for that (which is why the tests are all failing).

jakkdl:
> Well the main purpose is that it improves memory usage, since the `__dict__` for attributes don't need to be created. And as a side effect which can also be useful, additional attribute names can't be used. Though, in 3.11+ optimisations to regular objects means that the memory improvement is basically negligible. The other issue is that they break weakrefs unless you add a slot for that (which is why the tests are all failing).

that sounds ... not important enough to bother then imho?

CoolCat467:
Personally I always make sure to include slots in all the classes I can, because first off when you use `help()` on classes with slots, it tells you about all the non-private variables without having to dig through the docs, and additionally because it allows users down the line to use slots effectively themselves in sub-classes.

TeamSpen210:
For `help()`, really we should just add the attributes to the docstring, and/or add annotations to declare them for the class. `__slots__` just tells you the names, not what they do so it's kinda limited. Most of Trio's classes can't be subclassed, so that doesn't exactly apply. I don't really think this is necessary, it would only help on classes that are being spawned heaps of times. A lot of those use `attrs` anyway which does this automatically. 

The big issue is backwards compatibility - any user assigning custom attributes or using weakrefs is going to be broken by this. You can manually add the `__weakref__` slot back, but then the memory saving is effectively zeroed out.

jakkdl:
Closing this, adding `__slots__` does not seem to have clear benefits and been voted down, and `-> None` on `__init__`s with typed arguments doesn't do anything.

Issue #2699, opened 2023-07-12
Use tox and/or pre-commit for static analysis checks


jakkdl:
`check.sh` is getting kinda bloated with the addition of `isort`, `codespell`, and `check_type_completeness` - and trying to read the output from it to figure out what actually failed is kind of a pain. I've been using tox and pre-commit locally to avoid having to deal with the shell scripts, and been copying over configuration and commands, and maybe it's about time to share that and maybe make it official instead.

Setting up test environments in tox would also help people testing on different python versions and the like, although I've had some troubles with `test_exports/` from inside tox envs (having to mess with `PY_IGNORE_IMPORTMISMATCH`, custom `install_command`s, and the like. It's definitely doable though)

webknjaz:
Did you mean to close this via #3196?

jakkdl:
yeah this can be closed 🚀

Issue #2687, opened 2023-07-06
How much runtime introspection of types is a good thing?


A5rocks:
As I write this, one of the open typing PRs has a function reference a type behind a `if typing.TYPE_CHECKING` branch. This means that the function is not very introspectable.

I'm just opening this issue as a quick way to dump thoughts if I get any and to resolve that specific conversation. I don't have solutions and I don't really understand the tradeoffs. Here's the problem as I understand it:

The upsides are:
 - nicer types
 - nicer type syntax
 - *slightly* lower import times (... there's probably some low-hanging fruit here so I don't think this is too much a concern)

The downsides are:
 - ... runtime introspection?

See, I don't really understand the downsides. I know that there's use cases related to this but I don't know what or if they're even relevant here!

TeamSpen210:
In general it means the annotations at runtime are invalid, and can't be evaluated. Anyone passing an affected Trio type or function into a runtime typing library is going to unexpectedly get strange errors. It's also not really possible at all in some cases, like if you wanted to inherit from `Protocol`.

Additionally, I know Sphinx does runtime introspection for the `autoXXX` feature, so if we enable showing annotations there'll be piles of errors there as Sphinx tries to resolve the non-existent types. Not a problem for the syntax changes since docs would be run on the latest versions, but `TYPE_CHECKING` would break there.

A5rocks:
We talked a bit about this on Gitter and my current opinion is that we should allow language/syntax changes (e.g. newer aliases) but not a blunt application of `if typing.TYPE_CHECKING`. I'm still fine with using it to lie to type checkers about the type of something, but I don't want imaginary types floating around annotations. This way, if something does come up, people can just use a new version of Python (e.g. if Sphinx works better with runtime introspection of types instead of just doing its own thing with the string-version annotation).

jakkdl:
While hiding imports being a type-checking guard does improve import speed, in my experience the main reason is to avoid import cycles. I've many times encountered straight up two-way import dependencies A <-> B introduced due to adding type hints, and guarding the imports (or doing very fancy stuff like putting imports at the bottom of the file) has been the best way of resolving them. I also like the distinction in the import list between what's actually needed for the program to run, vs what's just type imports, to get a better sense of what's going on.


[...]

PR #2724, opened 2023-07-29
Add typing for `_highlevel_open_tcp_listeners.py`

CoolCat467:
This PR adds type annotations to `_highlevel_open_tcp_listeners.py`.

The only thing that might be a bit off is the type of `task_status` in the `serve_tcp` function. I annotated it as `trio.lowlevel.Task`, but I am not sure this is right.


A5rocks:
FYI I think `verify_types.json` is messed up a bit. Maybe your local env is a bit strange?

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -150,14 +155,14 @@ async def open_tcp_listeners(port, *, host=None, backlog=None):
...
+    handler: Callable[[trio.SocketStream], Awaitable[Any]],
+    port: int,
     *,
-    host=None,
-    backlog=None,
-    handler_nursery=None,
-    task_status=trio.TASK_STATUS_IGNORED,
-):
+    host: str | bytes | None = None,
+    backlog: int | None = None,
```
TeamSpen210, line 162: For `backlog` here, we need to check usages - a lot of Trio functions accept int + `math.inf`, which does look like it'll work here. So this should be a `float` actually. Perhaps that should be deprecated in favour of `None`.

TeamSpen210 (changes requested):
A few things that aren't quite right. `verify_types.json` is a bit picky, it doesn't give the same results for me either.

trio/_highlevel_open_tcp_listeners.py
CoolCat467: I agree, accepting float would be confusing and would not represent what we are trying to do here

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -62,7 +67,7 @@ async def open_tcp_listeners(port, *, host=None, backlog=None):
           :func:`open_tcp_listeners` will bind to both the IPv4 wildcard
           address (``0.0.0.0``) and also the IPv6 wildcard address (``::``).
 
-      host (str, bytes-like, or None): The local interface to bind to. This is
+      host (str, bytes, or None): The local interface to bind to. This is
```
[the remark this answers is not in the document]
CoolCat467: I originally had it handle bytes-convertable objects, but then `_socket.getaddrinfo` said it only handles str, bytes, or None.

trio/_highlevel_open_tcp_listeners.py
jakkdl: deprecating inf->None should be done in a separate PR, as it's used widely across the codebase. Several other functions have been typed as accepting float, so if it makes sense here I agree it should be added - though you can add a check that if it's `float` it's `math.inf`

trio/_highlevel_open_tcp_listeners.py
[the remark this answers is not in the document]
jakkdl: double-checked typeshed and it indeed doesn't take `buffer` so the docstring seems incorrect.

jakkdl:
> FYI I think `verify_types.json` is messed up a bit. Maybe your local env is a bit strange?

#2699 tox would help here.
Getting the environment right can actually be quite tricky IME

jakkdl (changes requested):
Update `pyproject.toml` to add this file to the list of files that has increased strictness.

CoolCat467:
Interesting, Windows pypy 3.9 nightly x64 failed with a segfault from the tests script: https://github.com/python-trio/trio/actions/runs/5705575738/job/15460400267?pr=2724

jakkdl:
> Interesting, Windows pypy 3.9 nightly x64 failed with a segfault from the tests script: https://github.com/python-trio/trio/actions/runs/5705575738/job/15460400267?pr=2724

it does that every now and then in my experience - no clue why. It's always been fixed by rerunning for me, but somebody should maybe open an issue and/or try to investigate the flakiness.

jakkdl (approved):
`Task` -> `TaskStatus` as TeamSpen pointed out, but otherwise mostly good. No need for a re-review from me.

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -150,14 +155,14 @@ async def open_tcp_listeners(port, *, host=None, backlog=None):
 
 
 async def serve_tcp(
-    handler,
-    port,
+    handler: Callable[[trio.SocketStream], Awaitable[Any]],
```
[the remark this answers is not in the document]
CoolCat467: Fixed in 43d229c

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -150,14 +155,14 @@ async def open_tcp_listeners(port, *, host=None, backlog=None):
...
     *,
-    host=None,
-    backlog=None,
-    handler_nursery=None,
-    task_status=trio.TASK_STATUS_IGNORED,
-):
+    host: str | bytes | None = None,
+    backlog: int | None = None,
+    handler_nursery: trio.Nursery | None = None,
+    task_status: trio.lowlevel.Task = trio.TASK_STATUS_IGNORED,  # type: ignore[has-type]  # Cannot determine type of "TASK_STATUS_IGNORED"
```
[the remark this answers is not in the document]
CoolCat467: Changed in 43d229c

trio/_highlevel_open_tcp_listeners.py
A5rocks: Bump on this @CoolCat467

CoolCat467: So making sure, we are going to make it accept `float` here then, but only return `int`?

A5rocks: I don't understand what you mean by "return" here -- I'm just talking about supporting `inf` at least in this PR and unsupporting it in another one!

CoolCat467: Fixed in e90e4cc

jakkdl (changes requested):

jakkdl (approved):

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -1,11 +1,18 @@
...
+from collections.abc import Awaitable, Callable
 from math import inf
+from typing import TYPE_CHECKING
 
 import trio
 
 from . import socket as tsocket
 
+if TYPE_CHECKING:
+    from trio.lowlevel import TaskStatus
```
TeamSpen210, line 14: Does this import need to be guarded (due to import cycles), or can it be a real import? It's probably better to just import if we can, so that this can be evaluated at runtime.

TeamSpen210 (approved):
Looks, good, except for one minor quibble.

trio/_highlevel_open_tcp_listeners.py
CoolCat467: Does not need to be guarded

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -240,6 +237,7 @@ async def serve_tcp(
 
     """
     listeners = await trio.open_tcp_listeners(port, host=host, backlog=backlog)
+    # typecheck: no-untyped-call error: Call to untyped function "serve_listeners" in typed context
```
jakkdl, line 240: nit, but I don't think we want to include action comments that refers to checkers or options not used in CI. Users should generally handle those themselves, or request them be added to CI, as there's no checking that they're kept updated, removed if no longer used, etc.

trio/_highlevel_open_tcp_listeners.py
jakkdl: https://github.com/python-trio/trio/issues/2687

This is kinda up in the air, I'd love it if we definitively resolved it and added a CI check that made usage consistent and saved review cycles. https://pypi.org/project/flake8-type-checking/

trio/_highlevel_open_tcp_listeners.py
CoolCat467: Yea that comment is not supposed to be there, that's a remnant of using my mypy extension in the text editor, I'll remove it.

CoolCat467: Removed in 6ff13c4

trio/_highlevel_open_tcp_listeners.py
CoolCat467: I would suggest using https://pypi.org/project/ruff/ instead, it does the same thing and more but faster and with auto-fix support

jakkdl:
Given the number of open typing PRs, I'm merging this one. Any minor problems can be addressed later if you want to bother going through it @A5rocks

PR #2725, opened 2023-07-29
Typing highlevel open tcp stream

CoolCat467:
This PR adds type annotations to `_highlevel_open_tcp_stream.py`

### DIFF
--- a/trio/_highlevel_open_tcp_stream.py
+++ b/trio/_highlevel_open_tcp_stream.py
@@ -1,5 +1,9 @@
+from __future__ import annotations
+
 import sys
+from collections.abc import Generator
 from contextlib import contextmanager
+from socket import AddressFamily, SocketKind
 
 import trio
 from trio._core._multierror import MultiError
@@ -109,8 +113,8 @@ DEFAULT_DELAY = 0.250
 
 
 @contextmanager
-def close_all():
-    sockets_to_close = set()
+def close_all() -> Generator[set[socket], None, None]:
+    sockets_to_close: set[socket] = set()
     try:
         yield sockets_to_close
     finally:
@@ -126,7 +130,9 @@ def close_all():
             raise MultiError(errs)
 
 
-def reorder_for_rfc_6555_section_5_4(targets):
+def reorder_for_rfc_6555_section_5_4(
+    targets: list[tuple[AddressFamily, SocketKind, int, str, tuple[str, int]]]
+) -> None:
     # RFC 6555 section 5.4 says that if getaddrinfo returns multiple address
     # families (e.g. IPv4 and IPv6), then you should make sure that your first
     # and second attempts use different families:
@@ -144,7 +150,7 @@ def reorder_for_rfc_6555_section_5_4(targets):
             break
 
 
-def format_host_port(host, port):
+def format_host_port(host: str | bytes, port: int) -> str:
     host = host.decode("ascii") if isinstance(host, bytes) else host
     if ":" in host:
         return f"[{host}]:{port}"
@@ -173,8 +179,12 @@ def format_host_port(host, port):
 #   AF_INET6: "..."}
 # this might be simpler after
 async def open_tcp_stream(
-    host, port, *, happy_eyeballs_delay=DEFAULT_DELAY, local_address=None
-):
+    host: str | bytes,
+    port: int,
+    *,
+    happy_eyeballs_delay: float = DEFAULT_DELAY,
+    local_address: str | None = None,
+) -> trio.abc.Stream:
     """Connect to the given host and port over TCP.
 
     If the given ``host`` has multiple IP addresses associated with it, then
--- a/trio/_highlevel_open_tcp_stream.py
+++ b/trio/_highlevel_open_tcp_stream.py
@@ -7,7 +7,7 @@ from socket import AddressFamily, SocketKind
 
 import trio
 from trio._core._multierror import MultiError
-from trio.socket import SOCK_STREAM, getaddrinfo, socket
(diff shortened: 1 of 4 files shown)


TeamSpen210:
This branch includes the commits for #2724 for some reason?

CoolCat467:
Rats I must have not changed branch before committing

jakkdl (changes requested):
Update `pyproject.toml` to add this file to the list of files that has increased strictness.

trio/_highlevel_open_tcp_stream.py
```diff
@@ -173,8 +187,12 @@ def format_host_port(host, port):
 #   AF_INET6: "..."}
 # this might be simpler after
 async def open_tcp_stream(
-    host, port, *, happy_eyeballs_delay=DEFAULT_DELAY, local_address=None
-):
+    host: str | bytes,
+    port: int,
+    *,
+    happy_eyeballs_delay: float | None = DEFAULT_DELAY,
+    local_address: str | None = None,
```
jakkdl, line 194: No clue why the docstring says this must be a str, but since `bind` can take an `Address` this can too.

`Address` defined at
https://github.com/CoolCat467/trio/blob/51347e3772aa54cbc5e1b431c285e01a7a813fc7/trio/_socket.py#L43

trio/_highlevel_open_tcp_stream.py
```diff
@@ -283,7 +301,11 @@ async def open_tcp_stream(
     #   the next connection attempt to start early
     # code needs to ensure sockets can be closed appropriately in the
     # face of crash or cancellation
-    async def attempt_connect(socket_args, sockaddr, attempt_failed):
+    async def attempt_connect(
+        socket_args: tuple[AddressFamily, SocketKind],
+        sockaddr: tuple[str, int] | tuple[str, int, int, int],
```
jakkdl, line 306: Use `Address` ?

trio/_highlevel_open_tcp_stream.py
```diff
@@ -126,7 +130,17 @@ def close_all():
...
 
-def reorder_for_rfc_6555_section_5_4(targets):
+def reorder_for_rfc_6555_section_5_4(
+    targets: list[
+        tuple[
+            AddressFamily,
+            SocketKind,
+            int,
+            str,
+            tuple[str, int] | tuple[str, int, int, int],
```
jakkdl, line 140: `Address`

jakkdl (changes requested):

trio/_highlevel_open_tcp_stream.py
CoolCat467: `attempt_connect` is called from unpacking the results of `getaddrinfo`, and with that unpacking, `sockaddr` can only become `tuple[str, int] | tuple[str, int, int, int]` unless the return annotation for `getaddrinfo` is wrong. I do see that it uses `sockaddr` for `socket.connect`, and `socket.connect`'s annotation says it accepts the `Address` union, but I feel like annotating `attempt_connect`'s `sockaddr` argument to be `Address` wouldn't be telling the truth about what we expect `getaddrinfo` to tell us.

trio/_highlevel_open_tcp_stream.py
CoolCat467: See previous comment on `attempt_connect`

trio/_highlevel_open_tcp_stream.py
jakkdl: Informing what `getaddrinfo` gives should be up to `getaddrinfo`, no? `attempt_connect`s signatures responsibility is saying what it accepts and returns, and it can accept `Address` no problem. Though it's a nested function it barely matters, other than if `open_tcp_stream` was changed to modify the address it got from `getaddrinfo` before sending it to `attempt_connect` - but anybody attempting such a change shouldn't be hindered by a restricting type hint on it's helper functions IMO.

CoolCat467: Changed in e82ce69

trio/_highlevel_open_tcp_stream.py
CoolCat467: Changed in e82ce69

trio/_highlevel_open_tcp_stream.py
CoolCat467: Fixed annotation in 62b3c9a

trio/_highlevel_open_tcp_stream.py
CoolCat467: Basically, accepting this causes mypy errors, and so this change was reverted in afaf153

A5rocks (approved):
This looks good to me. Let's wait on @jakkdl though!

trio/_highlevel_open_tcp_stream.py
jakkdl: Ah, digging further into the docstring and https://docs.python.org/3/library/socket.html?highlight=bind#socket-families it looks like it indeed does have to be a string.

trio/_highlevel_open_tcp_stream.py
jakkdl: I think the error on this one could be ignored / worked around as it's just about invariance of List, but given that it's only used internally there's no reason to.

jakkdl:
Played around with it locally to figure out what was going on with `Address`, and realised that the call to `nursery.start_soon` being ~untyped masked what would become a pretty nasty type error if/when it gets typed.

jakkdl (approved):

A5rocks:
Joining the organization means https://github.com/python-trio/trio/pull/2727 is unnecessary and you can contribute to the base PR!

PR #2726, opened 2023-07-30
Add typing to `_util.py`

CoolCat467:
This PR adds type annotations to `_util.py`

I added a bunch of `Any` annotations for `generic_function`, I can't think of anything better at the moment.

### DIFF
--- a/trio/_util.py
+++ b/trio/_util.py
@@ -9,9 +9,12 @@ import threading
 import typing as t
 from abc import ABCMeta
 from functools import update_wrapper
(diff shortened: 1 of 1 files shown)


A5rocks:
Oh geh, another overlap with https://github.com/python-trio/trio/pull/2719!

Mind giving that a look-over? Given you've done the same process yourself, you should probably have opinions on how it should be!

jakkdl:
Closing this one to focus on #2719

PR #2732, opened 2023-08-02
Fix version selector and maybe search for RTD

A5rocks:
This also kinda cleans up the `docs-requirements.in` file.


A5rocks:
Looks like it does fix it. https://trio--2732.org.readthedocs.build/en/2732/

Fuyukai (approved):

PR #2733, opened 2023-08-03
Add types to `_core._run`

TeamSpen210:
This adds types to the `_run` module (as much as can be done without `TypeVarTuple`). In particular I made `TaskStatus` a protocol, not a concrete class - since it's just documented as being a non-specific object with the method. Users shouldn't be depending on a specific type. Though the positional arguments can't be checked, this does make `nursery.start(...)` produce the right result, check you use `TaskStatus` correctly etc.

There's a number of interesting type errors - `Clock._autojump()` is called despite that not being defined on the ABC, and a bunch of places where `_cancel_status` is accessed without checking if it's None.


trio/_core/_run.py
```diff
@@ -780,19 +791,50 @@ def cancel_called(self) -> bool:
...
-    _new_nursery = attr.ib()
-    _called_started = attr.ib(default=False)
-    _value = attr.ib(default=None)
+class _TaskStatus(TaskStatus[StatusT]):
+    _old_nursery: Nursery = attr.ib()
+    _new_nursery: Nursery = attr.ib()
+    # _called_started indicates whether _value can be accessed. The latter can't be Optional[],
+    # since None is a valid value too.
+    _called_started: bool = attr.ib(default=False)
+    _value: StatusT = attr.ib(default=cast(StatusT, None))
```
jakkdl, line 824: How about a `NoValue` singleton?
```python
_value: StatusT | NoValue = attr.ib(default=NoValue)
```

trio/_core/_run.py
```diff
@@ -1542,7 +1600,9 @@ def current_root_task(self):
     ################
 
     @_public
-    def reschedule(self, task, next_send=_NO_SEND):
+    def reschedule(
+        self, task: trio.lowlevel.Task, next_send: object = _NO_SEND
```
jakkdl, line 1604: `next_send: Outcome` ?

jakkdl (changes requested):
The `None` cases can presumable be addressed with various `assert x is not None`, at least adding those (or `type: ignore`) makes it easier to see in the code review where it happens instead of having to cross-reference with the CI output.

I'm not sure how to handle `_autojump` in a way that's terribly helpful for end-users, as the mechanism by how it's enabled in the runner to start with is kinda weird. But since it requires explicitly setting `runner.clock_autojump_treshold` from outside the class, you're probably not hitting it unintentionally. Maybe just a try/except with a helpful error message directing them to `MockClock`, and a `type: ignore`.

trio/_core/_run.py
```diff
@@ -780,19 +791,50 @@ def cancel_called(self) -> bool:
...
-    def started(self, value=None):
+    @overload
+    def started(self: TaskStatus[None]) -> None:
+        ...
+
+    @overload
+    def started(self, value: StatusT) -> None:
+        ...
+
+    def started(self, value: StatusT = None) -> None:  # type: ignore[assignment] # StatusT != None
```
CoolCat467, line 837: Use `value: StatusT | None = None` instead of ignore

trio/_core/_generated_run.py
```diff
@@ -39,7 +45,7 @@ def current_statistics():
         raise RuntimeError("must be called from async context")
 
 
-def current_time():
+def current_time() ->float:
```
CoolCat467, line 48: `() -> float:`

trio/_core/_generated_run.py
```diff
@@ -56,7 +62,7 @@ def current_time():
         raise RuntimeError("must be called from async context")
 
 
-def current_clock():
+def current_clock() ->trio._abc.Clock:
```
CoolCat467, line 65: `() -> trio._abc.Clock`

trio/_core/_generated_run.py
```diff
@@ -65,7 +71,7 @@ def current_clock():
         raise RuntimeError("must be called from async context")
 
 
-def current_root_task():
+def current_root_task() ->(trio.lowlevel.Task | None):
```
CoolCat467, line 74: `() -> trio.lowlevel.Task | None`

trio/_core/_generated_run.py
```diff
@@ -78,7 +84,7 @@ def current_root_task():
         raise RuntimeError("must be called from async context")
 
 
-def reschedule(task, next_send=_NO_SEND):
+def reschedule(task: trio.lowlevel.Task, next_send: object=_NO_SEND) ->None:
```
CoolCat467, line 87: `) -> None`

trio/_core/_generated_run.py
```diff
@@ -103,7 +109,9 @@ def reschedule(task, next_send=_NO_SEND):
         raise RuntimeError("must be called from async context")
 
 
-def spawn_system_task(async_fn, *args, name=None, context=None):
+def spawn_system_task(async_fn: Callable[..., Awaitable[object]], *args:
+    Any, name: object=None, context: (contextvars.Context | None)=None
+    ) ->trio.lowlevel.Task:
```
CoolCat467, line 114: Fix formatting

trio/_core/_generated_run.py
```diff
@@ -162,7 +170,7 @@ def spawn_system_task(async_fn, *args, name=None, context=None):
         raise RuntimeError("must be called from async context")
 
 
-def current_trio_token():
+def current_trio_token() ->trio.lowlevel.TrioToken:
```
CoolCat467, line 173: `() -> trio.lowlevel.TrioToken`

trio/_core/_generated_run.py
```diff
@@ -174,7 +182,7 @@ def current_trio_token():
         raise RuntimeError("must be called from async context")
 
 
-async def wait_all_tasks_blocked(cushion=0.0):
+async def wait_all_tasks_blocked(cushion: float=0.0) ->None:
```
CoolCat467, line 185: `) -> None`

trio/_core/_run.py
```diff
@@ -1049,7 +1100,13 @@ def aborted(raise_cancel):
                 # (see test_nursery_cancel_doesnt_create_cyclic_garbage)
                 del self._pending_excs
 
-    def start_soon(self, async_fn, *args, name=None):
+    def start_soon(
+        self,
+        # TODO: TypeVarTuple
```
CoolCat467, line 1105: I thought I saw something about mypy being able to understand `TypeVarTuple` in one of the newer updates unless I am mistaken

trio/_core/_run.py
CoolCat467: I second this

trio/_core/_run.py
```diff
@@ -1576,8 +1636,16 @@ def reschedule(self, task, next_send=_NO_SEND):
...
 
     def spawn_impl(
-        self, async_fn, args, nursery, name, *, system_task=False, context=None
-    ):
+        self,
+        # TODO: TypeVarTuple
+        async_fn: Callable[..., Awaitable[object]],
+        args: tuple[Any, ...],
+        nursery: Nursery | None,
+        name: object,
```
CoolCat467, line 1644: `name: str | functools.partial | None`, and maybe also accepts `types.FunctionType`, but if unsure also add `object` as a catch all

CoolCat467 (changes requested):

trio/_core/_generated_run.py
TeamSpen210: This file is generated and not edited manually, the other annotations will appear when we annotate the classes elsewhere.

trio/_core/_run.py
TeamSpen210: It accepts anything. For functions it uses the name, otherwise it `str()`ifys it.

trio/_core/_generated_run.py
TeamSpen210: This is auto-generated, it's more a problem with the `astor` package.

trio/_core/_run.py
TeamSpen210: It's got experimental support - definitions work, but when Mypy tries checking function calls it tends to crash with `AssertionError`s.

jakkdl: https://github.com/python/mypy/labels/topic-pep-646

TeamSpen210:
While typing further, found a few more issues.

* `sortedcontainers` doesn't have types, and it seems is unlikely to ever have typing - it does a lot of unusual things in the constructors apparently.
* Looks like we need either my `outcome` PR or `trio_typing` in CI to have types for `outcome`, the errors are fine with that present.

CoolCat467:
Why does it say `name-defined error: Name "_core.RaiseCancelT" is not defined`?

trio/_core/_run.py
```diff
@@ -1386,26 +1506,29 @@ class _RunStatistics:
...
-    run_sync_soon_not_threadsafe = attr.ib()
-    done_callback = attr.ib()
-    unrolled_run_gen = attr.ib()
-    _value_factory: Callable[[], Value] = lambda: Value(None)
-    unrolled_run_next_send = attr.ib(factory=_value_factory, type=Outcome)
-
-    def guest_tick(self):
+    runner: Runner = attr.ib()
+    run_sync_soon_threadsafe: Callable[[Callable[[], object]], object] = attr.ib()
+    run_sync_soon_not_threadsafe: Callable[[Callable[[], object]], object] = attr.ib()
```
CoolCat467, line 1511: I thought that it was `Callable[[Callable[[], None], Any]`

trio/_core/_run.py
```diff
@@ -1524,12 +1647,12 @@ def current_time(self):
...
 
     @_public
-    def current_clock(self):
+    def current_clock(self) -> trio._abc.Clock:
         """Returns the current :class:`~trio.abc.Clock`."""
         return self.clock
 
     @_public
-    def current_root_task(self):
+    def current_root_task(self) -> trio.lowlevel.Task | None:
```
CoolCat467, line 1655: Why `trio.lowlevel.Task` when `Task` is from this file

trio/_core/_run.py
```diff
@@ -1542,7 +1665,9 @@ def current_root_task(self):
     ################
 
     @_public
-    def reschedule(self, task, next_send=_NO_SEND):
+    def reschedule(
+        self, task: trio.lowlevel.Task, next_send: Outcome[Any] = _NO_SEND
```
CoolCat467, line 1669: `trio.lowlevel.Task` -> `Task`

trio/_core/_run.py
```diff
@@ -1643,10 +1776,10 @@ async def python_wrapper(orig_coro):
             self.instruments.call("task_spawned", task)
         # Special case: normally next_send should be an Outcome, but for the
         # very first send we have to send a literal unboxed None.
-        self.reschedule(task, None)
+        self.reschedule(task, None)  # type: ignore[arg-type]
```
CoolCat467, line 1779: I know it's a special case but why not just have reschedule's annotation accept `None`?

trio/_core/_run.py
TeamSpen210: It's because of `@_public` - that copies this definition to the `_generated_run.py` file, so by using the full name we only need `import trio` in that.

trio/_core/_run.py
TeamSpen210: Same reason as `current_root_task`.

trio/_core/_run.py
```diff
@@ -1695,7 +1829,14 @@ def task_exited(self, task, outcome):
...
     @_public
-    def spawn_system_task(self, async_fn, *args, name=None, context=None):
+    def spawn_system_task(
+        self,
+        # TODO: TypeVarTuple
+        async_fn: Callable[..., Awaitable[object]],
+        *args: Any,
+        name: object = None,
+        context: contextvars.Context | None = None,
+    ) -> trio.lowlevel.Task:
```
CoolCat467, line 1839: `trio.lowlevel.Task` -> `Task`

trio/_core/_run.py
```diff
@@ -1794,7 +1940,7 @@ async def init(self, async_fn, args):
     ################
 
     @_public
-    def current_trio_token(self):
+    def current_trio_token(self) -> trio.lowlevel.TrioToken:
```
CoolCat467, line 1943: `trio.lowlevel.TrioToken` -> `TrioToken`

trio/_core/_run.py
```diff
@@ -2117,23 +2266,26 @@ def run(
...
-    async_fn,
-    *args,
-    run_sync_soon_threadsafe,
-    done_callback,
-    run_sync_soon_not_threadsafe=None,
+    async_fn: Callable[..., Awaitable[RetT]],
+    *args: Any,
+    run_sync_soon_threadsafe: Callable[[Callable[[], object]], object],
+    done_callback: Callable[[Outcome[RetT]], object],
+    run_sync_soon_not_threadsafe: Callable[[Callable[[], object]], object]
```
CoolCat467, line 2281: See other comment

trio/_core/_run.py
TeamSpen210: Well that's exposed as a public API, so I didn't want to add the special case there too.

trio/_core/_run.py
```diff
@@ -2373,8 +2526,7 @@ def unrolled_run(
                     # more Context.run adds.
                     tb = task_exc.__traceback__
                     for _ in range(1 + CONTEXT_RUN_TB_FRAMES):
-                        if tb is None:
-                            break
+                        assert tb is not None
```
CoolCat467, line 2529: Next traceback can be None in some cases, I think this check is important

CoolCat467:
Knowing the exact type of platform-dependant Event is not super important, as we give said events back to the same module in due time and never modify it ourselves

trio/_core/_run.py
TeamSpen210: I thought I added this check, and it was tripping coverage. So I should revert.

trio/_core/_run.py
TeamSpen210: That would also work, but since we don't actually use the return type, doing it this way means that it accepts functions that do return some value other than `None`.

TeamSpen210:
> Knowing the exact type of platform-dependant Event is not super important, as we give said events back to the same module in due time and never modify it ourselves

It isn't particularly important no, but this way the type checker can confirm we don't modify them.

trio/_core/_run.py
jakkdl: https://typing.readthedocs.io/en/latest/source/best_practices.html#using-any-and-object
> [...] if the return value of a callback is ignored, annotate it with object

trio/_tests/verify_types.json
```diff
@@ -7,11 +7,11 @@
...
   },
   "typeCompleteness": {
-    "completenessScore": 0.9154704944178629,
+    "completenessScore": 0.9378980891719745,
     "exportedSymbolCounts": {
       "withAmbiguousType": 0,
-      "withKnownType": 574,
-      "withUnknownType": 53
+      "withKnownType": 589,
+      "withUnknownType": 39
```
jakkdl, line 14: :rocket: end is in sight!

trio/_core/_run.py
jakkdl: If so, we should leave it for a different PR that changes it in all files though and can figure out any issues.

trio/_core/_run.py
```diff
@@ -169,7 +202,8 @@ def collapse_exception_group(excgroup):
         )
         return exceptions[0]
     elif modified:
-        return excgroup.derive(exceptions)
+        # derive() returns Any for some reason.
+        return excgroup.derive(exceptions)  # type: ignore[no-any-return]
```
jakkdl, line 206: The signature in [argonholm/exceptiongroup](https://github.com/agronholm/exceptiongroup/blob/0878b8331a905f18b4d39b24b105b3e514ada65b/src/exceptiongroup/_exceptions.py#L213) is
```python
    def derive(self: Self, __excs: Sequence[_BaseExceptionT_co]) -> Self:
```
and in [typeshed](https://github.com/python/typeshed/blob/5e2fe131faae347e109b48f451e4b8521378d764/stdlib/builtins.pyi#L2109-L2113) it is
```python
# In reality it is `NonEmptySequence`:
@overload
def derive(self, __excs: Sequence[_ExceptionT]) -> ExceptionGroup[_ExceptionT]: ...
@overload
def derive(self, __excs: Sequence[_BaseExceptionT]) -> BaseExceptionGroup[_BaseExceptionT]: ...

```


I found https://github.com/agronholm/exceptiongroup/pull/40, https://github.com/python/typeshed/issues/9219, and https://github.com/python/typeshed/issues/9230 and reading those it seems that this should've been resolved?

It might be that `disallow_any_generics` would flag something further up if you enabled it

trio/_core/_generated_run.py
```diff
@@ -78,7 +85,8 @@ def current_root_task():
         raise RuntimeError("must be called from async context")
 
 
-def reschedule(task, next_send=_NO_SEND):
+def reschedule(task: trio.lowlevel.Task, next_send: Outcome[Any]=_NO_SEND
```
jakkdl, line 88: [commenting CI check run]
` trio/_core/_generated_run.py:88: error: Cannot determine type of "_NO_SEND"  [has-type]`
This error seems only to pop up when running `mypy -m trio` as opposed to `mypy trio`, so it seems to be a case of circular imports or something. I think we can transition to running the latter in a bit, but for now you can just `type: ignore[misc]` it and add a comment about it.

jakkdl (changes requested):

jakkdl:
> While typing further, found a few more issues.
> 
>     * `sortedcontainers` doesn't have types, and it seems is unlikely to ever have typing - it does a lot of unusual things in the constructors apparently.
> 
>     * Looks like we need either my `outcome` PR or `trio_typing` in CI to have types for `outcome`, the errors are fine with that present.

We could maybe generate stubs for relevant calls to `sortedcontainers`.
Sounds fine to wait for the `Outcome` PR to get merged and released, which hopefully shouldn't be too far off. Can disable `no_any_unimported` in the meantime.

docs/source/reference-core.rst
```diff
@@ -916,15 +916,21 @@ The nursery API
...
+
+   .. automethod:: start(async_fn, *args, name = None)
+
+   .. automethod:: start_soon(async_fn, *args, name = None)
 
 .. attribute:: TASK_STATUS_IGNORED
+   :type: TaskStatus
 
-   See :meth:`~Nursery.start`.
+   See :meth:`Nursery.start`.
```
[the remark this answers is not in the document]
TeamSpen210: What `-` does is just make this display as `start()`, instead of showing the full `Nursery.start()` text.

trio/_core/_run.py
TeamSpen210: You have to pass the undocumented `--enable-incomplete-feature=TypeVarTuple` + `=Unpack` to enable the incomplete code. If not, you get a warning raised for the code, `Unpack` becomes `Any`. So maybe we should indeed use them, if we can suppress the error.

docs/source/conf.py
```diff
@@ -62,6 +62,10 @@
     ("py:obj", "trio._abc.SendType"),
     ("py:obj", "trio._abc.T"),
     ("py:obj", "trio._abc.T_resource"),
+    ("py:class", "trio._core._run.StatusT"),
+    ("py:class", "trio._core._run.StatusT_co"),
+    ("py:class", "trio._core._run.StatusT_contra"),
+    ("py:class", "trio._core._run.RetT"),
```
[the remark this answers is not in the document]
TeamSpen210: Ideally we'd figure out the issue in Sphinx that produces these spurious errors, or do an extension that fixes it for us at least. It's possible to write a hook that intercepts all missing names, then does custom things. In one of my own projects I made it just redirect typevars to link to the `typing.TypeVar` docs. Couldn't quite get that working for Trio though...

docs/source/reference-lowlevel.rst
```diff
@@ -49,7 +49,11 @@ attributes, :meth:`trio.Lock.statistics`, etc.). Here are some more.
 Global statistics
 -----------------
 
-.. autofunction:: current_statistics
+.. function:: current_statistics() -> RunStatistics
+
+   Returns an object containing run-loop-level debugging information:
+
+.. autoclass:: trio.RunStatistics()
```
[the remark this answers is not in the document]
TeamSpen210: I manually added the docs here to make it so that the list of attributes wasn't repeated - since the class right below is going to show those. But for the docstring, it makes sense to repeat them since you're probably only looking at the function.

docs/source/reference-core.rst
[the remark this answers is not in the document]
jakkdl: thanks :+1:

TeamSpen210:
Why lowlevel? It's very much a public interface, it's necessary for `Nursery.start()` which is fairly important.

oremanj:
Strong agree, TaskStatus should not be in lowlevel. (Maybe you think it goes with Task? They have nothing to do with each other, despite the name similarity.)

trio/_core/_generated_run.py
TeamSpen210: This should just be because `Outcome` is untyped right now.

pyproject.toml
```diff
@@ -74,6 +75,7 @@ disallow_subclassing_any = true
 # Needs to use Any due to some complex introspection.
 module = [
     "trio._path",
+    "trio._core._run",
 ]
```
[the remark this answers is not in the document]
TeamSpen210: The problem here is mainly `disallow_any_decorated` - normally that would be fine, but it means if you have a decorator on a function that's deliberately annotated with `Any`, it gives a warning.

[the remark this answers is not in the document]
jakkdl: sure. So either disable it for all files, or add `type: ignore` where it's unavoidable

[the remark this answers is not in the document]
jakkdl: ah, that's what you did. :+1:

trio/_core/_generated_run.py
jakkdl: Huh, weird how it gets fixed by either.

trio/_core/_run.py
jakkdl: Seems like it didn't. @Zac-HD as the resident expert on ExceptionGroups, would you mind weighing in what's going on here? Not a blocker for merging the PR though

jakkdl (approved):
I vote for disabling `no_any_unimported` so the PR can be merged, and then when Outcome PR is released+merged it can be toggled in another PR that also generates stubs for `sortedcontainers`.

trio/_core/_run.py
```diff
@@ -1356,13 +1446,41 @@ class RunContext(threading.local):
...
-@attr.s(frozen=True)
-class _RunStatistics:
-    tasks_living = attr.ib()
-    tasks_runnable = attr.ib()
-    seconds_to_next_deadline = attr.ib()
-    io_statistics = attr.ib()
-    run_sync_soon_queue_size = attr.ib()
+class _IOStats(Protocol):
+    @property
+    def backend(self) -> str:
```
jakkdl, line 1451: change this to be a `Literal` as in #2743 ?

jakkdl (changes requested):

trio/_core/_run.py
Zac-HD: Looks like the `exceptiongroup` package hasn't done the equivalent of https://github.com/python/typeshed/issues/9219, which is my bad for not following up on https://github.com/agronholm/exceptiongroup/pull/40#discussion_r1024633541.

I think it's fine to ignore this for now, but if you wanted to open an upstream issue and revisit later that'd be great.

trio/_core/_run.py
TeamSpen210: Should actually just be able to eliminate this entirely, specify the concrete type depending on the platform.

jakkdl (approved):

trio/_tools/gen_exports.py
```diff
@@ -24,10 +24,17 @@
 HEADER = """# ***********************************************************
 # ******* WARNING: AUTOGENERATED! ALL EDITS WILL BE LOST ******
 # *************************************************************
-# isort: skip
+# isort: skip_file
+from __future__ import annotations
+
+from typing import Any, Callable, Awaitable
```
CoolCat467, line 30: `from collections.abc import Awaitable, Callable`, see other

CoolCat467 (changes requested):
Looks great other than noted changes

trio/_core/_generated_run.py
```diff
@@ -1,18 +1,25 @@
 # ***********************************************************
 # ******* WARNING: AUTOGENERATED! ALL EDITS WILL BE LOST ******
 # *************************************************************
-# isort: skip
+# isort: skip_file
+from __future__ import annotations
+
+from typing import Any, Callable, Awaitable
```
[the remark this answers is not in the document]
jakkdl: there's no real reason not to use them either, once we drop py39 (?) support and run `pyupgrade` they'll all get replaced. I'm weakly in favor of using the non-aliases, but don't think it's worth delaying PR's over it.

.coveragerc
```diff
@@ -23,6 +23,7 @@ exclude_lines =
   if _t.TYPE_CHECKING:
   if t.TYPE_CHECKING:
   @overload
+  class .*\bProtocol\b.*\):
```
A5rocks, line 26: Huh, why add this (I really really don't want to read through all the conversation on this PR and the blame doesn't help :sob:)

trio/_core/_io_epoll.py
```diff
@@ -12,12 +14,16 @@
...
+
 
 @attr.s(slots=True, eq=False, frozen=True)
 class _EpollStatistics:
-    tasks_waiting_read = attr.ib()
-    tasks_waiting_write = attr.ib()
-    backend = attr.ib(default="epoll")
+    tasks_waiting_read: int = attr.ib()
+    tasks_waiting_write: int = attr.ib()
+    backend: Literal["epoll"] = attr.ib(default="epoll")
```
A5rocks, line 26: I really feel like this should have `init=False`

trio/_tools/gen_exports.py
```diff
@@ -24,10 +24,19 @@
 HEADER = """# ***********************************************************
 # ******* WARNING: AUTOGENERATED! ALL EDITS WILL BE LOST ******
 # *************************************************************
-# isort: skip
+# isort: skip_file
```
A5rocks, line 27: I really think we could put this prelude through `isort` at least.

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -6,7 +6,7 @@
 from math import inf
 
 import trio
-from trio.lowlevel import TaskStatus
+from trio import TaskStatus
```
A5rocks, line 9: Note to self: this isn't actually a breaking change (I got concerned for a bit, oops).

trio/_core/_run.py
A5rocks: Made an upstream issue! See https://github.com/agronholm/exceptiongroup/issues/76

trio/_core/_run.py
```diff
@@ -780,28 +820,57 @@ def cancel_called(self) -> bool:
...
-        if self._called_started:
+    @overload
+    def started(self: _TaskStatus[None]) -> None:
+        ...
+
+    @overload
+    def started(self: _TaskStatus[StatusT], value: StatusT) -> None:
+        ...
+
+    def started(self, value: StatusT | None = None) -> None:
```
A5rocks, line 864: Technically the `| None` is unnecessary. I don't know if mypy supports that though (I remember seeing a still-open issue that looked roughly like this.)

trio/_core/_run.py
```diff
@@ -1022,8 +1103,9 @@ async def _nested_child_finished(self, nested_child_exc):
             # If we get cancelled (or have an exception injected, like
             # KeyboardInterrupt), then save that, but still wait until our
             # children finish.
-            def aborted(raise_cancel):
-                self._add_exc(capture(raise_cancel).error)
+            def aborted(raise_cancel: _core.RaiseCancelT) -> Abort:
+                # capture() needs an overload for NoReturn -> Error.
```
A5rocks, line 1107: You did this, the comment isn't necessary anymore.

trio/_core/_run.py
```diff
@@ -1048,8 +1130,15 @@ def aborted(raise_cancel):
...
                 # (see test_nursery_cancel_doesnt_create_cyclic_garbage)
                 del self._pending_excs
+        return None
 
-    def start_soon(self, async_fn, *args, name=None):
+    def start_soon(
+        self,
+        # TODO: TypeVarTuple
+        async_fn: Callable[..., Awaitable[object]],
+        *args: Any,
```
A5rocks, line 1139: I feel like `object` is fine here (unlike in the protocol, where I don't dare even think about variance xdd)

trio/_core/_run.py
```diff
@@ -845,6 +914,15 @@ def started(self, value=None):
         self._old_nursery._check_nursery_closed()
 
 
+class _NurseryStartFunc(Protocol[StatusT_co]):
+    """Type of functions passed to `nursery.start() <trio.Nursery.start>`."""
+
+    def __call__(
+        self, *args: Any, task_status: TaskStatus[StatusT_co]
```
A5rocks, line 921: Have you made sure the protocol matches a couple representative functions? I'm a bit scared the `*args` might require all callables passed to have `*args` of their own.

trio/_core/_run.py
```diff
@@ -1091,7 +1180,9 @@ def start_soon(self, async_fn, *args, name=None):
         """
         GLOBAL_RUN_CONTEXT.runner.spawn_impl(async_fn, args, self, name)
 
-    async def start(self, async_fn, *args, name=None):
+    async def start(
+        self, async_fn: _NurseryStartFunc[StatusT], *args: Any, name: object = None
```
A5rocks, line 1184: I'll guess `object` is fine here too?

trio/_core/_run.py
```diff
@@ -1643,10 +1768,10 @@ async def python_wrapper(orig_coro):
             self.instruments.call("task_spawned", task)
         # Special case: normally next_send should be an Outcome, but for the
         # very first send we have to send a literal unboxed None.
-        self.reschedule(task, None)
+        self.reschedule(task, None)  # type: ignore[arg-type]
         return task
 
-    def task_exited(self, task, outcome):
+    def task_exited(self, task: Task, outcome: Outcome[Any]) -> None:
```
A5rocks, line 1774: I .... think the `Any` might possibly be `object`-ified.

trio/_core/_run.py
```diff
@@ -1756,7 +1889,12 @@ def spawn_system_task(self, async_fn, *args, name=None, context=None):
             context=context,
         )
 
-    async def init(self, async_fn, args):
+    async def init(
+        # TODO: TypeVarTuple
+        self,
+        async_fn: Callable[..., Awaitable[object]],
+        args: tuple[Any, ...],
```
A5rocks, line 1896: `object` might work here?

trio/_core/_run.py
```diff
@@ -2018,13 +2158,13 @@ def setup_runner(
 
 
 def run(
-    async_fn,
-    *args,
-    clock=None,
-    instruments=(),
+    async_fn: Callable[..., RetT],
+    *args: Any,
```
A5rocks, line 2162: `object` again

trio/_core/_run.py
```diff
@@ -2117,23 +2258,26 @@ def run(
...
 
 
 def start_guest_run(
-    async_fn,
-    *args,
-    run_sync_soon_threadsafe,
-    done_callback,
-    run_sync_soon_not_threadsafe=None,
+    async_fn: Callable[..., Awaitable[RetT]],
+    *args: Any,
```
A5rocks, line 2270: `object` maybe

trio/_core/_run.py
```diff
@@ -2221,10 +2365,10 @@ def my_done_callback(run_outcome):
 # straight through.
 def unrolled_run(
     runner: Runner,
-    async_fn,
-    args,
+    async_fn: Callable[..., object],
+    args: tuple[Any, ...],
```
A5rocks, line 2369: `object` maybe

A5rocks (commented):
comment

.coveragerc
A5rocks: Nevermind I figured it out.

trio/_core/_run.py
TeamSpen210: Unfortunately no, `Outcome` can't be invariant because it uses them in both `unwrap()`'s return value and `send()`'s parameter.

trio/_core/_run.py
TeamSpen210: It might actually. So then even `TypeVarTuple` won't help, we'll need a heap of protocols or something.

.coveragerc
TeamSpen210: Yeah otherwise coverage complains that the method stubs weren’t run. Though maybe we should rethink this, if we add default implementations to any protocols in the future it won’t flag those.

trio/_core/_io_epoll.py
TeamSpen210: Actually did that originally, but didn’t want to do a functional change.

trio/_core/_run.py
CoolCat467: It is absolutely necessary in new versions of mypy. That is a legacy feature that I think they are trying very hard to depreciate.

A5rocks: Nah I'm not talking about strict optional: I'm saying that that matches the overloads (cause StatusT would have to be None in order for that default to work)

trio/_core/_run.py
A5rocks: Nah I've tried out TypeVarTuple, the difference here is that *args: object is an infinite amount vs *args: Unpack[Ts] is a finite (probably) amount)

trio/_core/_run.py
TeamSpen210: Unfortunately no, Mypy still complains.

trio/_tools/gen_exports.py
TeamSpen210: Perhaps what we should do is just make `gen_exports` run `black` + `isort` on the files it produces.

trio/_core/_run.py
```diff
@@ -10,21 +10,40 @@
...
+    final,
+    overload,
+)
 
 import attr
 from outcome import Error, Outcome, Value, capture
 from sniffio import current_async_library_cvar
 from sortedcontainers import SortedDict
 
+import trio._core._ki
```
oremanj, line 43: What is this import for? I don't see any change that would necessitate it; we already import the names we need from `_ki` a few lines down.

trio/_core/_run.py
```diff
@@ -1524,12 +1638,12 @@ def current_time(self):
         return self.clock.current_time()
 
     @_public
-    def current_clock(self):
+    def current_clock(self) -> trio._abc.Clock:
```
oremanj, line 1641: maybe `trio.abc.Clock` for a less underscore-y public-facing signature?

trio/_core/_run.py
```diff
@@ -2533,6 +2678,7 @@ async def checkpoint() -> None:
     await cancel_shielded_checkpoint()
     task = current_task()
     task._cancel_points += 1
+    assert task._cancel_status is not None, task
```
oremanj, line 2681: `checkpoint` and `checkpoint_if_cancelled` are very hot perf-wise. I think it might be better to use a type-ignore than an assertion here. In practice the only way to get `Task._cancel_status is None` is in the init task outside the outermost nursery, and there's no user code that runs there. Typing `Task` as if `_cancel_status` can never be None is also reasonable IMO.

oremanj (commented):
This generally looks great; thank you for working on these!

trio/_core/_run.py
TeamSpen210: Yeah it might make more sense to make it non-optional, then type-ignore the places where it is `None`.

trio/_core/_run.py
TeamSpen210: Good question, seems pointless.

oremanj:
Thanks! I have no further comments but will leave final approval to @A5rocks since they've been more involved in the process thus far.

trio/_core/_io_epoll.py
```diff
@@ -230,14 +235,14 @@ def force_wakeup(self) -> None:
     # Return value must be False-y IFF the timeout expired, NOT if any I/O
     # happened or force_wakeup was called. Otherwise it can be anything; gets
     # passed straight through to process_events.
-    def get_events(self, timeout: float) -> list[tuple[int, int]]:
+    def get_events(self, timeout) -> EventResult:
```
CoolCat467, line 238: Why is this and `process_events` missing type annotations? Is it WIP?

CoolCat467 (changes requested):
I am noticing a lot of missing annotations, in particular `_io_` files

trio/_core/_io_epoll.py
TeamSpen210: Yep, this PR isn't about typing those IO modules. I just did a few specifiy ones, because they're imported from `_run`.

jakkdl: They were handled in #2743

jakkdl: Oh, in the merge you accidentally wiped some types that were added in it though

CoolCat467 (approved):
Looks good other than a few missing annotations on at least the windows one, but if not here there can always be more PRs. Also of course making sure the CI passes

jakkdl:
> Looks good other than a few missing annotations on at least the windows one, but if not here there can always be more PRs. Also of course making sure the CI passes

_io_windows is listed as a TODO in https://github.com/python-trio/trio/issues/2734

PR #2735, opened 2023-08-04
Add typing for `_core/_asyncgens.py`

CoolCat467:
This PR adds type annotations to `_core/_asyncgens.py`


CoolCat467:
I do have one issue currently, mypy says `trio/_core/_asyncgens.py:111: error: "Awaitable[None]" has no attribute "send"  [attr-defined]`

trio/_core/_asyncgens.py
```diff
@@ -1,14 +1,21 @@
...
+from typing import TYPE_CHECKING, Any
 
 import attr
 
 from .. import _core
 from .._util import name_asyncgen
 from . import _run
 
+if TYPE_CHECKING:
+    from ._run import Runner
```
TeamSpen210, line 17: We've imported `_run` already above, so you could just skip this import and use `_run.Runner` below.

trio/_core/_asyncgens.py
```diff
@@ -22,17 +29,17 @@ class AsyncGenerators:
     # asyncgens after the system nursery has been closed, it's a
     # regular set so we don't have to deal with GC firing at
     # unexpected times.
-    alive = attr.ib(factory=weakref.WeakSet)
+    alive: weakref.WeakSet[AsyncGenerator[Any, Any]] | set[AsyncGenerator[Any, Any]] = attr.ib(factory=weakref.WeakSet)  # type: ignore[assignment]
```
TeamSpen210, line 32: Instead of a union, this could just be `collections.abc.MutableSet`, which both sets inherit from.

trio/_core/_asyncgens.py
```diff
@@ -22,17 +29,17 @@ class AsyncGenerators:
...
     # init task starting end-of-run asyncgen finalization.
-    trailing_needs_finalize = attr.ib(factory=set)
+    trailing_needs_finalize: set[AsyncGenerator[Any, Any]] = attr.ib(factory=set)
 
     prev_hooks = attr.ib(init=False)
 
-    def install_hooks(self, runner):
-        def firstiter(agen):
+    def install_hooks(self, runner: Runner) -> None:
+        def firstiter(agen: AsyncGenerator[Any, Any]) -> None:
```
TeamSpen210, line 42: Instead of `Any`, this could use `AsyncGenerator[object, NoReturn]` - that should accept any generator, but it'll keep it type-safe by preventing code here from actually iterating it and expecting a specific value. Same applies for the rest of the file of course.

trio/_core/_asyncgens.py
```diff
@@ -174,10 +183,12 @@ async def finalize_remaining(self, runner):
             for agen in batch:
                 await self._finalize_one(agen, name_asyncgen(agen))
 
-    def close(self):
+    def close(self) -> None:
         sys.set_asyncgen_hooks(*self.prev_hooks)
 
-    async def _finalize_one(self, agen, name):
+    async def _finalize_one(
+        self, agen: AsyncGenerator[Any, Any], name: str | object
```
TeamSpen210, line 190: `name` here could just be `object`, the `str` is redundant. It really doesn't have much to do with strings, the logging call below just does `repr()`.

trio/_core/_asyncgens.py
```diff
@@ -61,7 +70,7 @@ def finalize_in_trio_context(agen, agen_name):
                 # have hit it.
                 self.trailing_needs_finalize.add(agen)
 
-        def finalizer(agen):
+        def finalizer(agen: AsyncGenerator[Any, Any]) -> None:
```
TeamSpen210, line 73: To fix the `closer.send(None)` error, change this to `types.AsyncGeneratorType` (and probably revert the same change in `_util.name_asyncgen`. The subtle difference between them is that [`collections.abc.AsyncGenerator`](https://github.com/python/typeshed/blob/75c71f324ec564f2c6eaeee81f6b9618db5fb98c/stdlib/typing.pyi#L477) defines the methods as just returning an `Awaitable`, while [`types.AsyncGeneratorType`](https://github.com/python/typeshed/blob/75c71f324ec564f2c6eaeee81f6b9618db5fb98c/stdlib/types.pyi#L379C17-L379C17) corresponds specifically to the type produced by `async def` functions and defines the methods as returning coroutines. 

This finalizer function is called by the interpreter, we should be safe in assuming that they're Python-defined generators and not some other class. Though, the stubs for `set_asyncgen_hooks()` use the regular ABC type...

CoolCat467:
Now we have these errors with those changes:
```console
trio/_core/_asyncgens.py:30: error: Incompatible types in assignment (expression has type "WeakSet[_T]", variable has type "MutableSet[AsyncGeneratorType[object, NoReturn]]")  [assignment]
trio/_core/_asyncgens.py:37: error: Incompatible types in assignment (expression has type "Set[_T]", variable has type "Set[AsyncGeneratorType[object, NoReturn]]")  [assignment]
trio/_core/_asyncgens.py:184: error: Incompatible types in assignment (expression has type "Set[<nothing>]", variable has type "MutableSet[AsyncGeneratorType[object, NoReturn]]")  [assignment]
```

trio/_core/_asyncgens.py
CoolCat467: Changed in 0205c8f

trio/_core/_asyncgens.py
CoolCat467: Changed in 0205c8f

trio/_core/_asyncgens.py
CoolCat467: Changed in 0205c8f

trio/_core/_asyncgens.py
CoolCat467: Changed in 0205c8f

trio/_core/_asyncgens.py
CoolCat467: Changed in 0205c8f

trio/_core/_asyncgens.py
CoolCat467: Reverted in 2e255eb

trio/_core/_asyncgens.py
```diff
@@ -22,17 +39,21 @@ class AsyncGenerators:
     # asyncgens after the system nursery has been closed, it's a
     # regular set so we don't have to deal with GC firing at
     # unexpected times.
-    alive = attr.ib(factory=weakref.WeakSet)
+    alive: weakref.WeakSet[AsyncGeneratorType[object, NoReturn]] | set[
```
TeamSpen210, line 42: Since you made the type aliases above, perhaps use those here - would let this fit on one line.

trio/_util.py
```diff
@@ -370,8 +370,5 @@ def name_asyncgen(agen: AsyncGeneratorType[object, t.NoReturn]) -> str:
         module = agen.ag_frame.f_globals["__name__"]
     except (AttributeError, KeyError):
         module = f"<{agen.ag_code.co_filename}>"
-    try:
-        qualname = agen.__qualname__
-    except AttributeError:
-        qualname = agen.ag_code.co_name
+    qualname = getattr(agen, "__qualname__", agen.ag_code.co_name)
```
TeamSpen210, line 373: Not sure, but perhaps it might be better to leave this as a try-except, since then `agen.ag_code.co_name` doesn't have to be evaluated in most cases.

TeamSpen210 (approved):
Looks good overall.

trio/_core/_asyncgens.py
CoolCat467: They are not aliases, they are purely for being able to set the factory type so there are not type errors

TeamSpen210: Well they function as type aliases, so you could use them as such.

trio/_util.py
CoolCat467: Reverted with 82900e5

trio/_core/_asyncgens.py
```diff
@@ -112,9 +135,9 @@ def finalizer(agen):
                         )
 
         self.prev_hooks = sys.get_asyncgen_hooks()
-        sys.set_asyncgen_hooks(firstiter=firstiter, finalizer=finalizer)
+        sys.set_asyncgen_hooks(firstiter=firstiter, finalizer=finalizer)  # type: ignore[arg-type]  # Finalizer doesn't use AsyncGeneratorType
```
A5rocks, line 138: Now that I look at the comment history I see why we use `AsyncGeneratorType` (that was my question).

However, if this is indeed guaranteed to be correct (I still don't understand but I haven't really looked too closely at docs), then can someone raise an error on typeshed too?

trio/_core/_asyncgens.py
```diff
@@ -22,17 +39,21 @@ class AsyncGenerators:
...
+        AsyncGeneratorType[object, NoReturn]
+    ] = attr.ib(factory=_WEAK_ASYNC_GEN_SET)
 
     # This collects async generators that get garbage collected during
     # the one-tick window between the system nursery closing and the
     # init task starting end-of-run asyncgen finalization.
-    trailing_needs_finalize = attr.ib(factory=set)
+    trailing_needs_finalize: set[AsyncGeneratorType[object, NoReturn]] = attr.ib(
+        factory=_ASYNC_GEN_SET
+    )
```
A5rocks, line 51: I swear I remember inference working here. Does it not? Let's make a mypy issue about this.

trio/_core/_asyncgens.py
```diff
@@ -170,14 +193,16 @@ async def finalize_remaining(self, runner):
         # all are gone.
         while self.alive:
             batch = self.alive
-            self.alive = set()
+            self.alive = _ASYNC_GEN_SET()
```
A5rocks, line 196: I would rather use inference (`self.alive: ... = set()`) then call a thing we could change and never get told about errors about (cause it's `if not TYPE_CHECKING`).

A5rocks (approved):
Small followups but looks good. I think we could probably get rid of the nasty type-check-only time type aliases in a couple months after tools get fixed.

trio/_core/_asyncgens.py
TeamSpen210: Well any library could manually do `sys.get_asyncgen_hooks()` then call it with whatever, but that’s rather esoteric and probably their fault if they pass something wrong. In the standard library, it’s only called deep in the async generator code, when they’re being destroyed. 

We could check, but I think it might actually be better not to - the only type that might make sense to pass here would be something like Cython, which would probably duck-type well enough.

trio/_core/_asyncgens.py
CoolCat467: Yea just tested removing the annotation and we get this error:
```console
error: Need type annotation for "trailing_needs_finalize"  [var-annotated]
```

trio/_core/_asyncgens.py
CoolCat467: Just tried replacing with 
```python
self.alive: _ASYNC_GEN_SET = set()
```
and we get this error:
```console
error: Incompatible types in assignment (expression has type "set[<nothing>]", variable has type "WeakSet[AsyncGeneratorType[object, NoReturn]] | set[AsyncGeneratorType[object, NoReturn]]")  [assignment]
```

jakkdl (changes requested):
Up the strictness in `pyproject.toml` for the file

CoolCat467:
> Up the strictness in `pyproject.toml` for the file

Changed in cd19e84

pyproject.toml
```diff
@@ -50,6 +50,7 @@ disallow_untyped_defs = false
 # Fully typed, enable stricter checks
 module = [
     "trio._abc",
+    "trio._core._asyncgens.py",
```
TeamSpen210, line 53: The `.py` probably shouldn't be here, same with `_highlevel_open_tcp_stream`.

trio/_core/_asyncgens.py
```diff
@@ -12,6 +16,19 @@
 # Used to log exceptions in async generator finalizers
 ASYNCGEN_LOGGER = logging.getLogger("trio.async_generator_errors")
 
+if TYPE_CHECKING:
+    # TODO: Remove when dropping 3.8 support, __class_getitem__ added in 3.9
+    _WEAK_ASYNC_GEN_SET = weakref.WeakSet[AsyncGeneratorType[object, NoReturn]]
+    if sys.version_info < (3, 10):
```
TeamSpen210, line 22: Since we're in a `TYPE_CHECKING` block, there isn't really any point checking the version - just use the builtin, type checkers know what it means. Mypy should suppress the version error, I think. If not just unconditionally import from typing.

trio/_core/_asyncgens.py
A5rocks: Wait I don't understand the error. Mypy seems to think you're doing `self.alive`... oh I missed that it was `self`. Ouch.

trio/_core/_asyncgens.py
A5rocks: Hm. Okay.

pyproject.toml
CoolCat467: Fixed in 07b26fa

trio/_core/_asyncgens.py
CoolCat467: Changed in 07b26fa

trio/_core/_asyncgens.py
jakkdl: +1

trio/_core/_asyncgens.py
jakkdl: IIRC you can add a `type=...` parameter to `attr.ib`. It might also be that we're using somewhat outdated `attr` constructions

jakkdl (approved):
Nothing to add aside from what's already up for discussion

trio/_core/_asyncgens.py
CoolCat467: Changed in d2c5f3c

jakkdl:
:rocket:

trio/_core/_asyncgens.py
```diff
@@ -12,6 +16,15 @@
 # Used to log exceptions in async generator finalizers
 ASYNCGEN_LOGGER = logging.getLogger("trio.async_generator_errors")
 
+if TYPE_CHECKING:
```
oremanj, line 19: I don't understand why this branch is needed. Since you have `from __future__ import annotations`, I would expect that it works fine to use the fully-specified types in `x: Type` contexts and the base types in `x = Type()` contexts.

trio/_core/_asyncgens.py
```diff
@@ -22,17 +35,17 @@ class AsyncGenerators:
...
     # init task starting end-of-run asyncgen finalization.
-    trailing_needs_finalize = attr.ib(factory=set)
+    trailing_needs_finalize: _ASYNC_GEN_SET = attr.ib(factory=_ASYNC_GEN_SET)
 
     prev_hooks = attr.ib(init=False)
 
-    def install_hooks(self, runner):
-        def firstiter(agen):
+    def install_hooks(self, runner: _run.Runner) -> None:
+        def firstiter(agen: AsyncGeneratorType[object, NoReturn]) -> None:
```
oremanj, line 48: firstiter and finalizer in typeshed are typed as taking `collections.abc.AsyncGenerator[Any, Any]`. I think it would be better for us to use that name and avoid the need for a type-ignore. Using Any is not evil.

trio/_core/_asyncgens.py
oremanj: never mind, I see this was already discussed

PR #2737, opened 2023-08-05
Add slots to exceptions

CoolCat467:
This PR adds slots to all the exceptions defined in `_core/_exceptions.py`

### DIFF
--- a/trio/_core/_exceptions.py
+++ b/trio/_core/_exceptions.py
@@ -14,6 +14,8 @@ class TrioInternalError(Exception):
 
     """
 
+    __slots__ = ()
+
 
 class RunFinishedError(RuntimeError):
     """Raised by `trio.from_thread.run` and similar functions if the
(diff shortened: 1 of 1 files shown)


jakkdl:
Is there a stronger case for this than #2723?

CoolCat467:
In this one, nothing should be weak referencing exceptions

jakkdl:
but is there much/anything to gain? There's a small cost in code readability and there's no CI check that makes usage consistent or anything

CoolCat467:
I feel like it's best practice to use slots whenever possible, and more generally (not applicable here as much though) having slots on your classes means if people want to subclass them they can also take advantage of lower memory usage per object if they so choose. Not having slots eliminates that option for people down the road.

A5rocks:
I'm *very* ambivalent about this one. It's probably fine! But I don't feel very strongly about specifically adding `__slots__` to exceptions either way. Maybe someone else with strong(ish) opinions one way or another can figure this out...?

oremanj:
A BaseException already has a `__dict__`, so the *only* effect of adding `__slots__ = ()` is to disable weak references. Which only saves eight bytes per exception object, and could plausibly break something even though I agree it's unlikely.

I don't think this is worth the trouble.

TeamSpen210:
Also exception objects aren't exactly created / kept around long, so it's not really that big a deal how big they are.

Issue #2734, opened 2023-08-03
Remaining tasks until --verifytypes passes and `check_type_completeness.py` can be removed


jakkdl:
To get a bit of an overview of the remaining files until we hit Type Completeness :tm:

## Public API
* TypeVarTuple support - needs to be checked with pyright instead of mypy that they're correct, but mypy should be able to accept their presence.
  * not really possible to get mypy to accept their presence in my testing, we need to wait for mypy 1.7. Tracked in #2859
* [x] SocketType, see #2720 - @jakkdl 
* _core/
  * [x] _mock_clock #2740
  * [x] _run #2733 
    * [x] _generated_run #2733
  * [x] _io_epoll #2743
    * [x] _generated_io_epoll #2743
  * [x] _io_kqueue #2743
    * [x] _generated_io_kqueue #2743
  * [x] _io_windows - some work in #2763, but I would like somebody else to take over and finish.
    * [x] _generated_io_windows
  * [x] _thread_cache #2741
  * [x] _traps #2763
* [x] testing/ #2747
* _highlevel
  * [x] _socket
  * [x] _ssl_helpers
  * [x] _serve_listeners
  * [x] _open_tcp_listeners #2724
  * [x] _open_unix_stream
* [x] _ssl #2745
* [x] _subprocess
* [x] _threads
* [x] _signals #2794 (`_open_signal_receiver` is clearly exported in `trio/__init__.py`, but for some reason doesn't show up in the file list when I run `--verifytypes`)

~~Possibly more files if running with a --pythonplatform other than linux.~~

## Private or deprecated (lower priority)
May include files that should be in the above list, I didn't look through files in a fully complete way
* _core/
  * [x] _asyncgens #2735 
  * [x] _multierror #2742
  * [x] _io_common #2763
  * [x] _windows_cffi #2755
  * [x] _ki #2763 
  * [x] _wakeup_socketpair #2763
* _tools/
  * [x] _gen_exports.py #2738
* [x] _windows_pipes #2812
* [x] _wait_for_object (windows) #2755 
* testing/
  * [x] _fake_net #2763 (This file should maybe be moved to _tests/ as it's only used internally)
* [x] _subprocess_platform/waitid
* [x] various files 

## Tests
Typing the tests will check that it's possible to use the interface in a reasonable way, as well as catch any possible bugs in the tests themself of course.
* [x] trio/testing/_fake_net
* [x] trio/_core/_tests/
* [x] trio/_tests/

## docstrings
[...]

PR #2721, opened 2023-07-29
[WIP] combined typing branch

jakkdl:
I've had this branch as a development branch locally to check compatibility between my typing PRs, and be able to quickly write mostly-fine types without waiting for reviews. But with @teamspen210 also working on types, I figured I should push this so we don't duplicate work. But feel free to grab stuff from it!

Don't bother reviewing this one, at least not in detail or in full. It'll likely be full of merge conflicts and failing tests and the like as well. I'm pulling out one or a few files at a time, cleaning them up, and opening separate PRs.

PR #2738, opened 2023-08-05
Add typing to `_tools/gen_exports.py`

CoolCat467:
This PR adds type annotations to `_tools/gen_exports.py`

### DIFF
--- a/trio/_tools/gen_exports.py
+++ b/trio/_tools/gen_exports.py
@@ -9,7 +9,7 @@ import argparse
 import ast
 import os
 import sys
-from collections.abc import Generator, Iterable
+from collections.abc import Iterable, Iterator
 from pathlib import Path
 from textwrap import indent
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


TeamSpen210 (approved):

trio/_tools/gen_exports.py
```diff
@@ -36,7 +43,7 @@
 """
 
 
-def is_function(node):
+def is_function(node: ast.AST) -> TypeGuard[ast.FunctionDef | ast.AsyncFunctionDef]:
```
jakkdl, line 46: style: Since there's a bunch of `ast.FunctionDef | ast.AsyncFunctionDef` I created a TypeAlias for it, and while it does make the signatures a bit easier to parse I'm not even sure myself it's worth it.

trio/_tools/gen_exports.py
```diff
@@ -45,17 +52,18 @@ def is_function(node):
...
+        for decorator in node.decorator_list:
+            if isinstance(decorator, ast.Name) and decorator.id == "_public":
+                return True
     return False
 
 
-def get_public_methods(tree):
+def get_public_methods(
+    tree: ast.AST,
+) -> Generator[ast.FunctionDef | ast.AsyncFunctionDef, None, None]:
```
jakkdl, line 66: nit: `Iterable` instead of `Generator` means you don't have to specify `None, None`

jakkdl (approved):
Ah, I didn't cover this one in #2734 since it's not relevant to `--verifytypes`, but I'd quickly typed it up too in #2721. I'll make a push to open up a bunch of PRs soon.

Only a few small nits compared to my implementation.

jakkdl:
Note that this file currently isn't typechecked at all in CI, since we only run `mypy -m trio -m trio.testing` and this file isn't imported by either `__init__.py`

PR #2739, opened 2023-08-06
Add a few missing type annotations to `_core/_run.py`

CoolCat467:
This PR adds a few missing type annotations to `_core/_run.py`

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -10,7 +10,14 @@ import sys
 import threading
 import warnings
 from collections import deque
-from collections.abc import Callable, Coroutine, Iterator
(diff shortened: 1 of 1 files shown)


CoolCat467:
Oh rats, there is already work on this in #2733, I didn't see it because it wasn't listed in #2734

A5rocks:
Oh oops, thanks for updating that

jakkdl:
> Oh rats, there is already work on this in #2733, I didn't see it because it wasn't listed in #2734

it is? I guess you read `[_generated]_run` as only including `_generated_run` and not both `_run` and `_generated_run`

PR #2740, opened 2023-08-06
Add missing type annotations to `_core/_mock_clock.py`

CoolCat467:
This PR adds missing type annotations to `_core/_mock_clock.py`.

### DIFF
--- a/trio/_core/_mock_clock.py
+++ b/trio/_core/_mock_clock.py
@@ -62,7 +62,7 @@ class MockClock(Clock, metaclass=Final):
 
     """
 
-    def __init__(self, rate: float = 0.0, autojump_threshold: float = inf):
+    def __init__(self, rate: float = 0.0, autojump_threshold: float = inf) -> None:
         # when the real clock said 'real_base', the virtual time was
         # 'virtual_base', and since then it's advanced at 'rate' virtual
         # seconds per real second.
@@ -150,7 +150,7 @@ class MockClock(Clock, metaclass=Final):
         else:
             return 999999999
 
-    def jump(self, seconds) -> None:
+    def jump(self, seconds: float) -> None:
         """Manually advance the clock by the given number of seconds.
 
         Args:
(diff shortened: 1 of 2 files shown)


TeamSpen210 (approved):

trio/_core/_mock_clock.py
```diff
@@ -62,7 +62,7 @@ class MockClock(Clock, metaclass=Final):
 
     """
 
-    def __init__(self, rate: float = 0.0, autojump_threshold: float = inf):
+    def __init__(self, rate: float = 0.0, autojump_threshold: float = inf) -> None:
```
A5rocks, line 65: Err, are we doing this or not...? Maybe put this in your bigger PR about these and punt decisions to that central place?

A5rocks (commented):
Just one small comment

trio/_core/_mock_clock.py
CoolCat467: With the changes I made with `verify_types.json`, it said that this PR adds 2 new typed functions, so I think this change will be important for 100% type completion for the entire project

A5rocks: Oh does pyright require this? Could you check please? If so, I'm totally on board with adding `-> None` everywhere: that adds software to ensure we are consistently putting the `-> None` everywhere.

CoolCat467: This section of pep 484 talks about `__init__ -> None` annotations: https://peps.python.org/pep-0484/#the-meaning-of-annotations

jakkdl: > With the changes I made with `verify_types.json`, it said that this PR adds 2 new typed functions, so I think this change will be important for 100% type completion for the entire project

That's because it counts the class itself as another symbol that's now fully typed. If you reverted only this one change it wouldn't change the output.

`mypy` has resolved that parenthetical in a different way, and seems like `pyright` does too here, in them not requiring `-> None` if all other parameters to `__init__` are typed. I can't even get `mypy` to complain if I add `--strict`, so I'm a bit curious what configuration you're using that flags it.

PR #2741, opened 2023-08-06
Add types to `core._thread_cache`

TeamSpen210:
Adds types to this module.

### DIFF
--- a/trio/_core/_thread_cache.py
+++ b/trio/_core/_thread_cache.py
@@ -118,7 +118,7 @@ name_counter = count()
 
 class WorkerThread(Generic[T]):
     def __init__(self, thread_cache: ThreadCache) -> None:
-        self._job: tuple[
+        self._job: tuple[  # type: ignore[no-any-unimported]
             Callable[[], T],
             Callable[[outcome.Outcome[T]], object],
             str | None,
@@ -198,7 +198,7 @@ class ThreadCache:
     def __init__(self) -> None:
         self._idle_workers: dict[WorkerThread[Any], None] = {}
 
-    def start_thread_soon(
+    def start_thread_soon(  # type: ignore[no-any-unimported]
         self,
         fn: Callable[[], T],
         deliver: Callable[[outcome.Outcome[T]], object],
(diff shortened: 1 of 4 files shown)


TeamSpen210:
This needs `outcome` stubs to fully pass type checking.

trio/_core/_thread_cache.py
```diff
@@ -185,10 +195,16 @@ def _work(self):
 
 
 class ThreadCache:
-    def __init__(self):
-        self._idle_workers = {}
-
-    def start_thread_soon(self, fn, deliver, name: Optional[str] = None):
+    def __init__(self) -> None:
+        self._idle_workers: dict[WorkerThread[Any], None] = {}
```
A5rocks, line 199: I'm thinking `[object]` works here? Maybe? I'm can't see where we add stuff into this dictionary...

A5rocks (approved):
Looks fine, minor comment.

(the `Optional[T]` to `T | None` took a bit of time to parse xdd)

trio/_core/_thread_cache.py
TeamSpen210: It is actually type-unsafe. The problem is in `ThreadCache.start_thread_soon()` - when we pop off an existing worker thread, that effectively is changing the type of the generic `T` from whatever the worker was before. Maybe I should make it `object` here so I can add a type-ignore there to highlight that.

A5rocks: Ahh duh, I see.

A5rocks: Okay yeah I was thinking maybe there was some covariant coolness we could do because we don't really care about the T used in one function, so maybe it could be an object.

I think I see how that's not really possible here and IMO Any is fine then. Thanks for explaining.

TeamSpen210: I originally had `WorkerThread` non-generic, but that just moves the cast into the `worker._job =` assignment. Which is a bit less safe since you're simultaneously ignoring the type of both callables.

jakkdl (approved):

A5rocks (approved):
I still think this is good, feel free to merge!

PR #2743, opened 2023-08-06
type _io_epoll and _io_kqueue

jakkdl:
Also generates a type guard in `_generated_io_*` so they don't raise errors if checked in other ways than `mypy -m trio`


jakkdl:
I don't really care for writing a test for the decorator logic in `gen_exports`. The lacking patch coverage in kqueue for `monitor_kevent` was already a thing

trio/_core/_io_epoll.py
```diff
@@ -1,23 +1,38 @@
...
 
 
 @attr.s(slots=True, eq=False, frozen=True)
 class _EpollStatistics:
-    tasks_waiting_read = attr.ib()
-    tasks_waiting_write = attr.ib()
-    backend = attr.ib(default="epoll")
+    tasks_waiting_read: int = attr.ib()
+    tasks_waiting_write: int = attr.ib()
+    backend: str = attr.ib(default="epoll")
```
TeamSpen210, line 35: Perhaps we might want to make this a `Literal` (and the same with `kqueue`/`windows`), since then type checkers could understand narrowing checks using this? Maybe also change it to a final/classvar.

trio/_tools/gen_exports.py
```diff
@@ -24,11 +24,25 @@
...
 from ._run import _NO_SEND, GLOBAL_RUN_CONTEXT
 
+if TYPE_CHECKING:
+    import select
+    from socket import socket
+
+    from ._traps import Abort, RaiseCancelT
+
+    from .. import _core
+    from ._unbounded_queue import UnboundedQueue
```
TeamSpen210, line 44: In my `run` PR to avoid adding a lot of irrelevant imports to this file I just `import trio`, then changed the original types to do `trio.whatever.Class`. Perhaps we should just change the processing logic below to have a different set of imports for each file instead though.

trio/_tools/gen_exports.py
```diff
@@ -94,18 +108,41 @@ def create_passthrough_args(funcdef: ast.FunctionDef | ast.AsyncFunctionDef) ->
...
     generated = [HEADER]
+
+    # This is only triggered by check.sh, not test_gen_exports.py
+    if lookup_path == "runner.io_manager" and isinstance(
+        source_path, Path
+    ):  # pragma: no coverage
+        for file_indicator, platform in (
+            ("windows", "win32"),
+            ("kqueue", "darwin"),
+            ("epoll", "linux"),
```
TeamSpen210, line 125: Instead of doing a redundant lookup here, it'd be better to change the `sources_and_lookups`/`to_wrap` to also have the platform name. Maybe also change the type to a attrs class, so we can use keywords to identify the different strings passed in.

trio/_core/_io_kqueue.py
```diff
@@ -96,49 +105,54 @@ def process_events(self, events):
...
+    def monitor_kevent(
+        self, ident: int, filter: int
+    ) -> Iterator[_core.UnboundedQueue[select.kevent]]:
         key = (ident, filter)
         if key in self._registered:
             raise _core.BusyResourceError(
                 "attempt to register multiple listeners for same ident/filter pair"
             )
-        q = _core.UnboundedQueue()
+        q = _core.UnboundedQueue[select.kevent]()
```
TeamSpen210, line 121: I believe `monitor_kevent` is somewhat experimental/not fully tested and implemented, which is why it's failing coverage here.

TeamSpen210 (changes requested):
A few minor things.

trio/_tools/gen_exports.py
```diff
@@ -122,6 +159,9 @@ def gen_public_wrappers_source(source_path: Path, lookup_path: str) -> str:
         # Create the function definition including the body
         func = astor.to_source(method, indent_with=" " * 4)
```
CoolCat467, line 160: If it's possible to fix the type annotation formatting that would be great but if not I understand

CoolCat467 (changes requested):

trio/_core/_generated_io_epoll.py
```diff
@@ -1,31 +1,47 @@
...
+    from .. import _core
+    from ._unbounded_queue import UnboundedQueue
+
 # fmt: off
 
 
-async def wait_readable(fd):
+assert not TYPE_CHECKING or sys.platform=="linux"
+
+async def wait_readable(fd: (int | socket)) ->None:
```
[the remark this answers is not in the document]
TeamSpen210: These are generated from the AST, the library doing that produces this for some reason.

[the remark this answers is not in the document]
A5rocks: Yeah :(

```py
➜  cat x.py

def f(x: int | str) -> None:
  pass
```

```py
In [1]: import x

In [2]: x.f
Out[2]: <function x.f(x: 'int | str') -> 'None'>

...

In [4]: import astor

...

In [6]: astor.to_source(astor.code_to_ast(x.f))
Out[6]: 'def f(x: (int | str)) ->None:\n    pass\n'
```

jakkdl:
Awesome stuff @TeamSpen210 - I would probably not have bothered myself but this is much much nicer and means we can remove it as an exception from `F401` in `pyproject.toml`

trio/_core/_io_epoll.py
jakkdl: Ah yes, works well with the `Protocol` in #2733.
How do you do class variables with an `attr` class, just remove `attr.ib`?

jakkdl: I suppose one could also do a hard-coded property or something

TeamSpen210: I was thinking actually to do a 1-value literal, so the type checker knows that `if stats.backend == "epoll"` will fail for the others. For class variables, `attrs` is smart enough to detect `ClassVar` annotations and skip them.

trio/_tools/gen_exports.py
TeamSpen210: Could make `gen_exports` call `black` after generating even. But definitely a separate PR.

jakkdl:
good stuff, thanks @TeamSpen210!

PR #2711, opened 2023-07-19
Docs: Fix line numbers on tutorial.rst

kulothunganug:
Issue: Documentation doesn't aligns with the code block's line number.

In image format:
image
image

Issue #2681, opened 2023-07-03
python version for CI formatting check


jakkdl:
Is there any special reason we're staying on 3.8 for the CI tests? If we bumped it we would get additional typing features and not have to do `from __future__ import annotations`, and we're going to have to bump 3.8 *eventually* when it reaches EOL in 2024-10.
Does some tooling break if we were to typecheck with 3.10 or 3.11 features or something? Is there any best-practice wrt typechecking/linting your code on oldest/newest version supported?

A5rocks:
I don't quite understand the issue here :P 

Are you talking about just the formatting? Or the tests themselves. The tests themselves are due to our compatibility window. The formatting check shouldn't really care about the version:
 - `from __future__ import annotations` prevents runtime breakage in the *tests*
 - using unsupported type annotations means that 3rd party users cannot introspect our types at runtime on those other platforms and I'm not quite sure if that's a price I'm personally willing to pay.
 - mypy and black (probably flake8 too?) have configuration to change what version you are checking them on: they do not depend on the runtime version for more than syntax (at least for mypy, which is what I know)

However given all that: I think bumping the version for the formatting job would be fine, even if it doesn't change anything.

jakkdl:
> from __future__ import annotations prevents runtime breakage in the tests

oh right, derp.


So unless the feature set of mypy/black/flake8 were to differ between installed versions (due to some new feature not getting implemented when running against an old AST, or similar), it shouldn't change anything in theory. I have a suspicion that stuff would actually behave differently in practice though, and we'd also get access to new language features in general (wrt stacktraces, speedups, etc). I think most tooling should have good 3.11 support now, but could go with 3.10 to be more conservative.

This behavior seemed to be due to the python version on the CI, though I'm not 100% if it's the python version it's run on vs the runtime flag of version to check against.
https://github.com/python-trio/trio/blob/4816f0e47a17b996ca3d44319b593fa35320e55f/trio/_channel.py#L97-L100

A5rocks:
We probably didn't configure `mypy` to run against a certain python version. We do so for platforms (see `check.sh`) but nothing for python version as far as I can tell which means `mypy` will default to using the runtime version of Python. We didn't forget to do this for `black`, however.

jakkdl:
Ah, you're right - and we should definitely set that so it's easier for developers to locally typecheck their code without having to run `mypy` inside a `py38` virtualenv.
`mypy.ini` does have the lines
```ini
[...]

PR #2744, opened 2023-08-07
Add pre-commit

CoolCat467:
Up for debate, but I find pre-commit auto-formatting code for my own projects very useful, and thought I would suggest trio adopting it as well.

Basically pre-commit uses git's hooks system to do anything you want it too, all defined from one file. They have more information about it on their website (https://pre-commit.com/).
They also have a continuous integration system (more info at https://pre-commit.ci/) that means you can do things like automatically doing things like code formatting on pull requests and committing directly.

This PR adds the config file pre-commit looks for, and has pre-commit do black code formatting and isort, as well as flake8, but I'm not super sure about that because sometime I plan to make another PR for moving over to use ruff instead, so it's skipped in the ci run section currently.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+        types: [file]
+        types_or: [python, pyi]
+
+ci:
+  autofix_commit_msg: "[pre-commit.ci] auto fixes from pre-commit.com hooks"
+  autofix_prs: true
+  autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
+  autoupdate_schedule: weekly
+  skip: [flake8]
+  submodules: false
```
A5rocks, line 36: I don't think we want pre-commit CI (other than *maybe* for updating the pre-commit hooks). We're intentionally not pushing commits based on what autoformatters want since that might interfere with an WIP PR, for instance.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - repo: https://github.com/pycqa/isort
+    rev: 5.12.0
+    hooks:
+      - id: isort
+  - repo: https://github.com/pycqa/flake8
+    rev: 6.1.0
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          - "flake8-pyproject==1.2.3"
```
jakkdl, line 26: does this need to be pinned? Will it autoupdate?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - id: check-yaml
+      - id: check-toml
+      - id: check-merge-conflict
+      - id: mixed-line-ending
+      - id: check-case-conflict
+  - repo: https://github.com/psf/black
+    rev: 23.7.0
+    hooks:
+      - id: black
+        language_version: python3.8
```
jakkdl, line 16: this is specified in pyproject.toml, so don't think you should specify it here.

.pre-commit-config.yaml
jakkdl: Running Black on WIP PR's should not be a problem. `isort` could in theory be a problem, but I suspect it'd help more often than it hurts - and in any case it doesn't modify your local repo and you can review the commit before `isort` does its thing if you want.

But it can definitely be messy, it's probably a tradeoff between experienced contributors (that will have taken the decision whether to run the pre-commit hooks locally) vs new contributors (that might struggle with setting up tooling and will waste time doing several CI runs until stuff passes) and people maybe having to review unformatted code.

jakkdl (changes requested):
* Yeah I've been thinking about this in #2699 and created a `.pre-commit-config.yaml` for personal use myself.
* `pre-commit run --all-files` will trigger a lot of changes in the `black` and `isort` hooks on files not currently checked by `ci.sh`. Either ignore these or reformat them (in this or another PR). Flake8 also complains about a couple files
* You're missing `codespell`
* We probably want mypy in this once one can run `mypy .` without triggering errors. Running `mypy -m trio -m trio.testing --platform=x.y.z` should be possible, but clunky and slow

* If the long-term plan is to replace `ci.sh` (which I'm strongly in favor of) you also want the ability to run
  * `_tools/gen_exports.py`
  * `_tests/_check_type_completeness.py` or at least `pyright --verifytypes` (on all platforms) once that passes
  * `pip-compile -q ...` ... except good luck doing that without `tox` to manage python version, but there's discussion in #2681 and there's #2595
  * Several of these should probably be default-off though
* I personally also have `autoflake` enabled
* I've had `https://github.com/macisamuele/language-formatters-pre-commit-hooks` `--pretty-format-toml` enabled for this repo, but dropped it at some point for some reason.

.pre-commit-config.yaml
jakkdl: 2023-08-08T13:55:07,201929064+02:00
case in point :sweat_smile:

.pre-commit-config.yaml
CoolCat467: When pre-commit ci is enabled for a repository, yes, it will be like dependabot and auto-create PRs to update the dependencies.

A5rocks:
FYI mypy shouldn't be used as a pre-commit hook. (The whole point of pre-commit hooks is to run checks on some subset of files and yet mypy should really check all files)

.pre-commit-config.yaml
A5rocks: Yeah I keep using the web editor for small changes :( 

I do think for larger PRs changing the code from under their feet is a bit annoying.

CoolCat467:
Personally I think the best thing to do is have pre-commit do auto-fixes and actual github actions do the tests that don't have auto-fixes or things that shouldn't be auto-fixed

.pre-commit-config.yaml
CoolCat467: Removed python version specifier in 25e0281

jakkdl:
> FYI mypy shouldn't be used as a pre-commit hook. (The whole point of pre-commit hooks is to run checks on some subset of files and yet mypy should really check all files)

I personally very much like running mypy locally as a pre-commit hook to catch small type errors before I push, and only running it on a subset of files speeds it up so so much. But of course CI should check all files to catch errors introduced in other files

jakkdl:
> Personally I think the best thing to do is have pre-commit do auto-fixes and actual github actions do the tests that don't have auto-fixes or things that shouldn't be auto-fixed

hm, I personally like it for any check that's quick to save you having to wait for a failing CI. Esp as pre-commit will limit the subset of files it checks so it's much much faster than running them manually locally as well unless you set up something fancy. Esp something like codespell I see no reason at all not to put in pre-commit

I also think there's a strong case for having all checks configured and run from the same place, parsing the output of pre-commit is much better than digging into the text log of `check.sh`. You could set checks to be default-off unless in CI for stuff that's slow or otherwise.

jakkdl:
For reference, this is the config I've crafted for flake8-trio: https://github.com/Zac-HD/flake8-trio/blob/main/.pre-commit-config.yaml (most of the flake8 checks are redundant since the addition of ruff though)

.pre-commit-config.yaml
jakkdl: using the web editor seems like a strong case in favor of it. Leaving it off at the beginning seems reasonable regardless just to make sure it works and all, and then we can try toggling it later

A5rocks:
> > FYI mypy shouldn't be used as a pre-commit hook. (The whole point of pre-commit hooks is to run checks on some subset of files and yet mypy should really check all files)
> 
> I personally very much like running mypy locally as a pre-commit hook to catch small type errors before I push, and only running it on a subset of files speeds it up so so much. But of course CI should check all files to catch errors introduced in other files

As you've noted elsewhere (w/r/t `mypy trio` vs `mypy -m trio`) the files you pass and how you pass them is important: running mypy specifically over certain files means some errors might not show up and that others might appear when they don't actually exist. (If it means anything, the stance of mypy's core devs is that pre-commit doesn't work great for mypy)

jakkdl:
> > > FYI mypy shouldn't be used as a pre-commit hook. (The whole point of pre-commit hooks is to run checks on some subset of files and yet mypy should really check all files)
> > 
> > 
> > I personally very much like running mypy locally as a pre-commit hook to catch small type errors before I push, and only running it on a subset of files speeds it up so so much. But of course CI should check all files to catch errors introduced in other files
> 
> As you've noted elsewhere (w/r/t `mypy trio` vs `mypy -m trio`) the files you pass and how you pass them is important: running mypy specifically over certain files means some errors might not show up and that others might appear when they don't actually exist. (If it means anything, the stance of mypy's core devs is that pre-commit doesn't work great for mypy)

The only issue I've hit with false-positives so far is on files that were conditionally imported on different platforms, but that was easily resolved with adding platform type guards to the files. This wasn't isolated to pre-commit though, it also happened with `mypy trio` (which we might want to switch to in order to include tests and everything) and if you had your IDE configured to run type-checking (which is very common nowadays, so is a very strong argument IMO. My setup were blasting errors at me any time I had those files open). Not finding errors is of course abundant, but CI running mypy on all files will check for that.

I understand `mypy` not wanting to official support it (https://github.com/python/mypy/issues/13916), but they do link the pre-commit mirror directly in their readme: https://github.com/python/mypy#integrations

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+    rev: v2.2.5
+    hooks:
+      - id: codespell
+
+ci:
+  autofix_commit_msg: "[pre-commit.ci] auto fixes from pre-commit.com hooks"
+  autofix_prs: true
+  autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
+  autoupdate_schedule: weekly
+  skip: [flake8,codespell]
```
jakkdl, line 38: skipping these in CI doesn't make sense to me. The only thing you're achieving is hiding errors from them in the vast expanse of the output log of `check.sh` instead of more visible in the pre-commit output.

CoolCat467: Removed skips in [caa68e7](https://github.com/python-trio/trio/pull/2744/commits/caa68e76245cc4fd08e6e881a4b5f75de1d17416)

jakkdl (commented):
Mostly just the problems with running `pre-commit run --all-files` that remains a stopper.

CoolCat467:
I think I'll make another PR running it on all files and fixing anything that happens once this one is merged

CoolCat467:
> Mostly just the problems with running `pre-commit run --all-files` that remains a stopper.

Ok so I decided to fix all the files, that's changed as of d8d6a2d

jakkdl (changes requested):
To avoid another https://github.com/python-trio/trio/pull/2711, I think subdirectories under `docs/source` should be ignored. And not just a matter of updating references to them now, as I think extra empty lines from isort or black is quite unnecessary in those short examples.

This should be set in `pyproject.toml` (for both isort and black) so they're ignored by manual runs as well. To still get fixes in `conf.py` and `local_customization.py` the easiest is probably to explicitly list the three subdirectories.

reformatting notes-to-self shouldn't be a problem afaik

A5rocks (commented):
Just a reminder about https://github.com/python-trio/trio/pull/2744#discussion_r1288915259 :P

CoolCat467:
> Just a reminder about [#2744 (comment)](https://github.com/python-trio/trio/pull/2744#discussion_r1288915259) :P

Disabled black and isort for CI autofixes in [316838a](https://github.com/python-trio/trio/pull/2744/commits/316838a28f9b36fd9dde6653de083c9713c0900d) (though I do hope we will turn that on)

notes-to-self/proxy-benchmarks.py
```diff
@@ -135,7 +161,7 @@ def check(cls):
 
 COUNT = 1000000
 try:
-    import __pypy__
+    import __pypy__  # noqa: F401  # __pypy__ improrted but unused
```
jakkdl, line 164: typo

notes-to-self/proxy-benchmarks.py
```diff
@@ -15,53 +16,67 @@ def __getattr__(self, name):
...
-    code = textwrap.dedent("""
+    code = textwrap.dedent(
+        """
         def wrapper(self, *args, **kwargs):
             return self._wrapped.{}(*args, **kwargs)
-    """.format(method))
+    """.format(
+            method
+        )
+    )
```
jakkdl, line 60: golly this is ugly, but not worth doing anything about

notes-to-self/graceful-shutdown-idea.py
```diff
@@ -31,9 +33,11 @@ async def stream_handler(stream):
     while True:
         with gsm.cancel_on_graceful_shutdown():
             data = await stream.receive_some()
+            print(f"{data = }")
```
jakkdl, line 36: what's up with this random added print?

jakkdl (approved):
just a couple small nits

notes-to-self/graceful-shutdown-idea.py
CoolCat467: Otherwise the flake8 says variable is unused, and I thought it would be better to use the data than remove the variable assignment. Up to you what you think best action is.

notes-to-self/proxy-benchmarks.py
CoolCat467: Fixed in 6aaec43f

notes-to-self/proxy-benchmarks.py
CoolCat467: Yea definitely not very pretty

notes-to-self/graceful-shutdown-idea.py
jakkdl: sounds good :+1:

CoolCat467:
To ensure this doesn't get too far out of date, I am going to merge this.

A5rocks:
Mind making a followup that updates documentation on contributing to suggest this?

CoolCat467:
I don't mind at all, I'll be opening a new PR shortly for that.

PR #2745, opened 2023-08-07
Add type annotations for `_ssl.py`

CoolCat467:
This PR adds type annotations to `_ssl.py`

Three interesting thins I noticed when making sure the type annotations were good:
1. There was a type annotation bug in `trio.abc.Listener` not using its `T_resource` generic in accept, now fixed
2. Mypy says `SSLStream.receive_some` could have potentially returned `None` if my typing for `SSLStream._retry` is correct. I don't think this is entirely true, I think everything is supposed to raise a `trio.BrokenResourceError` exception, but if it raised something else it could have returned `None`. There is a type-narrowing `assert not None` there now, but I am curious if anyone has more information.
3. `SSLStream.unwrap` sets `self.transport_stream` to `None`, and nothing else checks to make sure it's not `None` anywhere. I assume this is ok since it also sets the internal state to `CLOSED`, so I put a type ignore there to silence the error. The reason I hesitate to allow `self.transport_stream` to be None is because the docstring says it's required to be a Stream object.


trio/_ssl.py
```diff
@@ -207,15 +214,19 @@ class NeedHandshakeError(Exception):
...
     """
 
+    __slots__ = ()
+
 
 class _Once:
-    def __init__(self, afn, *args):
+    __slots__ = ("_afn", "_args", "started", "_done")
+
+    def __init__(self, afn: Callable[..., Awaitable[Any]], *args: Any) -> None:
```
A5rocks, line 223: I think `object` can be used for `*args`, and we don't care about the result of `afn` so just `Awaitable[object]` is AOK.

trio/_ssl.py
```diff
@@ -331,18 +342,18 @@ class SSLStream(Stream, metaclass=Final):
...
+        ssl_context: _stdlib_ssl.SSLContext,
         *,
-        server_hostname=None,
-        server_side=False,
-        https_compatible=False,
-    ):
+        server_hostname: str | None = None,
+        server_side: bool = False,
+        https_compatible: bool = False,
+    ) -> None:
```
jakkdl, line 351: nit: inconsequential `-> None`. We ought to create a style guide somewhere for typing, and/or add a check to #2744 that removes/adds it.

trio/_ssl.py
```diff
@@ -741,11 +760,11 @@ async def unwrap(self):
             await self._handshook.ensure(checkpoint=False)
             await self._retry(self._ssl_object.unwrap)
             transport_stream = self.transport_stream
-            self.transport_stream = None
             self._state = _State.CLOSED
+            self.transport_stream = None  # type: ignore[assignment]  # Messy, but State closed, should not use, should be fine...
```
jakkdl, line 764: Make the comment a bit more verbose to explain what's going on here

jakkdl (changes requested):
Guess which file needs to be updated...? `pyproject.toml`!

Does your setup not use it for enabling checks? In my workflow I first need to update `pyproject.toml` in order to be able to get the errors and see what needs fixing.

Otherwise mostly just nits

trio/_ssl.py
```diff
@@ -408,16 +419,16 @@ def __getattr__(self, name):
         else:
             raise AttributeError(name)
 
-    def __setattr__(self, name, value):
+    def __setattr__(self, name: str, value: Any) -> None:
```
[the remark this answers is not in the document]
CoolCat467: Changed in edbf981

trio/_ssl.py
CoolCat467: Changed in edbf981

trio/_ssl.py
CoolCat467: Changed in edbf981

trio/_ssl.py
jakkdl: ?
That's a 3 day old commit and doesn't look to contain any changes to the comment

trio/_ssl.py
CoolCat467: It is not inconsequential, I am fairly certain that `verify_types.json` changes for the better when it's there

TeamSpen210:
> How should this issue with readthedocs be addressed?
> 
> ```
> /home/docs/checkouts/readthedocs.org/user_builds/trio/checkouts/2745/trio/abc.py:docstring of trio.abc.Listener.accept:1: WARNING: py:class reference target not found: trio._abc.T_resource
> ```

You've got to add it to the ignore list in `docs/conf.py`, rather annoying.

trio/_ssl.py
jakkdl: Are you referring to https://github.com/python-trio/trio/pull/2740#discussion_r1285180160 ?
I removed the `-> None` without any impact on `verify_types.json`

trio/_tests/verify_types.json
```diff
@@ -7,11 +7,11 @@
     "warningCount": 0
   },
   "typeCompleteness": {
-    "completenessScore": 0.9186602870813397,
+    "completenessScore": 0.9202551834130781,
     "exportedSymbolCounts": {
-      "withAmbiguousType": 0,
-      "withKnownType": 576,
-      "withUnknownType": 51
+      "withAmbiguousType": 1,
```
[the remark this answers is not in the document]
CoolCat467: I can't seem to figure out how to run check_type_completeness properly. It says it can't find module trio to check.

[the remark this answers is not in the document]
jakkdl: adding a `py.typed` file and installing trio without the `-e` flag should do it (don't ask me why though). Running `pyright --verifytypes` probably fails the same way. Otherwise I can share my tox config for it tomorrow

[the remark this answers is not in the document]
TeamSpen210: Here's the error:
```
trio._ssl.SSLStream.transport_stream
  f:\Git\clones\python\trio\trio\_ssl.py:764:18 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "Stream"
  f:\Git\clones\python\trio\trio\_ssl.py:764:18 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "None"
```
Specifically, this line:
```py
self.transport_stream = None  # type: ignore[assignment]  # State is CLOSED now, nothing should use
```
You'll want to add a variable annotation for that, since it's just `Stream` in `__init__`.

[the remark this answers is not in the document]
CoolCat467: Have you read through point 3 in my first comment for discussing this PR? The docstring says it's required to be a `Stream` object and nothing anywhere else is checking to make sure `transport_stream` is valid.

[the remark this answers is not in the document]
TeamSpen210: It might be confusing Pyright still, since it's using both assignment statements to infer the type. If you add an annotation to the one in `__init__()`, it should fix the error. Despite the name, `type: ignore` just disables errors, type checkers still have to analyse that line of code to figure out what it does...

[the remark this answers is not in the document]
CoolCat467: Fixed in c896899

pyproject.toml
```diff
@@ -52,14 +52,15 @@ module = [
     "trio._abc",
     "trio._core._entry_queue",
     "trio._core._local",
-    "trio._core._unbounded_queue",
     "trio._core._thread_cache",
+    "trio._core._unbounded_queue",
```
A5rocks, line 56: Unnecessary movement

trio/_ssl.py
```diff
@@ -1,3 +1,18 @@
+from __future__ import annotations
```
A5rocks, line 1: I don't think moving the import block above the comment is necessary. (future imports can be after comments IIRC)

trio/_ssl.py
```diff
@@ -207,15 +214,19 @@ class NeedHandshakeError(Exception):
 
     """
 
+    __slots__ = ()
+
 
 class _Once:
-    def __init__(self, afn, *args):
+    __slots__ = ("_afn", "_args", "started", "_done")
```
A5rocks, line 221: Maybe you missing my initial comment: I think this and the `__slots__` above should probably be moved to your other PR.

trio/_ssl.py
```diff
@@ -431,7 +442,13 @@ def _check_status(self):
     # comments, though, just make sure to think carefully if you ever have to
     # touch it. The big comment at the top of this file will help explain
     # too.
-    async def _retry(self, fn, *args, ignore_want_read=False, is_handshake=False):
+    async def _retry(
+        self,
+        fn: Callable[..., T],
+        *args: Any,
```
A5rocks, line 448: `object` probably works here

pyproject.toml
CoolCat467: Yes, but that's from sorting them alphabetically. I can change it back of course but I thought there should be at least some order to it.

trio/_ssl.py
CoolCat467: This happened after running isort. Correct, __future__ can be after only module docstrings and `#` comments.

pyproject.toml
A5rocks: Ah missed that. My bad!

trio/_ssl.py
A5rocks: Didn't realize isort did that (I assumed it would keep everything under a comment as long as everything before it was under the comment), alright!

pyproject.toml
jakkdl: (https://github.com/macisamuele/language-formatters-pre-commit-hooks would automatically do this)

jakkdl (changes requested):
+1 to a5rock on leaving `__slots__` to separate PR as they've been controversial, otherwise seems ready to merge.

trio/_ssl.py
CoolCat467: Moved to #2754

trio/_ssl.py
CoolCat467: To be honest I am confused why `Any` would be any worse

TeamSpen210 (approved):

trio/_ssl.py
```diff
@@ -331,18 +342,18 @@ class SSLStream(Stream, metaclass=Final):
...
-        ssl_context,
+        transport_stream: Stream,
+        ssl_context: _stdlib_ssl.SSLContext,
         *,
-        server_hostname=None,
-        server_side=False,
-        https_compatible=False,
-    ):
-        self.transport_stream = transport_stream
+        server_hostname: str | None = None,
```
TeamSpen210, line 348: From `_highlevel_ssl_helpers.open_ssl_over_tcp_stream()`, it seems that `server_hostname` should actually be `str | bytes | None`.

trio/_ssl.py
A5rocks: Oops, I think you forgot to remove the changes from here :P

trio/_ssl.py
A5rocks: If we ever accidentally use an `Any` value, then that's really un-type-safe. Also, inference and everything can lead to badly inferred types with `Any` (people normally don't want that inferred).

A5rocks (commented):
Just a couple comments.

trio/_ssl.py
CoolCat467: Changed in acfcead1

trio/_ssl.py
CoolCat467: Thank you for pointing that out. It turns out there is a bug in typeshed that lead me to thinking its only `str | None`

Type ignore in it's place will be fixed with https://github.com/python/typeshed/pull/10590

trio/_ssl.py
CoolCat467: Yep, forgot to push the commit

trio/_ssl.py
jakkdl: https://typing.readthedocs.io/en/latest/source/best_practices.html#using-any-and-object
https://adamj.eu/tech/2021/05/07/python-type-hints-use-object-instead-of-any/

it took me a bit to grok the difference, but A5rocks has shown me the light

jakkdl (approved):

trio/_ssl.py
```diff
@@ -207,15 +214,17 @@ class NeedHandshakeError(Exception):
 
     """
 
+    __slots__ = ()
+
```
A5rocks, line 218: Geh this should probably also move out. I won't make this review "requested changes" though :P

trio/_ssl.py
A5rocks: Just bumping on this!

trio/_ssl.py
```diff
@@ -331,24 +340,24 @@ class SSLStream(Stream, metaclass=Final):
...
         self._outgoing = _stdlib_ssl.MemoryBIO()
-        self._delayed_outgoing = None
+        self._delayed_outgoing: bytes | None = None
         self._incoming = _stdlib_ssl.MemoryBIO()
         self._ssl_object = ssl_context.wrap_bio(
             self._incoming,
             self._outgoing,
             server_side=server_side,
-            server_hostname=server_hostname,
+            server_hostname=server_hostname,  # type: ignore[arg-type]  # Typeshed bug, does accept bytes as well (typeshed#10590)
```
A5rocks, line 360: Looks like that PR was merged, meaning next mypy release should support this! Nice!

trio/_ssl.py
```diff
@@ -897,16 +914,16 @@ class SSLListener(Listener[SSLStream], metaclass=Final):
...
         self,
-        transport_listener,
-        ssl_context,
+        transport_listener: Listener[Stream],
+        ssl_context: _stdlib_ssl.SSLContext,
         *,
-        https_compatible=False,
-    ):
+        https_compatible: bool = False,
+    ) -> None:
```
A5rocks, line 921: FWIW this `-> None` is probably also unnecessary.

A5rocks (commented):
Just some comments!

trio/_ssl.py
CoolCat467: Removed in cf7a777

trio/_ssl.py
CoolCat467: https://mypy.readthedocs.io/en/stable/config_file.html#confval-local_partial_types

trio/_ssl.py
jakkdl: Yeah, I haven't seen any reason why it would change anything.

jakkdl:
merging main ...
```
$ tox -e verifytypes
******************** 
Checking type completeness hasn't gone down...
creating py.typed
completenessScore has gone up from 0.9904458598726115 to 1
withUnknownType has gone down from 6 to 0
withKnownType has gone up from 622 to 628
  File "trio/_tests/check_type_completeness.py", line 190, in <module>
    sys.exit(main(args))
  File "trio/_tests/check_type_completeness.py", line 131, in main
    assert (
AssertionError: Fully type complete! Delete this script and instead directly run `pyright --verifytypes=trio` (consider `--ignoreexternal`) in CI and checking exit code.
```
:rocket: :rocket:

PR #2746, opened 2023-08-08
Address an old comment waiting for `Self` support

A5rocks:
Noticed this quite a while ago but I've been getting super side tracked :(

This hopefully won't fail anything.


TeamSpen210 (approved):
Well in this case it's not really necessary since both are `NoPublicConstructor` (and therefore `Final`), but it's a little nicer to read intent-wise.

PR #2747, opened 2023-08-08
Add types to most of `trio.testing`

TeamSpen210:
This adds types to `trio.testing`, except for `_fake_net` (which is considerably more complex and should probably go in its own PR). It's fairly straightforward, though in a few cases I made functions forward on both `*args`/`**kwargs` so they can use `ParamSpec`, instead of just one or the other. For `_check_streams`, to correctly type things the functions now expect/return tuples instead of other iterables that just happen to be 2-long - in practice I don't think people would do anything else anyway. It'll just fail type checking if so.


trio/_tests/verify_types.json
```diff
@@ -112,36 +112,6 @@
...
-      "trio.testing.check_one_way_stream",
-      "trio.testing.check_two_way_stream",
-      "trio.testing.lockstep_stream_one_way_pair",
-      "trio.testing.lockstep_stream_pair",
-      "trio.testing.memory_stream_one_way_pair",
-      "trio.testing.memory_stream_pair",
-      "trio.testing.memory_stream_pump",
-      "trio.testing.open_stream_to_socket_listener",
-      "trio.testing.trio_test",
       "trio.testing.wait_all_tasks_blocked",
```
jakkdl, line 115: If anybody else is confused: `wait_all_tasks_blocked` is defined in `_core/[_generated]_run.py` but exported in `testing/__init__.py` and is removed from this file in #2733

trio/testing/_check_streams.py
```diff
@@ -28,13 +36,13 @@ async def __aexit__(
...
-            await aclose_forcefully(self._both[0])
+            await aclose_forcefully(self._first)
         finally:
-            await aclose_forcefully(self._both[1])
+            await aclose_forcefully(self._second)
 
 
 @contextmanager
-def _assert_raises(exc):
+def _assert_raises(exc: type[BaseException]) -> Generator[None, None, None]:
```
jakkdl, line 45: Could do `Iterator[None]` here

jakkdl (approved):
Mostly just a bunch of nits, looks great overall! And imo the types increase readability of a lot of this code

trio/testing/_check_streams.py
CoolCat467: Personally I would keep is annotated as Generator, since it is a generator, I feel like Iterator is best used to describe the return type for `__iter__` on classes

trio/testing/_check_streams.py
```diff
@@ -67,18 +79,18 @@ async def check_one_way_stream(stream_maker, clogged_stream_maker):
         assert isinstance(s, SendStream)
         assert isinstance(r, ReceiveStream)
 
-        async def do_send_all(data):
-            with assert_checkpoints():
-                assert await s.send_all(data) is None
+        async def do_send_all(data: bytes) -> None:
```
CoolCat467, line 82: If `s` is a Stream object here, it actually accepts `bytes | bytearray | memoryview` apparently.

trio/testing/_memory_streams.py
```diff
@@ -245,21 +253,26 @@ def close(self):
...
             self.close_hook()
 
-    async def aclose(self):
+    async def aclose(self) -> None:
         """Same as :meth:`close`, but async."""
         self.close()
         await _core.checkpoint()
 
-    def put_data(self, data):
+    def put_data(self, data: bytes | bytearray) -> None:
```
CoolCat467, line 261: Might also support `memoryview`, but not certain

trio/testing/_memory_streams.py
```diff
@@ -448,15 +463,15 @@ async def _wait_for(self, fn):
...
         self._sender_closed = True
         self._something_happened()
 
-    def close_receiver(self):
+    def close_receiver(self) -> None:
         self._receiver_closed = True
         self._something_happened()
 
-    async def send_all(self, data):
+    async def send_all(self, data: bytes | bytearray) -> None:
```
CoolCat467, line 474: Again, might support `memoryview`

trio/testing/_memory_streams.py
```diff
@@ -515,39 +530,39 @@ async def receive_some(self, max_bytes=None):
...
+    def close(self) -> None:
         self._lbq.close_sender()
 
-    async def aclose(self):
+    async def aclose(self) -> None:
         self.close()
         await _core.checkpoint()
 
-    async def send_all(self, data):
+    async def send_all(self, data: bytes | bytearray) -> None:
```
CoolCat467, line 543: Maybe `memoryview` as well not certain

CoolCat467 (approved):
Looks good except noted, most of noted is that I think a lot of places might accept `memoryview` objects as well

trio/testing/_check_streams.py
Zac-HD: Personally I'm a fan of using `Iterable[T]` (not `Iterator`!) when you use `yield` but don't intend people to use [generator-only interfaces like `.send(...)`, `.throw(...)`, or `.close()`](https://docs.python.org/3/reference/expressions.html#generator-iterator-methods).  In this case it doesn't matter though, since we immediately wrap with `@contextmanager`.

TeamSpen210: The contextmanager stubs are actually slightly wrong - a generic iterator won't work, only a generator (since they use send/throw). There's an [open issue in typeshed](https://github.com/python/typeshed/issues/2772) about fixing this, but there's the problem of causing a bunch of nuisance errors in existing code.

Outside this context I do also prefer `Iterable` to avoid promising `send`/`throw`.

trio/testing/_check_streams.py
TeamSpen210: It might, but it doesn't really matter - these are internal nested functions, they're only being called with `bytes` objects.

docs/source/conf.py
```diff
@@ -90,13 +91,34 @@
...
+                r"StreamMaker\[([a-zA-Z ,]+)]",
+                lambda match: f"typing.Callable[[], typing.Awaitable[tuple[{match.group(1)}]]]",
+                signature,
+            )
+            signature = signature.replace(
+                "AsyncHook", "typing.Callable[[], typing.Awaitable[object]]"
+            )
+            signature = signature.replace("SyncHook", "typing.Callable[[], object]")
+
+    return signature, return_annotation
```
jakkdl, line 111: Oh wow, having to do this whenever we introduce type aliases in public functions is a major pain. Thankfully not too bad to expand this function, but :<

TeamSpen210: There's probably a better approach, we could parse with the AST module, introspect the function object etc. Though the advantage is we can do whatever we want to the signature here.

The alternative is to publicly document type aliases, might need a custom directive to properly add them as a "class"...

jakkdl: some of the type aliases might make sense to publicly document, as the same logic of "make the signature readable" and more specifically "make it obvious that all these functions take the same type of parameter" is valid just as much, if not more, in reading the docs as when reading the code.

And yeah, looking at https://trio--2747.org.readthedocs.build/en/2747/reference-testing.html#trio.testing.MemorySendStream it's ... far from readable.

(just realized: would be great if readthedocs provided a diff view)

TeamSpen210: Sphinx 7.0 has the functionality to automatically swap to 1 arg per line for long signatures. Once the [next version of `sphinx_rtd_theme`](https://github.com/readthedocs/sphinx_rtd_theme/pull/1464) releases, we should be able to use that.

trio/_core/_run.py
```diff
@@ -2152,7 +2152,7 @@ def setup_runner(
 
 
 def run(
-    async_fn: Callable[..., RetT],
+    async_fn: Callable[..., Awaitable[RetT]],
```
A5rocks, line 2155: Geh. I assume there's no other things like this?

trio/testing/_check_streams.py
```diff
@@ -2,24 +2,33 @@
...
-class _ForceCloseBoth:
-    def __init__(self, both):
-        self._both = list(both)
+    ArgsT = ParamSpec("ArgsT")
 
-    async def __aenter__(self):
-        return self._both
+Res1 = TypeVar("Res1", bound=AsyncResource)
+Res2 = TypeVar("Res2", bound=AsyncResource)
+StreamMaker: TypeAlias = "Callable[[], Awaitable[tuple[Res1, Res2]]]"
```
A5rocks, line 23: If this is a string only because of `tuple`, maybe use capital-T `Tuple`? (I've been so out of it I forgot what the convention here is...)

trio/testing/_memory_streams.py
```diff
@@ -1,16 +1,29 @@
...
 from .._highlevel_generic import StapledStream
 from ..abc import ReceiveStream, SendStream
 
+if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
+
+AsyncHook: TypeAlias = "Callable[[], Awaitable[object]]"
+# Would be nice to exclude awaitable here, but currently not possible.
+SyncHook: TypeAlias = "Callable[[], object]"
```
A5rocks, line 17: Do these have to be strings?

A5rocks (approved):
Looks good, though I'm not really able to review this too well without spending way more time than I'd like :(

(as a sidenote we should probably go through a couple of the tests and type them to make sure the user-side types are sane...)

trio/testing/_check_streams.py
TeamSpen210: It's actually because of Callable + Awaitable too.

A5rocks: Oh, is this maybe cause docs?

jakkdl:
yeah, once we're done with the public API we should
1. ask users to clone master and run against their code & report problems
  * are there any typed open source trio repos currently using trio-typing we can test against?
2. type tests
3. run tests in trio-typing to see if any of those stubs disagree

I think there's diminishing returns in reviewing PRs extremely thoroughly and it's more or less guaranteed that some problems will slip through, but it's more efficient to just catch those later imho - esp as type errors don't affect functionality.

pyproject.toml
```diff
@@ -70,6 +70,13 @@ module = [
     "trio._sync",
     "trio._tools.gen_exports",
     "trio._util",
+    "trio.testing",
+    "trio.testing._check_streams",
+    "trio.testing._checkpoints",
+    "trio.testing._memory_streams",
+    "trio.testing._network",
+    "trio.testing._sequencer",
+    "trio.testing._trio_test",
```
jakkdl, line 79: can `_highlevel_generic` be added here in this PR? given that #2756 doesn't touch it

jakkdl: added in 2b669c7

trio/testing/_memory_streams.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import operator
-from typing import TYPE_CHECKING, Awaitable, Callable
+from typing import TYPE_CHECKING, Awaitable, Callable, TypeVar
```
CoolCat467, line 4: I would do `from collections.abc import Awaitable, Callable`

CoolCat467 (changes requested):

trio/testing/_memory_streams.py
TeamSpen210: In this case it's necessary because of the type aliases below.

CoolCat467: That kind of sucks but I understand

CoolCat467 (approved):

jakkdl (changes requested):
Remove the ignores from `pyproject.toml`:
```
# 2747
"trio/testing/_network",
"trio/testing/_trio_test",
"trio/testing/_checkpoints",
"trio/testing/_check_streams",
"trio/testing/_memory_streams",
```

jakkdl (approved):
or I can just do it myself when there's no other problem :p

PR #2751, opened 2023-08-10
add types to _core/_traps

jakkdl:
See discussion in gitter for typing `_async_yield`: https://matrix.to/#/!OqDVTrmPstKzivLwZW:gitter.im/$-gvj3zldyiWi_2Aacz-Cepwuf2k3506BkiSj3EpZ36k?via=gitter.im&via=matrix.org&via=ancap.tech

Unfortunately lots of `Any`, but afaik it's not really possible to get rid of them with the current structure.


jakkdl:
just some `Outcome`/`no-any-unimported` failing in `check_formatting`.
readthedocs is also failing due to `Outcome` - it might be possible to link through to the Outcome docs, if there are any?

jakkdl:
> Looks as good as it can get. Half the types rely on the semantics of some other async library, so by definition we have no clue what it'll produce. Though, perhaps we could instead make everything `object`, require users to check what types get produced.

I guess it's a matter of how common false-positives will be and trivially inconveniencing users vs how many errors will be caught. My impression with these functions is that they most of the time will return what users expect and as such not worth requiring everybody to cast/isinstance the value? If they do want to be maximally careful it should get caught by having maximum strictness wrt `Any` in type checker strictness

jakkdl:
While digging in RTD to fix `Outcome` I also fixed an issue with `OpenSSL.SSL.Context` partly caused by `TaskStatus` weirdness

PR #2752, opened 2023-08-12
Add a Sphinx extension to specially handle typevars in docstrings

TeamSpen210:
Instead of just ignoring missing references for typevars, this applies specific formatting to typevars in type hints. We parse the code to get a list of typevar names, then modify the document node to instead link to `typing.TypeVar`/`ParamSpec`, and tweak the formatting.

This could do a lot of things, not sure if linking to the typevar is exactly what we want, and the CSS is rather temp - visuals is not my expertise.

Here's an [example of the rendering](https://trio--2752.org.readthedocs.build/en/2752/reference-io.html#trio.abc.SendChannel), hover over `SendType`.


docs/source/conf.py
```diff
@@ -53,16 +53,6 @@
...
-    ("py:class", "trio._abc.SendType"),
-    ("py:class", "trio._abc.T"),
-    ("py:obj", "trio._abc.ReceiveType"),
-    ("py:obj", "trio._abc.SendType"),
-    ("py:obj", "trio._abc.T"),
-    ("py:obj", "trio._abc.T_resource"),
-    ("py:class", "trio._threads.T"),
     # why aren't these found in stdlib?
     ("py:class", "types.FrameType"),
     ("py:class", "P.args"),
```
A5rocks, line 58: This and the next 2 should also be unnecesary

docs/source/typevars.py
```diff
@@ -0,0 +1,103 @@
...
+def identify_typevars() -> None:
+    """Record all typevars in trio."""
+    trio_folder = Path(__file__, "..", "..", "..", "trio").resolve()
+    print(trio_folder)
+    for filename in trio_folder.rglob("*.py"):
+        with open(filename, encoding="utf8") as f:
+            for line in f:
+                # A simple regex should be sufficient to find them all, no need to actually parse.
+                match = re.search(
+                    r"^\s*(\w+)\s*=\s*(TypeVar|TypeVarTuple|ParamSpec)\(",
```
A5rocks, line 26: it might be easier to use the string passed into the typevar

docs/source/typevars.py
```diff
@@ -0,0 +1,103 @@
...
+                    name = match.group(1)
+                    typevars_qualified[f'{".".join(relative.parts)}.{name}'] = kind
+                    existing = typevars_named.setdefault(name, kind)
+                    if existing != kind:
+                        print("Mismatch: {} = {}, {}", name, existing, kind)
+
+
+# All our typevars, so we can suppress reference errors for them.
+typevars_qualified = {
+    "_os.PathLike": "typing.TypeVar",
```
A5rocks, line 44: Why?

TeamSpen210: Seems to not be necessary, thought it was at some point to stop this errorring.

docs/source/typevars.py
TeamSpen210: Perhaps it might, but since this is anchored at the start of the string, it'll quickly fail for any non-assignment lines.

docs/source/typevars.py
```diff
@@ -0,0 +1,101 @@
...
+
+from sphinx.addnodes import pending_xref, Element
+from sphinx.application import Sphinx
+from sphinx.environment import BuildEnvironment
+from sphinx.errors import NoUri
+
+
+def identify_typevars() -> None:
+    """Record all typevars in trio."""
+    trio_folder = Path(__file__, "..", "..", "..", "trio").resolve()
```
jakkdl, line 19: could maybe query `git rev-parse --show-toplevel` for the root directory instead of hardcoding this monstrosity, like in
https://github.com/Zac-HD/shed/blob/73ea9159c5cec2c462ce0779457231dc69a79466/src/shed/_cli.py#L26-L34

jakkdl (approved):
Seems somewhat vulnerable to breakage and a bit opaque to fix, but not sure there's a ton that can be done.

docs/source/typevars.py
TeamSpen210: Hmm actually just `trio.__file__` too, already got that imported. Ideally we’d find where Sphinx stores the path, use that…

docs/source/typevars.py
```diff
@@ -0,0 +1,101 @@
...
+    print(trio_folder)
```
jakkdl, line 20: we probably don't want this + the typevars print in the final version

jakkdl:
> I could use the ast module or something, but a regex seemed simpler and faster. Type checkers aren’t going to allow any complicated expressions anyway. Though I need to add support for `typing.TypeVar()` etc, forgot that.

I could whip up a quick AST visitor if you want, I've written tons of those for flake8 plugins, but the regex is probably fine

TeamSpen210:
> > I could use the ast module or something, but a regex seemed simpler and faster. Type checkers aren’t going to allow any complicated expressions anyway. Though I need to add support for `typing.TypeVar()` etc, forgot that.
> 
> I could whip up a quick AST visitor if you want, I've written tons of those for flake8 plugins, but the regex is probably fine

Feel free to if you want, might be more robust. Though there's the problem of it breaking with future Python versions potentially...

jakkdl:
> > > I could use the ast module or something, but a regex seemed simpler and faster. Type checkers aren’t going to allow any complicated expressions anyway. Though I need to add support for `typing.TypeVar()` etc, forgot that.
> > 
> > 
> > I could whip up a quick AST visitor if you want, I've written tons of those for flake8 plugins, but the regex is probably fine
> 
> Feel free to if you want, might be more robust. Though there's the problem of it breaking with future Python versions potentially...

hah, true. It might not actually be more robust

PR #2753, opened 2023-08-14
Add types to `trio._subprocess`

TeamSpen210:
Another one of the remaining modules. During typing I did find two type issues:

* The IO modules might need slightly looser types - `wait_readable()` and similar method should take `int | HasFileno`, not just `int | socket`.
* `subprocess.Popen` takes `str | bytes | os.PathLike` or a `Sequence` of those, but `open_process()` does an `isinstance(..., str)` check so it won't correctly accept those.


TeamSpen210:
Missing coverage is an existing issue, `trio._subprocess_platform.waitid.wait_child_exiting()` doesn't seem to be exercised by any tests.

jakkdl:
> The IO modules might need slightly looser types - `wait_readable()` and similar method should take `int | HasFileno`, not just `int | socket`.

go ahead and change those if you want, I think `int|socket` was just a qualified guess of mine and only looking at usage within the same file
> `subprocess.Popen` takes `str | bytes | os.PathLike` or a `Sequence` of those, but `open_process()` does an `isinstance(..., str)` check so it won't correctly accept those.

I don't see any reason not to make the `isinstance` check wider?

trio/_subprocess.py
```diff
@@ -687,23 +719,28 @@ async def my_deliver_cancel(process):
...
+        chunks: list[bytes | bytearray],
+    ) -> None:
         async with stream:
             async for chunk in stream:
                 chunks.append(chunk)
 
     async with trio.open_nursery() as nursery:
-        proc = await open_process(command, **options)
+        # options needs a complex TypedDict.
+        proc = await open_process(command, **options)  # type: ignore
```
jakkdl, line 743: restrict the `type: ignore`. Actually, we could maybe enable a setting that warns about unrestricted `type: ignore`s

trio/_subprocess.py
```diff
@@ -439,17 +470,18 @@ async def _posix_deliver_cancel(p):
...
-    deliver_cancel=None,
-    task_status=trio.TASK_STATUS_IGNORED,
-    **options,
-):
+    stdin: bytes | bytearray | memoryview | int | HasFileno | None = b"",
+    capture_stdout: bool = False,
+    capture_stderr: bool = False,
+    check: bool = True,
+    deliver_cancel: Callable[[Process], Awaitable[object]] | None = None,
+    task_status: TaskStatus = trio.TASK_STATUS_IGNORED,  # type: ignore[assignment]  # TODO
```
jakkdl, line 482: bunch of bare `# TODO`s without any explanation - but I assume the majority of them are just waiting for #2733

pyproject.toml
```diff
@@ -65,6 +65,7 @@ module = [
     "trio._ki",
     "trio._socket",
     "trio._sync",
+    "trio._subprocess",
```
jakkdl, line 68: if the `subprocess_platform` files are fully typed they can also be added.

trio/_subprocess_platform/waitid.py
```diff
@@ -101,7 +101,7 @@ async def wait_child_exiting(process: "_subprocess.Process") -> None:
     #   process.
 
     if process._wait_for_exit_data is None:
-        process._wait_for_exit_data = event = Event()  # type: ignore
```
jakkdl, line 104: restrict the `type: ignore`

jakkdl (approved):
mostly just a bunch of smallish nits

trio/_subprocess.py
TeamSpen210: Yeah these are just reminders to make `TaskStatus` generic.

trio/_subprocess_platform/waitid.py
TeamSpen210: This is actually removing the type-ignore, it’s no longer required.

jakkdl: oh derp

trio/_subprocess.py
jakkdl: I've merged #2733 so you can go ahead with it now

trio/_subprocess.py
TeamSpen210: Unfortunately, since Unix has an overloaded `open_process` and Windows does not the type code varies depending on platform. Realised I can ignore the unused ignores though :P

A5rocks: Fwiw you can ignore multiple type codes, just use a comma

TeamSpen210: Yep, the issue is that on Windows you get `arg-type`, but on Linux/Mac you get `arg-type` + `call-overload`. So I had to use `type: ignore[arg-type, call-overload, unused-ignore]` to ignore the ignore not being used.

PR #2754, opened 2023-08-14
Add slots to `_Once`

CoolCat467:
This PR adds slots to `ssl`


A5rocks:
BTW mind transferring the slots from `NeedHandshakeError` too?

A5rocks:
Sorry, I haven't really changed on my mind on these being very slightly net negative. I'm going to close this just so it isn't hanging forever... But feel free to reopen if you feel strongly about this. Or anyone else can. (Basically this isn't a rejection of the implementation or technical merits, just I don't think consistency / maintenence / potential breakage is worth a diminishing memory decrease and speedup)

PR #2756, opened 2023-08-15
Type remaining `_highlevel` modules

TeamSpen210:
A bunch of little things that weren't typed yet. I did make one functional change: adding a `length` parameter to `SocketStream.setsockopt()` to match the underlying socket. I believe `trio.serve_listeners()` requires higher kinded types / generic typevar bounds to actually be typed correctly, currently it allows passing a handler that doesn't match the listener's results.


docs/source/conf.py
```diff
@@ -79,6 +81,7 @@
     # these are not defined in https://docs.python.org/3/objects.inv
     ("py:class", "socket.AddressFamily"),
     ("py:class", "socket.SocketKind"),
+    ("py:class", "Buffer"),  # collections.abc.Buffer, in 3.12
```
jakkdl, line 84: huh, I'm a bit surprised it's not available in the inv file given that it's in the online docs https://docs.python.org/3.12/library/collections.abc.html#collections.abc.Buffer. But I guess it's just a matter of waiting for the release

TeamSpen210: It's in the `inv` file, but we're linking to `/3/` which is still 3.11.

jakkdl (approved):

Issue #2750, opened 2023-08-10
Run black+isort on code generated by `gen_exports.py`


jakkdl:
See https://github.com/python-trio/trio/pull/2733#discussion_r1289625533 and https://github.com/python-trio/trio/pull/2743#discussion_r1285299714 and https://github.com/python-trio/trio/pull/2743#discussion_r1285266764

Should maybe also report issues that aren't just waiting for release upstream to https://github.com/berkerpeksag/astor

TeamSpen210:
Would be a nice tidyup, especially since `isort` should collapse multiple imports, so you can get rid of the basic check I did to prevent those. To maximise cross-platformness, might be best to call them via `subprocess.run([sys.executable, "-m", ...])` so we don't require `python` or the executable to be on `PATH`.

jakkdl:
Both of them look to have an API, so we shouldn't need to involve external processes at all

Zac-HD:
You might want to use [`shed`](https://github.com/Zac-HD/shed)?  It includes Black and Isort, ensures that the results are idempotent, does some further cleanups, and the Python API is easier to use than those of the wrapped tools.

jakkdl:
> It'd make more sense to use `shed` on the whole repo - if we just used it here, we might end up in a situation where it produces different results to running black/isort standalone.

I was about to make that comment, but even if it does produce slightly different code the problem would only happen if `shed` output code that doesn't pass black/isort. That in general shouldn't happen - but `shed` ignores project configs iirc, which *could* cause a diff if we were to turn on `--preview` for black, or modify isort settings.

A5rocks:
I think this is completed now?

PR #2759, opened 2023-08-17
Implement #2750: Run `black`+`isort` on `gen_exports` results

TeamSpen210:
This makes `gen_exports` run both linters on the generated files, improving the formatting a little. 

I ran `black` via subprocess because its API doesn't provide a nice way to fetch configuration from `pyproject.toml` etc to maintain consistency, while also processing an in-memory string. Still doesn't remove the redundant parentheses, but oh well.


jakkdl (approved):
Looks great!

TeamSpen210:
Adding this caused all the pypy tests to fail, since we don't even install black/isort there. I just skipped the tests there, the module is an internal tool that only needs to run on one version.

jakkdl:
needs pragma: skip on the exceptions

TeamSpen210:
Hmm actually I could just add tests for those cases - mock out `isort` to produce an error, send in an invalid file to make `black` fail.

jakkdl:
> Hmm actually I could just add tests for those cases - mock out `isort` to produce an error, send in an invalid file to make `black` fail.

sure .. but do we need thorough tests on an internal code generation tool? :p

trio/_tools/gen_exports.py
```diff
@@ -19,22 +21,21 @@
...
-# fmt: off
 from __future__ import annotations
 
 from ._ki import LOCALS_KEY_KI_PROTECTION_ENABLED
 from ._run import GLOBAL_RUN_CONTEXT
 """
 
-FOOTER = """# fmt: on
+FOOTER = """
 """
```
A5rocks, line 39: Why keep the newline?

trio/_tools/gen_exports.py
```diff
@@ -104,6 +105,40 @@ def create_passthrough_args(funcdef: ast.FunctionDef | ast.AsyncFunctionDef) ->
...
 
 
+def run_linters(file: File, source: str) -> str:
+    """Run isort and black on the specified file, returning the new source.
+
+    :raises ValueError: If either failed.
+    """
+    # Black has an undocumented API, but it doesn't easily allow reading configuration from
+    # pyproject.toml, and simultaneously pass in / receive the code as a string.
+    # https://github.com/psf/black/issues/779
```
A5rocks, line 115: FYI I played around with black's API (just using `format_file_contents` with `Mode` which was just a retyped config file; just line length and preview though) but a dev on black told me to use the CLI interface like you're doing instead :^)

trio/_tools/gen_exports.py
```diff
@@ -104,6 +105,40 @@ def create_passthrough_args(funcdef: ast.FunctionDef | ast.AsyncFunctionDef) ->
...
+    # isort does have a public API, makes things easy.
+    try:
+        isort_res = isort.api.sort_code_string(
+            result.stdout,
+            file_path=file.path,
+        )
+    except isort.exceptions.ISortError as exc:
+        print("Failed to run isort!")
+        traceback.print_exception(type(exc), exc, exc.__traceback__)
+        sys.exit(1)
```
A5rocks, line 138: Why can't we re-raise? Or even `raise RuntimeError("Failed to run isort!") from exc`!

A5rocks (commented):
Couple of nitpicks, looks good otherwise.

trio/_tools/gen_exports.py
TeamSpen210: I don't think that results in a nonzero exit code, I wanted to make sure CI etc fails.

trio/_tools/gen_exports.py
TeamSpen210: Good question.

trio/_tools/gen_exports.py
A5rocks: FWIW it does, it seems!

```py
a5@hostname:/tmp$ python3 t.py
  File "t.py", line 1, in <module>
    raise Exception("hi!")
Exception: hi!
a5@hostname:/tmp$ echo $?
1
```
(at least on my Ubuntu WSL env...)

A5rocks:
builds.sr.ht is back!?!

jakkdl:
I'll merge this unless somebody thinks otherwise, re-raising the exception instead of printing traceback seems like a minor enough problem not worth stalling over

PR #2761, opened 2023-08-17
add types to io_windows

jakkdl:
I hit a wall implementing this and I found it a hassle not being on windows, so would be lovely if anybody would take it over


trio/_core/_io_windows.py
```diff
@@ -185,13 +188,15 @@ class CKeys(enum.IntEnum):
     USER_DEFINED = 4  # and above
 
 
-def _check(success):
+def _check(success: bool) -> Literal[True]:
```
TeamSpen210, line 191: At some point I figured out that this needs to be a TypeVar, it's used on return values to raise when they indicate errors.

A5rocks:
I can maybe try to work on this (as I mainly use Windows) -- I would use GitHub's assignment thing that I remember it having but I can't find the buttons on mobile :(

TeamSpen210:
Pushed an attempt to type the CFFI parts, not sure if we want to go that route (or just leave them dynamic)? With a bunch of `NewType` we could check all the types.

A5rocks:
I don't understand the pypy failure but that does seem to be non-transient

A5rocks:
Looks like pypy on Windows is broken on master too, but since https://github.com/python-trio/trio/commit/0c108f94da7fc82008f5e5a034f48f4a7854269e so I think it's an upstream issue.

trio/_core/_io_windows.py
```diff
@@ -292,130 +405,37 @@ def _afd_helper_handle():
...
 
 
 class WindowsIOManager:
-    def __init__(self):
+    def __init__(self) -> None:
         # If this method raises an exception, then __del__ could run on a
         # half-initialized object. So we initialize everything that __del__
         # touches to safe values up front, before we do anything that can
         # fail.
         self._iocp = None
```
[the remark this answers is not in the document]
jakkdl: Could also add a `self._is_initialized` variable that's initially set to `False`, and set to `True` after `self._iocp` is initialized.

trio/_core/_windows_cffi.py
```diff
@@ -327,13 +517,14 @@ def raise_winerror(
     filename: str | None = None,
     filename2: str | None = None,
 ) -> NoReturn:
+    assert sys.platform == "win32"  # this should work starting next mypy release
```
jakkdl, line 520: oh, TIL! I was struggling with this in #2774

trio/_core/_generated_io_windows.py
```diff
@@ -8,77 +8,87 @@
...
+    from typing_extensions import Buffer
+
+    from ._unbounded_queue import UnboundedQueue
 import sys
 
 assert not TYPE_CHECKING or sys.platform=="win32"
 
 
-async def wait_readable(sock):
+async def wait_readable(sock: (socket.socket | int)) ->None:
```
jakkdl, line 24: In https://github.com/python-trio/trio/pull/2774/files#diff-175d00891570043e8c2954032ac59ba7aca294153052943cb309e07460969d74 I'm changing these to accept `_HasFileNo` instead of `socket.socket` - might as well do that directly here (and all other methods accepting `socket.socket`)

trio/_core/_io_windows.py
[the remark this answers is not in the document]
A5rocks: > Since `iocp` can only be `None` if `__init__` fails (so in `__del__`), it might be better to just cast/type-ignore the none assignment, so we avoid asserts everywhere.

FWIW it can also be `None` after `.close` but IDK if the object still exists then.

trio/_core/_windows_cffi.py
A5rocks: Yea, I believe this falls under https://github.com/python/mypy/pull/15386

jakkdl: yeah I suspect that's a different fix

jakkdl:
Ah, thought there was something wrong with #2750 but it's rather that this PR is pushing unformatted code - i.e. need to rerun `gen_exports`

jakkdl:
I can't approve a PR I created, but LGTM :heavy_check_mark:

CoolCat467 (approved):
Looks good to me

trio/_core/_windows_cffi.py
```diff
@@ -222,12 +222,196 @@
...
+    Offset: int
+    OffsetHigh: int
+
+
+class _DummyUnion(Protocol):
+    DUMMYSTRUCTNAME: _DummyStruct
+    Pointer: object
+
+
+class _Overlapped(Protocol):
```
TeamSpen210, line 405: Might want to just make this public, if it's being imported elsewhere.

trio/_core/_io_windows.py
```diff
@@ -724,12 +748,14 @@ def notify_closing(self, handle):
...
     @_public
-    def register_with_iocp(self, handle):
+    def register_with_iocp(self, handle: int | CData) -> None:
         self._register_with_iocp(handle, CKeys.WAIT_OVERLAPPED)
 
     @_public
-    async def wait_overlapped(self, handle, lpOverlapped):
-        handle = _handle(handle)
+    async def wait_overlapped(
+        self, handle_: int | CData, lpOverlapped: CData | int
```
TeamSpen210, line 756: Instead of casting to/from `CData`, maybe we should type it as our `Overlapped` / `Handle` types, so this can be statically checked? Or just add those to the union alongside `CData`, to be more loose.

trio/_core/_windows_cffi.py
```diff
@@ -327,13 +516,14 @@ def raise_winerror(
     filename: str | None = None,
     filename2: str | None = None,
 ) -> NoReturn:
+    # assert sys.platform == "win32"  # TODO: make this work in MyPy
     if winerror is None:
-        err = ffi.getwinerror()
+        err = ffi.getwinerror()  # type: ignore[attr-defined,unused-ignore]
```
TeamSpen210, line 521: May want a comment in this func to explain that `getwinerror()` is only defined on windows, hence the type-ignore.

trio/_tools/gen_exports.py
```diff
@@ -310,6 +315,17 @@ def main() -> None:  # pragma: no cover
 
 """
 
+IMPORTS_WINDOWS = """\
+from typing import TYPE_CHECKING, ContextManager
+
+if TYPE_CHECKING:
+    from .._file_io import _HasFileNo
```
TeamSpen210, line 322: Seems iffy to be importing a private name from a different module, might want to make this also public.

TeamSpen210 (approved):
Looks good, only have a few quibbles.

trio/_core/_windows_cffi.py
A5rocks: Yeah I'm not sure how to think about `Protocol` publicness as they are meant to be purely structural and any use of them could be replaced with a copy-pasted copy

trio/_core/_io_windows.py
A5rocks: I would push out `lpOverlapped` casting to a boundary... but looking at the code, this is public :(

The type casting can probably be improved though.

A5rocks:
@TeamSpen210 just wanted to make sure you saw my responses and think it's still fine!

(mainly about `Protocol` public-ness...)

TeamSpen210:
That does make sense, it’s of course easier to make something public than the reverse.

PR #2762, opened 2023-08-17
Add missing docstrings

jakkdl:
I started trying to write these ... and while I could maybe figure out what the functions are supposed to do it would be lovely if somebody else chipped in with suggestions (@A5rocks? @njsmith?) 

This also made me notice that there were some problems in #2753 with `open_process` and `run_process`, I noted them here but they should probably be fixed by @TeamSpen210 in a separate PR


jakkdl:
This is now the only thing remaining for `pyright --verifytypes` to be run without any issues, and make it possible to remove the endlessly infuriating `trio/_tests/verify_types_*.json` files. Would love somebody to take a stab at it, if any suggested docstrings are incorrect somebody will probably come along and correct them in a review anyhow.

CoolCat467:
> why are we getting that warning (is something weirdly configured?)

I think this might be because of differences in the installed version of ruff and the pre-commit ruff hook version, not completely sure to be honest though.

> pre-commit.ci is failing ... and ruff is saying there are 7 problems, but not saying what they are. @CoolCat467 something needs to be changed here

This is because it's attempting to autofix the errors, and right now the config says it doesn't need to comment about what fixes it applies.
`--fix-only` says:
> Fix any fixable lint violations, but don't report on leftover violations. Implies `--fix`.
So I think the "leftover violations" are the warnings, or maybe we need to specifically say `--no-preview`

jakkdl:
> > why are we getting that warning (is something weirdly configured?)
> 
> I think this might be because of differences in the installed version of ruff and the pre-commit ruff hook version, not completely sure to be honest though. At least for me, I always have to run the gen_exports script with python3.11 instead of just running it directly, because my system defaults to python3.9, and the version of packages used in the different python installations is different. Not sure if this applies but I thought I'd mention it.

Ah shit, this is my bad. I had a `ruff.toml` lying around from when I was running ruff locally myself before it was added to `pyproject.toml`.


> > pre-commit.ci is failing ... and ruff is saying there are 7 problems, but not saying what they are. @CoolCat467 something needs to be changed here
> 
> This is because it's attempting to autofix the errors, and right now the config says it doesn't need to comment about what fixes it applies. `--fix-only` says:
> 
> > Fix any fixable lint violations, but don't report on leftover violations. Implies `--fix`.
> 
> So I think the "leftover violations" are the warnings, or maybe we need to specifically say `--no-preview`

Yeah, it's autofixing instead of reporting the errors but we've disabled autofix in the pre-commit CI and it's failing because files are getting changed. So we either need to disable autofix for ruff when it's run through pre-commit, or for pre-commit to display the changes to the files that caused it to fail.

Re: "@CoolCat467
[Success should still be false for warnings](https://github.com/python-trio/trio/pull/2762/commits/c0875c3c568fd91ac1ba8f8105cf644e19b4d1f8)"
my change was specifically to not have it fail when giving warnings, but given I've resolved that issue we can probably revert to what the file was like originally.

TeamSpen210:
I think I got the overloads from the `trio_typing` stubs. The difference is that on Linux/Mac, you can pass a string command only with 'shell=True`, while on Windows you can pass either a string or list of strings regardless of the option. You can't use a single overload for a function, so I had to do the `_open_process` business. Looking at the typeshed stubs they don't make this distinction for `Popen` etc, maybe we should too and just simplify the mess.

jakkdl:
> I think I got the overloads from the `trio_typing` stubs. The difference is that on Linux/Mac, you can pass a string command only with `shell=True`, while on Windows you can pass either a string or list of strings regardless of the option. You can't use a single overload for a function, so I had to do the `_open_process` business. Looking at the typeshed stubs they don't make this distinction for `Popen` etc, maybe we should too and just simplify the mess.

Simplifying the mess sounds like a big win to me, cause this is indeed really really messy

Zac-HD (approved):
Looks great to me - thanks, @jakkdl!

CoolCat467 (approved):
I think that for the information we have these look great

trio/_core/_io_epoll.py
```diff
@@ -310,14 +310,69 @@ def abort(_: RaiseCancelT) -> Abort:
...
+        or else an object with a ``.fileno()`` method which returns an
+        integer file descriptor. Any kind of file descriptor can be passed,
+        though the exact semantics will depend on your kernel. For example,
+        this probably won't do anything useful for on-disk files.
+
+        On Windows systems, ``obj`` must either be an integer ``SOCKET``
+        handle, or else an object with a ``.fileno()`` method which returns
+        an integer ``SOCKET`` handle. File descriptors aren't supported,
+        and neither are handles that refer to anything besides a
+        ``SOCKET``.
```
A5rocks, line 326:
```suggestion
        On Unix systems, ``fd`` must either be an integer file descriptor,
        or else an object with a ``.fileno()`` method which returns an
        integer file descriptor. Any kind of file descriptor can be passed,
        though the exact semantics will depend on your kernel. For example,
        this probably won't do anything useful for on-disk files.

        On Windows systems, ``fd`` must either be an integer ``SOCKET``
        handle, or else an object with a ``.fileno()`` method which returns
        an integer ``SOCKET`` handle. File descriptors aren't supported,
        and neither are handles that refer to anything besides a
        ``SOCKET``.
```

trio/_core/_io_epoll.py
```diff
@@ -310,14 +310,69 @@ def abort(_: RaiseCancelT) -> Abort:
...
+            if another task calls :func:`notify_closing` while this
+            function is still working.
+        """
         await self._epoll_wait(fd, "read_task")
 
     @_public
     async def wait_writable(self, fd: int | _HasFileNo) -> None:
+        """Block until the kernel reports that the given object is writable.
+
+        See `wait_readable` for the definition of ``obj``.
```
A5rocks, line 341:
```suggestion
        See `wait_readable` for the definition of ``fd``.
```

trio/_core/_io_epoll.py
```diff
@@ -310,14 +310,69 @@ def abort(_: RaiseCancelT) -> Abort:
 
     @_public
     async def wait_readable(self, fd: int | _HasFileNo) -> None:
+        """
+        Block until the kernel reports that the given object is readable.
```
A5rocks, line 314:
```suggestion
        """Block until the kernel reports that the given object is readable.
```

trio/_core/_io_epoll.py
```diff
@@ -310,14 +310,69 @@ def abort(_: RaiseCancelT) -> Abort:
...
+            function is still working.
+        """
         await self._epoll_wait(fd, "write_task")
 
     @_public
     def notify_closing(self, fd: int | _HasFileNo) -> None:
+        """Call this before closing a file descriptor (on Unix) or socket (on
+        Windows). This will cause any `wait_readable` or `wait_writable`
+        calls on the given object to immediately wake up and raise
+        `~trio.ClosedResourceError`.
```
A5rocks, line 357: First line should be a description but not exactly sure what would work.

trio/_core/_io_kqueue.py
```diff
@@ -181,14 +193,68 @@ def abort(_: RaiseCancelT) -> Abort:
...
+        or else an object with a ``.fileno()`` method which returns an
+        integer file descriptor. Any kind of file descriptor can be passed,
+        though the exact semantics will depend on your kernel. For example,
+        this probably won't do anything useful for on-disk files.
+
+        On Windows systems, ``obj`` must either be an integer ``SOCKET``
+        handle, or else an object with a ``.fileno()`` method which returns
+        an integer ``SOCKET`` handle. File descriptors aren't supported,
+        and neither are handles that refer to anything besides a
+        ``SOCKET``.
```
A5rocks, line 208:
```suggestion
        On Unix systems, ``fd`` must either be an integer file descriptor,
        or else an object with a ``.fileno()`` method which returns an
        integer file descriptor. Any kind of file descriptor can be passed,
        though the exact semantics will depend on your kernel. For example,
        this probably won't do anything useful for on-disk files.

        On Windows systems, ``fd`` must either be an integer ``SOCKET``
        handle, or else an object with a ``.fileno()`` method which returns
        an integer ``SOCKET`` handle. File descriptors aren't supported,
        and neither are handles that refer to anything besides a
        ``SOCKET``.
```

trio/_core/_io_kqueue.py
```diff
@@ -181,14 +193,68 @@ def abort(_: RaiseCancelT) -> Abort:
...
+            if another task calls :func:`notify_closing` while this
+            function is still working.
+        """
         await self._wait_common(fd, select.KQ_FILTER_READ)
 
     @_public
     async def wait_writable(self, fd: int | _HasFileNo) -> None:
+        """Block until the kernel reports that the given object is writable.
+
+        See `wait_readable` for the definition of ``obj``.
```
A5rocks, line 223:
```suggestion
        See `wait_readable` for the definition of ``fd``.
```

trio/_core/_io_kqueue.py
```diff
@@ -181,14 +193,68 @@ def abort(_: RaiseCancelT) -> Abort:
...
+            function is still working.
+        """
         await self._wait_common(fd, select.KQ_FILTER_WRITE)
 
     @_public
     def notify_closing(self, fd: int | _HasFileNo) -> None:
+        """Call this before closing a file descriptor (on Unix) or socket (on
+        Windows). This will cause any `wait_readable` or `wait_writable`
+        calls on the given object to immediately wake up and raise
+        `~trio.ClosedResourceError`.
```
A5rocks, line 239: I think there should be a summary here as I commented elsewhere.

trio/_core/_io_windows.py
```diff
@@ -728,14 +728,68 @@ def abort_fn(_: RaiseCancelT) -> Abort:
...
+        or else an object with a ``.fileno()`` method which returns an
+        integer file descriptor. Any kind of file descriptor can be passed,
+        though the exact semantics will depend on your kernel. For example,
+        this probably won't do anything useful for on-disk files.
+
+        On Windows systems, ``obj`` must either be an integer ``SOCKET``
+        handle, or else an object with a ``.fileno()`` method which returns
+        an integer ``SOCKET`` handle. File descriptors aren't supported,
+        and neither are handles that refer to anything besides a
+        ``SOCKET``.
```
A5rocks, line 743:
```suggestion
        On Unix systems, ``sock`` must either be an integer file descriptor,
        or else an object with a ``.fileno()`` method which returns an
        integer file descriptor. Any kind of file descriptor can be passed,
        though the exact semantics will depend on your kernel. For example,
        this probably won't do anything useful for on-disk files.

        On Windows systems, ``sock`` must either be an integer ``SOCKET``
        handle, or else an object with a ``.fileno()`` method which returns
        an integer ``SOCKET`` handle. File descriptors aren't supported,
        and neither are handles that refer to anything besides a
        ``SOCKET``.
```

trio/_core/_io_windows.py
```diff
@@ -728,14 +728,68 @@ def abort_fn(_: RaiseCancelT) -> Abort:
...
+            if another task calls :func:`notify_closing` while this
+            function is still working.
+        """
         await self._afd_poll(sock, "read_task")
 
     @_public
     async def wait_writable(self, sock: _HasFileNo | int) -> None:
+        """Block until the kernel reports that the given object is writable.
+
+        See `wait_readable` for the definition of ``obj``.
```
A5rocks, line 758:
```suggestion
        See `wait_readable` for the definition of ``sock``.
```

trio/_core/_io_windows.py
```diff
@@ -728,14 +728,68 @@ def abort_fn(_: RaiseCancelT) -> Abort:
...
+            function is still working.
+        """
         await self._afd_poll(sock, "write_task")
 
     @_public
     def notify_closing(self, handle: Handle | int | _HasFileNo) -> None:
+        """Call this before closing a file descriptor (on Unix) or socket (on
+        Windows). This will cause any `wait_readable` or `wait_writable`
+        calls on the given object to immediately wake up and raise
+        `~trio.ClosedResourceError`.
```
A5rocks, line 774: Also missing summary.

trio/_subprocess.py
```diff
@@ -815,6 +815,56 @@ async def open_process(
...
+                  attached to whatever the ``stdout`` option says to attach it to.
+              **options: Other :ref:`general subprocess options <subprocess-options>`
+                  are also accepted.
+
+            Returns:
+              A new `trio.Process` object.
+
+            Raises:
+              OSError: if the process spawning fails, for example because the
+                 specified command could not be found.
```
A5rocks, line 865: Is this docstring format correct? You use the plain sphinx format elsewhere in this PR.

trio/_subprocess.py
```diff
@@ -835,10 +885,203 @@ async def run_process(
...
+            Raises:
+              UnicodeError: if ``stdin`` is specified as a Unicode string, rather
+                  than bytes
+              ValueError: if multiple redirections are specified for the same
+                  stream, e.g., both ``capture_stdout=True`` and
+                  ``stdout=subprocess.DEVNULL``
+              subprocess.CalledProcessError: if ``check=False`` is not passed
+                  and the process exits with a nonzero exit status
+              OSError: if an error is encountered starting or communicating with
+                  the process
```
A5rocks, line 1070: Again note about docstring format

A5rocks (changes requested):
General formatting things, content looks fine.

trio/_subprocess.py
jakkdl: This is a copy-paste of the docstring at line 346 for `_open_process` - but maybe that one is incorrectly formatted? Or all the other ones I copy-pasted from the docs are?

trio/_subprocess.py
jakkdl: same thing - copypasted from `run_process` earlier in the file. (and yes duplicate docstrings is incredibly ugly and bad... but I don't know how to improve @TeamSpen210's magic to avoid needing to and have it work both on runtime, in static analysis, and on all platforms.)

trio/_core/_io_epoll.py
jakkdl: > Notify waiters of the given object that it will be closed.

works?

CoolCat467 (approved):

trio/_core/_io_epoll.py
A5rocks: Yeah

trio/_subprocess.py
A5rocks: TBH as long as sphinx renders this fine then whatever. And it does seem to.

trio/_subprocess.py
A5rocks: TBH sphinx renders this fine so whatever.

A5rocks (approved):
Looks fine, if things don't like multiple docstring formats then we can fix them then. Not worth the time ATM I think.

(also the summary looks fine, so maybe add that before merging ^^)

jakkdl:
Sorry yeah, I added them :)

Time to get rid of the jsons!! :rocket:

PR #2763, opened 2023-08-17
Fix various files so the entire trio/ directory can be type-checked

jakkdl:
This is a squash of #2721 to avoid having to navigate a huge and messy PR. It consists of a ton of small fixes or `type: ignores` to any file that raised errors when running `mypy trio/`, and it also transitions `pyproject.toml` from being a list of files to check strictly, to an ignore-list with files to disable strict checks for. This doubles as a TODO-list, and it's easy to explode wildcards, or remove a disabled check, or similar.

The test changes has caught several small typing errors introduced in various PRs that haven't been checked in CI, or when files have been forgotten to get added.

All the changes should be non-controversial, if anything is up for debate I'll just mark it with a comment and/or revert it, and add it to #2734 or directly split out to a separate PR

### DIFF
--- a/trio/_core/_tests/test_io.py
+++ b/trio/_core/_tests/test_io.py
@@ -1,6 +1,9 @@
+from __future__ import annotations
+
 import random
 import socket as stdlib_socket
 from contextlib import suppress
+from typing import Callable
 
 import pytest
 
@@ -47,15 +50,15 @@ def using_fileno(fn):
     return fileno_wrapper
 
 
-wait_readable_options = [trio.lowlevel.wait_readable]
-wait_writable_options = [trio.lowlevel.wait_writable]
-notify_closing_options = [trio.lowlevel.notify_closing]
+wait_readable_options: list[Callable] = [trio.lowlevel.wait_readable]
+wait_writable_options: list[Callable] = [trio.lowlevel.wait_writable]
+notify_closing_options: list[Callable] = [trio.lowlevel.notify_closing]
 
-for options_list in [
+for options_list in (
     wait_readable_options,
     wait_writable_options,
     notify_closing_options,
-]:
+):
     options_list += [using_fileno(f) for f in options_list]
 
 # Decorators that feed in different settings for wait_readable / wait_writable
--- a/trio/_core/_tests/test_io.py
+++ b/trio/_core/_tests/test_io.py
@@ -2,8 +2,8 @@ from __future__ import annotations
 
 import random
 import socket as stdlib_socket
+from collections.abc import Callable
 from contextlib import suppress
-from typing import Callable
 
 import pytest
 
--- a/trio/_core/_tests/test_multierror_scripts/apport_excepthook.py
+++ b/trio/_core/_tests/test_multierror_scripts/apport_excepthook.py
@@ -3,13 +3,13 @@
 # make sure it's on sys.path.
 import sys
 
-import _common
+import _common  # isort: split
 
 sys.path.append("/usr/lib/python3/dist-packages")
 import apport_python_hook
 
 apport_python_hook.install()
 
-import trio
+from trio._core._multierror import MultiError  # Bypass deprecation warnings
 
-raise trio.MultiError([KeyError("key_error"), ValueError("value_error")])  # type: ignore[attr-defined]
+raise MultiError([KeyError("key_error"), ValueError("value_error")])
--- a/trio/_core/_tests/test_multierror_scripts/apport_excepthook.py
+++ b/trio/_core/_tests/test_multierror_scripts/apport_excepthook.py
@@ -12,4 +12,4 @@ apport_python_hook.install()
 
 import trio
 
-raise trio.MultiError([KeyError("key_error"), ValueError("value_error")])
+raise trio.MultiError([KeyError("key_error"), ValueError("value_error")])  # type: ignore[attr-defined]
--- a/trio/_tests/test_contextvars.py
+++ b/trio/_tests/test_contextvars.py
@@ -2,7 +2,9 @@ import contextvars
 
 from .. import _core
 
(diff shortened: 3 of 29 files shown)


trio/testing/_fake_net.py
jakkdl, line 1: without (possibly a subset of) the changes to _fake_net:
```
trio/testing/_fake_net.py:107: error: Signature of "socket" incompatible with supertype "SocketFactory"  [override]
trio/testing/_fake_net.py:107: note:      Superclass:
trio/testing/_fake_net.py:107: note:          def socket(self, family: Union[AddressFamily, int, None] = ..., type: Union[SocketKind, int, None] = ..., proto: Optional[int] = ...) -> _SocketType
trio/testing/_fake_net.py:107: note:      Subclass:
trio/testing/_fake_net.py:107: note:          def socket(self, family: int, type: int, proto: int) -> FakeSocket
trio/testing/_fake_net.py:116: error: Argument 1 of "getaddrinfo" is incompatible with supertype "HostnameResolver"; supertype defines the argument type as "Union[bytes, str, None]"  [override]
trio/testing/_fake_net.py:116: note: This violates the Liskov substitution principle
trio/testing/_fake_net.py:116: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
trio/testing/_fake_net.py:116: error: Argument 2 of "getaddrinfo" is incompatible with supertype "HostnameResolver"; supertype defines the argument type as "Union[bytes, str, int, None]"  [override]
trio/testing/_fake_net.py:131: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
trio/testing/_fake_net.py:179: error: Need type annotation for "_packet_sender"  [var-annotated]
trio/testing/_fake_net.py:179: error: Need type annotation for "_packet_receiver"  [var-annotated]
Found 5 errors in 1 file (checked 129 source files)
```

trio/testing/_fake_net.py
```diff
@@ -176,7 +193,7 @@ def __init__(self, fake_net: FakeNet, family: int, type: int, proto: int):
 
         self._closed = False
 
-        self._packet_sender, self._packet_receiver = trio.open_memory_channel(
+        self._packet_sender, self._packet_receiver = trio.open_memory_channel[object](
```
TeamSpen210, line 196: Using object here (and the other similar calls) isn't really right, it'll cause problems later when we type these functions further?

TeamSpen210 (approved):

trio/_core/_tests/test_io.py
```diff
@@ -1,6 +1,9 @@
+from __future__ import annotations
+
 import random
 import socket as stdlib_socket
 from contextlib import suppress
+from typing import Callable
```
CoolCat467, line 6: It's better to do `from collections.abc import Callable`

trio/_core/_tests/test_multierror_scripts/apport_excepthook.py
```diff
@@ -12,4 +12,4 @@
 
 import trio
 
-raise trio.MultiError([KeyError("key_error"), ValueError("value_error")])
+raise trio.MultiError([KeyError("key_error"), ValueError("value_error")])  # type: ignore[attr-defined]
```
CoolCat467, line 15: For this and the similar thing in some similar tests, I am curious what the error is in particular. If you could add a comment after the ignore about why, that would be handy.

trio/_tests/test_contextvars.py
```diff
@@ -18,7 +22,7 @@ async def child():
     assert record == ["main"]
 
 
-async def test_contextvars_set():
+async def test_contextvars_set() -> None:
```
CoolCat467, line 25: Probably not super necessary, but internal function `child` is missing it's annotation as well.

trio/_tests/test_exports.py
```diff
@@ -301,7 +303,7 @@ def lookup_symbol(symbol):
             with mod_cache.open() as f:
                 return json.loads(f.read())["names"][name]
 
-    errors: dict[str, object] = {}
+    errors: Dict[str, object] = {}
```
CoolCat467, line 306: Why use `Dict` when we have the `from __future__ import annotations` block at the top?

trio/_tests/test_exports.py
```diff
@@ -503,7 +505,7 @@ def test_classes_are_final():
                 continue
             # These are classes that are conceptually abstract, but
             # inspect.isabstract returns False for boring reasons.
-            if class_ in {trio.abc.Instrument, trio.socket.SocketType}:
+            if class_ in (trio.abc.Instrument, trio.socket.SocketType):
```
CoolCat467, line 508: I thought `set.__contains__` was slightly more efficient than tuple's, but I might be wrong

CoolCat467 (changes requested):

trio/_core/_tests/test_multierror_scripts/apport_excepthook.py
TeamSpen210: Since `MultiError` is deprecated, we hid it from type checkers to make them error out on any code that uses it.

trio/_tests/test_exports.py
TeamSpen210: It would be, but for sets mypy evaluates the type here as something like `set[type[object]]`, whereas for tuples it can be `tuple[type[trio.abc.Instrument], type[trio.socket.SocketType]]`, so it doesn't error. Might be even better to just unroll the two compares.

CoolCat467: If it's helping the type checker understand I think it's fine then

CoolCat467 (approved):

jakkdl:
merging this one so any additional minor problem in other typing PRs can be caught, gonna lead to a bunch of merge conflicts in `pyproject.toml` either way

PR #2767, opened 2023-08-21
check type completeness on all platforms

jakkdl:
Fully type complete, other than docstrings (#2762), on both Mac and Linux! :tada: 
Windows still needs works (#2761), but otherwise we're pretty much there wrt to the public API :rocket: :rocket:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,5 +2,3 @@
 trio/_core/_generated*    linguist-generated=true
 # Treat generated files as binary in git diff
 trio/_core/_generated*    -diff
-# don't merge the generated json file, let the user (script) handle it
-trio/_tests/verify_types.json merge=binary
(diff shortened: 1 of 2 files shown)


jakkdl:
Not bothering with forcing binary merges on `verify_types*.json` as the diffs should at this point be small and doable to update manually if somebody wants to - and they'll be gone soon too.

trio/_tests/verify_types_darwin.json
```diff
@@ -0,0 +1,221 @@
...
+      "withKnownType": 682,
+      "withUnknownType": 0
+    },
+    "packageName": "trio",
+    "symbols": [
+      {
+        "category": "function",
+        "diagnostics": [
+          {
+            "file": "/home/h/Git/trio/typing_improvements/.tox/verifytypes/lib/python3.8/site-packages/trio/_core/__init__.py",
```
A5rocks, line 58: By the way! Maybe you missed this

PR #2769, opened 2023-08-21
Fix pyright output

A5rocks:
cc @jakkdl hopefully this still accomplishes the goals you wanted when you originally made all the diagnostics happen.


jakkdl:
yeah this is great, thanks for the quick fix!

TeamSpen210 (approved):

jakkdl:
I think I must've broken `check.sh`, as otherwise it should've given an error when the path differed in CI

jakkdl:
oh nvm, it turned up in #2768. Not sure why it didn't get triggered in #2767 though

PR #2771, opened 2023-08-24
Add typing for some of the tests

CoolCat467:
Work in progress, adding typing for multiple tests
`test_channel` had an error, was awaiting a synchronous function


jakkdl:
You planning on pulling these out into separate PRs, or do a huge chunky one? If planning on doing this one wholesale it probably shouldn't get much larger

CoolCat467:
> You planning on pulling these out into separate PRs, or do a huge chunky one? If planning on doing this one wholesale it probably shouldn't get much larger

Got it

trio/_core/_tests/test_asyncgen.py
```diff
@@ -48,7 +52,7 @@ async def async_main():
         assert collected.pop() == "abandoned"
 
         # aclosing() ensures it's cleaned up at point of use
-        async with contextlib.aclosing(example("exhausted 1")) as aiter:
+        async with contextlib.aclosing(example("exhausted 1")) as aiter:  # type: ignore[attr-defined]  # skipif makes sure we are at least 3.10
```
jakkdl, line 55: It's possible to use https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks
though unfortunately you can't just plop an `assert sys.version_info > 3.X` in the function body (is it possible to do `if sys.version_info > 3.x; return`?). If it requires indenting the whole test with a redundant check it's probably not worth it.
The code then also won't be type-checked in CI at all since we're specifying the version, so maybe net loss overall.

trio/_core/_tests/test_asyncgen.py
```diff
@@ -242,13 +249,13 @@ async def step_outside_async_context(aiter):
     # NB: the strangeness with aiter being an attribute of abort_fn is
     # to make it as easy as possible to ensure we don't hang onto a
     # reference to aiter inside the guts of the run loop.
-    def abort_fn(_):
+    def abort_fn(_: object) -> _core.Abort:
         with pytest.raises(StopIteration, match="42"):
-            abort_fn.aiter.asend(None).send(None)
-        del abort_fn.aiter
+            abort_fn.aiter.asend(None).send(None)  # type: ignore[attr-defined]  # Callables don't have attribute "aiter"
+        del abort_fn.aiter  # type: ignore[attr-defined]
```
jakkdl, line 255: why is the parameter `object` instead of `Callable[..., object]`?

trio/_tests/test_channel.py
```diff
@@ -45,26 +49,26 @@ async def test_channel():
     with pytest.raises(trio.ClosedResourceError):
         await r.receive()
     with pytest.raises(trio.ClosedResourceError):
-        await r.receive_nowait()
+        r.receive_nowait()
     await r.aclose()
```
jakkdl, line 53: why the change of functionality?

trio/_tests/test_deprecate.py
```diff
@@ -11,23 +14,32 @@
...
+
 
 @pytest.fixture
-def recwarn_always(recwarn):
+def recwarn_always(recwarn: object) -> object:
     warnings.simplefilter("always")
     # ResourceWarnings about unclosed sockets can occur nondeterministically
     # (during GC) which throws off the tests in this file
     warnings.simplefilter("ignore", ResourceWarning)
     return recwarn
```
jakkdl, line 27: idk if it matters, but could be `(recwarn: T) -> T:`

jakkdl (approved):
Looks good overall, most of it very straightforward

trio/_core/_tests/test_asyncgen.py
CoolCat467: Because it's unused in the function body

trio/_tests/test_channel.py
CoolCat467: Because mypy said that `receive_nowait` is not an awaitable function and raises an error if attempted

trio/_tests/test_deprecate.py
CoolCat467: Changed in [2a0a355](https://github.com/python-trio/trio/pull/2771/commits/2a0a355168b2ab542cdd85161b31bf21aaf7164b)

trio/_core/_tests/test_asyncgen.py
CoolCat467: Changed in 9c09d80

trio/_core/_tests/test_asyncgen.py
CoolCat467: Changed in 9c09d80

TeamSpen210: It might be easier to just avoid using `aclosing()`, it's a trivial function to inline:
```py
aiter = example("exhausted 1")
try:
    assert 42 == await aiter.asend(None)
finally:
    await aiter.close()
```
Or import it from `async_generator`.

CoolCat467: Inlined in 5da3819

trio/_core/_tests/test_asyncgen.py
```diff
@@ -60,9 +64,12 @@ async def async_main():
         gc_collect_harder()
 
         # No problems saving the geniter when using either of these patterns
-        async with contextlib.aclosing(example("exhausted 3")) as aiter:
-            saved.append(aiter)
+        aiter = example("exhausted 3")
+        saved.append(aiter)
```
TeamSpen210, line 68: `saved.append()` should be inside the `try:` to preserve original semantics, though in practice that line's unlikely to raise exceptions.

trio/_core/_tests/test_asyncgen.py
```diff
@@ -85,10 +92,10 @@ async def async_main():
         assert agen.ag_frame is None  # all should now be exhausted
 
 
-async def test_asyncgen_throws_during_finalization(caplog):
+async def test_asyncgen_throws_during_finalization(caplog) -> None:
```
TeamSpen210, line 95: You can annotate the fixture with `pytest.LogCaptureFixture`.

trio/_core/_tests/test_asyncgen.py
```diff
@@ -242,13 +252,13 @@ async def step_outside_async_context(aiter):
     # NB: the strangeness with aiter being an attribute of abort_fn is
     # to make it as easy as possible to ensure we don't hang onto a
     # reference to aiter inside the guts of the run loop.
-    def abort_fn(_):
+    def abort_fn(_: Callable[..., object]) -> _core.Abort:
```
TeamSpen210, line 255: This should probably be `RaiseCancelT`.

trio/_core/_tests/test_asyncgen.py
```diff
@@ -257,16 +267,16 @@ def abort_fn(_):
 
 
 @pytest.mark.skipif(buggy_pypy_asyncgens, reason="pypy 7.2.0 is buggy")
-async def test_fallback_when_no_hook_claims_it(capsys):
```
TeamSpen210, line 260: `capsys` is [`pytest.CaptureFixture`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.capture.capsys)

trio/_core/_tests/test_windows.py
```diff
@@ -216,7 +220,7 @@ async def test_too_late_to_cancel():
         assert target[:6] == b"test2\n"
 
 
-def test_lsp_that_hooks_select_gives_good_error(monkeypatch):
+def test_lsp_that_hooks_select_gives_good_error(monkeypatch) -> None:
```
TeamSpen210, line 223: This one is [pytest.MonkeyPatch](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch).

trio/_tests/test_deprecate.py
```diff
@@ -11,23 +14,35 @@
...
+if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
+
+T = TypeVar("T")
+
 
 @pytest.fixture
-def recwarn_always(recwarn):
+def recwarn_always(recwarn: T) -> T:
```
TeamSpen210, line 25: Actually, `recwarn` is [pytest.WarningsRecorder](https://docs.pytest.org/en/stable/reference/reference.html#pytest.recwarn.recwarn), so we can drop the typevar and alias.

TeamSpen210 (changes requested):
A few things I noticed - PyTest does publicly expose the classes that fixtures produce, so we can just annotate with those.

trio/_core/_tests/test_asyncgen.py
jakkdl: well, technically it is used - but I was mostly thinking about getting type errors on the callers side if they're passing something that isn't a `Callable`.

trio/_tests/test_deprecate.py
CoolCat467: Changed in c06e788

trio/_core/_tests/test_windows.py
CoolCat467: Changed in c06e788

trio/_core/_tests/test_asyncgen.py
CoolCat467: Changed in c06e788

trio/_core/_tests/test_asyncgen.py
CoolCat467: Changed in c06e788

trio/_core/_tests/test_asyncgen.py
CoolCat467: Changed in c06e788

trio/_core/_tests/test_asyncgen.py
CoolCat467: Changed in c06e788

TeamSpen210 (approved):
Looks good, though you might want to remove these test modules from `pyproject.toml` so we get strict type checking.

trio/_tests/test_deprecate.py
```diff
@@ -11,23 +14,32 @@
...
     warnings.simplefilter("always")
     # ResourceWarnings about unclosed sockets can occur nondeterministically
     # (during GC) which throws off the tests in this file
     warnings.simplefilter("ignore", ResourceWarning)
     return recwarn
 
 
-def _here():
-    info = inspect.getframeinfo(inspect.currentframe().f_back)
+RecWarnObj: TypeAlias = Any
```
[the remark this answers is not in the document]
CoolCat467: Changed in 4fcc74e

trio/_core/_tests/test_asyncgen.py
TeamSpen210: Actually it should be `CaptureFixture[str]`, the arg determines if the results are strings or bytes.

CoolCat467:
@jakkdl @TeamSpen210 Is a re-review with most recent changes required before merging?

TeamSpen210:
For `CaptureFixture`, you should be able to use `str`, meaning that the `skipif` error will probably fix itself.

trio/_core/_tests/test_asyncgen.py
CoolCat467: Ah, that makes sense, I couldn't figure it out from looking at it what the generic was trying to communicate

CoolCat467: Changed in 5656eaf

PR #2773, opened 2023-08-24
Add types to `test_io`, `test_run`, various other fixes

TeamSpen210:
Adds types for two tests, and implements some TODOs.


trio/_tests/verify_types_darwin.json
```diff
@@ -76,7 +76,7 @@
     ],
     "otherSymbolCounts": {
       "withAmbiguousType": 0,
-      "withKnownType": 682,
+      "withKnownType": 680,
```
jakkdl, line 79: if checking manually (I realize now I removed generating `full_output.json` feel free to add that back if it helps), what's the removed symbols?

trio/_core/_tests/test_io.py
```diff
@@ -305,12 +322,12 @@ async def test_wait_on_invalid_object():
...
 
 
-async def test_io_manager_statistics():
-    def check(*, expected_readers, expected_writers):
+async def test_io_manager_statistics() -> None:
+    def check(*, expected_readers: int, expected_writers: int) -> None:
         statistics = _core.current_statistics()
         print(statistics)
         iostats = statistics.io_statistics
-        if iostats.backend in ["epoll", "windows"]:
```
A5rocks, line 313: Using a tuple here should work just as well

A5rocks: Nvm, I forgot `in` doesn't narrow :(

trio/_tests/verify_types_darwin.json
TeamSpen210: It'd be the `nursery.start()` protocol, and its `__call__` method.

trio/_core/_tests/test_io.py
TeamSpen210: It could/should, just Mypy doesn't implement it yet.

trio/_tests/verify_types_darwin.json
jakkdl: :+1: 
We might want to keep a version of these `json` files around in the end after all, though just as the list of symbols. Kinda similar to what `test_exports` is doing but not quite. I'm not sure if dropping private symbols is ever really a problem, but could be good to keep track of it.

TeamSpen210: Not really, as long as everything works (= tests pass). If it's private, by definition it can change whenever we want.

trio/_core/_run.py
```diff
@@ -1173,7 +1164,10 @@ def start_soon(
         GLOBAL_RUN_CONTEXT.runner.spawn_impl(async_fn, args, self, name)
 
     async def start(
-        self, async_fn: _NurseryStartFunc[StatusT], *args: object, name: object = None
+        self,
+        async_fn: Callable[..., Awaitable[object]],
+        *args: object,
+        name: object = None,
```
jakkdl, line 1170: So this drops the check for it accepting a `task_status` parameter. If that's planned to be added back with `ParamSpec`/`TypeVarTuple` it probably warrants a `#TODO` explicitly mentioning that. Or is there a reason for dropping it?

trio/_tests/verify_types_darwin.json
jakkdl: Well, main reason would be checking the public types - the private types mostly just incidental but I don't think it hurts (at least if the updating of the file can be automated in pre-commit or so)

jakkdl (changes requested):
My eyes glazed over reading `test_run`, that file is insanely long! The rest looks good though

trio/_core/_run.py
TeamSpen210: Yeah it needs TypeVarTuple. That protocol would only match callables that actually themselves use `*args`, which is not ideal.

jakkdl (approved):
Looks ready to be merged

trio/_core/_tests/test_io.py
A5rocks: Yeah there's an issue about that :(

trio/_core/_tests/test_run.py
```diff
@@ -208,26 +227,26 @@ async def crasher():
...
+    async def child1() -> None:
         nonlocal t1, t2
         t1 = _core.current_task()
         print("child1 start")
         x = await sleep_forever()
         print("child1 woke")
         assert x == 0
         print("child1 rescheduling t2")
-        _core.reschedule(t2, outcome.Error(ValueError()))
+        _core.reschedule(not_none(t2), outcome.Error(ValueError()))
```
A5rocks, line 242: Why not put the `assert` above this call? (similar for the one below)

Just a nitpick though, really.

trio/_core/_tests/test_run.py
```diff
@@ -249,46 +268,46 @@ async def test_current_time():
...
+        assert not_none(_core.current_task().parent_nursery).parent_task is parent_task
 
     async with _core.open_nursery() as nursery:
         nursery.start_soon(child)
 
 
-async def test_root_task():
-    root = _core.current_root_task()
+async def test_root_task() -> None:
+    root = not_none(_core.current_root_task())
```
A5rocks, line 294: Similar to above comment.

I'm not commenting on `test_current_task` I feel like getting a `None` there would be less an issue with the `current_task()` but idk, maybe that too?

trio/_core/_tests/test_run.py
```diff
@@ -777,20 +796,20 @@ async def task2():
...
     # without affecting any state
-    async def task3(task_status):
+    async def task3(task_status: _core.TaskStatus[_core.CancelScope]) -> None:
         with _core.CancelScope() as scope:
             task_status.started(scope)
             await sleep_forever()
 
     async with _core.open_nursery() as nursery:
-        scope = await nursery.start(task3)
+        scope: _core.CancelScope = await nursery.start(task3)
```
A5rocks, line 805: Why the explicit type annotation?

trio/_core/_tests/test_run.py
```diff
@@ -1940,25 +1999,23 @@ async def inner():
     _core.run(inner)
 
 
-def test_Nursery_init():
+def test_Nursery_init() -> None:
     with pytest.raises(TypeError):
-        _core._run.Nursery(None, None)
+        _core._run.Nursery(None, None, False)  # type: ignore[arg-type]
```
A5rocks, line 2004: Why the type ignore?

trio/_core/_tests/test_run.py
```diff
@@ -2025,15 +2078,18 @@ async def test_Task_custom_sleep_data():
...
-async def test_permanently_detach_coroutine_object():
-    task = None
-    pdco_outcome = None
+async def test_permanently_detach_coroutine_object() -> None:
+    task: _core.Task | None = None
+    pdco_outcome: outcome.Outcome[str] | None = None
 
-    async def detachable_coroutine(task_outcome, yield_value):
+    async def detachable_coroutine(
+        task_outcome: outcome.Outcome[Any],
```
A5rocks, line 2090: I *think* this can just be `task_outcome[None]`...? But honestly I'm a bit confused by this test...

trio/_core/_tests/test_run.py
```diff
@@ -2075,21 +2132,18 @@ async def bad_detach():
         nursery.start_soon(bad_detach)
 
 
-async def test_detach_and_reattach_coroutine_object():
-    unrelated_task = None
-    task = None
+async def test_detach_and_reattach_coroutine_object() -> None:
+    tasks: dict[str, _core.Task] = {}
```
A5rocks, line 2136: Why did you not use `not_none` here when you did so for similar situations elsewhere?

A5rocks (approved):
A couple nitpicks, a couple questions. Nothing that seems like it absolutely needs to be fixed, though I'd like you to go through my comments before merging ^^

trio/_core/_tests/test_run.py
TeamSpen210: I must have edited this one earlier, before I swapped to `not_none()`. Will switch this one to that too.

trio/_core/_tests/test_run.py
TeamSpen210: We're testing that the nursery raises if you try to construct it (I think?), but `None` isn't a valid parameter. Maybe I should tweak this to pass in valid objects instead.

A5rocks: Yeah I think you should.

trio/_core/_tests/test_run.py
TeamSpen210: Mypy demanded an annotation here, since `nursery.start()` isn't properly typed - it doesn't know what it returns yet.

trio/_core/_tests/test_run.py
TeamSpen210: I added `not_none()` to try and distinguish between asserts used to test Trio, and ones used to indicate what should be happening here. If `t2` never gets set, there's either something wrong with the test or I guess Trio's core internals.

trio/_core/_tests/test_run.py
A5rocks: Makes sense.

A5rocks (approved):
LGTM!

PR #2774, opened 2023-08-25
Typing SocketType, test_socket, test_highlevel_[socket, open_tcp_stream, open_tcp_listeners]

### LINKED ISSUE #2720: Handling _SocketType and SocketType
[`SocketType`](https://github.com/python-trio/trio/blob/a9a9170a0142ccf865ceaff10bd3ac0b4f30911e/trio/_socket.py#L425) is currently an empty type and what's meant for end-users to inherit from to show compatibility with trio sockets, but [`_SocketType`](https://github.com/python-trio/trio/blob/a9a9170a0142ccf865ceaff10bd3ac0b4f30911e/trio/_socket.py#L433) (which inherits from `SocketType` is what's actually used internally and has all the methods needed to get code to actually pass type checking.

So we probably want to merge these one way or another, I started trying to tackle it in #2705 but all my ideas hit problems and will likely have some effect on user code that inherits from either of them, so I think it deserves it's own issue + PR.

In tests we have three different `FakeSocket` classes inheriting from `SocketType`: [test_highlevel_socket.py](https://github.com/python-trio/trio/blob/a9a9170a0142ccf865ceaff10bd3ac0b4f30911e/trio/_tests/test_highlevel_socket.py#L206), [test_highlevel_open_tcp_stream.py](https://github.com/python-trio/trio/blob/a9a9170a0142ccf865ceaff10bd3ac0b4f30911e/trio/_tests/test_highlevel_open_tcp_stream.py#L180) and [test_highlevel_open_tcp_listeners.py](https://github.com/python-trio/trio/blob/a9a9170a0142ccf865ceaff10bd3ac0b4f30911e/trio/_tests/test_highlevel_open_tcp_listeners.py#L116)

Options include:
1. Make `SocketType` an abstract base class with abstract methods. Will break any code that has methods inheriting from it and not defined - so would need to only mark the methods that are absolutely *required* to be abstract, and others could instead raise `NotImplementedError`.
   *. Will almost certainly raise a lot of type checking errors.
2. Remove _SocketType, moving it all into `SocketType`. I think it's impossible to avoid this being a breaking API change.
3. Don't make any runtime changes at all, but during type checking assign `SocketType = _SocketType`.
   *. will raise a ton of type checking errors, might be possible to reduce the number with some effort.
4. Rename `_SocketType` something like `InternalTrioSocketType` and have that be the class used in signatures that require some methods to be available.
5. ...

jakkdl:
Resolves #2720 

Quite messy, and looks like there's some signatures in `_socket.py` that might warrant changing. Will write a summary when I've finished up `test_highlevel_open_tcp_listeners`

TODO:
* test generic-ness raises type errors
* document how users might want to approach the generic type


jakkdl:
Hm, looking at https://docs.python.org/3/library/socket.html#socket-families I now understand why methods like `recv` have `Any` in their return type when they're returning a socket address. Might warrant changing a couple signatures to use `Any` and/or `object`

trio/_highlevel_open_tcp_stream.py
```diff
@@ -155,11 +172,12 @@ def reorder_for_rfc_6555_section_5_4(
...
             # so that it becomes the second item on the list.
             if i != 1:
-                targets.insert(1, targets.pop(i))
+                # invariant workaround in arguments leads to type issues here
+                targets.insert(1, targets.pop(i))  # type: ignore[arg-type]
             break
 
 
-def format_host_port(host: str | bytes, port: int) -> str:
+def format_host_port(host: str | bytes, port: int|str) -> str:
```
CoolCat467, line 180: `int | str`

trio/_socket.py
```diff
@@ -520,16 +528,211 @@ async def _resolve_address_nocp(
...
+        def detach(self: SocketType) -> int:
+            ...
+
+        def fileno(self: SocketType) -> int:
+            ...
+
+        def getpeername(self: SocketType) -> Any:
+            ...
+
+        def getsockname(self: SocketType) -> Any:
```
CoolCat467, line 550: Not type checking that return type is instance of `object` or does this lead to issues?

CoolCat467 (changes requested):
Looks good other than a few formatting things and a question about `Any` vs `object` usage. I also think it would be good to annotate the `__init__` return types while we are at it so we don't have to do it later.

jakkdl:
> I also think it would be good to annotate the __init__ return types while we are at it so we don't have to do it later.

I have repeatedly asked you why you think this, while giving you explanations why it doesn't. It does *not* matter for `pyright --verifytypes` and it does *not* matter for mypy. Would you please elaborate, and show it with a reproducible sample + program invocation, so we can settle this?
They don't *hurt* in any way, but I don't see any reason to add them, and I have yet to find any check that can be turned on in any type-checker that flags either, when you have parameters that are typed.

trio/_highlevel_open_tcp_stream.py
```diff
@@ -303,7 +303,7 @@ async def open_tcp_stream(
     # face of crash or cancellation
     async def attempt_connect(
         socket_args: tuple[AddressFamily, SocketKind, int],
-        sockaddr: Address,
+        sockaddr: Any,
```
CoolCat467, line 306: Use `AddressFormat` alias

trio/_highlevel_open_tcp_stream.py
```diff
@@ -138,7 +138,7 @@ def reorder_for_rfc_6555_section_5_4(
             SocketKind,
             int,
             str,
-            tuple[str, int] | tuple[str, int, int, int],
+            Any,
```
CoolCat467, line 141: Use `AddressFormat` alias

trio/_dtls.py
```diff
@@ -1198,7 +1198,7 @@ def __init__(self, socket: _SocketType, *, incoming_packets_buffer: int = 10):
         # as a peer provides a valid cookie, we can immediately tear down the
         # old connection.
         # {remote address: DTLSChannel}
-        self._streams: WeakValueDictionary[Address, DTLSChannel] = WeakValueDictionary()
+        self._streams: WeakValueDictionary[Any, DTLSChannel] = WeakValueDictionary()
```
CoolCat467, line 1201: Use `AddressFormat` alias

trio/_dtls.py
```diff
@@ -828,7 +828,7 @@ class DTLSChannel(trio.abc.Channel[bytes], metaclass=NoPublicConstructor):
 
     """
 
-    def __init__(self, endpoint: DTLSEndpoint, peer_address: Address, ctx: Context):
+    def __init__(self, endpoint: DTLSEndpoint, peer_address: Any, ctx: Context):
```
CoolCat467, line 831: Use `AddressFormat` alias

trio/_dtls.py
```diff
@@ -664,7 +664,7 @@ def _read_loop(read_fn: Callable[[int], bytes]) -> bytes:
 
 
 async def handle_client_hello_untrusted(
-    endpoint: DTLSEndpoint, address: Address, packet: bytes
+    endpoint: DTLSEndpoint, address: Any, packet: bytes
```
CoolCat467, line 667: Use `AddressFormat` alias

trio/_dtls.py
```diff
@@ -603,7 +603,7 @@ def valid_cookie(
 
 
 def challenge_for(
-    key: bytes, address: Address, epoch_seqno: int, client_hello_bits: bytes
+    key: bytes, address: Any, epoch_seqno: int, client_hello_bits: bytes
```
CoolCat467, line 606: Use `AddressFormat` alias

trio/_dtls.py
```diff
@@ -581,7 +581,7 @@ def _make_cookie(
 
 
 def valid_cookie(
-    key: bytes, cookie: bytes, address: Address, client_hello_bits: bytes
+    key: bytes, cookie: bytes, address: Any, client_hello_bits: bytes
```
CoolCat467, line 584: Use `AddressFormat` alias

trio/_dtls.py
```diff
@@ -563,7 +563,7 @@ def _signable(*fields: bytes) -> bytes:
 
 
 def _make_cookie(
-    key: bytes, salt: bytes, tick: int, address: Address, client_hello_bits: bytes
+    key: bytes, salt: bytes, tick: int, address: Any, client_hello_bits: bytes
```
CoolCat467, line 566: Use `AddressFormat` alias

CoolCat467 (changes requested):
Looks good to me, but I would replace the `Address` -> `Any` change with `Address` -> `AddressFormat` alias. One of the review comments has more details why I think this should be done.

trio/_tests/test_highlevel_open_tcp_stream.py
```diff
@@ -313,43 +377,45 @@ async def run_scenario(
         return (exc, scenario)
 
 
-async def test_one_host_quick_success(autojump_clock):
+async def test_one_host_quick_success(autojump_clock: trio.testing.MockClock) -> None:
```
TeamSpen210, line 380: Could use `MockClock` directly to be consistent.

trio/testing/_fake_net.py
```diff
@@ -362,7 +378,8 @@ def __enter__(self):
 
     def __exit__(
         self,
-        exc_type: type[BaseException] | None,
+        # builtin `type` is shadowed by the property
+        exc_type: Type[BaseException] | None,
```
TeamSpen210, line 382: Could do `import builtins`, then `exc_type: builtins.type[...]`.

TeamSpen210 (approved):

jakkdl (commented):
uh, turns out I had a very old review-in-progress and it's a mess to find where I commented on what, so I'll submit this now and it might be outdated.

trio/socket.py
```diff
@@ -34,9 +34,7 @@
 
 # import the overwrites
 from ._socket import (
-    Address as Address,
```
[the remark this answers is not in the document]
jakkdl: Well, depends on if/how it's ever changed. I hadn't properly understood what a mess Address/AddressFormat was, and I don't think we can ever have a singular exported type alias that's useful. If we did want to type this harder I'd go with making SocketType generic depending on the protocol - and for that you'd have different aliases depending on the protocol.

jakkdl:
> > I also think it would be good to annotate the `__init__` return types while we are at it so we don't have to do it later.
> 
> I have repeatedly asked you why you think this, while giving you explanations why it doesn't. It does _not_ matter for `pyright --verifytypes` and it does _not_ matter for mypy. Would you please elaborate, and show it with a reproducible sample + program invocation, so we can settle this? They don't _hurt_ in any way, but I don't see any reason to add them, and I have yet to find any check that can be turned on in any type-checker that flags either, when you have parameters that are typed.

@CoolCat467

A5rocks:
(It might just be cause I haven't fully woken up yet, but quoted message feels a little bit too aggressive... it's alright and can stay since it's obviously not meant to be such: just for CoolCat, don't take it in an aggressive way ^^)

I'll make sure to actually review this in a while!! Sorry about the delay.

trio/socket.py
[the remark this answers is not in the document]
CoolCat467: Alright, that makes sense.

trio/_abc.py
```diff
@@ -211,10 +211,10 @@ class SocketFactory(metaclass=ABCMeta):
...
-        type: socket.SocketKind | int | None = None,
-        proto: int | None = None,
-    ) -> _SocketType:
+        family: socket.AddressFamily | int = socket.AF_INET,
+        type: socket.SocketKind | int = socket.SOCK_STREAM,
+        proto: int = 0,
+    ) -> SocketType:
         """Create and return a socket object.
 
         Your socket object must inherit from :class:`trio.socket.SocketType`,
```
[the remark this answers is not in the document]
CoolCat467: I think having them default to the values they would have became originally is a good thing

CoolCat467 (approved):
With the issues with `AddressFormat` explained, I think this looks good!

CoolCat467:
> > > I also think it would be good to annotate the `__init__` return types while we are at it so we don't have to do it later.
> > 
> > 
> > I have repeatedly asked you why you think this, while giving you explanations why it doesn't. It does _not_ matter for `pyright --verifytypes` and it does _not_ matter for mypy. Would you please elaborate, and show it with a reproducible sample + program invocation, so we can settle this? They don't _hurt_ in any way, but I don't see any reason to add them, and I have yet to find any check that can be turned on in any type-checker that flags either, when you have parameters that are typed.
> 
> @CoolCat467

```python
# example.py

class ClassName:
    def __init__(self):
        pass
```

```console
dmypy run "example.py" -- --disallow-untyped-defs
```
yields the following:
```
Daemon started
example.py:4: error: Function is missing a return type annotation  [no-untyped-def]
example.py:4: note: Use "-> None" if function does not return a value
Found 1 error in 1 file (checked 40 source files)
```

TeamSpen210:
That's a special case where `-> None` must be used, since the `__init__()` method has no parameters defined at all other than `self`. Try `def __init__(self, a: int, b: str):` and you'll see there should be no complaints. Alternatively perhaps it's a problem with daemon mode?

CoolCat467:
> That's a special case where `-> None` must be used, since the `__init__()` method has no parameters defined at all other than `self`. Try `def __init__(self, a: int, b: str):` and you'll see there should be no complaints. Alternatively perhaps it's a problem with daemon mode?

It indeed does not raise errors when it has arguments in dmypy 1.5.1, but I feel like this was not always the case. I always make sure it has the return type annotated mostly because I don't like the idea of special cases. Everything else has the return types annotated, so why shouldn't this? And plus, then any code that looks at the `__annotations__` of it has to have special exceptions, and things aren't nearly as clean.

Confirming this is [one of the comments in PEP-484](https://peps.python.org/pep-0484/#the-meaning-of-annotations):
> (Note that the return type of `__init__` ought to be annotated with -> None. The reason for this is subtle. If `__init__` assumed a return annotation of -> None, would that mean that an argument-less, un-annotated `__init__` method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that `__init__` ought to have a return annotation; the default behavior is thus the same as for other methods.)

jakkdl:
I am sorry for the overly aggressive message, but I did get a bit frustrated about it having come up several times :) [1](https://github.com/python-trio/trio/pull/2745#discussion_r1288886189) [2](https://github.com/python-trio/trio/pull/2723) [3](https://github.com/python-trio/trio/pull/2740#discussion_r1285180160) [4](https://github.com/python-trio/trio/pull/2719#discussion_r1278447388)



pyright:
https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness

> Type annotations can be omitted in a few specific cases where the type is obvious from the context:
> ...
> The return type for an __init__ method does not need to be specified, since it is always None.

mypy:

https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods

> The __init__ method is somewhat special – it doesn’t return a value. This is best expressed as -> None. However, since many feel this is redundant, it is allowed to omit the return type declaration on __init__ methods if at least one argument is annotated.


The comment in PEP-484 is specifically about "whether a type checker should check the function, or treat it as an untyped function and ignore the content", and not any confusion about what the return type actually is.
This is the norm in the most widely used type checkers, and in all projects I know of (as it isn't even possible to enable a warning for not having it!) so anybody inspecting `__annotations__` for whatever reason would have to special-case this behavior anyway.

(returning non-None will also raise a TypeError: https://docs.python.org/3/reference/datamodel.html?highlight=init#object.__init__ )

trio/_abc.py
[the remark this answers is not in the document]
A5rocks: I think this is fine. I approach type hints as not bound by semver (i.e. can narrow them or expand them as long as we're not passing something with a new type in) and subclasses can still default to `None`. Hopefully that approach is sane...

trio/_core/_io_kqueue.py
```diff
@@ -149,7 +148,7 @@ def abort(raise_cancel: RaiseCancelT) -> Abort:
         # wait_task_rescheduled does not have its return type typed
         return await _core.wait_task_rescheduled(abort)  # type: ignore[no-any-return]
 
-    async def _wait_common(self, fd: int | socket, filter: int) -> None:
+    async def _wait_common(self, fd: int | _HasFileNo, filter: int) -> None:
```
A5rocks, line 151: I'm a bit weirded out by codecov because this *should* be covered.

trio/_socket.py
```diff
@@ -40,10 +38,17 @@
...
+# _stdlib_socket.socket supports 13 different socket families, see
+# https://docs.python.org/3/library/socket.html#socket-families
+# and the return type of several methods in SocketType will depend on those. Typeshed
+# has ended up typing those return types as `Any` in most cases, but for users that
+# know which family/families they're working in we could make SocketType a generic type,
+# where you specify the return values you expect from those methods depending on the
+# protocol the socket will be handling.
+# But without the ability to default the value to `Any` it will be overly cumbersome for
+# most users, so currently we just specify it as `Any`.
+# AddressFormat = TypeVar("AddressFormat")
```
A5rocks, line 50: Probably forgot to delete this.

trio/_socket.py
```diff
@@ -510,26 +511,213 @@ async def _resolve_address_nocp(
...
-        raise TypeError(
-            "SocketType is an abstract class; use trio.socket.socket if you "
-            "want to construct a socket object"
-        )
+    def __init__(self) -> None:
+        if type(self) == SocketType:
+            raise TypeError(
+                "SocketType is an abstract class; use trio.socket.socket if you "
+                "want to construct a socket object"
+            )
```
A5rocks, line 528: Is there a reason you switched this to behind this `if`?

I think this makes sense but a bit confused what led to this change in specific.

trio/_socket.py
```diff
@@ -510,26 +511,213 @@ async def _resolve_address_nocp(
...
+        raise NotImplementedError
+
+    def set_inheritable(self, inheritable: bool) -> None:
+        raise NotImplementedError
+
+    if sys.platform == "win32" or (
+        not TYPE_CHECKING and hasattr(_stdlib_socket.socket, "share")
+    ):
+
+        def share(self, /, process_id: int) -> bytes:
```
A5rocks, line 586: This coverage warning here also doesn't make sense given we're running the tests on Windows...?

trio/_socket.py
```diff
@@ -510,26 +511,213 @@ async def _resolve_address_nocp(
...
+
+    @property
+    def proto(self) -> int:
+        raise NotImplementedError
+
+    @property
+    def did_shutdown_SHUT_WR(self) -> bool:
+        raise NotImplementedError
+
+    def __repr__(self) -> str:
```
A5rocks, line 616: I think the ordering of methods is a bit haphazard (normally the dunders are together) but that's just a nitpick.

trio/_socket.py
```diff
@@ -510,26 +511,213 @@ async def _resolve_address_nocp(
...
+    async def wait_writable(self) -> None:
+        raise NotImplementedError
+
+    async def accept(self) -> tuple[SocketType, AddressFormat]:
+        raise NotImplementedError
+
+    async def connect(self, address: AddressFormat) -> None:
+        raise NotImplementedError
+
+    def recv(__self, __buflen: int, __flags: int = 0) -> Awaitable[bytes]:
```
A5rocks, line 643: Pos-only args here, probably

trio/_socket.py
```diff
@@ -510,26 +511,213 @@ async def _resolve_address_nocp(
...
+        raise NotImplementedError
+
+    def recv_into(
+        __self, buffer: Buffer, nbytes: int = 0, flags: int = 0
```
A5rocks, line 647: For this and all the subsequent ones, `__self` can just be `self, /` I believe.

A5rocks (commented):
I didn't review the tests yet but generally I'm just gonna trust you're doing the right thing 😅

trio/_socket.py
jakkdl: Nope, due to typeshed using `__self` -> `_SocketType` using `__self` so `_make_simple_sock_method_wrapper` can validate the signature -> `SocketType` using `__self`. See comment on line 1055

trio/_socket.py
jakkdl: So this code doesn't error and require you to manually specify that you want to call `MyOtherType.__init__`:
```python
class MyOtherType:
  def __init__(self):
    """does something"""

class MySocketType(SocketType, MyOtherType):
  def __init__(self):
    super().__init__()
    """more code"""
```

trio/_socket.py
jakkdl: I kinda kept it in intentionally as what the line would otherwise look like. I could clarify the comment to make that more obvious, or delete the line as it's far from a very complicated one.

trio/_socket.py
jakkdl: Yeah no clue :<

trio/_socket.py
A5rocks: Makes sense. Looks like a mypy bug at first glance xd

(`__arg` should be equal to `arg, /` in all cases... typeshed just can't use pos-only cause not 3.8+)

trio/_socket.py
CoolCat467: Oh that's interesting, I wasn't aware you could do that. What order does it call both init functions in? MRO?

TeamSpen210: The definition here should probably also call `super()`, so it chains to `MyOtherType` too.

jakkdl: > Oh that's interesting, I wasn't aware you could do that. What order does it call both init functions in? MRO?

MRO, yes https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance

trio/_socket.py
jakkdl: clarified

trio/_socket.py
jakkdl: on a rerun there were no coverage warning, so seems like a temporary glitch.

trio/_core/_io_kqueue.py
jakkdl: on a rerun there were no coverage warning, so seems like a temporary glitch.

jakkdl:
uh, thought for a minute that `_public` could be skipped in coverage ... but since the implementation should get called by the generated files that's certainly not the case.

TeamSpen210 (approved):

PR #2777, opened 2023-08-26
Fix typecheck failures in `master`

TeamSpen210:
This fixes the issues that were causing mypy to fail, extracted from #2773.

### DIFF
--- a/trio/_tests/check_type_completeness.py
+++ b/trio/_tests/check_type_completeness.py
@@ -6,19 +6,20 @@ import argparse
 import json
 import subprocess
 import sys
+from collections.abc import Mapping
 from pathlib import Path
 
 # the result file is not marked in MANIFEST.in so it's not included in the package
 failed = False
 
 
-def get_result_file_name(platform: str):
+def get_result_file_name(platform: str) -> Path:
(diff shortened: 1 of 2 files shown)


trio/_tests/check_type_completeness.py
```diff
@@ -33,7 +34,13 @@ def run_pyright(platform: str):
     )
 
 
-def check_less_than(key, current_dict, last_dict, /, invert=False):
+def check_less_than(
+    key: str,
+    current_dict: Mapping[str, object],
+    last_dict: Mapping[str, object],
```
A5rocks, line 40: My PR used `float` here instead of `object` xd

trio/_tests/check_type_completeness.py
```diff
@@ -57,7 +64,7 @@ def check_less_than(key, current_dict, last_dict, /, invert=False):
     )
 
 
-def check_zero(key, current_dict):
+def check_zero(key: str, current_dict: Mapping[str, object]) -> None:
```
A5rocks, line 67: And here

A5rocks (approved):
Essentially the same types I came up with independently for https://github.com/python-trio/trio/pull/2761 so looks good.

PR #2779, opened 2023-08-31
Fix unconditional openssl import in DTLS

VincentVanlaer:


TeamSpen210 (approved):
Makes sense, if we're already documenting support for this. Though, we should probably add a test to confirm that importing `trio` is possible without OpenSSL installed. Perhaps by setting `sys.modules["OpenSSL"] = None` before importing in a subprocess?

jakkdl:
Ah, thanks for the PR! Looks like another reason for setting up tox environments for testing (#2699), to check that we don't get hard dependencies on anything in `test-requirements`. And another nail in the coffin for runtime type-checking #2687

Issue #2775, opened 2023-08-26
Please check our new types!


A5rocks:
[`trio` now has some level of typed-ness](https://github.com/python-trio/trio/pulls?q=label%3Atyping+is%3Aclosed+is%3Apr)! Unfortunately, we did not just copy over `trio-typing` meaning the type hints are slightly different.

If your project:
 1. uses `trio-typing` or some custom type stubs (!?!)
 2. is typechecked against a recent `trio` version

then please help us check our new typed interface by checking your code against `git+https://github.com/python-trio/trio` and telling us your results here! ~~I expect a couple minor-ish typing changes before releasing 0.23.0 (`TypeVarTuple`s, https://github.com/python-trio/trio/pull/2761, and fixes for any type issues) but~~ the current state of things should be final-ish.

agronholm:
I ran this against AnyIO and got 7 errors about unused `type: ignore` comments :+1:

A5rocks:
Checking against [tricycle](https://github.com/oremanj/tricycle), it just looks like over-adaptation to `trio-typing` AND that `_NurseryStartFunc` isn't typed correctly:

```
(venv) PS C:\Users\A5rocks\Documents\Temp\tricycle> mypy --strict --implicit-reexport -p tricycle
tricycle\_service_nursery.py:4: error: Cannot find implementation or library stub for module named "trio_typing"  [import]
tricycle\_service_nursery.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
tricycle\_service_nursery.py:121: error: Unused "type: ignore" comment
tricycle\_service_nursery.py:128: error: Unused "type: ignore" comment
tricycle\_service_nursery.py:132: error: Argument 2 to "start" of "Nursery" has incompatible type "Callable[[NamedArg(Any, 'task_status')], Coroutine[Any, Any, None]]"; expected "_NurseryStartFunc[Any]"  [arg-type]
tricycle\_tests\test_tree_var.py:5: error: Cannot find implementation or library stub for module named "trio_typing"  [import]
tricycle\_tests\test_tree_var.py:96: error: Need type annotation for "target_nursery"  [var-annotated]
tricycle\_tests\test_tree_var.py:96: error: Argument 1 to "start" of "Nursery" has incompatible type "Callable[[Any], Coroutine[Any, Any, None]]"; expected "_NurseryStartFunc[<nothing>]"  [arg-type]
[...]

PR #2780, opened 2023-08-31
Add py.typed in MANIFEST.in

GalaxySnail:
#2775

`py.typed` was not included in distributions, so add it in `MANIFEST.in`.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ include LICENSE LICENSE.MIT LICENSE.APACHE2
 include README.rst
 include CODE_OF_CONDUCT.md CONTRIBUTING.md
 include test-requirements.txt
+include trio/py.typed
 recursive-include trio/_tests/test_ssl_certs *.pem
 recursive-include docs *
 prune docs/build


A5rocks (approved):
Thank you! I was wondering why this wasn't working a few days ago but couldn't figure it out.

Previously:

```
(venv) PS path> pip install mypy git+https://github.com/python-trio/trio >o 2>e ; mypy -c "import trio"
<string>:1: error: Skipping analyzing "trio": module is installed, but missing library stubs or py.typed marker  [import]
<string>:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

Now!:

```
(venv) PS path> pip install mypy git+https://github.com/GalaxySnail/trio@package-py.type >o 2>e ; mypy -c "import trio"
Success: no issues found in 1 source file
```

(for reference i believe sdist works nonetheless, which is how i initially tried things? i forget...)

TeamSpen210 (approved):

PR #2781, opened 2023-08-31
Fix KI protection decorator

A5rocks:
This should fix the `trio.Lock` issue in https://github.com/python-trio/trio/issues/2775.

### DIFF
--- a/trio/_core/_ki.py
+++ b/trio/_core/_ki.py
@@ -6,13 +6,13 @@ import sys
 import types
 from collections.abc import Callable
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Final, Protocol, TypeVar
+from typing import TYPE_CHECKING, Final, Protocol, TypeVar
 
 import attr
 
 from .._util import is_main_thread
 
-CallableT = TypeVar("CallableT", bound="Callable[..., Any]")
+CallableT = TypeVar("CallableT", bound="Callable[..., object]")
 RetT = TypeVar("RetT")
 
 if TYPE_CHECKING:
--- a/trio/_core/_ki.py
+++ b/trio/_core/_ki.py
@@ -6,7 +6,7 @@ import sys
 import types
 from collections.abc import Callable
 from functools import wraps
-from typing import TYPE_CHECKING, Final, TypeVar
+from typing import TYPE_CHECKING, Final, TypeVar, Protocol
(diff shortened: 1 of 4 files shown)


apollo13:
Can confirm that it fixes my `Lock` issue.

TeamSpen210 (approved):
Makes sense as a solution, I had a little difficulty too when doing this originally. Actually though, it'd be better to use a typevar with `bound=Callable`, instead of ParamSpec - in that case, the checker can preserve the entire function type, including things like overloads.

trio/_core/_ki.py
```diff
@@ -183,14 +183,18 @@ def wrapper(*args: ArgsT.args, **kwargs: ArgsT.kwargs) -> RetT:
     return decorator
 
 
-enable_ki_protection: Callable[
-    [Callable[ArgsT, RetT]], Callable[ArgsT, RetT]
-] = _ki_protection_decorator(True)
+# pyright work around: https://github.com/microsoft/pyright/issues/5866
```
jakkdl, line 186: [nit] you probably want "workaround" or "work-around"

jakkdl (approved):
though if teamspen's suggestion of a typevar works that sounds like it would be even better

A5rocks:
> Makes sense as a solution, I had a little difficulty too when doing this originally. Actually though, it'd be better to use a typevar with `bound=Callable`, instead of ParamSpec - in that case, the checker can preserve the entire function type, including things like overloads.

I'm not entirely sure this is possible due to use using `*args: P.args`, etc. inside. It might be possible though? And that might be better external-facing behavior so probably worth a `# type: ignore` if CI running atm says it doesn't work.

trio/_core/_ki.py
```diff
@@ -6,16 +6,17 @@
...
 from collections.abc import Callable
 from functools import wraps
-from typing import TYPE_CHECKING, Final, Protocol, TypeVar
+from typing import TYPE_CHECKING, Any, Final, Protocol, TypeVar
 
 import attr
 
 from .._util import is_main_thread
 
+CallableT = TypeVar("CallableT", bound=Callable[..., Any])
```
TeamSpen210: This can just use `object`, if you want to avoid the `Any`.

CoolCat467 (approved):

PR #2784, opened 2023-09-01
make imports in tests on libraries not in install_requires skip on ImportError

jakkdl:
This opens up for a tox and/or CI environment that runs tests without optional libraries, see #2779


trio/_tests/tools/test_gen_exports.py
```diff
@@ -3,13 +3,17 @@
...
+try:
+    from trio._tools.gen_exports import (
+        File,
+        create_passthrough_args,
+        get_public_methods,
+        process,
+        run_linters,
+    )
+except ImportError as error:
+    assert error.msg == "No module named 'astor'"
```
TeamSpen210, line 15: Instead of testing an error message, it would be more robust to just import `astor` here, skip on `ImportError`, then do the import of `gen_exports` to run after. You might want to import `isort` and `black` too, to check those are installed.

jakkdl: I was thinking something like this would be easier to maintain, in case `gen_exports` is changed not to import `astor` and somebody forgets to update imports in `test_gen_exports`. But you're right that `isort` and `black` can also raise import errors.

jakkdl: eh, it got messy when `black` isn't even imported in `gen_exports` - so went with the simpler implementation + added a `noqa`d import for black in `gen_exports` + added a comment to make sure the list in `test_gen_exports` is kept up to date.

trio/_tools/gen_exports.py
```diff
@@ -16,14 +16,22 @@
 from textwrap import indent
 from typing import TYPE_CHECKING
 
+import attr
+
 if TYPE_CHECKING:
     from typing_extensions import TypeGuard
 
+# keep these imports up to date with conditional imports in test_gen_exports
+# isort: off
```
TeamSpen210, line 25: Instead of turning isort off/on, you can use `isort: split` to ensure imports don't get intermixed across the comment. Then it'll still sort the external tool imports.

TeamSpen210 (approved):
Looks good, though perhaps we might want to make these skips conditional in some way. Then we'd only skip if/when we run a deliberately minimal list of libraries, and still raise/fail otherwise. That'd ensure missing libraries don't suddenly just silently skip tests.

TeamSpen210 (approved):

jakkdl:
Added a CI environment for it: https://github.com/python-trio/trio/actions/runs/6084671235/job/16507092173?pr=2784

jakkdl:
hm, I'm a bit confused why `test_process` passes: https://github.com/python-trio/trio/actions/runs/6084671235/job/16507092173?pr=2784#step:5:1339
Seems like `black` is installed??

TeamSpen210:
Since there's a lot of places with optional imports, it might be better to just move the code in the `except:` branch into a common function. That way `SKIP_OPTIONAL_IMPORTS` doesn't even need to be exposed anywhere, and you can skip coverage in one place. Plus add comments to explain what's happening.

jakkdl:
@TeamSpen210 wanna reapprove a third time after the changes before I merge? :heart:

TeamSpen210 (approved):

Issue #2783, opened 2023-09-01
typing error on master in _core/_multierror since exceptiongroup==1.1.3


jakkdl:
https://github.com/python-trio/trio/actions/runs/6043678194/job/16401102375#step:5:840
```
 ++ mypy trio --platform linux
trio/_core/_multierror.py:264: error: Signature of "derive" incompatible with supertype "BaseExceptionGroup"  [override]
trio/_core/_multierror.py:264: note:      Superclass:
trio/_core/_multierror.py:264: note:          @overload
trio/_core/_multierror.py:264: note:          def [_ExceptionT <: Exception] derive(self, Sequence[_ExceptionT], /) -> ExceptionGroup[_ExceptionT]
trio/_core/_multierror.py:264: note:          @overload
trio/_core/_multierror.py:264: note:          def [_BaseExceptionT <: BaseException] derive(self, Sequence[_BaseExceptionT], /) -> BaseExceptionGroup[_BaseExceptionT]
trio/_core/_multierror.py:264: note:      Subclass:
trio/_core/_multierror.py:264: note:          @overload
trio/_core/_multierror.py:264: note:          def derive(self, Sequence[Exception], /) -> NonBaseMultiError
trio/_core/_multierror.py:264: note:          @overload
trio/_core/_multierror.py:264: note:          def derive(self, Sequence[BaseException], /) -> MultiError
trio/_core/_run.py:204: error: Unused "type: ignore" comment  [unused-ignore]
Found 2 errors in 2 files (checked 129 source files)
++ EXIT_STATUS=1
```
bumped in #2782

A5rocks:
Actually it's cause of a `exceptiongroups` bump -- I'm just not sure exactly why we use the whole assignment thing instead of just subclassing so I left that for a bit.

jakkdl:
@CoolCat467 wrote the type stubs in #2742

CoolCat467:
Basically there was a bit of a mess with overloads and the different overloads accepting things that were too broad and caught other branches of it.

jakkdl:
> Basically there was a bit of a mess with overloads and the different overloads accepting things that were too broad and caught other branches of it.

you up for writing a PR to fix it?

CoolCat467:
> > Basically there was a bit of a mess with overloads and the different overloads accepting things that were too broad and caught other branches of it.
> 
> you up for writing a PR to fix it?

Sure, I'll take a crack at it.

CoolCat467:
#2787

PR #2786, opened 2023-09-03
Require static checks to pass in CI

richardsheridan:
Recently the static checks do more than just black and isort. Recent errors like #2783 and #2777 would be caught this way, and it would make it less error prone to rebase some old PRs that lack typing.

Notably this PR won't pass until #2783 and maybe others are fixed. Whenever master shows green on status checks, someone needs to tickle the update and merge buttons.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,8 +104,7 @@ jobs:
     continue-on-error: >-
       ${{
         (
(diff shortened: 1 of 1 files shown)


A5rocks (approved):
Looks good to me!

richardsheridan:
I'll take that approval as permission to self-merge :)

PR #2787, opened 2023-09-03
Attempt to fix multierror typing

CoolCat467:
At the moment, this is only ignoring generics not matching properly in the subclass. If anyone has any better ideas on how to handle this, feel free to try it out.

### DIFF
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -265,7 +265,7 @@ class MultiError(_BaseExceptionGroup):
     def derive(self, excs: Sequence[Exception], /) -> NonBaseMultiError:
         ...
 
-    @overload  # type: ignore[override]  # 'BaseException' != '_BaseExceptionT'
+    @overload
     def derive(self, excs: Sequence[BaseException], /) -> MultiError:
         ...
 
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -261,11 +261,11 @@ class MultiError(_BaseExceptionGroup):
     def __repr__(self) -> str:
         return f"<MultiError: {self}>"
 
-    @overload
+    @overload  # type: ignore[override]  # 'Exception' != '_ExceptionT'
     def derive(self, excs: Sequence[Exception], /) -> NonBaseMultiError:
         ...
 
-    @overload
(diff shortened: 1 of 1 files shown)


TeamSpen210:
I think the correct solution is to change `MultiError`/`NonBaseMultiError` to also be generic, and propagate that throughout the code.

A5rocks:
I tried and it was annoying (some branch in my fork...)

CoolCat467:
> I think the correct solution is to change `MultiError`/`NonBaseMultiError` to also be generic, and propagate that throughout the code.

That sounds like a good idea, I'll start trying to do that now.

TeamSpen210:
It probably is fine if we add more ignores here, since it's deprecated anyway. Users should only be annotating/interacting with `ExceptionGroup`.

CoolCat467:
It's looking like it might not be possible to make `NonBaseMultiError` generic, because `ExceptionGroup` which we are subclassing itself needs to be using the generic we are defining

CoolCat467:
I've given up attempting to make `NonBaseMultiError` generic, I don't think it's easily possible without doing some nasty things at runtime with monkeypatching

trio/_core/_run.py
```diff
@@ -1705,13 +1701,17 @@ def spawn_impl(
             assert self.init_task is None
 
         ######
-        # Propagate contextvars
+        # Propagate contextvars, and make sure that async_fn can use sniffio.
```
A5rocks, line 1704: Err, maybe a bad merge happened or something?

A5rocks (approved):
Assuming this passes type checking and that the weird previous changes showing in the diff thing is fixed, this looks good to me.

A5rocks:
Actually the merge looks quite messed up -- maybe reset to a clean state, rebase on master, and copy over the type ignore deletion + new type ignore?

newsfragments/2700.misc.rst
A5rocks, line 1: Uhhhhh... I think you might have messed up the merge even more 😰

trio/_core/_run.py
CoolCat467: Yes, very bad, but it's fixed now.

A5rocks (approved):
Looks good

PR #2788, opened 2023-09-03
Fix type errors introduced when #2696 and #2700 were merged

CoolCat467:
This PR aims to fix the type errors introduced in #2696 and #2700, while leaving the multierror fixes for #2787.

### DIFF
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -2357,7 +2357,7 @@ def start_guest_run(
     # spawn_system_task. We don't actually run any user code during
     # this time, so it shouldn't be possible to get an exception here,
     # except for a TrioInternalError.
-    next_send: EventResult = cast(
+    next_send = cast(
         EventResult, None
     )  # First iteration must be `None`, every iteration after that is EventResult
     for tick in range(5):  # expected need is 2 iterations + leave some wiggle room
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -200,8 +200,7 @@ def collapse_exception_group(
         )
         return exceptions[0]
     elif modified:
-        # derive() returns Any for some reason.
-        return excgroup.derive(exceptions)  # type: ignore[no-any-return]
+        return excgroup.derive(exceptions)
     else:
         return excgroup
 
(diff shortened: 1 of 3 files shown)


trio/_core/_run.py
```diff
@@ -200,8 +200,7 @@ def collapse_exception_group(
         )
         return exceptions[0]
     elif modified:
-        # derive() returns Any for some reason.
-        return excgroup.derive(exceptions)  # type: ignore[no-any-return]
+        return excgroup.derive(exceptions)
```
A5rocks, line 203: You might have forgotten to move this bit to the other PR?

CoolCat467: This is connected in a way yes, but the type ignore is not needed anymore because of updating exceptiongroup. The only mypy failures after this PR are the ones I intend to fix in #2787

A5rocks: Ah sorry I assumed that other PR was for fixing type errors due to updating exceptiongroup instead of fixing that single one. Makes sense!

trio/_core/_run.py
```diff
@@ -2358,7 +2357,9 @@ def my_done_callback(run_outcome):
     # spawn_system_task. We don't actually run any user code during
     # this time, so it shouldn't be possible to get an exception here,
     # except for a TrioInternalError.
-    next_send = None
+    next_send: EventResult = cast(
+        EventResult, None
+    )  # First iteration must be `None`, every iteration after that is EventResult
```
A5rocks, line 2362: do we need really need to explicitly type this?

A5rocks (approved):
Changes seem fine, though nitpicks obviously...

I'll let someone else merge this (this isn't like emergency-level priority or anything :^)

trio/_core/_run.py
CoolCat467: Casting None to EventResult yes, but you are probably right about not having to explicitly type it after the cast.

CoolCat467: Changed in dacd4b3

richardsheridan:
Thank you!

Issue #2683, opened 2023-07-04
Add static analysis support to _util.Final


jakkdl:
See discussion in https://github.com/python-trio/trio/pull/2668#issuecomment-1607558210

`@final` was added in 3.8, so it would be good to make classes decorated with `_util.Final` also be treated as `@final` by type checkers. I didn't find a good way of applying both with a single decorator or metaclass, so unless I missed a way we might have to have both on all classes. Or possibly skip one of them.

A5rocks:
Might be a nice case for just a lint to ensure we have both? I can't think of a solution here unfortunately.

TeamSpen210:
There is actually a better way, but it'd require some churn. By using `__init_subclass__`, we [can implement](https://github.com/python-trio/trio/commit/b6c4089d6066047bb18814980a8ac8e0eb743466) `_util.Final` as a decorator. Then we can just make it an alias to the `typing` version `if TYPE_CHECKING`. A decorator's better anyway since we don't have to worry about metaclass conflicts. 

The biggest downside is that we have to remember to use `@final` alongside `NoPublicConstructor`, since I can't think of a way to make the latter a decorator. We could fairly easily add a check to the `test_exports` logic to look for these cases though.

I don't think we need to care about backwards compatibility, since these are both private.

jakkdl:
Sounds like a decent enough solution to me :+1:

PR #2793, opened 2023-09-06
#2683: Make `_util.Final` support static analysis by converting to a decorator.

TeamSpen210:
#2683: By converting `_util.Final` to be a decorator instead of a metaclass, we can alias it with the `typing.final` decorator. This will make type checkers also refuse to allow subclasses, and allow them to do type narrowing in some cases. Unfortunately I can't see a way to do the same with `NoPublicConstructor`, so instead I added a test which verifies all uses of that are also `@final`.

Aside from the static typing benefits, this simplifies the implementation a little and allows `_util.final` to be used on any class since we don't have metaclass conflicts.

There's a lot of changes here, since I modified every place `Final` or `NoPublicConstructor` is used. The actual important changes are in `trio/_util.py` and `trio/_tests/test_util.py`. The `verify_types_*.json` tables drop 2 private symbols because of the removal of `_util.Final`.


trio/_tests/test_util.py
```diff
@@ -170,13 +170,14 @@ def test_func(arg):
...
+def test_final_decorator() -> None:
+    @final
+    class FinalClass:
         pass
 
     with pytest.raises(TypeError):
 
-        class SubClass(FinalClass):
+        class SubClass(FinalClass):  # type: ignore[misc]
             pass
```
jakkdl, line 181: Add a comment here explaining a bit more what's being tested (particularly that it's two tests in one, both a runtime check and a type-check), and that `misc` is for "inherit from final"

jakkdl (approved):

trio/_core/_parking_lot.py
```diff
@@ -79,12 +79,13 @@
...
 
 from .. import _core
-from .._util import Final
+from .._util import final
 
 if TYPE_CHECKING:
     from ._run import Task
 
 
+@final
```
A5rocks, line 88: Just noting that this adds a `@final` where there were none before. It might have been nicer to do so in another PR for a nicer diff but eh, not a priority. Up to you.

trio/_core/_unbounded_queue.py
```diff
@@ -6,14 +6,15 @@
...
-from .._util import Final
+from .._util import final
 
 T = TypeVar("T")
 
 if TYPE_CHECKING:
     from typing_extensions import Self
 
 
+@final
```
A5rocks, line 17: Another `@final` addition

trio/_tests/test_exports.py
```diff
@@ -484,6 +496,16 @@ def lookup_symbol(symbol):
...
 
 
+def test_nopublic_is_final() -> None:
+    """Check all NoPublicConstructor classes are also @final."""
+    assert class_is_final(_util.NoPublicConstructor)  # This is itself final.
+
+    for module in ALL_MODULES:
+        for name, class_ in module.__dict__.items():
+            if isinstance(class_, _util.NoPublicConstructor):
+                assert class_is_final(class_)
```
A5rocks, line 506: Should we print out the class being tested so that when the test fails and stdout is printed, it's easier to check? (Or actually, `pytest` says what `class_` is through its `assert` debug stuff right?)

trio/_tests/test_exports.py
```diff
@@ -519,4 +541,4 @@ def test_classes_are_final() -> None:
             if name.endswith("Statistics"):
                 continue
 
-            assert isinstance(class_, _util.Final)
+            assert class_is_final(class_), f"{class_.__module__}.{class_.__qualname__}"
```
A5rocks, line 544: Hm, see previous comment, would rather this and that be consistent on how much debug info they provide!

trio/_util.py
```diff
@@ -325,39 +331,39 @@ class SomeClass(metaclass=Final):
     ------
     - TypeError if a subclass is created
     """
+    # Override the method blindly. We're always going to raise, so it doesn't
+    # matter what the original did (if anything).
+    decorated.__init_subclass__ = classmethod(_init_final_cls)  # type: ignore[assignment]
+    # Python 3.11+ sets this attr, do it on all versions so this being present can be tested.
+    decorated.__final__ = True  # type: ignore[attr-defined]
```
A5rocks, line 338: We could just call `typing.final` in case it ever grows new functionality.

A5rocks: ... Or actually that won't help with testing things. Uh.... I mean we can check `__init_subclass__` maybe?

A5rocks (commented):
Mostly minor comments. RE: making a new PR, just rebasing so you have like 3 commits to show specifically which commit is adding new final classes would work too.

trio/_util.py
TeamSpen210: Definitely should call `typing.final` also, in case it does have additional runtime effects. Checking `__init_subclass__` feels a bit iffy since it's sorta leaking implementation details. Though these are tests. I could try actually subclassing every class, see how slow that is :P.

trio/_core/_parking_lot.py
TeamSpen210: Should probably revert these two, `test_exports` currently just allows the stats classes to be final.

trio/_tests/test_exports.py
TeamSpen210: It does, but I found that it was failing to include the full module name in some cases.

trio/_util.py
A5rocks: We could unconditionally set `__final__` and call `typing.final` too ig

A5rocks (approved):
Any remaining threads from my past review are just opinions on implementation: choose what you want! I think anything sounds good and this PR looks good.

A5rocks:
Not sure if you want this squashed and merged or if you want to rebase to have several commits!

PR #2794, opened 2023-09-07
Add types to `_signals`, `test_signals`

TeamSpen210:
This is fairly straightforward, but I did extract a test helper method into a function. This allows `open_signal_receiver()` to be typed as producing an `AsyncIterator`, keeping `SignalReceiver` internal.

### DIFF
--- a/trio/_signals.py
+++ b/trio/_signals.py
@@ -90,7 +90,7 @@ class SignalReceiver:
         # And then redeliver any that are sitting in pending. This is done
         # using a weird recursive construct to make sure we process everything
         # even if some of the handlers raise exceptions.
-        def deliver_next():
+        def deliver_next() -> None:
             if self._pending:
                 signum, _ = self._pending.popitem(last=False)
                 try:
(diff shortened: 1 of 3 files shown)


CoolCat467 (approved):
Looks good to me. If anything, the change removing `_pending_signal_count` could potentially cause issues if people are using this function even though it basically says not to because it was only there for the tests, but I think this is an acceptable change because:
1) People shouldn't be using said function anyways
2) Making it a separate function that isn't publicly exposed is a better solution for the tests than having it be a class function that technically could exposed

trio/_signals.py
```diff
@@ -111,8 +118,17 @@ async def __anext__(self):
             return signum
 
 
+def get_pending_signal_count(rec: AsyncIterator[int]) -> int:
+    """Helper for tests, not public or otherwise used."""
+    # open_signal_receiver() always produces SignalReceiver, so cast.
```
A5rocks, line 123: Out of date comment (you assert, not cast here).

A5rocks (approved):
Looks like a pretty simple file to type. Essentially no comments!

A5rocks (approved):
Looks good to me.

jakkdl (approved):

PR #2796, opened 2023-09-09
Replace public usages of `functools.wraps`

A5rocks:
I'm not *entirely* sure this works right but I can try figuring out how to configure pyright in a day or something unless someone else wants to check!

Unfortunately yet another pyright incompatibility -- while we could just replace usages around methods this is probably easier?


A5rocks:
Huh the typing impact is kinda shocking?

jakkdl:
> Huh the typing impact is kinda shocking?

whoa, great you took the initiative to do this! I didn't know `@wraps` was *that* broken in typeshed. I can look at the socket errors on monday or tuesday if you don't wanna try and delve into that thorny mess before then. Could maybe pull it into #2774 and handle it all there

jakkdl:
> > Huh the typing impact is kinda shocking?
> 
> whoa, great you took the initiative to do this! I didn't know `@wraps` was _that_ broken in typeshed. I can look at the socket errors on monday or tuesday if you don't wanna try and delve into that thorny mess before then. Could maybe pull it into #2774 and handle it all there

oh, so there was no problem and it was just a matter of your first stab at `_wraps` being incorrect?

Hnasar:
> I didn't know `@wraps` was _that_ broken in typeshed.

Yeah unfortunately the stubs break the wrapped functions signature. This comment explains the situation well: https://github.com/python/mypy/issues/5107#issuecomment-1355954910

jakkdl:
> > I didn't know `@wraps` was _that_ broken in typeshed.
> 
> Yeah unfortunately the stubs break the wrapped function's signature. This comment explains the situation well: [python/mypy#5107 (comment)](https://github.com/python/mypy/issues/5107#issuecomment-1355954910)

thank you! It broke my brain a little bit, but I now I get the distinction.

jakkdl:
no wait, I'm confused again. I think I'm getting different behavior if not renaming the import to `_wraps`. Could you try just doing a straight `from functools import wraps` and see if that makes a difference, and if not I'd love a repro of what exactly it is that breaks.

TeamSpen210:
Just checked, MyPy doesn't have a plugin for `wraps`, but it does cherry-pick an older implementation in [its copy of typeshed](https://github.com/python/mypy/blob/88ae1e4c1541e5b03d695cf63d1265b972e427d9/mypy/typeshed/stdlib/functools.pyi#L108), which just treats it as an identity function.

jakkdl:
> no wait, I'm confused again. I think I'm getting different behavior (in pyright) if not renaming the import to `_wraps`. Could you try just doing a straight `from functools import wraps` and see if that makes a difference, and if not I'd love a repro of what exactly it is that breaks.

oh derp, the repro is https://github.com/python-trio/trio/issues/2775#issuecomment-1702267589

But replacing `_wraps` with `wraps` does seem to fix it?

A5rocks:
Just quick comment: I don't think it's worthwhile having both in CI as they have many intentional discrepancies. The only benefit we would have is that our public API would be guaranteed to work... which is probably something we can guarantee another way? (Just typecheck a subset of tests? Type tests that work in both mypy and pyright? Something like that)

jakkdl:
> Just quick comment: I don't think it's worthwhile having both in CI as they have many intentional discrepancies. The only benefit we would have is that our public API would be guaranteed to work... which is probably something we can guarantee another way? (Just typecheck a subset of tests? Type tests that work in both mypy and pyright? Something like that)

yeah agreed, we'd probably want to strip one of them down a lot. But I've been [running both in flake8-trio](https://github.com/Zac-HD/flake8-trio/blob/a652441fb3a9a92f14da3542bb2a0a1ba8f03f87/pyproject.toml#L21) without too many problems, though that project is definitely doing a lot fewer complex things type-wise - and started with pyright.

A5rocks:
Trying out pyright locally, I cannot reproduce `wraps` working! Not sure what's different about my setup, just using VSCode, ctrl+click then editing the file (e.g. replacing our `def _wraps` with the import from `functools` broke things with `Argument of type "list[bytes]" cannot be assigned to parameter "self" of type "_SocketType" in function "__call__"`, then renaming `_wraps` -> `wraps` then getting rid of the `as` didn't fix anything).

jakkdl:
I don't know what I did last time, but I made sure to do everything properly now ... and you are indeed right! Sorry for the confusion

We should totally add a test for this though, and even if properly setting up pyright is a bigger task - we could for the moment create a dedicated test file with the contents of https://github.com/python-trio/trio/issues/2775#issuecomment-1702267589, and for the moment run pyright only on that file. Maybe a directory `trio/_tests/typed_files/` or even more specifically `trio/_tests/pyright_type_checking/`.

A5rocks:
Well, that's probably not the implementation you wanted but I went ahead and added a regression test @jakkdl

TeamSpen210:
With `wraps`, instead of duplicating the definition it would be better to just put that in `_utils` or something, then import.
For the type-testing framework, it might be a little easier to just add a `type_tests/` folder or the like, then just run each checker over that whole folder in the regular CI.

jakkdl:
> Well, that's probably not the implementation you wanted but I went ahead and added a regression test @jakkdl

lol yeah, I'm not a fan of tests with code in strings. Teamspen210's idea is the same as mine.

A5rocks:
Well hopefully that works, I should learn my lesson at this point about not using the web editor ever... It's just too tempting :(

trio/_core/_run.py
```diff
@@ -72,7 +72,7 @@
 DEADLINE_HEAP_MIN_PRUNE_THRESHOLD: Final = 1000
 
 # Passed as a sentinel
-_NO_SEND: Final = cast("Outcome[Any]", object())
+_NO_SEND: Final["Outcome[Any]"] = cast("Outcome[Any]", object())
```
A5rocks, line 75: This should *probably* be reported as an issue but gehhhhh

CoolCat467 (approved):
I think this makes sense, looks good to me.

check.sh
```diff
@@ -62,6 +62,10 @@ mypy trio --show-error-end --platform darwin | python ./trio/_tools/mypy_annotat
     || { echo "* Mypy (Mac) found type errors." >> $GITHUB_STEP_SUMMARY; MYPY=1; }
 mypy trio --show-error-end --platform win32 | python ./trio/_tools/mypy_annotate.py --dumpfile mypy_annotate.dat --platform Windows \
     || { echo "* Mypy (Windows) found type errors." >> $GITHUB_STEP_SUMMARY; MYPY=1; }
+
+# for now, just run interface typing checks on the native platform...
+mypy trio/_tests/type_tests --show-error-end | python ./trio/_tools/mypy_annotate.py --dumpfile mypy_annotate.dat --platform Interface \
+    || { echo "* Mypy (Interface) found type errors." >> $GITHUB_STEP_SUMMARY; MYPY=1; }
```
jakkdl, line 68: Reading this made me think `Interface` was a magic string, so would probably be better to leave it out.

But isn't this fully covered by the above mypy runs? `trio/_tests/type_tests` shouldn't be excluded or anything in `pyproject.toml` AFAICT.

jakkdl (approved):
Other than `check.sh` this looks great!

check.sh
A5rocks: Oh oops, I forgot that it would be covered. Yeah that makes sense.

CoolCat467: 93e7d80 fixes this

A5rocks:
I'm going to merge this. Pypy 3.9 nightly failing on Windows happens elsewhere and this way we can ask for more testing of the types!

I think any further changes can be done in a new PR.

TeamSpen210 (approved):

Issue #2790, opened 2023-09-05
Switch packaging to PDM/Hatch/Poetry


jakkdl:
There's some discussion in #2681, and some more in Gitter, but let's make a dedicated issue for this.

Alternatives:
1. Status quo
   * dependabot is a hassle
   * sort of impossible to properly define different package version requirements on different versions
   * regenerating `*-requirements.txt` locally, as is currently required to pass CI whenever changing something in `*-requirements.in`, requires having a matching python version and possibly other requirements on your environment.
2. Poetry
   * Does not comply with PEP 621: https://github.com/python-poetry/roadmap/issues/3
   * there's an open PR for switching to Poetry: #2595
   * Supports lockfiles, quoting @njsmith on gitter:
     > lockfiles for a library are very valuable for keeping CI functioning
you don't ship them, but if you don't have them in CI and don't have very rapid development, then every time you do try to merge a PR you discover that actually CI has been broken for a while and no-one noticed so now you have to untangle that before you can finish your original thought
3. Hatch
   * Does not support lockfiles, but there are different ways to get around the above problem
     quoting @apollo13
     > Fair enough. But with a nightly run you would also notice if a dependency update is breaking stuff and you could react faster by limiting the dependency till you fixed it

     quoting @agronholm 
     > what I do in my projects instead is cache the dependencies
     > so they only get updated if I update my pyproject.toml or manually delete the caches
     > => no constant dependabot churn
4. PDM
   * supports lockfiles

apollo13:
FWIW, one should differentiate between the management tool and the build backend. For instance it is perfectly feasible to use PDM to manage the lockfile etc and `hatchling` (the build backend from hatch) to actually build the project. This is enabled via the following:
```
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```
and works for any project using PEP 621. In that sense it is also possible to switch between build backends easily (ie use setuptools instead of hatchling). In reality though one usually ends up with one build backend since one often uses one or two plugins to read the version from SCM etc… I like `hatch`/`hatchling` for being standards compliant and having reproducible builds by default. I am not using `hatch` currently though but PDM & `hatchling` since I often need lockfiles as well.

apollo13:
[...]

PR #2798, opened 2023-09-13
Move to src-layout.

apollo13:
The intention of this is better isolation. `trio` is no longer accidentally on the (python) path and as such requires an explicit installation for usage, this helps uncover issues in packaging data etc early on (see
https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/).

This serves as the basis for switching to another build system as well; refs #2790.


altendky:
In CI this is already addressed by changing into a subdirectory when testing.  I personally prefer `src/`, but not doing that has been an intentional decision thus far.  Though, I forget where I saw this discussed.  `:[`  I'm just suggesting that the discussion is had since I don't see a reference from here to it..

altendky:
https://github.com/python-trio/trio/blob/c16003f08285d0b5b3b5b16d5bdc4f00b1e3cd52/ci.sh#L106-L110

apollo13:
@altendky I see, I will still finish this for now and put it up for discussion, because while CI might address it, it can still be pretty much an annoyance. Let's see what the rest says :)

jakkdl:
I think #274 is the only place where it's been discussed

CoolCat467:
I definitely agree, a `src/` based layout would be much nicer, +1

jakkdl:
@njsmith do you want to weigh in on the matter?

jakkdl:
@A5rocks any opinion on the matter?

A5rocks:
I'm neutral on this change cause while it does align with what most people do, so would e.g. not shipping tests... so I don't think that's a very convincing argument.

And while it would take a bit of magic away from the tests it would also mean having an extra "src" folder in the path for every file in trio. (Then of course, I already have to type out trio/trio for the folder...)

So yeah arguments for and against but none of them feel strong, so neutral.

jakkdl:
if this makes switching packaging system easier, I think we can just go ahead with it. The PRs been open a month and the only issue that has been brought up are minor workflow ones.

CoolCat467:
@apollo13 I would be glad to finish working on getting this merged this if you would like

apollo13:
@CoolCat467 That would be great, I am currently blocked by personal things. My goal with the PR was to be as minimal as possible. Ie moving `_tests` (if wanted at all) into another top level package would be something done after this PR (same for switching to whatever other tooling). I can give you access to my repo if you want to work on that branch directly or we close this PR and you open your own.

apollo13:
> if this makes switching packaging system easier, I think we can just go ahead with it. The PRs been open a month and the only issue that has been brought up are minor workflow ones.

I guess it doesn't make switching much easier (if at all), tooling can usually handle both.

altendky:
> Ie moving `_tests` (if wanted at all) into another top level package

I don't want that.  `:]`  I like the stuff I import to have been installed so it's not all fiddly about how you launch Python or what your cwd is.  Also, why not be able to run tests on deployed packages?  Also also, fewer directories to both including and excluding from all the myriad of tools and packaging definitions etc.

CoolCat467:
> @CoolCat467 That would be great, I am currently blocked by personal things. My goal with the PR was to be as minimal as possible. Ie moving `_tests` (if wanted at all) into another top level package would be something done after this PR (same for switching to whatever other tooling). I can give you access to my repo if you want to work on that branch directly or we close this PR and you open your own.

If I could have access to the repo that would be great, I would prefer not creating another PR if it could be avoided.

apollo13:
Sent an invite

A5rocks:
fyi (cause review was requested) i want to merge this in after the next trio release, in order to minimize possible disruption.... unless someone wants this urgently

A5rocks:
OK release is out. Still ambivalent but if people think this would better and there isn't dissent I guess this should happen? Even if we don't end up using a pep 612 backend that requires the `src/` dir, it still might be better to stick to the beaten path here.

(Obviously conflicts need to be fixed)

CoolCat467 (approved):
Slightly maybe in warning of "review your own PRs" since I've taken over continued work on this, but all the important changes were already made before that point and the tests are passing.

jakkdl (approved):
we might want to wait a day to see if anybody else pipes up about trio 0.23 being broken and needing a hotfix patch, but otherwise I don't see any problems with merging this.

.coveragerc
A5rocks, line 1: We'll need to move this to `pyproject.toml`

setup.py
A5rocks, line 1: We dropped this so you might need to wrangle the new config in `pyproject.toml`

A5rocks (commented):
Conflicts again :(

But mainly just that new exciting `pyproject.toml` changes mean you have to migrate some of the changes here.

.coveragerc
A5rocks: Oops, you're still creating `.coveragerc`

CoolCat467:
@A5rocks How do we fix this error?
```console
  + coverage report -m --rcfile ../pyproject.toml
  No data to report.
  Error: Process completed with exit code 1.
```

A5rocks:
Make sure that coverage is outputting its data files in the cwd, maybe? Not sure tbh

CoolCat467:
@A5rocks I fixed the coverage report and resolved conflicts

pyproject.toml
```diff
@@ -228,15 +225,15 @@ showcontent = true
...
     # These are run in subprocesses, but still don't work. We follow
     # coverage's documentation to no avail.
-    "*/trio/_core/_tests/test_multierror_scripts/*",
+    "*/src/trio/_core/_tests/test_multierror_scripts/*",
     # Omit the generated files in trio/_core starting with _generated_
-    "*/trio/_core/_generated_*",
+    "*/src/trio/_core/_generated_*",
     # Script used to check type completeness that isn't run in tests
-    "*/trio/_tests/check_type_completeness.py",
+    "*/src/trio/_tests/check_type_completeness.py",
```
A5rocks, line 236: I presume these /src/ changes are for source code? But don't we run coverage against trio as a package? I hope we're not shipping /src/ in the wheel...

A5rocks (commented):
GitHub says I have a pending review so let's see what it is...

pyproject.toml
CoolCat467: Should be fixed in 3b582ca

jakkdl:
let's not delay this any further and fix any other potential errors as they crop up. :rocket:

PR #2799, opened 2023-09-16
Add comment and assertion to `pidfd_open` workaround

jakkdl:
EDIT: assumptions were wrong

Ended up researching this during discussion in #2792, and thought I might as well add a comment (and an assertion to double-check assumptions). Also cleared up a coverage miss for an error that shouldn't trigger in test runs.

One could possibly raise a warning instead of the assertion to be *extra* defensive, but afaict it *really* shouldn't trigger on 3.9+ so the assertion+message is already very defensive IMO.

https://docs.python.org/3/library/os.html#os.pidfd_open

### DIFF
--- a/trio/_subprocess.py
+++ b/trio/_subprocess.py
@@ -49,6 +49,12 @@ else:
         from os import pidfd_open
     except ImportError:
         if sys.platform == "linux":
+            # pidfd_open is defined in 3.9+, so this workaround can be removed when
+            # removing support for 3.8
+            assert sys.version_info < (3, 9), (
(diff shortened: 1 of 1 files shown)


jakkdl:
Ah, turns out `pypy` doesn't have `os.pidfd_open`

TeamSpen210 (approved):

A5rocks:
Wait so are we assuming 3.9+ won't run on Linux >5.3 or am I misunderstanding what "availability" means? (Probably the latter)

PR #2803, opened 2023-09-27
Fix wrongly highlighted lines in reference-core.rst

lyphrowny:
Fixes wrongly highlighted lines in reference-core.rst

In both versions of online documentation ([`stable`](https://trio.readthedocs.io/en/stable/reference-core.html#clean-shutdown-with-channels), [`latest`](https://trio.readthedocs.io/en/latest/reference-core.html#clean-shutdown-with-channels)), the lines in "Clean shutdown with channels" are wrongly highlighted.

Stable:
stable

Latest:
latest

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -1174,7 +1174,7 @@ the previous version, and then exits cleanly. The only change is the
 addition of ``async with`` blocks inside the producer and consumer:
 
 .. literalinclude:: reference-core/channels-shutdown.py
-   :emphasize-lines: 11,17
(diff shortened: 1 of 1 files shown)


TeamSpen210 (changes requested):
Based on the text, it's actually the `async with` lines that should be highlighted.

lyphrowny:
@TeamSpen210 Yes, however in the docs these lines are not highlighted. The provided screenshots are from the docs, not from the proposed change. Sorry for the confusion

TeamSpen210 (approved):
Ahh, my mistake.

PR #2805, opened 2023-09-28
get _core/_run to 100% coverage

jakkdl:
Just some random lines I noticed in codecov that should pretty clearly be excluded.


TeamSpen210 (approved):
It might be possible to add a test to exercise the `gc.get_ referents()` branch, but that would be real tricky and not really worth it.

PR #2809, opened 2023-10-03
Enable pyupgrade rule

CoolCat467:
This PR enables the pyupgrade ruff rule and fixes all the existing issues that were raised.


trio/_core/_io_epoll.py
```diff
@@ -201,7 +201,7 @@ class _EpollStatistics:
 class EpollIOManager:
     _epoll: select.epoll = attr.ib(factory=select.epoll)
     # {fd: EpollWaiters}
-    _registered: DefaultDict[int, EpollWaiters] = attr.ib(
+    _registered: defaultdict[int, EpollWaiters] = attr.ib(
```
A5rocks, line 204: Isn't `defaultdict` a 3.9+ thing? Why is ruff doing this? Cause the `from __future__ import annotations`? or is this a bug.

trio/_core/_io_windows.py
```diff
@@ -245,9 +244,9 @@ class CKeys(enum.IntEnum):
 # operation and start a new one.
 @attr.s(slots=True, eq=False)
 class AFDWaiters:
-    read_task: Optional[_core.Task] = attr.ib(default=None)
-    write_task: Optional[_core.Task] = attr.ib(default=None)
-    current_op: Optional[AFDPollOp] = attr.ib(default=None)
+    read_task: _core.Task | None = attr.ib(default=None)
+    write_task: _core.Task | None = attr.ib(default=None)
+    current_op: AFDPollOp | None = attr.ib(default=None)
```
A5rocks, line 249: Eh, same as above. I won't make any more comments about 3.9+ features sneaking in.

trio/_dtls.py
```diff
@@ -841,10 +841,7 @@ def __init__(self, endpoint: DTLSEndpoint, peer_address: Any, ctx: Context):
         # support and isn't useful anyway -- especially for DTLS where it's equivalent
         # to just performing a new handshake.
         ctx.set_options(
-            (
-                SSL.OP_NO_QUERY_MTU
-                | SSL.OP_NO_RENEGOTIATION  # type: ignore[attr-defined]
-            )
+            SSL.OP_NO_QUERY_MTU | SSL.OP_NO_RENEGOTIATION  # type: ignore[attr-defined]
```
A5rocks, line 844: Huh, why did ruff change this? I think it was intentionally formatted to showcase which specific attribute wasn't defined.

trio/_tests/type_tests/check_wraps.py
```diff
@@ -1,8 +1,7 @@
 # https://github.com/python-trio/trio/issues/2775#issuecomment-1702267589
 # (except platform independent...)
-import typing_extensions
-
 import trio
+import typing_extensions
```
A5rocks, line 5: This is fixed on master btw, might want to pull that in

trio/_util.py
```diff
@@ -135,28 +135,26 @@ def _return_value_looks_like_wrong_library(value: object) -> bool:
...
                 "Trio was expecting an async function, but instead it got "
-                "a coroutine object {async_fn!r}\n"
+                f"a coroutine object {async_fn!r}\n"
                 "\n"
                 "Probably you did something like:\n"
                 "\n"
-                "  trio.run({async_fn.__name__}(...))            # incorrect!\n"
-                "  nursery.start_soon({async_fn.__name__}(...))  # incorrect!\n"
+                f"  trio.run({async_fn.__name__}(...))            # incorrect!\n"
+                f"  nursery.start_soon({async_fn.__name__}(...))  # incorrect!\n"
```
A5rocks, line 143: Mind checking the commit that these came in with (or the PR more generally) and seeing if there's other stuff like this?

trio/_util.py
```diff
@@ -174,15 +172,15 @@ def _return_value_looks_like_wrong_library(value: object) -> bool:
         # Give good error for: nursery.start_soon(func_returning_future)
         if _return_value_looks_like_wrong_library(coro):
             raise TypeError(
-                "Trio got unexpected {!r} – are you trying to use a "
+                f"Trio got unexpected {coro!r} – are you trying to use a "
                 "library written for asyncio/twisted/tornado or similar? "
-                "That won't work without some sort of compatibility shim.".format(coro)
+                "That won't work without some sort of compatibility shim."
```
A5rocks, line 177: Wow messed up upgrade bugs everywhere. Mind checking the commit/PR that this came in with too?

A5rocks (commented):
Couple of comments but overall looks good

trio/_core/_io_epoll.py
TeamSpen210: It’s probably that yes. There’s an option to decide whether to preserve the hints making sense at runtime.

CoolCat467: Yes, it decides what's allowed based on both the version you specify and if a future annotations import exists.

trio/_dtls.py
CoolCat467: I saw this and the diff and was a bit confused myself, not sure about this one.

trio/_util.py
CoolCat467: I don't think it's actually a bug, I think the format call would have worked for it previously because it was applying to the concatenation of all of the strings, since they have parenthesis wrapping them. It definitely looks kind of strange though, and I think using f-strings here makes it much more clear what the idea is than this format call that looks a bit out of place.

trio/_util.py
CoolCat467: Same string concat weirdness (see other comment)

trio/_core/_io_epoll.py
CoolCat467: There is a setting to make sure runtime typing works, mostly so a specific library can work properly, but in my opinion I agree with ruff's behavior here. Why have outdated versions of the type annotations if we don't have to worry about runtime type annotations?

trio/_dtls.py
CoolCat467: Another one of my PRs hit this and I looked into it further, this change was by black, not ruff.

jakkdl:
You can merge this after fixing B904.
The codecov fail looks dumb, windows pypy-3.9-nightly is still bad, and macOS 3.10 is hitting the timeout issues that have started plaguing us. So none of those are related to the PR.

CoolCat467:
pre-commit.ci autofix

PR #2811, opened 2023-10-04
Add typing classifier to trio + add typing changelog

A5rocks:
I think we can say the type hints for the library are complete, even if they are not entirely correct.

(Also I resorted the list to be alphabetical again...)

... Also I added a changelog entry.


TeamSpen210 (approved):

CoolCat467 (approved):
Looks good.

PR #2812, opened 2023-10-05
Add type annotations for windows_pipes and its test

CoolCat467:
This PR adds type annotations for `_windows_pipes.py` and `_tests/test_windows_pipes.py`.

### DIFF
--- a/trio/_tests/test_windows_pipes.py
+++ b/trio/_tests/test_windows_pipes.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import sys
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 import pytest
 
@@ -20,12 +20,6 @@ if sys.platform == "win32":
 
     from .._core._windows_cffi import _handle, kernel32
     from .._windows_pipes import PipeReceiveStream, PipeSendStream
-else:
-    pipe = Any
-    _handle = Any
-    kernel32 = Any
-    PipeReceiveStream = Any
-    PipeSendStream = Any
 
 
 async def make_pipe() -> tuple[PipeSendStream, PipeReceiveStream]:
--- a/trio/_tests/test_windows_pipes.py
+++ b/trio/_tests/test_windows_pipes.py
@@ -14,6 +14,8 @@ if sys.platform == "win32":
 else:
     pytestmark = pytest.mark.skip(reason="windows only")
     pipe: Any = None
+    _handle: Any = None
+    kernel32: Any = None
     PipeSendStream: Any = None
     PipeReceiveStream: Any = None
 
@@ -26,9 +28,9 @@ async def make_pipe() -> Tuple[PipeSendStream, PipeReceiveStream]:
 
 async def test_pipe_typecheck() -> None:
     with pytest.raises(TypeError):
-        PipeSendStream(1.0)
+        PipeSendStream(1.0)  # type: ignore[arg-type]
     with pytest.raises(TypeError):
-        PipeReceiveStream(None)
+        PipeReceiveStream(None)  # type: ignore[arg-type]
 
 
 async def test_pipe_error_on_close() -> None:
--- a/trio/_tests/test_windows_pipes.py
+++ b/trio/_tests/test_windows_pipes.py
@@ -24,14 +24,14 @@ async def make_pipe() -> Tuple[PipeSendStream, PipeReceiveStream]:
     return PipeSendStream(w), PipeReceiveStream(r)
 
 
-async def test_pipe_typecheck():
+async def test_pipe_typecheck() -> None:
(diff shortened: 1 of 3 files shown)


trio/_tests/test_windows_pipes.py
```diff
@@ -14,24 +16,51 @@
 else:
     pytestmark = pytest.mark.skip(reason="windows only")
     pipe: Any = None
-    PipeSendStream: Any = None
-    PipeReceiveStream: Any = None
+    _handle: Any = None
+    kernel32: Any = None
+    from .._abc import ReceiveStream, SendStream
+
+    class _fake_recieve(ReceiveStream):
```
A5rocks, line 23: Hmm typo here. I'm on the edge on whether this making a class really is worth it over just assigning a `None` and type ignoring it... (or... does it even need an assignment?)

A5rocks (commented):
Looks simple enough, just one small comment

trio/_tests/test_windows_pipes.py
CoolCat467: I always forget how to spell receive, I can fix that.

CoolCat467: Fixed in bc72284c

jakkdl (approved):
Types on `_windows_pipes.py` looks trivially correct. The test seems overcomplicated and gives codecov misses, but otherwise perfectly fine as well.

trio/_tests/test_windows_pipes.py
CoolCat467: The only reason we have a subclass here is because the default `ReceiveStream` and `SendStream`'s init takes no arguments, while the `Pipe` versions require an integer.

jakkdl: wait... because `pytestmark` https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules is used, the tests aren't run on non-windows at all (that surely deserves a comment anywhere it's used IMHO), so can just skip typechecking the file on non-windows.

CoolCat467: Made to skip typechecking on non-windows in 6fc51bd

jakkdl: awesome :+1:

trio/_tests/test_windows_pipes.py
```diff
@@ -1,37 +1,47 @@
...
 else:
-    pytestmark = pytest.mark.skip(reason="windows only")
-    pipe: Any = None
-    PipeSendStream: Any = None
-    PipeReceiveStream: Any = None
+    pipe = Any
+    _handle = Any
+    kernel32 = Any
+    PipeReceiveStream = Any
+    PipeSendStream = Any
```
A5rocks, line 28: Types here shouldn't matter. In fact, we probably can delete these?

trio/_tests/test_windows_pipes.py
```diff
@@ -1,37 +1,47 @@
...
 
 from .. import _core
 from ..testing import check_one_way_stream, wait_all_tasks_blocked
 
+# Mark all the tests in this file as being windows-only
+pytestmark = pytest.mark.skipif(sys.platform != "win32", reason="windows only")
+
+assert (
+    sys.platform == "win32" or not TYPE_CHECKING
+)  # Skip type checking when not on Windows
```
A5rocks, line 16: nitpick: I think comment above might look nicer for this.

A5rocks (approved):
nitpicks but overall looks good (I don't really get the `= Any` bits)

trio/_tests/test_windows_pipes.py
CoolCat467: I wanted to delete them as well, but Mypy has failures otherwise. e74adb211d9733128c5bd337e890c9b26931356f added this to fix the errors in this run: https://github.com/python-trio/trio/actions/runs/6445455847/job/17499532948

trio/_tests/test_windows_pipes.py
CoolCat467: It was originally, but then when black formatted it it moved it below and I didn't notice. Fixed in a92b9c8

trio/_tests/test_windows_pipes.py
A5rocks: Huh, that's strange. Maybe now it will work after https://github.com/python-trio/trio/pull/2812/commits/ee64fe3a68f25f6ba62827d2dd3d5f7087f14d2f

CoolCat467: Seems to work now after removing in 5c4529c, good catch!

jakkdl:
agh, why is codecov being so dumb. It's incredibly frustrating that you can't trust the warnings displayed in the `files changed` view

PR #2813, opened 2023-10-08
Enable flake8-type-checking rule

CoolCat467:
This PR enables the ruff flake8-type-checking rule (TCH).
From [flake8-type-checking's description](https://pypi.org/project/flake8-type-checking/), the main benefit of this would be avoiding the overhead of type hint imports at runtime.


TeamSpen210:
I would argue against this, type hints are useful to have at runtime. This will break Sphinx for instance.

jakkdl:
> I would argue against this, type hints are useful to have at runtime. This will break Sphinx for instance.

is Sphinx == ReadTheDocs here? There are (only) 4 errors, or is the signatures in the documentation broken?

Digging around a bit, it seems you can get Sphinx to run with `TYPE_CHECKING = True`?
https://github.com/sphinx-doc/sphinx/blob/bb74aec2b6aa1179868d83134013450c9ff9d4d6/sphinx/ext/autodoc/importer.py#L107

CoolCat467:
pre-commit.ci autofix

jakkdl:
RTD build is failing, which I'm pretty sure it's not on master, so something is being messed up somewhere.

CoolCat467:
> RTD build is failing, which I'm pretty sure it's not on master, so something is being messed up somewhere.

Errors I'm seeing in the most recent run RTD had:
```
trio/__init__.py:docstring of trio.run_process:1: WARNING: py:class reference target not found: StrOrBytesPath
trio/__init__.py:docstring of trio.run_process:1: WARNING: py:class reference target not found: StrOrBytesPath
trio/lowlevel.py:docstring of trio.lowlevel.Task.iter_await_frames:1: WARNING: py:class reference target not found: FrameType
trio/lowlevel.py:docstring of trio.lowlevel.start_guest_run:1: WARNING: py:class reference target not found: Outcome
```
Not completely sure what this means, something about something missing from docstrings? I'm sure if I looked into it deeper I could figure it out but if anyone has more information I would be glad for the advise. Otherwise I'll do some research later and we can get this taken care of.

jakkdl:
similar problems required aliases or similar in `docs/source/conf.py`, or shuffling around imports. You can run the build process locally with `make html` standing in the `docs/` directory, which should make it much quicker to try to playing around with the code and figure out what is erroring where and why.

TeamSpen210:
Sphinx isn't able to find those names when it's trying to evaluate the annotations - possibly because those functions were imported into a different class or something. Probably need to move the imports out of `TYPE_CHECKING`. It could also be caused by us importing `trio` in `conf.py` to get `__version__` - maybe change that to read the file itself, so the module isn't loaded?

CoolCat467:
https://github.com/google/pytype/issues/1066 seems to be having a similar issue with `StrOrBytesPath`

jakkdl:
I kind of expected the RTD build to fail, but apparently it runs on 3.9+, whoop whoop.

> [google/pytype#1066](https://github.com/google/pytype/issues/1066) seems to be having a similar issue with `StrOrBytesPath`

I didn't manage to figure out any way of getting `StrOrBytesPath` working, but it does not seem related to the above afaict.

CoolCat467:
> I kind of expected the RTD build to fail, but apparently it runs on 3.9+, whoop whoop.
> 
> > [google/pytype#1066](https://github.com/google/pytype/issues/1066) seems to be having a similar issue with `StrOrBytesPath`
> 
> I didn't manage to figure out any way of getting `StrOrBytesPath` working, but it does not seem related to the above afaict.

I thought I would mention it, one of the only internet search results with anything potentially related. Maybe I was just looking in the wrong places, but it seems like sphinx doesn't have very much information about fixing errors.

Thank you so much for helping with this, I was definitely struggling a bit!

TeamSpen210:
For `StrOrBytesPath`, try adding it to `autodoc_type_aliases`, or we could use `autodoc_process_signature()` to expand it. That function gets passed the entire function signature as a string, so you can edit it however you like.

jakkdl:
> For `StrOrBytesPath`, try adding it to `autodoc_type_aliases`, or we could use `autodoc_process_signature()` to expand it. That function gets passed the entire function signature as a string, so you can edit it however you like.

I tried `autodoc_type_aliases` without any success

jakkdl:
teamwork! :rocket:

PR #2815, opened 2023-10-09
Fix some type issues with `trio.Path`

TeamSpen210:
After testing my own code against the new hints, I noticed that I messed up the path stubs a little. This fixes them in a more principled way, by going through the definitions in the same order that the typeshed stubs do for `pathlib.Path`. Things inherited from `PurePath` should be sync, while `Path` methods are async (some were mixed up). To ensure all the definitions are correct, I wrote a type-test which just uses every method to check it's doing the right thing.


TeamSpen210:
Also fixing an interesting docs build error caused by Python 3.12 releasing. `pathlib.Path.link_to()` was deprecated and removed, causing our link in the docs to fail. Added a manual deprecation to the docs.

jakkdl (approved):

trio/_path.py
```diff
@@ -446,6 +472,17 @@ async def with_segments(self, *pathsegments: StrPath) -> Path: ...
...
+    # Since we synthesise methods from the stdlib, this automatically will
+    # have deprecation warnings, and disappear entirely in 3.12+.
+    Path.link_to.__doc__ = """
+        Like Python 3.8-3.11's :meth:`~pathlib.Path.link_to`, but async.
+
+    :deprecated: This method was deprecated in Python 3.10 and entirely \
+    removed in 3.12. Use :meth:`hardlink_to` instead which has \
+    a more meaningful parameter order.
+    """
+
```
A5rocks, line 485: Why do we still overwrite doc with the new fix?

TeamSpen210: Well it seems good to have the deprecation warning replicated in our docs, so users looking there will be redirected to the better function.

trio/_tests/test_exports.py
```diff
@@ -494,9 +494,25 @@ def lookup_symbol(symbol):
                 missing.remove("__aiter__")
                 missing.remove("__anext__")
 
-        # intentionally hidden behind type guard
+            # __getattr__ is intentionally hidden behind type guard. That hook then
+            # forwards property accesses to PurePath, meaning these names aren't directly on
+            # the class.
         if class_ == trio.Path:
```
A5rocks, line 500: Is the indentation above a bit wonky or is that just GitHub's diff?

trio/_tests/test_exports.py
```diff
@@ -494,9 +494,25 @@ def lookup_symbol(symbol):
...
+                "name",
+                "parent",
+                "parents",
+                "parts",
+                "root",
+                "stem",
+                "suffix",
+                "suffixes",
+            }
+            assert len(extra) == before - 10
```
A5rocks, line 515: Codecov warnings here and above seem out of place. Is that just some codecov flakiness?

trio/_tests/type_tests/path.py
```diff
@@ -0,0 +1,118 @@
+"""Path wrapping is quite complex, ensure all methods are understood as wrapped correctly."""
```
A5rocks, line 1: I wish there was a nicer way to check we conform to typeshed's type hints and I think there might be other places where that would benefit us. But I'm not sure really how and it's probably worth punting that to another PR if we ever figure it out.

A5rocks (approved):
I don't see anything glaringly obvious but this is a big diff for what feels like pretty minor changes so I also glossed over a ton...

trio/_tests/test_exports.py
TeamSpen210: Yep, bad indentation.

trio/_tests/test_exports.py
TeamSpen210: It seems fine [on the Codecov site](https://app.codecov.io/gh/python-trio/trio/pull/2815?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=python-trio#c49ba6e4fd250a135d068a5402b778e7-R515), so it's probably just GitHub integration not working perfectly.

trio/_tests/type_tests/path.py
TeamSpen210: It's tricky, since we're not actually entirely conforming - we switch to async of course, and automatically replace `pathlib.Path` → `trio.Path`. Perhaps the best way might be to do things like code generation to create the tests?

jakkdl:
I think you should take this opportunity to enable type-checking on `trio/_tests/test_path` and fix any problems in it - it might have caught these errors if it was enabled originally.

trio/_tests/type_tests/path.py
A5rocks: Let's just defer on this. Let's merge this PR!

PR #2817, opened 2023-10-09
Bump RTD builder Python version

A5rocks:
This is erroring on master due to (AFAICT) trio doing magic to make Sphinx realize `link_to` exists, but then using Python 3.12's Sphinx index to find it (which obviously doesn't work).

This does raise the interesting question of whether functions that are removed will show up in the output, as we cannot expect everyone to be using Python 3.12. I'm not sure what to do about this.


TeamSpen210:
What might be better is to just hardcode the older URL for this specific link. I just disabled the link and added a deprecation warning in #2815...

A5rocks:
Ack I didn't read that PR yet. Ideally we would use an old intersphinx inventory but still link to the latest version of the Python documentation... but I'm not sure if that's possible.

Personally I would super duper rather just build our docs on 3.12 but unfortunately RTD doesn't support that yet. I think your solution is more pragmatic though (while it's probably more correct to not have a mismatch... it's not worth linking to older docs). I have to think about this a bit xd

CoolCat467 (approved):
Looks simple enough, but I feel like @TeamSpen210's solution in #2815 might be better in the end, and then we aren't dependent on "has to run docs on py3.11 only"

A5rocks:
I've gone ahead and submitted https://github.com/readthedocs/readthedocs.org/pull/10808 which might help things along

TeamSpen210:
Figured out an even better solution. We can just add an event listener which manually adds the definition into intersphinx's list of objects, with a 3.11 URL. Actually that probably would work for most of the other entries in `nitpick_ignore`.

A5rocks:
That sounds like a plan. I'm not comfortable enough with Sphinx to do that though... maybe you could? :P

webknjaz:
```console
/home/docs/checkouts/readthedocs.org/user_builds/trio/checkouts/2817/trio/__init__.py:docstring of trio.Path:1: WARNING: py:meth reference target not found: pathlib.Path.link_to
```

Didn't it get removed? It was deprecated since 3.10.

webknjaz:
How about replacing the reference with `symlink_to`?

https://webknjaz.github.io/intersphinx-untangled/docs.python.org/

webknjaz:
> What might be better is to just hardcode the older URL for this specific link.

FYI it is entirely possible to add several intersphinx entries against different python version docs.

A5rocks:
TBF this isn't really needed anymore due to fixes elsewhere, but this also (should, IIRC RTD releases on Tuesdays) bumps our builder to 3.12.

TeamSpen210:
We should wait until RTD does run on 3.12 - `link_to` will probably disappear at that point. What we could do is add some special case code to `_path.py` - if sphinx is in sys.modules or an environment var is set, unconditionally define the method?

CoolCat467:
Only one issue from read the docs build: 
```console
src/trio/__init__.py:docstring of trio.Path:1: WARNING: py:obj reference target not found: .
```

A5rocks:
Yeah I'm not sure if I even want to merge this PR anymore because it brings up so many possible issues (such as old forwarded methods not appearing in documentation). I'll just set it to WIP for a bit at least.

jakkdl:
> (such as old forwarded methods not appearing in documentatio

old forwarded issues are relatively straightforward to add exceptions for in `add_mapping` in `docs/source/conf.py` now. Though `Buffer` is failing here now

A5rocks:
This isn't going to be merged now cause it doesn't fix anything and just complicates things. Let's figure out how to handle older methods at some point in the future. Maybe in a similar way to how we might handle anything Windows-specific (if we have anything like that? Not sure.)

PR #2820, opened 2023-10-18
Type all remaining tests

jakkdl:
Decided to do a speedrun through all the remaining test files, with the help of https://github.com/JelleZijlstra/autotyping

If you see anything incorrect, you're very welcome to edit it straight away - I didn't put a lot of effort on average into the type hints, so there's probably a decent amount of them that are bad in some way or another.


trio/testing/_fake_net.py
```diff
@@ -12,9 +12,19 @@
...
+    Iterable,
+    NoReturn,
+    Optional,
+    TypeVar,
+    Union,
+    overload,
+)
 
 import attr
+from typing_extensions import Buffer, Self
```
TeamSpen210, line 27: We probably want to guard this import, since `trio.testing` is public.

trio/_ssl.py
```diff
@@ -239,9 +239,12 @@ def done(self) -> bool:
 
 _State = _Enum("_State", ["OK", "BROKEN", "CLOSED"])
 
+# TODO: variance
+T_Stream = TypeVar("T_Stream", bound=Stream)
```
TeamSpen210, line 243: I think this does need to be invariant. For `SSLStream` it could be covariant since we never use it as a return value, but `SSLListener` does return it. Could be covariant for the former and invariant for the latter, but that's just confusing.

trio/_tests/test_threads.py
```diff
@@ -426,6 +461,11 @@ async def test_run_in_worker_thread_limiter(MAX, cancel, use_default_limiter):
         # Mutating them in-place is OK though (as long as you use proper
         # locking etc.).
         class state:
+            if TYPE_CHECKING:
```
TeamSpen210, line 465: No need to guard this, they'll just set `__annotations__`?

trio/_tests/test_threads.py
```diff
@@ -500,55 +540,57 @@ async def run_thread(event):
...
 
-        def release_on_behalf_of(self, borrower):
+        def release_on_behalf_of(self, borrower: Task) -> None:
             record.append("release")
             assert borrower == self._borrower
 
-    await to_thread_run_sync(lambda: None, limiter=CustomLimiter())
+    # TODO: should CapacityLimiter have an abc or protocol so users can modify it?
+    # because currently it's `final` so writing code like this is not allowed.
+    await to_thread_run_sync(lambda: None, limiter=CustomLimiter())  # type: ignore[arg-type]
```
TeamSpen210, line 560: Obviously not something to do here, but I'd think it'd make most sense to do a protocol, with only the 6 `acquire_*`/`release_*` methods. That's all that a task needs to be limited, the others are only relevant for `CapacityLimiter`'s simple token strategy. The implementations of `acquire()`/`release()` could be included in the protocol since they defer to the `on_behalf_of` methods.

CoolCat467:
How do we fix the following errors `test_exports` raises?

>   jedi can't see the following symbols in trio.lowlevel:
  {'trio.lowlevel.RunVarToken': {'extra': {'previous_value', 'redeemed'},
                                 'missing': set()}}

>  mypy can't see the following symbols in trio.lowlevel:
  {'trio.lowlevel.RunVarToken': {'extra': {'previous_value', 'redeemed'},
                                 'missing': set()}}

trio/_core/_tests/test_guest_mode.py
```diff
@@ -505,11 +540,11 @@ async def crash_in_io(in_host):
 
     with monkeypatch.context() as m:
 
-        async def crash_in_worker_thread_io(in_host):
+        async def crash_in_worker_thread_io(in_host: InHost) -> None:
             t = threading.current_thread()
             old_get_events = trio._core._run.TheIOManager.get_events
 
-            def bad_get_events(*args):
+            def bad_get_events(*args: Any) -> object:
```
CoolCat467, line 547: Any reason not to use `object` instead of `Any`?

trio/_core/_tests/test_instrumentation.py
```diff
@@ -1,40 +1,45 @@
...
         assert task is _core.current_task()
         self.record.append(("after", task))
 
-    def after_run(self):
-        self.record.append(("after_run",))
+    def after_run(self) -> None:
+        self.record.append(("after_run", None))
 
-    def filter_tasks(self, tasks):
+    def filter_tasks(self, tasks: Container[Task]) -> Iterable[tuple[str, Task | None]]:
```
CoolCat467, line 34: Doesn't matter a whole lot, but I think `Iterable` is best used in return type if it's returning an object that defines an `__iter__` method. Even though it's more text, I think `Generator` would be more applicable here.

trio/_core/_tests/test_ki.py
```diff
@@ -4,7 +4,7 @@
 import inspect
 import signal
 import threading
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, AsyncIterator, Callable, Iterator
```
CoolCat467, line 7: I would import from `collections.abc` instead

trio/_core/_tests/test_multierror.py
```diff
@@ -391,7 +398,7 @@ def simple_filter(exc):
         gc.garbage.clear()
 
 
-def assert_match_in_seq(pattern_list, string):
+def assert_match_in_seq(pattern_list: List[str], string: str) -> None:
```
CoolCat467, line 401: Why old generic if we are using `__future__` annotations?

trio/_tests/test_highlevel_ssl_helpers.py
```diff
@@ -33,24 +41,48 @@ async def echo_handler(stream):
...
+            SocketKind,
+            int,
+            str,
+            tuple[str, int] | tuple[str, int, int, int],
+        ]
+    ]:
         return [(AF_INET, SOCK_STREAM, IPPROTO_TCP, "", self.sockaddr)]
 
-    async def getnameinfo(self, *args):  # pragma: no cover
+    async def getnameinfo(self, *args: Any) -> NoReturn:  # pragma: no cover
```
CoolCat467, line 65: `object`?

trio/_tests/test_subprocess.py
```diff
@@ -488,7 +515,7 @@ def test_waitid_eintr():
     got_alarm = False
     sleeper = subprocess.Popen(["sleep", "3600"])
 
-    def on_alarm(sig, frame):
+    def on_alarm(sig: object, frame: object) -> None:
```
CoolCat467, line 518: Isn't `sig` a signal type object and `frame` a `FrameType` (from `types`) object?

trio/_tests/test_threads.py
CoolCat467: That should be correct yes, as far as I am aware.

trio/_tests/test_threads.py
```diff
@@ -500,55 +541,57 @@ async def run_thread(event):
...
     assert isinstance(excinfo.value.__context__, KeyError)
     assert record == ["acquire", "release"]
 
 
-async def test_run_in_worker_thread_fail_to_spawn(monkeypatch):
+async def test_run_in_worker_thread_fail_to_spawn(monkeypatch: MonkeyPatch) -> None:
     # Test the unlikely but possible case where trying to spawn a thread fails
-    def bad_start(self, *args):
+    def bad_start(self: object, *args: object) -> NoReturn:
         raise RuntimeError("the engines canna take it captain")
```
CoolCat467, line 595: Unrelated to the types side of things but noticed this, could we change `canna` -> `cannot`

trio/_tests/test_timeouts.py
```diff
@@ -1,4 +1,5 @@
 import time
+from typing import Awaitable, Callable, TypeVar
```
CoolCat467, line 2: `collections.abc` for everything but TypeVar

trio/_tests/test_tracing.py
```diff
@@ -1,3 +1,5 @@
+from typing import AsyncGenerator
```
CoolCat467, line 1: `collections.abc`

trio/testing/_fake_net.py
```diff
@@ -12,9 +12,19 @@
...
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Iterable,
+    NoReturn,
+    Optional,
+    TypeVar,
+    Union,
+    overload,
+)
```
CoolCat467, line 24: `collections.abc` for a lot of these

CoolCat467 (changes requested):
Looks pretty good except a few things here and there.

CoolCat467:
All of the `collections.abc` imports I noted would have been able to be fixed automatically if #2809 would go through, with ruff's rule UP035 (https://docs.astral.sh/ruff/rules/deprecated-import/)

trio/_core/_tests/test_instrumentation.py
TeamSpen210: I prefer to generally use `Iterator`, since it leaves you open to changing the returned object to something else in the future. The fact it's a generator is an implementation detail. Unless you do want the users to use generator-specific methods.

jakkdl:
yeah I intentionally don't bother in large part because I expect it to get autofixed, but also doesn't really matter.

trio/_core/_tests/test_guest_mode.py
jakkdl: I'm pretty sure you'd get a typing error on line 516/551 in the `old_get_events(*args)` call. The high-effort solution (here and elsewhere in this PR) would be to copy the signature of it and not use `*args` at all.

trio/_core/_tests/test_ki.py
jakkdl: going to leave that for some autofixer / I don't care enough to bother fixing it manually.

trio/_core/_tests/test_multierror.py
jakkdl: uh, no reason at all. Might've been something `pyannotate` did

trio/_tests/test_highlevel_ssl_helpers.py
jakkdl: it might raise a type error about the signature not matching the superclass, not sure. But with an empty body it doesn't matter either way.

trio/_tests/test_subprocess.py
jakkdl: :+1:

trio/_tests/test_threads.py
jakkdl: lol, I suspect it might be an intentional joke - objectively the error message is pretty unhelpful. I vote to keep it :pirate_flag:

trio/_core/_tests/test_multierror.py
```diff
@@ -362,16 +369,16 @@ def catch_RuntimeError(exc):
...
     gc.collect()
     old_flags = gc.get_debug()
 
-    def make_multi():
+    def make_multi() -> NoReturn:
         # make_tree creates cycles itself, so a simple
         raise MultiError([get_exc(raiser1), get_exc(raiser2)])
 
-    def simple_filter(exc):
+    def simple_filter(exc: BaseException) -> Exception | RuntimeError:
```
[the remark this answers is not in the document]
jakkdl: yeah, I did it intentionally to mirror the function body. I realize now the proper way to do it would maybe be `(exc: T_Exception) -> T_Exception | RunTimeError`, but that's kinda overkill.

trio/testing/_fake_net.py
jakkdl: hmm, this should ideally be picked up by the `no test-requirements` CI run. Will look at it.

trio/_tests/test_threads.py
jakkdl: Created #2824

jakkdl:
> How do we fix the following errors `test_exports` raises?
> 
> > jedi can't see the following symbols in trio.lowlevel:
> > {'trio.lowlevel.RunVarToken': {'extra': {'previous_value', 'redeemed'},
> > 'missing': set()}}
> 
> > mypy can't see the following symbols in trio.lowlevel:
> > {'trio.lowlevel.RunVarToken': {'extra': {'previous_value', 'redeemed'},
> > 'missing': set()}}

(I/Somebody should update this error message to be more helpful)
This is because `dir()` doesn't see the attributes in it. I don't really know why, but setting `slots=True` seems to fix it and I don't think there's any particular reason not to.

trio/_tests/test_file_io.py
```diff
@@ -230,9 +244,9 @@ async def test_aclose_cancelled(path):
     assert f.closed
 
 
-async def test_detach_rewraps_asynciobase():
+async def test_detach_rewraps_asynciobase() -> None:
     raw = io.BytesIO()
-    buffered = io.BufferedReader(raw)
+    buffered = io.BufferedReader(raw)  # type: ignore[arg-type] # ????????????
```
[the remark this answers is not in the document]
jakkdl: thanks, that worked!

trio/_tests/test_threads.py
jakkdl: correct, thanks~

trio/testing/_fake_net.py
jakkdl: and indeed it did :tada: 
https://github.com/python-trio/trio/actions/runs/6567804909/job/17841165556?pr=2820#step:5:720
I just didn't see it because all runs were failing

trio/_core/_tests/test_multierror.py
jakkdl: should also get autofixed by ruff pyupgrade

trio/_tests/test_highlevel_ssl_helpers.py
```diff
@@ -33,24 +41,48 @@ async def echo_handler(stream):
...
         )
+        (listener,) = res
         async with listener:
-            sockaddr = listener.transport_listener.socket.getsockname()
+            # listener.transport_listener is of type Listener[Stream]
+            tp_listener: SocketListener = listener.transport_listener
+
+            sockaddr = tp_listener.socket.getsockname()
             hostname_resolver = FakeHostnameResolver(sockaddr)
             trio.socket.set_custom_hostname_resolver(hostname_resolver)
```
jakkdl, line 87: @TeamSpen210 @A5rocks anybody have an idea what to do here? Looking at `SSLListener` it seems to correctly require the generic to be a `Stream`, as it explicitly says it wraps a `Stream` and calls both `self.transport_listener.send_all` and `self.transport_listener.receive_some`. So is this test doing something completely illegal/illogical solely for the sake of testing and should just be `type: ignore'd`, or is this a limitation in the API and something `SSLStream`-like should be able to wrap a `Listener`, or something else?

trio/testing/_fake_net.py
```diff
@@ -352,31 +377,56 @@ def getsockname(self):
             assert self.family == trio.socket.AF_INET6
             return ("::", 0)
 
-    def getpeername(self):
+    def getpeername(self) -> None:
         self._check_closed()
         if self._binding is not None:
             if self._binding.remote is not None:
```
jakkdl, line 383: > trio/testing/_fake_net.py:(383:16 - 383:35): "UDPBinding" has no attribute "remote"  [attr-defined]

this method is never run in CI, so I suspect it's the code rather than the type hints that are wrong?

trio/testing/_fake_net.py
```diff
@@ -400,45 +450,49 @@ async def sendto(self, *args):
...
-    async def recvfrom(self, bufsize, flags=0):
+    async def recvfrom(self, bufsize: int, flags: int = 0) -> tuple[bytes, Any]:
         data, ancdata, msg_flags, address = await self.recvmsg(bufsize, flags)
         return data, address
 
-    async def recvfrom_into(self, buf, nbytes=0, flags=0):
+    async def recvfrom_into(
+        self, buf: Buffer, nbytes: int = 0, flags: int = 0
+    ) -> tuple[int, Any]:
         if nbytes != 0 and nbytes != len(buf):
```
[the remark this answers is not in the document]
TeamSpen210: That is correct, buffers only need the `__buffer__()` method. What you can do is `memoryview(buf).nbytes`,

trio/_core/_tests/test_instrumentation.py
TeamSpen210: Yep for `contextlib.contextmanager` it needs to be a generator, since that decorator uses `.send()`/`.throw()`. The type stubs are wrong, but changing them would require too much churn...

trio/_tests/test_threads.py
TeamSpen210: It's a Star Trek reference indeed.

trio/testing/_fake_net.py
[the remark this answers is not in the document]
jakkdl: thanks :+1:

trio/testing/_fake_net.py
jakkdl: added a bunch of asserts that will hopefully elucidate anybody trying to use this about the weirdness

trio/testing/_fake_net.py
```diff
@@ -299,6 +300,10 @@ async def sendmsg(self, *args: Any) -> int:
         ancdata = []
         flags = 0
         address = None
+
+        # This does *not* match up with socket.socket.sendmsg (!!!)
+        # https://docs.python.org/3/library/socket.html#socket.socket.sendmsg
+        # they always have (buffers, ancdata, flags, address)
```
jakkdl, line 306: I started writing coverage tests for `_fake_net` which is how I noticed this, but leaving those for another PR.

trio/_tests/test_timeouts.py
```diff
@@ -34,6 +37,8 @@ async def check_takes_about(f, expected_dur):
     # value above is exactly 128 ULPs below 1.0, which would make sense if it
     # started as a 1 ULP error at a different dynamic range.)
     assert (1 - 1e-8) <= (dur / expected_dur) < 1.5
+
+    # outcome is not typed
```
CoolCat467, line 41: Comment does not appear to be necessary any longer

CoolCat467 (changes requested):
Looks great except a few small things I noticed

jakkdl:
hrm, pre-commit.ci is running ruff on 0.1.1 whereas check.sh is running ruff with 0.1.0 - this feels destined to bring a conflict at some point.

CoolCat467:
> hrm, pre-commit.ci is running ruff on 0.1.1 whereas check.sh is running ruff with 0.1.0 - this feels destined to bring a conflict at some point.

I quite agree, we should update that soon.

CoolCat467 (approved):
Looks great, I'm sure the collections.abc things will get fixed at a later point if/when we run pyupgrade.

CoolCat467:
In reference to commit 6170384, if you look at the pre-commit run in the PR checks area, at the very bottom of the page it displays, it says
>autofix pushing was disabled in the configuration. to autofix this pr, comment "pre-commit.ci autofix" on the pull request

jakkdl:
I'll merge this soonish unless anybody has something to comment on, because this touches so much it very easily get merge conflicts. (e.g. other people adding test cases that aren't typechecked without this PR)

trio/_core/_tests/test_instrumentation.py
```diff
@@ -1,40 +1,45 @@
...
 from ... import _abc, _core
+from ...lowlevel import Task
 from .tutil import check_sequence_matches
 
 
 @attr.s(eq=False, hash=False)
-class TaskRecorder:
-    record = attr.ib(factory=list)
+class TaskRecorder(_abc.Instrument):
+    record: list[tuple[str, Task | None]] = attr.ib(factory=list)
```
A5rocks, line 15: I think you can keep the variable length tuples and just use tuple length to differentiate (I believe that narrows in mypy? If not, ignore this comment!)

Specifically `list[tuple[str] | tuple[str, Task]]`

trio/_core/_tests/test_instrumentation.py
```diff
@@ -193,26 +199,27 @@ async def main():
...
 
-    async def main():
-        record = []
+    async def main() -> None:
+        record: list[Task] = []
 
         # Changing the set of hooks implemented by an instrument after
         # it's installed doesn't make them start being called right away
-        instrument.before_task_step = record.append
+        instrument.before_task_step = record.append  # type: ignore[assignment]
```
A5rocks, line 222: I'm not sure why this is type erroring?

trio/_core/_tests/test_ki.py
```diff
@@ -145,38 +146,38 @@ def protected_manager():
 
 
 @pytest.mark.skipif(async_generator is None, reason="async_generator not installed")
-async def test_async_generator_agen_protection():
+async def test_async_generator_agen_protection() -> None:
     @_core.enable_ki_protection
-    @async_generator
-    async def agen_protected1():
+    @async_generator  # type: ignore[misc] # untyped generator
```
A5rocks, line 151: Just checking understanding, this is cause `async_generator` isn't typed?

trio/_core/_tests/test_ki.py
```diff
@@ -465,19 +466,19 @@ async def main():
             with pytest.raises(_core.Cancelled):
                 await _core.checkpoint()
 
-    _core.run(main)
+    _core.run(main_8)
```
A5rocks, line 469: TBH we should experiment with the mypy option that allows for reassignment...

trio/_core/_tests/test_ki.py
```diff
@@ -488,21 +489,22 @@ async def main():
 
 
 # Regression test for #461
-def test_ki_with_broken_threads():
+# don't know if _active not being visible is a problem
```
A5rocks, line 492: Probably more something to bring up with typeshed? But it's not documented AFAICT so probably intentional.

trio/_core/_tests/test_multierror.py
```diff
@@ -34,46 +36,47 @@ def __eq__(self, other: object) -> bool:
...
     raise NameError
 
 
-def get_exc(raiser):
+def get_exc(raiser: Callable[[], NoReturn]) -> BaseException:
     try:
         raiser()
     except Exception as exc:
         return exc
+    raise AssertionError("raiser should always raise")  # pragma: no cover
```
A5rocks, line 72: Again the comment about `-> BaseException | None` -- if mypy requires you explicitly return `None` then nvm on that.

trio/_core/_tests/tutil.py
```diff
@@ -98,9 +100,7 @@ def restore_unraisablehook() -> Generator[None, None, None]:
 
 # template is like:
 #   [1, {2.1, 2.2}, 3] -> matches [1, 2.1, 2.2, 3] or [1, 2.2, 2.1, 3]
-def check_sequence_matches(
-    seq: Sequence[object], template: Iterable[object | set[object]]
-) -> None:
+def check_sequence_matches(seq: Sequence[T], template: Iterable[T | set[T]]) -> None:
```
A5rocks, line 103: Hm. I'm not sure I quite follow why the `T` here -- but I don't remember the variance of `Iterable` so maybe this makes sense... Though I guess maybe for `set`.... aaa whatever, might as well ask.

trio/_highlevel_ssl_helpers.py
```diff
@@ -40,7 +40,7 @@ async def open_ssl_over_tcp_stream(
     data.
 
     Args:
-      host (bytes or str): The host to connect to. We require the server
+      host (str): The host to connect to. We require the server
```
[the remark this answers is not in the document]
A5rocks: Does this passing bytes error at runtime?

trio/_subprocess_platform/waitid.py
```diff
@@ -8,9 +8,14 @@
 from .._sync import CapacityLimiter, Event
 from .._threads import to_thread_run_sync
 
-assert (sys.platform != "win32" and sys.platform != "darwin") or not TYPE_CHECKING
+if TYPE_CHECKING:
+
+    def sync_wait_reapable(pid: int) -> None:
```
A5rocks, line 13: Mind leaving a comment about why this function specifically is up here?

A5rocks (approved):
Well, I skipped all the test files after the actual code changes because... I mean they're just tests. Comments but none of them block anything.

trio/_core/_tests/test_instrumentation.py
TeamSpen210: It does as of last week - https://github.com/python/mypy/pull/16237. So not quite yet.

trio/_subprocess_platform/waitid.py
TeamSpen210: I think it's to provide a definition if imported on other platforms.

TeamSpen210 (approved):

trio/_core/_tests/test_ki.py
CoolCat467: The reassignment option can open the door to lots of issues though. I will say it is convenient in a lot of cases to simply reassign something, and it can become a bit of type gymnastics with especially complex code, but I think that in the end leaving it off might be the best option. I feel like a lot of subtle issues might get suppressed with the reassignment option enabled.

CoolCat467:
pre-commit.ci autofix

trio/_highlevel_ssl_helpers.py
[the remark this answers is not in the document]
jakkdl: strange, can't repro any problem when reverting this edit.

trio/_core/_tests/test_guest_mode.py
```diff
@@ -450,7 +483,11 @@ async def trio_main():
                 aio_task.cancel()
                 return "trio-main-done"
 
-    async def aio_pingpong(from_trio, to_trio):
+        raise AssertionError("should never be reached")  # pragma: no cover
```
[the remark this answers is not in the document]
jakkdl: `-> str|None` would still give "missing return statement", if you have an explicit return somewhere in the function it wants explicit returns in all branches.

trio/_core/_tests/test_instrumentation.py
jakkdl: I'll leave it as is then.

trio/_core/_tests/test_instrumentation.py
jakkdl:
```
trio/_core/_tests/test_instrumentation.py:222: error: Cannot assign to a method  [method-assign]
trio/_core/_tests/test_instrumentation.py:222: error: Incompatible types in assignment (expression has type "Callable[[Task], None]", variable has type "Callable[[Arg(Task, 'task')], None]")  [assignment]
```

If I edit `_abc.Instrument.before_task_step` to make `task` positional-only then I only get the first error. Seems like `ignore[assignment]` covers both, but made it extra explicit by ignoring both, and added comment that explains the errors.

trio/_core/_tests/test_ki.py
jakkdl: yeah, clarifying comment

trio/_core/_tests/test_ki.py
jakkdl: Hmm, it's only `_active` and not `__active`, so I'm pretty sure a lot of code actually inspects it or modifies it in a subclass, but yeah makes sense that typeshed sticks to the documented parts.

trio/_core/_tests/tutil.py
jakkdl: I.. don't quite follow why the signature would be confusing? Though the comment above reversing them might be the cause of confusion.
You pass it a sequence of some thing, e.g. `[1, 2, 3, 4, 5]` or `[1, 2, 3, 5, 4]`, and use this function to see whether that matches a template - which can include either objects of the same type, or if you expect the order to sometimes be jumbled you can specify them inside a set*: `[1, 2, 3, {4, 5}]`. So the template should be an `Iterable` which contains objects of the same type your sequence consists of, or sets of that object. Variance is not related in any way.

The `Sequence` vs `Iterable` is because the function does slicing on `seq`, but only iterates over `template`.

\* I did not expect this to be how the function actually worked, and wrote an incorrect description at first where I assumed the sets were about unions. Not a great name imho.

Improved the comments accompanying the function.

trio/_subprocess_platform/waitid.py
jakkdl: It's for `trio/_tests/test_subprocess.py:test_waitid_eintr()`:
```python
@slow
def test_waitid_eintr() -> None:
    # This only matters on PyPy (where we're coding EINTR handling
    # ourselves) but the test works on all waitid platforms.
    from .._subprocess_platform import wait_child_exiting

    if TYPE_CHECKING and sys.platform == "win32":
        return

    if not wait_child_exiting.__module__.endswith("waitid"):
        pytest.skip("waitid only")
    from .._subprocess_platform.waitid import sync_wait_reapable

    ...
```

where despite having a
`if TYPE_CHECKING and sys.platform == "win32": return` it will still complain about the import on line 522. I guess this one could be reported upstream?

Though I realize now it's probably better to silence the error in the test than to make people think they can import `sync_wait_reapable` globally on all platforms only to get an `ImportError` at runtime.

trio/_tests/test_exports.py
```diff
@@ -445,6 +446,16 @@ def lookup_symbol(symbol):
             extra = {e for e in extra if not e.endswith("AttrsAttributes__")}
             assert len(extra) == before - 1
 
+        import enum
```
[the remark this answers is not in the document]
jakkdl: nope. moving it up.

trio/_tests/test_highlevel_serve_listeners.py
```diff
@@ -1,29 +1,45 @@
...
+    MemoryReceiveStream,
+    MemorySendStream,
+    MockClock,
+    memory_stream_pair,
+    wait_all_tasks_blocked,
+)
+
+# types are somewhat tentative - I just bruteforced them until I got something that didn't
+# give errors
+TypeThing = StapledStream[MemorySendStream, MemoryReceiveStream]
```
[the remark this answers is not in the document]
jakkdl: Also found two more places where it could be used in the file.

trio/_tests/test_ssl.py
```diff
@@ -116,10 +116,8 @@ def ssl_echo_serve_sync(
                     # respond in kind but it's legal for them to have already
                     # gone away.
                     exceptions = (BrokenPipeError, ssl.SSLZeroReturnError)
```
TeamSpen210, line 118: Putting these into a tuple only to immediately unpack seems a little redundant.

jakkdl: indeed

trio/_core/_tests/tutil.py
A5rocks: Well the `T` w/ `Sequence` (covariant) means that something like this would pass typechecking: `check_sequence_matches([5.0, 4.0], [5, 4])` but I was in fact being blind to the fact that `object` has that problem too.

jakkdl: Ah, huh. Yeah I don't think we really care about that level of precision, as it's only really used in a few tests and not a public function. I need to get better at variance, for sure.

PR #2821, opened 2023-10-19
Enable flake8-pyi

CoolCat467:
"But wait CoolCat, why do we need pyi if we aren't shipping stubs?"
Because some of the rule checks it does look like they might help notice bugs in PRs in process.
Full rule list section: https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ select = [
   "F",  # pyflakes
   "I",  # isort
   "YTT",  # flake8-2020
+  "PYI",  # flake8-pyi
 ]
 extend-ignore = [
     'F403',  # undefined-local-with-import-star
--- a/trio/_channel.py
+++ b/trio/_channel.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
 
 
 def _open_memory_channel(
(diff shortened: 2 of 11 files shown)


A5rocks:
>   `trio/_tests/test_socket.py:31:5: PYI042 Type alias `getaddrinfoResponse` should be CamelCase`

Who's going to tell them that's pascal case, not camel case 😰

TeamSpen210:
The float | int unions should be kept. They’re integer parameters, except they also allow `math.inf`. Maybe we should have a type alias defined somewhere to make that more explicit.

jakkdl:
> The float | int unions should be kept. They’re integer parameters, except they also allow `math.inf`. Maybe we should have a type alias defined somewhere to make that more explicit.

+1, esp as we might possibly want to move to `int|None` at some point in the future with corresponding API deprecations.

CoolCat467:
Planning to do another PR soon to make it `int | None` eventually, for the time being I am having the errors be ignored.

jakkdl (approved):

TeamSpen210 (approved):

PR #2822, opened 2023-10-19
Enable flake8-pytest-style

CoolCat467:
This PR enables the flake8-pytest-style ruff rule.


TeamSpen210:
Setting the [fixture-parentheses](https://docs.astral.sh/ruff/settings/#flake8-pytest-style-fixture-parentheses) option to `false` in `pyproject.toml` will make Ruff use `@pytest.fixture`, and instead warn if empty parentheses are used. It's probably better to be configuring linters to match project conventions, when the existing code is perfectly correct.

pyproject.toml
```diff
@@ -135,6 +136,9 @@ extend-exclude = [
 [tool.ruff.isort]
 combine-as-imports = true
 
+[tool.ruff.flake8-pytest-style]
+fixture-parentheses = false
```
webknjaz, line 140: :tada:

I also like the setting that forces using a sequence of strings in parametrize rather than a comma-separated string...

trio/_core/_tests/test_instrumentation.py
```diff
@@ -255,7 +255,7 @@ def after_run(self) -> NoReturn:
             raise ValueError("oops")
 
     async def main() -> None:
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="oops"):
```
webknjaz, line 258: People keep forgetting that it's a regex. Could we make that explicit?
```suggestion
        with pytest.raises(ValueError, match="^oops$"):
```

trio/_core/_tests/test_io.py
```diff
@@ -319,7 +319,7 @@ async def test_wait_on_invalid_object() -> None:
             fileno = s.fileno()
         # We just closed the socket and don't do anything else in between, so
         # we can be confident that the fileno hasn't be reassigned.
-        with pytest.raises(OSError):
+        with pytest.raises(OSError, match="Bad file descriptor"):
```
webknjaz, line 322:
```suggestion
        with pytest.raises(OSError, match="^Bad file descriptor$"):
```

trio/_tests/tools/test_mypy_annotate.py
```diff
@@ -13,7 +13,7 @@
 
 
 @pytest.mark.parametrize(
-    "src, expected",
+    ("src", "expected"),
```
webknjaz, line 16: :rocket:

trio/_tests/test_windows_pipes.py
```diff
@@ -45,9 +45,9 @@ async def test_pipe_error_on_close() -> None:
     assert kernel32.CloseHandle(_handle(r))
     assert kernel32.CloseHandle(_handle(w))
 
-    with pytest.raises(OSError):
+    with pytest.raises(OSError, match="TODO: replace with exception"):
         await send_stream.aclose()
-    with pytest.raises(OSError):
+    with pytest.raises(OSError, match="TODO: replace with exception"):
```
webknjaz, line 50: Could you turn all the `match=` arg values into full start-to-end matching regexps?

CoolCat467:
Interesting, the pypy-3.9 runner is segfaulting inside pytest from a `from .test_socket import setsockopt_tests` line. Any ideas?

src/trio/_tests/test_socket.py
```diff
@@ -819,7 +834,8 @@ async def test_address_in_socket_error() -> None:
         try:
             await sock.connect((address, 2))
         except OSError as e:
-            assert any(address in str(arg) for arg in e.args)
+            # the noqa is "Found assertion on exception `e` in `except` block"
+            assert any(address in str(arg) for arg in e.args)  # noqa: PT017
```
jakkdl, line 838: It wants you to do this:
```python
with pytest.raises(OSError, match=...) as excinfo:
  ...
assert any(... for arg in excinfo.value.args)
```

Looks like there's a few more instances of PT017 you can fix similarly

src/trio/_tests/test_socket.py
```diff
@@ -1116,7 +1132,8 @@ async def test_many_sockets() -> None:
         try:
             a, b = stdlib_socket.socketpair()
         except OSError as e:  # pragma: no cover
-            assert e.errno in (errno.EMFILE, errno.ENFILE)
+            # the noqa is "Found assertion on exception `e` in `except` block"
+            assert e.errno in (errno.EMFILE, errno.ENFILE)  # noqa: PT017
```
jakkdl, line 1136: .

src/trio/_core/_tests/test_run.py
```diff
@@ -140,7 +140,8 @@ async def erroring() -> NoReturn:
         async with _core.open_nursery() as nursery:
             nursery.start_soon(erroring)
     except ValueError as e:
-        assert e is exc
+        # the noqa is for "Found assertion on exception `multi_exc` in `except` block"
+        assert e is exc  # noqa: PT017
```
jakkdl, line 145: .

src/trio/_core/_tests/test_run.py
```diff
@@ -456,7 +457,8 @@ async def crasher() -> NoReturn:
                 # nursery block exited, all cancellations inside the
                 # nursery block continue propagating to reach the
                 # outer scope.
-                assert len(multi_exc.exceptions) == 4
+                # the noqa is for "Found assertion on exception `multi_exc` in `except` block"
+                assert len(multi_exc.exceptions) == 4  # noqa: PT017
```
jakkdl, line 461: oh derp, I should've thought of this one myself. You can leave this one as a `noqa` though and let me handle it in the multierror PRs

src/trio/_core/_tests/test_multierror.py
```diff
@@ -136,7 +136,7 @@ async def test_MultiErrorNotHashable() -> None:
     assert exc1 != exc2
     assert exc1 != exc3
 
-    with pytest.raises(MultiError):
+    with pytest.raises(MultiError):  # noqa: PT012
```
jakkdl, line 139: PT012 is a bit of a pain when `trio` so often tests context managers, so not entirely sure it's worth having enabled. It did catch a couple ugly cases this one time though.

src/trio/_tests/test_highlevel_socket.py
```diff
@@ -283,7 +289,7 @@ async def accept(self) -> tuple[SocketType, object]:
 
     for code in [errno.EMFILE, errno.EFAULT, errno.ENOBUFS]:
         with assert_checkpoints():
-            with pytest.raises(OSError) as excinfo:
+            with pytest.raises(OSError) as excinfo:  # noqa: PT011  # missing `match`
                 await listener.accept()
             assert excinfo.value.errno == code
```
jakkdl, line 294:
```python
for code, match in zip((errno.EMFILE, errno.EFAULT, errno.ENOBUFS), ("Out of file descriptors", "attempt to write to read-only memory", "out of buffers"):
  with assert_checkpoints():
    with pytest.raises(OSError, match=match) as excinfo:
      ...
    ...
```
should work?

src/trio/_tests/test_socket.py
```diff
@@ -648,11 +648,20 @@ async def res(
                 )
                 netlink_sock.close()
 
-            with pytest.raises(ValueError):
+            with pytest.raises(
+                ValueError,
+                match=r"^address should be a \(host, port(, \[flowinfo, \[scopeid\]\])*\) tuple$",
```
jakkdl, line 653: could save this in a string variable and reuse it

src/trio/_tests/test_subprocess.py
```diff
@@ -339,15 +338,30 @@ async def test_run() -> None:
     # invalid combinations
     with pytest.raises(UnicodeError):
         await run_process(CAT, stdin="oh no, it's text")
-    with pytest.raises(ValueError):
+    with pytest.raises(
+        ValueError,
+        match=r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$",
```
jakkdl, line 343: this string also makes sense to save in a variable and reuse

jakkdl (commented):
A couple nits, but overall great!

Going to be a ton of merge conflicts though :upside_down_face:

jakkdl:
> Interesting, the pypy-3.9 runner is segfaulting inside pytest from a `from .test_socket import setsockopt_tests` line. Any ideas?
> 
> Edit: https://github.com/python-trio/trio/actions/runs/7054572593/job/19203654087

it worked on a rerun... :woman_shrugging:

src/trio/_tests/test_highlevel_socket.py
webknjaz: By the way, it could be useful to have `@pytest.mark.parametrize` or `pytest-subtests` here so that separate checks would show up in the test report distinctly.

webknjaz:
> > Interesting, the pypy-3.9 runner is segfaulting inside pytest from a `from .test_socket import setsockopt_tests` line. Any ideas?
> > Edit: [python-trio/trio/actions/runs/7054572593/job/19203654087](https://github.com/python-trio/trio/actions/runs/7054572593/job/19203654087)
> 
> it worked on a rerun... 🤷‍♀️

I've been observing flaky PyPy 3.9 segfaults on GHA in other projects FTR.

src/trio/_tests/test_socket.py
CoolCat467: This one is quite interesting actually. This one is a loop that breaks prematurely only if the kernel responds with an `Out of socket file ids` error.

src/trio/_tests/test_subprocess.py
CoolCat467: Changed in f89cd0a

src/trio/_tests/test_socket.py
CoolCat467: Changed in f89cd0a

src/trio/_tests/test_highlevel_socket.py
CoolCat467: Changed in f89cd0a. I didn't do the parameterize because it would be a lot of setup code re-running, but if this is important I can still change it.

src/trio/_core/_tests/test_run.py
CoolCat467: Changed in f89cd0a

src/trio/_tests/test_socket.py
CoolCat467: Changed in f89cd0a

src/trio/_tests/test_socket.py
jakkdl: hmm, this is a weird test. it loops `_x` until it gets `total//2-1` unless it throws an exception - ~~in which case it'd break and then the test would fail later in the last~~ nowait that's just a print! That print should totally be inside the `except` catcher and `_x` can be renamed to `_`. Probably also a `pragma: no cover`? codecov is acting weird so idk if it's getting covered or not.

Okay, so, it loops until it fails to create a socket - or it hits `total//2`. So supposedly it's not an error if it raises earlier, nor is it a problem if it doesn't raise, which means that `pytest.raises` is not relevant.

It'd maybe be nice to add some comments. Why `pytest.raises` isn't relevant, and maybe also nice to add that EMFILE is "Too many open files" and ENFILE is "File table overflow".

jakkdl (approved):
Can *probably* ignore the codecov check failing, but it might be nice to make use of it to check out what's not properly getting run - and figure out if it should have `pragma: no cover` or if you messed with some check that now doesn't run anymore.

src/trio/_tests/test_socket.py
CoolCat467: Fixed I think with 48b771b4

jakkdl:
I fixed the codecov fails - this PR didn't touch any of those files, but it changed the length of the files which modified the % of misses in the files, which caused it to fail :upside_down_face:

So, unrelated, but:
@agronholm looking at `git blame` it looks like your code for testing `open_tcp_listeners` have had dead code since inception. It always has an `ExceptionGroup` as the `__cause__`:
https://github.com/python-trio/trio/blob/60b9e1423b5f1a80798e214af96926a80c7d94d2/src/trio/_highlevel_open_tcp_listeners.py#L166-L173
but the code testing it (that I removed in my commit) has a branch for if the `__cause__` is not wrapped in an exceptiongroup:
https://github.com/python-trio/trio/blob/master/src/trio/_tests/test_highlevel_open_tcp_listeners.py#L326-L335

I don't really see much any upside to changing the behaviour, but that could be done easily, so I opted to simply remove the redundant test code.

jakkdl:
ah no, it was codecov apparently being smart enough to figure out that the assertion in the test was irrelevant given that the `raises` now has a `match=` ?!

jakkdl:
I would like to merge this unless somebody else wants to review

TeamSpen210 (approved):
Looks all good. Perhaps we might want to enable the other config too to prevent parentheses in `@pytest.mark.whatever()` also, but not important.

CoolCat467:
> Perhaps we might want to enable the other config too to prevent parentheses in `@pytest.mark.whatever()` also, but not important.

I think this PR is a bit big, if we would like to later I think another PR would be best.

CoolCat467:
Merging, we have two approvals.

PR #2823, opened 2023-10-19
Enable ruff flake8-simplify rule

CoolCat467:
This PR enables the ruff flake8-simplify rule (SIM) and fixes all of the new errors from enabling it.


trio/_core/_multierror.py
```diff
@@ -328,7 +328,7 @@ def catch(
         return MultiErrorCatcher(handler)
 
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # noqa: SIM108
```
A5rocks, line 331: I agree ternaries don't make sense for `if TYPE_CHECKING` blocks but do type checkers even support it?

Purely out of curiosity!

trio/_core/_run.py
```diff
@@ -792,15 +792,19 @@ def cancel_called(self) -> bool:
...
+            or not self._has_been_entered
+            and not self._cancel_called
+            and current_time() >= self._deadline
+        ):
             # Scope is active or not yet entered: make sure cancel_called
             # is true if the deadline has passed. This shouldn't
             # be able to actually change behavior, since we check for
             # deadline expiry on scope entry and at every checkpoint,
             # but it makes the value returned by cancel_called more
             # closely match expectations.
```
A5rocks, line 806: Comment applies to the 2 conditions in the `if` statement, not the code

trio/_core/_tests/test_asyncgen.py
```diff
@@ -44,14 +43,14 @@ async def async_main() -> None:
         with pytest.warns(
             ResourceWarning, match="Async generator.*collected before.*exhausted"
         ):
-            assert 42 == await example("abandoned").asend(None)
+            assert await example("abandoned").asend(None) == 42
```
A5rocks, line 46: I don't get how this simplifies anything?

trio/_core/_tests/test_multierror.py
```diff
@@ -283,7 +283,9 @@ def noop(_):
 
     # Simple pass-through of all exceptions
     m = make_tree()
-    with pytest.raises(MultiError) as excinfo:
+    with pytest.raises(  # noqa: SIM117  # multiple-with-statements
```
A5rocks, line 286: I think combining the with statements (here and before) is probably fine, even if slightly less readable.

trio/_core/_tests/test_run.py
```diff
@@ -621,12 +619,11 @@ async def test_basic_timeout(mock_clock: _core.MockClock) -> None:
...
 async def test_cancel_scope_nesting() -> None:
     # Nested scopes: if two triggering at once, the outer one wins
-    with _core.CancelScope() as scope1:
-        with _core.CancelScope() as scope2:
-            with _core.CancelScope() as scope3:
-                scope3.cancel()
-                scope2.cancel()
-                await sleep_forever()
+    with _core.CancelScope() as scope1, _core.CancelScope() as scope2:  # noqa: SIM117  # multiple-with-statements
+        with _core.CancelScope() as scope3:
```
A5rocks, line 623: Why combine 2 and leave one behind?

trio/_subprocess.py
```diff
@@ -171,7 +172,11 @@ def __init__(
             else:
                 # It worked! Wrap the raw fd up in a Python file object to
                 # make sure it'll get closed.
-                self._pidfd = open(fd)
+                # fmt: off
+                self._pidfd = open(  # noqa: SIM115  # open-file-with-context-handler
```
A5rocks, line 176: I think ruff shouldn't error on this case, this is just silly. Not every class should be constructable solely using a context manager.

trio/_tests/test_highlevel_open_tcp_stream.py
```diff
@@ -250,7 +245,7 @@ def __init__(
         assert len(set(ip_order)) == len(ip_list)
         ip_dict: dict[str | int, tuple[float, str]] = {}
         for ip, delay, result in ip_list:
-            assert 0 <= delay
+            assert delay >= 0
```
A5rocks, line 248: I don't get this.

trio/_tests/test_path.py
```diff
@@ -53,7 +53,7 @@ async def test_magic() -> None:
 async def test_cmp_magic(cls_a: EitherPathType, cls_b: EitherPathType) -> None:
     a, b = cls_a(""), cls_b("")
     assert a == b
-    assert not a != b
+    assert a == b
```
A5rocks, line 56: This is testing `!=` which is its own method on `a`

A5rocks (commented):
I think overall this set of lints is kinda silly. Horizontal space really really shouldn't matter, especially with an autoformatter.

I can kinda see how a few might be helpful (e.g. `suppress`... and maybe multiple `with` statements) but talking about the multiple `with` statements thing, the stance here is super inconsistent.

Finally, I really don't get the rule that reorders the `==` or `>=` things.

trio/_core/_multierror.py
CoolCat467: I'm not completely sure off the top of my head, hence the noqa I added. I would assume not unless they are doing some very fine grained logic analysis.

trio/_core/_tests/test_asyncgen.py
jakkdl: I like this as a general style thing, though I think the developers of this repo already adheres to it decently well. E.g. code like
```python
if (a == 3 || 4 == b):
  ...
```
is very silly, and in spirit feels like something `Black` would enforce - except they are explicitly AST-neutral. So I think there's good that there's a tool that enforces it.

jakkdl: Oh, I actually very much prefer the changes to this set of comparisons. Because they go against the standard, I was thinking maybe it's because the developer wants to place an emphasis on the left term because the right term is constant ... but no the literal is just a random magic number and the expression on the right hand side very much does change.

trio/_core/_tests/test_guest_mode.py
```diff
@@ -593,10 +594,8 @@ async def agen(label):
             yield 1
         finally:
             library = sniffio.current_async_library()
-            try:
+            with contextlib.suppress(trio.Cancelled):
```
jakkdl, line 597: TIL. Me like

trio/_subprocess.py
jakkdl: Yeah this one could be reported upstream. Why is the `# fmt: off` needed ?

trio/_tests/test_dtls.py
```diff
@@ -173,7 +170,7 @@ async def route_packet(packet: UDPPacket) -> None:
                         break
 
             def route_packet_wrapper(packet: UDPPacket) -> None:
-                try:
+                try:  # noqa: SIM105  # suppressible-exception
```
jakkdl, line 173: loss of coverage information on whether a suppressed except is caught or not is actually a pretty good reason not to enable this check.

trio/_tests/test_dtls.py
```diff
@@ -292,9 +289,8 @@ async def null_handler(_: object) -> None:  # pragma: no cover
 
 
 async def test_dtls_over_dgram_only() -> None:
-    with trio.socket.socket() as s:
-        with pytest.raises(ValueError):
-            DTLSEndpoint(s)
+    with trio.socket.socket() as s, pytest.raises(ValueError):
```
jakkdl, line 292: this is definitely worse imho, even though the original isn't super pretty either

trio/_tests/test_highlevel_open_tcp_stream.py
jakkdl: I like the change

trio/_core/_multierror.py
jakkdl: ternaries aren't actually very complex AST-wise (pretty much identical to handling `if`-statements), so I wouldn't be shocked if it did work in some of them. But I could see them intentionally not supporting them.

jakkdl (changes requested):
I think the `combine-with` and `combine-if` checks should be disabled and the changes reverted:
1. They suppress coverage info
2. when the ifs/withs are doing completely different types of things, squashing them together makes them seem related - which they aren't. This is especially bad with `pytest.raises` + testing a with statement, and just in general with a library like Trio where careful thinking about scopes is really important.

trio/_subprocess.py
CoolCat467: `# fmt: off` needed because basically ruff and black end up fighting over the formatting if I remember correctly. I think it's because of differences in the way the two handle trying to make it not exceed a line number limit.

trio/_tests/test_dtls.py
CoolCat467: Just looking at the diff for this particular one, yes I completely agree.

trio/_core/_run.py
CoolCat467: Reverted as of 76d741357e82eed4ffd8de3e540804ca5346a5b3

trio/_core/_multierror.py
CoolCat467: Allowed to change into ternary in most recent commit as of now

trio/_core/_run.py
CoolCat467: Reverted as of most recent commit

trio/_core/_tests/test_multierror.py
CoolCat467: Disabled combining with statements in 76d7413

trio/_core/_tests/test_run.py
CoolCat467: Disabled combining with statements in 76d7413

trio/_tests/test_path.py
CoolCat467: Fixed as of 76d7413

trio/_tests/test_dtls.py
CoolCat467: Disabled combining with statements in 76d7413

CoolCat467:
> I think the `combine-with` and `combine-if` checks should be disabled and the changes reverted:
> 
>     1. They suppress coverage info
> 
>     2. when the ifs/withs are doing completely different types of things, squashing them together makes them seem related - which they aren't. This is especially bad with `pytest.raises` + testing a with statement, and just in general with a library like Trio where careful thinking about scopes is really important.

Reverted as of 76d7413

trio/_core/_multierror.py
```diff
@@ -328,10 +328,7 @@ def catch(
         return MultiErrorCatcher(handler)
 
 
-if TYPE_CHECKING:  # noqa: SIM108
-    _ExceptionGroup = ExceptionGroup[Exception]
-else:
-    _ExceptionGroup = ExceptionGroup
+_ExceptionGroup = ExceptionGroup[Exception] if TYPE_CHECKING else ExceptionGroup
```
A5rocks, line 331: FYI latest ruff will not do this, so this change is fine even if type checkers don't like it.

CoolCat467: Latest ruff does still apply interestingly enough, so I added a noqa for it.

CoolCat467: Fixed in dd7b0ed

trio/_subprocess.py
jakkdl: that *also* sounds like something that should be reported upstream :sweat_smile: 

if we're `fmt: off`ing then I think `fd)` should be moved onto line 176, kinda silly to split it up this much.

jakkdl (approved):
Minor formatting change, and I think you should open upstream issues. Otherwise looks great~

jakkdl:
I think we can merge this unless @A5rocks or others have any comments.

TeamSpen210 (approved):
Looks like improvements, though we might want to use a bit of caution with `contextlib.suppress`. It's definitely tidier, but I imagine it could have a performance impact.

PR #2826, opened 2023-10-23
Don't double-wrap exceptions in nurseries

### LINKED ISSUE #2611: `strict_exception_groups=True` "leaks" the implementation-detail nursery in `nursery.start(fn)`
Currently, exceptions raised from a `nursery.start()`'ed function will be wrapped in *extra* `ExceptionGroup` (aka `MultiError`), *iff* it's raised before `task_status.started()` is called.

I think that for consistency, we should pull such exceptions (which might themselves be `ExceptionGroup`s) out of the group added by the starter-nursery.

```python
import trio

async def startable_fn_which_doesnt_contain_a_nursery(task_status):
    if ...:
        print("Something went wrong")
        raise RuntimeError  # this might be wrapped in an ExceptionGroup
    task_status.started()
    raise ValueError  # ...but this will never be wrapped!

async def main():
        try:
            # Uh-oh!  With strict_exception_groups=True, the hidden inner nursery
            # in `nursery.start` wraps our RuntimeError in an ExceptionGroup :-/
            await nursery.start(startable_fn_which_doesnt_contain_a_nursery)
        except RuntimeError:
            print("No worries, we'll try something else.")

if __name__ == "__main__":
    trio.run(main, strict_exception_groups=False)  # passes
    print("\n==============================================================\n")
    trio.run(main, strict_exception_groups=True)  # fails!
```

Related to #2544 and possibly #1457.

jakkdl:
Fixes #2611 

I should probably remove the extraneous tests, and merge it into one of the existing test files, or at the very least rename the test file.

@Zac-HD suggested
> "if we're using strict exception groups and the inner nursery raises an exception group with a single exception in it (subtree, not necessarily leaf!) unwrap it concat the tracebacks" 

and I suppose that's maybe slightly different behavior from what I implemented?
I added a test case for when the child case itself raises an `ExceptionGroup`, should the traceback be modified in any way in that case or is it fine to just pass it through?


Zac-HD:
I think it's fine to pass through.

For the tests, there are currently enough separate cases that I can imagine missing that something is backwards.  Could we write something parameterized, such that it obviously does each possible set of operations, and then the asserts only depend on whether we're in strict mode?  (I expect this to be harder to read individually, but the confidence in consistency is what I'm after)

Zac-HD (approved):
This looks great to me - the tests still take a moment of thought, but suffice to convince me that it's now correct in general 🎉 

The CI failures look like one flake, and one mypy issue to fix - `trio/_core/_tests/test_run.py:2563: Missing type parameters for generic type "ExceptionInfo"  [type-arg]`, I think `ExceptionInfo[BaseException]` would be easiest - and once that's done let's merge!

A5rocks:
Oh just realized this should probably also contain a newsfragment?

trio/_core/_tests/test_run.py
```diff
@@ -2538,3 +2538,62 @@ async def test_cancel_scope_no_cancellederror() -> None:
...
+    raise ValueError
+
+
+async def raise_custom_exception_group_before(
+    *, task_status: _core.TaskStatus[None]
+) -> None:
+    raise ExceptionGroup("my group", [ValueError()])
+
+
+def _check_exception(exc: pytest.ExceptionInfo) -> None:
```
graingert, line 2563:
```suggestion
def _check_exception(exc: pytest.ExceptionInfo[BaseException]) -> None:
```

trio/_core/_tests/test_run.py
```diff
@@ -2538,3 +2538,62 @@ async def test_cancel_scope_no_cancellederror() -> None:
...
+# it was only when run from `trio.run` that the double wrapping happened
+@pytest.mark.parametrize("strict", [False, True])
+@pytest.mark.parametrize(
+    "raiser", [raise_before, raise_after_started, raise_custom_exception_group_before]
+)
+def test_trio_run_strict_before_started(
+    strict: bool, raiser: Callable[[], Awaitable[None]]
+) -> None:
+    expect_group = strict or raiser is raise_custom_exception_group_before
+    with pytest.raises(BaseExceptionGroup if expect_group else ValueError) as exc:
```
graingert, line 2596: I'd be tempted to split this into two separate testes

trio/_core/_tests/test_run.py
```diff
@@ -2538,3 +2538,62 @@ async def test_cancel_scope_no_cancellederror() -> None:
...
+    async with _core.open_nursery(strict_exception_groups=strict) as nursery:
+        await nursery.start(raiser)
+
+
+@pytest.mark.parametrize("strict", [False, True])
+@pytest.mark.parametrize("raiser", [raise_before, raise_after_started])
+async def test_strict_before_started(
+    strict: bool, raiser: Callable[[], Awaitable[None]]
+) -> None:
+    with pytest.raises(BaseExceptionGroup if strict else ValueError) as exc:
```
graingert, line 2581: I'd be tempted to split this into two tests otherwise the test is too similar logically to the code change

trio/_core/_tests/test_run.py
graingert: ah I see now in review this was specifically requested to be parametrized

trio/_core/_tests/test_run.py
graingert: ah I see now in review this was specifically requested to be parametrized

jakkdl:
Ah shoot, forgot the newsfragment. Opening a new PR for it

PR #2828, opened 2023-10-23
fix B904

jakkdl:
this snuck in with some PR, I blame having half a dozen tests failing currently.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -39,7 +39,7 @@ if ! ruff check .; then
     echo "* ruff found issues." >> "$GITHUB_STEP_SUMMARY"
     EXIT_STATUS=1
     if $ON_GITHUB_CI; then
-        ruff check --format github --diff .
+        ruff check --output-format github --diff .
     else
         ruff check --diff .
     fi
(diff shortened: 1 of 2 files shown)


A5rocks:
I'm confused cause formatting check should block merge nowadays! But whatever.

A5rocks:
(could you update `ruff check` invocation to use `--output-format` instead of `--format`?)

A5rocks (approved):
Looks good. Still spooky that this snuck by.

But makes sense about the automerge thing actually.

A5rocks:
We could try out GitHub merge queue after the next release (don't want workflow changes before then, ofc) maybe

jakkdl:
> We could try out GitHub merge queue after the next release (don't want workflow changes before then, ofc) maybe

ooh, that looks fancy. This crazy pace of PRs per day can't be kept up that much longer though, and there's only so many checks in ruff that can be enabled.

jakkdl:
oh, I don't seem to be able to merge this due to failing checks. I think it might never have been blocking before?

A5rocks:
If this is due to a flakey test I think we should rerun it cause any fix to the flakey test (a new PR, presumably) will need this fix too!

CoolCat467 (approved):

PR #2829, opened 2023-10-23
nursery_cancel doesn't seem to create cyclic garbage on 3.12 anymore

### LINKED ISSUE #2646: Fix Python 3.12 errors

jakkdl:
At least if I'm reading pytest output correctly.

Fixes #2646 and presumably https://github.com/python/cpython/issues/100964 can be closed as well.

### DIFF
--- a/trio/_core/_tests/test_run.py
+++ b/trio/_core/_tests/test_run.py
@@ -2374,10 +2374,6 @@ async def test_cancel_scope_exit_doesnt_create_cyclic_garbage() -> None:
         gc.garbage.clear()
 
 
-@pytest.mark.xfail(
-    sys.version_info >= (3, 12),
-    reason="Waiting on https://github.com/python/cpython/issues/100964",
-)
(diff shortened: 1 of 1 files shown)


CoolCat467 (approved):

A5rocks:
FYI I did some bisecting and this turns out not to be due to an upstream change but instead due to https://github.com/python-trio/trio/commit/a3d9d2536a4fefd9bfce2b66fa6a557304c54c81 being merged in.

PR #2827, opened 2023-10-23
timeout coverage/pytest with sigint to help debug CI timeout issues

jakkdl:
didn't find any dead simple ways of doing this within pytest. I hope this works on other platforms

PR #2832, opened 2023-10-24
Fix thread reentry deadlocks

richardsheridan:
Should fix the timeouts @jakkdl worked on in #2827, which were introduced in #2392.  Some discussion from chat:

> @richardsheridan: the examples used seem to be extremely simple use cases where  we do `to_thread_run_sync(from_thread_run_X(afn))` back to back. if we can make a test that just loops for a few seconds doing that maybe it could reliably trigger the race for debugging? I won't be able to get to it for a while but just thinking out loud in case someone else wants to try first.

> @oremanj: richardsheridan: I just took another look at the threading code changes and I think there is a race here that I missed. When a host-task `run` completes, the host task immediately writes the result to the queue, which unblocks the thread and makes it possible for it to terminate. The thread's termination will reschedule the host task. But there's one checkpoint that the host task executes in between putting the result on the queue and going back to sleep:
> 
> ```
>     async def run(self) -> None:
>         # we use extra checkpoints to pick up and reset any context changes
>         task = trio.lowlevel.current_task()
>         old_context = task.context
>         task.context = self.context.copy()
>         try:
>             await trio.lowlevel.cancel_shielded_checkpoint()
>             result = await outcome.acapture(self.unprotected_afn)
>             self.queue.put_nowait(result)
>         finally:
>             task.context = old_context
>             await trio.lowlevel.cancel_shielded_checkpoint() # <-- this one
> ```
> 
> If the `report_back_in_trio_thread` callback is run during that checkpoint, then we will try to reschedule an already-scheduled task, which violates the task invariants and is probably what's causing this hang.

I'm going to first see how the test suite goes on my new regression test, then push the fix.

### DIFF
--- a/trio/_tests/test_threads.py
+++ b/trio/_tests/test_threads.py
@@ -13,9 +13,17 @@ from typing import Callable, Optional
 import pytest
 import sniffio
 
-from .. import CapacityLimiter, Event, _core, fail_after, sleep, sleep_forever
+from .. import (
+    CapacityLimiter,
+    Event,
+    _core,
+    fail_after,
+    move_on_after,
+    sleep,
+    sleep_forever,
+)
 from .._core._tests.test_ki import ki_self
-from .._core._tests.tutil import buggy_pypy_asyncgens
+from .._core._tests.tutil import buggy_pypy_asyncgens, slow
 from .._threads import (
     current_default_thread_limiter,
     from_thread_check_cancelled,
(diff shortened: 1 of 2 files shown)


richardsheridan:
The test wasn't quite 100% reliably triggered in CI but much more frequently than the intermittent failures we had been seeing.

oremanj:
Thanks for fixing this!

jakkdl:
Awesome, thanks for the quick fix! :rocket:

PR #2833, opened 2023-10-25
Stop running `pypy-3.9-nightly` for windows CI

CoolCat467:
This PR aims to fix the 3.9 nightly test failure.
For future reference, this failure has been happening in a lot of the CI runs lately because of something with the internal version of cffi that's pre-installed with pypy.

Current idea on how to fix is to do the suggestion pip has and force-reinstall cffi.


A5rocks:
pffft and they told me Python was a memory safe language :sob:

A5rocks:
For what it's worth, I've been intending to drop old nightlies from CI matrix (so that means add 3.11 nightly, remove 3.10/3.9 nightly) for pypy but if that doesn't sound good to you feel free to figure out the root cause ^^

ci.sh
jakkdl, line 1: you seem to be running some autofixer here? The changes are minor enough it's probably fine to just include them, but it would be cleaner to run it on all files in a separate PR + possibly add it to pre-commit.

ci.sh
```diff
@@ -51,7 +51,7 @@ else
     # Actual tests
     # expands to 0 != 1 if NO_TEST_REQUIREMENTS is not set, if set the `-0` has no effect
     # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02
-    if [ ${NO_TEST_REQUIREMENTS-0} == 1 ]; then
```
jakkdl, line 54: IIRC this one in particular doesn't *need* quotes for weird curly braces reasons, but I don't think they can hurt and `no-test-requirements` looks to run correctly, so let's just keep them.

jakkdl (approved):

jakkdl:
> For what it's worth, I've been intending to drop old nightlies from CI matrix (so that means add 3.11 nightly, remove 3.10/3.9 nightly) for pypy but if that doesn't sound good to you feel free to figure out the root cause ^^

separate PR for pypy-3.11 sounds good

ci.sh
CoolCat467: You are correct, I am using https://www.shellcheck.net/, which actually does have a pre-commit hook, and I was actually planning on making a separate PR for adding said hook, but then when I was testing locally it turns out it needs docker to run and I didn't like the security details of that I was seeing. I could indeed move these changes to a separate PR if you would like me to.

jakkdl: Ah, in that case it seems reasonable to just run it as a one-time thing on all the shell files in the repo in a dedicated PR.
The changes in ci.sh are so minor though I don't care much either way what you want to do with this PR.

jakkdl: Oh actually, #2837 is a pretty good case for leaving it for a dedicated PR.

jakkdl:
merging to get rid of the :x:s on runs

PR #2834, opened 2023-10-25
Sort pre-commit config file

CoolCat467:
This PR enables the `sort-simple-yaml` pre-commit hook, and has it sort the `.pre-commit-config.yaml` file.
The intention is to give it order if changes ever happen in the future, but in the end it's personal preference.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,10 @@
+ci:
+  autofix_commit_msg: "[pre-commit.ci] auto fixes from pre-commit.com hooks"
+  autofix_prs: false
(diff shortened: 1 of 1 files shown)


A5rocks:
TBH not sure how useful this is given it doesn't seem to sort lists. We won't be adding new keys that often, surely...

jakkdl (approved):
the more the merrier!

CoolCat467:
> TBH not sure how useful this is given it doesn't seem to sort lists. We won't be adding new keys that often, surely...

I think it intentionally doesn't sort lists because they might be being used as command arguments where order might be important.

PR #2838, opened 2023-10-26
extend apport excepthook detection to also detect partial_apport_excepthook

### LINKED ISSUE #2464: Apport ExeptHook detection stopped working.
The detection via the name `apport_excepthook` introduced in #1065,#1528 stopped working for me. 
Checking my system it seems apport has changed the name:
```
>>> import sys
>>> sys.excepthook.__name__
'partial_apport_excepthook'
```

Relevant change: https://github.com/canonical/apport/commit/ad6d801d264e9a7430d8d655fd04284a8871d68a#diff-262abd69bbc20e14d09d4b9041c97e2586e3f3f3db59acf4389b2485abe6dd7cR233

I suggest adding both names to the detection.

This started after updating from Ubuntu 22.04 to Ubuntu 22.10.

Thanks! 

Paging @pquentin, as you are the original Author of the workaround.

jakkdl:
fixes #2464 

A slighly more general fix would be `if "apport_excepthook" in getattr(sys.excepthook, "__name__", "")` in case they make another slight change to the name of the function.

The issue mentioned that this broke as they updated to Ubuntu 22.10, and all the ubuntu builds in our CI runs on 22.04, which I assume is why this wasn't picked up by the test suite.
It would be great if @Adrien-Luxey or somebody else with easy access to an ubuntu 22.10 machine could run the test suite on master and then this PR to see if it is fixed.

Checking https://github.com/actions/runner-images#available-images it looks like 22.04 is the only available image, so would require a bunch of extra work to get this tested in CI.

### DIFF
--- a/trio/_core/_multierror.py
+++ b/trio/_core/_multierror.py
@@ -474,9 +474,9 @@ def concat_tb(
 # hook.
 #
 # More details: https://github.com/python-trio/trio/issues/1065
-if (
-    sys.version_info < (3, 11)
-    and getattr(sys.excepthook, "__name__", None) == "apport_excepthook"
+if sys.version_info < (3, 11) and getattr(sys.excepthook, "__name__", None) in (
+    "apport_excepthook",
+    "partial_apport_excepthook",
 ):
     from types import ModuleType
 


A5rocks (approved):
This looks like it works. I'll see if I have an up-to-date ubuntu system anywhere...

A5rocks:
Looks like ubuntu 23.10 reverted this change at least

PR #2839, opened 2023-10-27
Convert markdown link in README.rst to reST

richardsheridan:


A5rocks (approved):
I think I suggested this change, sorry! Good catch!

PR #2841, opened 2023-10-28
In `to_thread_run_sync()`, add `abandon_on_cancel=` as an alias for the `cancellable=` flag

richardsheridan:
As suggested in https://github.com/python-trio/trio/pull/2392#discussion_r1354105279


newsfragments/2841.feature.rst
```diff
@@ -0,0 +1,3 @@
+To better reflect the underlying thread handling semantics,
+`trio.to_thread.run_sync` gained a new keyword argument ``abandon_on_cancel``
+that overrides the older (now deprecated) ``cancellable`` flag.
```
A5rocks, line 3: Hmm it's not exactly deprecated. I'm not sure how to best phase "replaced temporary kwarg name" or whatever though. I feel like it's important information that `cancellable` was never released.

richardsheridan: I'm happy to simply not say deprecated anywhere!

A5rocks: Yeah that probably works

A5rocks: Oh wait gahhh I shouldn't do this while sleepy, I only saw the changes that replaced the kwarg, didn't realize you kept it for compatibility 😭 

Sorry... if we are going to keep the kwarg, then deprecation is good. I thought the function that "cancelled" is on was added in this release meaning we could just replace the name without deprecation period.

richardsheridan: Haha no worries, but I won't be able to revert it today. Feel free to merge either version in the mean while.

TeamSpen210 (approved):
That's a much more understandable name. It's not necessary, but we could use `@overload` to have type checkers prevent you from passing both parameters:
```py
@overload
async def to_thread_run_sync(
    sync_fn: Callable[..., RetT],
    *args: object,
    thread_name: str | None = None,
    abandon_on_cancel: bool,
    limiter: CapacityLimiter | None = None,
) -> RetT: ...
@overload
async def to_thread_run_sync(
    sync_fn: Callable[..., RetT],
    *args: object,
    thread_name: str | None = None,
    cancellable: bool = False,
    limiter: CapacityLimiter | None = None,
) -> RetT: ...
```

A5rocks (approved):

jakkdl (changes requested):
If `cancellable` is actually deprecated, as it's phrased in the newsfragment, then it should raise a `DeprecationWarning` if `cancellable` is specified.
Alternatives would be to forever support both arguments (yuck), or to make it deprecated later for some reason.

richardsheridan:
I'd sooner abandon this pull request than start raising warnings everywhere. The extra maintenance burden is 3 easy lines of code on our end, but a warning will send our users scrambling to perform busywork. Someone else should add a warning in a future PR for a real reason, like `cancellable will be removed in 1.0`, at the time such a decision is made to balance user and maintainer burdens.

newsfragments/2841.feature.rst
jakkdl, line 3:
```suggestion
To better reflect the underlying thread handling semantics,
`trio.to_thread.run_sync` gained a new keyword argument ``abandon_on_cancel``
that can be used instead of the older ``cancellable`` flag (that might get fully deprecated in the future).
```

docs/source/reference-core.rst
```diff
@@ -1827,16 +1827,16 @@ to spawn a child thread, and then use a :ref:`memory channel
...
+   around to perform the work. If you pass ``abandon_on_cancel=True`` at the outset, or if
    you provide a :class:`~trio.lowlevel.TrioToken` when calling back in to Trio, your
    functions will be executed in a new system task. Therefore, the
    :func:`~trio.lowlevel.current_task`, :func:`current_effective_deadline`, or other
    task-tree specific values may differ depending on keyword argument values.
 
 You can also use :func:`trio.from_thread.check_cancelled` to check for cancellation from
 a thread that was spawned by :func:`trio.to_thread.run_sync`. If the call to
-:func:`~trio.to_thread.run_sync` was cancelled (even if ``cancellable=False``!), then
+:func:`~trio.to_thread.run_sync` was cancelled (even if ``abandon_on_cancel=False``!), then
```
oremanj, line 1839: I think we can remove this parenthetical, it no longer makes sense since the name isn't confusing anymore

newsfragments/2841.feature.rst
```diff
@@ -0,0 +1,3 @@
+To better reflect the underlying thread handling semantics,
+`trio.to_thread.run_sync` gained a new keyword argument ``abandon_on_cancel``
```
oremanj, line 2: This newsfragment should get a `.removal.` name (not `.feature.`) so it shows up in the "Deprecations and removals" section of the release notes.

I would write this as
```
To better reflect the underlying thread handling semantics,
the keyword argument for `trio.to_thread.run_sync` that was
previously called ``cancellable`` is now named ``abandon_on_cancel``.
It still does the same thing -- allow the thread to be abandoned
if the call to `trio.to_thread.run_sync` is cancelled -- but since we now
have other ways to propagate a cancellation without abandoning
the thread, "cancellable" has become somewhat of a misnomer.
The old ``cancellable`` name is now deprecated.
```

trio/_threads.py
```diff
@@ -197,8 +198,10 @@ async def to_thread_run_sync(  # type: ignore[misc]
       sync_fn: An arbitrary synchronous callable.
       *args: Positional arguments to pass to sync_fn. If you need keyword
           arguments, use :func:`functools.partial`.
-      cancellable (bool): Whether to allow cancellation of this operation. See
-          discussion below.
+      abandon_on_cancel (bool): Whether to abandon this thread upon
+          cancellation of this operation. See discussion below.
+      cancellable (bool): *Deprecated* synonym for ``abandon_on_cancel``.
```
oremanj, line 203: Please remove discussion of `cancellable` from the documentation (deprecated options don't get docs), and raise a deprecation warning if it is passed. You will need to default it to None in order to notice if someone passed False explicitly. You can make it a hard error to supply both `cancellable` and `abandon_on_cancel`, so you don't have to worry about which one overrides the other one.

richardsheridan:
Guess I'm outvoted here. Hope everything goes ok downstream!

richardsheridan:
> ```trio/_threads.py:295:9: B028 No explicit `stacklevel` keyword argument found```

I guess I don't know at what level of the stack to warn?

trio/_tests/test_threads.py
```diff
@@ -1068,9 +1071,25 @@ async def test_reentry_doesnt_deadlock() -> None:
...
             for _ in range(4):
                 nursery.start_soon(child)
+
+
+async def test_cancellable_and_abandon_raises() -> None:
+    with pytest.raises(ValueError):
+        await to_thread_run_sync(bool, cancellable=True, abandon_on_cancel=False)
+
+    with pytest.raises(ValueError):
+        await to_thread_run_sync(bool, cancellable=True, abandon_on_cancel=True)
```
A5rocks, line 1087: The two `to_thread_run_sync`s should have a `# type: ignore[call-overload]`

trio/_tests/test_threads.py
```diff
@@ -881,15 +884,15 @@ async def test_trio_token_weak_referenceable() -> None:
...
+async def test_unsafe_abandon_on_cancel_kwarg() -> None:
     # This is a stand in for a numpy ndarray or other objects
     # that (maybe surprisingly) lack a notion of truthiness
     class BadBool:
         def __bool__(self) -> bool:
             raise NotImplementedError
 
     with pytest.raises(NotImplementedError):
-        await to_thread_run_sync(int, cancellable=BadBool())  # type: ignore[arg-type]
+        await to_thread_run_sync(int, abandon_on_cancel=BadBool())  # type: ignore[arg-type]
```
A5rocks, line 895: Because `to_thread_run_sync` is an overloaded function, the error code associated with it changed from `arg-type` to `call-overload`.

This also applies to line 586.

A5rocks (commented):
Re: mypy errors

trio/_threads.py
```diff
@@ -171,12 +172,35 @@ def run_in_system_nursery(self, token: TrioToken) -> None:
...
+    *args: object,
+    thread_name: str | None = None,
+    cancellable: bool = False,
+    limiter: CapacityLimiter | None = None,
+) -> RetT:
+    ...
+
+
 @enable_ki_protection  # Decorator used on function with Coroutine[Any, Any, RetT]
 async def to_thread_run_sync(  # type: ignore[misc]
```
A5rocks, line 198: For the two above I think you'll need to transfer over this `# type: ignore` over.

oremanj:
> I'd sooner abandon this pull request than start raising warnings everywhere. The extra maintenance burden is 3 easy lines of code on our end, but a warning will send our users scrambling to perform busywork. Someone else should add a warning in a future PR for a real reason, like cancellable will be removed in 1.0, at the time such a decision is made to balance user and maintainer burdens.

Sorry, I reviewed without having seen this comment or I would've responded to it initially. I have empathy for this view, but it doesn't match how we've handled all other deprecations in the past, and I think the old name is actively harmful/confusing under the new semantics. (`cancellable=False` threads are IMO "more cancellable" than `cancellable=True` threads, because `cancellable=False` threads might actually terminate their operation if the `to_thread.run_sync` call is cancelled, while `cancellable=True` threads definitely won't.) I think most users of `to_thread.run_sync` probably don't specify `cancellable` at all, and for those that do, the warning is a nice prompt to reconsider whether they still need abandon-on-cancel now that actually-cancel-on-cancel is available.

trio/_threads.py
```diff
@@ -266,9 +287,20 @@ async def to_thread_run_sync(  # type: ignore[misc]
...
     await trio.lowlevel.checkpoint_if_cancelled()
-    if abandon_on_cancel is not None:
-        cancellable = abandon_on_cancel
-    abandon_on_cancel = bool(cancellable)  # raise early if cancellable.__bool__ raises
+    if cancellable is not None:
+        if abandon_on_cancel is not None:
+            raise ValueError(
+                "Cannot set `cancellable` and `abandon_on_cancel` simultaneously."
+            )
+        warnings.warn(
```
oremanj, line 295: Suggest you use `trio._deprecate.warn_deprecated()` which provides a standard message format and warning class (so users can suppress Trio deprecation warnings independently of others).

trio/_threads.py
```diff
@@ -292,11 +292,11 @@ async def to_thread_run_sync(  # type: ignore[misc]
             raise ValueError(
                 "Cannot set `cancellable` and `abandon_on_cancel` simultaneously."
             )
-        warnings.warn(
-            DeprecationWarning(
-                "`cancellable` keyword argument is deprecated, "
-                "use `abandon on cancel` instead."
-            )
+        warn_deprecated(
+            "`cancellable=` keyword argument",
```
oremanj, line 296:
```suggestion
            "The `cancellable=` keyword argument to `trio.to_thread.run_sync`",
```

trio/_threads.py
```diff
@@ -292,11 +292,11 @@ async def to_thread_run_sync(  # type: ignore[misc]
...
+            "0.23.0",
+            issue=2841,
+            instead="`abandon_on_cancel=` keyword argument",
```
oremanj, line 299:
```suggestion
            instead="`abandon_on_cancel=`",
```

oremanj: These make the message be:
```
The `cancellable=` keyword argument to `trio.to_thread.run_sync` is deprecated since Trio 0.23.0; use `abandon_on_cancel=` instead (<issue url>)
```
which I think reads nicely.

oremanj (approved):

A5rocks:
I'll merge this in soon-ish unless someone has other thoughts!

jakkdl (approved):
Sorry for being a pain in the butt, looks great!

richardsheridan:
sorry thought the merge was a merge to master  not from master!

jakkdl:
> sorry thought the merge was a merge to master not from master!

No worries! With the ongoing updates to linters etc I wanted to double check that this passes before merging even if GitHub deemed it a safe merge :)

PR #2842, opened 2023-10-28
Remove infinity support for tcp listener backlogs

CoolCat467:
This PR makes due on my comment in 6f9ab95349044dbb9d63408d2245dc95ad167f7e, removing support for `math.inf` as an argument for the tcp listener backlog, leaving only `int | None` as valid options.

### DIFF
--- a/newsfragments/2842.removal.rst
+++ b/newsfragments/2842.removal.rst
@@ -1,0 +1,1 @@
+Removed support for ``math.inf`` for the ``backlog`` argument in ``open_tcp_listeners``, making it's docstring correct in the fact that only ``TypeError``s are raised if invalid arguments are passed.
--- a/newsfragments/2842.removal.rst
+++ b/newsfragments/2842.removal.rst
@@ -1,1 +1,1 @@
-Removed support for ``math.inf`` for the ``backlog`` argument in ``open_tcp_listeners``, making it's docstring correct in the fact that only ``TypeError``s are raised if invalid arguments are passed.
+Removed support for ``math.inf`` for the ``backlog`` argument in ``open_tcp_listeners``, making its docstring correct in the fact that only ``TypeError``s are raised if invalid arguments are passed.
--- a/trio/_highlevel_open_tcp_listeners.py
+++ b/trio/_highlevel_open_tcp_listeners.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import errno
 import sys
 from collections.abc import Awaitable, Callable
-from math import inf
 
 import trio
 from trio import TaskStatus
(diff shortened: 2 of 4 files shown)


A5rocks:
While I get using `int | None` to work around flaws in the type system, I think there should at least be a deprecation period.

trio/_tests/test_highlevel_open_tcp_listeners.py
```diff
@@ -385,6 +383,6 @@ async def test_open_tcp_listeners_backlog_float_error() -> None:
     tsocket.set_custom_socket_factory(fsf)
     for should_fail in (0.0, 2.18, 3.14, 9.75):
         with pytest.raises(
-            ValueError, match=f"Only accepts infinity, not {should_fail!r}"
+            TypeError, match=f"backlog must be an int or None not {should_fail!r}"
```
A5rocks, line 386:
```suggestion
            TypeError, match=f"backlog must be an int or None, not {should_fail!r}"
```

njsmith:
Let's just keep accepting `inf` even if it's not documented or listed as a valid type, as a little backcompat thing? no reason to break existing code and it's not really any extra maintenance burden.

CoolCat467:
> Let's just keep accepting `inf` even if it's not documented or listed as a valid type, as a little backcompat thing? no reason to break existing code and it's not really any extra maintenance burden.

Added warning for `inf` but still accepts it as of [daada74](https://github.com/python-trio/trio/pull/2842/commits/daada74ec7e50f91069b764ba7e3352a2c6642ef)

A5rocks (approved):

jakkdl:
I'd kinda like making this consistent across the whole codebase with corresponding `DeprecationWarning`s*, but this seems like a good thing on it's own.

\* probably making it a very long deprecation period, since this mostly just is about the typing limitation (though I also do think it's much easier to specify `None` than having to import `math.inf`). Or not even raising a warning and just continue to accept `math.inf` without issue indefinitely, but add support for `None` (where that's not already the case) and specify the signature as only `int|None`.

trio/_highlevel_open_tcp_listeners.py
```diff
@@ -41,16 +42,18 @@
...
     # missing overflow protection, so we apply our own overflow protection.
     # https://github.com/golang/go/issues/5030
-    if isinstance(backlog, float):
-        # TODO: Remove when removing infinity support
-        # https://github.com/python-trio/trio/pull/2724#discussion_r1278541729
-        if backlog != inf:
-            raise ValueError(f"Only accepts infinity, not {backlog!r}")
+    if backlog == math.inf:
         backlog = None
+        warnings.warn(
```
jakkdl, line 51: Add test or `# pragma: no cover`

jakkdl (approved):

newsfragments/2842.removal.rst
```diff
@@ -0,0 +1 @@
+Removed support for ``math.inf`` for the ``backlog`` argument in ``open_tcp_listeners``, making it's docstring correct in the fact that only ``TypeError``s are raised if invalid arguments are passed.
```
A5rocks, line 1:
```suggestion
Removed support for ``math.inf`` for the ``backlog`` argument in ``open_tcp_listeners``, making its docstring correct in the fact that only ``TypeError``s are raised if invalid arguments are passed.
```

A5rocks (approved):
nit: feel free to merge regardless.

trio/_highlevel_open_tcp_listeners.py
CoolCat467: Added test in 60f6956

PR #2845, opened 2023-10-31
Fix regressions introduced when hiding `Nursery.start` implementation-detail nursery

### LINKED ISSUE #2844: (Regression on master) `strict_exception_groups` is not respected
`strict_exception_groups` is documented as making the exception structure that comes out of a nursery aexit not depend on whether `== 1` or `> 1` tasks raised, by having a strict nursery always wrap all exception(s) in a new group:

> If [`strict_exception_groups` is] true, nurseries will always wrap even a single raised exception in an exception group.

but on master this isn't happening:

## minimized reproducer 1

```python
import trio


async def eg() -> None:
    async def raiser() -> None:
        raise RuntimeError("boom")

    async def f() -> None:
        async with trio.open_nursery(strict_exception_groups=True) as inner:
            inner.start_soon(raiser)

    async with (
        # A similar bug also occurs if there are more strict nurseries open here:
        # trio.open_nursery(strict_exception_groups=True),
        trio.open_nursery(strict_exception_groups=True) as nursery,
    ):
        n = 2
        # try uncommenting me! it completely changes the exception structure.
        # n = 1
        for _ in range(n):
            nursery.start_soon(f)


trio.run(eg)
```

**expected**: the same exception structure with `n = 1` and `n > 1`, i.e. two levels of exception groups (or three if the third nursery is uncommented).

e.g. with `n = 1`:

```
  |   ...
  | trio.NonBaseMultiError: <MultiError: RuntimeError('boom')>
    | Exception Group Traceback (most recent call last):
    |   ...
    | trio.NonBaseMultiError: RuntimeError('boom')
      |   ...
      | RuntimeError: boom
```

this is the behavior on v0.22.2.

**observed**: (w/ trio @ 6c50dc2, CPython 3.11.6.) different exception structure depending on `n = 1` vs. `n > 1`.

e.g. with `n = 1`:

```
  |   ...
  | trio.NonBaseMultiError: RuntimeError('boom')
    |   ...
    | RuntimeError: boom
```

the task's crash is in only one one exception group, even though it occurred two strict nurseries down the stack. (similarly, if you uncomment the extra nursery, it's *still* only one when it should be three.)

while this new behavior will not break `except*` usage (because `except*` recurses), it is still a problem. for example:

## minimized reproducer 2

```python
import trio


async def eg() -> None:
    async def raiser() -> None:
        exc_group_from_some_task = ExceptionGroup(
            "some error message", (RuntimeError("boom"),)
        )
        exc_group_from_some_task.add_note("some note")
        raise exc_group_from_some_task

    try:
        async with trio.open_nursery(strict_exception_groups=True) as nursery:
            nursery.start_soon(raiser)
    except BaseExceptionGroup as nursery_exc_group:
        raise BaseExceptionGroup(
            "a useful error message instead of just the (necessarily) vague MultiError repr",
            nursery_exc_group.exceptions,
        ) from None
        # Note: `strict_exception_groups=True` guarantees that the `nursery_exc_group` we caught
        # here is a brand new one that the nursery created, and NOT an exception group
        # that came out of one of `nursery`'s tasks.


trio.run(eg)
```

**expected**: two exception groups, and trio doesn't cause your exception group messages/notes to be hidden:

```
  |   ...
  | ExceptionGroup: a useful error message instead of just the (necessarily) vague MultiError repr (1 sub-exception)
    | ...
    | ExceptionGroup: some error message (1 sub-exception)
    | some note
      | RuntimeError: boom
```

**observed**: one exception group, and trio causes some exception group messages/notes to be hidden:

```
  |   ...
  | ExceptionGroup: a useful error message instead of just the (necessarily) vague MultiError repr (1 sub-exception)
    | RuntimeError: boom
```

## details

i encountered this regression while testing a downstream work against trio master in preparation for the upcoming v0.23 release.

this bisects to #2826 (CC @jakkdl). from its newsfragment, tests, and surrounding discussions, my understanding is that #2826's primary intent was to hide the implementation-detail nursery of `start` (i.e. to resolve #2611)—but that it did not intend to make trio violate its `strict_exception_groups` docs & also hide third-party exception messages from tracebacks.

gschaffner:
proposed alternative fix and tests for #2611. fixes #2844.

note that these tests are parametrized:

> For the tests, there are currently enough separate cases that I can imagine missing that something is backwards. Could we write something parameterized, such that it obviously does each possible set of operations, and then the asserts only depend on whether we're in strict mode? (I expect this to be harder to read individually, but the confidence in consistency is what I'm after) --- https://github.com/python-trio/trio/pull/2826#issuecomment-1776252294

### DIFF
--- a/newsfragments/2611.bugfix.rst
+++ b/newsfragments/2611.bugfix.rst
@@ -1,4 +1,5 @@
 When a starting function raises before calling :func:`trio.TaskStatus.started`,
 :func:`trio.Nursery.start` will no longer wrap the exception in an undocumented
-:exc:`ExceptionGroup`. Previously, an :exc:`ExceptionGroup` was added when using
-``trio.run(..., strict_exception_groups=True)``.
+:exc:`ExceptionGroup`. Previously, :func:`trio.Nursery.start` would incorrectly
+raise an :exc:`ExceptionGroup` containing it when using ``trio.run(...,
+strict_exception_groups=True)``.
--- a/newsfragments/2611.bugfix.rst
+++ b/newsfragments/2611.bugfix.rst
@@ -1,1 +1,4 @@
-With ``strict_exception_groups=True``, when you ran a function in a nursery which raised an exception before calling ``task_status.started()``, it previously got wrapped twice over in ``ExceptionGroup`` in some cases. It no longer does that, and also won't wrap any ``ExceptionGroup`` raised by the function itself.
+When a starting function raises before calling :func:`trio.TaskStatus.started`,
+:func:`trio.Nursery.start` will no longer wrap the exception in an undocumented
+:exc:`ExceptionGroup`. Previously, an :exc:`ExceptionGroup` was added when using
+``trio.run(..., strict_exception_groups=True)``.
--- a/trio/_core/_run.py
+++ b/trio/_core/_run.py
@@ -1103,13 +1103,6 @@ class Nursery(metaclass=NoPublicConstructor):
 
         popped = self._parent_task._child_nurseries.pop()
         assert popped is self
-
-        # don't unnecessarily wrap an exceptiongroup in another exceptiongroup
-        # see https://github.com/python-trio/trio/issues/2611
-        if len(self._pending_excs) == 1 and isinstance(
-            self._pending_excs[0], BaseExceptionGroup
-        ):
-            return self._pending_excs[0]
         if self._pending_excs:
             try:
                 return MultiError(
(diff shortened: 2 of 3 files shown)


newsfragments/2611.bugfix.rst
```diff
@@ -1 +1,4 @@
-With ``strict_exception_groups=True``, when you ran a function in a nursery which raised an exception before calling ``task_status.started()``, it previously got wrapped twice over in ``ExceptionGroup`` in some cases. It no longer does that, and also won't wrap any ``ExceptionGroup`` raised by the function itself.
+When a starting function raises before calling :func:`trio.TaskStatus.started`,
+:func:`trio.Nursery.start` will no longer wrap the exception in an undocumented
+:exc:`ExceptionGroup`. Previously, an :exc:`ExceptionGroup` was added when using
+``trio.run(..., strict_exception_groups=True)``.
```
jakkdl, line 4: I think it's also good to mention in particular that this led to double-wrapped exceptions if used together with `trio.Nursery.start` with `strict_exception_groups` set to `True` or `None`.

Zac-HD (approved):
This looks great to me - thanks so much @gschaffner!

Issue #2846, opened 2023-10-31
`removal` newsfragments do not show in history


richardsheridan:
We have several removals pending and I was directed to make one in #2841, but `removal` is [not a valid towncrier type right now](https://github.com/python-trio/trio/blob/6c50dc2b747802fc3ad98a55071dad00f9fbaec9/pyproject.toml#L121-L154). Do we want to rename newsfragments, or add `removal` and remove `deprecated`, or maybe there is a way to alias them?

A5rocks:
Yeah I noticed this and my conclusion is we should add a new removal type that mirrors deprecated. If you check the readme in the newsfragments folder it mentions both are for different things after all.

oremanj:
My suggestion of `removal` in the `cancellable` -> `abandon_on_cancel` diff was incorrect; that should have been `deprecated`. I got confused by the other `removal` newsfragments, which I agree that we should support.

PR #2850, opened 2023-11-02
Add removals as a newsfragment type

A5rocks:
Fixes https://github.com/python-trio/trio/issues/2846

cc @CoolCat467 about changing the category for your newsfragment.


CoolCat467 (approved):

CoolCat467:
I know it says there should be two reviews, but I think this should be simple enough and doesn't even change code, so I am going to merge this.

PR #2852, opened 2023-11-02
add tests and coverage ignores for testing/_fake_net to reach full coverage

jakkdl:
When writing #2820 I noticed the bad shape of `_fake_net`. While it's only for internal use, and unclear if it's ever going to be added to the public API, I figured it'd be worth cleaning it up a bit and fix coverage so as not to drag everything else down.


CoolCat467 (approved):
This all makes sense, I approve.

TeamSpen210 (approved):

PR #2856, opened 2023-11-03
Remove incorrect assertion for Anaconda interpreters

### LINKED ISSUE #2855: trio 0.23.0 - AssertionError in Conda
### To reproduce

```bash
docker run -it continuumio/miniconda3 /bin/bash

pip install trio

python -c "import trio"
```

**Error**

> Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/site-packages/trio/_subprocess.py", line 50, in <module>
    from os import pidfd_open
ImportError: cannot import name 'pidfd_open' from 'os' (/opt/conda/lib/python3.11/os.py)

> During handling of the above exception, another exception occurred:

> Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/conda/lib/python3.11/site-packages/trio/__init__.py", line 23, in <module>
    from . import abc, from_thread, lowlevel, socket, to_thread
  File "/opt/conda/lib/python3.11/site-packages/trio/lowlevel.py", line 48, in <module>
    from ._subprocess import open_process as open_process
  File "/opt/conda/lib/python3.11/site-packages/trio/_subprocess.py", line 54, in <module>
    assert sys.version_info < (3, 9) or sys.implementation.name != "cpython"
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AssertionError

### Why

It seems related to https://github.com/python-trio/trio/blob/eef24cb9787aa8f4f1707b4889ccd2a71419e3ce/trio/_subprocess.py#L49-L54

- the `os` package in Conda deviates from std python and does not have `pidfd_open` --> `ImportError` (see a [similar issue](https://github.com/conda/conda/issues/10085#issuecomment-661297467)).
- `sys.platform == "linux"` is `True`
- `sys.version_info < (3, 9) or sys.implementation.name != "cpython"` is `False or False` --> `AssertionError`

A5rocks:
cc @jakkdl (didn't realize Anaconda interpreters were weird!)

We should add Anaconda to the CI matrix, though. And put out a patch fix soon after this!

Fixes #2855


jakkdl (approved):
Hm, in hindsight I probably shouldn't have bothered with that assertion once testing the PR was done. There's no real gain from knowing it's needed in more weird cases, unless pypy started supporting it after we'd dropped `3.8`.

Maaaybe it can be removed whenever support for that older linux kernel is dropped, and so the assertion check could be extended to check `platform.system` / `platform.release` /`platform.version`. But we should probably just give up on asserting, and maybe reconsider removing it only when we cannot find any way to run it in CI.

PR #2860, opened 2023-11-04
switch to PEP 621

graingert:
This is a simpler version of https://github.com/python-trio/trio/pull/2449/files that keeps setuptools and continues parsing the version out of `trio/_version.py` rather than adding `setuptools-scm`


ci.sh
```diff
@@ -37,11 +37,11 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
...
 
 echo "::group::Install dependencies"
-python -m pip install -U pip setuptools wheel
+python -m pip install -U pip build
 python -m pip --version
 
-python setup.py sdist --formats=zip
-python -m pip install dist/*.zip
+python -m build --sdist
+python -m pip install dist/*.tar.gz
```
graingert, line 44: I'd be tempted to do this:

```suggestion
python -m build
python -m pip install dist/*.whl
```

pyproject.toml
```diff
@@ -1,3 +1,76 @@
...
+[tool.setuptools]
+# This means, just install *everything* you see under trio/, even if it
+# doesn't look like a source file, so long as it appears in MANIFEST.in:
+include-package-data = true
+
+[tool.setuptools.packages]
+find = {namespaces = false}
+
+[tool.setuptools.dynamic]
+version = {attr = "trio._version.__version__"}
```
graingert, line 72: this is the magic to avoid setuptools-scm

ci.sh
TeamSpen210: Could even do `pip install .`, though it might be better to keep it separate to get better logs.

graingert: `pip install .` does not detect problems with your MANIFEST.in

pyproject.toml
```diff
@@ -1,3 +1,76 @@
...
+authors = [{name = "Nathaniel J. Smith", email = "njs@pobox.com"}]
+license = {text = "MIT OR Apache-2.0"}
+keywords = [
+    "async",
+    "io",
+    "networking",
+    "trio",
+]
+classifiers = [
+    "Development Status :: 3 - Alpha",
```
agronholm, line 17: Is this still correct?

pyproject.toml
```diff
@@ -1,3 +1,76 @@
+[build-system]
+requires = ["setuptools>=61.2"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "trio"
+description = "A friendly Python library for async concurrency and I/O"
+authors = [{name = "Nathaniel J. Smith", email = "njs@pobox.com"}]
```
agronholm, line 8: Maybe `maintainers` should be filled in too?

pyproject.toml
```diff
@@ -1,3 +1,76 @@
+[build-system]
+requires = ["setuptools>=61.2"]
```
agronholm, line 2: I recommend setuptools >= 64 since that release added editable installs for PEP 621 based projects.

pyproject.toml
graingert: definitely worth fixing the README.rst and this at once in a separate PR. I want to keep the metadata identical in the PR

pyproject.toml
graingert:
```suggestion
requires = ["setuptools>=64"]
```

pyproject.toml
```diff
@@ -1,3 +1,76 @@
...
+    "attrs >= 20.1.0",
+    "sortedcontainers",
+    "idna",
+    "outcome",
+    "sniffio >= 1.3.0",
+    # cffi 1.12 adds from_buffer(require_writable=True) and ffi.release()
+    # cffi 1.14 fixes memory leak inside ffi.getwinerror()
+    # cffi is required on Windows, except on PyPy where it is built-in
+    "cffi>=1.14; os_name == 'nt' and implementation_name != 'pypy'",
+    "exceptiongroup >= 1.0.0rc9; python_version < '3.11'",
```
agronholm, line 50: Let's not have pre-release versions in specifiers unless absolutely necessary.
```suggestion
    "exceptiongroup; python_version < '3.11'",
```

pyproject.toml
graingert: that's probably quite a long list, and I want to keep the metadata identical in this PR

pyproject.toml
graingert: I'll do this in a followup PR

.coveragerc
```diff
@@ -2,7 +2,6 @@
 branch=True
 source=trio
 omit=
-  setup.py
```
agronholm, line 5: Do we want to keep a separate .coveragerc instead of putting the config to `pyproject.toml`?

graingert: moving all config files into pyproject.toml proved controversial in gitter - I'd like to move that config in a followup PR

pyproject.toml
graingert: where's the issue to switch from alpha 3 to final? and be more confident in introducing trio in the docs?

ci.sh
agronholm: What problems does `python -m build` detect that `pip install .` does not?

agronholm: Hm, I think this relates to building the sdist which IIRC pip install . does not do, yes?

graingert: Building an sdist can omit files in the source tree missing from MANIFEST.in `pip install .` doesn't omit those files

agronholm: I was thoroughly confused by this response. Are you saying that `pip install .` skips parts of the build process somehow, or not? As you said "pip install . _doesn't_ omit those files".

graingert: Yeah pip install . skips the sdist phase which is where MANIFEST.in is used

agronholm: I don't think this argument holds water since `python -m build` happily builds the sdist even when files included in `MANIFEST.in` are missing.

graingert: Right but then those files will be missing from the install

agronholm: They'll be missing (without warnings) no matter how you do the build.

graingert: If you do `pip install .` they won't be missing

agronholm: How can nonexistent files not be missing from the install?

graingert: the point is that `pip install .` skips the sdist stage and we want to test the sdist stage

agronholm: Ok, so: `pip install .` does not go through the sdist packaging, so `python -m build` and then installing the wheel would seem like the way to go.

graingert: Probably we want a comment on this bit

pyproject.toml
webknjaz: Could write something like "project collaborators" as a single entry. But you're right. It's out of the scope.

agronholm (approved):

newsfragments/2860.misc.rst
```diff
@@ -0,0 +1 @@
+Moved the metadata into ``PEP 621``-compliant ``pyproject.toml``.
```
webknjaz, line 1: Nit: could use Sphinx-specific roles here.
```suggestion
Moved the metadata into :pep:`PEP 621`-compliant :file:`pyproject.toml`.
```

pyproject.toml
```diff
@@ -1,3 +1,76 @@
+[build-system]
+requires = ["setuptools>=64"]
```
webknjaz, line 2: (Personal preference, non-blocking) 
```suggestion
requires = ["setuptools >= 64"]
```

pyproject.toml
```diff
@@ -1,3 +1,76 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Topic :: System :: Networking",
+    "Typing :: Typed",
+]
+requires-python = ">=3.8"
+dependencies = [
+    # attrs 19.2.0 adds `eq` option to decorators
```
webknjaz, line 39: Do we really need to keep this comment about some older version boundary?

pyproject.toml
```diff
@@ -1,3 +1,76 @@
...
+    # attrs 20.1.0 adds @frozen
+    "attrs >= 20.1.0",
+    "sortedcontainers",
+    "idna",
+    "outcome",
+    "sniffio >= 1.3.0",
+    # cffi 1.12 adds from_buffer(require_writable=True) and ffi.release()
```
webknjaz, line 46: Same: the older version comment is probably unnecessary.

pyproject.toml
webknjaz: Technically, this shouldn't matter, since library deps semantically mean compatibility, not demand of pre-releases... From that PoV, if a necessary feature that is used appeared in an rc, then it's more accurate to use rc.

pyproject.toml
webknjaz: Yea. Though, setuptools-scm provide many more things, than just generating a version.. So it's not exactly a replacement, just sourcing version from importables.

trio/_version.py
```diff
@@ -1,3 +1,3 @@
-# This file is imported from __init__.py and exec'd from setup.py
+# This file is imported from __init__.py and ast parsed by setuptools
```
webknjaz, line 1: Typo?
```suggestion
# This file is imported from __init__.py and parsed by setuptools
```

graingert: Not a typo I meant ast parsed

pyproject.toml
graingert: Comments were in the original setup.py so I maintained them in the pyproject.toml

pyproject.toml
graingert: Comments were in the original setup.py so I maintained them in the pyproject.toml

pyproject.toml
agronholm: It will matter if I ever make another pre-release of `exceptiongroup`, as using a pre-release in a version specifier essentially gives pip permission to install the latest matching pre-release.

trio/_version.py
graingert: I wanted to say something instead of eval/exec describing what setuptools does here

newsfragments/2860.misc.rst
```diff
@@ -0,0 +1 @@
+Moved the metadata into :pep:`PEP 621`-compliant :file:`pyproject.toml`.
```
webknjaz, line 1: Commenting from phone, I didn't fully clean up the previous suggested change. This should be better:
```suggestion
Moved the metadata into :pep:`621`-compliant :file:`pyproject.toml`.
```

webknjaz (approved):

newsfragments/2860.misc.rst
agronholm: You should remove the word `PEP ` from between the backticks (not sure if it work with that inside).

A5rocks:
Sorry, didn't see this. Looks good!

.coveragerc
graingert: @agronholm https://github.com/python-trio/trio/pull/2867

pyproject.toml
graingert: @agronholm @webknjaz https://github.com/python-trio/trio/issues/2868

PR #2866, opened 2023-11-05
use the regular readme for the long description

graingert:


A5rocks:
Personally I like having a customized version of README for pypi even if it has caused issues due be being out of sync in the past.

(people seeing trio in pypi have different goals than people seeing trio on github. but idk if we adequately address that as is...)

Fuyukai:
Genuinely, do people ever actually read the pypi readmes? I nearly always just click through to the repository instead (usually because the readme is outdated and wrong).

A5rocks:
I'm not sure actually -- for the longest time the pypi readme wasn't rendered as rst and I don't remember anyone talking about it (I fixed it cause it annoyed me)

Maybe I'm just forgetting things though

pyproject.toml
```diff
@@ -52,7 +52,7 @@ dependencies = [
 dynamic = ["version"]
 
 [project.readme]
-file = "LONG_DESCRIPTION.rst"
+file = "README.rst"
 content-type = "text/x-rst"
```
A5rocks, line 56: Do we really need to specify `content-type`?

A5rocks (approved):
OK I've come around on this, I agree with @Fuyukai that nobody really looks at these and that any upkeep burden is too much.

pyproject.toml
CoolCat467: I think it's best practice, but I am not certain.

CoolCat467 (approved):
I think this makes sense

PR #2867, opened 2023-11-05
move coveragrc to pyproject.toml

graingert:


TeamSpen210 (approved):

newsfragments/2867.misc.rst
```diff
@@ -0,0 +1 @@
+move .coveragerc into pyproject.toml
```
A5rocks, line 1: Noticed this with another newsfragment you wrote, but I believe our convention here is to capitalize the first letter (?)

Shouldn't block merge though, if you want you can fix this but there's other ones to fix too (in another PR).

graingert:
```suggestion
Move .coveragerc into pyproject.toml
```

PR #2871, opened 2023-11-05
Minor refactors in _threads.py

richardsheridan:
Since I've been staring at the threading code for a long time now, I've found a couple of changes that are probably good for maintainability and good style, but were not a priority for release. I recommend looking at this PR commit by commit rather than as a whole; I'll try to make sure CI gives a green check for each discrete refactor.


richardsheridan:
Mypy boffins: the diff of 5422f61a doesn't touch the `abandon_on_cancel` variable but somehow mypy can't tell anymore that the variable cannot be None past [line 302](https://github.com/python-trio/trio/blob/5422f61aae54fd179c82da0dcb34babf7c5bc95d/trio/_threads.py#L302).

A5rocks:
> Mypy boffins: the diff of [5422f61](https://github.com/python-trio/trio/commit/5422f61aae54fd179c82da0dcb34babf7c5bc95d) doesn't touch the `abandon_on_cancel` variable but somehow mypy can't tell anymore that the variable cannot be None past [line 302](https://github.com/python-trio/trio/blob/5422f61aae54fd179c82da0dcb34babf7c5bc95d/trio/_threads.py#L302). Is this a mypy bug? I'll just ignore it for now.

mypy doesn't let you redefine the variable type. however, you can stick a `assert abandon_on_cancel is not None` because mypy narrowing will work.

richardsheridan:
I'd rather ignore than assert because 

 - if a `None` sneaks through, it likely has the correct semantics, so an assertion would be unnecessary brittleness
 - we can save a bytecode on every function call for something we can determine statically

Happy to hear an argument otherwise though!

A5rocks:
I don't see why mypy started warning you just now, so that's probably a sign of a mypy bug.

As for adding an `assert`, people can always use `-O` and get rid of them (... assuming trio doesn't rely on `assert`s having side effects, which it really shouldn't) so I regard point 2 as irrelevant and point 1 is impossible AFAIK. But idk if it's worth addressing.

richardsheridan:
> mypy doesn't let you redefine the variable type.

Actually this was the right answer all along, I just needed a new variable to carry along the narrowed type!

CoolCat467 (approved):
You are quite right, looking at as one big change is confusing. After going through each commit, this looks great!

CoolCat467 (approved):

src/trio/_threads.py
```diff
@@ -369,29 +361,30 @@ def worker_fn() -> RetT:
             del PARENT_TASK_DATA.task_register
 
     context = contextvars.copy_context()
-    # Partial confuses type checkers, coerce to a callable.
-    contextvars_aware_worker_fn: Callable[[], RetT] = functools.partial(context.run, worker_fn)  # type: ignore[assignment]
+    # Trio doesn't use current_async_library_cvar, but if someone
+    # else set it, it would now shine through since
+    # sniffio.thread_local isn't set in the new thread. Make sure
+    # the new thread sees that it's not running in async context.
+    context.run(current_async_library_cvar.set, None)
```
A5rocks, line 368: I thought we didn't use this anymore. As of https://github.com/python-trio/trio/pull/2700

A5rocks (changes requested):
One small thing I noticed, looks good other than this.

src/trio/_threads.py
A5rocks: Ohh sorry I should have read the comment better. Ack.

A5rocks (approved):
Sorry about the confusion, looks good.

CoolCat467:
With a second approval, I am going to merge this.

Issue #2873, opened 2023-11-08
`pyright` does not understand `trio.open_memory_channel[int](0)`


Zac-HD:
```python
# pyright: strict
import trio

send_chan, recv_chan = trio.open_memory_channel[int](1e999)
```

`mypy --strict t.py` succeeds, but `pyright t.py` fails with:

```
/Users/zac/t.py
  /Users/zac/t.py:4:49 - error: Expected no type arguments for class "open_memory_channel" (reportGeneralTypeIssues)
  /Users/zac/t.py:4:1 - error: Type of "send_chan" is partially unknown
    Type of "send_chan" is "MemorySendChannel[Unknown]" (reportUnknownVariableType)
  /Users/zac/t.py:4:12 - error: Type of "recv_chan" is partially unknown
    Type of "recv_chan" is "MemoryReceiveChannel[Unknown]" (reportUnknownVariableType)
3 errors, 0 warnings, 0 informations 
```

This is on Trio 0.23.1; if I install `trio-typing` I don't have any issues - but I'd prefer to drop the extra dependency.

See also: https://github.com/python-trio/trio/issues/2734

Zac-HD:
I also saw a bizarre `error: Argument missing for parameter "cancelled_caught" (reportGeneralTypeIssues)` for `trio.CancelScope()`, implying a bug in handling of `attrs` classes since `cancelled_caught` is an `init=False` attribute... but couldn't get an isolated repro.  I'm just going to ignore that for now, but noting it here in case someone else runs into it.

A5rocks:
For the attrs thing that's very possible because `pyright` made a PEP based on dataclasses behavior and then applies it to `attrs`. `mypy` just has an `attrs` plugin builtin that does essentially the same transformations.

As for `open_memory_channel` it looks like we might be forgetting a `Generic` there? I find it surprising `mypy` doesn't warn; does `mypy` let us imply a generic class?

A5rocks:
OK something's fishy here:
```py
# pyright: strict
from trio._channel import MemorySendChannel, MemoryReceiveChannel, open_memory_channel
import typing

T = typing.TypeVar("T")

class omc(typing.Tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]):
    def __new__(  # type: ignore[misc]  # "must return a subtype"
        cls, max_buffer_size: int | float
    ) -> tuple[MemorySendChannel[T], MemoryReceiveChannel[T]]:
        ...

[...]

PR #2874, opened 2023-11-08
Make pyright recognize `open_memory_channel` as generic

A5rocks:
Fixes https://github.com/python-trio/trio/issues/2873

Minimal changes required. It's also possible to reorganize code but meh.

Not sure if there's a way to scan `trio` for other occurances of this... other than running `pyright` over it and fixing everything which I absolutely do not want to do. 2 type checkers is 1 too many!


A5rocks:
pre-commit.ci autofix

Zac-HD (approved):
I've confirmed locally that this fixes the issue!  Thanks for the incredibly fast fix @A5rocks 🤩

Zac-HD:
Let's hold 0.23.2 until https://github.com/python-trio/trio/issues/2734 is finished, to make sure we haven't missed any similar issues.

PR #2875, opened 2023-11-10
Update black and ruff

CoolCat467:
This PR updates `black` to `23.11.0` and `ruff` to `0.1.5` in pre-commit as well as in the test requirements files.

Pre-commit autoupdate on its weekly update schedule will be running shortly regardless, but it won't change the test requirements files on its own.


jakkdl:
Maybe time to reconsider pulling those from `check.sh` and `test-requirements.txt`? Or if people still like `check.sh` can have that call `pre-commit`...... except that fucks with `test_gen_exports.py`. Though IMHO we can just skip that one automatically if ruff is not installed, since pre-commit/check.sh will raise errors if the files are not formatted correctly.

A5rocks:
Punting reconsidering to later, I'm going to merge this instead of #2876.

PR #2879, opened 2023-11-18
Test that Trio works around some CPython `.throw` bugs present in >= 3.9

gschaffner:
the current tests that ensure Trio works around CPython `.throw` bugs are all for `.throw` bugs that were fixed in 3.9.0 or earlier. however `.throw` is still buggy in every CPython release supported by Trio today.

to test the new tests (i.e. to verify that they do indeed fail if Trio removes its CPython `.throw` workaround before Trio drops support for 3.12.0 (at least)), one can apply 256568ea4e70c8a895f9add5cb9e45f58dace7e0 and run the test suite on CPython 3.9.x, 3.10.x, 3.11.x, 3.12.x, and/or 3.13-dev. when one does this, the two new tests fail (and they are the only tests that fail).

### DIFF
--- a/src/trio/_core/_tests/test_run.py
+++ b/src/trio/_core/_tests/test_run.py
@@ -1074,12 +1074,19 @@ async def test_exc_info() -> None:
     ]
 
 
-# Before CPython 3.9, using .throw() to raise an exception inside a
-# coroutine/generator causes the original exc_info state to be lost, so things
-# like re-raising and exception chaining are broken.
+# On all CPython versions (at time of writing), using .throw() to raise an
+# exception inside a coroutine/generator can cause the original `exc_info` state
+# to be lost, so things like re-raising and exception chaining are broken unless
+# Trio implements its workaround. At time of writing, CPython main (3.13-dev)
+# and every CPython release (excluding releases for old Python versions not
+# supported by Trio) is affected (albeit in differing ways).
 #
-# https://bugs.python.org/issue29587
-async def test_exc_info_after_yield_error() -> None:
+# If the `ValueError()` gets sent in via `throw` and is suppressed, then CPython
+# loses track of the original `exc_info`:
+#   https://bugs.python.org/issue25612 (Example 1)
+#   https://bugs.python.org/issue29587 (Example 2)
+# This is fixed in CPython >= 3.7.
+async def test_exc_info_after_throw_suppressed() -> None:
     child_task: _core.Task | None = None
 
     async def child() -> None:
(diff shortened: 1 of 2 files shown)


src/trio/_core/_tests/test_run.py
```diff
@@ -1088,21 +1095,28 @@ async def child() -> None:
...
-# Similar to previous test -- if the ValueError() gets sent in via 'throw',
-# then Python's normal implicit chaining stuff is broken.
-async def test_exception_chaining_after_yield_error() -> None:
+# Similar to previous test -- if the `ValueError()` gets sent in via 'throw' and
+# propagates out, then CPython doesn't set its `__context__` so normal implicit
+# exception chaining is broken:
+#   https://bugs.python.org/issue25612 (Example 2)
+#   https://bugs.python.org/issue25683
+#   https://bugs.python.org/issue29587 (Example 1)
+# This is fixed in CPython >= 3.9.
```
jakkdl, line 1118: I'm slightly confused that the test doesn't have any calls to `sys.version_info` when it talks about differing behaviour on different python versions. Can maybe clarify the comment a bit more in that respect?

src/trio/_core/_tests/test_run.py
```diff
@@ -1123,6 +1137,71 @@ async def child() -> None:
...
+            await wait_all_tasks_blocked()
+            _core.reschedule(not_none(child_task), outcome.Error(ValueError()))
+
+    assert isinstance(excinfo.value.__context__, IndexError)
+    assert isinstance(excinfo.value.__context__.__context__, KeyError)
+
+
 @pytest.mark.skipif(
     sys.version_info < (3, 6, 2), reason="https://bugs.python.org/issue29600"
 )
```
jakkdl, line 1207: hah, this has been missed a bunch of times. Opened #2884

jakkdl (approved):
I'm not able to comment on the actual logic or internals, but seems fine at a glance.

It would be great in theory to be able to automate applying the mentioned patch and xfail tests (something like env variables or monkeypatching some functions), but I'm not sure that's worth the effort and upkeep.

src/trio/_core/_tests/test_run.py
jakkdl: Right. It's only if we applied that patch that we'd get different behaviour depending on python version. And the workaround implemented by trio, for this specific bug/test, is redundant on py3.9+, and when we drop support for 3.8 in oct 2024 we *could* remove this test and any corresponding workaround.

PR #2880, opened 2023-11-18
Add trio.CapacityLimiter.wait_no_borrows

VincentVanlaer:
This is the equivalent of trio.testing.wait_all_tasks_blocked but for anything wrapped by a CapacityLimiter. This is useful when writing tests that use to_thread


njsmith:
I'm worried this will be an attractive nuisance. In tests where you know
exactly what tasks are supposed to be where, it makes total sense, but it
also *sounds* like something that would let you get exclusive access to
whatever the limiter is protecting. But that's not true; by the time it
returns, another task could have already claimed a token.

If the immediate goal is to be able to use it in a few tests, maybe we can
handle those with a less-elegant sleep/check loop, or some other way to
tell when the tasks are done?

VincentVanlaer:
> I'm worried this will be an attractive nuisance. In tests where you know exactly what tasks are supposed to be where, it makes total sense, but it also *sounds* like something that would let you get exclusive access to whatever the limiter is protecting. But that's not true; by the time it returns, another task could have already claimed a token.

Doesn't the `while` loop guarantee that when the function returns, there are no borrowed tokens? Meaning this is just a more efficient `while limiter.borrowed_tokens: await trio.lowlevel.checkpoint()`. Or am I misunderstanding something here?

In any case, perhaps another possibility that would communicate the intent better, is to add a function to `trio.testing` which does the same as `wait_no_borrowers`. It would take a capacity limiter and hooks into the internals of the capacity limiter to provide similar functionality without a busy/semi-busy loop. This is less clean on the implementation side though.

For context, I'm currently using this function as follows:

```python
async def wait_all_blocked():
    capacity_limiter = trio.to_thread.current_default_thread_limiter()
    while True:
        await capacity_limiter.wait_no_borrowers()
        if (capacity_limiter.borrowed_tokens == 0):
            break
```

VincentVanlaer:
@njsmith  Does my explanation or alternative proposal alleviate your concerns?

njsmith:
Would a `trio.testing.wait_all_threads_complete` be a good solution?

VincentVanlaer:
I have implemented your suggestion in https://github.com/python-trio/trio/pull/2937 as it's a complete reimplementation

VincentVanlaer:
Superseded by #2937

PR #2881, opened 2023-11-19
Use TypeVarTuple in our APIs

TeamSpen210:
With Mypy 1.7 being released, variadic generics are now supported. This works real well for `Nursery.start_soon()`, `to_thread.run_sync()`, etc. However the complex types for `Nursery.start()` + `TaskStatus` are producing a bunch of spurious errors for code that just calls it and ignores the return type. That might need to be fixed first in Mypy...


src/trio/_core/_run.py
```diff
@@ -70,19 +76,25 @@
     # for some strange reason Sphinx works with outcome.Outcome, but not Outcome, in
     # start_guest_run. Same with types.FrameType in iter_await_frames
     import outcome
-    from typing_extensions import Self
+    from typing_extensions import Self, TypeVarTuple, Unpack
+
+    PosArgT = TypeVarTuple("PosArgT")
+
+    # Needs to be guarded, since Unpack[] would be evaluated at runtime.
+    class _NurseryStartFunc(Protocol[Unpack[PosArgT], StatusT_co]):
```
A5rocks, line 84: Curious about why you need an Unpack here

A5rocks: Oh sorry, I should have paid more attention to pep 646: https://peps.python.org/pep-0646/#type-variable-tuples-must-always-be-unpacked

TeamSpen210: TypeVarTuple is basically always used in unpacked form. If we were doing 3.11+ only code, it'd be `Protocol[*PosArgT, StatusT_co]`, like we're unpacking the TypeVars inside the TVT.

A5rocks:
Could you add some type tests?

src/trio/_core/_run.py
```diff
@@ -70,19 +76,25 @@
...
+        """Type of functions passed to `nursery.start() <trio.Nursery.start>`."""
+
+        def __call__(
+            self, *args: Unpack[PosArgT], task_status: TaskStatus[StatusT_co]
```
A5rocks, line 88:
```suggestion
            self, *args: Unpack[PosArgT], *, task_status: TaskStatus[StatusT_co]
```

This should fix the mypy errors. This actually makes sense!

TeamSpen210: Unfortunately no, the `*args` already makes it keyword-only.

A5rocks: Yeah sorry about the repeated confusion with this thread, `TypeVarTuple` is getting to my head :(

TeamSpen210: Yeah it's real confusing. Since `task_status` is keyword-only it shouldn't get mixed up. As you can see with the type-tests, if it's being assigned to an annotated type, Mypy can do bidirectional inference and figure it out. But if it's not annotated or just a bare call, it gets confused and assumes it's `Never`. I think it's a bug.

A5rocks: I agree. I took a look at mypy's code for the last 45 minutes and spotted:
 1) It looks like it isn't generating constraints for the typevar.
 2) This is because the constraints only get generated for a callback protocol if the type being assigned is a subtype of the callback protocol w/ erased tvars (I assume this is to make less confusing error messages). Unfortunately, `TypeVarTuple` gets erased to `tuple[Any, ...]` and our callable doesn't take in infinite parameters.
 3) Even if that problem were to be solved, it looks like mypy only generates constraints for either a bunch of type vars or a single unpack. This isn't a problem normally (`Callable[[Unpack[Ts], T], None]` has an implicit `Unpack`) but means that mypy focuses on the typevartuple, not making a constraint for our type variable.

I'm not really sure how to solve this myself, a bug report would probably work nicely.

jakkdl: @A5rocks probably worth cross-posting your analysis in the issue that teamspen opened.

jakkdl:
for this PR: probably worth merging the signatures that work / type:ignoring the fails, so we don't have this lying around as a draft until the bug is resolved in mypy and get the incremental improvement. Especially as it pertains to deprecation of trio-typing

A5rocks:
I think it depends on whether we want to release our next bug fix release sooner or later. Specifically due to the pyright issue w/ current release. Might not be good to have regressions in a common function!

TeamSpen210:
I'll split off the `start()` annotations to their own PR, since that's the only part that's not working right.

TeamSpen210:
I removed the `start()` signatures, making Mypy pass this fine. However, Pyright seems to be failing - if I do the full check offline, it seems to not be recognising `typing_extensions.Unpack` as something special?

pyproject.toml
```diff
@@ -127,6 +127,7 @@ extend-exclude = [
 'src/trio/lowlevel.py' = ['F401']
 'src/trio/socket.py' = ['F401']
 'src/trio/testing/__init__.py' = ['F401']
+'**/type_tests/*' = ['F841']  # Annotated local vars are used to specify return values.
```
A5rocks, line 130: Is this still necessary? (should we be using `typing.assert_type` instead?)

A5rocks (commented):
I'm not sure why you're getting that from pyright, given it accepts this following program in VSCode on my machine:

```py
from typing import Callable
from typing_extensions import TypeVarTuple, Unpack

Ts = TypeVarTuple("Ts")

def x(f: Callable[[Unpack[Ts]], None], *args: Unpack[Ts]) -> None:
    return f(*args)

x(lambda x: print(x + 42), 42)
```

\-\-\-

re: pyright looks like it's not seeing the TypeVarTuple right?

pyproject.toml
TeamSpen210: In this case I was deliberately testing `start()` calls with/without the return type explicitly defined, to see if Mypy could infer them. It's not necessary right now after I removed that though.

A5rocks:
Looks like pyright just hates the name `PosArgT`? If I ctrl+f and replace then all with `PosArgTs` then it all works ...???

A5rocks:
Yeah I have no clue beyond that import being wacky -- FYI pyright also dislikes line 73 of the `nursery_start.py` type test.

Could you make issues on pyright/smaller reproducers xdd

(Also I think you should add this to `ci.sh` under the existing pyright run on type tests: `pyright src/trio/_core/_tests/type_tests || EXIT_STATUS=$?`)

TeamSpen210:
Okay, adding `__all__` lists to the `_generated_*` modules fixed the problem. I expect because the `PosArgsT` typevar is no longer being imported back into `_run`. It's probably good to do anyway, so a bunch of internal stuff doesn't get re-imported elsewhere in a confusing way.

A5rocks:
Just a bump on this before I review!:

> (Also I think you should add this to `check.sh` under the existing pyright run on type tests: `pyright src/trio/_core/_tests/type_tests || EXIT_STATUS=$?`)

jakkdl (approved):

TeamSpen210:
We may want to hold off on merging until the next Pyright version. It’ll have support for handling a callable with default arguments, making those optional parts of the TypeVarTuple. (microsoft/pyright#6623) We can then enable the test for that.

TeamSpen210:
In bbd436de90a0d80f9ec51953cb00ff4f61aa2f94 I added a `pyright:ignore`, since it was failing the test case I added for this. So we'd want to wait, so I can remove that ignore. Otherwise it'll start failing next time we update Pyright.

A5rocks:
Well pyright released a new version so...

pyproject.toml
A5rocks:
```suggestion
```

A5rocks (approved):
Looks good!

CoolCat467:
Oop, there's that segfault issue in pypy3.9 again. https://github.com/python-trio/trio/actions/runs/7120585043/job/19388115101

A5rocks:
> Oop, there's that segfault issue in pypy3.9 again. https://github.com/python-trio/trio/actions/runs/7120585043/job/19388115101

This one's different from the other one you might be thinking about (iirc?) and [it's one where I've provided more info in the issue](https://github.com/pytest-dev/pytest/issues/11168#issuecomment-1835317095) so hopefully it'll resolve itself soon!

A5rocks:
Mind checking my changes seem fine then merging? @TeamSpen210

Zac-HD:
Woohoo!  Can we release this soon, to get the lovely typing benefits without waiting to sort out the ExceptionGroup transitions?

PR #2882, opened 2023-11-19
Don't use absolute path for sleep binary in tests

VincentVanlaer:
While most OSes will have the `sleep` binary in `/bin`, this is not guaranteed (e.g. `/bin` only contains `sh` on NixOS). While in principle one might have an incompatible `sleep` somewhere in PATH, this seems unlikely.

### DIFF
--- a/src/trio/_tests/test_subprocess.py
+++ b/src/trio/_tests/test_subprocess.py
@@ -73,7 +73,7 @@ CAT = python("sys.stdout.buffer.write(sys.stdin.buffer.read())")
 if posix:
 
     def SLEEP(seconds: int) -> list[str]:
-        return ["/bin/sleep", str(seconds)]
+        return ["sleep", str(seconds)]
 
 else:
 


jakkdl:
https://docs.python.org/3/library/subprocess.html#popen-constructor says to use fully qualified path; and I think users aliasing or overriding `sleep` in various ways is not too crazy and would be way nastier to debug than getting a "file not found" that can be resolved with a symlink. I also have vague memories that fully qualified are preferred to avoid security risks.
But I think that `/usr/bin/sleep` might be more universal than `/bin/sleep`?

Fuyukai:
If you really want to do this properly it should be ``/usr/bin/env sleep``.

jakkdl:
> Possibly, but in my case that wouldn't work, as on NixOS those binaries are located in `/nix/store/...`.


whoa - this really surprised me. I dug into threads like https://discourse.nixos.org/t/why-doesnt-e-g-bin-link-to-run-current-system-sw-bin/1562/6 and  https://discourse.nixos.org/t/add-bin-bash-to-avoid-unnecessary-pain/5673 to understand the motivation; and while I'm not sure I fully buy into the philosophy it makes some sense. Given that NixOS is relatively prevalent and we're only talking about tests I'm fine with the change.
If this turns out to break some other distro that complains louder we can reverse it and NixOS users will have to deal with ActivationScripts or whatever.


I'm not sure how `env` plays into this and

VincentVanlaer:
> > Possibly, but in my case that wouldn't work, as on NixOS those binaries are located in `/nix/store/...`.
> 
> whoa - this really surprised me. I dug into threads like https://discourse.nixos.org/t/why-doesnt-e-g-bin-link-to-run-current-system-sw-bin/1562/6 and https://discourse.nixos.org/t/add-bin-bash-to-avoid-unnecessary-pain/5673 to understand the motivation; and while I'm not sure I fully buy into the philosophy it makes some sense. Given that NixOS is relatively prevalent and we're only talking about tests I'm fine with the change. If this turns out to break some other distro or system setup that complains louder we can reverse it and NixOS users will have to deal with ActivationScripts or whatever.

Perhaps, as a best of both worlds case, it's still possible to check if `/bin/sleep` exists and otherwise fall back to just using whatever is in `PATH`.

A5rocks (approved):
As long as this works on people's setups this looks fine to me!

(I think the main reason you'd want a full-qualified path is security which isn't really a factor here because it's just test cases and an attacker able to change your path should already have execution rights at your level (I forget if the CWD is added to path... but I don't think it's worth making this secure). But maybe I'm wrong.)

jakkdl (approved):
feel free to change it to using `/bin/sleep` if it exists (probably need a `# pragma: no coverage` for it though), but seems fine either way.

PR #2884, opened 2023-11-22
remove skipif not relevant since py<3.6.2 was supported

jakkdl:


A5rocks (approved):
Thanks, I noticed this too when looking another PR but didn't have the motivation to make this!

Issue #2785, opened 2023-09-02
Remove `MultiError` and set `strict_exception_groups=True` by default


Zac-HD:
## Goal

Our long-term goal is for Trio and our users to use the builtin `ExceptionGroup` type and `except*` syntax, and to _always_ raise `ExceptionGroup` where we might _ever_ do so.  https://github.com/python-trio/trio/pull/2213 started moving us towards this goal, following discussion in https://github.com/python-trio/trio/issues/2211 (and https://github.com/python-trio/trio/issues/611), and this issue is about how and when we might finish it.

## Proposal

Remove `MultiError`, using the builtin (or backported) `ExceptionGroup` type and catching logic instead.  At the same time, change the default value of `strict_exception_groups` to `True`, to match the behavior of asyncio and anyio `TaskGroup`.

Later, when `strict_exception_groups=False` has become very rare or Python 3.10 reaches end of life, we'll remove the `strict_exception_groups` argument entirely.  That's a future issue!

## Timing and todos

With the growing popularity of `ExceptionGroup` in general and `anyio` in particular, I'm keen to ship this in the next few months - making Trio as "normal" as possible makes it much easier for prospective users to adopt and learn.  That said, I think there are a few things we should or could get done first:

- [x] Wait for `anyio` 4.0, which uses the standard `ExceptionGroup` type
- [x] IPython now handles `ExceptionGroup` so we can remove our workaround https://github.com/python-trio/trio/pull/2702
- [x] Release Trio 0.23 and get our new-and-improved type annotations shaken out.
  - [x] Fix https://github.com/python-trio/trio/issues/2783
- [x] Fix https://github.com/python-trio/trio/issues/2611
  - [x] check for other places where Trio itself might raise groups when a single exception is intended
  - [ ] consider documenting any design patterns we like or dislike for interfaces with an "inner nursery", c.f. https://github.com/python-trio/trio-websocket/issues/132
- [x] Propose moving our `apport` monkeypatch to the `exceptiongroup` backport, https://github.com/agronholm/exceptiongroup/issues/23 / https://github.com/agronholm/exceptiongroup/pull/88

jakkdl:
In #2301 you previously said
[...]

PR #2886, opened 2023-11-24
change strict_exception_groups default to True

jakkdl:
this commit ... is a monster. I could try to split it up some things into separate PRs, but a lot of it is hard to separate.

* See discussion in #2785, which this fixes most of.
* I've written an implementation for `pytest.raises` that handles `exceptiongroup`s. This should probably be added upstream, or in a separate package, or do a way less illegal way of interaction. `mypy` also goes mad on it, as it should. Don't focus on reviewing `_exceptiongroup_util.py` for now. See https://github.com/pytest-dev/pytest/issues/11538
* A *ton* of tests had to be changed with the functionality change. I started out by passing `strict_exception_groups=False`, but decided that embracing the future is much better in the long term and it unveiled a bunch of issues.
* Most tests are fine, but I haven't looked super closely at all instances where exceptiongroups now get raised, or don't, so the tests diff should be looked over very closely I've marked a couple instances where it seemed clearly suspect that exceptions did or didn't get wrapped - even got an instance of triple-wrapped exceptions that probably should be fixed in a separate PR.
* `_assert_raises` has been removed. It seems to be a relic with no reason for existing
* several tests were likewise ... weirdly written, or simply didn't have the almighty power of `ExpectedExceptionGroup`, and have been simplified.
* I left actually removing `MultiError` and changing nurseries etc to raise `ExceptionGroup`s instead to a separate PR. But changed some tests to check for identity with `ExceptionGroup` instead of `MultiError`. (ignore the name of the branch >.>)

This is a pretty big change, and touches on lots and lots of stuff, and should figure out a solution for `raises`/`ExpectedExceptionGroup`, so I expect this to stay as a draft for quite a while - maybe with other PRs splitting out from it and implementing parts of it or fixing weird [non-]wrapping.


jakkdl:
Other than `ExpectedExceptionGroup` not being final, all tests pass for me locally. Remains to see if I messed up some mac/windows/pypy-only test

src/trio/_core/_tests/test_ki.py
```diff
@@ -331,7 +334,8 @@ def kill_during_shutdown() -> None:
 
         token.run_sync_soon(kill_during_shutdown)
 
-    with pytest.raises(KeyboardInterrupt):
+    # Does not wrap in an ExceptionGroup(!!)
+    with testing.raises(KeyboardInterrupt):
```
A5rocks, line 338: I would love an explanation in a comment for both this and the subsequent one. I think I figured out why ("if it's from a sync function, then you cannot have code running concurrently") and hopefully that's right?

src/trio/_core/_tests/test_run.py
```diff
@@ -123,10 +124,11 @@ async def test_nursery_warn_use_async_with() -> None:
 async def test_nursery_main_block_error_basic() -> None:
     exc = ValueError("whoops")
 
-    with pytest.raises(ValueError) as excinfo:
+    with pytest.raises(ExpectedExceptionGroup(ValueError)) as excinfo:
         async with _core.open_nursery():
             raise exc
-    assert excinfo.value is exc
+    assert len(excinfo.value.exceptions) == 1
```
A5rocks, line 130: I noticed you removed this kind of assertation (checking that the `ExceptionGroup` has *exactly* one exceptions) elsewhere. Is this redundent? Or should `ExpectedExceptionGroup` has a `strict` kwarg that means it doesn't take other exceptions (in a specific ordering? not sure about the impl...)?

src/trio/_core/_tests/test_run.py
```diff
@@ -178,16 +181,15 @@ async def main() -> None:
             nursery.start_soon(looper)
             nursery.start_soon(crasher)
 
-    with pytest.raises(ValueError) as excinfo:
+    with pytest.raises(ExpectedExceptionGroup(ValueError("argh"))):
```
A5rocks, line 184: Hm. Earlier you checked `ExpectedExceptionGroup(ValueError)` then checked `excinfo.value.exceptions[0] is exc` but wouldn't that be `ExpectedExceptionGroup(exc)` to be similar to this?

Probably not because `is` vs `==`? (I am being very confused right now :( )

src/trio/_core/_tests/test_run.py
```diff
@@ -214,19 +211,15 @@ async def main() -> None:
             nursery.start_soon(crasher, KeyError)
             nursery.start_soon(crasher, ValueError)
 
-    with pytest.raises(MultiError) as excinfo:
+    with pytest.raises(ExpectedExceptionGroup((ValueError, KeyError))):
         _core.run(main)
-    assert {type(exc) for exc in excinfo.value.exceptions} == {
-        ValueError,
-        KeyError,
-    }
```
A5rocks, line 222: These are significantly cleaner, nice!

src/trio/testing/_exceptiongroup_util.py
```diff
@@ -0,0 +1,385 @@
...
+            self.expected_exceptions = exceptions
+        else:
+            self.expected_exceptions = (exceptions, *args)
+        if not all(
+            isinstance(exc, (type, BaseException, ExpectedExceptionGroup))
+            for exc in self.expected_exceptions
+        ):
+            raise ValueError(
+                "All arguments must be exception instances, types, or ExpectedExceptionGroup."
+            )
```
A5rocks, line 72: Both errors could be checked in `test_testing`, probably.

src/trio/testing/_exceptiongroup_util.py
```diff
@@ -0,0 +1,385 @@
...
+
+    # TODO: TypeGuard
+    def matches(
+        self,
+        exc_val: Optional[BaseException],
+    ) -> bool:
```
A5rocks, line 78:
```suggestion
    ) -> TypeGuard[BaseExceptionGroup]:
```

?

src/trio/testing/_exceptiongroup_util.py
```diff
@@ -0,0 +1,385 @@
...
+        if exc_val is None:
+            return False
+        if not isinstance(exc_val, BaseExceptionGroup):
+            return False
+        if len(exc_val.exceptions) != len(self.expected_exceptions):
```
A5rocks, line 83: Oh gah, I assumed this would be like `except*` where you're just checking if certain exceptions are in the final group. Is that more useful behavior? With stricter matching sometimes? (though... I can't think of a scenario when you would *want* that non-strict behavior).

src/trio/testing/_exceptiongroup_util.py
```diff
@@ -0,0 +1,385 @@
...
+                        and isinstance(rem_e, ExpectedExceptionGroup)
+                        and rem_e.matches(e)
+                    )
+                    or (
+                        isinstance(rem_e, BaseException)
+                        and isinstance(e, type(rem_e))
+                        and re.search(str(rem_e), str(e))
+                    )
+                ):
+                    remaining_exceptions.remove(rem_e)  # type: ignore # ??
```
A5rocks, line 103: What's the typechecker error?

A5rocks (commented):
Some initial comments

src/trio/_core/_tests/test_run.py
jakkdl: Yeah `ExpectedExceptionGroup` by default checks that *exactly* the exceptions you specify get raised, no more, no less. A kwarg that changes that behaviour would make sense though

src/trio/_core/_tests/test_run.py
jakkdl: yeah, `ExpectedExceptionGroup` checks with `isinstance` - but this test previously did an identity check w/ the caught exception so I thought I'd keep that check. Will add a comment

Most tests can't check with identity as they're not the one raising, so makes some sense that this one does when it can, to check that a different exception with the same type didn't get raised

src/trio/testing/_exceptiongroup_util.py
jakkdl: yeah I was confused that https://github.com/pytest-dev/pytest/pull/11424 works in that loose sense - it seems completely backwards to me and, for testing, almost never what you want.

src/trio/testing/_exceptiongroup_util.py
jakkdl: yeah I have unit tests in my wip pytest PR. I'll get that opened and depending on reception decide how to proceed with defining & testing it here

src/trio/_core/_tests/test_ki.py
jakkdl: I don't know! (and haven't looked into them yet)

And regardless we might consider whether it's the best interface to have them differ, or if inner nurseries should be hidden in some cases (either by reraising, or running stuff with `strict_exception_groups=False`)

docs/source/reference-core.rst
```diff
@@ -802,9 +802,9 @@ to set the default behavior for any nursery in your program that doesn't overrid
   wrapping, so you'll get maximum compatibility with code that was written to
   support older versions of Trio.
 
-To maintain backwards compatibility, the default is ``strict_exception_groups=False``.
-The default will eventually change to ``True`` in a future version of Trio, once
-Python 3.11 and later versions are in wide use.
+The default is set to ``strict_exception_groups=True`` in line with the default behaviour
+of ``TaskGroup`` in asyncio and anyio. This is also to avoid any bugs caused by only
+catching one type of exceptions/exceptiongroups.
```
Zac-HD, line 807:
```suggestion
The default is set to ``strict_exception_groups=True``, in line with the default behaviour
of ``TaskGroup`` in asyncio and anyio.  We've also found that non-strict mode makes it 
too easy to neglect the possibility of several exceptions being raised concurrently,
causing nasty latent bugs when errors occur under load.
```

src/trio/testing/_exceptiongroup_util.py
jakkdl: `error: Argument 1 to "remove" of "list" has incompatible type "Union[BaseException, Type[E], ExpectedExceptionGroup[E]]"; expected "Union[E, Type[E], ExpectedExceptionGroup[E]]"  [arg-type]`
i.e. `BaseException != E` where `E = TypeVar(BaseException, bound=BaseException)`

Is that something to do with the variance? (I still don't understand variance at all >.>)

A5rocks: Yeah this has to do with variance. `list[int]` will not accept `.remove(<float>)`. Not sure how to fix it though.

src/trio/_core/_tests/test_run.py
```diff
@@ -1587,16 +1583,30 @@ async def main() -> None:
...
-
         async def async_gen(arg: T) -> AsyncGenerator[T, None]:  # pragma: no cover
             yield arg
 
-        with pytest.raises(
-            TypeError, match="expected an async function but got an async generator"
-        ):
-            bad_call(async_gen, 0)  # type: ignore
+        # this is obviously horribly ugly code
+        # but one raising an exceptiongroup and one not doing so is probably bad
```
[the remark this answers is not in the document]
jakkdl: wait, no. Looking closer at this `bad_call_spawn` explicitly opens a nursery in its `main` function which it then passes to `_core.run`, where as `bad_call_run` calls `_core.run(func, *args)` directly. So I don't think this is a problem at all and I just need to rewrite the test more cleanly. (and add a comment explaining why the results differ)

src/trio/_core/_tests/test_ki.py
jakkdl: Looked into it:
1. `check_unprotected_kill` opens a nursery, and only inside of it does it call `raiser`, so it will be wrapped in an exceptiongroup from the nursery.
2. `check_protected_kill` does the same.
3. `check_kill_during_shutdown` does not touch any nurseries, so there's nothing that would wrap exceptions
4. `main_1` likewise does not do anything with nurseries.

So it seems very straightforward, will update/add comments.

src/trio/_core/_tests/test_run.py
jakkdl: in this case this is more of a question of the behavior of ExpectedExceptionGroup/RaisesGroup. But once the RaisesGroup PR is merged and I rebase this on top of it I will rewrite the former to do
```python
with RaisesGroup(Matcher(ValueError, check=lambda x: x is exc)):
  ...
```

jakkdl:
Reviewed two cases where I thought ExceptionGroups were missing, and they turned out to be fairly simple cases. Next steps:
- [ ] update comments & docs from review
- [ ] rebase on top of #2898 once merged
- [ ] rebase on top of #2891 once merged

CoolCat467:
pre-commit.ci autofix

jakkdl:
Ah right, a plethora of PT012. Well, first gotta settle on what message to add to them. PT017 in the helper is pretty amusing.

src/trio/_core/_tests/test_run.py
```diff
@@ -2547,14 +2575,10 @@ async def main() -> NoReturn:
 
 async def test_nursery_strict_exception_groups() -> None:
     """Test that strict exception groups can be enabled on a per-nursery basis."""
```
jakkdl, line 2577: TODO: parametrize/change/duplicate these and related ones for `strict_exception_groups=False`

src/trio/_core/_tests/test_run.py
```diff
@@ -940,10 +942,14 @@ async def main() -> None:
...
 
+    # not wrapped in ExceptionGroup
     with pytest.raises(_core.TrioInternalError) as excinfo:
         _core.run(main)
 
-    assert RaisesGroup(KeyError, ValueError).matches(excinfo.value.__cause__)
+    # TODO: triple-wrapped exceptions ?!?!
+    assert RaisesGroup(RaisesGroup(RaisesGroup(KeyError, ValueError))).matches(
+        excinfo.value.__cause__
+    )
```
jakkdl, line 952: The reason for the triple-wrapping turns out to be quite straightforward:
https://github.com/python-trio/trio/blob/aadd1ea56f2b64d920a70897abb1a7178b54cfab/src/trio/_core/_run.py#L1903-L1913

doesn't make for the cleanest `__cause__`s on `TrioInternalError`s, but don't think it's a problem. Updating comments.

src/trio/_subprocess.py
```diff
@@ -760,6 +760,9 @@ async def read_output(
             async for chunk in stream:
                 chunks.append(chunk)
 
+    # TODO: if this nursery raises an exceptiongroup, manually raise a CalledProcessError,
+    # or maybe a InternalTrioError, with the exceptiongroup as a cause, so run_process
+    # never raises an exceptiongroup. (?)
     async with trio.open_nursery() as nursery:
         # options needs a complex TypedDict. The overload error only occurs on Unix.
         proc = await open_process(command, **options)  # type: ignore[arg-type, call-overload, unused-ignore]
```
jakkdl, line 768: .

src/trio/_tests/test_subprocess.py
```diff
@@ -644,14 +646,24 @@ async def test_for_leaking_fds() -> None:
...
     assert set(SyncPath("/dev/fd").iterdir()) == starting_fds
 
+    # TODO: should `run_process` always, never, or sometimes raise exceptiongroups?
+    # With `strict_exception_groups=false` `read_output` could currently return an
+    # exceptiongroup, though it's not trivial to write a test case that would trigger it.
+
+    # Easiest way I see to is to monkeypatch `open_process` to return a faulty `trio.Process`
+    # when `run_process` calls it, that will raise errors when its stdin/stdout/stderr are
+    # accessed if we've specified capture_[stdin/stdout/stderr]=True.
+
```
jakkdl, line 666: .

jakkdl (commented):
I've been procrastinating on cleaning up the language in docstrings/documentation. But other than that I think this is the only remaining problematic case where a function sometimes raises a group and sometimes doesn't.
There's maybe better solutions, but I'd just wrap the nursery call in a try/except and raise a new error with the caught exceptiongroup as a cause, as I described in the comment in _subprocess.py.

docs/source/reference-core.rst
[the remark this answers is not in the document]
Zac-HD: I think that we'll eventually want to rewrite this whole section from the new strict-by-default perspective, under the heading "Deprecated: non-strict `ExceptionGroup`s" - to explain that it only exists for backwards-compatibility, will be removed in future, and that we recommend against it for all new code.

I don't want to delay this PR further though, so let's open an issue to revisit the docs on this topic rather than waiting to merge.

src/trio/_core/_run.py
```diff
@@ -2207,9 +2207,9 @@ def run(
           main thread (this is a Python limitation), or if you use
           :func:`open_signal_receiver` to catch SIGINT.
 
-      strict_exception_groups (bool): If true, nurseries will always wrap even a single
-          raised exception in an exception group. This can be overridden on the level of
-          individual nurseries. This will eventually become the default behavior.
+      strict_exception_groups (bool): Unless set to false, nurseries will always wrap
+          even a single raised exception in an exception group. This can be overridden
+          on the level of individual nurseries.
```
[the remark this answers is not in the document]
Zac-HD: I think it's important to add that "`=False` is deprecated, and will be removed in a future version of Trio."

src/trio/_subprocess.py
Zac-HD: (surfacing from an ~hour-long dive into `run_process()`)  I think it's sufficiently unlikely to get an `ExceptionGroup` here from any reasonable usage, that I'd favor `raise InternalTrioError from the_group`.

Zac-HD (commented):
Getting very close now 😀

src/trio/_subprocess.py
jakkdl: Getting multiple exceptions from inside the nursery may be unlikely, but it seems relatively straightforward to pass a bad command that will get an exception from `open_process` - and we probably don't want to raise `InternalTrioError` in those cases. Pushing a commit with a suggested fix, but I suspect there might be a reason the `open_process` call was placed inside the nursery in the first place (?).

jakkdl: (the `PermissionError` in the test is raised inside `open_process`, and there's validation of inputs that `run_process` punts to `open_process` as well)

newsfragments/2786.breaking.rst
```diff
@@ -0,0 +1,2 @@
+``strict_exception_groups`` now defaults to True in ``trio.run`` and ``trio.start_guest_run``, as well as ``trio.open_nursery`` as a result of that.
```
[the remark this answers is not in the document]
jakkdl: kinda awkward wording, but should be better now

src/trio/_core/_run.py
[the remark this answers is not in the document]
jakkdl: I think we need to add a `DeprecationWarning` when specifying `False` to use a wording like that. But if we wanna do that as well in the same release I can get going on it in a separate PR.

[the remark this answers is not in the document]
jakkdl: added note about `=False` being deprecated&removed in the future, and updated `open_nursery` docstring.

src/trio/_subprocess.py
```diff
@@ -760,38 +763,45 @@ async def read_output(
...
+                            await deliver_cancel(proc)
+
+                    nursery.start_soon(killer)
+                    await proc.wait()
+                    killer_cscope.cancel()
+                    raise
+    except BaseExceptionGroup as exc:
+        if all(isinstance(e, Cancelled) for e in exc.exceptions):
+            raise exc
+        raise TrioInternalError("Error interacting with opened process") from exc
```
jakkdl, line 804: I'd love some extra eyes on this particular change from people more familiar with the inner workings of trio. Maybe @richardsheridan ?

jakkdl (commented):
Thorough rewrite of docs and an eventual DeprecationWarning for setting `strict_exception_groups=False` will have to wait for another PR. But other than uncertainty on messing with the nursery in `open_process` I think everything has been resolved.

src/trio/_subprocess.py
richardsheridan: I had thoughts on this. the first one is opening the process outside of the nursery, which you already did. the second one is that if `deliver_cancel` raises something other than `Cancelled`, it would be ideal to raise that error not in an exception group (or as a trio internal error.) As you might have seen in chat, there are several ways to accomplish this and none are very nice. Finally, I'm not sure it's valuable to convert pipe reader task error groups to a trio internal error. They may well be critical/fatal OS errors outside of trio's control. But since either the old or new way will be practically uncatchable, I don't have a strong preference.

jakkdl: It's fairly easy to add a
```python
[ re-raise cancelled ]
if len(exc.exceptions) == 1:
  raise exc.exceptions from None
[ raise triointernalerror ]
```
clause to pass through single errors, and only resort to `TrioInternalError` (or something else) in group cases. But then we're causing users a headache if the same underlying exception can cause them to get two different exceptions depending on if a second exception occurred at the same time.
`TrioInternalError` does seem somewhat off, especially if we consider getting a single error not to be an internal error. But not sure what else to raise. RuntimeError? Or be upfront that we would raise a group in that case? Always wrap exceptions in a group? Just raise the first non-cancelled error, and have the group as a `__cause__`?

richardsheridan: sanity check: if I set the nursery in question here to `strict_exception_groups=False`, and then fudge `test_custom_deliver_cancel` so line 589 reads `assert False`, shouldn't I get a single unwrapped `AssertionError`? Right now I'm getting an EG

jakkdl: which nursery is "the nursery in question", the one in test_subprocess L591, or the one in _subprocess line ~771?
If you're changing the former, you should get an EG, because it doesn't set the global contect of strict_exception_groups=False, so the one on 771 will use the default of True (in this PR).

jakkdl: but passing a bad `deliver_cancel` is indeed a very trivial way to get an EG atm, hrm.

richardsheridan:
oh and I meant the nursery in  _subprocess line ~771

jakkdl:
> oh and I meant the nursery in _subprocess line ~771

huh, then I'm confused.

jakkdl:
Any remaining semi-blockers have been cleared for this:
* https://github.com/python-trio/trio/releases/tag/v0.24.0 was released with #2891, and that has not seen much (any?) negative responses.
* #2923 seems to be settling down to preserving status quo of sometimes raising an ExceptionGroup in `run_process`

So I'll rebump review request from @A5rocks and/or @TeamSpen210. But unless @Zac-HD finds anything else on a re-review I think this can be merged.

Zac-HD (approved):
Looks great!  Since this PR has been open and at times growing for two months, I'm going to merge, and open an issue for the docs followup mentioned above.  Post-merge reviews entirely welcome; I suggest comments here if you want to discuss or just open an issue for some followup action if you spot something we should change further.

Huge thanks to @jakkdl for all your work on this, and to @A5rocks, @CoolCat467, @richardsheridan, and @TeamSpen210 for reviews 🤩  I'm really looking forward to getting this into production at work, and seeing the ecosystem improve handling of concurrent errors!

Issue #2885, opened 2023-11-24
CI changes


A5rocks:
Fixing existing runs:
- [ ] Don't run Jedi in PyPy.
- [ ] Only run autodeps in `python-trio/trio`

Adding more runs:
- [ ] Add an anaconda run
- [ ] Try out qemu for a freebsd run?

These feel like they will take more motivation than I have :(

jakkdl:
A more fun change: Add python 3.13

TeamSpen210:
We also aren't running 3.11/3.12 on Windows.

jakkdl:
weird CI fail: https://github.com/python-trio/trio/actions/runs/7232062095/job/19705878765?pr=2891

jakkdl:
Add back Alpine Linux for testing musl libc.
https://github.com/python-trio/trio/pull/2917

jakkdl:
Sleep timing fail on MacOS pypy3.10 https://github.com/python-trio/trio/actions/runs/7448486091/job/20262980216?pr=2886
```
FAILED ../../../_temp/116e3acf-3f1d-45b1-afa8-48d74a477716/pypy-c-jit-184208-0c87577e022f-macos_x86_64/lib/pypy3.10/site-packages/trio/_tests/test_timeouts.py::test_sleep - assert (1.70860500699996 / 1.0) < 1.5
```

Not especially weird, and only problematic if it's encountered regularly.

jakkdl:
Is it possible to trigger actions on "push or pull request" rather than "push and pull request"? I started #2942 on a branch in `trio/` to be able to run tests w/o spamming a PR, but once I opened the PR I'm getting double runs.

TeamSpen210:
A few possible ways:
* Using [concurrency groups](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) - create a group name including the commit hash, then it'd only allow one of the pair to run at a time. Then we'd set "cancel-in-progress" to true only on the PR run, causing the push to be cancelled. Downside is that the push events are still run and cancelled, wasting time and cluttering the runs list.
* We could set the push event to only apply to `master`, so you'd need to open a PR to get CI. May not be helpful.
* We could add a job to the workflow that runs before everything else on push requests. It'd fetch all PRs via the API, get their hash, figure out if we match and then skip the rest. Shouldn't be too complex?

A5rocks:
[...]

PR #2887, opened 2023-11-24
Export test fixes

A5rocks:
Was looking at export tests for https://github.com/python-trio/trio/issues/2885 and spotted a few changes to make!

### DIFF
--- a/src/trio/_tests/test_exports.py
+++ b/src/trio/_tests/test_exports.py
@@ -229,10 +229,6 @@ def test_static_tool_sees_all_symbols(tool: str, modname: str, tmp_path: Path) -
     else:  # pragma: no cover
         raise AssertionError()
 
-    # mypy handles errors with an `assert` in its branch
-    if tool == "mypy":
-        return
-
     # It's expected that the static set will contain more names than the
     # runtime set:
     # - static tools are sometimes sloppy and include deleted names
--- a/src/trio/_tests/test_exports.py
+++ b/src/trio/_tests/test_exports.py
@@ -157,6 +157,9 @@ def test_static_tool_sees_all_symbols(tool: str, modname: str, tmp_path: Path) -
         ast = linter.get_ast(module.__file__, modname)
         static_names = no_underscores(ast)  # type: ignore[arg-type]
     elif tool == "jedi":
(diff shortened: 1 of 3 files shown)


A5rocks:
pre-commit.ci autofix

CoolCat467 (approved):
Looks good to me

src/trio/_tests/test_exports.py
```diff
@@ -157,6 +157,9 @@ def no_underscores(symbols: Iterable[str]) -> set[str]:
         ast = linter.get_ast(module.__file__, modname)
         static_names = no_underscores(ast)  # type: ignore[arg-type]
     elif tool == "jedi":
+        if sys.implementation.name != "cpython":
+            pytest.skip("jedi does not support pypy")
```
jakkdl, line 161:
```suggestion
        if sys.implementation.name != "cpython":  # pragma: no branch
            pytest.skip("jedi does not support pypy")
```

src/trio/socket.py
```diff
@@ -67,13 +67,19 @@
     ntohs as ntohs,
 )
 
+# TODO: update python docs to reflect this
```
jakkdl, line 70: mostly just this that's a blocker

jakkdl (approved):

src/trio/_tests/test_exports.py
A5rocks: This feels weird though, like I don't think codecov should be complaining about this :(

src/trio/socket.py
A5rocks: Hopefully done with https://github.com/python/typeshed/pull/11073 and https://foss.heptapod.net/pypy/pypy/-/issues/4040.

A5rocks: Well I made a typeshed PR thinking "oh well the stdlib docs are the spec, so PyPy is deviating from it and should not have the docs update for it" but then now I realize `sys.implementation.name` isn't a guard typeshed allows :(

And it looks like these functions just never exist on PyPy. Lol.

TeamSpen210 (approved):

CoolCat467:
With ca46414 eliminating jakkdl's blocker, I am going to merge this.

PR #2891, opened 2023-11-27
remove multierror

jakkdl:
The companion to #2886, see #2785

- [x] Clean up comments and test names to not refer to multierror [in present tense]
- [x] update docs
- [x] add newsfragment
- [x] and probably some more tests in `test_multierror` that could be retired.
- [x] And files/directories referencing "multierror" to be renamed.


jakkdl:
do we want a newsfragment for now relying on exceptiongroup's apport excepthook monkeypatching?

richardsheridan:
> I experimented with removing the code that's supposedly there to fix `test_MultiError_catch_doesnt_create_cyclic_garbage` and it didn't seem to break it - but `test_locals_destroyed_promptly_on_cancel` did start failing. @A5rocks are those workarounds still needed? Can the cyclic garbage test be retired? Should I change those comments to refer to the destroyed_promptly test instead?

Ya these cyclic gc tests have somewhat overlapping behavior, since we made them by tracking down cycles basically one-by-one in various scenarios. I'd say it's not trio's job to check if exception groups create garbage (anymore). The question is whether a similar test exists in the suite for exception groups in cpython or the backport.

richardsheridan:
> Should I change those comments to refer to the destroyed_promptly test instead?
 
oh and yes please on this

CoolCat467:
I totally agree with using python's built in ExceptionGroups, but what about python versions pre-3.11? My worry is for one of my own projects, I support 3.9 and up, but ExceptionGroup doesn't exist pre-3.11.

My particular use case of MultiError:
Shim code:
https://github.com/CoolCat467/Scanner-Server/blob/0a64a32fee92a4b9e3d80ad46d9c9b7fcb595f65/src/sanescansrv/server.py#L68-L71
Main usage:
https://github.com/CoolCat467/Scanner-Server/blob/0a64a32fee92a4b9e3d80ad46d9c9b7fcb595f65/src/sanescansrv/server.py#L645-L657

If I were using 3.11 + I would just use `except *` but that's not a thing in older python versions.

CoolCat467:
> I totally agree with using python's built in ExceptionGroups, but what about python versions pre-3.11? My worry is for one of my own projects, I support 3.9 and up, but ExceptionGroup doesn't exist pre-3.11.
> 
> My particular use case of MultiError: Shim code: https://github.com/CoolCat467/Scanner-Server/blob/0a64a32fee92a4b9e3d80ad46d9c9b7fcb595f65/src/sanescansrv/server.py#L68-L71 Main usage: https://github.com/CoolCat467/Scanner-Server/blob/0a64a32fee92a4b9e3d80ad46d9c9b7fcb595f65/src/sanescansrv/server.py#L645-L657
> 
> If I were using 3.11 + I would just use `except *` but that's not a thing in older python versions.

Oh I see, we can just use external `exceptiongroup` module. Sorry about that, I didn't know that was a thing.

CoolCat467 (approved):
Looks good other than a few small things I noticed.

CoolCat467:
> Oh I see, we can just use external `exceptiongroup` module. Sorry about that, I didn't know that was a thing.

Fixed as of https://github.com/CoolCat467/Scanner-Server/commit/9ad2774fe7aab3d70a879ad3323685d89d1dd88f

src/trio/_core/_concat_tb.py
```diff
@@ -0,0 +1,128 @@
...
+        _ctypes.Py_DECREF(old_tb_frame)  # type: ignore[attr-defined]
+
+        c_new_tb.tb_lasti = base_tb.tb_lasti
+        c_new_tb.tb_lineno = base_tb.tb_lineno
+
+        try:
+            return new_tb
+        finally:
+            # delete references from locals to avoid creating cycles
+            # see test_ExceptionGroup_catch_doesnt_create_cyclic_garbage
```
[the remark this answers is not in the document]
richardsheridan: right, should be changed to `test_locals_destroyed_promptly_on_cancel`

jakkdl:
> > Oh I see, we can just use external `exceptiongroup` module. Sorry about that, I didn't know that was a thing.
> 
> Fixed as of [CoolCat467/Scanner-Server@9ad2774](https://github.com/CoolCat467/Scanner-Server/commit/9ad2774fe7aab3d70a879ad3323685d89d1dd88f)

trio also has an exceptiongroup dependency on <3.11

src/trio/_core/_run.py
```diff
@@ -1108,8 +1112,11 @@ def aborted(raise_cancel: _core.RaiseCancelT) -> Abort:
         assert popped is self
         if self._pending_excs:
             try:
-                return MultiError(
-                    self._pending_excs, _collapse=not self._strict_exception_groups
+                if not self._strict_exception_groups and len(self._pending_excs) == 1:
+                    return self._pending_excs[0]
+                return BaseExceptionGroup(
+                    "collapsible" if not self._strict_exception_groups else "TODO",
```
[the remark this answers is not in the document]
jakkdl: The exception requires specifying a message, so we probably want some informative message on the cause of the exception or something.

jakkdl:
> > I experimented with removing the code that's supposedly there to fix `test_MultiError_catch_doesnt_create_cyclic_garbage` and it didn't seem to break it - but `test_locals_destroyed_promptly_on_cancel` did start failing. @A5rocks are those workarounds still needed? Can the cyclic garbage test be retired? Should I change those comments to refer to the destroyed_promptly test instead?
> 
> Ya these cyclic gc tests have somewhat overlapping behavior, since we made them by tracking down cycles basically one-by-one in various scenarios. I'd say it's not trio's job to check if exception groups create garbage (anymore). The question is whether a similar test exists in the suite for exception groups in cpython or the backport.

I'm not finding anything with a quick search in the backport @agronholm 

I'm not familiar with the inner workings of gc and the implementation of multierror/exceptiongroup, but wouldn't the need for such tests depend on the manner of implementation? If the garbage was caused by hacky stuff that multierror needed to do in order to work, and cpython manages to sidestep that completely by implementing stuff """properly""", then there's no direct need for the tests on their side and/or they might have to look quite different.

agronholm:
> > > I experimented with removing the code that's supposedly there to fix `test_MultiError_catch_doesnt_create_cyclic_garbage` and it didn't seem to break it - but `test_locals_destroyed_promptly_on_cancel` did start failing. @A5rocks are those workarounds still needed? Can the cyclic garbage test be retired? Should I change those comments to refer to the destroyed_promptly test instead?
> > 
> > 
> > Ya these cyclic gc tests have somewhat overlapping behavior, since we made them by tracking down cycles basically one-by-one in various scenarios. I'd say it's not trio's job to check if exception groups create garbage (anymore). The question is whether a similar test exists in the suite for exception groups in cpython or the backport.
> 
> I'm not finding anything with a quick search in the backport @agronholm
> 
> I'm not familiar with the inner workings of gc and the implementation of multierror/exceptiongroup, but wouldn't the need for such tests depend on the manner of implementation? If the garbage was caused by hacky stuff that multierror needed to do in order to work, and cpython manages to sidestep that completely by implementing stuff """properly""", then there's no direct need for the tests on their side and/or they might have to look quite different.

I'd add such tests to the backport only as regression tests if such a problem is detected there.

jakkdl:
oh actually, I *am* getting cyclic garbage test fails in some cases. Will push a commit with updated comments

jakkdl:
> but `test_locals_destroyed_promptly_on_cancel` did start failing. 

Strangely enough I'm not managing to make this test fail now

richardsheridan:
Just to be completely clear, every spooky `del` in the codebase removes a cycle or shortens the lifetime of a strong reference in a way that can be statically reasoned about. _Some_ of these we were clever enough to also create regression tests.  I did my best at the time to cross reference these, by going through each `del` of the PR and commenting out the fix and seeing which test failed, or putting a TODO where nothing failed.

src/trio/_core/_run.py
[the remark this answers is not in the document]
jakkdl: Made it a somewhat generic "Exceptions from Trio nursery". Am very open to suggestions for what else could be added.

richardsheridan (approved):
I think all the cyclic garbage changes are making sense. Hanging on to `test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage` even though it maybe can't fail currently is fine IMO since we want to cover `outcome` and protect future refactoring.

src/trio/_core/_run.py
```diff
@@ -1128,10 +1128,12 @@ def aborted(raise_cancel: _core.RaiseCancelT) -> Abort:
...
                     return self._pending_excs[0]
-                return BaseExceptionGroup(
-                    "collapsible" if not self._strict_exception_groups else "TODO",
-                    self._pending_excs,
+                exception = BaseExceptionGroup(
+                    "Exceptions from Trio nursery", self._pending_excs
                 )
+                if not self._strict_exception_groups:
+                    exception.add_note("collapsible")
+                return exception
```
CoolCat467, line 1136: I thought `add_note` was 3.11+? Or does the `exceptiongroup` package have add_note implemented?

src/trio/_core/_tests/test_multierror.py
Zac-HD, line 1: Let's rename this something like `test_exceptiongroup_gc.py`?

src/trio/_core/_tests/test_run.py
```diff
@@ -2554,6 +2563,26 @@ async def test_nursery_strict_exception_groups() -> None:
     assert exc.value.exceptions[0].args == ("foo",)
 
 
+async def test_nursery_loose_exception_groups() -> None:
+    """Test that loose exception groups can be enabled on a per-nursery basis."""
```
Zac-HD, line 2567: This function tests that we get an `ExceptionGroup` when raising two errors in non-strict mode; can we add another which tests that raising a single error _doesn't_ result in a group?

src/trio/_core/_run.py
Zac-HD: Yep, [it's in the backport](https://github.com/agronholm/exceptiongroup/blob/c741f0b81c56d3b7a75b6096859e9e1463e02bbb/src/exceptiongroup/_exceptions.py#L87), so one way or another this is definitely valid.

Zac-HD (approved):
Two trivial nitpicks, but overall I think this is ready to merge.

I'd also be keen to ship a new version with this change ASAP, so that we can see it working in the wild _before_ shipping https://github.com/python-trio/trio/pull/2886.  That should reduce the risk of complications for everyone...

src/trio/_core/_run.py
TeamSpen210: Seems kinda odd to be adding a note like this, if the user's able to see it it'll be rather confusing.

CoolCat467: If it's in the backport everything should be fine, just wanted to be sure!

src/trio/_core/_tests/test_multierror.py
jakkdl: Ah yes, missed that one. It is a bit of a sad file, testing concat_tb and a gc test that may be redundant, but I suppose that name works. It'd probably be good to reorganize some tests, moving out a bunch from the 2.7k line long `test_run.py`, and putting these with them, but that's for another day.

src/trio/_core/_run.py
TeamSpen210: Another way to tag them would be to use a `WeakKeyDict` (or `WeakSet` for this case), so it's not even visible outside our code.

src/trio/__init__.py
```diff
@@ -120,26 +116,7 @@
...
-        version="0.22.0",
-        issue=2211,
-        instead=(
-            "BaseExceptionGroup (on Python 3.11 and later) or "
-            "exceptiongroup.BaseExceptionGroup (earlier versions)"
-        ),
-    ),
-    "NonBaseMultiError": _deprecate.DeprecatedAttribute(
-        value=_NonBaseMultiError,
-        version="0.22.0",
```
A5rocks, line 135: So I guess any release with these should be 0.24.0? Release preparation page says "two months or two releases, whichever is longer") and I assume it means minor releases.

src/trio/_core/_concat_tb.py
```diff
@@ -0,0 +1,128 @@
...
+# NB: 'tail' might be None, which requires some special handling in the ctypes
+# version.
+#
+# The complication here is that Python doesn't actually support copying or
+# modifying traceback objects, so we have to get creative...
+#
+# On CPython, we use ctypes. On PyPy, we use "transparent proxies".
+#
+# Jinja2 is a useful source of inspiration:
+#   https://github.com/pallets/jinja/blob/master/jinja2/debug.py
```
A5rocks, line 23:
```suggestion
#   https://github.com/pallets/jinja/blob/main/src/jinja2/debug.py
```

src/trio/_core/_concat_tb.py
```diff
@@ -0,0 +1,128 @@
...
+            return new_tb
+        finally:
+            # delete references from locals to avoid creating cycles
+            # see test_cancel_scope_exit_doesnt_create_cyclic_garbage
+            del new_tb, old_tb_frame
+
+else:
+    # http://doc.pypy.org/en/latest/objspace-proxies.html
+    def copy_tb(base_tb: TracebackType, tb_next: TracebackType | None) -> TracebackType:
+        # tputil.ProxyOperation is PyPy-only, but we run mypy on CPython
```
A5rocks, line 88: Unfortunately, as I found out recently, type checkers don't have a way of saying some code is PyPy / CPython specific *anyways*... So this wouldn't work regardless. I think this comment could probably be removed (or reworded) cause it insinuates that it is possible with the current state of typing.

src/trio/_core/_run.py
```diff
@@ -2615,18 +2625,17 @@ def unrolled_run(
                         # protocol of unwrapping whatever outcome gets sent in.
                         # Instead, we'll arrange to throw `exc` in directly,
                         # which works for at least asyncio and curio.
-                        # TODO: remove [unused-ignore] when Outcome is typed
-                        runner.reschedule(task, exc)  # type: ignore[arg-type, unused-ignore]
+                        runner.reschedule(task, exc)  # type: ignore[arg-type]
                         task._next_send_fn = task.coro.throw
                     # prevent long-lived reference
-                    # TODO: develop test for this deletion
+                    # There's no regression test checking the necessity of this
```
A5rocks, line 2631: For this and the following, the `TODO`s provide a way to find these via a simple search. Is there a reason not to include that?

src/trio/_core/_tests/test_exceptiongroup_gc.py
```diff
@@ -0,0 +1,101 @@
...
+    assert extract_tb(concat_tb(tb1, None)) == entries1
+    assert concat_tb(None, None) is None
+
+    # Make sure the original tracebacks didn't get mutated by mistake
+    assert extract_tb(get_tb(raiser1)) == entries1
+    assert extract_tb(get_tb(raiser2)) == entries2
+
+
+# Unclear if this can still fail, removing the `del` from _concat_tb.copy_tb does not seem
+# to trigger it (on a platform where the `del` is executed)
```
A5rocks, line 77: Huh, it might be a good idea to remove this test then and add a `TODO` to that `del` that it isn't tested for.

It seems like a good idea to try to remove unnecessary `del`s if things have improved!

A5rocks (commented):
Sorry about the wait; essentially just nitpicks (even about code you just copied over from elsewhere instead of writing :-) and I paid less attention to the tests.

src/trio/_core/_tests/test_run.py
```diff
@@ -2379,6 +2384,8 @@ async def test_cancel_scope_deadline_duplicates() -> None:
         await sleep(0.01)
 
 
+# I don't know if this one can fail anymore, the `del` next to the comment that used to
+# refer to this only seems to break test_cancel_scope_exit_doesnt_create_cyclic_garbage
```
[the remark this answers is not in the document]
jakkdl: @richardsheridan said in https://github.com/python-trio/trio/pull/2891#pullrequestreview-1785190814
> Hanging on to `test_simple_cancel_scope_usage_doesnt_create_cyclic_garbage` even though it maybe can't fail currently is fine IMO since we want to cover outcome and protect future refactoring.

I'd probably weakly be in favor of removing the test, but if not should probably update the comment to include sheridan's reasoning.

though `concat_tb` is only used for `collapse_exception_group`, which is only used for `strict_exception_groups=False`, which we might deprecate soonish. So not sure there's much future refactoring to consider.

src/trio/__init__.py
jakkdl: makes sense to do major (or whatever 0.X releases are called) release on both this and on #2886 since they can break stuff.

src/trio/_core/_tests/test_exceptiongroup_gc.py
jakkdl: see https://github.com/python-trio/trio/pull/2891#issuecomment-1831844141 on removing unnecessary `del`s.
But `copy_tb` is possibly an exception since it no longer uses `MultiError`.

src/trio/_core/_run.py
jakkdl: Ah of course. Not sure why I did this.

src/trio/_core/_run.py
jakkdl: I think it is good for it to be visible to end users though, if we plan on deprecating `strict_exception_groups=False`. With a custom exception it additionally makes it easy to add a deprecation warning (although it's maybe easier to just do that where they specify `strict_exception_groups=False`).
And e.g. if somebody is only catching single exceptions with loose groups, and then it crashes with a group, giving more context on that is great.

src/trio/_core/_concat_tb.py
jakkdl: reworded it

src/trio/__init__.py
agronholm: I still don't get why Trio doesn't have a 1.0.0 release yet if compatibility guarantees are this stringent anyway.

Zac-HD: I'd be inclined to go 1.0 when we've got this and strict-by-default in; I agree but it seems odd to make that release right before some substantial changes.

src/trio/_core/_tests/test_run.py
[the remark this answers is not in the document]
A5rocks: Yeah updating the comment would be nice

A5rocks (approved):

src/trio/_core/_tests/test_run.py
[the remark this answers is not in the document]
jakkdl: Ah oops, I mixed up the tests - the latter half of my previous comment pertains to `test_ExceptionGroup_catch_doesnt_create_cyclic_garbage`.
Updated the comment though.

PR #2892, opened 2023-11-29
Try to prevent autodeps from running in forked repos

A5rocks:


TeamSpen210 (approved):
Seems like there isn't a better way to do this, though forks will default to not running actions at all. In that case you don't get tests of course.

richardsheridan (approved):
It will help old forks from before not running actions was the default too.

A5rocks:
(well, as long as they merge in changes from upstream!)

PR #2898, opened 2023-12-06
Add RaisesGroup, a helper for catching ExceptionGroups in tests

jakkdl:
pytest does not seem overly excited to add my proposed extensions to `pytest.raises` in a timely manner, although I did get good feedback on the design - see https://github.com/pytest-dev/pytest/issues/11538

So I think we might want to settle with having it in trio (or possible in a separate sub-project? catching exceptiongroups is not exclusive to trio users) for some time, so I thought I'd make a proper implementation+tests and rewrite any current tests that can make use of this. This PR is largely to prep for #2886, together with #2891, and splits out some code from it (except this uses my `RaisesGroup` implementation from https://github.com/pytest-dev/pytest/pull/11671 instead of `ExpectedExceptionGroup` from https://github.com/pytest-dev/pytest/pull/11656).

TODO:
- [x] Documentation
- [x] Changelog
- [x] explicit tests for _ExceptionInfo
- [x] add docstrings (ETA 2023-12-30)
  - [x] also why doesn't pyright complain about them missing


jakkdl:
Failing tests are due to `RaisesGroup` inheriting from `BaseExceptionGroup` when type-checking, but not on runtime, which confuses the export tests. Might move away from that messiness and add a couple properties to `RaisesGroup` so it can be treated as an `ExceptionGroup` by end users using `exc_info`, or add exceptions to the export tests.

jakkdl:
```
   FAILED ../../../../hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/trio/_tests/test_exports.py::test_static_tool_sees_class_members[mypy-trio.testing] - AssertionError: assert not {'trio.testing.Matcher': {'extra': {'check', 'exception_type', 'match'}, 'missing': set()}}
```
wait, I think this is the same thing as a lot of the other weird exceptions in the test - `inspect.getmembers` and `dir()` only gives us static class members - not runtime properties/members/attributes/etc, but the way we look it up with mypy gives us the runtime members. I feel a bit silly for not realizing this earlier.

src/trio/_core/_tests/test_run.py
```diff
@@ -784,14 +773,22 @@ async def task2() -> None:
...
-            assert isinstance(exc, RuntimeError)
-            assert "closed before the task exited" in str(exc)
-            cancelled_in_context |= isinstance(exc.__context__, _core.Cancelled)
-        assert cancelled_in_context  # for the sleep_forever
+
+    def no_context(exc: RuntimeError) -> bool:
+        return exc.__context__ is None
+
+    msg = "closed before the task exited"
+    subexceptions = (
```
TeamSpen210, line 781: Might be simpler to just do: `group = RaisesGroup(`, avoiding the pack/unpack.

src/trio/_tests/test_testing_raisesgroup.py
```diff
@@ -0,0 +1,267 @@
...
+        trio.testing._raises_group._ExceptionInfo,
+    )
+    with trio.testing.RaisesGroup(ValueError) as excinfo:
+        raise ExceptionGroup("", (ValueError("hello"),))
+    assert excinfo.type is ExceptionGroup
+    assert excinfo.value.exceptions[0].args == ("hello",)
+    assert isinstance(excinfo.tb, TracebackType)
+
+
+if TYPE_CHECKING:
```
TeamSpen210, line 213: It'd be better to move these to the `type_tests` folder, which is also run under Pyright. Use bare `# type: ignore` so we check that it errors for both checkers.

TeamSpen210 (commented):
Looks good and useful. In regard to the typing, it does seem better to avoid inheriting from `BaseExceptionGroup`, if it doesn't actually use that behaviour.

jakkdl:
@TeamSpen210 
the reason for the inheritance from BaseExceptionGroup and lies in general is that `excinfo.value` *is*  an `ExceptionGroup`, not a `RaisesGroup`.
We're essentially saying we expect a structure with `RaisesGroup`s, but in fact we're not - we are expecting a structure of `ExceptionGroup`s, and `excinfo.value` will be a structure of `ExceptionGroup`s.

```python
with RaisesGroup(RaisesGroup(ValueError)) as excinfo:
   ...
print(type(excinfo.value))  # prints ExceptionGroup
```
so what I kind of want is to be able to search&replace instances of `RaisesGroup` with `ExceptionGroup` in the `TypeVar` that `RaisesGroup` accepts - in order to set the type on the `ExceptionInfo` correctly, but I have no clue how to do that.

TeamSpen210:
Ahh, that is tricky, and probably not possible without a plugin or something.

src/trio/_core/_tests/test_run.py
jakkdl: Ah oops, relic of old version.

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,232 @@
...
+        exc_info = cast(
+            "tuple[type[BaseExceptionGroup[E]], BaseExceptionGroup[E], TracebackType]",
+            (exc_type, exc_val, exc_tb),
+        )
+        self.excinfo.fill_unfilled(exc_info)
+        return True
+
+    def __repr__(self) -> str:
+        # TODO: [Base]ExceptionGroup
+        return f"ExceptionGroup{self.expected_exceptions}"
```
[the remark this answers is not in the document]
jakkdl: split it off to a whole new method

jakkdl:
huh, why is pyright not complaining about missing docstrings. Neither `Matcher` nor `RaisesGroup` has ~any

jakkdl:
> huh, why is `pyright --verifytypes` not complaining about missing docstrings. Neither `Matcher` nor `RaisesGroup` has ~any

opened an issue in pyright after debugging this. I'm pretty sure we'll get warnings for lots of random symbols if that gets fixed :sweat_smile:

jakkdl:
Quite confused by the pyright error, it appears to be a problem with pyright+exceptiongroup backport and not by `RaisesGroup`. Opened an issue https://github.com/microsoft/pyright/issues/6795 to see what that's about.

The `TracebackType` *shouldn't* be causing a problem in sphinx given that they explicitly have a workaround for it https://github.com/sphinx-doc/sphinx/pull/9015


The only remaining thing now is writing docstrings .... which I wrote on my desktop computer and then forgot to push before traveling to family over the holidays. Not gonna bother rewriting them, so they'll have to wait until the ~30th of december when I'm back.

jakkdl:
The error was in exceptiongroup, now pyright. So wrote a PR to fix it: https://github.com/agronholm/exceptiongroup/pull/101

jakkdl:
@Zac-HD would you mind reviewing this? It's the blocker for continuing work on #2886

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def check_inheritance_and_assignments() -> None:
+    # Check inheritance
+    _: BaseExceptionGroup[ValueError] = RaisesGroup(ValueError)
+    _ = RaisesGroup(RaisesGroup(ValueError))  # type: ignore
+
+    a: BaseExceptionGroup[BaseExceptionGroup[ValueError]]
+    a = RaisesGroup(RaisesGroup(ValueError))
+    a = BaseExceptionGroup("", (BaseExceptionGroup("", (ValueError(),)),))
```
Zac-HD, line 35:
```
  ++ pyright src/trio/_tests/type_tests
    /home/runner/work/trio/trio/src/trio/_tests/type_tests/raisesgroup.py:35:32 - error: Argument type is partially unknown
      Argument corresponds to parameter "args" in function "__init__"
      Argument type is "tuple[BaseExceptionGroup[Unknown]]" (reportUnknownArgumentType)
    /home/runner/work/trio/trio/src/trio/_tests/type_tests/raisesgroup.py:99:13 - error: Argument type is partially unknown
      Argument corresponds to parameter "args" in function "__init__"
      Argument type is "tuple[ExceptionGroup[Unknown]]" (reportUnknownArgumentType)
  2 errors, 0 warnings, 0 informations 
```

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def matches(
+        self,
+        exc_val: BaseException | None,
+    ) -> TypeGuard[BaseExceptionGroup[E]]:
+        if exc_val is None:
+            return False
+        # TODO: print/raise why a match fails, in a way that works properly in nested cases
+        # maybe have a list of strings logging failed matches, that __exit__ can
+        # recursively step through and print on a failing match.
```
Zac-HD, line 229: Do we need to do this before merging?

docs/source/reference-testing.rst
```diff
@@ -219,3 +219,16 @@ Testing checkpoints
...
+----------------------
+
+.. autoclass:: RaisesGroup
+   :members:
+
+.. autoclass:: Matcher
+   :members:
+
+.. autoclass:: trio.testing._raises_group._ExceptionInfo
+   :members:
```
Zac-HD, line 234: These classes should have docstrings aimed at explaining their use to end-users, if we're exposing them publicly.

I'd also prefer to mark this as a provisional interface, and note that we hope for it to be obsoleted by upstream work in pytest itself someday soon.  (note: carefully noncommital about what happens or when)

Zac-HD (approved):
Looks great - I've left a few nitpick comments, merge when you feel they've been addressed.

(I'm about to leave on a ten-day camping trip, and will probably be offline even if I have internet access, so don't wait for me!)

src/trio/_tests/type_tests/raisesgroup.py
jakkdl: bug in exceptiongroup: agronholm/exceptiongroup#101

src/trio/testing/_raises_group.py
jakkdl: I'd say it depends on expected usage - if it's a helper that lots of people start using, and/or if it takes a long time until it's supplied by pytest, then it's probably worth chugging through now.

But I don't know how much of it would be applicable to any eventual solution in pytest, or if anybody cares to use this, so I'm hesitant to go all-in on polishing this.

docs/source/reference-testing.rst
jakkdl: I wrote extensive docstrings... on a computer I'm currently 500km away from, and forgot to push. So I'll add them once back the 30th of december.
But good call to explicitly mark them as provisional

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,305 @@
...
+        assert (
+            self._excinfo is not None
+        ), ".tb can only be used after the context manager exits"
+        return self._excinfo[2]
+
+
+# this may bite users with type checkers not using pytest, but that should
+# be rare and give quite obvious errors in tests trying to do so.
+# Ignoring "incorrect import of pytest", it makes sense in this situation.
+if TYPE_CHECKING:
```
TeamSpen210, line 77: If the class above is sufficient, why use the pytest version at all?

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,305 @@
...
+            raise ValueError("You must specify at least one parameter to match on.")
+        if exception_type is not None and not issubclass(exception_type, BaseException):
+            raise ValueError(
+                f"exception_type {exception_type} must be a subclass of BaseException"
+            )
+        self.exception_type = exception_type
+        self.match = match
+        self.check = check
+
+    def matches(self, exception: E) -> TypeGuard[E]:
```
TeamSpen210, line 115: Shouldn't this be `exception: BaseException` to allow this to be used to check something that may not match?

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,305 @@
...
+@final
+class Matcher(Generic[E]):
+    def __init__(
+        self,
+        exception_type: type[E] | None = None,
+        match: str | Pattern[str] | None = None,
+        check: Callable[[E], bool] | None = None,
+    ):
+        if exception_type is None and match is None and check is None:
+            raise ValueError("You must specify at least one parameter to match on.")
```
TeamSpen210, line 106: You could type `init` here as having 3 overloads, each requiring one parameter. That way type checkers will disallow using it with no args. Also put match last, then type `self: Matcher[BaseException]` so E gets something valid in that case.

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,305 @@
...
+        if exception_type is not None and not issubclass(exception_type, BaseException):
+            raise ValueError(
+                f"exception_type {exception_type} must be a subclass of BaseException"
+            )
+        self.exception_type = exception_type
+        self.match = match
```
TeamSpen210, line 112: Minor and probably not necessary, but you could set `exception_type` to `BaseException` if None to avoid a condition below, and if `match` is a str do `re.compile()` so that's cached.

TeamSpen210 (approved):
Looks all good, though noticed a few things in `Matcher`.

src/trio/testing/_raises_group.py
CoolCat467: Type guards are special functions that take an input that could be a union of lots of things and return a boolean. If the value is True, the input is supposed to be confirmed to be of the type given as a parameter of the type guard. I think I agree with TeamSpen210 here, this type guard doesn't make a lot of sense just by looking at the snippet shown here, because it's as if we are just confirming that `E` is indeed `E`, but if it wasn't then type checkers should raise a `invalid parameter value` or something error when we try to call this function.

src/trio/testing/_raises_group.py
jakkdl: the first one ended up harder to read, and might interfere with future `print-why-it-didn't-match` logic - so skipping that one. Second one sounds good though :+1:

src/trio/testing/_raises_group.py
jakkdl: ah, that caught `check` being off as well - instead of `Callable[[E], bool]` it's now `Callable[[BaseException], TypeGuard[E]]`

src/trio/testing/_raises_group.py
jakkdl: [pytest.ExceptionInfo](https://docs.pytest.org/en/stable/reference/reference.html#exceptioninfo) has a couple helper methods I didn't bother reproducing - `match`, `getrepr`, `exconly`. And somebody writing tests might very well want to make use of those after catching an `ExceptionGroup`

src/trio/testing/_raises_group.py
TeamSpen210: I don't think you want to make it use typeguard, since it's unlikely people will be using that annotation. Not sure if checkers will infer that for lambdas also. Actually this probably needs an overload - if `exception_type` is supplied it's `Callable[[E], bool]` (so you can check that exception's attrs), but if not it would be `Callable[[BaseException], bool]`. 

Best would be to type the attribute as `Callable[[E], bool]`, do the overloads I mentioned below, and in the case that `exception_type` isn't supplied type `self: Match[BaseException]`. You'll need a cast if you leave `exception_type` able to be None I think.

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -92,6 +93,19 @@ def check_matcher_transparent() -> None:
...
+def check_matcher_tostring() -> None:
+    assert str(Matcher(ValueError)) == "Matcher(ValueError)"
+    assert str(Matcher(match="[a-z]")) == "Matcher(match='[a-z]')"
+    pattern_no_flags = re.compile("noflag", 0)
+    assert str(Matcher(match=pattern_no_flags)) == "Matcher(match='noflag')"
+    pattern_flags = re.compile("noflag", re.IGNORECASE)
+    assert str(Matcher(match=pattern_flags)) == f"Matcher(match={pattern_flags!r})"
+    assert (
+        str(Matcher(ValueError, match="re", check=bool))
+        == f"Matcher(ValueError, match='re', check={bool!r})"
```
CoolCat467, line 105: Is it intended that check is set to boolean constructor? I think you intended to set check to either True or False as with a pytest.parametrize. If so, don't use `bool` as a variable name, because it would be shadowing a python builtin.

CoolCat467 (changes requested):

src/trio/_tests/type_tests/raisesgroup.py
TeamSpen210: Yes, I was using bool as an arbitrary function. Maybe should have used something else.

jakkdl: I think it's fine

jakkdl: ...though why is this in `type_tests`

src/trio/testing/_raises_group.py
jakkdl: Forcing everybody to use `TypeGuard` would be amusing, but probably wiser not to.

src/trio/_tests/type_tests/raisesgroup.py
TeamSpen210: ... whoops. That explains the coverage.

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,341 @@
...
+    # needed for pyright, since BaseExceptionGroup.__new__ takes two arguments
+    if TYPE_CHECKING:
+
+        def __new__(cls, *args: object, **kwargs: object) -> RaisesGroup[E]:
+            ...
+
+    def __init__(
+        self,
+        exceptions: type[E] | Matcher[E] | E,
+        *args: type[E] | Matcher[E] | E,
```
A5rocks, line 206:
```suggestion
        exception: type[E] | Matcher[E] | E,
        *other_exceptions: type[E] | Matcher[E] | E,
```

(or something, idk)

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,341 @@
...
+                    continue
+                # Matcher __init__ assures it's a subclass of BaseException
+                self.is_baseexceptiongroup |= not issubclass(
+                    exc.exception_type, Exception
+                )
+            elif isinstance(exc, type) and issubclass(exc, BaseException):
+                self.is_baseexceptiongroup |= not issubclass(exc, Exception)
+            else:
+                raise ValueError(
+                    "Invalid argument {exc} must be exception type, Matcher, or"
```
A5rocks, line 239:
```suggestion
                    f"Invalid argument {exc!r} must be exception type, Matcher, or"
```

src/trio/testing/_raises_group.py
```diff
@@ -0,0 +1,341 @@
...
+# impossible. The type being RaisesGroup[RaisesGroup[ValueError]] is probably also fine,
+# as long as I add fake properties corresponding to the properties of exceptiongroup. But
+# I had trouble with it handling recursive cases properly.
+
+# Current solution settles on the above giving BaseExceptionGroup[RaisesGroup[ValueError]], and it not
+# being a type error to do `with RaisesGroup(ValueError()): ...` - but that will error on runtime.
+if TYPE_CHECKING:
+    SuperClass = BaseExceptionGroup
+else:
+    SuperClass = Generic
```
A5rocks, line 192: Why not just always inherit from `BaseExceptionGroup`?

And I feel like inheriting from that's super fishy given we don't actually have the information `BaseExceptionGroup` would have. But TBH I don't really see another solution. The typing on this is really icky.

A5rocks: I gave typing a go and ran across:
```py
import contextlib
from typing import ContextManager, Generic, TypeVar, overload

E = TypeVar("E", bound=BaseException)

class FakeExcInfo(Generic[E]):
    value: E

class RaisesGroup(ContextManager[FakeExcInfo[BaseExceptionGroup[E]]], Generic[E]):
    @overload
    def __init__(self, a: type[E], *b: type[E]):
        ...
    
    @overload
    def __init__(self: RaisesGroup[BaseExceptionGroup[E]], a: RaisesGroup[E], *b: RaisesGroup[E]):
        ...
    
    def __init__(self: object, *args: object):
        ...
    
    def __exit__(self, *args: object) -> bool:
        ...


with RaisesGroup(RaisesGroup(ValueError)) as excinfo:
    pass

reveal_type(excinfo.value)
```

it has a couple issues (not accepting both `RaisesGroup` and normal errors, plus whatever else kind of exceptions we allow) and also mypy doesn't like the overloads idk why but like. maybe that's cleaner?

A5rocks: For what it's worth, pyright seems to dislike that (probably a bug) but there's an easy workaround for it:
```py
from typing import ContextManager, Generic, TypeVar, overload

E = TypeVar("E", bound=BaseException)
E2 = TypeVar("E2", bound=BaseException)

class FakeExcInfo(Generic[E]):
    value: E

class RaisesGroup(ContextManager[FakeExcInfo[BaseExceptionGroup[E]]], Generic[E]):
    @overload
    def __init__(self, a: type[E], *b: type[E]):
        ...
    
    @overload
    def __init__(self: "RaisesGroup[BaseExceptionGroup[E2]]", a: "RaisesGroup[E2]", *b: "RaisesGroup[E2]"):
        ...
    
    def __init__(self: object, *args: object):
        ...
    
    def __exit__(self, *args: object) -> bool:
        ...


with RaisesGroup(RaisesGroup(ValueError)) as excinfo:
    pass

reveal_type(excinfo.value)
```

src/trio/testing/_raises_group.py
CoolCat467: +1

src/trio/testing/_raises_group.py
CoolCat467: +1, makes more sense

src/trio/testing/_raises_group.py
jakkdl: > Why not just always inherit from `BaseExceptionGroup`?

This gives errors due to the funky `__new__` stuff that `BaseExceptionGroup` does.
```
>       with RaisesGroup(ValueError, KeyError):
E       TypeError: BaseExceptionGroup.__new__() argument 1 must be str, not type
```

implementing a `__new__` that works like `object.__new__`:
```
>       obj = object.__new__(cls)
E       TypeError: object.__new__(RaisesGroup) is not safe, use BaseExceptionGroup.__new__()
```

and since we don't actually care for any of the functionality in `BaseExceptionGroup`, not inheriting at runtime seems clearly correct.

jakkdl: > And I feel like inheriting from that's super fishy given we don't actually have the information BaseExceptionGroup would have. But TBH I don't really see another solution. The typing on this is really icky.

Yeah this is clearly doing stuff that typing wasn't meant to handle.
The underlying problem is getting this to pass typing:
```python
with RaisesGroup(RaisesGroup(ValueError)) as excinfo:
  ...
myexc: BaseExceptionGroup = excinfo.value.exceptions[0]
```
and the only way I see to manage this is to lie to the type checker that it can treate `RaisesGroup` like a `BaseExceptionGroup`.

I'll update comments with reasons for some of the shady stuff.

src/trio/testing/_raises_group.py
jakkdl: realized that anybody using this without pytest would have zero types on exceptioninfo, so reversed it to always pretend we return `_ExceptionInfo` and added stubs for the helper methods. Maybe easier to simply always use `_ExceptionInfo`.

src/trio/testing/_raises_group.py
jakkdl: also see https://github.com/python-trio/trio/pull/2898#issuecomment-1848932771

docs/source/reference-testing.rst
jakkdl: Added them, and a note about them being provisional in the newsfragment.

src/trio/testing/_raises_group.py
A5rocks: What did you think of my proposed workaround using overloads btw?

Rest makes sense though.

CoolCat467 (approved):
As far as I can tell reading through this, everything makes sense.

src/trio/testing/_raises_group.py
A5rocks: Ugh I poked a bit more and the `E2` thing above that pyright accepts, mypy doesn't. I'm fine with this hacky inheritance thing I guess, hopefully inheritance doesn't fall under semver.

CoolCat467: I wouldn't think so for this particular case, because there would be no need to preform `isinstance` checks because you would not expect RaisesGroup to be in a list mixed with other objects.

jakkdl: > What did you think of my proposed workaround using overloads btw?
> 
> Rest makes sense though.

Oh I didn't trying to run it and failed to grok at a glance how it resolved the issue, sorry! I'll try and understand it and see if I can get it to work, otherwise I'll stick with the ugly inheritance thing.

jakkdl: The E2 version + pyright doesn't work for me with a third level of nesting
```
with RaisesGroup(RaisesGroup(RaisesGroup(ValueError))) as excinfo3:
    pass

reveal_type(excinfo3.value)
```
gives
```
information: Type of "excinfo3.value" is "BaseExceptionGroup[BaseExceptionGroup[Unknown]]"
```

jakkdl: but yeah let's stick with the hacky inheritance for now, and we can revisit this if a version of this is accepted for upstream pyright.

jakkdl:
Thanks everybody for the help, reviews, and feedback! :rocket: :rocket: :rocket:

PR #2900, opened 2023-12-09
open_tcp_listeners exception __cause__ is always an ExceptionGroup

jakkdl:
2 years ago, 4bd09bc346ee44ac3588ebcc2caaa3ab4ab46314 changed `open_tcp_listeners` to use `ExceptionGroup` instead of `MultiError` for the `__cause__`, which led to a behavior change in single-fail cases where `MultiError`s would collapse into the contained exception - something that `ExceptionGroup` doesn't do.

I only noticed this due to some confusion on my part when interpreting codecov in #2822.

I suspect we want to stick with the current behavior and just fix the test, but splitting it out so it can be discussed.

Current `open_tcp_listeners`
https://github.com/python-trio/trio/blob/60b9e1423b5f1a80798e214af96926a80c7d94d2/src/trio/_highlevel_open_tcp_listeners.py#L166-L173

and the test
https://github.com/python-trio/trio/blob/60b9e1423b5f1a80798e214af96926a80c7d94d2/src/trio/_tests/test_highlevel_open_tcp_listeners.py#L326-L335


CoolCat467:
From what I understand from what you have said in your comment above, the way exceptions are handled for `open_tcp_listeners` is different from how the rest of Trio handles exceptions, no? If this is indeed the case, I think it would be best to make this function handle exceptions like the rest of Trio (how Trio handles things when `strict_exception_groups` is True I mean), unless there would be a good reason not to.

jakkdl:
> From what I understand from what you have said in your comment above, the way exceptions are handled for `open_tcp_listeners` is different from how the rest of Trio handles exceptions, no? If this is indeed the case, I think it would be best to make this function handle exceptions like the rest of Trio (how Trio handles things when `strict_exception_groups` is True I mean), unless there would be a good reason not to.

Yeah that's correct. But the current behaviour having been in place for two years with nobody complaining means that it's not especially confusing or anything, and changing the behaviour might break code that's working perfectly fine at the moment - which I think is a very good reason not to change it.

CoolCat467:
> Also, re: "rest of trio" - I'm pretty sure it's only Nurseries that raise exceptiongroups and respect `strict_exception_groups`. I suppose we could add a parameter `strict_exception_groups` to `open_tcp_listeners` as well, defaulting to `True`, in which case the only change would be to users passing `strict_exception_groups=False` to `trio.run`.. but I still don't see much to gain from that.

I don't think we should have a parameter for `open_tcp_listeners`, I think it should be fine as is as you have said. Public statement of agreement.

Zac-HD (approved):
I agree with the comments above that keeping the current behavior and just fixing this test is the best way forward at the moment.

CoolCat467 (commented):
Looks good. Not sure what type of review this is, Github's UI seems to have a bug at the moment.

jakkdl:
> Looks good. Not sure what type of review this is, Github's UI seems to have a bug at the moment.

I suspect it's because the PR was already merged when you did the review.

PR #2901, opened 2023-12-09
add small tests and codecov skips to work on reaching 100%

jakkdl:
We're tantalizingly close, and I didn't find a setting for requiring the patch diff to be 100% without the project needing to be 100%, so might as well slowly work towards it. I noticed I've been authoring PRs with very minor misses - and easier to enforce 100% at time of authoring a PR than fix afterward.

We're at 52 missed lines on master/ currently


jakkdl:
PyPy 7.3.0 was released in december 2019, and I have no clue how long it's been since trio last ran 7.2.X in CI, so I think users still running 7.2.X have bigger problems than buggy asyncgens. But pypy versioning is weird so ... maybe warrants a release note?? @A5rocks

A5rocks:
Looks like pypy 7.2.0 is only Python 3.6 so it's not in our supported versions anyways

src/trio/_tests/test_highlevel_open_unix_stream.py
```diff
@@ -11,10 +11,11 @@
 
 assert not TYPE_CHECKING or sys.platform != "win32"
 
-if not hasattr(socket, "AF_UNIX"):
-    pytestmark = pytest.mark.skip("Needs unix socket support")
+has_unix = hasattr(socket, "AF_UNIX")
+reason_unix = "Needs unix socket support"
```
TeamSpen210, line 15: Instead of splitting these out, we could just store the `skipif` call here unconditionally, then reuse it lower down.

TeamSpen210 (approved):
Increasing test coverage is good indeed.

src/trio/_tests/test_highlevel_open_unix_stream.py
jakkdl: ah of course, thanks!

CoolCat467 (approved):
I think this looks great other than making sure we have a full discussion about the dtls `handle_client_hello_untrusted` thing

CoolCat467:
Pypy 3.10 segfaulted again: https://github.com/python-trio/trio/actions/runs/7200698782/job/19615211426

jakkdl:
> I think this looks great other than making sure we have a full discussion about the dtls `handle_client_hello_untrusted` thing

I think that requires more time/effort to fully delve into than it is worth, so I personally don't mind punting that to an unspecified time when somebody else works on the dtls stuff and stumbles upon it. Or if somebody reading this knows the dtls code well enough that they care to investigate and open a separate issue/PR.

PR #2907, opened 2023-12-19
Make ``start_soon`` a protocol.

Fuyukai:
Don't care for bikeshedding over the name of the protocol. The definition of ``start_soon`` is directly copied from ``Nursery``. 

I would've liked Nursery to explicitly implement TaskSpawner but it can't due to metaclass clashing.

### DIFF
--- a/src/trio/_abc.py
+++ b/src/trio/_abc.py
@@ -4,12 +4,12 @@ import socket
 from abc import ABCMeta, abstractmethod
 from typing import TYPE_CHECKING, Generic, Protocol, TypeVar
 
-from typing_extensions import TypeVarTuple
+if TYPE_CHECKING:
+    from typing_extensions import TypeVarTuple
+
+    PosArgT = TypeVarTuple("PosArgT")
 
 import trio
-
-PosArgT = TypeVarTuple("PosArgT")
-
 
 if TYPE_CHECKING:
     from collections.abc import Awaitable, Callable
--- a/src/trio/_abc.py
+++ b/src/trio/_abc.py
@@ -2,14 +2,20 @@ from __future__ import annotations
 
 import socket
 from abc import ABCMeta, abstractmethod
-from typing import TYPE_CHECKING, Generic, TypeVar
+from typing import TYPE_CHECKING, Generic, Protocol, TypeVar
+
+from typing_extensions import TypeVarTuple
 
 import trio
 
+PosArgT = TypeVarTuple("PosArgT")
+
+
 if TYPE_CHECKING:
+    from collections.abc import Awaitable, Callable
     from types import TracebackType
 
-    from typing_extensions import Self
+    from typing_extensions import Self, Unpack
 
     # both of these introduce circular imports if outside a TYPE_CHECKING guard
     from ._socket import SocketType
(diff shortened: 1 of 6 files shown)


src/trio/_abc.py
```diff
@@ -698,3 +704,53 @@ class Channel(SendChannel[T], ReceiveChannel[T]):
     `ReceiveChannel` interfaces, so you can both send and receive objects.
 
     """
+
+
+class TaskSpawner(Protocol):
```
graingert, line 709: SupportsStartSoon is the standard

graingert (changes requested):

src/trio/_abc.py
Fuyukai: SupportsStartsSoon is a stupid name

A5rocks: Yeah the `SupportsWhatever` names are mostly because ATM we use protocols for self-types, so we can be like `def f(self: Something[SupportsSomething])` and it reads nicer (we want the generic argument to support `.something`). IDK if that's a good naming idea for protocols that aren't used like that.

src/trio/_abc.py
```diff
@@ -698,3 +704,53 @@ class Channel(SendChannel[T], ReceiveChannel[T]):
...
+                  :meth:`start_soon` will try to make it one. A
+                  common use case is if you're wrapping a function
+                  before spawning a new task, you might pass the
+                  original function as the ``name=`` to make
+                  debugging easier.
+
+        Raises:
+            RuntimeError: If this task spawner is no longer open
+                          (i.e. its ``async with`` block has
+                          exited).
```
A5rocks, line 755: I don't think exceptions are really part of the interface exposed by protocols (unless we depend on this specific protocol raising this?)

src/trio/_abc.py
A5rocks: Alternative idea: `NurseryLike` (the "task spawner" in the `start_soon` docstring is a bit awkward, though it's workable so `TaskSpawner` is... fine.)

src/trio/_abc.py
```diff
@@ -698,3 +704,53 @@ class Channel(SendChannel[T], ReceiveChannel[T]):
...
+        *args: Unpack[PosArgT],
+        name: object = None,
+    ) -> None:
+        """Creates a child task, scheduling ``await async_fn(*args)``.
+
+        If you want to run a function and immediately wait for its result,
+        then you don't need a task spawner; just use ``await async_fn(*args)``.
+        If you want to wait for the task to initialize itself before
+        continuing, see :meth:`start`, the other fundamental method for
+        creating concurrent tasks in Trio.
```
A5rocks, line 724: This bit doesn't make much sense since the protocol doesn't expose `start`.

src/trio/_abc.py
Fuyukai: I just copy-pasted the docstring. I think it's worth having it mentioned in some form though, because there's no way of making a task spawner that isn't eventually closed.

Fuyukai:
I'm not changing the name.

src/trio/_abc.py
oremanj: `SupportsClose` is the first example of a protocol in the PEP that defined protocols. https://peps.python.org/pep-0544/#defining-a-protocol

The `typing` module defines a bunch more using the same naming scheme: https://docs.python.org/3/library/typing.html#protocols

I think being consistent/recognizable is valuable here. I think `SupportsStartSoon` and `NurseryLike` are both fine. I disprefer `TaskSpawner` (a nursery is arguably the *target* of task spawning, not the agent of it) but you've made your position abundantly clear so <shrug>.

A5rocks (changes requested):
I really don't care enough about the name for that to block my approval. But:

1) does this really need to be exported? I imagine any normal usage of this will just have the concrete other nursery as a type hint:

```py
class BetterNursery:
    ...

async def whatever(n: BetterNursery) -> None:
  await something_that_takes_task_spawner(n)
```

This doesn't work for libraries but (1) libraries may have other requirements on the task spawner they are given (e.g. supports `await n.start(task)`) and (2) it's trivial to redefine.

If this isn't exported, I doubt anyone cares about name. And docstring doesn't need to exist!

2) if this *really* has to be exported, can you rewrite the docstring a bit? Thanks!

A5rocks:
Hi, just a bump! re: my last comment I still can't think of normal usage that requires this to be exported.

Fuyukai:
> does this really need to be exported? I imagine any normal usage of this will just have the concrete other nursery as a type hint:

I already changed type hints for listeners to accept this, so yes, it does.

I'll get around to fixing the docstring at some point soon.

Fuyukai:
Yeah, I don't really care to do this anymore. I'll just close it.

PR #2909, opened 2023-12-20
Sphinx cleanups, bump version requirements, typecheck docs source files

jakkdl:
I was struggling with sphinx in #2898 and noticed that sphinx was pinned to an ancient version, and unpinning & updating it seemed to fix a couple of them. My struggles also made me understand sphinx innards better than I had before, so could fix a bunch of references at the same time.

* add missing type annotations to docs/source/*.py
* specify files to check with mypy in pyproject.toml (docs/source/typevars.py was typed already, but not checked in CI)
* change sphinx version limits to >5.3
* clean up & fix a bunch of references in sphinx
* add comment explaining workarounds

let's see if RTD manages this as well

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -60,6 +60,7 @@ if "READTHEDOCS" in os.environ:
 # 4. add_intersphinx manually modifies the intersphinx mappings after
 #    objects.inv has been parsed, to resolve bugs and version differences
 #    that causes some objects to be looked up incorrectly.
+# 5. docs/source/typevars.py handles redirecting `typing_extensions` objects to `typing`, and linking `TypeVar`s to `typing.TypeVar` instead of sphinx wanting to link them to their individual definitions.
 # It's possible there's better methods for resolving some of the above
 # problems, but this works for now:tm:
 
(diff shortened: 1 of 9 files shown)


docs/source/conf.py
```diff
@@ -41,6 +49,20 @@
             check=True,
         )
 
+# Sphinx is very finicky, and somewhat buggy, so we have several different
+# methods to help it resolve links.
```
TeamSpen210, line 53: Might want to mention the `typevar` module also, which redirects `typing_extensions` → `typing`, and all typevars to `typing.TypeVar` etc.

docs/source/conf.py
```diff
@@ -16,8 +16,16 @@
...
+from __future__ import annotations
+
+import collections.abc
 import os
 import sys
+from typing import TYPE_CHECKING, cast
+
+if TYPE_CHECKING:
+    from sphinx.application import Sphinx
+    from sphinx.util.typing import Inventory
```
TeamSpen210, line 28: Not sure if the guard is necessary. If this is running, Sphinx is installed?

Fuyukai: Ruff does this automatically for any objects that are only found in type hints for some reason

TeamSpen210: Ahh right.

jakkdl: Hah, I did it manually out of habit. I like denoting typing-only imports

TeamSpen210 (approved):

Issue #2908, opened 2023-12-19
Cython has defeated our Cython workaround


njsmith:
When we spawn a task, we inject a magic local variable into the top stack frame's local namespace, as part of our tricky control-C magic:

https://github.com/python-trio/trio/blob/d4ce2f9e42c386f1b64969f0495fafe2a2183b22/src/trio/_core/_run.py#L1753

However, this doesn't work for `async def` functions defined in C or Cython, because they don't have a Python frame. So we a simple workaround, we detect this case and inject a no-op Python frame where we can insert our magic local (cf #550):

https://github.com/python-trio/trio/blob/d4ce2f9e42c386f1b64969f0495fafe2a2183b22/src/trio/_core/_run.py#L1747-L1750

However, the bug came back at some point with newer versions of Cython:

```
             |   File "/root/.pyenv/versions/3.11.6/lib/python3.11/site-packages/trio/_core/_run.py", line 1599, in spawn_impl
             |     coro.cr_frame.f_locals.setdefault(LOCALS_KEY_KI_PROTECTION_ENABLED, system_task)
             |     ^^^^^^^^^^^^^^^^^^^^^^
             | AttributeError: 'NoneType' object has no attribute 'f_locals'
```

From the error message, it appears that Cython has added a `cr_frame` attribute to its fake coroutine objects... just it's `None`. So it's still not really there, but it's enough to sneak past our `hasattr(coro, "cr_frame")` test.

I think the fix here is just to replace the `if not hasattr(coro, "cr_frame"):` with `if getattr(coro, "cr_frame", None) is None:`. But it would be good to also run a test for this in CI. We do have a test for the interpreter's built-in C coroutines (`test_async_function_implemented_in_C`) and I originally assumed that would be enough, but obviously it wasn't, so I guess we need to do the legwork to install cython in CI and build a tiny Cython module.

Edit by CoolCat467:
Things to do:
- [x] Fix existing issue (#2911 is handling)
- [ ] Add CI runner for cython so this sort of thing doesn't happen again in the future

TeamSpen210:
This might be fixed in newer versions potentially. Found this issue: https://github.com/cython/cython/issues/2306. As of 3.0, `cr_frame` should actually produce a valid frame object. Locals will be null but it looks like accessing it will make a dict for us.

[...]

PR #2911, opened 2023-12-21
Cython Workaround

CoolCat467:
This pull request partially completes changes that were noted in #2908.

Either added on to this pull request or in another pull request, we still need to make sure to test that cython will continue working and not find these sorts of issues after they are already a problem.

Excerpt from original comment in #2908:
>But it would be good to also run a test for this in CI. We do have a test for the interpreter's built-in C coroutines (`test_async_function_implemented_in_C`) and I [njsmith] originally assumed that would be enough, but obviously it wasn't, so I guess we need to do the legwork to install cython in CI and build a tiny Cython module.

### DIFF
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -1758,7 +1758,7 @@ class Runner:
             except AttributeError:
                 name = repr(name)
 
-        if not hasattr(coro, "cr_frame"):
+        if getattr(coro, "cr_frame", None) is None:
             # This async function is implemented in C or Cython
             async def python_wrapper(orig_coro: Awaitable[RetT]) -> RetT:
                 return await orig_coro


A5rocks (approved):
Looks good, I think actually doing the CI work can be done later (just repurpose the already made issue).

Zac-HD:
I'd be OK with shipping this as a fix for the current problem, but I'm also keen to get some regression tests into our CI so we can tell that it's effective and notice if it breaks later.  Maybe @jakkdl can help out?

jakkdl:
> I'd be OK with shipping this as a fix for the current problem, but I'm also keen to get some regression tests into our CI so we can tell that it's effective and notice if it breaks later. Maybe @jakkdl can help out?

I'll take a look at it, seems quite doable. But agree on merging this and getting CI setup in a later PR.

jakkdl (approved):

CoolCat467:
Ok, with two approvals I am going to merge this.

PR #2917, opened 2024-01-04
fix test regexes for musl libc

tornaria:
Error messages are slightly different when using musl libc.

The new regexes fix that.

### DIFF
--- a/src/trio/_tests/test_fakenet.py
+++ b/src/trio/_tests/test_fakenet.py
@@ -10,7 +10,7 @@ from trio.testing._fake_net import FakeNet
 
 # ENOTCONN gives different messages on different platforms
 if sys.platform == "linux":
-    ENOTCONN_MSG = r"^\[Errno 107\] Transport endpoint is not connected$"
+    ENOTCONN_MSG = r"^\[Errno 107\] (Transport endpoint is|Socket) not connected$"
 elif sys.platform == "darwin":
     ENOTCONN_MSG = r"^\[Errno 57\] Socket is not connected$"
 else:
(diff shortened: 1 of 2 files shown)


A5rocks:
Where did you run into this? (So we can add that to CI!)

Fuyukai:
WHY is this using regexes. OSError literally has an ``errno`` property. Against this with vehemence. Somebody fix the tests properly.

CoolCat467:
> WHY is this using regexes. OSError literally has an `errno` property. Against this with vehemence. Somebody fix the tests properly.

It's using regexes because `pytest.expect`'s `match` argument is a regular expression, not literal string compare. This was a part of my work enabling ruff's `flake8-pytest-style` rule in #2822 so we can be even more precise with what errors we expect to happen in the test suite.

CoolCat467 (approved):
Thank you for fixing this, we should really add CI testing for this platform!

jakkdl:
> WHY is this using regexes. OSError literally has an `errno` property. Against this with vehemence. Somebody fix the tests properly.

Disapprove of the tone.

And scrolling through the file it looks like pretty much all tests does check the `errno` property. Checking the string in addition is maybe overkill, but I like the check in most all other instances - and it's doing what it's supposed to by letting us know about a new system that behaves differently.

jakkdl:
> Where did you run into this? (So we can add that to CI!)

some quick searching for "musl libc":
https://peps.python.org/pep-0656/
https://github.com/FRRouting/frr/issues/8442

So seems like the easiest way is to add Alpine Linux to the CI. Perhaps something like https://github.com/marketplace/actions/setup-alpine-linux-environment

tornaria:
> Where did you run into this? (So we can add that to CI!)

Upgrading `python3-trio` package for void linux (which supports both glibc and musl) in https://github.com/void-linux/void-packages/pull/47946.

Since void linux is rolling release, alpine is probably easiest to implement.

A5rocks:
Yeah I was confused cause we *should* be running Alpine but then I remembered our sr.ht CI stuff stopped working...

I agree that it's somewhat strange that we check the strings via regex rather than check errnos but this is also tests and tbh being more strict is fine. (I note that the 2nd and 3rd error doesn't have one specific errno... and it's easier to look at a test and see "Socket not connected" than see "107" and have to do another lookup).

Issue #2903, opened 2023-12-12
Test invocation fails with Python 3.13.0~alpha2


befeleme:
In Fedora Linux we try to build the Python libraries with Python 3.13.0 alpha versions. 
Trio fails to build in our environment, more concretely the test invocation fail with a traceback below.

```
+ /usr/bin/pytest --pyargs trio -p trio._tests.pytest_plugin --verbose --skip-optional-imports
Traceback (most recent call last):
  File "/usr/bin/pytest", line 8, in <module>
    sys.exit(console_main())
             ^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/_pytest/config/__init__.py", line 192, in console_main
    code = main()
           ^^^^^^
  File "/usr/lib/python3.13/site-packages/_pytest/config/__init__.py", line 150, in main
    config = _prepareconfig(args, plugins)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/_pytest/config/__init__.py", line 331, in _prepareconfig
    config = pluginmanager.hook.pytest_cmdline_parse(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/pluggy/_hooks.py", line 493, in __call__
    return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/pluggy/_manager.py", line 115, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/pluggy/_callers.py", line 130, in _multicall
    teardown[0].send(outcome)
  File "/usr/lib/python3.13/site-packages/_pytest/helpconfig.py", line 104, in pytest_cmdline_parse
    config: Config = outcome.get_result()
                     ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/pluggy/_result.py", line 114, in get_result
    raise exc.with_traceback(exc.__traceback__)
[...]

PR #2918, opened 2024-01-06
Make trio run on Python 3.13 (probably)

A5rocks:
I'm pretty sure I tested this 3 weeks ago locally. Unfortunately it's not possible to run 3.13 in CI ATM because there's no compatible `cryptography` release.

This fixes https://github.com/python-trio/trio/issues/2903. (hopefully)

cc @befeleme can you make sure this is fine?


src/trio/_path.py
```diff
@@ -199,6 +199,7 @@ def generate_iter(cls, attrs: dict[str, object]) -> None:
                 setattr(cls, attr_name, wrapper)
 
 
+# TODO: `pathmod` points to `os.path`... is that what we want?
```
A5rocks, line 202: I left this because ATM `trio.Path` has a `pathmod` pointing to a synchronous path module. Which is weird but also we don't have an equivalent module IIRC.

TeamSpen210: This might get changed, see https://github.com/python/cpython/pull/113893.

CoolCat467: https://github.com/python/cpython/pull/113893 was merged

A5rocks: Looks like the TODO is irrelevant because AFAICT the `pathmod` or `flavour` or `flavor` or whatever it will be only does operations like joining paths. Seemingly nothing that interacts with FS?

CoolCat467: Yea, as far as I remember looking at it a bit ago I think they've just make an abstract class for the path interface, at least in that particular PR

TeamSpen210 (approved):
This seems like the right approach for `pathmod`, but we might want to consider expanding `trio.Path` a bit in anticipation of 3.13. Since `pathlib` paths can now be subclassed (to allow say paths inside a zip, or on a remote server), we might want to make it possible to wrap subclasses of path using this.

illia-v:
FYI, this patch let me fix urllib3's test with CPython 3.13 in https://github.com/urllib3/urllib3/pull/3156. Thanks!

CoolCat467 (approved):
These changes look good to me!

befeleme:
I'm sorry, this has slipped my radar. I can run the tests with Python 3.13 now. (8 of them fail, but that's another tale.)

PR #2920, opened 2024-01-10
Bump version to 0.24.0

A5rocks:
I finally realized `UnboundedQueue` is just sitting there deprecated. But it's used by the Windows IO thing so removal looked more complicated than just "remove any references".


jakkdl:
> I finally realized `UnboundedQueue` is just sitting there deprecated. But it's used by the Windows IO thing so removal looked more complicated than just "remove any references".

Open an issue about it?

Issue #2922, opened 2024-01-10
Remove `UnboundedQueue`


A5rocks:
This has been deprecated for a long time and I don't think that's been reversed. Just, it's used in the rest of trio so it would be a good idea to replace those.

richardsheridan:
xref: https://github.com/python-trio/trio/issues/40#issuecomment-1819829567

arthur-tacca:
I stumbled across this while researching #3500. Trio currently has [the rule that]:

> there’s a very strict internal layering: the `trio._core` package is a fully self-contained implementation of the core scheduling/cancellation/IO handling logic, and then the other trio.* modules are implemented in terms of the API it exposes.

#2925 violates by importing memory channels from outside of it. Actually even the existing `ParkingLot` in `_core` comes with [a comment that says](https://github.com/python-trio/trio/blob/50b9825f7ef6630c2326001cc9323c714c05355d/src/trio/_core/_parking_lot.py#L24-L35) says its use in `UnboundedQueue` is the main reason its in `_core`. (It also mentions its theoretical role in ensuring fairness which is coupled to the task scheduling algorithm, but I suspect that wouldn't have been enough to justify its placement in `_core` by itself). I think, if anything, we should be aiming to replace the two usages of `UnboundedQueue` with a simpler alternative. Some options:

* Keep `UnboundedQueue` but make it simpler (only one allowed recipient) (NJS suggested this [at the end of this comment](https://github.com/python-trio/trio/issues/578#issuecomment-408760107)) so that at least the dependence on `ParkingLot` can be removed (for clarity: NJS never mentioned that bit).
* Allow the caller to pass a memory channel (NJS mentioned this [here](https://github.com/python-trio/trio/pull/937#issuecomment-465397988) as a way to allow different kevents to be sent to one channel) in which case `_core` only needs a protocol for it rather than importing the whole thing. `monitor_kevent()` now just waits forever since it no longer needs to return an iterable, and this gives an avenue for solving what to do with an exception raised by sending to a closed (or full) channel: it can be captured in `process_events()` and re-raised out of `monitor_kevent`.
[...]

PR #2925, opened 2024-01-16
Remove deprecated unbounded queue

A5rocks:
Very much based off of https://github.com/python-trio/trio/pull/937; fixes https://github.com/python-trio/trio/issues/2922; should be in the next 0.x.0 release so don't merge this yet!


src/trio/_core/_parking_lot.py
```diff
@@ -21,12 +21,11 @@
...
-# UnboundedQueue, and UnboundedQueue is used for a number of things in the
-# core. And second, it's responsible for providing fairness to all of our
-# high-level synchronization primitives (locks, queues, etc.). For now with
-# our FIFO scheduler this is relatively trivial (it's just a FIFO waitqueue),
-# but in the future we ever start support task priorities or fair scheduling
+# This is in the core because it's responsible for providing fairness to all
+# of our high-level synchronization primitives (locks, queues, etc.). For now
+# with our FIFO scheduler this is relatively trivial (it's just a FIFO
+# waitqueue), but in the future we ever start support task priorities or fair
+# scheduling
```
CoolCat467, line 28: The grammar here is a bit odd. I would change it to something like
"but in the future if we ever support task priorities or
fair scheduling it should be fairly simple."

src/trio/_core/_generated_io_kqueue.py
```diff
@@ -43,7 +43,7 @@ def current_kqueue() -> select.kqueue:
 
 def monitor_kevent(
     ident: int, filter: int
-) -> ContextManager[_core.UnboundedQueue[select.kevent]]:
+) -> ContextManager[MemoryReceiveChannel[select.kevent]]:
```
A5rocks, line 46: Huh actually I'm just realizing this is potentially a compat break. I can work around this using a kwarg but I need to see if this currently warns or not...

Zac-HD: Private subclass which adds the unbounded queue methods but warns if you call them?

src/trio/_core/_io_kqueue.py
```diff
@@ -98,7 +99,7 @@ def process_events(self, events: EventResult) -> None:
             if isinstance(receiver, _core.Task):
                 _core.reschedule(receiver, outcome.Value(event))
             else:
-                receiver.put_nowait(event)  # TODO: test this line
+                receiver.send_nowait(event)  # TODO: test this line
```
graingert, line 102: it's possible for this to fail if someone calls:

```python
with trio.lowlevel.monitor_kevent() as q:
    q.close()
```

we should either catch the exception here or yield a clone

I'd add it to this PR for you but I just cannot work out how to cover this line - it should be easy but I'm new to kqueue so might be missing something

graingert: something like this:

```patch
from 9ddb5ecc47dda3a7502851967bb67247a99aefe5 mon sep 17 00:00:00 2001
from: thomas grainger <tagrain@gmail.com>
date: tue, 24 dec 2024 08:24:41 +0000
subject: [patch] make sure closing monitor queues is a noop

---
 src/trio/_core/_io_kqueue.py  | 12 ++++++------
 src/trio/_core/_io_windows.py | 13 +++++++------
 2 files changed, 13 insertions(+), 12 deletions(-)

diff --git a/src/trio/_core/_io_kqueue.py b/src/trio/_core/_io_kqueue.py
index fed4da83..08b5d966 100644
--- a/src/trio/_core/_io_kqueue.py
+++ b/src/trio/_core/_io_kqueue.py
@@ -137,12 +137,12 @@ class kqueueiomanager:
                 "attempt to register multiple listeners for same ident/filter pair",
             )
         send, recv = open_memory_channel[select.kevent](math.inf)
-        self._registered[key] = send
-        try:
-            yield recv
-        finally:
-            send.close()
-            self._registered.pop(key, none)
+        with send, recv:
+            self._registered[key] = send
+            try:
+                yield recv.clone()
+            finally:
+                self._registered.pop(key, none)
 
     @_public
     async def wait_kevent(
diff --git a/src/trio/_core/_io_windows.py b/src/trio/_core/_io_windows.py
index 4676e6c5..824e91f5 100644
--- a/src/trio/_core/_io_windows.py
+++ b/src/trio/_core/_io_windows.py
@@ -1040,9 +1040,10 @@ class windowsiomanager:
 
         key = next(self._completion_key_counter)
         send, recv = open_memory_channel[object](math.inf)
-        self._completion_key_queues[key] = send
-        try:
-            yield (key, recv)
-        finally:
-            send.close()
-            del self._completion_key_queues[key]
+        with send, recv:
+            self._completion_key_queues[key] = send
+            try:
+                yield (key, recv.clone())
+            finally:
+                send.close()
+                del self._completion_key_queues[key]
-- 
2.43.0
```

A5rocks:
@graingert the reason this stalled (and the reason I think your latest changes don't quite work) is because we need to provide a layer that has both memory channel and unbounded queue apis. Alternatively, a boolean kwarg to switch over.

Otherwise, we're breaking backwards compatibility.

graingert:
These apis were sketches barely documented and not suitable for production use, they also warned if they were used. I think it's acceptable to break them in this case.

Issue #2848, opened 2023-11-01
AttributeError: module 'select' has no attribute 'epoll'


kushalmraut:
**version : trio==0.22.2**

```
[2023-10-10 12:06:56 +0000] [4254] [INFO] Worker exiting (pid: 4254)
[2023-10-10 12:07:00 +0000] [4232] [INFO] Shutting down: Master
[2023-10-10 12:07:00 +0000] [4232] [INFO] Reason: Worker failed to boot.
[2023-10-10 12:07:04 +0000] [4264] [DEBUG] Current configuration:
 config: None
 bind: ['unix:/run/anahit.sock']
 backlog: 2048
 workers: 1
 worker_class: gevent
 threads: 1
 worker_connections: 1000
 max_requests: 0
 max_requests_jitter: 0
 timeout: 120
 graceful_timeout: 30
 keepalive: 2
 limit_request_line: 4094
 limit_request_fields: 100
 limit_request_field_size: 8190
 reload: False
 reload_engine: auto
 reload_extra_files: []
 spew: False
 check_config: False
 preload_app: False
 sendfile: None
 reuse_port: False
 chdir: /home/anahit_web/anahit_website/app
 daemon: False
 raw_env: []
 pidfile: None
 worker_tmp_dir: None
 user: 1000
 group: 33
 umask: 0
 initgroups: False
 tmp_upload_dir: None
 secure_scheme_headers: {'X-FORWARDED-PROTOCOL': 'ssl', 'X-FORWARDED-PROTO': 'https', 'X-FORWARDED-SSL': 'on'}
 forwarded_allow_ips: ['127.0.0.1']
 accesslog: /home/anahit_web/logs/access.log
 disable_redirect_access_to_syslog: False
 access_log_format: %(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"
 errorlog: /home/anahit_web/logs/anahit_gunicorn.log
 loglevel: debug
 capture_output: True
 logger_class: gunicorn.glogging.Logger
 logconfig: None
 logconfig_dict: {}
 syslog_addr: udp://localhost:514
 syslog: False
 syslog_prefix: None
 syslog_facility: user
 enable_stdio_inheritance: True
 statsd_host: None
 dogstatsd_tags: 
 statsd_prefix: 
 proc_name: None
 default_proc_name: anahit.wsgi:application
 pythonpath: None
 paste: None
[...]

PR #2928, opened 2024-01-17
Just delay `select.epoll` to fix import after gevent monkey patching

A5rocks:
This seems to work:
```
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110] on linux
>>> import gevent.monkey
>>> gevent.monkey.patch_all()
True
>>> import trio
>>> exit()
```

Fixes: https://github.com/python-trio/trio/issues/2848

... While trio still doesn't work under gevent, I think not erroring when importing is a good property to have.


A5rocks:
cc @jeongilpark could you see if this works for you? I'm not sure if you're actually ending up using trio. (if you do, maybe we should do a better solution, like... I don't know, maybe gevent stuffs old attributes somewhere? or... idk this level of trio very well.)

jakkdl:
> @jakkdl @CoolCat467 - is this issue and the one specifically listed below resolved now?
> 
> [gevent/gevent#2008](https://github.com/gevent/gevent/issues/2008)

no? I don't see any reason why they've would've been fixed with no changes to any code.

CoolCat467 (approved):
Seems simple enough, while obviously not fixing the underlying issue I think not crashing on import is important.

TeamSpen210:
Might be a good idea to add a comment as to why we're using a lambda here. Otherwise someone coming along in the future might "optimise" that out, especially since I know PyLint and therefore Ruff [has a rule](https://docs.astral.sh/ruff/rules/unnecessary-lambda/) which would flag this.

PR #2930, opened 2024-01-22
Enable ruff `flake8-builtins` rule

CoolCat467:
This pull request enables ruff's `flake8-builtins` rule, sorts the enabled rules and ignored rules sections, and fixes or ignores a lot of newly generated "argument or variable is shadowing a python builtin" errors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,8 +94,10 @@ select = [
     "A",     # flake8-builtins
     "ASYNC", # flake8-async
     "B",     # flake8-bugbear
+    "C4",    # flake8-comprehensions
     "E",     # Error
     "F",     # pyflakes
+    "FA",    # flake8-future-annotations
     "I",     # isort
     "PT",    # flake8-pytest-style
     "PYI",   # flake8-pyi
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,6 +130,7 @@ extend-exclude = [
 'src/trio/socket.py' = ['F401']
 'src/trio/testing/__init__.py' = ['F401']
 'src/trio/_core/_generated_io_kqueue.py' = ['A002']
+'src/trio/_core/_io_kqueue.py' = ['A002']
 
 [tool.ruff.isort]
 combine-as-imports = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,27 +91,27 @@ allowed-confusables = ["–"]
 src = ["src/trio", "notes-to-self"]
 
 select = [
-  "RUF",  # Ruff-specific rules
-  "F",  # pyflakes
-  "E",  # Error
-  "W",  # Warning
-  "I",  # isort
-  "UP",  # pyupgrade
-  "B",  # flake8-bugbear
-  "YTT",  # flake8-2020
-  "ASYNC",  # flake8-async
-  "PYI",  # flake8-pyi
-  "SIM",  # flake8-simplify
-  "TCH",  # flake8-type-checking
-  "PT",  # flake8-pytest-style
+    "RUF",   # Ruff-specific rules
+    "F",     # pyflakes
+    "E",     # Error
+    "W",     # Warning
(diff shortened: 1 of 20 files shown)


Fuyukai:
Why is this three separate PRs that will all conflict due to all three reformattting the pyproject.toml?

CoolCat467:
Because they are separate rules, and if they conflict I will deal with that.

CoolCat467:
The segfault in pypy-3.9 happened again: https://github.com/python-trio/trio/actions/runs/7607032357/job/20713702030#step:5:1186
For future reference, where should these be recorded?

A5rocks:
Gah, well see the segfaults don't exactly provide much information... But I agree that it's annoying. At this point ig we might as well make an issue here (on trio) to track things cause upstream pytest hasn't been able to solve this yet.

---

Also, I don't really see the use of this rule. We have type checkers; I don't see the issue with this shadowing. Maybe others see issue though?

CoolCat467:
In all reality we might not need `flake8-builtins`, but I wanted to throw it out there because it's a part of best practice not to shadow builtins.

jakkdl:
I'm in favor of not shadowing builtins in general, type checking might catch most errors it might lead to (only if the object is different though!) - but I can definitely see it causing confusion/lead to bugs/etc.

pyproject.toml
```diff
@@ -128,6 +129,8 @@ extend-exclude = [
 'src/trio/lowlevel.py' = ['F401']
 'src/trio/socket.py' = ['F401']
 'src/trio/testing/__init__.py' = ['F401']
+'src/trio/_core/_generated_io_kqueue.py' = ['A002']
+'src/trio/_core/_io_kqueue.py' = ['A002']
```
jakkdl, line 133: probably good to have comments here (for the F401 too) saying what the rule is and why it's ignored.

src/trio/_socket.py
```diff
@@ -356,14 +356,16 @@ def socket(
...
+            family, type, proto, fileno
+        )
     stdlib_socket = _stdlib_socket.socket(family, type, proto, fileno)
     return from_stdlib_socket(stdlib_socket)
 
 
 def _sniff_sockopts_for_fileno(
     family: AddressFamily | int,
-    type: SocketKind | int,
+    type_: SocketKind | int,  # name shadowing builtin
```
jakkdl, line 368: Don't think there's a need for a long-lasting comment here.

src/trio/_core/_io_kqueue.py
```diff
@@ -118,7 +118,9 @@ def current_kqueue(self) -> select.kqueue:
     @contextmanager
     @_public
     def monitor_kevent(
-        self, ident: int, filter: int
+        self,
+        ident: int,
+        filter: int,  # Name shadows builtin
```
jakkdl, line 123: this comment gives zero info unless you know that it refers to A002, and that it's ignored in this file. So probably remove it, and/or add a longer comment saying that A002 is ignored in the file and why. But not sure either is warranted

jakkdl (changes requested):
actually, after reading through all the diffs I think we should ignore `A002` - but we could keep `A001`. We're never going to change the argument names for external-facing functions, and I'm not sure it's worth the need for lots of ugly `noqa`s to be extra safe on internal functions.

src/trio/_core/_io_kqueue.py
CoolCat467: Changed in 67116bc

src/trio/_socket.py
CoolCat467: Changed in 67116bc

pyproject.toml
CoolCat467: Changed in 67116bc

jakkdl (approved):
Looks great!

TeamSpen210 (approved):

PR #2932, opened 2024-01-22
Enable `flake8-future-annotations`

CoolCat467:
This pull request enables ruff's `flake8-future-annotations` rule, building off of the formatting changes done in https://github.com/python-trio/trio/pull/2930.

### DIFF
--- a/src/trio/socket.py
+++ b/src/trio/socket.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 # here.
 # We still have some underscore names though but only a few.
 # Uses `from x import y as y` for compatibility with `pyright --verifytypes` (#2625)
+#
 # Dynamically re-export whatever constants this particular Python happens to
 # have:
 import socket as _stdlib_socket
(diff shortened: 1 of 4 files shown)


src/trio/socket.py
```diff
@@ -1,14 +1,13 @@
...
 # annoying to carefully always use underscored names for module-level
 # temporaries, imports, etc. when implementing the module. So we put the
 # implementation in an underscored module, and then re-export the public parts
 # here.
 # We still have some underscore names though but only a few.
-
-
 # Uses `from x import y as y` for compatibility with `pyright --verifytypes` (#2625)
-
 # Dynamically re-export whatever constants this particular Python happens to
```
A5rocks, line 11: Is there a reason for removing whitespace? The comments are distinct so I think it was better with that (or maybe replacing the empty lines with `#`?)

CoolCat467: No, I can fix that.

CoolCat467: Should be fixed with [46b0c34](https://github.com/python-trio/trio/pull/2932/commits/46b0c343e675e50d22a2141c3997b655d3021c61)

jakkdl (approved):

PR #2933, opened 2024-01-22
Add Alpine to CI

A5rocks:
Ref: https://github.com/python-trio/trio/issues/2885

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,8 +189,10 @@ jobs:
         uses: actions/checkout@v3
       - name: Install necessary packages
         run: apk update && apk add python3-dev bash nodejs musl-dev
+      - name: Enter virtual environment
+        run: python -m venv .venv
       - name: Run tests
-        run: ./ci.sh
+        run: source .venv/bin/activate && ./ci.sh
       - if: always()
         uses: codecov/codecov-action@v3
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,7 +188,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Install necessary packages
-        run: apk update && apk add bash
+        run: apk update && apk add bash nodejs
+      - run: pip install pyright && pyright --version
       - name: Run tests
         run: ./ci.sh
       - if: always()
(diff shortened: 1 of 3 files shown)


CoolCat467 (approved):

src/trio/_tests/test_socket.py
```diff
@@ -629,6 +629,9 @@ async def res(
                 # Linux
                 if hasattr(tsocket, "EAI_ADDRFAMILY"):
                     expected_errnos.add(tsocket.EAI_ADDRFAMILY)
+                # musl libc
+                if hasattr(tsocket, "EAI_NODATA"):
+                    expected_errnos.add(tsocket.EAI_NODATA)
```
jakkdl, line 634: codecov says this comparison never fails, so we should either add it unconditionally (but some quick searching turns up https://github.com/python/cpython/issues/69325) or add `# pragma: no branch`

.github/workflows/ci.yml
```diff
@@ -179,6 +179,28 @@ jobs:
...
+  # run CI on a musl linux
+  Alpine:
+    name: "Alpine"
+    runs-on: ubuntu-latest
+    container: alpine
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install necessary packages
+        run: apk update && apk add python3-dev bash nodejs musl-dev
```
jakkdl, line 191: maybe good to add a comment saying why these packages in particular are needed, or if it was suggested by someplace online a link to that.

jakkdl (approved):
Skimmed the test logs, and IPv6 tests are being skipped
```
  SKIPPED [1] ../.venv/lib/python3.11/site-packages/trio/_tests/test_dtls.py:87: need IPv6
```
probably not a big deal, but would be nice if it was possible to enable ipv6.


So other than the minor stuff, looks great!

src/trio/_tests/test_socket.py
A5rocks: I think it's fine to have tests fail then add in the branch at that time

A5rocks:
Doesn't look like we can do IPv6. Making the socket works, but binding fails. It seems that this is a docker thing and that we'd need to configure the docker daemon? I found this stackoverflow post but it has no answers: https://stackoverflow.com/questions/75205327/configuring-docker-daemon-for-github-actions

jakkdl:
> Doesn't look like we can do IPv6. Making the socket works, but binding fails. It seems that this is a docker thing and that we'd need to configure the docker daemon? I found this stackoverflow post but it has no answers: https://stackoverflow.com/questions/75205327/configuring-docker-daemon-for-github-actions
> 

Seems very doable in theory then, but definitely not worth the effort.

PR #2935, opened 2024-01-24
change all instances of 'import typing as ...' to 'from typing import ...'

jakkdl:
Noticed this when reviewing #2932. For context, `from typing import` is what's used everywhere else, with 110 matches in the repository when `grep`ing.

also one instance of `import typing_extensions`, where statistics are similar.


jakkdl:
right, ofc there are export/visibility reasons for the leading underscore... why didn't we simply settle on using `__all__` like everybody else 🙃

TeamSpen210 (approved):

CoolCat467 (approved):

PR #2937, opened 2024-01-25
Add trio.testing.wait_all_threads_completed

VincentVanlaer:
Alternative to #2880

### DIFF
--- a/src/trio/_threads.py
+++ b/src/trio/_threads.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Generic, TypeVar, overload
 
 import attr
 import outcome
+from attrs import define
 from sniffio import current_async_library_cvar
 
 import trio
@@ -17,16 +18,17 @@ import trio
 from ._core import (
     RunVar,
     TrioToken,
+    checkpoint,
     disable_ki_protection,
     enable_ki_protection,
     start_thread_soon,
 )
 from ._deprecate import warn_deprecated
-from ._sync import CapacityLimiter
+from ._sync import CapacityLimiter, Event
 from ._util import coroutine_or_error
 
 if TYPE_CHECKING:
-    from collections.abc import Awaitable, Callable
+    from collections.abc import Awaitable, Callable, Generator
 
     from trio._core._traps import RaiseCancelT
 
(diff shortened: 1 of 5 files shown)


CoolCat467:
This looks pretty good, but one part I worry about is the incrementing and decrementing thread count system. As it stands I do think it would work, but I would feel a lot better if it were using a context manager instead.
You could pretty easily make it a context manager with the following:
```python3
from collections.abc import Generator

@contextmanager
def handling_thread() -> Generator[None, None, None]:
    _increment_active_threads()
    yield
    _decrement_active_threads()

with handling_thread():
    # do all the things
    ....
```

Context managers are guaranteed to run their close methods, similar to a try finally block, but way nicer to work with.

src/trio/_threads.py
```diff
@@ -52,6 +54,79 @@ class _ParentTaskData(threading.local):
 _thread_counter = count()
 
 
+class _ActiveThreadCount:
+    count: int
+    event: Event
+
+    def __init__(self) -> None:
+        self.count = 0
+        self.event = Event()
```
jakkdl, line 63: @CoolCat467 recommended NamedTuple to not have to have an `__init__`, though code in this repo generally uses attr to handle default values in "dataclasses". See e.g. `EpollWaiters` in `src/trio/_core_io_epoll.py`.

src/trio/_threads.py
```diff
@@ -397,12 +473,14 @@ def abort(raise_cancel: RaiseCancelT) -> trio.lowlevel.Abort:
...
         ] = await trio.lowlevel.wait_task_rescheduled(abort)
         if isinstance(msg_from_thread, outcome.Outcome):
+            _decrement_active_threads()
             return msg_from_thread.unwrap()
         elif isinstance(msg_from_thread, Run):
             await msg_from_thread.run()
         elif isinstance(msg_from_thread, RunSync):
             msg_from_thread.run_sync()
         else:  # pragma: no cover, internal debugging guard TODO: use assert_never
+            _decrement_active_threads()
```
jakkdl, line 483: After looking at this for a bit I agree with @CoolCat467 that the increment/decrement does seem unreliable. if `await msg_from_thread.run()` or `msg_from_thread.run_sync` raises exceptions it won't decrement. So either a contextmanager, or wrapping all the involved lines in `to_thread_run_sync` in a try/finally, would probably be good.

src/trio/_threads.py
```diff
@@ -52,6 +54,79 @@ class _ParentTaskData(threading.local):
...
+    try:
+        active_threads_local = _active_threads_local.get()
+        active_threads_local.count -= 1
+        if active_threads_local.count == 0:
+            active_threads_local.event.set()
+            active_threads_local.event = Event()
+    except LookupError as e:
+        raise TrioInternalError(
+            "Tried to decrement active threads while _active_threads_local is unset"
+        ) from e
```
jakkdl, line 89: this seems fine to have a `pragma: no cover` on - but think it'd be fairly straightforward to manipulate `_active_threads_local` to force an error.

jakkdl (changes requested):
Also I'd like you to add tests so you get 100% patch coverage.

Don't have any particular insight or opinion on if this version could be an "attractive nuisance" or whatever. But I do like that it's exposed in testing.

VincentVanlaer:
I believe I have addressed all comments. The codecov fail seems to be spurious (the diff is in lines with only comments in a complete unrelated file if I understand it correctly).

CoolCat467:
For future reference, force pushing makes it far more difficult to compare the previous version to new changes.

src/trio/_threads.py
```diff
@@ -52,6 +54,72 @@ class _ParentTaskData(threading.local):
...
+
+    try:
+        active_threads_local = _active_threads_local.get()
+    except LookupError:
+        # If there would have been active threads, the
+        # _active_threads_local would have been set
+        return
+
+    while active_threads_local.count != 0:
+        await active_threads_local.event.wait()
```
CoolCat467, line 109: Just wanting to be sure, I know event wait call is doing what we want and will block until thread count is zero and it's set, but is there any way with race conditions that `active_threads_local.event` is reset to a new lock before the `wait` call sees that? Would it be beneficial to move the event resetting to before the `active_threads_local.count += 1` line?

CoolCat467 (commented):
Looks good but I have a question about a part of the reset logic

jakkdl:
> I believe I have addressed all comments. The codecov fail seems to be spurious (the diff is in lines with only comments in a complete unrelated file if I understand it correctly).

yeah the codecov can be weird sometimes, if there seems to be duplicates of codecov in the run list some may be stale and not getting updated / checking vs the wrong set of files / something like that.

src/trio/_threads.py
jakkdl: ooh. Yeah shouldn't it maybe be something like this:
```python
    try:
        active_threads_local = _active_threads_local.get()
    except LookupError:
        active_threads_local = _ActiveThreadCount(1, Event())
        _active_threads_local.set(active_threads_local)
    else:
        active_threads_local.count += 1
```

CoolCat467: Yea, probably something like that and adding a part in the else block there where it resets the event object if the event it has has already fired.

richardsheridan: I don't understand where a race could arise here. This is all happening in the main thread so everything between awaits is effectively atomic.

I have similar code in a different project and it hasn't given me trouble... Although speaking of that, it would be good to `assert active_threads_local.count >= 0` after decrementing it, because if a negative number sneaks in it'd be better to fail fast.

A5rocks: Yeah I don't think we have to care about thread safety here. Rather, the only times `_active_threads_local` can be changed (that we worry about) is at an `await` point.

After all `to_thread_run_sync` can be only run from the main thread (that has your `trio` event loop).

A5rocks (approved):
I didn't look at this too hard but the strategy makes sense to me.

jakkdl (approved):

CoolCat467 (approved):
If we don't have to worry about thread safety I think this is great!

PR #2938, opened 2024-01-29
Replace usage of strict_exception_groups=False in Nursery.start

jakkdl:
In preparation for eventual deprecation of `strict_exception_groups=False` (https://github.com/jakkdl/trio/tree/deprecate_exceptiongroups_false) we probably shouldn't rely on it ourselves in the internal code.

#2611 - the original issue where errors raised before `task_status.started()` got double-wrapped
#1599 - for why the old nursery *can* have multiple tasks.
#2844 - the regression error that occured after my initial fix for #2611 in #2826

Fairly straightforward implementation, other than perhaps exactly what to raise when multiple exceptions do occur. The main code that tests this is `test_trio_run_strict_before_started`, I added the new test after it only to trigger the case where multiple exceptions do pop up in the old nursery.

### DIFF
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -1249,9 +1249,10 @@ class Nursery(metaclass=NoPublicConstructor):
             except BaseExceptionGroup as exc:
                 if len(exc.exceptions) == 1:
                     raise exc.exceptions[0] from None
-                # TODO: give a deprecationwarning instead?
                 raise TrioInternalError(
-                    "internal nursery should not have multiple tasks"
+                    "Internal nursery should not have multiple tasks. This can be "
+                    'caused by the user managing to access the "old" nursery in '
+                    "`task_status` and spawning tasks in it."
                 ) from exc
 
             # If we get here, then the child either got reparented or exited
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -1231,19 +1231,29 @@ class Nursery(metaclass=NoPublicConstructor):
             raise RuntimeError("Nursery is closed to new arrivals")
         try:
             self._pending_starts += 1
-            # `strict_exception_groups=False` prevents the implementation-detail
-            # nursery from inheriting `strict_exception_groups=True` from the
-            # `run` option, which would cause it to wrap a pre-started()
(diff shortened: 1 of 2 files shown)


src/trio/_core/_run.py
```diff
@@ -1231,19 +1231,29 @@ async def async_fn(arg1, arg2, *, task_status=trio.TASK_STATUS_IGNORED):
...
+                    task._eventual_parent_nursery = self
+                    # Wait for either TaskStatus.started or an exception to
+                    # cancel this nursery:
+            except BaseExceptionGroup as exc:
+                if len(exc.exceptions) == 1:
+                    raise exc.exceptions[0] from None
+                # TODO: give a deprecationwarning instead?
+                raise TrioInternalError(
+                    "internal nursery should not have multiple tasks"
+                ) from exc
```
A5rocks, line 1255: I liked how one of the linked PRs proposed warning in `.start_soon` for that internal nursery (re: todo)

src/trio/_core/_run.py
```diff
@@ -1231,19 +1231,29 @@ async def async_fn(arg1, arg2, *, task_status=trio.TASK_STATUS_IGNORED):
...
+                    task = GLOBAL_RUN_CONTEXT.runner.spawn_impl(
+                        thunk, args, old_nursery, name
+                    )
+                    task._eventual_parent_nursery = self
+                    # Wait for either TaskStatus.started or an exception to
+                    # cancel this nursery:
+            except BaseExceptionGroup as exc:
+                if len(exc.exceptions) == 1:
+                    raise exc.exceptions[0] from None
+                # TODO: give a deprecationwarning instead?
```
Zac-HD, line 1252: I think making this an internal error immediately is fine - it's _possible_ to do this, but it's never been something we documented or that we think is reasonable to have done for some reason.

Zac-HD (approved):

jakkdl:
sounds like we're settling on internalerror, and leaving any eventual warning for #1600 or similar.
removed the comment and expanded the error message to make it clearer for the user that it might be their fault.

Zac-HD:
Very nice!  Merging now 😁

PR #2939, opened 2024-01-29
Fix finding pthread_*name_np on vanilla musl libc

mgorny:
Fix the `pthread_getname_np` and `pthread_setname_np` search logic to support vanilla versions of musl and CPython, e.g. as used on Gentoo musl systems.  On such systems, there is no "libpthread.so" (there is only a static library) and the relevant functions are found in "libc.so".  Additionally, `ctypes.util.find_library("c")` does not work because of an old unsolved bug in CPython (linked in the code).

To resolve the problem, add a fallback to trying `libc.so` if no pthread library can be found.  This roughly covers three possibilities:

- a "typical" system with `libpthread.so` will find that library and use it

- a musl system will fall back to `libc.so`, load that library and find pthread functions there

- any other system will try to load `libc.so`, and fail

The code in `get_os_thread_name_func()` remains fully relaxed, allowing either CDLL construction (i.e. finding the library) to fail, or the library not to contain `pthread_setname_np`.

The code in `test_threads.py` was made more relaxed — rather than skipping if `libpthread.so` does not exist, it tries to load `libc.so` as a fallback, and skips if that fails.

Originally reported as https://bugs.gentoo.org/923257.

### DIFF
--- a/src/trio/_tests/test_threads.py
+++ b/src/trio/_tests/test_threads.py
@@ -237,9 +237,17 @@ def _get_thread_name(ident: int | None = None) -> str | None:
 
     libpthread_path = ctypes.util.find_library("pthread")
     if not libpthread_path:
-        print(f"no pthread on {sys.platform})")
+        # musl includes pthread functions directly in libc.so
+        # (but note that find_library("c") does not work on musl,
+        #  see: https://github.com/python/cpython/issues/65821)
+        # so try that library instead
+        # if it doesn't exist, CDLL() will fail below
+        libpthread_path = "libc.so"
+    try:
+        libpthread = ctypes.CDLL(libpthread_path)
+    except Exception:
+        print(f"no pthread on {sys.platform}")
         return None
-    libpthread = ctypes.CDLL(libpthread_path)
 
     pthread_getname_np = getattr(libpthread, "pthread_getname_np", None)
 
--- a/newsfragments/2939.bugfix.rst
+++ b/newsfragments/2939.bugfix.rst
@@ -1,0 +1,1 @@
+The pthread functions are now correctly found on systems using vanilla versions of musl libc.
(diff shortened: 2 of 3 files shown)


A5rocks:
I ran into this for https://github.com/python-trio/trio/pull/2933 and ended up just making that require `musl-dev` instead of figuring this out. Thanks for putting in the work!

src/trio/_tests/test_threads.py
```diff
@@ -237,9 +237,17 @@ def _get_thread_name(ident: int | None = None) -> str | None:
...
+        # musl includes pthread functions directly in libc.so
+        # (but note that find_library("c") does not work on musl,
+        #  see: https://github.com/python/cpython/issues/65821)
+        # so try that library instead
+        # if it doesn't exist, CDLL() will fail below
+        libpthread_path = "libc.so"
+    try:
+        libpthread = ctypes.CDLL(libpthread_path)
+    except Exception:
+        print(f"(no pthread on {sys.platform})")
```
A5rocks, line 249: I think this looks better (nitpick)
```suggestion
        print(f"no pthread on {sys.platform}")
```

A5rocks (approved):

src/trio/_tests/test_threads.py
mgorny: Sure. I've noticed the missing opening parenthesis, so I've just fixed it but I'm fine either way.

mgorny:
Fixed and added a newsfragment.

CoolCat467 (approved):
Seeing this and your description, this makes sense and I think it looks good!

jakkdl (approved):

mgorny:
Thanks!

jakkdl:
Thank you for the contribution!
Can also see that it works in CI over in #2933
https://github.com/python-trio/trio/actions/runs/7712303460/job/21019556518?pr=2933

Issue #2929, opened 2024-01-22
Ensure that our docs present `strict_exception_groups=True` as the natural default


Zac-HD:
Follow-up for https://github.com/python-trio/trio/pull/2886#discussion_r1444122554:

> I think that we'll eventually want to rewrite this whole section from the new strict-by-default perspective, under the heading "Deprecated: non-strict ExceptionGroups" - to explain that it only exists for backwards-compatibility, will be removed in future, and that we recommend against it for all new code.

We should also skim through other mentions of `ExceptionGroup` to ensure that our messaging is consistent across all the docs.

Zac-HD:
Relatedly, we should document any design patterns we like or dislike for interfaces with an "inner nursery", c.f. https://github.com/python-trio/trio-websocket/issues/132

The basic problems are that (1) forcing users to handle errors which _may or may not_ be an `ExceptionGroup` has horrible ergonomics, and that (2) usually one or the other is more common, so it's really easy to end up with latent bugs from only handling the common kind.  That's why we made `strict_exception_groups=True` the default!

Unfortunately, this _also_ means that using a nursery internally means that error handling doesn't quite follow [the black box rule](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/#nurseries-a-structured-replacement-for-go-statements) - [you can still catch an ungrouped exception with `except*:`](https://peps.python.org/pep-0654/#naked-exceptions), but most code has good reason to prefer `except:`.

**Example patterns:** 
1. Pass through the `ExceptionGroup`, letting users handle it.
    - Pros: this is the default behavior, so it's trivial to implement
    - Cons: forcing users to deal with maybe-a-group is unergonomic and constrains internal changes
2. `raise SomeInternalError(...) from group`
    - Pros: reasonably simple; easy to implement with a context manager
    - Cons: users still have to deal with maybe getting a `SomeInternalError` instead of a group, so it's not clear that we've actually gained much - `except Exception:` will catch an `ExceptionGroup` directly, after all.
3. Raise the contained exception, if there's only one, else raise a group
    - Pros: status quo ante
    - Cons: see above for why we deprecated this approach
4. Raise the contained exception, if there's only one, else ???
    - it's not clear what the "???" would be, if not isomorphic to (2)
5. Avoid using an inner nursery
    - Pros: avoids the problem entirely
    - Cons: ...by avoiding concurrency.  Good advice when it works but can't be our only solution.
[...]

PR #2941, opened 2024-01-31
add deprecation warning for specifying strict_exception_groups=False

jakkdl:
Wasn't totally sure which issue to refer to, ended up going with #2929

the removed `with` block in `test_subprocess` was added to try and track down some weirdness in https://github.com/python-trio/trio/pull/2886#discussion_r1449035885 - but I don't think it served much of a purpose in the end, so opted now to remove it instead of wrapping it in a `pytest.warns`.

### DIFF
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -998,13 +998,13 @@ def open_nursery(
           and ultimately removed in a future version of Trio.
 
     """
-    # only warn if explicitly set to False, not if we get it from the global context.
-    if strict_exception_groups is False:
+    # only warn if explicitly set to falsy, not if we get it from the global context.
+    if strict_exception_groups is not None and not strict_exception_groups:
         warn_deprecated(
             "open_nursery(strict_exception_groups=False)",
             version="0.24.1",
             issue=2929,
-            instead="Use the default value of True and rewrite exception handlers to handle ExceptionGroups",
+            instead="the default value of True and rewrite exception handlers to handle ExceptionGroups",
         )
 
     if strict_exception_groups is None:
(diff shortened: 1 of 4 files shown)


src/trio/_core/_run.py
```diff
@@ -2244,6 +2254,13 @@ def run(
           propagates it.
 
     """
+    if strict_exception_groups is False:
+        warn_deprecated(
+            "trio.run(..., strict_exception_groups=False)",
+            version="0.24.1",
+            issue=2929,
+            instead="the default value of True and rewrite exception handlers to handle ExceptionGroups",
```
Zac-HD, line 2262: Other messages begin "Use", otherwise identical.  copy/paste error?

Zac-HD (approved):
Looks good - thanks @jakkdl!

CoolCat467:
Making sure, nothing can go wrong with the literal `is False` compare? No one could set strict exception groups to a value that could be interpreted as the `False` branch?

jakkdl:
> Making sure, nothing can go wrong with the literal `is False` compare? No one could set strict exception groups to a value that could be interpreted as the `False` branch?

Right, if somebody is supplying a falsy value of an incorrect type they're not gonna get a warning. So I suppose I could do
```python
if not strict_exception_groups and strict_exception_groups is not None:
    ...
```
Don't think we need to be that defensive, but I guess it also doesn't hurt.

src/trio/_core/_run.py
jakkdl: looking at `warn_deprecated`
```python
        msg += f"; use {_stringify(instead)} instead"
```
I might have noticed that supplies a `use` and only removed it in one of the three places. Fixed

PR #2942, opened 2024-01-31
Add Cython to CI

jakkdl:
after some wrangling I've got Cython going in the CI. Though so much of CI is ~magic I probably have something configured badly.

* Commented out all non-cython stuff for sake of testing, obviously will return those before merging.
* Reverted the change of #2911 for the sake of reproducing the failure
* Cython==0.29.23, python 3.8 & 3.10 fails because of reverting #2911 
* The other five failing tests are because old Cython versions does not work with 3.11/3.12

I vote for testing cython<3 and cython>=3 on latest python, but if @njsmith or @Zac-HD or somebody else has any reasons to test older versions, or on non-ubuntu, or on other python versions (and pypy?). Could even go so far as to add it to `ci.sh` and test it on everything, which I think would add something like 4-7 seconds.

`test_cython.pyx` could of course also be much more extensive, though grepping for `Cython` I'm only getting a hit for the one comment in `src/trio/_core_run.py` so doesn't seem like there's a lot of code that has been specially adapted for Cython. Maybe catch some exceptions? pytest does not seem to be a thing for cython, so running the full test suite or whatever I think is out of the question.


jakkdl:
pypy+cython seems to sort of be a thing: https://docs.cython.org/en/latest/src/userguide/pypy.html
but I'm getting a fancy error locally, so if that's something we want to support we probably want it in the matrix (after fixing the error):
```
python -c 'import test_cython'
hello...
 world !
  |   File "<string>", line 1, in <module>
  |   File "test_cython.pyx", line 20, in init test_cython
  |     main()
  |   File "test_cython.pyx", line 18, in test_cython.main
  |     trio.run(trio_main)
  |   File "/home/h/Git/trio/cython_ci/.tox/pypy3-cython/lib/pypy3.10/site-packages/trio/_core/_run.py", line 2274, in run
  |     raise runner.main_task_outcome.error
  |   File "test_cython.pyx", line 12, in trio_main
  |     async with trio.open_nursery() as nursery:
  |   File "/home/h/Git/trio/cython_ci/.tox/pypy3-cython/lib/pypy3.10/site-packages/trio/_core/_run.py", line 958, in __aexit__
  |     raise combined_error_from_nursery
  | exceptiongroup.ExceptionGroup: Exceptions from Trio nursery (1 sub-exception)
    |   File "test_cython.pyx", line 13, in test_cython.trio_main
    |     nursery.start_soon(foo)
    | AttributeError: 'NoneType' object has no attribute 'start_soon'
```

.github/workflows/ci.yml
```diff
@@ -11,209 +11,239 @@ concurrency:
...
-          flags: Alpine,3.12
+            python-version: '${{ matrix.python }}'
+            cache: pip
+      - name: install cython
+        run: |
+          # setuptools is needed to get distutils on 3.12
+          python -m pip install --upgrade pip setuptools
+          python -m pip install "cython${{ matrix.cython}}"
+      - name: install trio
+          # use -e?
```
A5rocks, line 229: I think editable installs isn't the normal case and those are weird:tm:

So I think it's better to test non-editable installs of trio

tests/cython/test_cython.pyx
```diff
@@ -0,0 +1,20 @@
...
+
+    async with trio.open_nursery() as nursery:
+        nursery.start_soon(foo)
+        nursery.start_soon(foo)
+        nursery.start_soon(foo)
+
+def main() -> None:
+    trio.run(trio_main)
+
+main()
```
A5rocks, line 20:
```suggestion
if __name__ == "__main__":
    main()
```

maybe? not sure why there's a `main` function to begin with

.github/workflows/ci.yml
```diff
@@ -11,209 +11,239 @@ concurrency:
...
-      ${{
-        (
-          endsWith(matrix.python, '-dev')
-          || endsWith(matrix.python, '-nightly')
-        )
-        && true
-        || false
-      }}
+        cython: ['==0.29.23', '==0.29.26', '==0.29.30', '<3', '>=3']
+        python: ['3.8', '3.10', '3.11', '3']
```
A5rocks, line 214: Bit confused about versions here:
- weren't you just talking about testing `<3` and `>=3`?
- Why are we testing `'3'`? (and how about 3.12?)
- What about 3.9? (wondering if there's a reason other than matrix size -- I don't really care if 3.9's tested or not)

.github/workflows/ci.yml
```diff
@@ -11,209 +11,239 @@ concurrency:
...
+      - name: install cython
+        run: |
+          # setuptools is needed to get distutils on 3.12
+          python -m pip install --upgrade pip setuptools
+          python -m pip install "cython${{ matrix.cython}}"
+      - name: install trio
+          # use -e?
+        run: python -m pip install .
+      - name: compile pyx file
+        run: cythonize -i tests/cython/test_cython.pyx
```
A5rocks, line 232: It should be fine to run the compiled file across multiple python versions -- maybe that'll make for a faster CI? (probably just a future optimization tbh)

.github/workflows/ci.yml
jakkdl: sorry, should've clarified with a comment. These are just a semi-random set of versions to test/show different behaviour in the CI, not the versions I'm actually suggesting.

tests/cython/test_cython.pyx
jakkdl: lol, good question.

.github/workflows/ci.yml
jakkdl: I was randomly using Black's CI as a reference: https://github.com/search?q=repo%3Apsf%2Fblack+%22python+-m+pip+install+-e%22&type=code
I have no clue why they're doing `-e` tho

.github/workflows/ci.yml
jakkdl: looking at the tests, that step is taking ~ 1-3s. So negligible time save, and all but two of the tests are failing on the compile step (though due to super old cython + new python version, which maybe isn't terribly relevant for us).

.github/workflows/ci.yml
jakkdl: specifying `'3'` could make sense if we only bother trying on two python versions, oldest and newest. So like
```
cython: ['<3', '>=3']
python: ['3.8', '3']
```
but given that all other CI steps need to get updated on highest-version bumps that's probably silly optimization.

.github/workflows/ci.yml
jakkdl: What would likely save a significant amount of time is only having separate machines for the python versions, and testing all relevant cython versions on the same machine. Since the job itself is so short the gain from parallelization is vastly outweighed by having to wait for acquisition of the machine and setup/teardown.

.github/workflows/ci.yml
A5rocks: Yeah I'd personally rather things be similar (e.g. either all use `'3.12'` or all use `'3'`)

.github/workflows/ci.yml
A5rocks: meh we'll find out if we need it

A5rocks (approved):
If we ever find some way to run the whole test suite under Cython that might be worth switching over (too much effort if it doesn't already exist though), but for now this works.

CoolCat467 (approved):
Looks good to me

jakkdl:
Looking around more for cython+pytest I found https://shwina.github.io/cython-testing/ and after messing around with that for a bit I made some progress and managed to run sync test functions, so "just" need to inject our async magic somewhere in there as well and I might close in on running the full test suite.

I also somehow triggered the error from #2908 due to running against an old trio version in my messing around, despite cython being >3, so I wanna track that down and potentially report it upstream.

But we can merge this for now, and then I'll open another PR if/when I make progress on running the full test suite.

PR #2945, opened 2024-02-03
Support Pyright detection of constructor kwargs in `attrs`-based data classes

mikenerone:
[The changes related to this update fall into three categories:
1. `attrs>=23.2.0` is now required (based on [this comment thread](https://github.com/python-trio/trio/pull/2945#pullrequestreview-1861112927) plus needing to bump to at _least_ 22.2.0 anyway to support Pyright).
2. Adds non-underscored aliases for a few selected private (i.e. leading underscore) attributes in `CancelStatus` and `CancelScope` (this is required for Pyright to recognize private attributes as constructor kwargs).
3. Switching to the "NG" interfaces in recent versions of `attrs` per previous chat discussion.
  - `import attr` -> `import attrs` everywhere
  - `attr.ib` -> `attrs.field`
  - `attr.s` -> `attrs.define` or `attrs.frozen`, depending on existing `frozen` kwarg and with adjustments made to all calls to account for `slots=True` now being the default).
  - `attrs.field(factory=X)` -> `attrs.Factory(X)`
  - `attrs.field(default=X)` -> simply assign X
4. One typing correction that was uncovered by that last "NG" update (see [related comment](https://github.com/python-trio/trio/pull/2945#issuecomment-1925577702)).

Note: there's no new logic to be tested. All existing tests pass unchanged (except for non-logic name changes resulting from item 3 above).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ requires-python = ">=3.8"
 dependencies = [
     # attrs 19.2.0 adds `eq` option to decorators
     # attrs 20.1.0 adds @frozen
-    "attrs >= 20.1.0",
+    "attrs >= 21.2.0",
     "sortedcontainers",
     "idna",
     "outcome",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,9 @@ requires-python = ">=3.8"
 dependencies = [
     # attrs 19.2.0 adds `eq` option to decorators
     # attrs 20.1.0 adds @frozen
-    "attrs >= 21.2.0",
+    # attrs 21.1.0 adds a dataclass transform for type-checkers
+    # attrs 21.3.0 adds `import addrs`
+    "attrs >= 23.2.0",
     "sortedcontainers",
     "idna",
     "outcome",
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -160,7 +160,7 @@ class SystemClock(Clock):
     # Add a large random offset to our clock to ensure that if people
     # accidentally call time.perf_counter() directly or start comparing clocks
     # between different runs, then they'll notice the bug quickly:
-    offset: float = attr.ib(factory=lambda: _r.uniform(10000, 200000))
+    offset: float = attr.field(factory=lambda: _r.uniform(10000, 200000))
 
     def start_clock(self) -> None:
         pass
@@ -226,9 +226,9 @@ class Deadlines:
     """
 
     # Heap of (deadline, id(CancelScope), CancelScope)
-    _heap: list[tuple[float, int, CancelScope]] = attr.ib(factory=list)
+    _heap: list[tuple[float, int, CancelScope]] = attr.field(factory=list)
     # Count of active deadlines (those that haven't been changed)
-    _active: int = attr.ib(default=0)
+    _active: int = attr.field(default=0)
 
     def add(self, deadline: float, cancel_scope: CancelScope) -> None:
         heappush(self._heap, (deadline, id(cancel_scope), cancel_scope))
@@ -318,7 +318,7 @@ class CancelStatus:
     # Our associated cancel scope. Can be any object with attributes
     # `deadline`, `shield`, and `cancel_called`, but in current usage
     # is always a CancelScope object. Must not be None.
-    _scope: CancelScope = attr.ib(alias="scope")
+    _scope: CancelScope = attr.field(alias="scope")
 
     # True iff the tasks in self._tasks should receive cancellations
     # when they checkpoint. Always True when scope.cancel_called is True;
@@ -328,29 +328,29 @@ class CancelStatus:
     # effectively cancelled due to the cancel scope two levels out
     # becoming cancelled, but then the cancel scope one level out
     # becomes shielded so we're not effectively cancelled anymore.
-    effectively_cancelled: bool = attr.ib(default=False)
+    effectively_cancelled: bool = attr.field(default=False)
 
     # The CancelStatus whose cancellations can propagate to us; we
     # become effectively cancelled when they do, unless scope.shield
     # is True.  May be None (for the outermost CancelStatus in a call
     # to trio.run(), briefly during TaskStatus.started(), or during
     # recovery from mis-nesting of cancel scopes).
-    _parent: CancelStatus | None = attr.ib(default=None, repr=False, alias="parent")
+    _parent: CancelStatus | None = attr.field(default=None, repr=False, alias="parent")
 
     # All of the CancelStatuses that have this CancelStatus as their parent.
-    _children: set[CancelStatus] = attr.ib(factory=set, init=False, repr=False)
+    _children: set[CancelStatus] = attr.field(factory=set, init=False, repr=False)
 
     # Tasks whose cancellation state is currently tied directly to
     # the cancellation state of this CancelStatus object. Don't modify
     # this directly; instead, use Task._activate_cancel_status().
     # Invariant: all(task._cancel_status is self for task in self._tasks)
-    _tasks: set[Task] = attr.ib(factory=set, init=False, repr=False)
+    _tasks: set[Task] = attr.field(factory=set, init=False, repr=False)
 
     # Set to True on still-active cancel statuses that are children
     # of a cancel status that's been closed. This is used to permit
     # recovery from mis-nested cancel scopes (well, at least enough
     # recovery to show a useful traceback).
-    abandoned_by_misnesting: bool = attr.ib(default=False, init=False, repr=False)
+    abandoned_by_misnesting: bool = attr.field(default=False, init=False, repr=False)
 
     def __attrs_post_init__(self) -> None:
         if self._parent is not None:
@@ -523,15 +523,15 @@ class CancelScope:
     has been entered yet, and changes take immediate effect.
     """
 
-    _cancel_status: CancelStatus | None = attr.ib(default=None, init=False)
-    _has_been_entered: bool = attr.ib(default=False, init=False)
-    _registered_deadline: float = attr.ib(default=inf, init=False)
-    _cancel_called: bool = attr.ib(default=False, init=False)
-    cancelled_caught: bool = attr.ib(default=False, init=False)
+    _cancel_status: CancelStatus | None = attr.field(default=None, init=False)
+    _has_been_entered: bool = attr.field(default=False, init=False)
+    _registered_deadline: float = attr.field(default=inf, init=False)
+    _cancel_called: bool = attr.field(default=False, init=False)
+    cancelled_caught: bool = attr.field(default=False, init=False)
 
     # Constructor arguments:
-    _deadline: float = attr.ib(default=inf, kw_only=True, alias="deadline")
-    _shield: bool = attr.ib(default=False, kw_only=True, alias="shield")
+    _deadline: float = attr.field(default=inf, kw_only=True, alias="deadline")
+    _shield: bool = attr.field(default=False, kw_only=True, alias="shield")
 
     @enable_ki_protection
     def __enter__(self) -> Self:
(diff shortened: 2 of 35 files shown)


pyproject.toml
```diff
@@ -38,7 +38,7 @@ requires-python = ">=3.8"
 dependencies = [
     # attrs 19.2.0 adds `eq` option to decorators
     # attrs 20.1.0 adds @frozen
-    "attrs >= 20.1.0",
+    "attrs >= 21.2.0",
```
Fuyukai, line 41: There's not really any reason to pin this to a version from three years ago instead of 23.2, especially since 21.3 *did* actually make a backwards incompatible change wrt converters.

Fuyukai:
If you're replacing ``ib`` -> ``define``, you might as well go through and replace ``import attr`` with ``import attrs`` everywhere too.

pyproject.toml
mikenerone: It's usually my habit to base constraints on real compatibility requirements to keep the library as widely-usable for others as possible, but as long as I'm at it, if the team would like like me to constraint to >23.2 for other reasons, I'm happy to. Can we reach some consensus on that in this thread?

mikenerone:
> If you're replacing `ib` -> `define`, you might as well go through and replace `import attr` with `import attrs` everywhere too.

Happy to do that, as well, but that really does require increasing the constraint to `>=21.3.0`, so I'll wait for consensus on what version to constrain to in the [other thread](https://github.com/python-trio/trio/pull/2945#pullrequestreview-1861112927).

pyproject.toml
mikenerone: For others: as @Fuyukai [mentioned](https://github.com/python-trio/trio/pull/2945#issuecomment-1925433171), I could update all `import attr` to the newer `import attrs`, as well. This does sound like a good update to me IMO, but it requires `attrs` 21.3.0. IMO, that's still old enough to seem reasonable, so it's a second reason to make it `attrs>=21.3.0` (in addition to the aforementioned converters change). Anyone have a reason for something even more recent than that? Keep in mind that based on the Changelog, 21.3.0 was the last release that introduced any backward-incompatible changes. This makes it seems like a prime candidate to me.

If anyone's wondering, the converters change, while technically backward-incompatible, is actually one that doesn't require any changes in Trio code (obviously, Trio's been tested for two years since then :P).

mikenerone: Or all the way to the latest (23.2) is fine by me, as well. I just want to make sure there's some consensus on that.

mikenerone: Actually look at the changes more, there are bug-fixes we really should ensure. I'm signing onto the 23.2 constraint.

mikenerone: Added a commit including the `attrs>=23.2.0` constraint _and_ `import attr` -> `import attrs` everywhere.

src/trio/_tests/test_socket.py
```diff
@@ -512,12 +512,12 @@ def gai_without_v4mapped_is_buggy() -> bool:  # pragma: no cover
...
+@attrs.define(slots=False)
 class Addresses:
-    bind_all: str = attr.ib()
-    localhost: str = attr.ib()
-    arbitrary: str = attr.ib()
-    broadcast: str = attr.ib()
+    bind_all: str = attrs.field()
+    localhost: str = attrs.field()
+    arbitrary: str = attrs.field()
+    broadcast: str = attrs.field()
```
TeamSpen210, line 520: With the new attrs version, we don't need the `field()` calls if they don't have any options set. For tests, might as well also remove `slots=False`, since it can't cause backwards compatibility problems.

mikenerone: Re `field`: orly? I did that transform for `attrs.define` and `attrs.frozen` decorators, but didn't realize it applied to non-decorator `attrs.field`, as well, but confirmed. :)

Re: `slots=False`, I'm not sure what you mean. `slots=False` was the default for `attr.s`, so all that didn't specify had it `False`. But with the new ones, `slots=True` is now the default, so I think all of those that didn't specify before now need `slots=False` to guarantee the same behavior. Can confirm that removing that _does_ immediately break a bunch of tests.

mikenerone: Correction: It doesn't break the tests - my mistake. Can you enlighten me as to why this is actually safe to do? It seems to be that if they were slotted before, we want them to still not be slotted now.

mikenerone: I deleted a comment here because I thought I'd made a mistake about that breaking the tests, but no, confirmed it does break them.

mikenerone: It also turns out that while the bare `x = attrs.field` construct works at runtime, it makes both `pyright` and `mypy` very unhappy, and it fails CI. Apparently it's not supported by type-checkers yet. I've removed a3550387fc902f2cae61cb231e17a57ee07ac3d8 from the branch.

TeamSpen210: For `field`, I was meaning using an annotation, no assignment at all. I was thinking that making them slotted would likely work in the tests, but I guess not.

CoolCat467: I would be in favor of enabling slots everywhere where it makes sense to do so

src/trio/_core/_run.py
```diff
@@ -328,29 +328,29 @@ class CancelStatus:
...
-    effectively_cancelled: bool = attr.ib(default=False)
+    effectively_cancelled: bool = attrs.field(default=False)
 
     # The CancelStatus whose cancellations can propagate to us; we
     # become effectively cancelled when they do, unless scope.shield
     # is True.  May be None (for the outermost CancelStatus in a call
     # to trio.run(), briefly during TaskStatus.started(), or during
     # recovery from mis-nesting of cancel scopes).
-    _parent: CancelStatus | None = attr.ib(default=None, repr=False)
+    _parent: CancelStatus | None = attrs.field(default=None, repr=False, alias="parent")
```
CoolCat467, line 338: Why are we exposing this?

src/trio/_core/_run.py
```diff
@@ -523,15 +523,15 @@ class CancelScope:
...
+    _has_been_entered: bool = attrs.field(default=False, init=False)
+    _registered_deadline: float = attrs.field(default=inf, init=False)
+    _cancel_called: bool = attrs.field(default=False, init=False)
+    cancelled_caught: bool = attrs.field(default=False, init=False)
 
     # Constructor arguments:
-    _deadline: float = attr.ib(default=inf, kw_only=True)
-    _shield: bool = attr.ib(default=False, kw_only=True)
+    _deadline: float = attrs.field(default=inf, kw_only=True, alias="deadline")
+    _shield: bool = attrs.field(default=False, kw_only=True, alias="shield")
```
CoolCat467, line 534: Why are we adding alias here?

CoolCat467 (commented):
Looks pretty good, but I have a few questions.
First, why are we adding aliases in a few parts?
Secondly, are we able to enable slots on more of these?

src/trio/_core/_run.py
TeamSpen210: This changes the name of the parameter defined in `__init__`. Attrs already strips the underscore, so this doesn't actually change anything. But explicitly specifying the alias makes Pyright recognise this correctly.

pyproject.toml
A5rocks: I'd rather pin based on what we require rather than what's latest, because by default people will get the latest (with pip) and all we're causing is incompatibility with things that pin older attrs.

Fuyukai: > and all we're causing is incompatibility with things that pin older attrs.

This is a fake problem.

A5rocks: You say that as though tools like Poetry doesn't put an upper bound by default :'(

(e.g. I clicked 2 arbitrary libraries that depend on `attrs` and one only depended on it in test extras (that were thankfully pinned) and the other used `~=` which has an upper bound)

src/trio/_tests/test_socket.py
mikenerone: Sure, but that seems like a different issue.

mikenerone: @TeamSpen210 Ohh, of course! Now that you point it out, I don't know how I missed it. Type checkers have no issue with that, added 88473eaafac2350f65e1fd51824e0218183ad17e.

pyproject.toml
mikenerone: I'm happy with whatever y'all decide.

src/trio/_core/_run.py
mikenerone: Correct. This is explained in the PR description.

src/trio/_core/_run.py
mikenerone: As explained in the PR description, my main goal for this PR was to make Pyright stop complaining about constructor kwargs for these classes. These aliases + `attrs>=21.2.0` is what actually accomplishes that (though I was perfectly happy to handle some other `attrs`-related updates that were agreed upon in chat and comments).

CoolCat467 (approved):
After explanation of alias, looks good to me!

TeamSpen210 (approved):
Much nicer indeed.

mikenerone:
Added two more commits for additional shorter syntaxes:
- `attrs.field(factory=X)` -> `attrs.Factory(X)`
- `attrs.field(default=X)` -> simply assign X

pyproject.toml
mikenerone: After a few more commits for even more places where more concise syntaxes are now enabled, I believe this discussion about the `attrs` version constraint is the only open question.

CoolCat467 (approved):

pyproject.toml
mikenerone: Running tests with the different versions of `attrs`, I discovered that these changes to make Pyright work actually require `attrs=22.2.0`, because that's when support for the `alias` parameter for `attrs.field()` [was introduced](https://www.attrs.org/en/stable/changelog.html#id7) (note: link correct at time of writing but looks unstable).

Would 22.2.0, only two releases back from current and released in December 2022, be an acceptable middle ground for everyone?

@Fuyukai @A5rocks @TeamSpen210

A5rocks: Pinning to 22.2 is what I proposed (in my edit for clarity), so yeah!

Fuyukai: 23.2 switched to using the PEP dataclass transform as opposed to the version only supported by Pyright, which makes it pretty obviously the only correct option in a statically typed Python world.

A5rocks: Oh I didn't realize. I don't really know whether bumping versions makes sense then (if one of our deps didn't work on 3.12 at the earliest version we support, would we bump its pin?)

But I'm alright with this change cause at least it's got a technical reason (not +1, not -1).

mikenerone: I certainly agree that _users_ should want to be updating to 23.2 for that reason among others, but (and I hate to say this because I'd like to see this PR merged) I'm not so sure that means that it's Trio's place to enforce it on them.

src/trio/_core/_run.py
jakkdl, line 534: if it was me I'd perhaps do `from attrs import define, Field, Factory` etc, since we're only using a couple different objects from `attrs`, reuse them a bunch of times, and are widely used enough in the codebase I don't think there'd be much confusion about what library they're from. (and as opposed to `attr.ib` where `ib` on its own doesn't make sense). E.g. in this block it would cut down on the sheer number of characters a decent amount.

But seems plenty fine to stick with `import attrs`

jakkdl (approved):
Only skimmed the changes, but seems fine to merge. For ease of review it's perhaps not great to bundle a bugfix with a verbose style change, (and slots changes, with is coolcat's hobbyhorse :P) but probably not worth splitting it up at this point.

Another plus would be a regression test in `src/trio/_tests/type_tests/`, which are run with both pyright and mypy. The overkill approach would be something that comprehensively tested all parameters to functions, but given how often `trio/_tests/test_exports` is a hassle it might even be net negative.

Very happy to be moving to modern usage where reading the official docs for `attrs` is much easier, it confused me a bit when I first started familiarizing myself with the code.

mikenerone:
> Only skimmed the changes, but seems fine to merge. For ease of review it's perhaps not great to bundle a bugfix with a verbose style change, (and slots changes, with is coolcat's hobbyhorse :P) but probably not worth splitting it up at this point.

@jakkdl By way of explanation, because the Pyright fix required a bump to the `attrs` dep version, this PR really became an "update `attrs` and use all of the next-gen interfaces that are now prescribed in this new version." From that perspective, this really was _just_ a style PR resulting from the dep update. The "bugfix" part we got for _free_ with the update - there are zero extra lines here to accomplish that. Also, there are no slots changes here, and in fact intentionally zero changes in any functional way like that.

PR #2947, opened 2024-02-06
Enable ruff's `flake8-trio` rule

CoolCat467:
Probably should enable the TRIO rules? Maybe in a separate PR.

_Originally posted by @TeamSpen210 in https://github.com/python-trio/trio/issues/2946#issuecomment-1928600531_

This pull request enables the `flake8-trio` ruff rule (but in ruff's system it's marked as `TRIO`).

### DIFF
--- a/src/trio/_tests/test_threads.py
+++ b/src/trio/_tests/test_threads.py
@@ -24,6 +24,8 @@ from typing import (
 import pytest
 import sniffio
 
+from trio import lowlevel
+
 from .. import (
     CancelScope,
     CapacityLimiter,
@@ -165,7 +167,7 @@ def test_run_in_trio_thread_ki() -> None:
         thread.start()
         print("waiting")
         while thread.is_alive():
-            await sleep(0.01)
+            await lowlevel.checkpoint()
         print("waited, joining")
         thread.join()
         print("done")
@@ -535,7 +537,7 @@ async def test_run_in_worker_thread_limiter(
             # check below won't fail due to scheduling issues. (It could still
             # fail if too many threads are let through here.)
             while state.parked != MAX or c.statistics().tasks_waiting != MAX:
-                await sleep(0.01)  # pragma: no cover
+                await lowlevel.checkpoint()  # pragma: no cover
             # Then release the threads
             gate.set()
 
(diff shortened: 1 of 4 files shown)


CoolCat467:
At the moment the tests will fail until #2946 gets merged, and even then there will be a failure detailed in https://github.com/astral-sh/ruff/issues/9855 that is a bug.

jakkdl:
Disclaimer: I'm the one that wrote 99% of the code in flake8-trio.

Ruffs implementation of flake8-trio is only a small set (6 out of ~30) of the full rules: https://github.com/Zac-HD/flake8-trio Although a few of them are handled under other rules.

The plugin is intended for *users* of trio or anyio, so some rules might not make sense in trio itself, but can ofc then just disable those.

It is also possible to run flake8-trio as a standalone, so we don't have to bring back flake8. (because we wanted to implement autofix with libcst, which isn't possible when running with flake8)

(so yes, both parts of the flake8-trio name are incorrect... https://github.com/Zac-HD/flake8-trio/issues/124#issuecomment-1454034177 )

src/trio/_tests/test_threads.py
```diff
@@ -165,7 +167,7 @@ def external_thread_fn() -> None:
         thread.start()
         print("waiting")
         while thread.is_alive():
-            await sleep(0.01)
+            await lowlevel.checkpoint()
```
A5rocks, line 170: I think it's probably better to use `sleep` here because otherwise we'll be busy looping I think?

Like this isn't a matter of "trio just needs another cycle or two of the available tasks" this is "we need to wait for the thread to start and do its stuff"

This happens 3 other places too.

A5rocks (changes requested):
Just quick thing I think isn't great

CoolCat467:
Given that ruff's version has a few issues and other points mentioned, I don't think this particular rule is fit for Trio at the moment.

src/trio/_tests/test_threads.py
jakkdl: this is also a bug in ruffs implementation, flake8-trio only triggers when the literal is exactly `0`

jakkdl: wait no, ruff gave TRIO110, which suggests to await an `Event` instead of looping a sleep. Resolving that by replacing with a checkpoint is ... definitely not the way to go.
Maybe I should widen TRIO110 to also catch checkpoints to make sure people don't confuse them and thinks that because it stops warning on `checkpoint` that it "solves" it

CoolCat467: I manually changed the 0.02 to a checkpoint because an error was being raised from the sleep there.

jakkdl: yeah an Event could plausibly not be possible or make sense in a context, in which case `noqa` would be the solution :)

PR #2953, opened 2024-02-12
Make sure we're running on M1 MacOS

A5rocks:
`macos-latest` doesn't seem to include [`macos-14`](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/): https://github.com/python-trio/trio/actions/runs/7865413592/job/21458313729#step:1:8

So this should make things run on M1 macs.

Reference: https://github.com/python-trio/trio/issues/2885


A5rocks:
Oh despite 3.8 and 3.9 having installers that work on arm, it looks like setup-python doesn't recognize that yet: https://github.com/actions/setup-python/issues/808

From the issue it sounds like all that's missing is someone needs to build for 3.8 and 3.9 arm64? So I'm comfortable letting this sit until that happens.

A5rocks:
This should be unblocked now

A5rocks:
Actually this is no longer necessary

macos-latest will now be M1 macos

PR #2954, opened 2024-02-13
Added link to `logot` in docs

etianen:
I've just released [`logot`](https://github.com/etianen/logot), a testing library with some nice tools for testing highly-concurrent async code using log capture. It includes a [`trio` integration](https://logot.readthedocs.io/latest/integrations/trio.html), so it would be great to add it to the list of testing tools in your docs. ❤️ 

Here's a quick example of how the `trio` integration looks:

``` py
from logot import Logot, logged

async def test_app(logot: Logot) -> None:
      nursery.start_soon(app.start())
      await logot.await_for(logged.info("App started"))
```

Of course I'd welcome any feedback you have on the `trio` integration too! 🙇

### DIFF
--- a/docs/source/awesome-trio-libraries.rst
+++ b/docs/source/awesome-trio-libraries.rst
@@ -94,6 +94,7 @@ Testing
 * `hypothesis-trio <https://github.com/python-trio/hypothesis-trio>`__ - Hypothesis plugin for trio.
 * `trustme <https://github.com/python-trio/trustme>`__ - #1 quality TLS certs while you wait, for the discerning tester.
 * `pytest-aio <https://github.com/klen/pytest-aio>`_ - Pytest plugin with support for trio, curio, asyncio
+* `logot <https://github.com/etianen/logot>`_ - Test whether your async code is logging correctly.
 
 
 Tools and Utilities


A5rocks (approved):
Just for the record, in general we recommend that you guarantee an `await` point in every `async` function: it looks like `await_for` doesn't.

But this looks good.

etianen:
Thanks for the approval. Um... what happens now.. I can't actually merge! 😅 

> Just for the record, in general we recommend that you guarantee an await point in every async function: it looks like await_for doesn't.

Thanks for the feedback! Can you clarify what you mean here? The point of the early return is to avoid creating an `Event` and waiting on it when the expected log are already satisfied. Are you suggesting an `await trio.lowlevel.checkpoint()` call? What does that gain?

A5rocks:
Yeah I'm suggesting that if `_start_waiting` returns `None`, then you `await trio.lowlevel.checkpoint()`. I don't think the benefits are massive for tests, but it's a convention we suggest: it's nice to know where cancellation will be propagated and when other tasks will run. (and certainty about both is worth a lot).

> Thanks for the approval. Um... what happens now.. I can't actually merge! 😅

Yeah I'm just waiting for someone else to make sure its fine.

etianen:
Thanks for the explanation! I've not come across that convention before, but all my async work has been in `asyncio`. (I find the arguments for `trio` very convincing, btw, but I was already too deep on my various async codebases to change).

No rush on the PR... I just wanted to make sure you hadn't forgotten to press the button because _I do that to people all the time on my projects!_ 🤦

PR #2955, opened 2024-02-15
Fixing `Path` for Python 3.13

etianen:


.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['pypy-3.9', 'pypy-3.10', '3.8', '3.9', '3.10', '3.11', '3.12']
+        python: ['pypy-3.9', 'pypy-3.10', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13-dev']
```
A5rocks, line 90: I think it's fine to add fixes to 3.13 without adding it to CI, if doing so is failing.

A5rocks (approved):

.github/workflows/ci.yml
etianen: I've reverted the change to CI, leaving just the Python 3.13 fix. I'm hoping this will all pass nicely now. ❤️

PR #2957, opened 2024-02-16
Switch to use `uv pip install`

### LINKED ISSUE #2956: Switch to ``uv`` in CI
[uv](https://astral.sh/blog/uv) is a new tool from the Ruff guys that replaces ``pip`` and ``pip-tools``.

This is mostly interesting because it's actually being benchmarked on Trio itself. Sounds like it's already been tested for us.

CoolCat467:
This pull request follows up on #2956 and is attempting to switch the continuous integration system to use astral.sh' new tool [uv](https://astral.sh/blog/uv).


A5rocks:
This seems nontrivial, can we break this into 2 prs?
1) replace pip-compile
2) replace pip install

Personally I see (1) as much more valuable-- (2) doesn't change much. Personally I'm more excited by uv's ability to get the lowest compatible version so we can add a new CI run than by it replacing `pip install`.

ci.sh
```diff
@@ -37,25 +37,43 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
...
+fi
+
+# Make sure pip is installed (pypy weirdness)
+python -m ensurepip
+
+# Install uv in venv
+python -m pip install --upgrade uv
+
+# Make sure build is there
+uv pip install build
```
jakkdl, line 58: Why do we need to create a virtualenv? In CI it shouldn't be necessary, and for developers I feel like it's better to let the developer be responsible for handling their venv and not recreate the venv every time you run `ci.sh`

CoolCat467: I think there were some issues because I didn't know about `--system` mode. Otherwise, on default, `uv` uses a virtual environment instead of touching system python installation. Thanks for reminding me about this, I'll get that fixed.

CoolCat467: I believe I have fixed this

ci.sh
```diff
@@ -37,25 +39,53 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
...
+    python -m pip install --upgrade uv
+fi
+
+if [ "$ON_WINDOWS" = "True" ]; then
+    PYTHON_PATH=$(python -c "import os, sys; print(os.path.dirname(sys.executable))")
+else
+    PYTHON_PATH=$(which python)
+fi
+
+uv pip install --python="$PYTHON_PATH" build
```
jakkdl, line 70: Is this needed?

> When running pip sync or pip install, uv will search for a virtual environment in the following order:
> 1. An activated virtual environment based on the VIRTUAL_ENV environment variable.
> 2. An activated Conda environment based on the CONDA_PREFIX environment variable.
> 3. A virtual environment at .venv in the current directory, or in the nearest parent directory.

https://github.com/astral-sh/uv?tab=readme-ov-file#advanced-usage

ci.sh
```diff
@@ -37,25 +39,53 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
...
+# expands to 0 != 1 if MAKE_VENV is not set, if set the `-0` has no effect
+# https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02
+if [ "${MAKE_VENV-0}" == 1 ]; then
+    uv venv .venv --seed
+
+    if [ "$ON_WINDOWS" = "True" ]; then
+        source ./.venv/Scripts/activate
+    else
+        source .venv/bin/activate
+    fi
```
jakkdl, line 55: I'm skeptical we need to keep the `MAKE_VENV` code at all, if we want isolated build&testing environment locally we should be using tox (though uv plans to reimplement that themselves at some point :tm:).
But I guess I'm fine with leaving this around until then, even though I see very marginal use for it.

Ideally environmental variables like this should be documented in the contributing guide, if we expect developers to use it at all. If this is only a CI thing... then we truly can remove it since you don't seem to set it in the CI config?

The contributor docs should maybe also be changed to instruct users to use `uv` rather than pip directly, don't think it makes a difference now but might break in the future if we start using uv-specific features?

CoolCat467: The reason I left it is because on my machine, I installed python with system package manager, which causes both pip and uv to raise warnings about how modifying system python installation can be harmful, and using a virtual environment means that doesn't happen.

jakkdl: whenever working with packages it's highly recommended you as a developer create a separate virtualenv for each repo. I try to have that as a step in any contributor docs I write/edit, but maybe haven't given trio that treatment

jakkdl: https://trio.readthedocs.io/en/stable/contributing.html#tests does only a weak suggestion atm... that's easy to miss due to the length of the line 🙃:

̀̀```
cd path/to/trio/checkout/
pip install -r test-requirements.txt # possibly using a virtualenv
pytest trio 
```

I'd change that to

̀̀```
cd path/to/trio/checkout/
python -m venv .venv
source .venv/bin/activate
pip install -r test-requirements.txt
pytest trio 
```

ci.sh
CoolCat467: Removed in 57b8d2f

ci.sh
CoolCat467: Removed `MAKE_ENV` in 57b8d2f

ci.sh
```diff
@@ -37,7 +37,7 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
 echo "::endgroup::"
 
 echo "::group::Install dependencies"
-python -m pip install -U pip $(grep "uv==" test-requirements.txt)
+python -m pip install -U pip "$(grep 'uv==' test-requirements.txt)"
```
A5rocks, line 40: We can use a similar technique to https://github.com/python-trio/trio/pull/3061 and use `python -m pip install pip uv -c test-requirements.txt`.

CoolCat467: First off, uv implements constraints through `uv pip -c constraints.txt`.
[I was looking into what exactly constraints do](https://stackoverflow.com/questions/34645821/pip-constraints-files), and from what I can tell, the way we are using uv, installing from a locked dependencies file, there is no reason to use constraints. They seem purely limit the version of a dependency that is able to be installed, and if we are specifying and pinning all dependencies with the lock file, I'm pretty sure using constraints shouldn't be necessary.
Please correct me if I am wrong, I could be misunderstanding something.

A5rocks: I'm only recommending using constraints when initially installing `uv` rather than grepping the test requirements for consistency reasons.

CoolCat467: Changed in bc3a94e

ci.sh
```diff
@@ -37,25 +37,29 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
...
+python -m pip install -U pip uv -c test-requirements.txt
 python -m pip --version
+python -m uv --version
+
+python -m uv pip install build
 
 python -m build
-python -m pip install dist/*.whl
+wheel_package=$(ls dist/*.whl)
+python -m uv pip install "trio @ $wheel_package"
```
A5rocks, line 48: We can probably lead the path with ./ to get this to work like before: https://github.com/astral-sh/uv/issues/6052

A5rocks (commented):
Some of your changes remind me that we should probably run shellcheck over our scripts...

A5rocks:
> > Some of your changes remind me that we should probably run shellcheck over our scripts...
> 
> Problem is, it's written in haskell and the pre-commit integration for it uses docker, which is a mess

Yeah I think it's fine not to check it in CI. It's not like our shell scripts are that important or change often; it's just a nicety.

ci.sh
CoolCat467: Is there a particular reason not to leave it at my current solution?

A5rocks: Personal preference TBH. Maybe you disagree but I prefer moving logic out of shell scripts and into program arguments.

Feel free not to do it though.

A5rocks (approved):

Issue #2944, opened 2024-02-02
[Heads up] Test failures with Python 3.13.0a3


befeleme:
I used the fix from https://github.com/python-trio/trio/pull/2918 to attempt to build trio in Fedora with python 3.13.0a3. This is an ongoing effort to integrate new Python as soon as possible and provide feedback to all interested parties.
The result of the attempted build is that 8 tests fail:

```
=================================== FAILURES ===================================
___________________ test_compare_async_stat_methods[is_dir] ____________________

method_name = 'is_dir'

    @pytest.mark.parametrize("method_name", ["is_dir", "is_file"])
    async def test_compare_async_stat_methods(method_name: str) -> None:
>       method, async_method = method_pair(".", method_name)

../../BUILDROOT/python-trio-0.23.1-4.fc40.x86_64/usr/lib/python3.13/site-packages/trio/_tests/test_path.py:119: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../BUILDROOT/python-trio-0.23.1-4.fc40.x86_64/usr/lib/python3.13/site-packages/trio/_tests/test_path.py:26: in method_pair
    return getattr(sync_path, method_name), getattr(async_path, method_name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = trio.Path('.'), name = 'is_dir'

    def __getattr__(self, name):
        if name in self._forward:
            value = getattr(self._wrapped, name)
            return rewrap_path(value)
>       raise AttributeError(name)
E       AttributeError: is_dir

../../BUILDROOT/python-trio-0.23.1-4.fc40.x86_64/usr/lib/python3.13/site-packages/trio/_path.py:236: AttributeError
___________________ test_compare_async_stat_methods[is_file] ___________________

method_name = 'is_file'

    @pytest.mark.parametrize("method_name", ["is_dir", "is_file"])
    async def test_compare_async_stat_methods(method_name: str) -> None:
>       method, async_method = method_pair(".", method_name)

../../BUILDROOT/python-trio-0.23.1-4.fc40.x86_64/usr/lib/python3.13/site-packages/trio/_tests/test_path.py:119: 
[...]

PR #2910, opened 2023-12-20
bump pyright, introducing warnings for tons of missing docstrings

jakkdl:
With https://github.com/microsoft/pyright/issues/6758 released, we now have warnings for missing docstrings on ... 22 classes and *133* functions.

A lot of them are sockets and path/fileIO, which we mostly can ignore. It's maybe time to write a program that filters the output, which would also enable us to run without `--ignoreexternal`. At least the first ~40 should probably get some simple docstrings written for them though.

```
trio._core._run.Task
trio._sync.AsyncContextManagerMixin
trio._sync._LockImpl
trio._core._io_epoll._EpollStatistics
trio._core._local._NoValue
trio._core._local.RunVarToken
trio.lowlevel.RunVarToken
trio.lowlevel.Task
trio._core._ki.KIProtectionSignature
trio._core._mock_clock.MockClock.start_clock
trio._core._mock_clock.MockClock.current_time
trio._core._mock_clock.MockClock.deadline_to_sleep_time

# protocol methods
trio._subprocess.HasFileno.fileno
trio._sync._HasAcquireRelease.acquire
trio._sync._HasAcquireRelease.release

# channels
trio.MemoryReceiveChannel
trio._channel.MemoryReceiveChannel
trio._channel.MemoryReceiveChannel.statistics
trio._channel.MemoryChannelStats
trio._channel.MemoryReceiveChannel.aclose
trio.MemorySendChannel
trio._channel.MemorySendChannel
trio._channel.MemorySendChannel.statistics
trio._channel.MemorySendChannel.aclose
trio.open_memory_channel
trio._channel.open_memory_channel

# streams
trio._highlevel_socket.SocketStream.send_all
trio._highlevel_socket.SocketStream.wait_send_all_might_not_block
trio._highlevel_socket.SocketStream.send_eof
trio._highlevel_socket.SocketStream.receive_some
trio._highlevel_socket.SocketStream.aclose
trio._unix_pipes.FdStream.send_all
trio._unix_pipes.FdStream.wait_send_all_might_not_block
trio._unix_pipes.FdStream.receive_some
trio._unix_pipes.FdStream.close
trio._unix_pipes.FdStream.aclose
trio._unix_pipes.FdStream.fileno


# path / fileIO
trio._file_io._HasFileNo
trio._file_io._HasFileNo.fileno
trio._file_io.AsyncIOWrapper.fileno
trio._file_io.AsyncIOWrapper.isatty
trio._file_io.AsyncIOWrapper.readable
[...]

PR #2959, opened 2024-02-17
`Path` refactor

etianen:
This PR implements the proposed refactor in #2944. As well as fixing the Python 3.13 problems, it adds some cool new things! 😎 

## The big change 💥 

`trio.Path` now subclasses `pathlib.PurePath`!

This means all the sync method forwarding can be completely removed!

Since it now fits into the `pathlib` class hierarchy, it can be used interchangeably with other pure paths and methods that take pure paths. This also has the nice effect of making the wrapped async methods _just work_ with regard to input types.

It's also now a non-virtual subclass of `os.PathLike[str]`, which might help some static analysis tools...?

Two concrete subclasses `PosixPath` and `WindowsPath` also now exist, matching how `pathlib` does things. Instantiating a `Path` actually gets you the appropriate one for the platform, just like `pathlib`.

## The medium change

Async method wrapping is now explicit, fixing Python 3.13. There are three high-level wrapper helpers that work for most cases (`_wrap_method`, `_wrap_method_path` and `_wrap_method_path_iterable`). For methods that don't fit into these, a low-level (`_wraps_async`) is used.

## Other changes

`Path` is now slotted, to match `pathlib`. A minor speed boost, with luck!

## Fixing Python 3.13

Unlike #2955, this fix actually works! Check it out:

```
================================================= test session starts ==================================================
platform darwin -- Python 3.13.0a3+, pytest-8.0.1, pluggy-1.4.0
rootdir: /Users/dave.hall/Workspace/trio
configfile: pyproject.toml
collected 33 items                                                                                                     

src/trio/_tests/test_path.py .................................                                                   [100%]

================================================== 33 passed in 0.05s ==================================================
```

## My battle with `pyright --verifytypes` 😭 

The `pyright --verifytypes` check does **not** like this syntax:

```py
stat = _wrap_method(pathlib.Path.stat)
```

It claims this is an ambiguous type, asking me to annotate it with something that _can't be expressed with the Python type system_ except for a per-method unweildy protocol!

The workaround would be to put in a dummy method and use a decorator syntax, which `pyright` thinks is absolutely okay despite _inferring to exactly the same thing_! It's also pretty disgusting to look at, and bloats the implementation with boilerplate.

```py
@_wrap_method(pathlib.Path.stat)
def stat(self) -> Any:
    raise AssertionError("unreachable!")
```

I can't see any way to disable this check for just this file, or just these lines. Any suggestions?! 😭


src/trio/_tests/test_path.py
```diff
@@ -168,41 +167,6 @@ async def test_repr() -> None:
...
-
-    assert not hasattr(MockWrapper, "_private")
-
-
-async def test_type_wraps_private() -> None:
-    WrapperType.generate_wraps(MockWrapper, {"unsupported": None})
-
-    assert not hasattr(MockWrapper, "_private")
-
-
```
etianen, line 205: All of this can be removed, now we're not using metaclasses.

src/trio/_tests/type_tests/path.py
```diff
@@ -39,7 +39,7 @@ def sync_attrs(path: trio.Path) -> None:
     assert_type(path.drive, str)
     assert_type(path.root, str)
     assert_type(path.anchor, str)
-    assert_type(path.parents[3], pathlib.Path)
```
etianen, line 42: I count this as a bugfix!

src/trio/_tests/test_exports.py
```diff
@@ -306,7 +306,8 @@ def lookup_symbol(symbol: str) -> dict[str, str]:
                     mod_cache = next_cache / "__init__.data.json"
                 else:
                     mod_cache = mod_cache / (modname[-1] + ".data.json")
-
```
etianen, line 309: Bugfix for `os.Pathlike`

src/trio/_path.py
```diff
@@ -32,219 +21,97 @@
...
-    from typing_extensions import Concatenate, Literal, ParamSpec, TypeAlias
-
-    from trio._file_io import AsyncIOWrapper as _AsyncIOWrapper
+    from typing_extensions import Concatenate, Literal, ParamSpec
 
     P = ParamSpec("P")
 
-T = TypeVar("T")
-StrPath: TypeAlias = Union[str, "os.PathLike[str]"]  # Only subscriptable in 3.9+
+    Self = TypeVar("Self", bound="Path")
```
CoolCat467, line 28: Why are we creating a `Self` type when `from typing_extensions import Self` exists?

src/trio/_path.py
```diff
@@ -32,219 +21,97 @@
...
+    wrapped: Callable[..., Any]
+) -> Callable[[Callable[P, T]], Callable[P, Awaitable[T]]]:
+    def decorator(fn: Callable[P, T]) -> Callable[P, Awaitable[T]]:
+        async def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
+            return await run_sync(partial(fn, *args, **kwargs))
 
+        update_wrapper(wrapper, wrapped)
+        wrapper.__doc__ = (
+            f"Like :meth:`~{wrapped.__module__}.{wrapped.__qualname__}`, but async."
+        )
```
CoolCat467, line 42: Would it be possible to include the original documentation after this header? I've personally always found it frustrating when I'm looking into what trio path objects can do and all the documentation just says "hey go look at this other thing" instead of it just being right there.

CoolCat467 (commented):
Questions about a few minor things

src/trio/_path.py
etianen: A good question! The `Self` type is special and can only be used in conventional class method declarations. Since we're using it here in a decorator/wrapper, it's not usable.

src/trio/_path.py
etianen: I'd prefer that too! I was just following convention! I wonder if anyone else has any opinions.

jakkdl: I suspect the current convention is out of laziness/complexity, rather than it being the optimal thing :tm:. If you can copy the old one and prepend that text, or something similar, I think it would be much superior.

src/trio/_tests/test_exports.py
jakkdl: would you mind elaborating on what the bug/fix is? This code should probably be commented cause reading it now I don't understand almost any of it x)

jakkdl:
> The pyright --verifytypes check does not like this syntax:
> `stat = _wrap_method(pathlib.Path.stat)`
> It claims this is an ambiguous type, asking me to annotate it with something that can't be expressed with the Python type system except for a per-method unweildy protocol!
> The workaround would be to put in a dummy method and use a decorator syntax, which pyright thinks is absolutely okay despite inferring to exactly the same thing! It's also pretty disgusting to look at, and bloats the implementation with boilerplate.

The error message is specifically about it being ambigious and could be inferred differently by *other type checkers*, so whether it's being inferred to the same thing *by pyright* is not really relevant. In most other cases we've resorted to doing the disgusting/bloated boilerplate to get around these errors, but I'll have a closer look later on whether I think it's worth it. We are after all running type checks on both mypy & pyright, so we can probably ignore it. 

> I can't see any way to disable this check for just this file, or just these lines. Any suggestions?! 😭

Maybe I should resuscitate https://github.com/python-trio/trio/pull/2910 and get the output-filtering script finished/merged.

src/trio/_path.py
etianen: I'll do that this evening. A good call! I actually had it set up this originally, then reigned in my ambitions and followed conventions. 😅

src/trio/_tests/test_exports.py
etianen: Guess what... I don't really understand it either! 😢 

Here's what I do know... this code is mapping a module name to a file in the `mypy` cache dir. For modules, this file is `foo.data.json`. for packages, this file is `package/__init__.data.json`.

You'll see a few lines above this change that the case of a package is handled for things like `foo.bar.baz` (i.e. if `modname` is truthy). But the package case is not otherwise handled. Now it is.

I found this check extremely confusing in general. I can see what it's trying to do, but it was thrown utterly by my completely reasonable _subclassing of a stdlib class_. Having to figure out what this complex linter was doing and then put in `missing -= PurePath.__dict__.keys()` just to keep it happy about _inheritance_ was a good 45 mins of frustration! And then I had to do it all again for the case of windows-only methods with overrides specific to `jedi` and `mypy`! In this latter case, the only way to test it was a push-to-CI-and-wait cycle.

It's working now, but are we sure this linter is worth the maintenance burden? An opinionated `ruff` coupled with very strict `mypy` is already some pretty heavy-duty guardrails. This discussion is out of scope for the PR, I think though.

etianen:
> The error message is specifically about it being ambiguous and could be inferred differently by other type checkers

I think I'm disagreeing here with `pyright`. I can write the same thing using a decorator syntax, which provides no additional typing information, and it's happy. It just seems to be complaining that I'm doing a variable assignment and not explicitly annotating the variable.

This is related to my point about the static visibility linter, really. This check seems to have a very uncompromising idea of what's _ambiguous_. The use of `ParamSpec` and `Concatenate` here has, AFAIK, absolutely only one possible meaning. That's not ambiguous! It's simply not _explicit_.

I'm a huge fan of static analysis tools in general, and favor very strict typing and linting. But to me, these checks feel overboard. I realize that this _exactly_ the argument people make to me about strict `mypy` checks, so maybe it's a case of comfort levels? But at the point inheritance and strictly-typed metaprograming become forbidden, it feels too much to me.

I'm left in a bit of a quandry for this PR though. Hopefully you have a solution in mind with #2910.

etianen:
Also, apologies if the tone of my messages comes over as overly critical! I'm a big fan of how welcoming this project is to new contributors, so this stems from a concern about placing too-high barriers in front of new contributions.

It's hard to express frustration without sounding... _frustrated_. 🥵

src/trio/_path.py
jakkdl: I kinda think it should be named differently to avoid confusion with the special type, maybe `PathT`?

src/trio/_tests/test_exports.py
jakkdl: They were introduced because we in fact have had errors several times, the logic for exported symbols is kinda overly complex. But for a PR like this that overhauls a complex part I do find it good to know that we have exhaustive tests that would catch if something got missed.
But the burden they impose on contributors is non-trivial, for sure. I can maybe try to add more comments to the code & improve error messages to make it easier to know how to handle new problems at least.

jakkdl:
disagreeing with pyright seems very fair in this case, and given how complex the Path wrapping is already I'm open to just ignoring what `--verifytypes` says. I don't know for sure why it forbids assignment here.
Will look at stuff tomorrow :)

src/trio/_tests/test_exports.py
etianen: It's definitely a call for core project maintainers to make! 🙇

src/trio/_path.py
etianen: Calling this sort of thing `Self` was fairly common before the _real_ `Self` type was introduced. But yeah, I'll do that, it makes sense now!

etianen:
Thanks for the feedback! I've implemented the suggested changes. The new docs in particular look very nice! 💪 

image

I'll await your findings on the `--verifytypes` problem. I couldn't find a way to selectively ignore this file, so hopefully you'll have a better time of it than me! 😅

CoolCat467:
One slightly interesting point https://trio--2959.org.readthedocs.build/en/2959/reference-io.html#trio.Path.link_to

etianen:
> One slightly interesting point https://trio--2959.org.readthedocs.build/en/2959/reference-io.html#trio.Path.link_to

I'm assuming these docs were built not on Python 3.12? The method is indeed gone in 3.12.

A5rocks:
Yeah we build docs on an older Python. No matter what method availability depends on Python version and we don't have a setup for documenting that. I think I had a PR open to build docs on 3.12 but closed it when I realized that it would mean our Path docs would be less useful for anyone not using 3.12 (due to method removals).

docs/source/reference-io.rst
```diff
@@ -631,6 +631,11 @@ Asynchronous path objects
 
 .. autoclass:: Path
    :members:
+   :inherited-members:
+
+.. autoclass:: PosixPath
+
+.. autoclass:: WindowsPath
```
A5rocks, line 638: What is the use case of using `trio.WindowsPath` or `trio.PosixPath`? Looking at the code, it looks like we only wrap methods from `PurePath` so it's not like they get more functionality (?)

I'm probably missing something trivial though. Though if it's not necessary to then I think having to put platform-specific types is weird for an exported API.

src/trio/_path.py
```diff
@@ -32,219 +22,102 @@
...
-def _forward_magic(
-    cls: AsyncAutoWrapperType, attr: Callable[..., T]
-) -> Callable[..., Path | T]:
-    sentinel = object()
+def _wrap_method(
+    fn: Callable[Concatenate[pathlib.Path, P], T],
+) -> Callable[Concatenate[Path, P], Awaitable[T]]:
+    @_wraps_async(fn)
+    def wrapper(self: Path, /, *args: P.args, **kwargs: P.kwargs) -> T:
+        return fn(pathlib.Path(self), *args, **kwargs)
```
A5rocks, line 57: Doing `pathlib.Path(self)` rather than caching it as `self._wrapped` might have implications? I'm not familiar with what the `pathlib.Path` `__init__` does though.

src/trio/_path.py
```diff
@@ -32,219 +22,102 @@
...
+        f"\n"
+        f"    for subpath in await mypath.{fn.__name__}():\n"
+        f"        ...\n"
+        f"\n"
+        f".. note::\n"
+        f"\n"
+        f"    The iterator is loaded into memory immediately during the initial\n"
+        f"    call, (see `issue #501\n"
+        f"    <https://github.com/python-trio/trio/issues/501>`__ for discussion).\n"
+    )
```
A5rocks, line 95: Maybe keep how it was before (multiline string) and then `textwrap.dedent`?

src/trio/_path.py
```diff
@@ -32,219 +22,102 @@
...
-                wrapper = iter_wrapper_factory(cls, attr_name)
-                assert isinstance(wrapper, types.FunctionType)
-                wrapper.__signature__ = inspect.signature(attr)
-                setattr(cls, attr_name, wrapper)
-
-
-@final
-class Path(metaclass=AsyncAutoWrapperType):
+class Path(pathlib.PurePath):
     """A :class:`pathlib.Path` wrapper that executes blocking methods in
```
A5rocks, line 100:
```suggestion
    """A :class:`pathlib.PurePath` wrapper that executes blocking methods in
```

I'm not too familiar with `pathlib`, does this have implications for back compat? Why can't this be `pathlib.Path`?

src/trio/_path.py
```diff
@@ -32,219 +22,102 @@
...
-# re-wrap return value from methods that return new instances of pathlib.Path
-def rewrap_path(value: T) -> T | Path:
-    if isinstance(value, pathlib.Path):
-        return Path(value)
-    else:
-        return value
 
+def _wraps_async(
+    wrapped: Callable[..., Any]
+) -> Callable[[Callable[P, T]], Callable[P, Awaitable[T]]]:
```
A5rocks, line 35: FYI, `pyright` tracks `P` through and attaches the docstring from the initial point to where it ends up, so that's a potential solution for getting `pyright` to recognize the docstrings.

src/trio/_path.py
```diff
@@ -32,219 +22,102 @@
...
-
-    def __dir__(self) -> list[str]:
-        return [*super().__dir__(), *self._forward]
+    __slots__ = ()
 
-    def __repr__(self) -> str:
-        return f"trio.Path({str(self)!r})"
+    def __new__(cls: type[PathT], *args: str | os.PathLike[str]) -> PathT:
+        if cls is Path:
+            cls = WindowsPath if os.name == "nt" else PosixPath  # type: ignore[assignment]
```
A5rocks, line 109: TBH if you just duplicate the return this shouldn't need the type ignore? I think

src/trio/_path.py
```diff
@@ -32,219 +22,102 @@
...
-                return rewrap_path(value)
-            raise AttributeError(name)
-
-    def __dir__(self) -> list[str]:
-        return [*super().__dir__(), *self._forward]
+    __slots__ = ()
 
-    def __repr__(self) -> str:
-        return f"trio.Path({str(self)!r})"
+    def __new__(cls: type[PathT], *args: str | os.PathLike[str]) -> PathT:
```
A5rocks, line 107: Surprised `Self` doesn't work here.

src/trio/_path.py
```diff
@@ -314,172 +187,78 @@ async def open(  # type: ignore[misc]  # Any usage matches builtins.open().
...
+class WindowsPath(Path, pathlib.PureWindowsPath):
+    """A :class:`pathlib.WindowsPath` wrapper that executes blocking methods in
+    :meth:`trio.to_thread.run_sync`.
+
+    """
+
+    __slots__ = ()
+
+
+final(Path)
```
A5rocks, line 264: Putting `final` down here prevents type checkers from seeing that `Path` is final. Maybe make an intermediate class? I'm not sure if `Path = final(BasePath)` works but something should work.

docs/source/reference-io.rst
etianen: If you look at the `pathlib` module, that's exactly what they do there. Windows and posix have very different ways of expressing a path (backslashes vs forwardslashes etc). This is making the `trio` path behavior just like the stdlib.

Practically, being able to make `trio.Path` a subclass of `PurePath` means it interops with all other path-derived things, and dramatically simplifies the implementation. However, you can't directly subclass `PurePath`... you have to subclass `PureWindowsPath` or `PurePosixPath`. And that means you need to have to concrete `PosixPath` and `WindowsPath` classes.

src/trio/_path.py
etianen: `Path` is spammed all over the place, as it's an immutable class and the everyday APIs create new ones all the time. Something like: `foo / bar / baz` creates a new `Path` and then throws it away at each intermediate `/` time.

So it's best to optimize this for cheap creation of new `trio.Path` objects, which is what this PR does. The new `trio.Path` is slotted, and does not wrap another class, so I'd expect non-blocking operations on the new `trio.Path` to be quite a bit faster.

The actual blocking operations are gonna be slow anyway. Compare to creating a `partial` and passing it to a thread via a thread-safe channel and then waiting for the result using an event loop, I'd expect it to be trivial.

If you'd like me to run some benchmarks, I'd be happy to. But my instinct here is that non-blocking operations will be faster, and wrapped blocking operations will functionally identical.

src/trio/_path.py
etianen: I did have a fiddle with using the textwrap tools. They do have a runtime overhead, and I found them finickity giving inconsistent results depending on the indentation of the first line. If you feel strongly, I can have another go.

src/trio/_path.py
etianen: It's not a `PurePath` wrapper though. It's a `PurePath` subclass *and* a `Path` wrapper.

Subclassing `Path` is impossible because the blocking methods (e.g. `absolute`) are not async, and replacing sync methods with async is forbidden by type checkers.

Now you mention it though, maybe we should stop documenting that it's a wrapper? That's an implementation detail. This is just an async `Path` implementation that slots into the `pathlib` class hierarchy as one would expect.

src/trio/_path.py
etianen: I don't understand. What's `pyright` doing with docstrings and paramspecs? And what does that have to do with inconsistent types?

If you have a solution to the `pyright` problem I'd love to know more! ❤️

src/trio/_path.py
etianen: It does! It just doesn't gain from a type-checking point of view, and I was avoiding using multiple `Self` types in this module. If you'd rather, I can replace the use of `PathT` with `Self` where possible. I have no strong opinions here.

src/trio/_path.py
etianen: Ah, nope, you just get a different type error.

``` py
def __new__(cls: type[PathT], *args: str | os.PathLike[str]) -> PathT:
        if cls is Path:
            return super().__new__(WindowsPath if os.name == "nt" else PosixPath, *args)
        return super().__new__(cls, *args)
```

error: Incompatible return value type (got "Path", expected "PathT")

And the slightly modified:

```py
def __new__(cls: type[Self], *args: str | os.PathLike[str]) -> Self:
        if cls is Path:
            return super().__new__(WindowsPath if os.name == "nt" else PosixPath, *args)
        return super().__new__(cls, *args)
```

error: Incompatible return value type (got "Path", expected "Self")

I'm not worried about the `type: ignore` here. I'm a big fan of static type checking, but sometimes you've gotta break it in hidden implementations while maintaining the static typing on the public-facing part.

etianen:
Thanks for the feedback @A5rocks ! 🙇 

I'm seeing two outstanding issues here, both related to the `trio` linting requirements:

1. `trio.Path` cannot be made `final`
2. `trio.Path` cannot be made _unambiguous_ according to `pyright` without adding a bunch of boilerplate.

Hopefully excluding `trio.Path` from the checks for (1) is not a blocker. 🙏 

Hopefully there is a way to exclude `trio.Path` from the checks for (2), since excluding it is the right answer here. The types are not ambiguous AFAIK.

In terms of it being a slightly funny API, I do agree, but it's the funny API from `pathlib`. 🤷 And if the purpose of `trio.Path` is to be an an async `pathlib.Path`, then being able to interop with and behave like the stdlib is pretty neat, IMO.

src/trio/_path.py
jakkdl: I have a weak preference for using `Self` where possible and only using `PathT` where necessary :)

src/trio/_path.py
jakkdl: pyright --verifytypes checks for docstrings, and while it doesn't highlight with an error message for each one you can see it in the summary at the end:
```
  Symbols without documentation:
    Functions without docstring: 86
    Functions without default param: 0
    Classes without docstring: 21
```
but that was broken before (see #2910) so not a hard requirement to get it passing. Although #2910 says 133 functions are missing docstrings, and here it's 86, so you maybe you did fix the ones from path? Or it's a matter of `--ignoreexternal`??

jakkdl:
Spent all day doing other stuff instead of doing #2910, but if the verifytypes test becomes the only blocker we can ignore it and/or temporarily make it not fail CI if it fails.

src/trio/_path.py
etianen: It is done! 🙇

src/trio/_path.py
etianen: I suspect that inheriting a bunch of methods from `PurePath` rather than wrapping them has indeed fixed the docstrings! 💪

etianen:
I've disabled the requirement for `trio.Path` to be final. Hopefully that's fine. 🙏 

As a bonus, I've added explicit support for subclassing `trio.Path`, making this a feature! @TeamSpen210 mentioned subclassing in [this comment](https://github.com/python-trio/trio/issues/2944#issuecomment-1948068833). Well, now people can subclass `trio.Path` to wrap `pathlib.Path` subclasses.

As a bonus, this speeds up async methods slightly by not performing an `os.platform` check ever time, since `trio.WindowsPath` and `trio.PosixPath` explicitly reference `pathlib.WindowsPath` and `pathlib.PosixPath`

etianen:
So I _think_ the only blocker here is `--verifytypes`. There's still a few open conversations in the air about minor issues, so it would be good to get some other opinions to get them resolved.

src/trio/_path.py
A5rocks: No benchmarks necessary. I was more thinking about any possible exceptions being raised!

src/trio/_path.py
A5rocks: I'm not too familiar with `pathlib` things (hence these slightly strange comments), but if you think we can, then "An async `pathlib.Path`" but more fleshed out might sound better.

src/trio/_path.py
A5rocks: Oh I was thinking something like:

```py
class PathBase:  # not exported
  # current `Path`

@final
class Path(PathBase):
  # just forward everything to `PathBase`.
  __slots__ = ()

@final
class WindowsPath(PathBase):
  __slots__ = ()
  
  ...
  
```

It's a bit convoluted and just making sure people get earlier warnings about subclassing `trio.Path` might not be worth this.

etianen: I think I've turned this into a feature now (see latest commit and comment).

src/trio/_path.py
etianen: Check out my latest commit and comment, which makes it being a wrapper actually part of the public API, and enables explicit subclassing support to match what's coming in Python 3.13.

I can take or leave my last commit. On average, I think it's nice to support subclassing to match what the stdlib does, but I'm also cool to make it an implementation detail and tweak the docstring. I'll go with the majority here!

TeamSpen210:
In regards to Pyright's checks, the intent I think is to warn about any public API that would require use of type inference to determine the type, meaning that it's possible that a less powerful/intelligent checker might give different results. In this case we deliberately do want to use inference to prevent having to manually write out the signatures, so ignoring it would be fine.

etianen:
@TeamSpen210 100% The only thing preventing me is I can't find *any* way to selectively disable it just for this file!

src/trio/_path.py
```diff
@@ -32,219 +22,110 @@
...
+        f".. code:: python\n"
+        f"\n"
+        f"    for subpath in await mypath.{fn.__name__}():\n"
+        f"        ...\n"
+        f"\n"
+        f".. note::\n"
+        f"\n"
+        f"    The iterator is loaded into memory immediately during the initial\n"
+        f"    call, (see `issue #501\n"
+        f"    <https://github.com/python-trio/trio/issues/501>`__ for discussion).\n"
```
A5rocks, line 94:
```suggestion
        f"    The iterator is loaded into memory immediately during the initial\n"
        f"    call. (see `issue #501\n"
        f"    <https://github.com/python-trio/trio/issues/501>`__ for discussion)\n"
```

I think the comma is weird grammatically. Nitpick though

src/trio/_path.py
```diff
@@ -314,172 +195,79 @@ async def open(  # type: ignore[misc]  # Any usage matches builtins.open().
...
-# :meth:~pathlib.Path.absolute, which does not exist. Removing this makes more
-# sense than inventing our own special docstring for this.
-del Path.absolute.__doc__
-
-# TODO: This is likely not supported by all the static tools out there, see discussion in
-# https://github.com/python-trio/trio/pull/2631#discussion_r1185612528
-os.PathLike.register(Path)
+    ) -> AsyncIOWrapper[IO[Any]]: ...
+
+    @_wraps_async(pathlib.Path.open)  # type: ignore[misc]  # Overload return mismatch.
```
A5rocks, line 200: If this is what I think it is (mismatch because we don't have overloads for `"r"` vs `"w"` vs whatever), then IMO this should have a TODO. I realize you didn't write this type ignore and just copied it over, but yeah.

src/trio/_tests/test_path.py
A5rocks, line 1: Can we add a basic test to make sure both `WindowsPath` and `PosixPath` are actually doing things asynchronously? (on their respective platforms, of course!) Would be embarassing if they don't.

We don't need the whole gamut later in this file, just something like `test_open_file_can_open_path` for both probably.

src/trio/_tests/type_tests/path.py
```diff
@@ -7,7 +7,7 @@
 from typing import IO, Any, BinaryIO, List, Tuple
 
 import trio
-from trio._path import _AsyncIOWrapper  # pyright: ignore[reportPrivateUsage]
+from trio._path import AsyncIOWrapper  # pyright: ignore[reportPrivateUsage]
```
A5rocks, line 10: pyright: ignore should be unnecessary now

src/trio/_tests/type_tests/path.py
A5rocks, line 1: I know `WindowsPath` and `PosixPath` essentially function as custom subclasses of `trio.Path` rn, but I think it would be nice to have an example subclass of `trio.Path` in here to make sure it works typing-wise even if we change the impl details of `WindowsPath`/`PosixPath`.

A5rocks (changes requested):
I forgot: could you add a newsfragment?

Also a couple comments, ranging from nitpicks to asking for a test or two

src/trio/_path.py
CoolCat467: Oh yea, I was thinking about this when I reviewed the other day, why are all of the strings in this block f-strings if only a few actually need it?

A5rocks: the consistent `f` prefixes help text alignment. `textwrap.dedent` would probably be nicer but it's unnecessary and can have its quirks (w/r/t how it decides how much to dedent)

A5rocks:
> The only solution I can see is the solution in https://github.com/python-trio/trio/pull/2910. Do we keep this open and wait for that to merge, or do we temporarily disable the check?

I'm sure specifically filtering pyright errors should be a pretty simple thing to do; let's wait a few days and see if that happens. If it doesn't, then temporarily disabling the check sounds good. (I'm hoping we can get a release out in ~a week, so that this and a bunch of changes can go out)

TeamSpen210:
Perhaps we should open an issue for Pyright. Thinking about it, it seems like a bug that `verifytypes` doesn't pick up on inference being required for the decorator, that's inconsistent.

src/trio/_path.py
etianen: > I think the comma is weird grammatically. Nitpick though

I think that creates a different grammatical problem (capitalization of `(see`.). So I checked with the _authority_ that is grammarly and apparently just removing the comma is correct.

https://www.grammarly.com/blog/comma-before-parenthesis-or-after/

So I did that.

src/trio/_path.py
TeamSpen210: This is probably actually because we return an `AsyncIOWrapper`, not the sync file.

etianen: It's neither! It's `Type of decorated function contains type "Any" ("Callable[[Path, str, int, Optional[str], Optional[str], Optional[str]], Coroutine[Any, Any, AsyncIOWrapper[IO[Any]]]]")`

For some reason, using any in a decorated function is naughty. It's not naughty if the function isn't decorated though. 🤷

src/trio/_tests/type_tests/path.py
etianen: Having taken a look at what's involved in creating a subclass, I've decided to take out mention of this being a wrapper, or being subclassable. It'll be a support headache, and the actual use-cases for creating more async path subclasses outside of trio are not worth it, IMO.

etianen:
Hopefully that's all points addressed! Some directly, others _laterally_.

A5rocks (approved):
Looks good, I like that this decreases the total amount of code!

(hopefully someone else can also give this a lookover because my passes might have missed something)

src/trio/_tests/type_tests/path.py
```diff
@@ -7,7 +7,7 @@
 from typing import IO, Any, BinaryIO, List, Tuple
 
 import trio
-from trio._path import _AsyncIOWrapper  # pyright: ignore[reportPrivateUsage]
+from trio._path import AsyncIOWrapper
```
CoolCat467, line 10: Why are we not doing `from trio._file_io import AsyncIOWrapper` here? And more interesting, why is this import not being marked as using a private import?

CoolCat467 (changes requested):
Looks pretty good other than one thing I noticed

src/trio/_tests/type_tests/path.py
etianen: No good reason, but I was trying to keep this PR somewhat self-contained rather than blarting changes anywhere I spotted something odd. 😅 

But you're quite right, this should be an import from a canonical location. Now it is. 💪 

Why `pyright` was okay with this I don't really know. I've had an argument with `pyright` recently and we're not talking to each other any more. 🙉

CoolCat467 (approved):
Very nice!

A5rocks:
pyright changes are merged, you should be able to `python src/trio/_tests/check_type_completeness.py --overwrite` and that should be it!

jakkdl:
I added some logic to ignore everything instead of listing all the errors in the json. I feel like `has_docstring_at_runtime` should've been able to pick up the docstrings, but I haven't bothered figuring out why it doesn't.

etianen:
It's in! Thanks all!

jakkdl:
Thank you! And sorry for the delay with finishing up #2910 :)

A5rocks:
fyi this went out in 0.25.0!

PR #2964, opened 2024-02-27
Remove incorrect `str` possibility from HostnameResolver.getaddrinfo

tjstum:
`_socket.getaddrinfo` takes care of encoding all passed in `str` host arguments before passing it to a custom resolver. Additionally, the docs say that a resolver will always receive IDNA-encoded bytes.

However, the type hints of `HostnameResolver.getaddrinfo` imply that a resolver may need to contend with a `str`

Remove the extraneous appearance

### DIFF
--- a/src/trio/_abc.py
+++ b/src/trio/_abc.py
@@ -168,7 +168,7 @@ class HostnameResolver(metaclass=ABCMeta):
     @abstractmethod
     async def getaddrinfo(
         self,
-        host: bytes | str | None,
+        host: bytes | None,
         port: bytes | str | int | None,
         family: int = 0,
         type: int = 0,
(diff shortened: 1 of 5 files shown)


A5rocks (approved):
I agree with this assessment. For others who want to double check:

https://github.com/python-trio/trio/blob/accaae421399ebb37a8411db0c06a5d0b5631f72/src/trio/_socket.py#L223-L230

... I was going to ask about putting `@typing.override` on these implementations of `HostnameResolver` but turns out that doesn't warn on the signatures being different! So nevermind.

mikenerone (commented):
Pretty straightforward, seems correct.

mikenerone (approved):

CoolCat467 (approved):

jakkdl:
thanks! Surprised there haven't been more typing errors that snuck through.

CoolCat467:
Type checker must be working well!

tjstum:
Thanks everyone!

I realize I forgot a changelog entry. Does this warrant one, or is it too small to care?

jakkdl:
> Thanks everyone!
> 
> I realize I forgot a changelog entry. Does this warrant one, or is it too small to care?

I guess it technically is a public-facing change, but I think it's probably too small to care.

Issue #2546, opened 2023-01-27
"Google Docstring Format" link goes to 404


BruceEckel:
Under this header: https://trio.readthedocs.io/en/stable/contributing.html#documentation
At the beginning of the 2nd paragraph: "For docstrings, we use [the Google docstring format](https://www.sphinx-doc.org/en/3.x/usage/extensions/example_google.html#example-google-style-python-docstrings)."

The link goes to a 404 page.

A5rocks:
It looks like v3 of the Sphinx docs was pulled? This link certainly works: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google-style-python-docstrings

I think we should probably check the old link 404ing is intentional. In the meantime, updating the docs should be simple.

kachida:
Hi @A5rocks , I've created a pull request (PR#2967) for this issue. Could you please take a look and review it

https://github.com/python-trio/trio/pull/2967

PR #2967, opened 2024-03-03
fix: updated the google docstring url

kachida:
Updated Google Docstring Format link since currently it points to a 404 page.

Fixes https://github.com/python-trio/trio/issues/2546


CoolCat467 (approved):
Looks good, thanks!

jakkdl:
Though maybe there might be a `stable` version of the docs or something, but seems like `master` is the only option so nothing to consider. Thanks!

PR #2969, opened 2024-03-11
Enable ruff's Perflint rule.

CoolCat467:
This pull request enables ruff's [perflint](https://docs.astral.sh/ruff/rules/#perflint-perf) rule and handles the associated fixes or ignores.


src/trio/_core/_instrumentation.py
```diff
@@ -98,7 +98,7 @@ def call(self, hookname: str, *args: Any) -> None:
         for instrument in list(self[hookname]):
             try:
                 getattr(instrument, hookname)(*args)
-            except BaseException:
+            except BaseException:  # noqa: PERF203  # try-except in loop
```
A5rocks, line 101: I'm surprised ruff doesn't like this. Like yeah, if we have a bunch of things to do and some might fail, then the try except belongs *in* the loop.

src/trio/_subprocess.py
```diff
@@ -723,7 +723,7 @@ async def read_output(
     ) -> None:
         async with stream:
             async for chunk in stream:
-                chunks.append(chunk)
+                chunks.append(chunk)  # noqa: PERF402
```
A5rocks, line 726: I feel like `PERF402` on variable passed in is incorrect and probably is a bug...

src/trio/_tests/test_socket.py
```diff
@@ -324,7 +324,9 @@ async def test_sniff_sockopts() -> None:
     sockets = []
     for family in [AF_INET, AF_INET6]:
         for type_ in [SOCK_DGRAM, SOCK_STREAM]:
-            sockets.append(stdlib_socket.socket(family, type_))
+            sockets.append(  # noqa: PERF401  # Makes sense hear, clarity.
```
A5rocks, line 327:
```suggestion
            sockets.append(  # noqa: PERF401  # Makes sense here, clarity.
```

src/trio/_core/_instrumentation.py
TeamSpen210: It's a rule we might want to turn off, it just warns whenever a try-except is found inside a loop. The recommendation is to try to refactor so that it's moved around the loop, but that's not always possible. It's also not applicable in 3.11+ due to "zero cost" exception handling.

src/trio/_tests/test_socket.py
```diff
@@ -324,7 +324,9 @@ async def test_sniff_sockopts() -> None:
     sockets = []
     for family in [AF_INET, AF_INET6]:
         for type_ in [SOCK_DGRAM, SOCK_STREAM]:
-            sockets.append(stdlib_socket.socket(family, type_))
+            sockets.append(  # noqa: PERF401  # Makes sense here, clarity.
+                stdlib_socket.socket(family, type_)
+            )
```
TeamSpen210, line 329: We could do:
```py
sockets = [
    stdlib_socket.socket(family, type_)
    for family in [AF_INET, AF_INET6]
    for type_ in [SOCK_DGRAM, SOCK_STREAM]
]
```
Or actually, since it's just immediately used, just change this to use `itertools.product()`, make the stdlib socket call the first line.

src/trio/_core/_instrumentation.py
jakkdl: https://docs.astral.sh/ruff/rules/try-except-in-loop/ does say
> This rule is only enforced for Python versions prior to 3.11, which introduced "zero cost" exception handling.

But I suppose we have ruff set to minimum supported python version. In either case the rule documentation also states
> Note that, as with all perflint rules, this is only intended as a micro-optimization, and will have a negligible impact on performance in most cases.

so we can be quite liberal with disabling rules.

src/trio/_tests/test_socket.py
CoolCat467: Thanks, added in 550518a

src/trio/_core/_instrumentation.py
CoolCat467: Disabled `PERF203` in e38b663

src/trio/_core/_tests/test_run.py
```diff
@@ -2044,9 +2044,7 @@ async def __anext__(self) -> list[int]:
 
             return items
 
-    result: list[list[int]] = []
-    async for vals in async_zip(it(4), it(2)):
-        result.append(vals)
+    result: list[list[int]] = [vals async for vals in async_zip(it(4), it(2))]
```
TeamSpen210, line 2047: Thinking about it, actually it'd be good to avoid comprehensions in test code, if they're the "main loop". If an error occurs, a comprehension would just be discarded while an explicit loop would leave the partial results behind to be printed out with other local variables.

CoolCat467: Reverted change in 17afaab

CoolCat467:
Updating branch to see if any changes with new ruff versions

CoolCat467:
pre-commit.ci autofix

src/trio/_subprocess.py
CoolCat467: Created https://github.com/astral-sh/ruff/issues/11316 for this

src/trio/_subprocess.py
```diff
@@ -726,7 +726,7 @@ async def read_output(
     ) -> None:
         async with stream:
             async for chunk in stream:
-                chunks.append(chunk)
+                chunks.append(chunk)  # noqa: PERF401
```
A5rocks, line 729: If we care about the noqa this could probably be `chunks.extend([chunk async for chunk in stream])` though if anything that's *slower* (I think?).

Maybe we should report this as a bug to ruff.

A5rocks (approved):
Just one thought

CoolCat467:
Given that https://github.com/astral-sh/ruff/issues/11316 is still open, I am thinking of merging this soon unless anyone has any other comments.

PR #2972, opened 2024-03-13
Add a trio repl

### LINKED ISSUE #1360: Provide "python -m trio" to drop into trio based REPL
Just like "python -m asyncio" so we can use things like "await foo" straight out of the box.

clint-lawrence:
Closes #1360 

This adds a repl that can be launched with `python -m trio`, similar to asyncio and curio.

I still need to add a news entry and update the documentation, but I was hoping for some feedback if this is likely to be accepted before doing that. Also, if there is any suggestions about where it would be best to document this that would be helpful.


oremanj:
I am in favor of accepting something like this. As the github bot has noted, please add type hints.

Does Ctrl+C work reliably to cancel the running thing? Does Ctrl+C at the REPL prompt just give you another prompt rather than crashing the REPL? I would say that both of those are important usability features, and you might have them without trying but it's worth a test.

A5rocks:
Also would it be worth it to have a nursery created by default in a `nursery` local so you can start something in the background for the REPL?

CoolCat467:
I like this pull request a lot more than #2407

clint-lawrence:
> I am in favor of accepting something like this. As the github bot has noted, please add type hints.

Yes, I have start on that. Just a few issues I need to come up with a fix for.

> Does Ctrl+C work reliably to cancel the running thing? Does Ctrl+C at the REPL prompt just give you another prompt rather than crashing the REPL? I would say that both of those are important usability features, and you might have them without trying but it's worth a test.

As far as I can tell, yes, Ctrl+C handling is robust and yes, it keeps you in the repl. The robustness comes from trio to begin with!

> Also would it be worth it to have a nursery created by default in a `nursery` local so you can start something in the background for the REPL?

This is actually what initially motivated me to start down this path, but I hadn't worked out how I was going to do it. So thanks for the suggestion. I will have a play with that idea after getting the basics sorted out.

Thanks for the positive feedback. It will probably take a day or two to tidy up the type hinting

jakkdl:
> I like this pull request a lot more than #2407

to save me+others some time comparing them if you already have - why? What's the difference between their approach?

TeamSpen210:
For the nursery, perhaps what it could do is that when control-C is pressed, first stop any foreground code that's running as normal. But if pressed again with nothing, cancel the nursery, close it, then replace it with a fresh nursery that's open. If a user causes the nursery to be cancelled, also replace it?

clint-lawrence:
> > I like this pull request a lot more than #2407
> 
> to save me+others some time comparing them if you already have - why? What's the difference between their approach?

I didn't see #2407 before creating this. I think the main implementation difference is this PR keeps the event loop running in a separate thread to the blocking calls to stdin. The other PR will block the event loop each time the repl is waiting for input on stdin. At the moment, that probably doesn't make much difference, but I have my eye on being able to run tasks in the background. I've also got reasonable test coverage here.

@CoolCat467  might have had other things in mind?

CoolCat467:
Very similar approaches, both based on tweaking asyncio's solution. The other one uses a nursery it sends tasks to with `nursery.start_soon`, adds change log message, and adds docs reference to the fact that this feature exists now, but is missing tests and I believe it doesn't work correctly because of among other things, main nursery never runs because `console.interact` is blocking trio from ever running.
Prior to me stepping in a few months ago after finding it when looking through currently open pull requests, it had been stale for quite some time and didn't have type annotations. I like this new proposed approach a lot more and suggest stealing the documentation and change log additions the other one made.

src/trio/_repl.py
```diff
@@ -10,14 +14,19 @@
...
         self.compile.compiler.flags |= ast.PyCF_ALLOW_TOP_LEVEL_AWAIT
 
-    def runcode(self, code):
-        async def _runcode_in_trio():
+    def runcode(self, code: types.CodeType) -> None:
+        async def _runcode_in_trio() -> BaseException | None:
+            # code.InteractiveInterpreter defines locals as Mapping[str, Any]
+            # However FunctionType expects a dict. We know our copy of
+            # locals will be a dict due to the annotation on repl_locals in __init__
+            # so the cast is safe.
```
A5rocks, line 26: You might be able to put a variable declaration and avoid this? Not sure how to phrase that, but like:

```py
class TrioInteractiveConsole(InteractiveConsole):
    locals: dict[str, object] | None
```

clint-lawrence: Yes, that worked. Must nicer than having to cast

src/trio/_repl.py
```diff
@@ -0,0 +1,85 @@
...
+        warnings.filterwarnings(
+            "ignore",
+            message=r"^coroutine .* was never awaited$",
+            category=RuntimeWarning,
+        )
+
+
+def main(original_locals: dict[str, object]) -> None:
+    with contextlib.suppress(ImportError):
+        import readline  # noqa: F401
```
CoolCat467, line 71: Curious what this import is for

mikenerone: Just importing `readline` enables readline support in some builtin functions like `input()`. This is to enable it if installed.

clint-lawrence: readline provides the basic editing function on the repl, if available. e.g. up arrow to access history etc.

CoolCat467 (approved):
Looks great!

src/trio/_repl.py
```diff
@@ -0,0 +1,100 @@
...
+        if maybe_exc_or_excgroup is not None:
+            # maybe_exc_or_excgroup is an exception, or an exception group.
+            # If it is SystemExit or if the exception group contains
+            # a SystemExit, quit the repl. Otherwise, print the traceback.
+            if isinstance(maybe_exc_or_excgroup, SystemExit):
+                raise SystemExit()
+            elif isinstance(maybe_exc_or_excgroup, BaseExceptionGroup):
+                for exc in maybe_exc_or_excgroup.exceptions:
+                    if isinstance(exc, SystemExit):
+                        raise SystemExit
```
TeamSpen210, line 54: I think this will need to use `ExceptionGroup.subgroup()`, since it's possible the `SystemExit` can be nested arbitrarily deep in exception groups. It'd also be good to raise the original, not create a fresh exception - you can pass an exit code or message to `sys.exit()`, which is stored in `SystemExit` and used as the process return code.

clint-lawrence: Oh, I did not realise exception groups could nest. I'll have a look.

TeamSpen210 (approved):

clint-lawrence:
I think all the previous feedback has been addressed. Please let me know if you need anything else for this to be merged.

newsfragments/2972.feature.rst
```diff
@@ -0,0 +1,16 @@
+Added an interactive interpreter ``python -m trio``.
+
+This makes it easier to try things and experiment with trio in the a python repl.
```
A5rocks, line 3:
```suggestion
This makes it easier to try things and experiment with trio in the a Python repl.
```

src/trio/__main__.py
```diff
@@ -0,0 +1,3 @@
+from trio._repl import main
+
+main(locals())
```
A5rocks, line 3:
```suggestion
variables = locals().copy()
del variables["main"]
main(variables)
```

This is of course only necessary if we change `main` to allow other locals to work. Also probably worth checking that's the only surprising entry!

src/trio/_repl.py
```diff
@@ -0,0 +1,114 @@
...
+    repl_locals: dict[str, object] = {"trio": trio}
+    for key in {
+        "__name__",
+        "__package__",
+        "__loader__",
+        "__spec__",
+        "__builtins__",
+        "__file__",
+    }:
+        repl_locals[key] = original_locals[key]
```
A5rocks, line 111: In conjunction with my above suggestion:
```suggestion
    repl_locals: dict[str, object] = {"trio": trio, **original_locals}
```

I'm not sure if this is useful. The single use case I can think of is allowing someone to run a trio repl with custom local vars but this isn't reexported anyways. Might be worth changing just to stop having a list of keys to keep?

src/trio/_repl.py
```diff
@@ -0,0 +1,114 @@
...
+            try:
+                coro = func()
+            except BaseException as e:
+                return e
+
+            if inspect.iscoroutine(coro):
+                try:
+                    await coro
+                except BaseException as e:
+                    return e
```
A5rocks, line 51: Can some of this exception stuff be replaced with stuff from `outcome`? I don't see it providing much technical benefit but it would decrease some code duplication.

src/trio/_repl.py
```diff
@@ -0,0 +1,114 @@
...
+            # a SystemExit, quit the repl. Otherwise, print the traceback.
+            if isinstance(maybe_exc_or_excgroup, SystemExit):
+                raise maybe_exc_or_excgroup
+            elif isinstance(maybe_exc_or_excgroup, BaseExceptionGroup):
+                sys_exit_exc = maybe_exc_or_excgroup.subgroup(SystemExit)
+                if sys_exit_exc:
+                    # There is a SystemExit exception, but it might be nested
+                    # If there are more than one SystemExit exception in
+                    # the group, this will only find and re-raise the first.
+                    raise next(_flatten_exception_group(sys_exit_exc))
```
A5rocks, line 68: It might be worth printing all non-system exit exceptions in the exception group?

A5rocks (commented):
I have a few more comments, if you feel up to it. Keep in mind most of them don't improve anything technically.

src/trio/_repl.py
clint-lawrence: I might not be understanding how outcomes works, but I don't see how it would simplify this. Wouldn't it just end up  calling capture method and then immediately unwrapping?
```python
try:
  coro = outcome.capture(func)
  coro.unwrap()
except BaseException as e:
  return e
```

The traceback printing is consistent with the standard python repl and the `python -m asyncio` so, I'm leaning towards keeping it as is.

src/trio/_repl.py
clint-lawrence: I like this idea. I just need to have a think how to do that in a way that isn't too messy.

src/trio/_repl.py
clint-lawrence: I prefer the explicit list. There is less risk of surprise. It taken from the similar code in the `python -m asyncio` implementation and I don't think it has changed there over many python versions, so don't I expect much churn.

src/trio/_repl.py
clint-lawrence: I think I should remove that elif entirely. Then the exception group will be printed and the repl won't exit. Ideally, there shouldn't be a call to `sys.exit()` buried down the call stack that would cause this, but if there is, it is unlikely the user intended to exit the interpreter.

Trying to get me head around this, there are the following cases...
1. User types `exit()`, intending to quit the repl.
2. Some code called from within the repl calls `sys.exit()` or otherwise raises `SystemExit()`
3. Some code raises an exception group which includes a `SystemExit()`. Perhaps a call to `sys.exit()` inside a task.

For cases 1 & 2, this repl, standard python and asyncio python all just exit with no traceback. So if there is a call to `sys.exit()` buried down the call stack somewhere, the repl just quits. It's not easy to tell the different between 1 & 2 so this is reasonable even if it would be better to show the traceback in the case of 2.

Case 3 is similar to case 2, it's just that there is most likely some async code involved. We don't have the problem case 2 has, having to distinguish from a "real" command to exit the repl. So we stop treating this as a special case, then the code is simpler and the user gets better information.

In the `python -m asyncio` repl, a `SystemExit()` inside a `BaseExceptionGroup` doesn't quit the interpreter. Instead, the exception group gets printed like any other exception. The standard python interpreter is the same. So the proposed change would be in line with the behavior there.

Does that sound reasonable?

clint-lawrence: I've made this change and pushed. Let me know what you think and I can keep it or roll back if needed.

src/trio/_repl.py
A5rocks: Well specifically I was thinking directly returning the Value/Error (and an `if` on whether to use `capture` or `acapture`? Which would disallow leaving out the `await` when calling an async function but that's already trio style. I'll make this a separate comment when I get back to looking at this). If the trace backs are the same then the only benefit would be a few less lines, which probably isn't worth the work.

src/trio/_repl.py
A5rocks: It feels a bit weird to me to have this different behavior, but now that I think of it it's probably nicer. If I write `exit()` inside a nursery I probably don't exist my whole repl to exit!

src/trio/_repl.py
A5rocks: Makes sense

src/trio/_repl.py
clint-lawrence: I came up with this. It's better in some ways, but at the same time bit strange.
```python
    def runcode(self, code: types.CodeType) -> None:
        async def _runcode_in_trio():
            func = types.FunctionType(code, self.locals)
            result = outcome.capture(func)
            
            if isinstance(result, outcome.Error):
                return result
            
            coro = result.unwrap()
            if inspect.iscoroutine(coro):
                return await outcome.acapture(lambda: coro)
        
            return outcome.Value(coro)

        try:
            value = trio.from_thread.run(_runcode_in_trio).unwrap()
        except SystemExit:
            # If it is SystemExit quit the repl. Otherwise, print the
            # traceback.
            # There could be a SystemExit inside a BaseExceptionGroup. If
            # that happens, it probably isn't the user trying to quit the
            # repl, but an error in the code. So we print the exception
            # and stay in the repl.
            raise
        except BaseException:
            self.showtraceback()
```
I think it is awkward because there is a single step for most expressions, but any repl input that is an `await` needs the second step of evaluation.

A5rocks:
This is probably here for good reason in fact I get a feeling of deja-vu so we've probably already discussed this. If there's previous discussion about this, just link that and I should be satisfied.

However, I don't really like the behavior of unconditionally `await`ing the result of the thing in the prompt. Is this really necessary? Or can we use some `inspect` function to find that there's an await?

I will eventually get around to poking at this and likely rediscover any issues with not unconditionally `await`ing the result, but if there's no issue then I personally would really prefer something where directly calling `trio.sleep(5)` doesn't automatically `await` unless you add an `await` yourself.

clint-lawrence:
> However, I don't really like the behavior of unconditionally `await`ing the result of the thing in the prompt. Is this really necessary? Or can we use some `inspect` function to find that there's an await?

This isn't automatically awaiting the input. For example below the first `trio.sleep(1)` does not get awaited or actually sleep. The second `await trio.sleep(1)` blocks for 1 second, before the repl prompts for the next input.

```
trio REPL 3.12.2 (v3.12.2:6abddd9f6a, Feb  6 2024, 17:02:06) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Use "await" directly instead of "trio.run()".
>>> import trio
>>> trio.sleep(1)
<coroutine object sleep at 0x10f3eda40>
>>> await trio.sleep(1)
>>>
```

A5rocks:
Oh! That's what I get for assuming things from the code, rather than testing things. Sorry!

src/trio/_repl.py
A5rocks: I finally tried poking at this a bit, ultimately what I was thinking about is:

```py
    def runcode(self, code: types.CodeType) -> None:
        async def _runcode_in_trio():
            func = types.FunctionType(code, self.locals)
            if inspect.iscoroutinefunction(func):
              return outcome.acapture(func)
            else:
              return outcome.capture(func)

        try:
            value = trio.from_thread.run(_runcode_in_trio).unwrap()
        except SystemExit:
            # If it is SystemExit quit the repl. Otherwise, print the
            # traceback.
            # There could be a SystemExit inside a BaseExceptionGroup. If
            # that happens, it probably isn't the user trying to quit the
            # repl, but an error in the code. So we print the exception
            # and stay in the repl.
            raise
        except BaseException:
            self.showtraceback()
```

probably doesn't work cause I haven't tested exactly that, but `inspect.iscoroutinefunction` seems to do the right thing (doesn't trigger for `trio.sleep(1)` or `async def x(): ...`, but triggers for `await trio.sleep(1)`)

A5rocks (approved):
Minor comment just to clarify what I meant previously but doing anything else is optional.

src/trio/_repl.py
clint-lawrence: Ah, `iscoroutinefunction` is the bit I was missing. That is much better. Thanks for the nudge in that direction. I've got a few bits to double check but it looks like that works right.

TeamSpen210: Normally it might not be the best idea (since it can’t detect code that just returns a coroutine object directly). But in this case we know `return` isn’t valid.

clint-lawrence:
I've simplified the logic a little with the suggestion from @A5rocks and I think the should be good now.

A5rocks:
Cc @mikenerone @TeamSpen210 any comments before this gets merged?

PR #2979, opened 2024-03-30
Path decorators: allow empty docstrings to support embedded Python

ameily:
I am developing an application that runs on the Windows embedded Python package. It appears that the embedded Python standard library does not include docstrings which means an unhandled exception is raised when importing trio because of a couple `assert X.__doc__ is not None` statements. This PR makes it so `__doc__` is allowed to be `None`.

To test, I am using [Python Windows Embedded 64-bit 3.11.8](https://www.python.org/downloads/release/python-3118/) package. I wrote a small script to show that `__doc__ is None` on every Path function:

```python
import inspect
from pathlib import Path

for name in dir(Path):
    if name.startswith("_"):
        continue

    attr = getattr(Path, name)
    if inspect.isfunction(attr) or inspect.ismethod(attr):
        if attr.__doc__ is None:
            print(f"Path.{name}.__doc__ is None")
```

When I run this through embedded Python I see every function in `Path` has no docstring.

### DIFF
--- a/newsfragments/2987.issue.rst
+++ b/newsfragments/2987.issue.rst
@@ -1,0 +1,1 @@
+Fix crash when importing trio in python windows embedded, and other installs that removes docstrings.
--- a/newsfragments/2987.issue.rst
+++ b/newsfragments/2987.issue.rst
@@ -1,1 +1,1 @@
-Fix crash when importing trio in python windows embedded, and other installs that removes docstrings.
+Fix crash when importing trio in embedded Python on Windows, and other installs that remove docstrings.
--- a/src/trio/_path.py
+++ b/src/trio/_path.py
@@ -38,11 +38,12 @@ def _wraps_async(
             return await run_sync(partial(fn, *args, **kwargs))
 
         update_wrapper(wrapper, wrapped)
-        wrapper.__doc__ = (
-            f"Like :meth:`~{wrapped.__module__}.{wrapped.__qualname__}`, but async.\n"
-            f"\n"
-            f"{cleandoc(wrapped.__doc__ or '')}\n"
-        )
+        if wrapped.__doc__:
+            wrapper.__doc__ = (
+                f"Like :meth:`~{wrapped.__module__}.{wrapped.__qualname__}`, but async.\n"
+                f"\n"
+                f"{cleandoc(wrapped.__doc__)}\n"
+            )
         return wrapper
 
     return decorator
--- a/src/trio/_path.py
+++ b/src/trio/_path.py
@@ -75,9 +75,7 @@ def _wrap_method_path_iterable(
     def wrapper(self: PathT, /, *args: P.args, **kwargs: P.kwargs) -> Iterable[PathT]:
         return map(self.__class__, [*fn(self._wrapped_cls(self), *args, **kwargs)])
 
-    if wrapper.__doc__ is None:
-        wrapper.__doc__ = ""
-    wrapper.__doc__ += (
+    wrapper.__doc__ = (wrapper.__doc__ or "") + (
         f"\n"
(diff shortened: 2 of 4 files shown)


A5rocks:
IMO we shouldn't set the docstring if there isn't one, rather than acting as though the docstring is blank.

TeamSpen210:
That would make sense yes. This might be tricky to test for though - Trio can't work in `-OO` mode, because `pycparser` (imported from `cffi`) needs docstrings to work. But it would be fine if only the standard library is stripped.

A5rocks (changes requested):
Sorry, I should have marked my previous comment properly. Here it is just as reference:
> IMO we shouldn't set the docstring if there isn't one, rather than acting as though the docstring is blank.

src/trio/_path.py
```diff
@@ -38,11 +38,10 @@ async def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
             return await run_sync(partial(fn, *args, **kwargs))
 
         update_wrapper(wrapper, wrapped)
-        assert wrapped.__doc__ is not None
         wrapper.__doc__ = (
             f"Like :meth:`~{wrapped.__module__}.{wrapped.__qualname__}`, but async.\n"
             f"\n"
-            f"{cleandoc(wrapped.__doc__)}\n"
+            f"{cleandoc(wrapped.__doc__ or '')}\n"
```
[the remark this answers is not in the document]
ameily: Sorry about that

A5rocks (approved):

jakkdl:
Could add tests which monkeypatch `Path` and then runs the decorator on them, which would appease codecov.

A5rocks:
I added a test case cause it can be annoying to make them. cc @jakkdl does it look good?

A5rocks:
@ameily I'm super flakey on these (so you can ignore this if you want) but you don't have a changelog attached! Might be nice to have that.

src/trio/_tests/test_path.py
```diff
@@ -252,3 +252,21 @@ async def test_classmethods() -> None:
...
+        trio._path._wrap_method,
+        trio._path._wrap_method_path,
+        trio._path._wrap_method_path_iterable,
+    ],
+)
+def test_wrapping_without_docstrings(
+    wrapper: Callable[[Callable[[], None]], Callable[[], None]]
+) -> None:
+    @wrapper
+    def func_without_docstring() -> None: ...
```
jakkdl, line 270:
```suggestion
    def func_without_docstring() -> None: ...  # pragma: no cover
```

jakkdl (changes requested):

jakkdl:
> I added a test case cause it can be annoying to make them. cc @jakkdl does it look good?
> 
> (the only iffy bit is passing a sync function to a function that assumes it is async -- but it's not like we run it and we can update that if we ever require passed in functions to satisfy `inspect.iscoroutinefunction`)

looks great! (other than needing a `# pragma: no cover` :P)

jakkdl (approved):

newsfragments/2987.issue.rst
```diff
@@ -0,0 +1 @@
+Fix crash when importing trio in python windows embedded, and other installs that removes docstrings.
```
A5rocks, line 1:
```suggestion
Fix crash when importing trio in embedded Python on Windows, and other installs that remove docstrings.
```

Grammar nitpick

A5rocks (approved):

newsfragments/2987.issue.rst
CoolCat467: I would have seconded this

jakkdl: (I merged the suggestion in b519645)

PR #2984, opened 2024-04-06
Document strict ExceptionGroups as the natural default

Zac-HD:
Closes https://github.com/python-trio/trio/issues/2929, by ensuring that our docs present `strict_exception_groups=True` as the natural default (and it is indeed the default in current Trio, as well as the _only_ behavior of asyncio and anyio).

Following discussion on that issue, I've also written up some notes on three ways of handling `ExceptionGroup`s before raising to user code: (1) defer-to, (2) single-error, optionally handling background and foreground tasks differently, and (3) simply letting it propagate to the caller.

2b might actually be a neat option to add to `trio.open_nusery()` itself, but I'd want to see a proof-of-concept judged useful downstream before we committed to supporting it there.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -639,8 +639,7 @@ crucial things to keep in mind:
   * The nursery is marked as "closed", meaning that no new tasks can
     be started inside it.
 
-  * Any unhandled exceptions are re-raised inside the parent task. If
-    there are multiple exceptions, then they're collected up into a
+  * Any unhandled exceptions are re-raised inside the parent task, grouped into a
     single :exc:`BaseExceptionGroup` or :exc:`ExceptionGroup` exception.
 
 Since all tasks are descendents of the initial task, one consequence
@@ -733,9 +732,9 @@ If you want to reraise exceptions, or raise new ones, you can do so, but be awar
 exceptions raised in ``except*`` sections will be raised together in a new exception
 group.
 
-But what if you can't use ``except*`` just yet? Well, for that there is the handy
-exceptiongroup_ library which lets you approximate this behavior with exception handler
-callbacks::
+But what if you can't use Python 3.11, and therefore ``except*``, just yet?
+The same exceptiongroup_ library which backports `ExceptionGroup`  also lets
+you approximate this behavior with exception handler callbacks::
 
     from exceptiongroup import catch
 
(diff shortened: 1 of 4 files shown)


clint-lawrence:
I'm not sure if this is directly relevant to these docs or not.

While working on #2972 I came across a subtle different between trio and asyncio. In asyncio, if a `SystemExit` or `KeyboardInterrupt` is what leads to a task group being cancelled, that individual exception will be re-raised, rather than being wrapped into an exception group. My understanding is that in trio, a nursery (with `strict_exception_groups=True`) will wrap those into a exception group.

Highlighting that difference would be nice, but I'm not sure where that fits best in the docs.

docs/source/reference-core.rst
```diff
@@ -768,43 +767,94 @@ inside the handler function(s) with the ``nonlocal`` keyword::
...
-++++++++++++++++++++++++++++++++++++++++++++++++
+..
+    TODO: what about `Cancelled`?  It's relevantly similar to `KeyboardInterrupt`,
+        but if you have multiple Cancelleds destined for different scopes, it seems
+        like it might be bad to abandon all-but-one of those - we might try to execute
+        some more code which then itself gets cancelled again, and incur more cleanup.
+
+        Maybe CancelScope itself should be ExceptionGroup-aware, and use .split() to
+        extract all and only matching Cancelled errors?  trio_util.defer_to_cancelled
+        only defers if there's exactly one Cancelled, which is a good intermediate step.
```
jakkdl, line 799: Skimming the code, there seems to be some splitting going on? Idk if you're talking about some slightly different scenario though
https://github.com/python-trio/trio/blob/564907b15d8569b738047e4337f1ef3a2350cc30/src/trio/_core/_run.py#L596-L610

Zac-HD: No, I think I was just looking in the wrong place.  Will adjust the comment 👍

Zac-HD:
> In asyncio, if a `SystemExit` or `KeyboardInterrupt` is what leads to a task group being cancelled, that individual exception will be re-raised, rather than being wrapped into an exception group. My understanding is that in trio, a nursery (with `strict_exception_groups=True`) will wrap those into a exception group.

That matches my understanding on the Trio side.  I think this should probably be a separate issue, where we can decide whether to match asyncio (probably yes?) and then implement + document whatever we do.

I've reached my timebox for this PR though, so if people agree it's an improvement on the status quo I'll merge it 🙂

docs/source/reference-core.rst
```diff
@@ -768,43 +767,91 @@ inside the handler function(s) with the ``nonlocal`` keyword::
...
+Note that the final pattern, simply raising an `ExceptionGroup`, is the most common -
+and nurseries automatically do that for you.
+
+**First**, you might want to 'defer to' a particular exception type, raising just that if
+there is any such instance in the group.  For example: `KeyboardInterrupt` has a clear
+meaning for the surrounding code, could reasonably take priority over errors of other
+types, and whether you have one or several of them doesn't really matter.
+
+This pattern can often be implemented using a decorator or a context manager, such
+as :func:`trio_util.multi_error_defer_to` or :func:`trio_util.defer_to_cancelled`.
```
belm0, line 786: (unfortunately trio-util itself is not caught up with the MultiError/ExceptionGroup transition)

docs/source/reference-core.rst
```diff
@@ -768,43 +767,91 @@ inside the handler function(s) with the ``nonlocal`` keyword::
...
+
+**Second**, you might want to treat the concurrency inside your code as an implementation
+detail which is hidden from your users - for example, abstracting a protocol which
+involves sending and receiving data to a simple receive-only interface, or implementing
+a context manager which maintains some background tasks for the length of the
+``async with`` block.
+
+The simple option here is to ``raise MySpecificError from group``, allowing users to
+handle your library-specific error.  This is simple and reliable, but doesn't completely
+hide the nursery.  *Do not* unwrap single exceptions if there could ever be multiple
```
belm0, line 806: it what way does it not hide the nursery?

docs/source/reference-core.rst
```diff
@@ -768,43 +767,91 @@ inside the handler function(s) with the ``nonlocal`` keyword::
...
+    For more on this pattern, see https://github.com/python-trio/trio/issues/2929
+    and the linked issue on trio-websocket.  We may want to provide a nursery mode
+    which handles this automatically; it's annoying but not too complicated and
+    seems like it might be a good feature to ship for such cases.
+
+**Third and most often**, the existence of a nursery in your code is not just an
+implementation detail, and callers *should* be prepared to handle multiple exceptions
+in the form of an `ExceptionGroup`, whether with ``except*`` or manual inspection
+or by just letting it propagate to *their* callers.  Because this is so common,
+it's nurseries' default behavior and you don't need to do anything.
```
belm0, line 830: most often -> most common

But if we're talking about an API, I don't think it's desirable for this to be the most common.  I'm afraid it will become the most common though, because it takes a lot of work to do otherwise.  I'd guess that an `async` call or context manager having concurrency inside is usually an implementation detail, and should not be exposed.  Exposing ExceptionGroup from an API is a significant burden for users, and should be considered carefully.

docs/source/reference-core.rst
Zac-HD:
```python
# end-user code
with library_function_wrapping_a_nursery():
    1/0
```
raises `MySpecificError from ExceptionGroup("", [ZeroDivisionError()])` instead of a bare `ZeroDivisionError`.

docs/source/reference-core.rst
Zac-HD: For many functions, internal concurrency is part of the documented semantics - for example it's not just an implementation detail that [`trio.serve_listeners()`](https://trio.readthedocs.io/en/stable/reference-io.html#trio.serve_listeners) has a nursery internally.  In such cases I think that responsibility for handling an `ExceptionGroup` _should_ lie with the caller.

I agree it's a fuzzy boundary though, and I'll remove the most-common claim.  I don't think the burden is particularly heavy once you're on Python 3.11+ and can use `except*`, though obviously this will vary by project.

belm0: No doubt there are API's that need to expose concurrency.

A few burden points I can think of:
  * "once you're on Python 3.11+" - for any package trying to support the active Python versions, that's 2 years off.  In the meantime you need the cumbersome exceptiongroup.  And on the other hand, people will complain if you depend on exceptiongroup for 3.11+ even though it's harmless, complicating pinned dependencies for CI, etc...
  * we've agreed that pytest doesn't have a good story for matching the semantics of `except*`
  * that the user has to learn `except*` at all, even for an API that doesn't otherwise expose concurrency

docs/source/reference-core.rst
```diff
@@ -768,43 +767,91 @@ inside the handler function(s) with the ``nonlocal`` keyword::
...
+case - with the result that things would work well in testing, and then crash under
+heavier load (typically in production).  `asyncio.TaskGroup` learned from this
+experience and *always* wraps errors into an `ExceptionGroup`, as does ``anyio``,
+and as of Trio 0.25 that's our default behavior too.
+
+We currently support a compatibility argument ``strict_exception_groups=False`` to
+`trio.run` and `trio.open_nursery`, which restores the old behavior (although
+``MultiError`` itself has been fully removed).  We strongly advise against it for
+new code, and encourage existing uses to migrate - we consider the option deprecated
+and plan to remove it after a period of documented and then runtime warnings.
```
A5rocks, line 854: Should we mention that we plan on keeping `strict_exception_groups` until 3.10 EOL (or when usage "has become very rare" which I don't personally know how to measure)? Mentioned in https://github.com/python-trio/trio/issues/2785:

> Later, when strict_exception_groups=False has become very rare or Python 3.10 reaches end of life, we'll remove the strict_exception_groups argument entirely. That's a future issue!

A5rocks:
Should the deprecation message for `strict_exception_groups=False` link to this handling exception groups section? https://github.com/python-trio/trio/blob/59b2eaca688510d478b0ed072f3d71b2938d96d9/src/trio/_core/_run.py#L1011

(if that gets updated, please also update the version number to be `0.25.0`!!)

docs/source/reference-core.rst
Zac-HD: I'd rather not make any specific claims about this; I think 'after 3.10 EOL' is the most likely timeline but we might want to wait for more or less time depending on how the migration is going.

A5rocks:
Oops, I totally missed this:

> I've reached my timebox for this PR though, so if people agree it's an improvement on the status quo I'll merge it 🙂

I'll add commits then! (was about to ask if I could without causing any sort of merge conflict with local changes)

Issue #2981, opened 2024-03-30
Move `pytest-trio`'s hook for deterministic Hypothesis tests upstream into Trio


Zac-HD:
`pytest-trio` does two things for property-based tests with Hypothesis:

1. Inserts the async-to-sync wrapper logic.  `pytest-trio` is clearly the right place to do that.
2. Registers Trio's internal `Random` instance so that the scheduler is deterministic and Hypothesis can replay failing examples (see https://github.com/python-trio/pytest-trio/pull/73 and https://github.com/python-trio/trio/pull/890/).

At the time, `pytest-trio` was the right place to do (2), but not anymore: since https://github.com/HypothesisWorks/hypothesis/pull/2556, it's possible for Trio to ship a _Hypothesis_ plugin which handles the registration if-and-only-if Hypothesis is imported.  [Here are the docs.](https://hypothesis.readthedocs.io/en/latest/strategies.html#hypothesis-integration-via-setuptools-entry-points)

This is useful because it works even for users without `pytest-trio` installed or enabled (c.f. https://github.com/HypothesisWorks/hypothesis/issues/3940), e.g. because they're using `unittest` or running a test function by hand in some debugging loop or the test function calls `trio.run()` internally, and it will allow `pytest-trio` to avoid importing Hypothesis if it wasn't otherwise going to be used (which is nice for performance).

PR #2985, opened 2024-04-08
Ensure we're deterministic for Hypothesis in Trio, not pytest-trio

Zac-HD:
This registration was previously handled by pytest-trio (before Hypothesis supported entry-point plugins).  Upstreaming it means that we'll get the behavior we want for unittest, for functions people call by hand, and so on.  Fixes https://github.com/python-trio/trio/issues/2981.  Manual testing with Hypothesis confirms this works, and our existing determinism-related unit tests continue to pass, so I haven't added more complicated tests which would require subprocesses and installing Hypothesis.

I'll open a PR to remove the relevant logic from pytest-trio once this has been released; double-registration is perfectly safe so it's better to overlap than leave a gap.

Plus some bonus updates to the awesome-trio-libraries page, since I noticed outdated information there.

### DIFF
--- a/docs/source/awesome-trio-libraries.rst
+++ b/docs/source/awesome-trio-libraries.rst
@@ -91,7 +91,8 @@ RPC
 Testing
 -------
 * `pytest-trio <https://github.com/python-trio/pytest-trio>`__ - Pytest plugin for trio.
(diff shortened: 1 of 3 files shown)


CoolCat467 (approved):
Looks pretty good to me, though I should say I'm not super familiar with how hypothesis plugins work but I think this looks reasonable.

PR #2988, opened 2024-04-13
Format using Ruff instead of Black.

Fuyukai:
Removes a dependency, which is always good.

Apparently it's meant to be drop-in compatible, but clearly not.

Most of the test code relating to Black has been removed, because that entire area is an atrocity and needs to be vaporised.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,7 @@ force-exclude = '''
 ignore-words-list = 'astroid,crasher,asend'
 
 [tool.ruff]
+target-version = "py38"
 respect-gitignore = true
 fix = true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,6 @@ force-exclude = '''
 ignore-words-list = 'astroid,crasher,asend'
 
 [tool.ruff]
-target-version = "py38"
 respect-gitignore = true
 fix = true
 
--- a/src/trio/_tests/tools/test_gen_exports.py
+++ b/src/trio/_tests/tools/test_gen_exports.py
@@ -19,7 +19,6 @@ from trio._tools.gen_exports import (
     create_passthrough_args,
     get_public_methods,
     process,
-    run_black,
     run_linters,
     run_ruff,
 )
@@ -93,12 +92,6 @@ skip_lints = pytest.mark.skipif(
 @skip_lints
 @pytest.mark.parametrize("imports", [IMPORT_1, IMPORT_2, IMPORT_3])
 def test_process(tmp_path: Path, imports: str) -> None:
-    try:
-        import black  # noqa: F401
-    # there's no dedicated CI run that has astor+isort, but lacks black.
-    except ImportError as error:  # pragma: no cover
-        skip_if_optional_else_raise(error)
-
     modpath = tmp_path / "_module.py"
     genpath = tmp_path / "_generated_module.py"
     modpath.write_text(SOURCE, encoding="utf-8")
@@ -121,23 +114,6 @@ def test_process(tmp_path: Path, imports: str) -> None:
     with pytest.raises(SystemExit) as excinfo:
         process([File(modpath, "runner", imports=imports)], do_test=True)
     assert excinfo.value.code == 1
-
-
-@skip_lints
-def test_run_black(tmp_path: Path) -> None:
-    """Test that processing properly fails if black does."""
-    try:
-        import black  # noqa: F401
-    except ImportError as error:  # pragma: no cover
-        skip_if_optional_else_raise(error)
-
-    file = File(tmp_path / "module.py", "module")
-
-    success, _ = run_black(file, "class not valid code ><")
-    assert not success
-
-    success, _ = run_black(file, "import waffle\n;import trio")
-    assert not success
 
 
 @skip_lints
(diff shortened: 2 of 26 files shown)


src/trio/_tools/gen_exports.py
```diff
@@ -184,8 +213,8 @@ def gen_public_wrappers_source(file: File) -> str:
 
     generated = ["".join(header)]
 
-    source = astor.code_to_ast.parse_file(file.path)
-    method_names = []
+    source = ast.parse(file.path.read_text())
```
A5rocks, line 218: Is there a reason why we're not using `astor` anymore?

If it's because we're running this through an autoformatter which should do AST -> code, then maybe we could use `ast.unparse` instead of `astor.to_source` and drop `astor` entirely. Though I see that `ast.unparse` is 3.9+ (cause type hints...) so I guess maybe that's just a `# TODO: ...` comment for when we drop 3.8.

src/trio/_tools/gen_exports.py
```diff
@@ -3,6 +3,9 @@
 Code generation script for class methods
 to be exported as public API
 """
+
+# mypy: disable-error-code="unused-ignore"
```
A5rocks, line 7: Is mypy saying that the `astor` call's `# type: ignore` is unused? Why/why is it wrong?

A5rocks (commented):
Not sure if we use them, but https://docs.astral.sh/ruff/formatter/#docstring-formatting  might be nice. If we do use code in docstrings, then we'll need to check line numbers in docs are accurate afterwards.

src/trio/_tools/gen_exports.py
Fuyukai: > Is there a reason why we're not using astor anymore?

astor is untyped. This function is literally just a wrapper around `ast.parse` that opens the file. Errors in my editor annoy me.

src/trio/_tools/gen_exports.py
Fuyukai: Mypy doesn't care about untyped libraries; pyright strict does.

Fuyukai:
I'd like to get this PR approved and merged as soon as possible so I can base my next PR off of it. I can't really imagine anything objectionable here.

A5rocks (approved):
I doubt any content here is bad, but someone might want you to add something. (I'll probably add the `TODO` about `ast.unparse` and maybe change the `# type: ignore` on `astor.to_source` to be `# type: ignore[unused-ignore]` in a seperate PR if you don't)

I think someone else should merge this as a final set of eyes. If you want to make the other PR now, you could rebase this to be fewer commits and we can use the "merge commit" merge here to make git history work out.

pyproject.toml
```diff
@@ -86,6 +86,7 @@ force-exclude = '''
 ignore-words-list = 'astroid,crasher,asend'
 
 [tool.ruff]
+target-version = "py38"
```
CoolCat467, line 89: Ruff is supposed to auto-detect your target version from the `requires-python = ">=3.8"`. Is this no longer the case?

src/trio/_tests/tools/test_gen_exports.py
```diff
@@ -170,7 +146,6 @@ def test_run_ruff(tmp_path: Path) -> None:
 def test_lint_failure(tmp_path: Path) -> None:
     """Test that processing properly fails if black or ruff does."""
```
CoolCat467, line 147: Missed docstring change

src/trio/_tools/gen_exports.py
CoolCat467: I would much rather type ignore the specific lines that are causing issues than the entire file, because if something in our assumptions change later that will make debugging more tricky, and plus I would consider it better practice.

src/trio/_tools/gen_exports.py
```diff
@@ -184,8 +179,9 @@ def run_linters(file: File, source: str) -> str:
       SystemExit: If either failed.
     """
```
CoolCat467, line 180: Nit, grammar also needs updating, `either` referrers to two, but now it's only one.

CoolCat467 (changes requested):
Looks pretty good other than a few things I noted in review comments, might want to look closer at a few docstrings. Not sure I particularly like the new way ruff is formatting the type annotations, would be interested in seeing if there are configuration options for the formatter to potentially change that.

Looks pretty good overall though!

pyproject.toml
Fuyukai: Didn't know that was a thing. I'll remove it.

CoolCat467 (approved):
Looks good, but I think I would particularly like to see if @jakkdl and/or @TeamSpen210 have any comments as well, they've done a lot of work in this sort of area if I'm not mistaken.

jakkdl (commented):
Yeah I, uh, think the vast majority of the formatting changes in this PR are negative wrt readability. I'm somewhat biased as I'm the author of the latest set of formatting improvements wrt type hints in Black, but the ruff authors have also intentionally skipped some of the (imo) improvements partly due to disagreeing with implications of style choices, and partly due to implementation details being difficult to replicate. Not finding the exact issue atm, but e.g. https://github.com/psf/black/issues/4123 has very long and incredibly in-the-weed discussion about type hints. I also have some weak long-term worries about ruff gobbling up Black, but idk.

If black being a testing requirement is bad, then we can mark the tests with `skipif`s. Or if having the tests in themselves is bad, then remove them. `test_run_black` does seem somewhat overkill, but /shrug

The astor->ast changes seem good though.

Overall I'm not a hard no, but I vote against switching away from Black.

jakkdl:
> Not finding the exact issue atm, but e.g. [psf/black#4123](https://github.com/psf/black/issues/4123) has very long and incredibly in-the-weed discussion about type hints. 

No that was the issue I was thinking about, it's linked at https://docs.astral.sh/ruff/formatter/black/#parenthesizing-long-nested-expressions

Fuyukai:
Guess this ain't gonna happen.

PR #2989, opened 2024-04-15
Add support to testing.RaisesGroup for catching unwrapped exceptions

jakkdl:
* Deprecates the `strict` parameter, renaming it to `flatten_subgroups`.
* Adds the `allow_unwrapped` parameter, adding the possibility to more closely mirror `except*`. This is a feature requested in https://github.com/pytest-dev/pytest/issues/11538#issuecomment-2024361910 by @belm0 
* Fixes a bug I noticed while reading the code, where e.g.
```python
with RaisesGroup(ValueError, TypeError, strict=False):
  raise ExceptionGroup("", [ExceptionGroup("", [ValueError(), TypeError()])])
```
would fail due to it checking if the number of exceptions was correct before flattening the structure of the raised exceptiongroup.

~~I also found an unrelated issue where mypy&pyright are not able to deduce the type when passing multiple `Matcher` objects matching against different exceptions to `RaisesGroup`. (or `Matcher`+class, but two classes work). Fiddled around a little bit but wasn't able to quickly find a fix, and explicitly setting `[Exception]` isn't especially onerous.~~
EDIT: this was due to `TypeVar` lacking covariance.


src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -134,3 +134,26 @@ def check_nested_raisesgroups_matches() -> None:
...
+    d = a
+    d = b
+    d = c
+    assert d
+
+
+def check_multiple_exceptions_2() -> None:
+    # "Cannot infer argument 1 of RaisesGroup"
+    # if removing overloads from Matcher then pyright stops warning, but mypy still
+    # complains
```
belm0, line 154: will you file an issue to these tools?

jakkdl:
Oh, turns out the unrelated typing error does work with pyright even with the overloads, and it's solely a mypy thing. I'll try to get a minimal repro and/or see if it's a known mypy issue and/or report it.

src/trio/testing/_raises_group.py
```diff
@@ -263,8 +263,9 @@ class RaisesGroup(ContextManager[ExceptionInfo[BaseExceptionGroup[E]]], SuperCla
     This works similar to ``pytest.raises``, and a version of it will hopefully be added upstream, after which this can be deprecated and removed. See https://github.com/pytest-dev/pytest/issues/11538
 
 
-    This differs from :ref:`except* <except_star>` in that all specified exceptions must be present, *and no others*. It will similarly not catch exceptions *not* wrapped in an exceptiongroup.
-    If you don't care for the nesting level of the exceptions you can pass ``strict=False``.
+    This differs from :ref:`except* <except_star>` in that all specified exceptions must be present, *and no others*. It will by default not catch exceptions *not* wrapped in an exceptiongroup.
```
belm0, line 266: consider "by default only catch exceptions wrapped in an exceptiongroup" to avoid double negative

src/trio/testing/_raises_group.py
```diff
@@ -383,6 +384,15 @@ def matches(
...
         if not isinstance(exc_val, BaseExceptionGroup):
+            if not self.strict and len(self.expected_exceptions) == 1:
+                exp_exc = self.expected_exceptions[0]
+                if isinstance(exp_exc, Matcher) and exp_exc.matches(exc_val):
+                    return True
+                if isinstance(exp_exc, type) and isinstance(exc_val, exp_exc):
+                    return True
+            # Consider printing a helpful message on unwrapped exception, strict=False and
+            # `len(self.expected_exceptions) > 1`; since they might expect the exceptions
+            # to be 'or'd like with `pytest.raises`.
```
belm0, line 395: if it's a valid usage, the message might be annoying or even interfere with tooling (where a project tries to keep test runs free of warning messages)

belm0 (approved):

src/trio/testing/_raises_group.py
jakkdl: It'd only be:
* if the code is `with RaisesGroups(excA, excB, strict=False): ...`
* and a single exception [that would match excA or excB] is raised

Either way your test case is failing, so would be similar to how `pytest.raises` tries to be helpful when it looks like a user has failed to `re.escape` a `match=` string.

In general I think the matching & construction of stuff like nested groups is complicated enough that the tool should be emitting when something is close to matching but doesn't. Also see
https://github.com/python-trio/trio/blob/62c0ea02e0bf98188544c6e0136caa6b46911b92/src/trio/testing/_raises_group.py#L383-L385

src/trio/_tests/type_tests/raisesgroup.py
jakkdl: yeah, see https://github.com/python-trio/trio/pull/2989#issuecomment-2057109799

jakkdl: The issue was in fact just that the typevar wasn't specified as covariant.

src/trio/testing/_raises_group.py
jakkdl: rewrote the section entirely, hopefully for the better.

jakkdl:
@TeamSpen210 after adding `covariant=True` to the `TypeVar` the `docs/source/typevars.py` fails to resolve it:
```
/home/h/Git/trio/looser_excgroups/src/trio/testing/__init__.py:docstring of trio.testing.RaisesGroup:1: WARNING: py:obj reference target not found: typing.Callable[[BaseExceptionGroup[+E]], bool] | None
/home/h/Git/trio/looser_excgroups/src/trio/testing/__init__.py:docstring of trio.testing.Matcher:1: WARNING: py:class reference target not found: type[+E] | None
/home/h/Git/trio/looser_excgroups/src/trio/testing/_raises_group.py:docstring of trio.testing._raises_group._ExceptionInfo.type:1: WARNING: py:class reference target not found: type[+E]
```
I'm a bit surprised as covariant typevars exist in a few other places in the codebase without issue. Unless you have any opinions I'll try do some ugly workaround where I either add a plussed [copy] to the dicts in `identify_typevars`, or strip `+`s from the target in `lookup_reference`

jakkdl:
I can make a theoretical case for splitting `strict=False` into two separate flags - one that allows unwrapped exceptions, and one that flattens nested exceptiongroups. But I suspect the added burden of having to set both might outweigh the gain from somebody that actually wants them split:
```python
# they want this to pass
with RaisesGroup(ValueError, strict=False):
    raise ExceptionGroup("outer", [ExceptionGroup("inner", [ValueError()])])
# but this to fail
with RaisesGroup(ValueError, strict=False):
    raise ValueError
```

Of course they can just do
```python
with RaisesGroup(..., strict=False) as e:
    ...
assert isinstance(e, ExceptionGroup)
```
but that's easy to forget and could silently introduce changes in behavior.


One, probably better, case in favor of splitting out `allow_unwrapped=True` is that we can raise an error in `RaisesGroup.__init__` if specifying several exceptions + `allow_unwrapped=True`. That's both easier to implement and faster to debug for end-users than trying to resolve that with a message on a failed catch https://github.com/python-trio/trio/pull/2989#discussion_r1565979668


Since typing `with RaisesGroup(ValueError, allow_unwrapped=True, flatten=True)` is a mouthful we (or leave that for pytest) could add convenience aliases, or just rely on end-users to write their own thin wrappers that change the defaults.

TeamSpen210:
Fixed the typevar-related issues, but now there's something else in `history.rst`. Not fully sure why the `+E` is only happening with this typevar, but I found a workaround. By using `autodoc_process_signatures()` I could insert the fully qualified name for the var, allowing Sphinx to find it.

newsfragments/2989.feature.rst
```diff
@@ -0,0 +1 @@
+:ref:`testing.RaisesGroup` can now catch unwrapped exceptions with ``strict=False``, similar to ``except*``.
```
Zac-HD, line 1: Undefined label is making Sphinx fail, maybe we want e.g.

```suggestion
:class:`trio.testing.RaisesGroup` can now catch unwrapped exceptions with ``strict=False``, similar to ``except*``.
```

Zac-HD (approved):
looks good to me, once the build passes let's merge!

jakkdl:
Seems like nobody else had opinions on `strict` vs `allow_unwrapped`+`flatten_subgroups`. The ability to raise a `ValueError` on users doing `RaisesGroup(SyntaxError, TypeError, allow_unwrapped=True)` with a helpful message on what else to do convinced me that this is the way to go, even if it will be more verbose when you want to fully emulate `except*` (i.e. you expect a single exception, but don't care if it's unwrapped, or in any level of nesting).

newsfragments/2989.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug where :class:`trio.testing.RaisesGroup` would check the number of exceptions in the raised `ExceptionGroup` before flattening subgroups, leading to incorrectly failed matches with ``flatten_subgroups=True`` (previously ``strict=False``).
```
belm0, line 1: This note seems odd, given that `flatten_subgroups` was just added in this PR.  I think it should only mention `strict=False`, which is the only way people can be hitting this bug.

newsfragments/2989.feature.rst
```diff
@@ -0,0 +1 @@
+:class:`trio.testing.RaisesGroup` can now catch an unwrapped exception with ``unwrapped=True``, similar to ``except*``.
```
belm0, line 1: It should also mention `flatten_subgroups`.  Please also update PR description with this info (including deprecation of `strict`.

newsfragments/2989.bugfix.rst
jakkdl: flipped the references to them

belm0 (approved):

jakkdl:
bumping exceptiongroup so whoever reviews the next dep bump don't have to spend any time tracking down why type tests start ""failing""

src/trio/_tests/test_testing_raisesgroup.py
```diff
@@ -78,33 +78,77 @@ def test_raises_group() -> None:
...
+            ...
+    # But it *can* be used to check for nesting level +- 1 if they move it to
+    # the nested RaisesGroup. Users should probably use `Matcher`s instead though.
+    with RaisesGroup(RaisesGroup(ValueError, allow_unwrapped=True)):
+        raise ExceptionGroup("", [ExceptionGroup("", [ValueError()])])
+    with RaisesGroup(RaisesGroup(ValueError, allow_unwrapped=True)):
+        raise ExceptionGroup("", [ValueError()])
+
+    # with allow_unwrapped=False (default) it will not be caught
+    with pytest.raises(ValueError, match="value error text"):
```
CoolCat467, line 143: I would change this to use a more strict regular expression `^value error text$` like the line above that was deleted.

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -84,9 +79,9 @@ def check_filenotfound(exc: FileNotFoundError) -> bool:
     Matcher(exception_type=ValueError)
     Matcher(match="regex")
     Matcher(check=check_exc)
-    Matcher(check=check_filenotfound)  # type: ignore
     Matcher(ValueError, match="regex")
     Matcher(FileNotFoundError, check=check_filenotfound)
+    Matcher(check=check_filenotfound)  # type: ignore # not narrowed
```
CoolCat467, line 84: What does "not narrowed" mean in this situation? Does it mean not ignoring specific type (such as `# type: ignore[misc]`)? If so why?

CoolCat467 (approved):
Looks pretty good other than a few small things I noted.

CoolCat467 (changes requested):
whoops wrong submit type

src/trio/_tests/type_tests/raisesgroup.py
jakkdl: I could clarify the comment, though there is a comment a few lines up not seen in the preview that explains it. I simply moved the test to be next to the other `check=check_filenotfound` line.
It fails because `check_filenotfound` has its arg typed as `FileNotFoundError` but the `Matcher` doesn't narrow the type from `Exception` when there's no type passed.

Specific ignore types are not used in type tests due to how pyright/mypy differs. Even if you ignore a specific type with `# type: ignore` pyright will ignore all warnings on the line, because it does not have the same mapping of warning names. I suppose it can still give some information and give warning if mypy warning changes, but it's also hella misleading wrt pyright.

CoolCat467: Thanks for clarifying, yea I didn't see the comment in the changes preview

CoolCat467 (approved):
Looks good. I think I saw a few regexes reading through the changes that could still have end specifiers added, but I think it should be ok.

src/trio/testing/_raises_group.py
```diff
@@ -307,28 +325,60 @@ def __init__(
...
+                " If you want to match one of multiple possible exceptions you should"
+                " use a `Matcher`."
+                " E.g. `Matcher(check=lambda e: isinstance(e, (...)))`"
+            )
+        if allow_unwrapped and isinstance(exception, RaisesGroup):
+            raise ValueError(
+                "`allow_unwrapped=True` has no effect when expecting a `RaisesGroup`."
+                " You might want it in the expected `RaisesGroup`, or"
+                " `flatten_subgroups=True` if you don't care about the structure."
+            )
```
A5rocks, line 366: Another bad construction: `allow_unwrapped` and `match`. Currently the behavior is incorrect (it ignores `match` if there's no exception group).

jakkdl: I think the current behaviour is somewhat sensible: "match the text of the group if it's in a group, if you want to match the text of the exception you should use a matcher object".
Same thing for `check`.

But I could see that being unexpected, and we're better off banning `allow_unwrapped` plus any of `match`/`check`. And in the error message instructing users to use a `Matcher` object if they intended to `match`/`check` on the exception; or do extra checks afterwards on the group if they wanted constraints on the group. The latter should be sufficiently rare that it's fine to force users to jump through hoops for it.

src/trio/testing/_raises_group.py
```diff
@@ -307,28 +325,60 @@ def __init__(
...
         self.is_baseexceptiongroup = False
 
+        if strict is not None:
+            warnings.warn(
+                DeprecationWarning(
+                    "`strict=False` has been replaced with `flatten_subgroups=True`"
+                    " with the introduction of `allow_unwrapped` as a parameter."
+                ),
+                stacklevel=2,
+            )
```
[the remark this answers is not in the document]
jakkdl: mostly just me being confused about the existence of `TrioDeprecationWarning` (hence #2992). One weird part with `warn_deprecated` is you have to know what the next version bump will be, but I'm guessing I should mark it as `0.25.1`.
But I also don't think this one is worthy of an official extended period of supporting it, it's a trivial change to use `flatten_subgroups=True` instead and this is explicitly a temporary helper until it's merged into pytest, so it's barely worth having a DeprecationWarning at all. It's not used at runtime either, so it can only break tests. But I guess `TrioDeprecationWarning` doesn't have an explicit promise of how long it will be kept deprecated anyhow

src/trio/_tests/test_testing_raisesgroup.py
```diff
@@ -78,39 +78,100 @@ def test_raises_group() -> None:
...
+    with RaisesGroup(
+        Matcher(check=lambda e: isinstance(e, (SyntaxError, ValueError))),
+        allow_unwrapped=True,
+    ):
+        raise ValueError
+
+    # Unwrapped nested `RaisesGroup` is likely a user error, so we raise an error.
+    with pytest.raises(ValueError, match="has no effect when expecting"):
+        with RaisesGroup(RaisesGroup(ValueError), allow_unwrapped=True):
+            ...
```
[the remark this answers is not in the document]
jakkdl: No yeah, it raises an error on construction of the group so the content of the with-block is never executed. 
You could work around the latter with matching against `^ *...$`.
But I can see marginal utility from seeing if-the-content-of-a-block-is-executed even when the block content is only `...`, so unless we have lots of `... # pragma: ignore` I'm not sure there's much reason to add it to global ignores.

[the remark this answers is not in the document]
jakkdl: There's zero instances of `^\.\.\. # pragma:`, though I guess there might be skips on a preceding `else` branch.

src/trio/_tests/test_testing_raisesgroup.py
```diff
@@ -118,8 +118,7 @@ def test_catch_unwrapped_exceptions() -> None:
     with pytest.raises(
         ValueError, match="^You cannot specify multiple exceptions with"
     ):
-        with RaisesGroup(SyntaxError, ValueError, allow_unwrapped=True):
-            ...
+        RaisesGroup(SyntaxError, ValueError, allow_unwrapped=True)
```
CoolCat467, line 121: For this one and other, why are we not using with block anymore?

jakkdl: error is raised on construction, not on  __enter__. With a with block I needed # pragma: no cover. See review comments by a5rocks

src/trio/_tests/test_deprecate.py
```diff
@@ -254,9 +254,19 @@ def test_module_with_deprecations(recwarn_always: pytest.WarningsRecorder) -> No
...
     assert isinstance(got.message, Warning)
     assert "instead-string instead" in got.message.args[0]
 
     with pytest.raises(AttributeError):
         module_with_deprecations.asdf  # type: ignore[attr-defined]  # noqa: B018  # "useless expression"
+
+
+def test_warning_class() -> None:
+    with pytest.warns(DeprecationWarning):
+        warn_deprecated("foo", "bar", issue=None, instead=None)
```
A5rocks, line 267: For this (and quite a bit above), we can probably use specialized pytest helpers.

src/trio/_tests/test_testing_raisesgroup.py
```diff
@@ -125,6 +185,37 @@ def test_check() -> None:
...
+        return True
+
+    msg = (
+        "`allow_unwrapped=True` bypasses the `match` and `check` parameters"
+        " if the exception is unwrapped. If you intended to match/check the"
+        " exception you should use a `Matcher` object. If you want to match/check"
+        " the exceptiongroup when the exception *is* wrapped you need to"
+        " do e.g. `if isinstance(exc.value, ExceptionGroup):"
+        " assert RaisesGroup(...).matches(exc.value)` afterwards."
+    )
```
A5rocks, line 199: Do we do this elsewhere? I'm pretty sure that elsewhere we try to catch only snippets of the message.

Though that's probably more because the message elsewhere is variable. I'm a bit concerned that this is brittle though I'm sure that if we update the message to be more descriptive the person doing that will be fine copying it here.

src/trio/_tests/test_testing_raisesgroup.py
```diff
@@ -78,39 +78,99 @@ def test_raises_group() -> None:
...
     with pytest.raises(ValueError, match="^value error text$"):
-        with RaisesGroup(ValueError, strict=False):
+        with RaisesGroup(ValueError):
             raise ValueError("value error text")
 
+    # code coverage
+    with pytest.raises(TypeError):
+        with RaisesGroup(ValueError, allow_unwrapped=True):
+            raise TypeError
+
```
A5rocks, line 150: One untested case: can we make sure that `allow_unwrapped` doesn't allow more than 1 unwrapping? I thought that was the case until I read the docstring about `flatten_subgroups`. Prior to that I had a comment here about checking that case.

Basically I don't think the only 1 unwrapping behavior is necessarily intuitive so I think it's a good idea to have a failing test case in case someone forgets and updates the code.

src/trio/testing/_raises_group.py
```diff
@@ -307,28 +325,68 @@ def __init__(
...
+        if allow_unwrapped and (match is not None or check is not None):
+            raise ValueError(
+                "`allow_unwrapped=True` bypasses the `match` and `check` parameters"
+                " if the exception is unwrapped. If you intended to match/check the"
+                " exception you should use a `Matcher` object. If you want to match/check"
+                " the exceptiongroup when the exception *is* wrapped you need to"
+                " do e.g. `if isinstance(exc.value, ExceptionGroup):"
+                " assert RaisesGroup(...).matches(exc.value)` afterwards."
+            )
```
A5rocks, line 374: One typing nicety is we can make it clear these parameters must not be mixed through an overload. (yes this complicates the signature but I think we'll need it anyways for better typing once [mypy supports the typevar stuff](https://github.com/python/mypy/pull/16753) we want)

A5rocks (commented):
A few comments. Sorry for taking a while, I wasn't sure whether we should merge this yet in case we want a patch release w/ those embedded Python fixes, but I haven't made that release yet so... Oh well.

jakkdl:
oh sorry for the re-request, didn't get notified about your review

src/trio/_tests/test_deprecate.py
jakkdl: this already uses `pytest.warns`, or what kind of helper do you mean?

src/trio/_tests/test_testing_raisesgroup.py
jakkdl: nah it's not them being variable, the match strings in `test_catch_unwrapped_exceptions` was mostly me being lazy - but also because in all those tests the error message in question is only matched against once, whereas here it's matched for twice.
It was a bit of a pain when iterating the message, but any future updates to the message should indeed be simply copy-paste jobs to update the string in the test.

Quoting the whole string also does have the positive side effect of effectively being a comment explaining the test case.

src/trio/_tests/test_testing_raisesgroup.py
jakkdl: added :+1:

src/trio/testing/_raises_group.py
jakkdl: added :+1:

jakkdl:
> oh sorry for the re-request, didn't get notified about your review

fixed everything now though, so re-review request is warranted

src/trio/_tests/test_deprecate.py
A5rocks: Specifically I was thinking of pytest.deprecated_call

Tbf this is more of a nitpick

jakkdl: Haha, that is an incredibly nitty nitpick. Though we might care about whether we raise `DeprecationWarning` or `PendingDeprecationWarning` so I actually think the current implementation is better. We can do both though

jakkdl:
I encountered some issues with typing for functions passed to the `check` argument, and thought about handling that in a separate pull request - but it got so thorny that I'll just add the failing tests now and maaaybe address it in a separate PR once I hear from https://github.com/python/mypy/issues/17251

src/trio/testing/_raises_group.py
```diff
@@ -303,32 +322,108 @@ class RaisesGroup(ContextManager[ExceptionInfo[BaseExceptionGroup[E]]], SuperCla
...
+    # RaisesGroup instance, match is None, check is None
+    @overload
+    def __init__(
+        self,
+        exception: type[E] | Matcher[E],
+        *,
+        allow_unwrapped: Literal[True],
+        flatten_subgroups: bool = False,
+        match: None = None,
+        check: None = None,
```
A5rocks, line 335:
```suggestion
```

I don't think it's necessary to provide these. a) we're adding the param now so no backcompat issues and b) I don't think we should care about backcompat for types because they're just hints.

A5rocks (approved):

A5rocks:
Also re: your mypy bug, my impression was that it just ignores `__new__` in favor of `__init__` if it's available.

src/trio/testing/_raises_group.py
jakkdl: Hmm, I think it doesn't hurt to have them, and allows users to do some automated stuff where they pass variables that they know to be `None`:
```python
for allow_unwrapped,match in zip([True, None], [False, "hello"]):
  with RaisesGroup(Exception, allow_unwrapped=allow_unwrapped, match=match):
    ...
```

jakkdl:
> Also re: your mypy bug, my impression was that it just ignores `__new__` in favor of `__init__` if it's available.

scrolling through https://github.com/python/mypy/blob/master/test-data/unit/check-classes.test I guess that is the current state of things with mypy. But after sleeping on it I think I have a better way of resolving it - but that's for a new PR. It's finally time to merge this one :rocket:

PR #2993, opened 2024-04-25
Remove `test_run_black`

A5rocks:
https://github.com/python-trio/trio/pull/2988 was closed, but I think we agree that `test_run_black` can be removed. I don't actually remember why it was introduced, but using blame all the way back leads to it not seeming necessary. It was made because `test_lint_failure` wouldn't check that `ruff` was actually running, so we duplicated that to have a `black`-specific test and a `ruff`-specific test. The `ruff` one is still useful (maybe), but the `black` one is duplicated.

I'm not so sure *any* of these tests should exist anymore. I think I kinda pushed for the `ruff`-specific one but now I think it's ultimately unnecessary. If `ruff` doesn't run in `gen_exports`, that's ... fine. We have CI jobs dedicated to checking our code is alright and if it isn't, then that will be caught. Removing that is another PR's job though.

### DIFF
--- a/src/trio/_tests/tools/test_gen_exports.py
+++ b/src/trio/_tests/tools/test_gen_exports.py
@@ -19,7 +19,6 @@ from trio._tools.gen_exports import (
     create_passthrough_args,
     get_public_methods,
     process,
-    run_black,
     run_linters,
     run_ruff,
 )
--- a/src/trio/_tests/tools/test_gen_exports.py
+++ b/src/trio/_tests/tools/test_gen_exports.py
@@ -124,23 +124,6 @@ def test_process(tmp_path: Path, imports: str) -> None:
 
 
 @skip_lints
-def test_run_black(tmp_path: Path) -> None:
-    """Test that processing properly fails if black does."""
-    try:
-        import black  # noqa: F401
-    except ImportError as error:  # pragma: no cover
-        skip_if_optional_else_raise(error)
-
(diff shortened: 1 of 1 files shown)


A5rocks:
Yikes looks like GHA sometimes bumps macos version to use the new m1 chips even though setup python doesn't like that: https://github.com/A5rocks/trio/actions/runs/8827503948/job/24234861793

CoolCat467:
> If `ruff` doesn't run in `gen_exports`, that's ... fine. We have CI jobs dedicated to checking our code is alright and if it isn't, then that will be caught. Removing that is another PR's job though.

I think running ruff in `gen_exports` is important, now that ruff handles import sorting

CoolCat467 (approved):
I would be for removing this test as well, but I think I remember I tried removing it or something in a previous pull request (I specifically remember the import waffle haha) and I think I recall it lead to a loss of test coverage that was an issue I think. Don't completely remember the details, but thought I'd share.

A5rocks:
Doesn't look like this loses test coverage. Removing `test_run_ruff` probably does though, and removing `test_lint_failures` definitely does. (but I'd argue that coverage loss is beneficial.)

For clarity, when I say "If `ruff` doesn't run in `gen_exports`, that's ... fine." I mean "if we break it accidentally" (which is the case the tests guard against).

jakkdl (approved):

Issue #1439, opened 2020-03-19
How to get "trio token" for trio.to_thread.run_sync?


Nikratio:
Hello,

I am trying to migrate from trio.BlockingTrioPortal to trio.from_thread(). Unfortunately I am running into some trouble because I am trying to call from a not-trio-created thread.

The documentation specifies that I can "pass a keyword argument, trio_token specifiying a specific trio.run loop to re-enter". But how do I get such a token? It doesn't seem mentioned anywhere else (including the description of trio.run).

I think I am looking for a function like `trio.get_current_token()` whose output I can save and pass to a thread for use with `trio.from_thread.run_sync.

oremanj:
Use `trio.hazmat.current_trio_token()` from within the Trio thread. For more details, see https://trio.readthedocs.io/en/stable/reference-hazmat.html#trio-tokens

njsmith:
Sounds like the docs could use a small update to point to that?

Nikratio:
Yeah, a pointer would be great. I still feels somewhat counterintuitive that I have to fallback on `trio.hazmat` though, especially because with the old interface that wasn't necessary.

pquentin:
@Nikratio We renamed hazmat to lowlevel recently to emphasize the fact that there's nothing wrong with that module :)

PR #3001, opened 2024-05-18
Updated docstring for from_thread_run to reflect how to get trio token

oakal510:
Added a documentation fix for the following issue: fixes https://github.com/python-trio/trio/issues/1439.


Zac-HD (approved):
Thanks for the fix, @oakal510!

PR #3002, opened 2024-05-18
tweaks to new REPL

richardsheridan:
A follow up to #2972. These would have been review comments earlier if I knew the code was going to be (a) this simple and (b) in my wheelhouse!

I think `runcode` can just be:

``` python

    def runcode(self, code: types.CodeType) -> None:
        try:
            func = types.FunctionType(code, self.locals)
            if inspect.iscoroutinefunction(func):
                trio.from_thread.run(func)
            else:
                trio.from_thread.run_sync(func)
        except SystemExit:
            # ...snip...
            raise
        except BaseException:
            self.showtraceback()

```

It's less code, less layers of onion unwrapping, and the more common `from_thread.run_sync` is more efficient (not that anyone would notice latency from two extra checkpoints at the REPL). That's the first commit.

I also noticed one fragility, which is that KI works thanks to the thread task reentry feature combined with the fact that we inject `KeyboardInterrupt` into the main task.  If we were to go the route of https://github.com/python-trio/trio/issues/733#issuecomment-629126671 and just cancel everything and transmute to `KeyboardInterrupt` when the run finishes, the run would be cancelled. I put a test for this in the second commit.

Finally, we can hide the threading code frames from the traceback, although I'm not sure about editing the global exception state like that. that's the last commit.

### DIFF
--- a/src/trio/_repl.py
+++ b/src/trio/_repl.py
@@ -8,8 +8,6 @@ import types
 import warnings
 from code import InteractiveConsole
 
-import outcome
-
 import trio
 import trio.lowlevel
 
@@ -25,15 +23,12 @@ class TrioInteractiveConsole(InteractiveConsole):
         self.compile.compiler.flags |= ast.PyCF_ALLOW_TOP_LEVEL_AWAIT
 
     def runcode(self, code: types.CodeType) -> None:
-        async def _runcode_in_trio() -> outcome.Outcome[object]:
+        try:
             func = types.FunctionType(code, self.locals)
             if inspect.iscoroutinefunction(func):
-                return await outcome.acapture(func)
+                trio.from_thread.run(func)
             else:
-                return outcome.capture(func)
-
-        try:
-            trio.from_thread.run(_runcode_in_trio).unwrap()
+                trio.from_thread.run_sync(func)
         except SystemExit:
             # If it is SystemExit quit the repl. Otherwise, print the
             # traceback.
--- a/src/trio/_repl.py
+++ b/src/trio/_repl.py
@@ -33,19 +33,19 @@ class TrioInteractiveConsole(InteractiveConsole):
         else:
             result = trio.from_thread.run_sync(outcome.capture, func)
         if isinstance(result, outcome.Error):
-            # If it is SystemExit quit the repl. Otherwise, print the traceback.
+            # If it is SystemExit, quit the repl. Otherwise, print the traceback.
+            # If there is a SystemExit inside a BaseExceptionGroup, it probably isn't
+            # the user trying to quit the repl, but rather an error in the code. So, we
+            # don't try to inspect groups for SystemExit. Instead, we just print and
+            # return to the REPL.
             if isinstance(result.error, SystemExit):
-                # There could be a SystemExit inside a BaseExceptionGroup. If
-                # that happens, it probably isn't the user trying to quit the
-                # repl, but an error in the code. So we print the exception
-                # and stay in the repl.
                 raise result.error
             else:
                 # Inline our own version of self.showtraceback that can use
                 # outcome.Error.error directly to print clean tracebacks.
                 # This also means overriding self.showtraceback does nothing.
-                sys.last_traceback = result.error.__traceback__
                 sys.last_type, sys.last_value = type(result.error), result.error
+                sys.last_traceback = result.error.__traceback__
                 # see https://docs.python.org/3/library/sys.html#sys.last_exc
                 if sys.version_info >= (3, 12):
                     sys.last_exc = result.error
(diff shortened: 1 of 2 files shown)


richardsheridan:
Looks like that traceback cleanup trick only works on 3.11 later. Either we could find a different trick or only assert it on recent python versions!

A5rocks:
Cc @clint-lawrence cause you made the original PR

CoolCat467:
Out of date branch so updated

richardsheridan:
The difference in editing exceptions that showed up in 3.11 is [documented](https://docs.python.org/3/library/sys.html#sys.exc_info) so I think we can just rely on it.

richardsheridan:
And to follow up on people interested in a nursery for background tasks, my test accidentally shows that you can just use the system nursery.

https://github.com/python-trio/trio/blob/8a0b573c57442b4ea5a28ae154d31b0ee8b0a08a/src/trio/_tests/test_repl.py#L106-L117

src/trio/_repl.py
```diff
@@ -42,7 +39,13 @@ async def _runcode_in_trio() -> outcome.Outcome[object]:
...
             # and stay in the repl.
             raise
-        except BaseException:
+        except BaseException as exc:
+            # remove our code from traceback
+            tb = exc.__traceback__
+            for _ in range(n):
+                assert tb is not None
+                tb = tb.tb_next
+            exc.__traceback__ = tb
```
A5rocks, line 48: I haven't put much thought into this so I might be wrong, but I think that using outcome would decrease the number of trio frames in all Python versions: this would mean nicer (but not optimally nice) tracebacks in pre-3.11 without messing around with tracebacks here.

clint-lawrence: The general change proposed makes sense to me. However, getting the traceback neatly formatted doesn't look straightforward.

I had a quick play using `outcome`. This has the advantage that the sync and async traceback have the same number of frames that need to be remove. So no need to use the count variable to know how many levels to strip.

```python
    def runcode(self, code: types.CodeType) -> None:
        try:
            func = types.FunctionType(code, self.locals)
            if inspect.iscoroutinefunction(func):
                trio.from_thread.run(outcome.acapture, func).unwrap()
            else:
                trio.from_thread.run_sync(outcome.capture, func).unwrap()
        except SystemExit:
            raise
        except BaseException as exc:
            self.showtraceback()
```

The `InteractiveConsole.showtraceback()` already strips one level of the traceback. So I tried copying that into the `TrioInteractiveConsole` subclass and modifying it to strip two levels. I think this works. One complication is the original logic checks to see if `sys.excepthook` has been set and if so, doesn't strip the traceback frame. However in any python prior to 3.12, Trio includes `exceptiongroup` which *does* override the default `sys.excepthook`.

In the implementation below, I'm stripping the two stack frame even when delegating to `sys.excepthook`. That is a little different to the stdlib version, but perhaps is a pragmatic middle ground.

```python
    def showtraceback(self):
        sys.last_type, sys.last_value, last_tb = ei = sys.exc_info()
        sys.last_traceback = last_tb
        # see https://docs.python.org/3/library/sys.html#sys.last_exc
        if sys.version_info >= (3, 12):
            sys.last_exc = ei[1]  
        
        try:
            lines = traceback.format_exception(ei[0], ei[1], last_tb.tb_next.tb_next)
            if sys.excepthook is sys.__excepthook__:
                self.write(''.join(lines))
            else:
                # pre 3.12, exceptgroup installs its own hook, so this
                # path gets taken.
                # was -> sys.excepthook(ei[0], ei[1], last_tb)
                sys.excepthook(ei[0], ei[1], last_tb.tb_next.tb_next)
        finally:
            last_tb = ei = None
```

richardsheridan: As long as we are using outcome to clean up tracebacks, we can save further machinations by not raising and catching the error again, see fff91de.

richardsheridan:
A really clean (IMO) implementation breaks certain features of `InteractiveConsole` intended to enable customization in subclasses. So I made it `final` to discourage people from trying and noted in comments which things won't work for future contributors to this class.

clint-lawrence (approved):
This looks good to me.

A5rocks (approved):

CoolCat467 (approved):
Looks good, only thing I can think of is that comment after `isinstance(result.error, SystemExit):` is a bit interesting:

> So we print the exception and stay in the repl

Not immediately obvious that raising the error in that branch causes error to be only printed and not close repl if I am reading the comment right.

richardsheridan:
> Not immediately obvious that raising the error in that branch causes error to be only printed and not close repl if I am reading the comment right.

Yes, I think I did a poor job of splitting up that comment. I'm going clarify that a bit and then someone can do a nice squash merge for us.

PR #3010, opened 2024-06-06
Change `fail_after`&`move_on_after` to set deadline relative to entering. Add CancelScope.relative_deadline

### LINKED ISSUE #2512: `fail_after` deadline is set on initialization not context entry
This is a duplicate of https://github.com/agronholm/anyio/issues/514 — they are following the convention established here in trio.

When using a `fail_after` context, the deadline is set at "initialization" time rather than `__enter__`. This behavior can result in unexpected bugs if the context is declared before it is entered. This is not particularly intuitive for a Python context manager — I'd expect the timer to start when the context is entered. I do not think changing it would be complex, but it could be considered breaking.

```python
import trio


async def main():
    ctx = trio.fail_after(5)
    await trio.sleep(5)
    with ctx:
        for i in range(1, 6):
            print(i)
            await trio.sleep(1)


trio.run(main)
```

```
❯ python example.py  
1
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_timeouts.py", line 106, in fail_at
    yield scope
  File "/Users/mz/dev/prefect/example.py", line 168, in main
    await trio.sleep(1)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_timeouts.py", line 76, in sleep
    await sleep_until(trio.current_time() + seconds)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_timeouts.py", line 57, in sleep_until
    await sleep_forever()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_timeouts.py", line 40, in sleep_forever
    await trio.lowlevel.wait_task_rescheduled(lambda _: trio.lowlevel.Abort.SUCCEEDED)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_core/_traps.py", line 166, in wait_task_rescheduled
    return (await _async_yield(WaitTaskRescheduled(abort_func))).unwrap()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/outcome/_impl.py", line 138, in unwrap
    raise captured_error
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_core/_run.py", line 1222, in raise_cancel
    raise Cancelled._create()
trio.Cancelled: Cancelled

  File "/Users/mz/dev/prefect/example.py", line 171, in <module>
    trio.run(main)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_core/_run.py", line 2010, in run
  File "/Users/mz/dev/prefect/example.py", line 168, in main
    await trio.sleep(1)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/contextlib.py", line 137, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-39/lib/python3.9/site-packages/trio/_timeouts.py", line 108, in fail_at
    raise TooSlowError
trio.TooSlowError
```

jakkdl:
~~Partially deals with #2512 by documenting current behavior.~~

This PR now works to fully resolve #2512 by adding a `relative_deadline` attribute to `CancelScope`, which is then used by `fail_after` and `move_on_after`. Upon entering the `CancelScope` the absolute deadline is then calculated as `min(deadline, current_time() + relative_deadline)`.

This PR also reverts an old workaround in `timeouts.py` due to [pycharm being unable to resolve `@contextmanager`](https://youtrack.jetbrains.com/issue/PY-36444/PyCharm-doesnt-infer-types-when-using-contextlib.contextmanager-decorator) for type checking.

TODO:
* [x] undo changes to/fix docstrings of `fail_after` and `move_on_after`
* [x] write docstring for `_RelativeCancelScope`
* [x] ~~resolve `None` vs `math.inf` for no deadline~~
* [x] settle on behavior ~~of `relative_deadline` after entering the cm. See https://github.com/python-trio/trio/issues/2512#issuecomment-2180329727~~


jakkdl:
(ignore the commit history, I created the branch in a messy way. Will squash the PR anyway)

src/trio/_timeouts.py
```diff
@@ -27,6 +27,9 @@ def move_on_after(seconds: float) -> trio.CancelScope:
     """Use as a context manager to create a cancel scope whose deadline is
     set to now + *seconds*.
```
A5rocks, line 28: Maybe say "set to creation time + \*seconds\*"? Alternate phrasing to what you proposed.

jakkdl: That would be cleaner. I was slightly worried about misinterpreting when the sentence starts with "Use as a context manager...", but I think that's a stretch.

jakkdl: Should `fail_after` also get a similar treatment? Something like
> Creates a cancel scope with the given timeout, relative to creation time, and raises an error if it is actually cancelled.

or maybe we want one of them to be super explicit

A5rocks (approved):

CoolCat467 (approved):
Makes sense

Zac-HD:
Per my comment on the issue,

> I think it would make sense for `trio.fail_after(x)` to be relative to entering the context manager; `trio.fail_at(trio.current_time() + x)` seems a more natural way to express the relative-to-construction-time semantics. (and the same for `trio.move_on_{at,after}`)
> 
> Either way, we should document the chosen semantics.

While I think that documenting this is a net improvement, it might also lead people to rely on the creation-time semantics for `*_after` in a way that makes changing that to enter-time semantics harder later.  Would you be up for making that change and documenting it as a one-step PR?

jakkdl:
> Per my comment on the issue,
> 
> > I think it would make sense for `trio.fail_after(x)` to be relative to entering the context manager; `trio.fail_at(trio.current_time() + x)` seems a more natural way to express the relative-to-construction-time semantics. (and the same for `trio.move_on_{at,after}`)
> > Either way, we should document the chosen semantics.
> 
> While I think that documenting this is a net improvement, it might also lead people to rely on the creation-time semantics for `*_after` in a way that makes changing that to enter-time semantics harder later. Would you be up for making that change and documenting it as a one-step PR?

Sure. Started thinking about different ways of implementing it in the issue: https://github.com/python-trio/trio/issues/2512#issuecomment-2157955978

jakkdl:
1. See https://github.com/python-trio/trio/issues/2512#issuecomment-2180329727 for discussion on the behavior of `relative_deadline` after entering.
2. I personally prefer `None` over `math.inf` quite a bit, but changing the behavior of `CancelScope.deadline` would be breaking for anybody accessing the value of the property in the `inf`/`None` case - so I should probably just change `relative_deadline` to also use `inf` instead of `None`.
3. Need to write docstring for `relative_deadline`, but can probably mostly just take the stuff I wrote in the newsfragment.
4. Need to update docstrings for `move_on_after` and `fail_after` to be explicit about deadline being relative to entering. (i.e. undo/invert what was originally changed in this PR).


The pycharm bug referred to in timeouts seem to be fixed https://youtrack.jetbrains.com/issue/PY-36444/PyCharm-doesnt-infer-types-when-using-contextlib.contextmanager-decorator and released over a year ago, so I think we can drop the workaround.

newsfragments/2512.breaking.rst
```diff
@@ -0,0 +1,18 @@
+:func:`trio.move_on_after` and :func:`trio.fail_after` now sets the deadline upon entering the context manager, instead of at initialization. The vast majority of usage initialized them when entering and is therefore unaffected. But if you have code that looks like:
```
A5rocks, line 1:
```suggestion
:func:`trio.move_on_after` and :func:`trio.fail_after` now set the deadline upon entering the context manager, instead of at initialization. The vast majority of usage initialized them when entering and is therefore unaffected. But if you have code that looks like:
```

newsfragments/2512.breaking.rst
```diff
@@ -0,0 +1,18 @@
...
+
+.. code-block:: python3
+
+    cs = trio._move_on_after(5)
+    ...
+    with cs:
+        ...
```
A5rocks, line 8: Couldn't we detect (most of) this usage? Might help this work better. IE compare construction time vs usage time? (and then a kwarg to silence + enable this new behavior?)

I'd just like something a bit more backwards compatible or at least that doesn't fail silently if people used it wrong.

newsfragments/2512.breaking.rst
```diff
@@ -0,0 +1,18 @@
+:func:`trio.move_on_after` and :func:`trio.fail_after` now sets the deadline upon entering the context manager, instead of at initialization. The vast majority of usage initialized them when entering and is therefore unaffected. But if you have code that looks like:
+
+.. code-block:: python3
+
+    cs = trio._move_on_after(5)
```
A5rocks, line 5:
```suggestion
    cs = trio.move_on_after(5)
```

src/trio/_core/_run.py
```diff
@@ -540,6 +540,11 @@ class CancelScope:
 
     # Constructor arguments:
     _deadline: float = attrs.field(default=inf, kw_only=True, alias="deadline")
+    # use float|None, or math.inf?
+    # or take it as an opportunity to also introduce deadline=None?
+    _relative_deadline: float | None = attrs.field(
+        default=None, kw_only=True, alias="relative_deadline"
+    )
```
A5rocks, line 550: I don't think `deadline` + `relative_deadline` should be supported: we could deny that using a post-init thing from `attrs`. Maybe someone has a use case for it? I can't imagine anything.

src/trio/_core/_run.py
```diff
@@ -550,6 +555,13 @@ def __enter__(self) -> Self:
                 "Each CancelScope may only be used for a single 'with' block"
             )
         self._has_been_entered = True
+
+        if self._relative_deadline is not None:
+            if self._relative_deadline + current_time() < self._deadline:
+                self._deadline = self._relative_deadline + current_time()
+            else:
+                self._relative_deadline = self.deadline - current_time()
```
A5rocks, line 566: We don't need to update `_relative_deadline` I think.

src/trio/_core/_run.py
```diff
@@ -744,6 +756,25 @@ def deadline(self, new_deadline: float) -> None:
...
+            return self._deadline - current_time()
+        return self._relative_deadline
+
+    @relative_deadline.setter
+    def relative_deadline(self, relative_deadline: float | None) -> None:
+        self._relative_deadline = relative_deadline
+        if self._has_been_entered:
+            with self._might_change_registered_deadline():
+                if relative_deadline is None:
+                    self._deadline = inf
```
A5rocks, line 770: I think this is confusing in conjunction with allowing people to set both `deadline` and `relative_deadline`.

A5rocks (commented):
Two major comments (somewhat touched on in the code comments):
 - I think this could be done with backwards compat
 - I really think that the interaction between `deadline` and `relative_deadline` is non-obvious. Prior to entering the context manager, setting one might override the other, depends on the other value. But when in the context manager, setting one will always override the other. I think we should disallow interacting with both through runtime errors; I can't think of any design that would fix this.

src/trio/_core/_run.py
A5rocks: cc @Zac-HD because you have opinions.

I think what @Zac-HD proposed of disallowing setting relative deadlines inside the cancel scope would help somewhat, even though I still think that absolute deadlines vs relative deadlines is a weird concept to have before the context manager. (So namely being able to pass both here and being able to set both with their properties)

src/trio/_core/_run.py
jakkdl: ah yeah, that's a relic of previous version.

newsfragments/2512.breaking.rst
jakkdl: Yeah that's probably a good call :+1:

jakkdl: Adding the functionality to `move_on_after` was quite straightforward.... but `fail_after` is being a pita, since `@contextmanager` makes the function not get executed at all until we try to enter it. The only way I could figure out how to resolve it was to change `fail_after` to be a full-on class-based contextmanager.

jakkdl:
RTD is failing because towncrier appends link to the github issue at the end of the newsfragments, but since we're inside a code block that leads to invalid python syntax. Trying to append newlines to the `rst` file doesn't help, they get stripped. Not sure the code blocks are *really* needed, could probably just move that stuff into the docstrings, but lets deal with that when we've decided on the precise implementation.

src/trio/_core/_run.py
```diff
@@ -543,16 +543,60 @@ class CancelScope:
...
                 "Each CancelScope may only be used for a single 'with' block"
             )
+
+        if (
+            self._creation_time is not None
+            and abs(self._creation_time - current_time()) > 0.01
+            and not self._timeout_from_enter
+        ):
+            # not using warn_deprecated because the message template is a weird fit
+            # TODO: mention versions in the message?
```
A5rocks, line 574: I think the trio deprecation util does this

src/trio/_timeouts.py
```diff
@@ -123,11 +141,14 @@ def fail_at(deadline: float) -> AbstractContextManager[trio.CancelScope]:  # typ
         raise TooSlowError
 
 
-if not TYPE_CHECKING:
-    fail_at = contextmanager(fail_at)
+# This was previously a simple @contextmanager-based cm, but in order to save
+# the current time at initialization it needed to be class-based.
+# Once that change of behaviour has gone through and the old functionality is removed
+# it can be reverted.
```
A5rocks, line 147: Might be worth putting a comment alongside related deprecation warning because someone might just find usages of the deprecation function and remove them rather than searching more in the file.

src/trio/_core/_run.py
jakkdl: possibly, or I could add some kwarg that overrides wording. But since not using `TrioDeprecationWarning` I don't really feel like `warn_deprecated` adds much of anything in a lot of cases
https://github.com/python-trio/trio/blob/a7db0e43ed9f5bbccabed7a06bf3717e3f8d093e/src/trio/_deprecate.py#L54-L75

src/trio/_core/_run.py
Zac-HD: My preferred semantics are:

- `CancelScope` gets a new `._relative_deadline: float` property, with no corresponding init argument
- the `*_after()` scopes create a `CancelScope` with `deadline=inf`, and then set the attribute
- on entry to the `CancelScope`, set the deadline to `min(deadline, _relative_deadline + current_time())`, and then set `_relative_deadline = inf`
  - this means that you get the same semantics for setting `deadline` as any other scope, but "relative deadlines" don't become a semantic thing for users to think about
- it's an error to set the `_relative_deadline` if the scope has ever been entered

The existing deprecation logic is a bit ugly, but seems like the best option available overall.

newsfragments/2512.deprecated.rst
```diff
@@ -0,0 +1,9 @@
+:func:`trio.move_on_after` and :func:`trio.fail_after` previously set the deadline relative to initialization time, instead of more intuitively upon entering the context manager. Any code where these two points in time do not line up (according to :func:`trio.current_time`) will now issue a `DeprecationWarning`. To silence the warning and adopt the new behaviour, pass ``timeout_from_enter = True``. If you want to retain current behaviour, you should instead use :func:`trio.move_on_at`/:func:`trio.fail_at`. For example:
+
+.. code-block:: python
+
+    # this is equivalent to how trio.move_on_after(5) would function previously
+    cs = trio.move_on_at(trio.current_time() + 5)
+    ...
+    with cs:
+        ...
```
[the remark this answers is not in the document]
jakkdl: turns out an issue was recently opened for it: https://github.com/twisted/towncrier/issues/614

src/trio/_core/_run.py
jakkdl: Why is this your preferred semantics? I'm mostly just seeing downsides to having the feature be this hidden, but still able to sneakily introduce confusing behavior if a user is sending around a `CancelScope`. If I have a function that takes a `CancelScope` as a parameter, I'd *have* to introspect the hidden attribute if I wanted to figure out what it would set the deadline to once entered. Maybe that's uncommon logic-wise, but surely people are writing logging statements like `print(f"this cs will time out at {cs.deadline}")` (before entering).

If we don't want to introduce relative deadlines as a primary feature, and avoid any stealthy gotchas, then I strongly think we should introduce a class that transforms into a `CancelScope` - to make sure that typing would catch any attempt at passing/saving the return value of `fail/move_on_after` to something that doesn't expect precisely that, and raise `AttributeError` if attempting to access `deadline`.

jakkdl:
I gave it a shot to instead implement this as a transforming class. The basic implementation is way cleaner, but the big downside is that it will break code that currently does

```python
cs = trio.move_on_after(5)
with cs:
  cs.deadline = 7  # AttributeError: _RelativeCancelScope does not have an attribute `deadline`
# you need to instead write
with cs as cs2:
  cs2.deadline = 7
```

There's a couple ways to fix that, though it's gonna be somewhat ugly. Either properties/methods that mirror the properties/methods, or dunder magic with `__getattr__`/`__setattr__`. We can make them invisible to type checkers and raise warnings though.

src/trio/_core/_run.py
jakkdl: Oh, the `min` variant also breaks trying to change the deadline, unless we unset `relative_deadline` upon setting `deadline`. So ultimately there's no reason to use `min` and we instead do some conditionals on whether values are `None`/`inf`.

```python
cs = move_on_after(5)
# wait no I want to make the deadline later
cs.deadline = trio.current_time() + 7
with cs:
    print(cs.deadline - trio.current_time()) # prints 5
```
this is regardless of if the time is relative to construction or entering.

and of course
```python
cs = move_on_after(5)
cs.deadline = cs.deadline + 2
with cs:
    print(cs.deadline)  # prints 5, not 7, which it would previously
```

Zac-HD: I now agree that my proposal upthread is clearly bad 😅

jakkdl:
Wrote a section in the documentation and added `_RelativeCancelScope`. We might want to consider re-export it as a non-hidden class for the sake of making type hints more ergonomic.
Updated newsfragments and docstrings to match current implementation.

The docs and docstrings wouldn't complain about a second pass to improve phrasing, but otherwise I think this is mostly ready to be merged.

jakkdl:
I have no clue why verifytypes is complaining about `fail_at`, and cannot repro it locally. But might want to restore the previous workaround, also because the documentation now displays the return type of the function as `Generator[trio.CancelScope, None, None]`

jakkdl:
(working on fixing the TODO for accessing deadline&shield on `_RelativeCancelScope` in case somebody is quick on reviewing)

jakkdl:
Perhaps somewhat overkill on transitional functions, but this should be very robust.
* I feel like supporting `cancel()` on unentered relative cancel scopes is sort of overkill, but think it can be implemented fine - unless the time at which it's actually cancelled matters.
* should `shield` raise `AttributeError` after entering, if called on the `_RelativeCancelScope`?

Opened https://github.com/microsoft/pyright/discussions/8905, think we're gonna end up ignoring it unless there's a quick fix released.

jakkdl:
The `pyright --verifytypes` issue has been fixed in https://github.com/microsoft/pyright/pull/8924 so just need to wait for version 1.1.380

Zac-HD:
I'm sorry this has taken me so long to get to; vacation & illness took their toll!  I'm aiming for a full review over the weekend, but here are some initial thoughts while I have them:

- I think that having two transitions (add deprecation, remove deprecation) is worse for users than just switching immediately to the new behavior.  Anyone who wants the previous behavior can already switch `fail_after(n)` to `fail_at(current_time() + n)` and have that backwards-compatible.
- I have a fairly strong aesthetic preference to _only_ have a `CancelScope` class, adding new `is_relative: bool | None` and `relative_deadline` attributes.
  - `deadline` and `relative_deadline` become properties, with setters that adjust a shared internal state in the obvious way - once the scope has been entered this is easy.
  - _before_ the scope is entered, I'd make it an error to access the relative-deadline on a non-relative scope.  Accessing the deadline on a relative scope before entry can raise a deprecation warning; assigning is deprecated _and_ converts to a non-relative scope.
  - better for introspection in IDEs, avoids the reassignment issues above, avoids duplicating some tricky code that might get out of sync later, etc.
- Assigning an `inspect.Signature()` object to the `__signature__` attribute of a callable does the trick for Sphinx, and can be guarded by an `if "sphinx" in sys.modules:` to save the important latency the rest of the time.  Neat trick to keep typecheckers _and_ docs users happy 😉

jakkdl:
I just realized:
```python
my_move_on_at = trio.move_on_at(5)
my_move_on_after = trio.move_on_after(5)
my_fail_at = trio.fail_at(5)
my_fail_after = trio.fail_after(5)
print(my_move_on_at.shield) # works
print(my_move_on_after.shield) # works
# these give AttributeError: _GeneratorContextManager object has no attribute shield
print(my_fail_at.shield)
print(my_fail_after.shield)
```
due to `fail_at` & `fail_after` being implemented as wrapping context managers. If people start passing around relative cancel scopes this might pop up as a pain point.

jakkdl:
> I'm sorry this has taken me so long to get to; vacation & illness took their toll! I'm aiming for a full review over the weekend, but here are some initial thoughts while I have them:

:heart:

> * I think that having two transitions (add deprecation, remove deprecation) is worse for users than just switching immediately to the new behavior.  Anyone who wants the previous behavior can already switch `fail_after(n)` to `fail_at(current_time() + n)` and have that backwards-compatible.

The downside to this is it being a silent breaking change, and if a library starts relying on it they will have differing behavior depending on what trio version it runs against. So will definitely need a "major" version bump.

> * _before_ the scope is entered, I'd make it an error to access the relative-deadline on a non-relative scope.  Accessing the deadline on a relative scope before entry can raise a deprecation warning; assigning is deprecated _and_ converts to a non-relative scope.
This does resolve some of the instances of silent breakage, though not all. I'm not sure I think a `DeprecationWarning` is 
> * better for introspection in IDEs, avoids the reassignment issues above, avoids duplicating some tricky code that might get out of sync later, etc.

This seems fine, and if end users really do want different typing between absolute and relative scopes they can get it with a protocol with `is_relative: Literal[True]` (which we could even add ourselves)

> * Assigning an `inspect.Signature()` object to the `__signature__` attribute of a callable does the trick for Sphinx, and can be guarded by an `if "sphinx" in sys.modules:` to save the important latency the rest of the time.  Neat trick to keep typecheckers _and_ docs users happy 😉

For which issue was this?

I pushed a commit that changes the implementation, I have not updated all docs/docstrings/newsfragments pending further changes.

src/trio/_core/_run.py
```diff
@@ -540,11 +540,21 @@ class CancelScope:
...
     _cancel_called: bool = attrs.field(default=False, init=False)
+    _is_relative: bool | None = attrs.field(default=False, init=False)
     cancelled_caught: bool = attrs.field(default=False, init=False)
 
     # Constructor arguments:
-    _deadline: float = attrs.field(default=inf, kw_only=True, alias="deadline")
-    _shield: bool = attrs.field(default=False, kw_only=True, alias="shield")
+    _relative_deadline: float = attrs.field(default=inf, kw_only=True)
+    _deadline: float = attrs.field(default=inf, kw_only=True)
+    _shield: bool = attrs.field(default=False, kw_only=True)
```
Zac-HD, line 549: I think we want to keep the non-underscored `alias=` for each of these.

src/trio/_core/_run.py
```diff
@@ -734,13 +750,48 @@ def deadline(self) -> float:
         this can be overridden by the ``deadline=`` argument to
         the :class:`~trio.CancelScope` constructor.
         """
+        if self._is_relative is True:
+            raise RuntimeError(
+                "unentered relative cancel scope does not have an absolute deadline",
+            )
```
Zac-HD, line 756: We could make this a bit smoother: emit a deprecation warning, and then return current_time() + relative_deadline

src/trio/_core/_run.py
```diff
@@ -734,13 +750,48 @@ def deadline(self) -> float:
...
         return self._deadline
 
     @deadline.setter
     def deadline(self, new_deadline: float) -> None:
+        if self._is_relative is True:
+            raise RuntimeError(
+                "unentered relative cancel scope does not have an absolute deadline",
+            )
```
Zac-HD, line 764: Here too, I'd emit a deprecation warning but set the deadline as expected, and if-not-entered also set ~~is_relative=False~~ _relative_deadline=inf.

src/trio/_core/_run.py
```diff
@@ -554,6 +564,12 @@ def __enter__(self) -> Self:
                 "Each CancelScope may only be used for a single 'with' block",
             )
         self._has_been_entered = True
+        assert self._is_relative is not None
+
+        if self._is_relative:
+            self._deadline = current_time() + self._relative_deadline
+            self._is_relative = None
```
Zac-HD, line 571: need to dedent this

src/trio/_tests/test_timeouts.py
```diff
@@ -173,3 +184,63 @@ async def test_timeouts_raise_value_error() -> None:
...
+
+    mock_clock.jump(3)
+    start = _core.current_time()
+    with rcs as cs:
+        assert cs.deadline == start + 5
+
+        assert rcs is cs
+
+
+async def test_invalid_acces_unentered() -> None:
```
Zac-HD, line 221:
```suggestion
async def test_invalid_access_unentered() -> None:
```

src/trio/_core/_run.py
```diff
@@ -540,11 +540,21 @@ class CancelScope:
     _has_been_entered: bool = attrs.field(default=False, init=False)
     _registered_deadline: float = attrs.field(default=inf, init=False)
     _cancel_called: bool = attrs.field(default=False, init=False)
+    _is_relative: bool | None = attrs.field(default=False, init=False)
```
Zac-HD, line 543: I think we can actually avoid tracking this attribute at all, and instead define:

```python
@property
def is_relative(self):
    if self.deadline < inf and not self._has_been_entered:
        assert self.relative_deadline == inf
        return True
    if self.relative_deadline < inf and not self._has_been_entered:
        return False
    return None
```

Zac-HD:
edited my comment above to point to the possible docs issue.  I wouldn't include the protocol, if anyone wants that they can define it downstream.

A5rocks:
> The downside to this is it being a silent breaking change, and if a library starts relying on it they will have differing behavior depending on what trio version it runs against. So will definitely need a "major" version bump.

Yeah, not sure this is great. Even across a large version bump. ... Is there any actual use case for constructing `move_on_after` before the `async with`? If not, we could just warn about the deprecation for that specifically (check whether CancelScope was constructed too long before `__enter__`, no way to opt out), and not have to un-deprecate anything! (after a version or two, we could remove the deprecation -- or switch it to be a hard error.)

Zac-HD:
Annoyingly, I think there's no good reason to construct a `move_on_after` in advance _yet_: with the current semantics you can more clearly express the behavior with a `move_on_at(current_time() + n)`.  Unfortunately the whole reason we're doing this is that you might want to express something like "when you eventually send this request, fail if it takes more than five seconds".

I guess we could make a deprecation-only release and then un-deprecate it when the semantics change, but I think it's less disruptive overall to just switch to the new behavior 😕

A5rocks:
> I guess we could make a deprecation-only release and then un-deprecate it when the semantics change, but I think it's less disruptive overall to just switch to the new behavior 😕

That's actually a great way to do this. Though if users *want* relative timeouts and realize this upon seeing the deprecation message, then they have no real alternative (other than restructuring things) which does suck. (... which is better than just silently being incorrect? but not as good as silently becoming correct -- the deprecation is for people who don't want relative timeouts).

... Alternatively, we could mention passing `relative_timeout` to `CancelScope` in the deprecation message and just hope that having to turn `trio.move_on_after(5)` into `trio.CancelScope(relative_timeout=5)` isn't very common. (not great, though stylistic things like that can always be caught by `flake8-async` once deprecation period is over).

Zac-HD:
It just seems extremely strange for someone who doesn't want a relative timeout to use `move_on_after()`, so I still favor a one-step update where we just make that do the right thing.

Good call on the linter though: we could have a lint rule against calling `*_after()` without immediately entering it, which only fires if the Trio version is older than whatever this patch ends up being.  And then add a config option for `trio_version` (and `anyio_version`?), which defaults to whatever is installed.

A5rocks:
Yeah ok I agree that it's fine to do this without deprecation. It's not that big of a behavior change (your code runs for just a bit longer) and deprecation strategies range from counter-productive to very weird (making people use the relative_timeout parameter). And if the linter could catch old usages that would be very nice.

src/trio/_core/_run.py
jakkdl: The non-underscored explicit alias doesn't do anything, attrs automatically exposes them as non-underscored init parameters unless you specify `init=False`.
We could keep them for the sake of being extra explicit, but attrs docs don't do it and it could lead to confusion in the case of somebody not having an alias.

jakkdl:
> Good call on the linter though: we could have a lint rule against calling `*_after()` without immediately entering it, which only fires if the Trio version is older than whatever this patch ends up being. And then add a config option for `trio_version` (and `anyio_version`?), which defaults to whatever is installed.

I think gating it on Trio version isn't great, because it's fairly easy to get mismatched versions between what your linter/ci setup is running (very easy if you're using pre-commit), and if you're supporting a range of trio versions you almost surely won't have different environments that lints against different versions of dependencies. So I'd go with a default-enabled check, that mentions that you should disable it if you only support Trio>0.x (and at some point we might default-disable that check)
Opened https://github.com/python-trio/flake8-async/issues/290, we can have further linter-specific discussions in there.

jakkdl:
>     * Assigning an `inspect.Signature()` object to the `__signature__` attribute of a callable does the trick for Sphinx, and can be guarded by an `if "sphinx" in sys.modules:` to save the important latency the rest of the time.  Neat trick to keep typecheckers _and_ docs users happy 😉
>       (edit: to address [Add `_RelativeCancelScope`. Change `fail_after`&`move_on_after` to set deadline relative to entering #3010 (comment)](https://github.com/python-trio/trio/pull/3010#issuecomment-2331056189) )

Neat! But I now realized we in fact probably shouldn't do this because it does actually matter: https://github.com/python-trio/trio/pull/3010#issuecomment-2346000138

The only way (other than adding a wrapping _FailingCancelScope :upside_down_face:) I can see to fix it is adding a parameter to `CancelScope` that makes it raise `TooSlowError` when canceled. But that's perhaps for another issue

src/trio/_core/_run.py
jakkdl: nice, that works.
Re `< inf` vs `!= inf`: we made the timeout helper functions have special handling for `nan`, but not `CancelScope`... oops

src/trio/_core/_run.py
jakkdl: not relevant since removal of `self._is_relative`

src/trio/_core/_run.py
jakkdl: For this one I have no problems with raising a warning instead.

src/trio/_tests/test_timeouts.py
```diff
@@ -169,7 +180,79 @@ async def test_timeouts_raise_value_error() -> None:
     ):
         with pytest.raises(
             ValueError,
-            match="^(duration|deadline|timeout) must (not )*be (non-negative|NaN)$",
+            match="^(duration|deadline|timeout|relative deadline) must (not )*be (non-negative|NaN)$",
         ):
```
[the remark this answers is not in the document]
Zac-HD: I think I'd just update the error messages to refer to the argument name in all cases, using backticks to make it clear that this is an argument name if needed (e.g. maybe for "seconds"?)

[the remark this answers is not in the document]
jakkdl: Since both `move_on_after` and `CancelScope` want to validate the same input parameter but currently have a different name for it, that would mean one of these has to be true:
* `move_on_after(seconds=-1)` gives `"relative_deadline" can't be negative`
* we duplicate validation code between `move_on_after` and `CancelScope.__attrs_post_init__`
* `move_on_after` catches and reraises the exception from `CancelScope` solely to replace the parameter name in the error message.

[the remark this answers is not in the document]
Zac-HD: Duplicating validation logic seems least bad of these options 🤔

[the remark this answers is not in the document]
TeamSpen210: There is another option we could do. Given how many names this parameter has, we could just add an additional parameter to `CancelScope` itself, with the name to use when reporting errors. Maybe that clutters the API too much though. But it does seem anyone else wrapping might also want to change the error reporting.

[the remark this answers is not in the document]
jakkdl: that sounds quite overcomplicated imo, esp since that would probably include both `deadline` and `relative_deadline`. If users are wrapping it and want pretty errors I don't think it's unreasonable to expect them to catch and reraise

jakkdl:
Once https://github.com/python-trio/flake8-async/pull/292 is merged I'll update the newsfragment to refer to `ASYNC122`, otherwise all docs/docstrings/etc should be good now.

Zac-HD (approved):

PR #3011, opened 2024-06-06
[docs] clarify checkpoint semantics for trio.open_nursery

jakkdl:
Documents the changes made in https://github.com/python-trio/trio/pull/1696 more explicitly.

### DIFF
--- a/docs/source/reference-core.rst
+++ b/docs/source/reference-core.rst
@@ -102,6 +102,8 @@ them. Here are the rules:
     only that one will act as a checkpoint. This is documented
     on a case-by-case basis.
 
+    * :func:`trio.open_nursery` is a further exception to this rule.
+
 * Third-party async functions / iterators / context managers can act
   as checkpoints; if you see ``await <something>`` or one of its
   friends, then that *might* be a checkpoint. So to be safe, you
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -993,7 +993,8 @@ def open_nursery(
     new `Nursery`.
 
     It does not block on entry; on exit it blocks until all child tasks
-    have exited.
+    have exited. If no child tasks are running on exit, it will insert a
+    schedule point (but no cancellation point).
 
     Args:
       strict_exception_groups (bool): Unless set to False, even a single raised exception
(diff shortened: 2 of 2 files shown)


docs/source/reference-core.rst
```diff
@@ -102,6 +102,8 @@ them. Here are the rules:
     only that one will act as a checkpoint. This is documented
     on a case-by-case basis.
 
+    * :func:`trio.open_nursery` is a further exception to this rule.
```
CoolCat467, line 105: Not sure if this is needed, I think the `Partial exception for async context managers` block handles this

jakkdl: It may be overkill, but `open_nursery` in itself as a context manager doesn't guarantee a full checkpoint on either of entry or exit - so imo it doesn't abide by the special rules in the `Partial exception for async context managers` block. There's some more discussion around this in #1457

A5rocks (approved):

src/trio/_core/_run.py
```diff
@@ -993,7 +993,8 @@ def open_nursery(
     new `Nursery`.
 
     It does not block on entry; on exit it blocks until all child tasks
-    have exited.
+    have exited. If no child tasks are running on exit, it will insert a
+    schedule point (but no cancellation point).
```
TeamSpen210, line 997: Since this is unique behaviour, maybe it'd be good to add an additional sentence explaining why this is the case - something like "A nursery is never the source of the cancellation exception, it only propagates it from sub-tasks."

jakkdl: Oh actually, is that not functionally the same as https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.cancel_shielded_checkpoint ?
If so, we can refer to that for a more thorough explanation

A5rocks: I don't think just referring to that builds any kind of intuition for why open_nursery is a special case. In fact I forgot the exact wording/problem that convinced me, but I think it was some message on the gitter (that pushed the old PR to be merged).

TeamSpen210 (approved):

src/trio/_core/_run.py
A5rocks: Not exactly, but I think docs as is are good enough:tm:

CoolCat467 (approved):
Makes sense

PR #3017, opened 2024-06-12
Add Python 3.11 and 3.12 to Windows and MacOS matrices

A5rocks:
Part of: https://github.com/python-trio/trio/issues/2885


A5rocks:
Oh, turns out the github branch rename feature doesn't work. Oh well

A5rocks:
pre-commit.ci autofix

CoolCat467 (approved):
Looks good to me, simple enough

jakkdl:
Huh, did we just forget to add these when adding 3.11/3.12 - or was something broken that now works? Not that it matters, good stuff.

A5rocks:
It looks like I added 3.12 at least and while I can't remember exactly, I'm pretty sure I just forgot to add it to Windows/MacOS (because step 1 was adding `3.12-dev` which we handle using a PPA, then step 2 (where I forgot) is just removing `-dev`)

PR #3019, opened 2024-06-16
Use Typevar defaults for `TaskStatus` and `Matcher`

TeamSpen210:
This causes annotations to default to `TaskStatus[None]`, `RaisesGroup[BaseException]` and `Matcher[BaseException]`, instead of `Any`. [`_socket`](https://github.com/python-trio/trio/blob/451393acec35c3a01cba0c19de6bf8f6f23dbc7b/src/trio/_socket.py#L41-L52) could also use this to make `SocketType` generic over `AddressFormat`, but I'm not really sure how that should behave.

### DIFF
--- a/src/trio/_util.py
+++ b/src/trio/_util.py
@@ -203,23 +203,6 @@ def coroutine_or_error(
     return coro
 
 
-class EmptySuperclass:
-    """Used as a runtime placeholder for classes which should be imported only during TYPE_CHECKING.
-
-    This can be subscripted with anything, then gets discarded entirely from the base classes list.
-    """
-
-    def __getitem__(self, item: object) -> EmptySuperclass:
-        return self
-
-    def __mro_entries__(
-        self,
-        bases: tuple[type[object], ...],
-    ) -> tuple[type[object], ...]:
-        """This is discarded."""
-        return ()
-
-
 class ConflictDetector:
     """Detect when two tasks are about to perform operations that would
     conflict.
--- a/src/trio/_util.py
+++ b/src/trio/_util.py
@@ -203,6 +203,23 @@ def coroutine_or_error(
     return coro
 
 
+class EmptySuperclass:
+    """Used as a runtime placeholder for classes which should be imported only during TYPE_CHECKING.
+
+    This can be subscripted with anything, then gets discarded entirely from the base classes list.
+    """
+
+    def __getitem__(self, item: object) -> EmptySuperclass:
+        return self
+
+    def __mro_entries__(
+        self,
+        bases: tuple[type[object], ...],
+    ) -> tuple[type[object], ...]:
+        """This is discarded."""
+        return ()
+
+
 class ConflictDetector:
     """Detect when two tasks are about to perform operations that would
     conflict.
--- a/src/trio/testing/_raises_group.py
+++ b/src/trio/testing/_raises_group.py
@@ -15,7 +15,7 @@ from typing import (
 )
 
 from trio._deprecate import warn_deprecated
-from trio._util import EmptySuperclass, final
+from trio._util import final
 
 if TYPE_CHECKING:
     import builtins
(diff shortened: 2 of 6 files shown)


TeamSpen210:
Seems `RaisesGroup` needs to be `Any` right now, otherwise a bunch of [type-tests break](https://github.com/TeamSpen210/trio/actions/runs/9531969991/job/26273545636) for Pyright. @jakkdl looks like it might not actually be inferring the generic, and previously I think it wasn't generic at all.

A5rocks:
`RaisesGroup` types are current all sorts of messed up IMO, though we still have to wait on https://github.com/python/mypy/pull/16753 + a release of mypy before making them better...

src/trio/_util.py
```diff
@@ -203,6 +203,23 @@ def _return_value_looks_like_wrong_library(value: object) -> bool:
     return coro
 
 
+class EmptySuperclass:
+    """Used as a runtime placeholder for classes which should be imported only during TYPE_CHECKING.
+
+    This can be subscripted with anything, then gets discarded entirely from the base classes list.
+    """
+
+    def __getitem__(self, item: object) -> EmptySuperclass:
```
A5rocks, line 212: (nitpick) really this is `Self`.

src/trio/testing/_raises_group.py
```diff
@@ -254,13 +267,15 @@ def __str__(self) -> str:
...
-    SuperClass = Generic
+    # At runtime, just discard this base class.
+    SuperClass = EmptySuperclass()
 
 
 @final
-class RaisesGroup(ContextManager[ExceptionInfo[BaseExceptionGroup[E]]], SuperClass[E]):
+class RaisesGroup(
+    SuperClass[E], ContextManager[ExceptionInfo[BaseExceptionGroup[E]]], Generic[E]
+):
```
A5rocks, line 278: Why do we need `Generic` here?

A5rocks (commented):
The code makes sense, though I'm not sure either `RaisesGroup` or (especially) `_ExceptionInfo` is used in type hints enough to make typevar defaults make sense. Definitely for `TaskStatus` though.

A5rocks:
Turns out the stuff necessary for `RaisesGroup`-but-good is merged in mypy! After the next mypy release I'll take my chances on nicer typing. I don't think it's worth thinking too much about type var defaults for them right now.

jakkdl:
> Seems `RaisesGroup` needs to be `Any` right now, otherwise a bunch of [type-tests break](https://github.com/TeamSpen210/trio/actions/runs/9531969991/job/26273545636) for Pyright. @jakkdl looks like it might not actually be inferring the generic, and previously I think it wasn't generic at all.

could you maybe add a test (with `type: ignore` or `pyright: ignore`) that demonstrates the issue, even without default?

src/trio/testing/_raises_group.py
TeamSpen210: I was thinking that generic classes need the `Generic` subclassing, but they don't. So some of this can be removed.

jakkdl: I might have added `Generic` as a superclass precisely because I wanted a superclass that did nothing (because its other superclasses themselves already inherit from it)

TeamSpen210:
Looking at the error again, it might just be the mypy issue we're already aware of. I'll just revert the changes to `RaiseGroup` for this PR.

src/trio/_util.py
TeamSpen210: Probably should have made this final, but actually I don't need it since I can remove `Generic`.

jakkdl:
It would be nice to have some addition to the typing tests that demonstrate what has changed. 

```python
def check_typevar_default(e: Matcher) -> None:
    assert e.exception_type is not None
    f = e.exception_type()
    # this would previously pass, as `f` would be `Any`
    f = 5  # type: ignore[assignment]
def check_typevar_default_explicit(e: Matcher) -> None:
    assert_type(e.exception_type, Optional[type[BaseException]])
```


Other than that it looks great.

A5rocks (approved):

jakkdl (approved):

PR #3022, opened 2024-06-25
Add PosArgT typing to trio.run

jakkdl:
I noticed this when going through the trio-typing test suite. I don't see any reason why this would've been intentionally skipped in #2881, so I suspect it just got missed?


Zac-HD (approved):
Looks good to me!

Should we consider moving over some of the trio-typing test suite, since apparently that catches some issues that the trio tests don't?

TeamSpen210 (approved):
Must have been missed, yes.

A5rocks (approved):
Maybe add a type test.

Feel free to merge this whether or not you add one though.

CoolCat467 (approved):
Looks great!

jakkdl:
> Looks good to me!
> 
> Should we consider moving over some of the trio-typing test suite, since apparently that catches some issues that the trio tests don't?

There's not much that's worth adapting, I don't think we want to adapt their method of [running internal mypy functions](https://github.com/python-trio/trio-typing/blob/master/trio_typing/_tests/test_typecheck.py), so it would consist of copying and adapting the [test-data](https://github.com/python-trio/trio-typing/tree/master/trio_typing/_tests/test-data). But we already cover most everything:
* We don't care for async_generator tests
* outcome [has its own type tests](https://github.com/python-trio/outcome/blob/master/tests/type_tests.py)
* we already have type tests for `trio.Path` and `TaskStatus`
* trio-basic:
  * we have type tests for open_memory_channel
  * we have type tests for nursery.start_soon

The only things left were `run`, that `trio-typing` raised a typing error if you tried to instantiate `Cancelled`, `Nursery.start`, and some trivial stuff that would raise typechecking errors when we run mypy on our test suite.

PR #3024, opened 2024-06-28
Ruff preview mode

CoolCat467:
This pull request enables ruff's preview mode.

Currently a work in progress, but I wanted to see if I've messed anything up in the unit tests.


src/trio/_core/_tests/test_mock_clock.py
```diff
@@ -92,7 +92,7 @@ async def test_mock_clock_autojump(mock_clock: MockClock) -> None:
     mock_clock.autojump_threshold = 0
     # if the above line didn't take affect immediately, then this would be
     # bad:
-    await sleep(100000)
+    await sleep_forever()
```
TeamSpen210, line 95: This isn't correct, the code is testing that the autojump clock is able to skip a large (but finite) amount of sleeping.

src/trio/_tests/test_ssl.py
```diff
@@ -711,7 +711,7 @@ async def sleep_then_wait_writable() -> None:
 
     async def sleeper_with_slow_wait_writable_and_expect(method: str) -> None:
         if method == "wait_send_all_might_not_block":
-            await trio.sleep(100000)
+            await trio.sleep_forever()
         elif method == "expect":
             await trio.sleep(1000)
```
jakkdl, line 716: surely both or neither of these should be changed, and not one of them.

src/trio/_tests/test_threads.py
```diff
@@ -895,6 +895,7 @@ async def main(use_system_task: bool) -> None:
 
 
 async def test_trio_token_weak_referenceable() -> None:
+    await _core.checkpoint()
     token = _core.current_trio_token()
     assert isinstance(token, _core.TrioToken)
     weak_reference = weakref.ref(token)
```
jakkdl, line 901: ~~I don't think there's any reason to enable ASYNC910/async-function-no-checkpoint on test files. Or is this a different check saying "async function with no awaits can be sync?"~~
oh this is RUF029/unused-async

src/trio/_tests/test_unix_pipes.py
```diff
@@ -279,4 +279,12 @@ async def test_bizarro_OSError_from_receive() -> None:
...
-    await check_one_way_stream(make_pipe, make_clogged_pipe)
+    def make_pipe_async() -> tuple[FdStream, FdStream]:
+        await _core.checkpoint()
+        return make_pipe()
+
+    def make_clogged_pipe_async() -> tuple[FdStream, FdStream]:
+        await _core.checkpoint()
+        return make_clogged_pipe()
+
+    await check_one_way_stream(make_pipe_async, make_clogged_pipe_async)
```
jakkdl, line 290: needing to duplicate `make_[clogged_]pipe` and having worthless checkpoints in them both looks awful. I vote for reverting

src/trio/_tests/test_util.py
```diff
@@ -147,7 +147,7 @@ def generator_based_coro() -> Any:  # pragma: no cover
 
         assert "appears to be synchronous" in str(excinfo.value)
 
-        async def async_gen(_: object) -> Any:  # pragma: no cover
+        async def async_gen(_: object) -> Any:  # pragma: no cover  # noqa: RUF029
```
jakkdl, line 150: add comment explaining what RUF029 is

src/trio/testing/_fake_net.py
```diff
@@ -466,15 +466,16 @@ async def send(self, data: Buffer, flags: int = 0) -> int:
...
     @overload
     async def sendto(
         self,
-        __data: Buffer,
-        __flags: int,
-        __address: tuple[object, ...] | str | None | Buffer,
+        data: Buffer,
+        flags: int,
+        address: tuple[object, ...] | str | None | Buffer,
+        /,
```
jakkdl, line 478: yeahno this will give typing errors. There's reasons for those double underscores, see comments in the socket file - might be good to copy them here.

jakkdl (commented):
I think I'd vote weakly against using ruff preview.
If you just want to run CI checks it's enough to push to a branch in this repo, without opening a PR. (or even better run tests locally ofc)

src/trio/_core/_tests/test_mock_clock.py
CoolCat467: Fixed in new version

src/trio/_tests/test_ssl.py
CoolCat467: Fixed in new version

src/trio/_tests/test_threads.py
CoolCat467: Fixed in new version

src/trio/_tests/test_unix_pipes.py
CoolCat467: Reverted in new version

src/trio/_tests/test_util.py
CoolCat467: Fixed in new version

src/trio/testing/_fake_net.py
CoolCat467: Fixed in new version with inline entire file ignore with comment

CoolCat467:
`hash` argument fixes being merged in this should fix what caused previous ci runs to fail

A5rocks:
I was looking at whether ruff catches a forgotten prefix for fstrings like `"hello {name}"` and found `RUF027`. It seems to be under preview, so I'm for this change now!

TeamSpen210:
Maybe we should enable `explicit-preview-rules` too. Then only rules we explicitly enable will be used. By definition preview rules might be incorrect, change, etc which would be bad if they suddenly start failing CI. Seems fine to run ruff ourselves when a new release comes out to try these rules though, most of the changes here are fine.

For these preview rules, we probably should disable `ANN004` for now, since it produces false positives for all `exceptiongroup` imports.
`RUF029` (async functions without any `await` etc) should probably be disabled for test code. We don't particularly care about guaranteeing checkpoints there, and often we'll be implementing tasks that just assert/raise, without any async logic. Constantly disabling the rule is just distracting noise.
Similarly `ASYNC116`, since we're using autojump it doesn't matter if sleeps are excessively long.

src/trio/_tests/test_ssl.py
```diff
@@ -707,7 +707,7 @@ async def expect(expected: bytes) -> None:
     # Our receive_some() call will get stuck when it hits send_all
     async def sleeper_with_slow_send_all(method: str) -> None:
         if method == "send_all":
-            await trio.sleep(100000)
+            await trio.sleep(100000)  # noqa: ASYNC116  # not sleep forever
```
A5rocks, line 710: Similar as for the other explanations for `ASYNC116`, could this explain the why rather than the what.

CoolCat467:
There are no `AN004`s raised with the most recent version of ruff

src/trio/testing/_memory_streams.py
```diff
@@ -347,7 +347,8 @@ def memory_stream_one_way_pair() -> tuple[MemorySendStream, MemoryReceiveStream]
     def pump_from_send_stream_to_recv_stream() -> None:
         memory_stream_pump(send_stream, recv_stream)
 
-    async def async_pump_from_send_stream_to_recv_stream() -> None:
+    # await not used
+    async def async_pump_from_send_stream_to_recv_stream() -> None:  # noqa
```
A5rocks, line 351: Missing code (but looks like RUF029).

A5rocks (commented):
Can we disable `RUF029`? We're very deliberate with what is marked `async` for `trio` and what isn't so nothing should be missed. Tests are a different story but I don't think it matters for them.  The false-positives are annoying! (and very noisy. probably like half of the line changes are adding noqa for this!)

Additionally, I don't think A005 brings any benefit. We *know* we're shadowing stdlib and that's intentional (well, except the underscore prefixed ones, which nobody cares about shadowing anyways).

src/trio/_tools/gen_exports.py
```diff
@@ -302,7 +302,7 @@ def process(files: Iterable[File], *, do_test: bool) -> None:
             print("Generated sources are up to date.")
     else:
         for new_path, new_source in new_files.items():
-            Path(new_path).write_text(new_source, encoding="utf-8", newline="\n")
+            Path(new_path).write_text(new_source, encoding="utf-8")
```
A5rocks, line 305: At least with `open`, `newline` is a good idea for Windows users so that `gen_exports.py` won't output files using `"\r\n"`. I'm not sure if that's the case or not with `Path#write_text`.

CoolCat467: Apparently, Path.write_text's `newline` argument was only added in 3.10 or something

A5rocks: Probably worth switching back to `open` then and then making an issue on ruff about availability of the parameter.

src/trio/_socket.py
```diff
@@ -1,3 +1,9 @@
+# ruff: noqa: A005
+# Module `_socket` is shadowing a Python builtin module
+
+# ruff: noqa: PYI063  # Several cases throughout file where
+# argument names with __ used because of typeshed, see comment for recv in _SocketType
```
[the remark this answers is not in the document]
jakkdl: Ooh this shouldn't be needed anymore. Opened #3094

src/trio/_core/_tests/test_asyncgen.py
```diff
@@ -189,7 +189,8 @@ def test_last_minute_gc_edge_case() -> None:
     record = []
     needs_retry = True
 
-    async def agen() -> AsyncGenerator[int, None]:
+    # async function missing await
+    async def agen() -> AsyncGenerator[int, None]:  # noqa: RUF029
```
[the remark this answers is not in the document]
CoolCat467: Disabled in d2a4031

src/trio/_core/_tests/test_run.py
```diff
@@ -1679,7 +1691,7 @@ async def main() -> None:
     async def f() -> None:  # pragma: no cover
         pass
 
-    async def async_gen(arg: T) -> AsyncGenerator[T, None]:  # pragma: no cover
+    async def async_gen(arg: T) -> AsyncGenerator[T, None]:  # pragma: no cover  # noqa
```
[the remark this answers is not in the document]
CoolCat467: RUF029, removed in d2a4031

src/trio/testing/_memory_streams.py
CoolCat467: RUF029, removed in d2a4031

src/trio/_tools/gen_exports.py
CoolCat467: Fixed in 924b6dd

src/trio/_tests/test_ssl.py
CoolCat467: Fixed in 924b6dd

src/trio/_socket.py
```diff
@@ -680,6 +686,7 @@ async def connect(self, address: AddressFormat) -> None:
     def recv(__self, __buflen: int, __flags: int = 0) -> Awaitable[bytes]:
         raise NotImplementedError
 
+    # argument names with __ used because of typeshed, see comment for recv in _SocketType
```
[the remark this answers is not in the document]
CoolCat467: Fixed in 924b6dd

src/trio/_core/_tests/test_mock_clock.py
```diff
@@ -92,7 +92,7 @@ async def test_mock_clock_autojump(mock_clock: MockClock) -> None:
     mock_clock.autojump_threshold = 0
     # if the above line didn't take affect immediately, then this would be
     # bad:
-    await sleep(100000)
+    await sleep(100000)  # noqa: ASYNC116  # not sleep_forever
```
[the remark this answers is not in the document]
CoolCat467: Fixed in 924b6dd

src/trio/_tests/test_subprocess.py
```diff
@@ -659,7 +659,10 @@ async def test_for_leaking_fds() -> None:
 async def test_run_process_internal_error(monkeypatch: pytest.MonkeyPatch) -> None:
     # There's probably less extreme ways of triggering errors inside the nursery
     # in run_process.
-    async def very_broken_open(*args: object, **kwargs: object) -> str:
+    async def very_broken_open(
+        *args: object,
+        **kwargs: object,
+    ) -> str:
```
A5rocks, line 665: Similar magic comma situation here

A5rocks (commented):
Left a few comments because some of the changes involved adding trailing commas. IIRC that's a separate PR? I left a comment on every file with the change rather than on every change.

pyproject.toml
```diff
@@ -147,10 +148,20 @@ extend-ignore = [
...
+# RUF029 is ignoring tests that are marked as async functions but
+# do not use an await in their function bodies. There are several
+# places where internal trio synchronous code relies on being
+# called from an async function, where current task is set up.
+'src/trio/_tests/*.py' = ['RUF029']
+'src/trio/_core/_tests/*.py' = ['RUF029']
+# A005 is ignoring modules that shadow stdlib modules.
+'src/trio/_abc.py' = ['A005']
+'src/trio/_socket.py' = ['A005']
+'src/trio/_ssl.py' = ['A005']
```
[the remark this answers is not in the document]
CoolCat467: I had the intention of disabling it on current triggers so that in the event a new module was added it would be triggered for the new one. To be honest I don't see that realistically happening any time soon, but that was my logic.

src/trio/_core/_tests/test_run.py
```diff
@@ -1658,7 +1658,9 @@ async def check(spawn_fn: Callable[..., object]) -> None:
     await check(_core.spawn_system_task)
 
 
-async def test_current_effective_deadline(mock_clock: _core.MockClock) -> None:
+async def test_current_effective_deadline(
+    mock_clock: _core.MockClock,
+) -> None:
```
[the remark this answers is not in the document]
CoolCat467: Fixed in d8ffd87f

src/trio/_tests/test_fakenet.py
```diff
@@ -64,7 +64,7 @@ async def test_msg_trunc() -> None:
     s2 = trio.socket.socket(type=trio.socket.SOCK_DGRAM)
     await s1.bind(("127.0.0.1", 0))
     await s2.sendto(b"xyz", s1.getsockname())
-    data, addr = await s1.recvfrom(10)
+    _data, _addr = await s1.recvfrom(10)
```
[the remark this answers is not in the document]
CoolCat467: Removed assignment in d8ffd87f

src/trio/_tests/test_subprocess.py
CoolCat467: Fixed in d8ffd87f

src/trio/_core/_tests/test_guest_mode.py
```diff
@@ -118,7 +118,9 @@ async def trio_return(in_host: InHost) -> str:
 
     assert trivial_guest_run(trio_return) == "ok"
 
-    async def trio_fail(in_host: InHost) -> NoReturn:
+    async def trio_fail(
+        in_host: InHost,
+    ) -> NoReturn:
```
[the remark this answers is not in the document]
CoolCat467: Fixed in d8ffd87f

pyproject.toml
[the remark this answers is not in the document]
A5rocks: Ah that would make sense. TBH using the built in setting of names to ignore might work with less noise for that case. I'm not sure what would be better.

A5rocks (approved):
Everything in my last review seems to be addressed.

I'm not quite sure I like having an explanation with every instance of `RUF029` but tbh just seeing that error code doesn't explain much -- I think having different formats of explanations visible at once is getting to me, while it wouldn't while programming (because I would only see a small portion of code).

CoolCat467:
Going to merge this soon unless anyone has any additional concerns.

TeamSpen210:
I think it would be good to merge the changes, but perhaps not actually enable `preview-mode`. By definition these rules are less certain, might have edge cases that don't work, or get reverted. So it'd be bad to update the pin, then suddenly have errors in CI that we might have to revert later. Instead just run them manually when there's a new version, see if we want to do anything at that point.

A5rocks:
> I think it would be good to merge the changes, but perhaps not actually enable `preview-mode`. By definition these rules are less certain, might have edge cases that don't work, or get reverted. So it'd be bad to update the pin, then suddenly have errors in CI that we might have to revert later. Instead just run them manually when there's a new version, see if we want to do anything at that point.

I think there are a few (like the fstring check rule) that would be nice to have, but maybe that's not worth the churn.

CoolCat467:
> I think it would be good to merge the changes, but perhaps not actually enable `preview-mode`. By definition these rules are less certain, might have edge cases that don't work, or get reverted. So it'd be bad to update the pin, then suddenly have errors in CI that we might have to revert later. Instead just run them manually when there's a new version, see if we want to do anything at that point.

That is a good point. What if instead, we specifically enabled all of the rules that are preview as of now, and then when new versions come out we can try adding the new ones then? Is it possible to do that with ruff?

TeamSpen210:
Looks like you can't no. You can specify them, but it won't enable. But we could enable them in preparation.

notes-to-self/socket-scaling.py
```diff
@@ -32,7 +32,7 @@ def pt(desc, *, count=total, item="socket"):
             now = time.perf_counter()
             total_ms = (now - last_time) * 1000
             per_us = total_ms * 1000 / count
-            print(f"{desc}: {total_ms:.2f} ms total, {per_us:.2f} µs/{item}")
+            print(f"{desc}: {total_ms:.2f} ms total, {per_us:.2f} μs/{item}")
```
graingert, line 35: what did this change?

TeamSpen210: Ahh, Ruff changed "U+00B5 MICRO SIGN" to "U+03BC GREEK SMALL LETTER MU". They look identical, but apparently the latter is the preferred encoding to use. According to section 2.5 of [this unicode doc](https://www.unicode.org/reports/tr25/#page11) at least.

CoolCat467: Correct

CoolCat467:
My current understanding of conversation is that it would be nice if we could specify which preview rules we would like to enable, but that is not possible as of writing. There was discussion about leaving the changes but not enabling preview mode, which I don't think would be a good solution because first if there are `noqa` lines for preview rule matches, then ruff's `unused-noqa` will get rid of them, and then we aren't checking future code against preview rules, leading to inconsistency. The idea behind leaving preview disabled is that new preview rules could be introduced which by definition are unstable and might not be good, but my counter is that we could always disable those specific new preview rules, but rightly said is that this will increase difficulty of updating ruff versions, but I don't think it would be too much of a hassle to disable said new rules if required.

Net result, I will continue with earlier stated merging, and in the event this is not the right course of action we can always revert this or apply a better solution on top of this.

PR #3030, opened 2024-07-05
Improve REPL KI

### LINKED ISSUE #3007: Ctrl+C behavior problems at the new REPL
It seemed a little too good to be true that Ctrl+C "just works" in the new Trio REPL (#2972, #3002), it turns out there are a couple weird problems. Interrupting running sync and async code seems to go fine, but when you are just sitting at the input prompt, KI cannot reach that code in the "usual" fashion. I've noticed 3 issues so far:

### `Runner._ki_pending` flag survives returning to the prompt

Consider executing a simple uninterruptible operation:

`>>> await trio.to_thread.run_sync(time.sleep, 5) `

This should and does last 5 seconds even if you try to interrupt it. However, Trio is still trying to get rid of the KI hot potato, so the next checkpoint will fire off a KI:

```
>>> await trio.sleep(0)
  File "<console>", line 1, in <module>
  <snip>
    raise KeyboardInterrupt
KeyboardInterrupt
>>>
```

This one seems to be easy enough to solve, either dig into the Runner and manually reset the flag every time we're done processing input, or run `trio.from_thread.run(trio.lowlevel.checkpoint_if_cancelled)` and it will safely propagate to the normal console KI code. However, after doing that exposed the next two issues.

### `input` on Windows sees Ctrl+C even though there is a handler AND it's not in the main thread

If you hit Ctrl+C any time the console is waiting for input, the Trio run ends:

```
>>> {Ctrl+C}
now exiting TrioInteractiveConsole...
  <snip>
KeyboardInterrupt
^C
C:\>
```

(Weirdly, this doesn't happen during PyCharm's terminal emulation!) This is very simply a result of `input` popping out `EOFError` as if you'd done Ctrl+Z+Enter:

```
C:\>python -c "import trio; trio.run(trio.to_thread.run_sync, input)" {then hit Ctrl+C}
  <snip>
    ret = context.run(sync_fn, *args)
EOFError

  <snip>
KeyboardInterrupt

C:\>
```

I suppose in the main thread this is overridden by the `KeyboardInterrupt` at some point for the python REPL behavior to work right. This seems like a matter of transforming the Exception in our console subclass:

```
    def raw_input(self, prompt=""):
        try:
            return input(prompt)
        except EOFError:
            # check if trio has a pending KI
            trio.from_thread.run(trio.lowlevel.checkpoint_if_cancelled)
            raise
```

This works fine as far as I can tell.

### `input` on Linux DOES NOT see Ctrl+C  (because it's not in the main thread)

This isn't so bad as it only means KI at the input prompt doesn't take effect until you finish entering a (possibly multiline) command, and even then the command runs to completion. This is weird, but maybe tolerable? I think the worst part is that you can't quickly discard/drop out of multiline edits. Furthermore, I have no idea how to fix this.

## Summary

This is an issue instead of a PR because there might be a smarter way to interact with KI that would also work on linux, and I wanted other people to follow up on this after experimenting a bit if there's any other weird behavior that isn't covered by the KI above checks (such as the linux case).

richardsheridan:
This commit has a draft method to address #3007. I have 0 ideas on how to test this thing: not only does it use the `raw_input` that gets patched for testing, but the way it injects a newline is basically impossible for a test runner to pass in. Also, it feels fragile somehow to dip in and out of the trio loop patching the signal handler and peeking for `KeyboardInterrupt`.


A5rocks:
Could we spawn the REPL as another process and send it a signal? (simulate ctrl+c)

I think we could first send it some data over stdin to get into specific scenarios.

richardsheridan:
The newline is sent to the terminal I think, rather than any specific process.

Potentially we could avoid the ioctl weirdness by allowing a small amount of user discomfort in needing to press enter after Ctrl+C.

richardsheridan:
Still no idea how to test the proposed fixes. I think it's relevant that cPython isn't able to interrupt the input thread either, unless it's using the new python repl:

https://github.com/python/cpython/blob/e65e9f90626a4c62da4d3500044f354b51e51dbb/Lib/asyncio/__main__.py#L132-L139

Was there ever talk about the `_pyrepl` going public? if so, we could use a version of that interrupt method instead.

src/trio/_repl.py
```diff
@@ -53,6 +53,42 @@ def runcode(self, code: types.CodeType) -> None:
...
+            import termios
+            from signal import SIGINT, signal
+
+            interrupted = False
+
+            def handler(sig: int, frame: types.FrameType | None) -> None:
+                nonlocal interrupted
+                interrupted = True
+                # Fake up a newline char as if user had typed it at terminal
+                fcntl.ioctl(sys.stdin, termios.TIOCSTI, b"\n")
```
graingert, line 83: you can't write to sys.stdin from a handler safely, I'm not sure if you can icotl to the .fileno() or not either, I think you need:


```python
@disable_ki_protection
def _newline():
    # Fake up a newline char as if user had typed it at terminal
    fcntl.ioctl(sys.stdin, termios.TIOCSTI, b"\n")


class TrioInteractiveConsole:
    token = trio.lowlevel.current_trio_token()

    def handler(...):
        nonlocal interrupted
        interrupted = True
        token.run_sync_soon(_newline, idempotent=True)
```

richardsheridan: Thanks! I think I understand why, but won't a keyboard interrupt in the newline function blow up trio and end the repl? If so, Would suppressing the exception be sufficient?

richardsheridan:
> Was there ever talk about the `_pyrepl` going public? if so, we could use a version of that interrupt method instead.

Uh turns out pyrepl is a separate package that cpython vendored: https://pypi.org/project/pyrepl/#history  

I have no intention of vendoring it or making a dependency for this pr but just FYI

A5rocks:
> The newline is sent to the terminal I think, rather than any specific process.

I was talking about just sending `SIGTERM` to a process running the REPL as a test that, well, the REPL supports ctrl+c. This makes it sound like this PR makes us depend on having a terminal attached (and maybe even behavior of that terminal) which sounds not great. If so, then there's probably no way other than maybe using tmux as a parent for the REPL and seeing if that does this special behavior?

A5rocks:
I would be happy to push this to the finish line. Here's a couple things I noticed:
 1. the extra newline after pressing ctrl+c after a prompt irritates me more than it should. After spending more time than is reasonable, I think the best approach is making `interrupted` be stored on the class and overwriting `write`.
 2. I started to work on the test case I was thinking about. It seems to imply we need the *user* to enter a newline or that the ioctl thing only works for shells (?). (I don't need to enter a newline when testing using my terminal...):
```py
import trio
import sys
import subprocess
import signal
from functools import partial

async def main():
        proc = await nursery.start(partial(
            trio.run_process,
            [sys.executable, "-m", "trio"],
            # shell=True,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            stdin=subprocess.PIPE,
        ))

        async with proc.stdout, proc.stderr:
            # setup
            await anext(proc.stderr)
            assert await anext(proc.stdout) == b'>>> '

            # ensure things work
            await proc.stdin.send_all(b'print("hello!")\n')
            assert await anext(proc.stdout) == b'hello!\n>>> '

            # ensure that ctrl+c on the REPL works
            proc.send_signal(signal.SIGINT)
            print("stdout:")
            with trio.move_on_after(0.5):
                async for c in proc.stdout:
                    print(c)

            print("stderr:")
            with trio.move_on_after(0.5):
                async for c in proc.stderr:
                    print(c)

            await proc.stdin.send_all(b'\n')
            print("stdout:")
            with trio.move_on_after(0.5):
                async for c in proc.stdout:
                    print(c)

            print("stderr:")
            with trio.move_on_after(0.5):
                async for c in proc.stderr:
                    print(c)

        # kill the process
        nursery.cancel_scope.cancel()

trio.run(main)
```
output:
```
stdout:
stderr:
stdout:
b'>>> '
stderr:
b'KeyboardInterrupt\n'
```

Generally this looks great though.

richardsheridan:
Thanks for taking interest in completing this!

> 1. the extra newline after pressing ctrl+c after a prompt irritates me more than it should. After spending more time than is reasonable, I think the best approach is making `interrupted` be stored on the class and overwriting `write`.

This is an effort/reward balance that wouldn't quite make the cut for me, but I won't try to stop you!

> 2. I started to work on the test case I was thinking about. It seems to imply we need the _user_ to enter a newline or that the ioctl thing only works for shells (?). (I don't need to enter a newline when testing using my terminal...):

I thought the ioctl somehow uses a kernel api to send bytes through the process's attached terminal so i'm not surprised that it only works in some cases, but I also don't know exactly what those cases would be without trying. What happened to your `tmux` idea?

A5rocks:
> What happened to your `tmux` idea?

Unfortunately that doesn't seem to help -- I've tried both `shell=True` (which implies `sh -c "..."`) and explicitly using `tmux -c "..."`, but the ioctl call doesn't seem to be inserting a newline. I suppose it's fine to not be able to test that, but that's really not ideal...

A5rocks:
Ok it seems this relies on having a PTY, which is something we can provide it. Maybe one day CPython will export pyrepl so we can do what asyncio does and take advantage of the fact they use a reimplemented readline...

A5rocks:
Turns out the reason my test code above wasn't working is somewhat trivial: `terminal_newline` is suppressing OSError!

So far I get:
 - using `subprocess.PIPE` for stdin, etc gives OSError 25 "Inappropriate ioctl for device" (makes sense...)
 - using a PTY yields OSError 1 "Operation not permitted"

The latter seems closer, but I'm not quite getting Linux's permission model 🤔

A5rocks:
It's actually impossible to test this in CI... because it doesn't work in CI! Linux 6.2 introduced a `sysctl` to make the `TIOCSTI` ioctl require superuser (`CAP_SYS_ADMIN`), and guess what Ubuntu 24.04 enabled :/ (I didn't notice because my Debian system is on Linux 6.1...)

You can check your own system with `sysctl dev.tty.legacy_tiocsti`. I guess we can just say the implementation is best-effort and call it a day? We can keep the test that ctrl+c works elsewhere + inject a newline in the test case itself for non-Windows. This is still an improvement over status quo I suppose...

Please review this!

EDIT: I really dislike how... not great the tests are. I think we should rely on vendored pyrepls given we support only CPython/PyPy, especially since the functionality we need is basically at a "rewrite readline" level. Though any changes should be on another PR.

richardsheridan:
> It's actually impossible to test this in CI... because it doesn't work in CI! Linux 6.2 introduced a `sysctl` to make the `TIOCSTI` ioctl require superuser (`CAP_SYS_ADMIN`), and guess what Ubuntu 24.04 enabled :/ (I didn't notice because my Debian system is on Linux 6.1...)

I'll try to review this weekend, but just quickly wanted to ask if this means that eventually the ioctl will break on (essentially) all linux users?

A5rocks:
Only if the distro disables the ioctl, so maybe not all? But yeah we will need to brainstorm another solution...

src/trio/_repl.py
```diff
@@ -27,7 +27,9 @@ def terminal_newline() -> None:
     import termios
 
     # Fake up a newline char as if user had typed it at the terminal
-    fcntl.ioctl(sys.stdin, termios.TIOCSTI, b"\n")
+    # on a best-effort basis
+    with contextlib.suppress(OSError):
+        fcntl.ioctl(sys.stdin, termios.TIOCSTI, b"\n")  # type: ignore[attr-defined, unused-ignore]
```
richardsheridan, line 31: Instead of simply suppressing OSError, maybe catch it and prompt the user to press enter to continue by printing text? We could also get the OSError code at that time which could be useful for understanding the situation.

src/trio/_tests/test_repl.py
```diff
@@ -239,3 +242,179 @@ def test_main_entrypoint() -> None:
...
+
+            print(buffer.decode())
+
+            # ensure that ctrl+c on a prompt works
+            # NOTE: for some reason, signal.SIGINT doesn't work for this test.
+            # Using CTRL_C_EVENT is also why we need subprocess.CREATE_NEW_PROCESS_GROUP
+            signal_sent = signal.CTRL_C_EVENT if sys.platform == "win32" else signal.SIGINT  # type: ignore[attr-defined,unused-ignore]
+            os.kill(proc.pid, signal_sent)
+            if sys.platform == "win32":
+                # we rely on EOFError which... doesn't happen with pipes.
```
richardsheridan, line 382: I never dug in to why EOFError pops out, maybe that knowledge could help us test here

richardsheridan (commented):
I suppose the tests make sense to me overall, it's just unfortunate that there are still some holes in them. If it is just a best-effort holdover until a pyrepl based solution can be stitched together then I don't think that should be a blocker. It's still better than the present behavior for all users.

src/trio/_tests/test_repl.py
A5rocks: I haven't dug into this yet unfortunately.

A5rocks:
I want to do this but still haven't:
 - make the test skip only if the relevant sysctl is enabled
 - add a bunch of codecov skips....

Sorry about the delay, I've been on vacation!

src/trio/_repl.py
A5rocks: Hopefully my change worked, I'm not on a Linux machine so I cannot check yet.

geofft:
This is a bit of a driveby comment, sorry, but TIOCSTI is indeed going to be blocked by most distros. The upstream kernel has a strong warning against it in drivers/tty/Kconfig:

```
config LEGACY_TIOCSTI
        bool "Allow legacy TIOCSTI usage"
        default y
        help
          Historically the kernel has allowed TIOCSTI, which will push
          characters into a controlling TTY. This continues to be used
          as a malicious privilege escalation mechanism, and provides no
          meaningful real-world utility any more. Its use is considered
          a dangerous legacy operation, and can be disabled on most
          systems.

          Say Y here only if you have confirmed that your system's
          userspace depends on this functionality to continue operating
          normally.

          Processes which run with CAP_SYS_ADMIN, such as BRLTTY, can
          use TIOCSTI even when this is set to N.

          This functionality can be changed at runtime with the
          dev.tty.legacy_tiocsti sysctl. This configuration option sets
          the default value of the sysctl.
```

Also, at least on my machine, a newline after Ctrl-C doesn't actually _interrupt_, it sends the current line for execution just as if I had not hit Ctrl-C.

 I think this is actually going to be hard to get right. The old REPL using Modules/readline.c wants to get interrupted with `EINTR`, but it also wants to specifically see `PyErr_CheckSignals()` return nonzero, which it simply will not do not on the main thread. The new REPL also does a little bit too much I/O on its own.

One option is to actually do the blocking I/O on the main thread. I know that's distasteful, but as I understand it, the REPL doesn't give you a nursery, so the only things that would be blocked are system tasks, right?

One option is to do the blocking I/O on the main thread, and take advantage of the fact that both REPLs call some hooks every 100 ms and call into Trio's guest mode. I think you can do this in pure Python for the new REPL; for the old REPL I think you can abuse ctypes against the running Python interpreter to set PyOS_InputHook but it's a little weird.

One option might be to try to get Trio to run on a secondary thread?

The nicest option would be to handle console I/O in Trio itself. For the old REPL you want to use the readline "alternate interface" (`rl_callback_handler_install` and `rl_callback_read_char`), which Modules/readline.c uses, but not quite flexibly enough. Follow the logic in that file, and call those two C functions via ctypes. For the new REPL there's a small chunk of code to copy/paste and sprinkle some awaits on, but at least it's pure Python.

A5rocks:
> One option is to actually do the blocking I/O on the main thread.

I think we discussed it a bit and this fails because someone may run this in their REPL:
```py
while True:
  pass
```

and we want to interrupt that.



> For the old REPL you want to use the readline "alternate interface" (`rl_callback_handler_install` and `rl_callback_read_char`), which Modules/readline.c uses

Huh does that work with libeditline or whatever the MacOS thing is?

---

Ultimately I think this is probably the best compromise for now, and then we rely on vendored pyrepls which have reimplemented readline in Python which means any sort of changes is easy. (this is what the asyncio REPL relies on)

geofft:
#3306 is an extremely messy attempt at doing this for the old REPL via calling readline ourselves.

> Huh does that work with libeditline or whatever the MacOS thing is?

The code in Modules/readline.c in the Python source seems to use them unconditionally, and the symbols appear to exist in editline, so I think so.

(But the more annoying thing is that this general approach doesn't work with python-build-standalone's builds i.e. what you get from uv and a few other tools. While we're using libedit, that's not the problem here; the problem is we're statically linking it and don't re-export those symbols. That is... something I'm going to have to think about, we have a similar problem with people who want to get at the underlying Tcl/Tk symbols from tkinter.)

geofft:
Re `while True: pass`, I think that would work fine with what I'm thinking - the blocking I/O is only when reading input, i.e., `InteractiveConsole.raw_input()`. Once that's over, you're back in normal async code and Ctrl-C works the way it normally does.

To try it out, change `line = await self.async_input(prompt)` in #3306 back to `line = self.raw_input(prompt)`, but keep the rest of the changes to run everything on the main thread. Seems to work for me. (If you want to go this route, you can ditch all the weird ctypes stuff in that PR, the relevant change is just that you want `async def ainteract()` instead of calling the existing `interact()` in a thread.

richardsheridan:
> I know that's distasteful, but as I understand it, the REPL doesn't give you a nursery, so the only things that would be blocked are system tasks, right?

The system nursery is available and may be expected to function for dispatching background tasks for both the interactive repl user and any number of libraries, so I think this is a nonstarter. unless maybe you do it by polling?

src/trio/_repl.py
A5rocks: Looks like this works now:
```
Press enter! Newline injection failed: [Errno 25] Inappropriate ioctl for deviceKeyboardInterrupt
```

A5rocks:
Turns out setting a sysctl on github actions seems not to persist (or my check for skipping tests is wrong. Probably the prior though).

Anyways this has been a lot of work for something that really doesn't matter that much. I'm going to fix up any codecov issues (by adding ignores), and hopefully then we can merge this!

At some point I want to check out https://github.com/python-trio/trio/pull/3306 / look at using pyrepl....

PR #3032, opened 2024-07-08
Use `--universal` flag with uv pip compile

CoolCat467:
Follow up to https://github.com/python-trio/trio/pull/2958

> We might want to try out this new `--universal` flag (I'm not sure when it was added? but seems useful)
> 
> ... but that can be a followup. This is good as a drop in replacement.

_Originally posted by @A5rocks in https://github.com/python-trio/trio/pull/2958#pullrequestreview-2161941408_

This pull request is adding the new  `--universal` flag for the uv pip compile runs.


docs-requirements.txt
```diff
@@ -12,12 +12,18 @@ beautifulsoup4==4.12.3
...
-    # via cryptography
+cffi==1.16.0 ; os_name == 'nt' or platform_python_implementation != 'PyPy'
+    # via
+    #   -r docs-requirements.in
+    #   cryptography
 charset-normalizer==3.3.2
     # via requests
 click==8.1.7
     # via towncrier
+colorama==0.4.6 ; platform_system == 'Windows' or sys_platform == 'win32'
```
A5rocks, line 23: Nice!

A5rocks (approved):
The changes look good, I can't speak of whether they're accurate/missing things, but definitely something nice. I'm fine merging this despite there not being many docs about it, but once this is documented more we can probably close the issue RE: replacing `pip-compile`.

A5rocks (changes requested):
I'm not sure this is stable enough to use as is. I ran `uv pip compile --universal --no-strip-markers --python-version=3.8 --upgrade
 test-requirements.in -o test-requirements.txt` twice locally and it started duplicating requirements in the requirements file.

A5rocks:
Ok so while investigating that I figured out that I can't fix it (the fix would require too much change in `uv` I think).

However! 1) the reason we have different results is because this wasn't a problem prior to 0.2.19, and 2) a workaround (because we want to stay current) is to just put Sphinx at the top of the requirements file. Though I haven't double checked for 2.

check.sh
```diff
@@ -78,10 +78,10 @@ fi
 
 # Check pip compile is consistent
 echo "::group::Pip Compile - Tests"
-uv pip compile --no-strip-markers --python-version=3.8 test-requirements.in -o test-requirements.txt
+uv pip compile --universal --no-strip-markers --python-version=3.8 test-requirements.in -o test-requirements.txt
```
A5rocks, line 81: I think `--universal` implies `--no-strip-markers` but I haven't checked yet. (I was checking which is why I noticed the duplicated requirements entries before)

A5rocks (changes requested):

check.sh
CoolCat467: Can't seem to find `--universal` mentioned anywhere in the docs, so I'm not sure either.

CoolCat467:
While looking through the documentation, I came across https://github.com/astral-sh/uv-pre-commit, might be interesting to try out using that.

check.sh
A5rocks: Yeah the docs are out of date. I suspect they'll stay this way at least until uv actually releases their "uv tool" thing.

jakkdl:
> While looking through the documentation, I came across https://github.com/astral-sh/uv-pre-commit, might be interesting to try out using that.

I think we want to bump dependencies in pull requests separate from other PRs, so adding a pre-commit check that plays around with requirements sounds like a bad idea to me.

A5rocks (approved):

check.sh
CoolCat467: Removed `--no-strip-markers` in d3b388c

CoolCat467:
I think there is a case of test flakiness again, the merge run failed on pypy-3.10 with a dtls issue again:
```
   ______________________ test_gc_before_system_task_starts _______________________
  
      @pytest.mark.filterwarnings("always:unclosed DTLS:ResourceWarning")
      async def test_gc_before_system_task_starts() -> None:
          e = endpoint()
      
  >       with pytest.warns(ResourceWarning):
  E       Failed: DID NOT WARN. No warnings of type (<class 'ResourceWarning'>,) were emitted.
  E        Emitted warnings: [].
  
  /opt/hostedtoolcache/PyPy/3.10.14/x64/lib/pypy3.10/site-packages/trio/_tests/test_dtls.py:757: Failed
```
I think something similar to this happened recently elsewhere as well, might need to make an issue collecting test flakiness for pypy-3.10

A5rocks:
Maybe we just have to bump GC counts yet again.

PR #3042, opened 2024-07-31
Fix trio.socket.SocketType.bind to work with PathLike arguments again

### LINKED ISSUE #3041: Regression: unix socket bind does not work with a PathLike argument
trio.socket.SocketType.bind no longer works with a PathLike (e.g. pathlib.Path) argument.

Simple test program:
```python
import socket, trio.socket
from pathlib import Path

async def main():
    sock = trio.socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
    await sock.bind(Path('/tmp/x.sock'))

trio.run(main)
```

Results in:
```
  File "/tmp/x.py", line 8, in <module>
    trio.run(main)
  File "/usr/lib/python3.12/site-packages/trio/_core/_run.py", line 2305, in run
  File "/tmp/x.py", line 6, in main
    await sock.bind(Path('/tmp/x.sock'))
  File "/usr/lib/python3.12/site-packages/trio/_socket.py", line 874, in bind
    address = await self._resolve_address_nocp(address, local=True)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/trio/_socket.py", line 922, in _resolve_address_nocp
    return await _resolve_address_nocp(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/trio/_socket.py", line 472, in _resolve_address_nocp
    assert isinstance(address, (str, bytes))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

This bugs seems to have been introduced in [cf1f3c745](https://github.com/python-trio/trio/commit/cf1f3c745#diff-300b958089de6b62596dad8182b7dfb234ac01d922fdbce3456f641c5a9964e1R466), released in v0.23.0.

regnarg:
fixes #3041


A5rocks (commented):
Could you add a regression test? I'm not sure exactly how to, but it's probably fine to use a tmpfile as a socket and actually do the IO.

---

EDIT: it looks like we test `open_unix_socket` already, maybe just make the name types a 2x2 parametrized thing? (not sure about keeping it in `test_highlevel_open_unix_stream` but honestly it's either that or `test_socket` and that's better.)

https://github.com/python-trio/trio/blob/20f92914da0a28a6e0ac6eb17dd7fa56c6602cc0/src/trio/_tests/test_highlevel_open_unix_stream.py#L54-L76

CoolCat467:
pre-commit.ci autofix

regnarg:
> Could you add a regression test?

It took a few tries but test is there and seems to pass.

CoolCat467:
Why so many force pushes?

regnarg:
Because I did not figure out how to successfully run the tests locally so it took me several tries to get the CI to pass. And I do not need to clutter commit history with trivial fixes of trivial mistakes.

CoolCat467:
Not really an issue, we use squash and merge most of the time I think

A5rocks (approved):

A5rocks:
> Because I did not figure out how to successfully run the tests locally so it took me several tries to get the CI to pass. And I do not need to clutter commit history with trivial fixes of trivial mistakes.

Was there any specific issues? We should make this better for new contributors. Was it just a matter of finding the right pytest invocation?

A5rocks (changes requested):
Sorry, I keep forgetting: can you add a changelog entry? (A newsfragment)

A5rocks:
@regnarg just a bump for this PR!

A5rocks:
Could someone else check the newsfragment looks fine?

CoolCat467 (approved):
Looks good to me. 
```python
import os
reveal_type(os.fspath)
```
with mypy yields
```
note: Revealed type is "Overload(
    def (path: builtins.str) -> builtins.str,
    def (path: builtins.bytes) -> builtins.bytes,
    def [AnyStr in (builtins.str, builtins.bytes)] (path: os.PathLike[AnyStr`-1]) -> AnyStr`-1
)"
```
so return is indeed str or bytes if pathlike is given

Issue #2837, opened 2023-10-26
Mark pure reformatting PRs in .git-blame-ignore-revs


jakkdl:
This is mostly relevant to @CoolCat467 as you're in the progress of cleaning up the repository, but I realize we in general haven't been great at keeping that file updated.

See discussion in https://github.com/python-trio/trio/pull/2686 and #2962

It would be great if one could have a check that warned about this ... but I have no clue how that would work in any way, and I'm also not sure if it's possible to learn the commit ID's before merging such that you can actually add a PR to the file before it gets merged.

A5rocks:
> I'm also not sure if it's possible to learn the commit ID's before merging such that you can actually add a PR to the file before it gets merged.

I don't think so, because it'll use the current timestamp. Is there anything you think should get ignored? I looked at the blame for the abc file and the only ones I see are at least 4 years old.

jakkdl:
> > I'm also not sure if it's possible to learn the commit ID's before merging such that you can actually add a PR to the file before it gets merged.
> 
> I don't think so, because it'll use the current timestamp.

I feel like it should be possible with some post-merge github action or something, but probs not worth setting up anything overly complicated.

> Is there anything you think should get ignored? I looked at the blame for the abc file and the only ones I see are at least 4 years old.

Did a quick look and there's a bunch of recent PR's that are somewhat borderline, like #2930 and #2932. #2823 has a mix of blocks of reformatting and other stuff.
Not sure about #3024. #3044 definitely merits it.

If we *really* wanted to we could split partial-reformatting PR's into two commits that we merge onto master, ignoring one of them, instead of squashing.

jakkdl:
https://github.com/python-trio/trio/commit/acdb7b2c424a2f91f8f98e63010e423b2dc49358 should be added imo. It's type-checking and not pure formatting, but I don't think it'll ever be useful to hit in a `git blame` (as I just did)

PR #3044, opened 2024-08-01
Enable ruff's `flake8-commas` rule

CoolCat467:
In this pull request, we enable ruff's `flake8-commas` rule and handle all the reformatting black does because of this.

~No functional changes~ Had to update `gen_exports` to run black a 2nd time, if/when this is merged it will probably need to be added to the purely formatting commits thing jakkdl was talking about a while ago.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,4 @@
 # sorting all imports with isort
 933f77b96f0092e1baab4474a9208fc2e379aa32
+# enabling ruff's flake8-commas rule
+b25c02a94e2defcb0fad32976b02218be1133bdf
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -132,7 +132,8 @@ def autodoc_process_signature(
             # and insert the fully-qualified name.
             signature = signature.replace("+E", "~trio.testing._raises_group.E")
             signature = signature.replace(
-                "+MatchE", "~trio.testing._raises_group.MatchE"
+                "+MatchE",
+                "~trio.testing._raises_group.MatchE",
             )
         if "DTLS" in name:
             signature = signature.replace("SSL.Context", "OpenSSL.SSL.Context")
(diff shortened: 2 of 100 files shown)


jakkdl:
> In this pull request, we enable ruff's `flake8-commas` rule and handle all the reformatting black does because of this.
> 
> ~No functional changes~ Had to update `gen_exports` to run black a 2nd time, if/when this is merged it will probably need to be added to the purely formatting commits thing jakkdl was talking about a while ago.

#2837

CoolCat467:
pre-commit.ci autofix

jakkdl (approved):

A5rocks (approved):
Spot-checks fine. I'm not sure how useful this will be, though forcing `noqa` and `type: ignores` onto lines with only that argument is actually useful.

jakkdl:
> Spot-checks fine. I'm not sure how useful this will be, though forcing `noqa` and `type: ignores` onto lines with only that argument is actually useful.
> 
> Though, can you put this in a blame-ignore file?

it's also generally useful for minimizing diffs when only changing a single argument.

And yes, this should be in blame-ignore. If you wanna be strictly correct you can reshuffle the commits so there's one that is pure refactoring, and one that contains the functional changes to `gen_exports`. This can for example be done with something like (untested, and might depend on the state of your local repo)
```
git reset origin/main
git stash push --patch src/trio/_tools/gen_exports.py
[interactively only stash the functional changes]
git commit -am [...]
git stash pop
git commit -am [...]
```

after some brief testing it appears it's possible to add the hash to `.git-blame-ignore-revs` and not have it change if the PR is merged with a merge commit, in which case you can also do that in the second commit (or in a third).

If you mess up, there's `git reflog`, or pulling from the remote branch. If you don't manage to work it out lmk and I can do it :)

A5rocks (approved):

jakkdl:
2024-08-31T13:03:13,100427169+02:00
https://github.com/python-trio/trio/blame/main/docs/source/conf.py
whoop whoop!

PR #3047, opened 2024-08-01
Enable ruff's `flake8-return` rule

CoolCat467:
In this pull request, we enable ruff's `flake8-return` rule and fix all the new issues from doing this.

### DIFF
--- a/notes-to-self/how-does-windows-so-reuseaddr-work.py
+++ b/notes-to-self/how-does-windows-so-reuseaddr-work.py
@@ -38,11 +38,10 @@ def table_entry(mode1, bind_type1, mode2, bind_type2):
         except OSError as exc:
             if exc.winerror == errno.WSAEADDRINUSE:
                 return "INUSE"
-            elif exc.winerror == errno.WSAEACCES:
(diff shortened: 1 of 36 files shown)


TeamSpen210:
Not sure if this is worth it, it just causes a lot of churn indenting/dedenting things for no real reason. The `elif` technically is redundant if the previous `if` always returns, but it makes it more clear that they're mutually exclusive, and might cause bugs in future if someone removes one of the returns. This also causes symmetric `if-else` checks to become non-symmetric now, which might be confusing to the reader.

A5rocks:
I agree with @TeamSpen210 that this doesn't really help anything. It doesn't make things clearer (unlike `flake8-commas`) and removes semantics. Looking through the error codes listed on <https://pypi.org/project/flake8-return/> (I'm sure `ruff` implements them smarter, ofc):
 - R501 conflicts with mypy
 - R502/R503 is just type checking
 - R504 is potentially useful, but in general assigning to a variable is useful because it allows us to show a nicer name in tracebacks. (maybe we could *just* enable this? I'm not too sure)
 - R505/R506/R507/R508 remove semantics

PR #3049, opened 2024-08-02
Use 3.12 for formatting check

A5rocks:
Fixes https://github.com/python-trio/trio/issues/2681. Not using the lowest version means:
 - this runs faster because Python is faster now
 - it's clear that this doesn't need to be the lowest supported version.

This works now because of the switch to `uv` which has a `--python-version` flag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
         no_test_requirements: ['0']
         extra_name: ['']
         include:
-          - python: '3.8'
+          - python: '3.12'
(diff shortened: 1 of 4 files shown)


CoolCat467 (approved):
Looks good provided that the CI succeeds

CoolCat467:
Shouldn't we also update `no test-requirements` while we are at it?

A5rocks:
> Shouldn't we also update `no test-requirements` while we are at it?

I think its python version isn't related to the formatting check.

Issue #2835, opened 2023-10-25
Replace Autodeps with pre-commit?


richardsheridan:
https://github.com/jazzband/pip-tools#version-control-integration

I don't know much about pre-commit but if we could get it to run on only a subset of commits it would replace the half-baked tool we have. Maybe using the "manual" stage? ([`stages: [manual]`](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages))

jakkdl:
It's in the plans to move away from the current dependency management, see #2790, which is partially blocked by #2798, which is waiting until we've done a release with the new typehints.

A5rocks:
Just popping by to mention that #2790 probably will end up not making a change to how we have requirements files (at least if I have my way and there isn't a great argument for it -- I really don't like having to install a 3rd party tool globally to install requirements ... or have to make 2 venvs, one for the tool and one for the project + deps).

Given that, I think making this change makes sense and shouldn't be blocked by anything! But I still don't know if we can have platform-independent lockfiles using pip-tools, which is my biggest gripe with it.

jakkdl:
(responded to A5rocks in #2790)

A5rocks:
I don't think using a manual stage would make sense, because either way we'd have to commit and push the changes which are the main issues with the current workflow.

CoolCat467:
Manual stage would work well for re-generating the autogenerated files, I might make a pull request implementing that later.

PR #3050, opened 2024-08-03
Regenerate generated files automatically if any are changed

CoolCat467:
Making true on https://github.com/python-trio/trio/issues/2835#issuecomment-2266175000, in this pull request I add a local pre-commit hook that automatically regenerates all the generated files if either the generated file itself or the files used to create them are edited. The only one it doesn't match at the moment is `gen_exports.py` itself, and I could add this if needed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@ ci:
   autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
   autoupdate_schedule: weekly
   submodules: false
+  skip: [regenerate-files]
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,3 +33,11 @@ repos:
     rev: v2.3.0
     hooks:
       - id: codespell
+  - repo: local
+    hooks:
+      - id: regenerate-files
+        name: regenerate generated files
+        language: system
+        entry: python src/trio/_tools/gen_exports.py
+        pass_filenames: false
+        files: ^src\/trio\/_core\/(_generated)?(_run|(_i(o_(common|epoll|kqueue|windows)|nstrumentation)))\.py$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,4 +41,4 @@ repos:
         language: system
         entry: python src/trio/_tools/gen_exports.py
         pass_filenames: false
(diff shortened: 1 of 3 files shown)


A5rocks:
Tbh you can probably add uv pre-commit to keep those up to date, just not have it with updating flags

.pre-commit-config.yaml
```diff
@@ -33,3 +34,11 @@ repos:
...
     hooks:
       - id: codespell
+  - repo: local
+    hooks:
+      - id: regenerate-files
+        name: regenerate generated files
+        language: system
+        entry: python src/trio/_tools/gen_exports.py
+        pass_filenames: false
+        files: ^src\/trio\/_core\/(_generated)?(_run|(_i(o_(common|epoll|kqueue|windows)|nstrumentation)))\.py$
```
A5rocks, line 44: I feel like "any file with generated in it" is probably better and not overfitted to the current things trio supports.

A5rocks (changes requested):

CoolCat467:
> Tbh you can probably add uv pre-commit to keep those up to date too, just not have it with updating flags

I'm afraid I don't quite understand what you mean by this. As far as I am aware, regenerating the generated code files doesn't have anything to do with uv, right?

.pre-commit-config.yaml
CoolCat467: I could certainly change the regex to only match if a file has `_generated` in it, but I am trying to have it match both generated files and the files that are used to create the generated files, mostly because I ran into a case working on https://github.com/python-trio/trio/pull/3044 where the new ruff rules changed the source files and I didn't realize and didn't regenerate the generated files.

A5rocks:
> > Tbh you can probably add uv pre-commit to keep those up to date too, just not have it with updating flags
> 
> 
> 
> I'm afraid I don't quite understand what you mean by this. As far as I am aware, regenerating the generated code files doesn't have anything to do with uv, right?

Sorry, was referring to the generated requirements files. I guess that could be another PR.

.pre-commit-config.yaml
A5rocks: Oh duh, that makes sense. I was thinking about this the wrong way around: you want to check for the inputs, not the outputs. Can we remove the `(_generated)?` bit entirely?

CoolCat467: Sure thing

CoolCat467: Changed with 78f93bdf

A5rocks (approved):

PR #3051, opened 2024-08-03
move_on_ and fail_ functions accepts shield kwarg

### LINKED ISSUE #3052: move_on_ and fail_ functions should accept shield as a keyword argument
Support a simpler syntax for:
```
async with trio.move_on_after(1) as scope:
  scope.shield = True
```
as a one-liner:
```
async with trio.move_on_after(1, shield=True) as scope:
  ...
```

agnesnatasya:
make `move_on_at`, `move_on_after`, `fail_at` and `fail_after` accepts shield as a kwarg.

Closes #3052


src/trio/_tests/test_timeouts.py
```diff
@@ -75,6 +75,66 @@ async def sleep_3() -> None:
     await check_takes_about(sleep_3, TARGET)
 
 
+@slow
+async def test_move_on_after_shields_from_outer() -> None:
+    duration = 0.1
```
A5rocks, line 80: Given that this is using sleep a checkpoint, can you use a checkpoint (or trio.sleep(0)) instead of a hardcoded small value?

A5rocks (commented):
Just a comment about the tests, I don't have much of an opinion on the feature itself.

src/trio/_tests/test_timeouts.py
agnesnatasya: If i use 0, i don't think i can make use of `check_takes_about` because it checks `dur / expected_dur < 1.5`, and `expected_dur = 0` in this case. The docstring also says that they observe overruns when they sleep for 0.05s, so I picked 0.1s here.
Alternatively, I can use a different way to test whether we actually sleep (instead of using `check_takes_about`)

newsfragments/3049.feature.rst
```diff
@@ -0,0 +1 @@
+`trio.move_on_at`, `trio.move_on_after`, `trio.fail_at` and `trio.fail_after` now accepts shield as a kwarg.
```
oremanj, line 1:
```suggestion
`trio.move_on_at`, `trio.move_on_after`, `trio.fail_at` and `trio.fail_after` now accept *shield* as a keyword argument. If specified, it provides an initial value for the `~trio.CancelScope.shield` attribute of the `trio.CancelScope` object created by the context manager.
```

src/trio/_tests/test_timeouts.py
oremanj: So don't use `check_takes_about` (and don't mark this test as slow). Instead do something like:
```
try:
except trio.Cancelled:
    pytest.fail("shield didn't work")
inner.shield = False
with pytest.raises(trio.Cancelled):
```

src/trio/_tests/test_timeouts.py
```diff
@@ -75,6 +75,66 @@ async def sleep_3() -> None:
...
+            outer.cancel()
+            # The outer scope is cancelled, but this task is protected by the
+            # shield, so it manages to get to sleep until deadline is met
+            await sleep_forever()
+
+    await check_takes_about(task, TARGET)
+
+
+@slow
+async def test_fail_after_exits_shields_from_outer() -> None:
```
oremanj, line 108: This appears to be exactly the same test as `test_move_on_after_shields_from_outer` except with substituting `fail_after` for `move_on_after`. It could be one test that uses `@pytest.mark.parametrize` to provide the identity of the context manager to use.

src/trio/_timeouts.py
```diff
@@ -7,36 +7,38 @@
...
 
 
-def move_on_at(deadline: float) -> trio.CancelScope:
+def move_on_at(deadline: float, *, shield: bool = False) -> trio.CancelScope:
     """Use as a context manager to create a cancel scope with the given
     absolute deadline.
 
     Args:
       deadline (float): The deadline.
+      shield (bool): Whether the cancel scope created is shielded.
```
oremanj, line 16:
```suggestion
      shield (bool): Initial value for the `~trio.CancelScope.shield` attribute
          of the newly created cancel scope.
```
Suggest you link to the documentation for `CancelScope.shield` to clarify what shielding is about (in all four of these).

CoolCat467:
I think these changes are good, but I am curious about the root issue behind this. Are there use cases where it's not sufficient to set the shield to True slightly after it's initialized? And even in that case, couldn't you just create the cancel scope, set the `shield` attribute, and *then* enter it?

agnesnatasya:
> Are there use cases where it's not sufficient to set the shield to True slightly after it's initialized? And even in that case, couldn't you just create the cancel scope, set the shield attribute, and then enter it?

This is sufficient, and this change is just to have a one-liner syntax for this behavior.

src/trio/_timeouts.py
```diff
@@ -7,36 +7,42 @@
...
-def move_on_at(deadline: float) -> trio.CancelScope:
+def move_on_at(deadline: float, *, shield: bool = False) -> trio.CancelScope:
     """Use as a context manager to create a cancel scope with the given
     absolute deadline.
 
     Args:
       deadline (float): The deadline.
+      shield (bool): Initial value for the `~trio.CancelScope.shield` attribute
+          of the newly created `cancel scope
+          <https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope.shield>`__.
```
CoolCat467, line 18: Shouldn't sphinx be able to detect `CancelScope` reference if you changed text to `~trio.CancelScope` or something? I'm pretty sure there's a way to do that, but I don't remember.

src/trio/_timeouts.py
```diff
@@ -7,36 +7,42 @@
...
-def move_on_after(seconds: float) -> trio.CancelScope:
+def move_on_after(seconds: float, *, shield: bool = False) -> trio.CancelScope:
     """Use as a context manager to create a cancel scope whose deadline is
     set to now + *seconds*.
 
     Args:
       seconds (float): The timeout.
+      shield (bool): Initial value for the `~trio.CancelScope.shield` attribute
+          of the newly created `cancel scope
+          <https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope.shield>`__.
```
CoolCat467, line 37: Same as prior comment

src/trio/_timeouts.py
```diff
@@ -110,14 +116,17 @@ def fail_at(deadline: float) -> AbstractContextManager[trio.CancelScope]:  # typ
 
     Args:
       deadline (float): The deadline.
+      shield (bool): Initial value for the `~trio.CancelScope.shield` attribute
+          of the newly created `cancel scope
+          <https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope.shield>`__.
```
CoolCat467, line 121: Same as prior comment

src/trio/_timeouts.py
```diff
@@ -140,6 +151,9 @@ def fail_after(seconds: float) -> AbstractContextManager[trio.CancelScope]:
 
     Args:
       seconds (float): The timeout.
+      shield (bool): Initial value for the `~trio.CancelScope.shield` attribute
+          of the newly created `cancel scope
+          <https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope.shield>`__.
```
CoolCat467, line 156: Same as prior comment

CoolCat467 (approved):
Looks good!

A5rocks:
I was looking at whether this has been discussed before and found https://github.com/python-trio/trio/issues/147 where @njsmith thinks that making `shield=True` too ergonomic is an issue because people shouldn't shield much. On the other hand, `flake8-async` suggests people shield async statements in `finally`, libraries might have to use shields to make things safe for cancellation, and that issue seems to have gotten nowhere.

On the other other hand, the "library" I'm looking at (hypercorn's TCP send implementation) manually sets `scope.shield = True` even though `shield=True` is a kwarg for `CancelScope()`. @oremanj since you talked in that issue and are here already, what do you think?

(really this comment belongs in the related issue but... already written, what can you do.)

oremanj:
Sorry for the delay, have been ill this week.

@A5rocks I would say that if we're trying to encourage use of shield+timeout over just shield-without-timeout -- which I think we are -- then we shouldn't make shield+timeout be cumbersome to write. Currently one must do either
```
with trio.CancelScope(shield=True) as scope:
    scope.deadline = trio.current_time() + 5
    <stuff>
```
or
```
with trio.move_on_after(5) as scope:
    scope.shield = True
    <stuff>
```
or
```
with trio.CancelScope(shield=True, deadline=trio.current_time() + 5):
    <stuff>
```
none of which exactly roll off the fingers. I think we should allow
```
with trio.move_on_after(5, shield=True):
    <stuff>
```
so people aren't pushed toward the more uncancellable-hang-prone
```
with trio.CancelScope(shield=True):
    <stuff>
```
If we don't want to make shielding visible in `move_on_after` and friends, then I think we should support `trio.CancelScope(timeout=5, shield=True)`. But that then has an ambiguity associated with it (does the timeout start when the cancel scope is constructed or does it start when you enter the `with` block?) which seems like more interface-design trouble than it's worth.

Overall I think shielding is more important in moderately-complex-and-up programs than Trio's initial design gave it credit for.

A5rocks (approved):
Sounds good then!

PR #3054, opened 2024-08-03
Don't use `hash` to avoid attrs warning

### LINKED ISSUE #3053: Using trio with attrs 24.1.0 raises a warning
This happens to break urllib3's CI which, like trio, fails on warnings.

```
ImportError while loading conftest '/.../urllib3/test/conftest.py'.
test/conftest.py:15: in <module>
    from dummyserver.app import hypercorn_app
dummyserver/app.py:14: in <module>
    import trio
.nox/test-3-12/lib/python3.12/site-packages/trio/__init__.py:23: in <module>
    from ._core import TASK_STATUS_IGNORED as TASK_STATUS_IGNORED  # isort: split
.nox/test-3-12/lib/python3.12/site-packages/trio/_core/__init__.py:9: in <module>
    from ._entry_queue import TrioToken
.nox/test-3-12/lib/python3.12/site-packages/trio/_core/_entry_queue.py:149: in <module>
    @attrs.define(eq=False, hash=False)
.nox/test-3-12/lib/python3.12/site-packages/attr/_next_gen.py:402: in wrap
    return do_it(cls, True)
.nox/test-3-12/lib/python3.12/site-packages/attr/_next_gen.py:348: in do_it
    return attrs(
.nox/test-3-12/lib/python3.12/site-packages/attr/_make.py:1291: in attrs
    warnings.warn(
E   DeprecationWarning: The `hash` argument is deprecated in favor of `unsafe_hash` and will be removed in or after August 2025.
```

CoolCat467:
Fixes #3053, this pull request replaces all uses of `@attrs.define(... hash=x)` with `@attrs.define(... unsafe_hash=x)`.
In attrs >= 24.1.0, a warning is raised that `hash` is going to be removed in August of 2025.
`unsafe_hash` is an alias for `hash` as of attrs v22.2.0, so there should be no functional changes.


richardsheridan:
I'm sure none of our code actually needs this argument, we should be using _just_ eq=false or frozen=true to get the behavior we need. These were all put in before attrs was giving clear guidance on the feature.

Surely the tests all pass if you just remove the hash=false text?

CoolCat467:
> Surely the tests all pass if you just remove the hash=false text?

No idea, haven't tested that, but I was trying to avoid changing functionality.

A5rocks:
Yeah given attrs says:
> The overarching theme is to never set the `@attrs.define(unsafe_hash=X)` parameter yourself. Leave it at `None` which means that attrs will do the right thing for you, depending on the other parameters:

it's probably a good idea to not use the parameter.

A5rocks (changes requested):
In addition to removing all `unsafe_hash=False`s, can you add a newsfragment? You added a tag not to, but I think it's a good thing to talk about for people just reading the changelogs.

EDIT: also, could you bump attrs in the test requirements so we can be sure there's no other warnings?

A5rocks (approved):

CoolCat467 (commented):
Looks good @A5rocks

A5rocks:
Let's let others check this, I'll only make a release tomorrow anyways.

CoolCat467:
Yea mostly commenting on the additional changes you made

agronholm (approved):
LGTM.

agronholm:
I need this for AnyIO to pass its test suite on 3.13 :smiley:

hynek:
Whoa, y'all are fast! I ran into the warning in stamina and wanted to reach out to you how obnoxious it is for you and if I should just soft-deprecate it instead.

Looks like you got most of them except (at least) https://github.com/python-trio/trio/blob/223e4f44431630b27f598c2cc689469701033c56/src/trio/_threads.py#L142

A fun fact on the side: *attrs*'s guidance on hashing goes entirely back to @njsmith opening my eyes and causing a big rewrite of that feature within *attrs*. So the chances are excellent that you don't need the argument at all.

Tangentially, `unsafe_hash` has been added to attrs already in 22.2.0, so if you want, you can relax that lower pin a bit.

A5rocks:
> I ran into the warning in stamina and wanted to reach out to you how obnoxious it is for you and if I should just soft-deprecate it instead.

We've already handled it and it's been fine: it revealed two configuration issues so that's nice! (one is that deprecation warnings get ignored if the module is imported as a plugin first. the other is that one of the CI runs should use `test-requirements.txt` as a constraints file)

CoolCat467:
> Looks like you got most of them except (at least)
> 
> https://github.com/python-trio/trio/blob/223e4f44431630b27f598c2cc689469701033c56/src/trio/_threads.py#L142

Unfortunate, surprised CI didn't catch that, but this is because when I made this pull request I did `git grep attrs.define | grep hash` if I remember correctly, so will need to do another pass looking for anything not using `attrs.define`

hynek:
I did grepped later for hash= and found only this one instance so I think you’ll be fine!

Issue #3048, opened 2024-08-02
Add a Github Action that checks for newsfragment


A5rocks:
I've been forgetting these quite a bit for reviews. While many changes don't need them, it's much easier to explicitly mark what PRs shouldn't (with a tag) than the other way around.

Of course, we can do some basic filtering so that only PRs that change things in `src/trio/` (ignoring paths with `/_tests/` in them) need them. This doesn't have to be perfect.

PR #3055, opened 2024-08-04
Check for newsfragments

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3048

This is something way too easy to forget: hence, automation. I don't think the screening done here is great, but hopefully it's good enough for now?

### DIFF
--- a/.github/workflows/check-newsfragment.yml
+++ b/.github/workflows/check-newsfragment.yml
@@ -1,0 +1,23 @@
+name: Check newsfragment
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


CoolCat467:
Why is on pull request trigger checking `type: closed`?

A5rocks:
Copy-paste from the default types + labelled, but good point.

CoolCat467 (approved):
Looks good, CI is currently failing because of #3053 but once #3054 is merged that should fix it.

PR #3058, opened 2024-08-05
Fix attrs redone

A5rocks:
https://github.com/python-trio/trio/pull/3054 didn't catch everything. However the root cause was that pytest imports trio during plugin phase and doesn't handle deprecation warnings by erroring then.

To fix that, we can just clear the module cache and try importing again. I'm not completely sold on this idea but it's the best one I had and it works.

### DIFF
--- a/src/trio/_tests/test_trio.py
+++ b/src/trio/_tests/test_trio.py
@@ -1,5 +1,8 @@
-import sys
+def test_trio_import():
+    import sys
 
-for module in list(sys.modules.keys()):
-    if module.startswith("trio"):
-        del sys.modules[module]
+    for module in list(sys.modules.keys()):
+        if module.startswith("trio"):
+            del sys.modules[module]
+    
+    import trio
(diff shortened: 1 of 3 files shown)


CoolCat467 (approved):
Looks good, but to be honest kind of confused about the reasoning for `test_endtoend` changes

A5rocks:
The mypy annotate tool uses pickle  so you don't want to use a cached version of trio otherwise it fails.

(It gets the cached version of trio even though that was deleted from sys.modules because it's cached by the tool)

CoolCat467:
Makes sense, thanks for explaining. In that case though, why is there another call to cached `main` just a few lines down then?

https://github.com/python-trio/trio/blob/72c1c684dfd7473cc936b3db9a6d8dc3a9b14ee3/src/trio/_tests/tools/test_mypy_annotate.py#L131

A5rocks:
Oh I missed that! I'm surprised that doesn't error.

I assume it's cause it loads the object from pickle which uses the version of trio I just used rather than the one cached in the `main`. That's just a guess though.

A5rocks:
Sorry about the review request, I tried something that seemed to work locally but it didn't work in CI (after I requested your review again).

PR #3069, opened 2024-08-21
Update `uv`'s output when it gets bumped in autodeps

A5rocks:
This avoids having to manually update the lockfile and allows for easier regression testing (rather than having to trawl through a failing check.sh run).


CoolCat467 (approved):
Looks good

PR #3074, opened 2024-08-29
Test PyPy 3.10 on both Windows and MacOS

A5rocks:

This seems to just work as is.


CoolCat467 (approved):
Looks good

PR #3075, opened 2024-08-29
Enable mypy and disable jedi on PyPy

A5rocks:

I remember jedi being somewhat flaky on PyPy so this should fix that.


CoolCat467 (approved):

PR #3076, opened 2024-08-31
Update open process types

A5rocks:
This should resolve the issue mentioned in Gitter

### DIFF
--- a/src/trio/_subprocess.py
+++ b/src/trio/_subprocess.py
@@ -333,9 +333,10 @@ async def _open_process(
           bytes such as ``['ls', '-l', 'directory with spaces']``, where the
           first element names the executable to invoke and the other elements
           specify its arguments. With ``shell=True`` in the ``**options``, or on
-          Windows, ``command`` may alternatively be a string or bytes, which
-          will be parsed following platform-dependent :ref:`quoting rules
-          <subprocess-quoting>`.
+          Windows, ``command`` can be a string or bytes, which will be parsed
+          following platform-dependent :ref:`quoting rules
+          <subprocess-quoting>`. In all cases ``command`` can be a path or a
+          sequence of paths.
       stdin: Specifies what the child process's standard input
           stream should connect to: output written by the parent
           (``subprocess.PIPE``), nothing (``subprocess.DEVNULL``),
@@ -369,6 +370,7 @@ async def _open_process(
             )
 
     if os.name == "posix":
+        # TODO: how do paths and sequences thereof play with `shell=True`?
         if isinstance(command, (str, bytes)) and not options.get("shell"):
             raise TypeError(
                 "command must be a sequence (not a string or bytes) if "
--- a/src/trio/_subprocess.py
+++ b/src/trio/_subprocess.py
@@ -329,9 +329,9 @@ async def _open_process(
     want.
 
     Args:
-      command (str or byte or sequence of either): The command to run. Typically
-          this is a sequence of strings or bytes such as ``['ls', '-l',
-          'directory with spaces']``, where the first element names the
+      command (str or bytes or Sequence[str] or Sequence[bytes]): The command to
+          run. Typically this is a sequence of strings or bytes such as ``['ls',
+          '-l', 'directory with spaces']``, where the first element names the
           executable to invoke and the other elements specify its arguments.
           With ``shell=True`` in the ``**options``, or on Windows, ``command``
           may alternatively be a string or bytes, which will be parsed following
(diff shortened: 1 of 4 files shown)


agronholm:
The updated documentation makes no mention of `PathLike` – shouldn't that be fixed?

src/trio/_subprocess.py
```diff
@@ -329,13 +329,13 @@ async def _open_process(
...
 
     Args:
-      command (list or str): The command to run. Typically this is a
-          sequence of strings such as ``['ls', '-l', 'directory with spaces']``,
-          where the first element names the executable to invoke and the other
-          elements specify its arguments. With ``shell=True`` in the
-          ``**options``, or on Windows, ``command`` may alternatively
-          be a string, which will be parsed following platform-dependent
-          :ref:`quoting rules <subprocess-quoting>`.
+      command (str or bytes or Sequence[str] or Sequence[bytes]): The command to
```
jakkdl, line 332: what's the upside of documenting the type in the docstring at all? It feels redundant

agronholm: I don't know, but if they are documented, they should be documented accurately.

jakkdl: oh for sure, but maybe better to remove them (unless we wanna leave that for a different issue+pr to do that systematically). The other parameters aren't specified in this docstring even.

A5rocks:
I figured it's fine to have more accurate documentation (type hint) and less accurate but clearer (what's written in the docstring). Both are visible in the built docs.

I can make them the same though.

src/trio/_subprocess.py
agronholm: Do what you feel is best, so long as consistency is maintained.

A5rocks:
Huh https://github.com/python-trio/trio/actions/runs/10649671109/job/29520147803#step:4:208 shouldn't be happening. I'll assume it's transient but it seems like some sort of `pip` bug probably?

TeamSpen210:
The cache has been working a few days ago, [this run](https://github.com/python-trio/trio/actions/runs/10628314082/job/29462930592) for example. We'd definitely want it if possible. Could the problem be that the Cython build isn't using `cache-dependency-path`?

A5rocks:
> The cache has been working a few days ago, [this run](https://github.com/python-trio/trio/actions/runs/10628314082/job/29462930592) for example. We'd definitely want it if possible. Could the problem be that the Cython build isn't using `cache-dependency-path`?

It wasn't Cython, it was PyPy on Windows (specifically). It should be using the `cache-dependency-path` right because it has the same hash (`c5182900c6da08ceac54233234be0232412d20dcd40abe0f886b892a46fd3c5e`) within the commit for the run I linked. Maybe it's just using the hash from the current branch rather than what `@actions/checkout` pulls (which is the state were this PR merged)?

TeamSpen210:
It's very mystifying. Looking at the code, the hash should be stored [here](https://github.com/actions/setup-python/blob/f677139bbe7f9c59b41e40162b753c062f5d49a3/src/cache-distributions/cache-distributor.ts#L53), and then reloaded [here](https://github.com/actions/setup-python/blob/f677139bbe7f9c59b41e40162b753c062f5d49a3/src/cache-save.ts#L48), so it shouldn't be affected if the requirements file was say edited. But I was thinking maybe there could be an issue if we had another action using `setup-python` internally. Not sure what though.

jakkdl (approved):

src/trio/_subprocess.py
```diff
@@ -304,7 +304,7 @@ def kill(self) -> None:
 
 
 async def _open_process(
-    command: list[str] | str,
+    command: StrOrBytesPath | Sequence[StrOrBytesPath],
```
CoolCat467, line 307: Wondering, why is type var here called `StrOrBytesPath`? Wouldn't it be `StrOrBytes`, or is `StrOrBytesPath` special somehow?

CoolCat467 (approved):
Looks great, I just have one question in a comment I made.

src/trio/_subprocess.py
agronholm: Because it's typed as `str | bytes | PathLike[str] | PathLike[bytes]`?

CoolCat467: Reading the docstring currently in place, I wasn't under the impression it supported pathlikes. If you could update that then, that would be great!

A5rocks: The docstring atm distinguishes what is possible with `shell=True` and what isn't, but I think paths are fine either way. I can update the docstring though. (this also appears in the function signature Sphinx makes)

agronholm: I would seem to be for the best to mention that, yes. After that, I have no objections.

Issue #3035, opened 2024-07-16
Problems with `trio.Lock()` across multiple tasks


Zac-HD:
Some time ago - never mind how long precisely - having a little coordination between tasks and nothing particular to interest me in semaphores, I saw our `RuntimeError("can't release a Lock you don't own")`.

I approve pretty strongly of keeping things local by default; it prevents all sorts of bugs and `trio.Semaphore` is right there if you need something without this limitation.  But two further observations prompted this issue:

1. `trio.Semaphore()` is considerably more powerful and flexible in general.  By the principle of least power, it might be nice to have `trio.Lock(*, allow_outside_release: bool = False)` or something along those lines.
2. If the owning task finishes (returns or raises) _without releasing the lock_, then you're stuck forever.  The toy program below demonstrates.  Perhaps we should raise an error or at least emit a warning when a task which owns a lock finishes?

```python
import trio

@trio.run
async def main():
    lock = trio.Lock()
    async with trio.open_nursery() as nursery:
        nursery.start_soon(lock.acquire)
    # The lock is now held, and can only be released by the acquiring task.
    # However, that task is now finished, so we can never release the lock!
    lock.release()
```

A5rocks:
I'm not sure how I feel about adding a keyword argument to `trio.Lock`. Does `trio.Semaphore` with a limit of `1` really have differences other than this? I like that `trio.Lock` is more easily debuggable and harder to shoot yourself with. (and if we make `acquire` error given the owner is dead, then even better!)

>  Perhaps we should raise an error or at least emit a warning when a task which owns a lock finishes?

I think that this could error when you run `lock.acquire_nowait()`. i.e. `lock.acquire_nowait` could check if `self._owner` is still running. This doesn't handle the following though:
- task 1 calls `lock.acquire`
- task 2 calls `lock.acquire`
- task 1 exits

Zac-HD:
Let's set `Semaphore` aside then, I'm more concerned that you can easily get a deadlock from `Lock`, e.g.

```python
    lock = trio.Lock()
    async with trio.open_nursery() as nursery:
        nursery.start_soon(lock.acquire)
        nursery.start_soon(lock.acquire)
```

IMO instead of hanging forever, we should raise `BrokenResourceError` in the task(s) waiting on a lock which can never be released, and perhaps also when exiting the task which holds that lock.

A5rocks:
[...]

PR #3080, opened 2024-09-02
raise BrokenResourceError if lock owner has exited.

jakkdl:
Partially addresses #3035 

* Idk if adding `test_lock_multiple_acquire` is very meaningful
* need to add newsfragment

It's possible that a full implementation for #3035 makes this redundant, but this was easy enough to write&test I thought I might as well push it as a PR that improves the status quo for now.

### DIFF
--- a/src/trio/_sync.py
+++ b/src/trio/_sync.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import math
 from typing import TYPE_CHECKING, Protocol
 
@@ -571,11 +572,19 @@ class _LockImpl(AsyncContextManagerMixin):
         """
 
         task = trio.lowlevel.current_task()
+
         if self._owner is task:
             raise RuntimeError("attempt to re-acquire an already held Lock")
         elif self._owner is None and not self._lot:
             # No-one owns it
             self._owner = task
+        elif (
+            self._owner is not None
+            and inspect.getcoroutinestate(self._owner.coro) is inspect.CORO_CLOSED
+        ):
+            raise trio.BrokenResourceError(
+                "Attempted acquire on Lock which will never be released, owner has exited.",
+            )
         else:
             raise trio.WouldBlock
 
--- a/src/trio/_tests/test_sync.py
+++ b/src/trio/_tests/test_sync.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Callable, Union
 import pytest
 
 from .. import _core
+from .._core._tests.tutil import slow
 from .._sync import *
 from .._timeouts import sleep_forever
 from ..testing import assert_checkpoints, wait_all_tasks_blocked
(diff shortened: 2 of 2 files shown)


src/trio/_tests/test_sync.py
```diff
@@ -586,3 +587,31 @@ async def lock_taker() -> None:
...
+        await lock.acquire()
+
+
+@slow
+async def test_lock_multiple_acquire() -> None:
+    """This currently hangs without throwing any error
+    See https://github.com/python-trio/trio/issues/3035
+    """
+    lock = trio.Lock()
+    with pytest.raises(trio.TooSlowError):
```
A5rocks, line 612: How about xfailing this test?

src/trio/_sync.py
```diff
@@ -571,11 +572,19 @@ def acquire_nowait(self) -> None:
...
             raise RuntimeError("attempt to re-acquire an already held Lock")
         elif self._owner is None and not self._lot:
             # No-one owns it
             self._owner = task
+        elif (
+            self._owner is not None
+            and inspect.getcoroutinestate(self._owner.coro) is inspect.CORO_CLOSED
+        ):
+            raise trio.BrokenResourceError(
+                "Attempted acquire on Lock which will never be released, owner has exited.",
```
A5rocks, line 586:
```suggestion
                "Attempted acquire on Lock which will never be released as the owner has exited.",
```

I'm not sure this is the ideal message either way but the alternative I can think of isn't great either: "Deadlock detected: acquiring lock which will never be released."

A5rocks (changes requested):
Can you add a newsfragment?

A5rocks:
> I'm somewhat concerned about the performance impact of `inspect.getcoroutinestate` in `acquire_nowait()`, which might be fine.

It should only be in the case where we would lock anyways.

> The larger problem is that it only addresses the case where the holding task is _already_ gone at acquire-time; I think we should probably go directly to the general solution of adding a `ParkingLot.mark_resource_broken()` method and using that from Lock.

Yeah, the problem is we can't just add a way to poison a `ParkingLot`, we also need a way to know a `Lock`'s task got ended. We can't use a `weakref.finalizer` because no guarantees on when `Task` gets deleted (it could be stored somewhere else other than `_owner`), so I think we need something in `trio`'s event loop itself.

jakkdl:
> I'm somewhat concerned about the performance impact of `inspect.getcoroutinestate` in `acquire_nowait()`, which might be fine.

I ran a simple benchmark now and could not detect any difference with or without a call to `inspect.getcoroutinestate`.

> The larger problem is that it only addresses the case where the holding task is _already_ gone at acquire-time; I think we should probably go directly to the general solution of adding a `ParkingLot.mark_resource_broken()` method and using that from Lock.

Yes this PR intentionally only made the easy fix from https://github.com/python-trio/trio/issues/3035#issuecomment-2312328838, but if the full fix turns out doable then we'll close this one.

jakkdl:
This will probably end up closed in favor of #3081

A5rocks:
Now unnecessary.

PR #3081, opened 2024-09-06
Add ability to break parking lots, stop locks from stalling

### LINKED ISSUE #3035: Problems with `trio.Lock()` across multiple tasks

jakkdl:
fixes #3035 

TODO:
* [x] newsfragment(s)
* [x] docstrings
* [x] docs
* [x] tests of directly using the parking lot functions
* [x] lock test that checks that handing over the owner works (this is indirectly tested by a lot of other tests, but good to have a standalone test)
* [x] reraise the exception in `Lock`/`_LockImpl` to get a better error message


src/trio/_core/_parking_lot.py
```diff
@@ -86,6 +87,28 @@
...
+        GLOBAL_PARKING_LOT_BREAKER[task].append(lot)
+
+
+def remove_parking_lot_breaker(task: Task, lot: ParkingLot) -> None:
+    """Deregister a task as a breaker for a lot. See :func:`add_parking_lot_breaker`."""
+    if task not in GLOBAL_PARKING_LOT_BREAKER:
+        raise RuntimeError(
+            "Attempted to remove parking lot breaker for task that is not registered as a breaker",
+        )
+    GLOBAL_PARKING_LOT_BREAKER[task].remove(lot)
```
TeamSpen210, line 109: Instead of directly checking each time, this could just catch the `KeyError` (then reraise and suppress), so it has zero overhead in the happy case. Might also want to catch `ValueError` from `remove()` failing.

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +266,23 @@ def repark_all(self, new_lot: ParkingLot) -> None:
...
 
+    def break_lot(self, task: Task) -> None:
+        """Break this lot, causing all parked tasks to raise an error, and any
+        future tasks attempting to park (and unpark? repark?) to error. The error
+        contains a reference to the task sent as a parameter."""
+        self.broken_by = task
+        # TODO: weird to phrase this one, we probably should reraise this error in Lock
+        error = outcome.Error(
+            _core.BrokenResourceError(f"Parking lot broken by {task}"),
+        )
```
TeamSpen210, line 277: I think this shouldn't be cached, each task should get its own outcome & exception object. Otherwise the traceback would get scrambled together.

src/trio/_core/_parking_lot.py
jakkdl: Fixed

and tbh it's probably better not to distinguish between `KeyError` and `ValueError`. Which one you'll get will depend on if the task has previously parked any lots or not, not much about the current state.

src/trio/_core/_parking_lot.py
jakkdl: thanks, fixed

src/trio/_core/_parking_lot.py
```diff
@@ -86,6 +87,31 @@
     from ._run import Task
 
 
+GLOBAL_PARKING_LOT_BREAKER: dict[Task, list[ParkingLot]] = {}
+
+
+def add_parking_lot_breaker(task: Task, lot: ParkingLot) -> None:
+    """Register a task as a breaker for a lot. This means that if the task exits without
+    having unparked from the lot, then the lot will break and raise an error for all tasks
+    parked in the lot, as well as any future task that attempt to park in it."""
```
A5rocks, line 96: Hm, this isn't exactly the behavior we're looking for: it's if it exits without removing itself as a parking lot breaker, rather than if it exits while parked on the lot.

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +269,23 @@ def repark_all(self, new_lot: ParkingLot) -> None:
...
 
+    def break_lot(self, task: Task) -> None:
+        """Break this lot, causing all parked tasks to raise an error, and any
+        future tasks attempting to park (and unpark? repark?) to error. The error
+        contains a reference to the task sent as a parameter."""
+        self.broken_by = task
+
+        # TODO: is there any reason to use self._pop_several?
+        for parked_task in self._parked:
+            # TODO: weird to phrase this one, we maybe should reraise this error in Lock?
```
A5rocks, line 280: I agree that it should be reraised -- I think it's best to use a subclass so that anyone using this can catch and reraise without any concern about catching things incorrectly + still have access to the task that caused the parking lot to break.

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +269,23 @@ def repark_all(self, new_lot: ParkingLot) -> None:
         """
         return self.repark(new_lot, count=len(self))
 
+    def break_lot(self, task: Task) -> None:
```
A5rocks, line 272: I think taking `task` optionally might be a nicer interface (cause we're making this public I guess), defaulting it to `trio.lowlevel.current_task()`.

A5rocks:
@njsmith cause this is interface design, do you think there's a better way to express all of this?

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +269,23 @@ def repark_all(self, new_lot: ParkingLot) -> None:
...
+        """Break this lot, causing all parked tasks to raise an error, and any
+        future tasks attempting to park (and unpark? repark?) to error. The error
+        contains a reference to the task sent as a parameter."""
+        self.broken_by = task
+
+        # TODO: is there any reason to use self._pop_several?
```
A5rocks, line 278: Probably some thin veneer of thread safety? Or guaranteeing that we'll remove every task we raise an error in, even if somehow `_core.reschedule` raises an error.

Neither is really something we care about but probably worth using cause it's a simple drop in replacement anyways.

TeamSpen210: `unpark()` above would have the same theoretical issue, if `reschedule()` failed it could drop tasks too. Probably not something to worry about, if that fails we've got bigger problems.

src/trio/_core/_parking_lot.py
jakkdl: Ah right. The latter is specifically for `Lock`.

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +269,23 @@ def repark_all(self, new_lot: ParkingLot) -> None:
         """
         return self.repark(new_lot, count=len(self))
 
+    def break_lot(self, task: Task) -> None:
+        """Break this lot, causing all parked tasks to raise an error, and any
+        future tasks attempting to park (and unpark? repark?) to error. The error
```
jakkdl, line 274: Should `ParkingLot.unpark`/`repark` raise `BrokenResourceError` if the lot is broken? Breaking the lot empties it, so `unpark`/`repark` becomes no-ops otherwise.

jakkdl:
> @njsmith cause this is interface design, do you think there's a better way to express all of this?

as we don't actually have anybody requesting the functionality on `ParkingLot` or `add_parking_lot_breaker`/`remove_parking_lot_breaker`, we could opt not to make these public for now and document them as an experimental API. Or I suppose make them public but with a warning text in the docstring that they're experimental.

newsfragments/3081.feature.rst
```diff
@@ -0,0 +1 @@
+Added :func:`trio.lowlevel.add_parking_lot_breaker` and :func:`trio.lowlevel.remove_parking_lot_breaker` to allow creating custom lock/semaphore implementations that will break their underlying parking lot if a task exits unexpectedly. :meth:`trio.lowlevel.ParkingLot.break_lot` is also added, to allow breaking a parking lot intentionally. Breaking a parking lot raises :exc:`trio.BrokenResourceError` for all tasks currently parked in the lot, and any tasks attempting to park in an already broken parking lot will also error. The breakage status of a lot can be viewed and manually modified with the ``trio.ParkingLot.broken_by`` attribute.
```
jakkdl, line 1: idk why sphinx isn't finding any of the things I added

src/trio/_core/_parking_lot.py
A5rocks: I don't think so...

It's hard to think of a concrete case that this matters for; locks will not be able to reach them. However, it's possible to come up with cases I suppose?:
 - case for raising errors: data structure that needs to have at least one task parked at all times. Obviously having none would be a resource error.
 - case against: people putting parks in a try, then unparking the next task in a finally. Throwing would just make the exception worse.

I think the second case convinces me that it shouldn't raise but idk.

src/trio/_sync.py
```diff
@@ -18,6 +19,11 @@
     from ._core._parking_lot import ParkingLotStatistics
 
 
+class StalledLockError(Exception):
+    """Raised by :meth:`Lock.acquire` and :meth:`StrictFIFOLock.acquire` if the owner
+    exits, or has previously exited, without releasing the lock."""
```
Zac-HD, line 24: I'm not convinced we need a new exception type here - wouldn't raising `BrokenResourceError` make sense?  If we do go with a new type, I'd at least inherit from that.

src/trio/_core/_parking_lot.py
Zac-HD: yeah, I think treating these as noops makes sense, even though my initial inclination was to error out.

newsfragments/3081.feature.rst
Zac-HD: I think they have to be explicitly documented somewhere in a .rst file, since Trio doesn't use `.. automodule::`.

src/trio/_core/_tests/test_parking_lot.py
```diff
@@ -215,3 +218,73 @@ async def test_parking_lot_repark_with_count() -> None:
...
+        async with _core.open_nursery() as nursery:
+            nursery.start_soon(bad_parker, lot, cs)
+            await wait_all_tasks_blocked()
+
+            nursery.start_soon(lot.park)
+            await wait_all_tasks_blocked()
+
+            cs.cancel()
+
+    # check that trying to park in brokena lot errors
```
Zac-HD, line 269:
```suggestion
    # check that trying to park in broken lot errors
```

src/trio/_core/_tests/test_parking_lot.py
```diff
@@ -215,3 +218,73 @@ async def test_parking_lot_repark_with_count() -> None:
...
+
+    with pytest.raises(
+        RuntimeError,
+        match="Attempted to remove task as breaker for a lot it is not registered for",
+    ):
+        trio.lowlevel.remove_parking_lot_breaker(task, lot)
+    trio.lowlevel.add_parking_lot_breaker(task, lot)
+    trio.lowlevel.add_parking_lot_breaker(task, lot)
+    trio.lowlevel.remove_parking_lot_breaker(task, lot)
+    trio.lowlevel.remove_parking_lot_breaker(task, lot)
```
Zac-HD, line 235: add a code comment to explain these duplicated lines?

Zac-HD (commented):
This looks good to me!  I'd appreciate a review by another experienced maintainer but I'd be inclined to merge once minor comments are addressed 😁

src/trio/_sync.py
A5rocks: Yeah this might have been because of my comment. Sorry for not being more explicit: IMO parking lots should raise a subclass of `BrokenResourceError`. This would mean that locks can do something like this (for better error messages):

```py
try:
  await parking_lot.park()
except BrokenParkingLot as e:
  raise BrokenResourceError(f"this lock was held by {e.task} when it exited")  # note specifically that task is accessible by `e.task`
```

src/trio/_core/_tests/test_parking_lot.py
A5rocks:
```suggestion
    # check that breaking a parking lot is re-entrant
    trio.lowlevel.add_parking_lot_breaker(task, lot)
    trio.lowlevel.add_parking_lot_breaker(task, lot)
    trio.lowlevel.remove_parking_lot_breaker(task, lot)
    trio.lowlevel.remove_parking_lot_breaker(task, lot)
```
(that's actually a bad explanation maybe, but it's not like test comments matter that much)

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +269,24 @@ def repark_all(self, new_lot: ParkingLot) -> None:
...
         return self.repark(new_lot, count=len(self))
 
+    def break_lot(self, task: Task | None = None) -> None:
+        """Break this lot, causing all parked tasks to raise an error, and any
+        future tasks attempting to park (and unpark? repark?) to error. The error
+        contains a reference to the task sent as a parameter.
+        """
+        if task is None:
+            task = _core.current_task()
+        self.broken_by = task
```
A5rocks, line 279: I'm just now realizing that this relationship is a bit flawed. Lots can only be marked as broken by one task, but multiple tasks could be marked as a parking lot breaker and exit (or a task multiple times, but that's handled already). I'm not sure what's a good idea. We can't raise any error because there's no good place to raise errors, but I suppose we could warn? Or maybe `broken_by` should be a list.

newsfragments/3081.feature.rst
jakkdl: ah yes, this sounds familiar. thanks

src/trio/_core/_parking_lot.py
jakkdl: should warning be raised if the same task breaks a lot multiple times?

src/trio/_core/_tests/test_parking_lot.py
jakkdl: added comment, slightly different from a5rocks suggestion

src/trio/_core/_parking_lot.py
A5rocks: Yeah that sounds like a fine warning and a task breaking a lot multiple times is probably fine. After all we allow nesting these add/remove_parking_lot_breaker

src/trio/_sync.py
jakkdl: removed `StalledLockError` and reverted to everything using `BrokenResourceError`, but Lock still reraises. I think that's what the conclusion was?

src/trio/_core/_run.py
```diff
@@ -1859,6 +1860,12 @@ def task_exited(self, task: Task, outcome: Outcome[Any]) -> None:
             assert task._parent_nursery is not None, task
             task._parent_nursery._child_finished(task, outcome)
 
+        # before or after the other stuff in this function?
+        if task in GLOBAL_PARKING_LOT_BREAKER:
+            for lot in GLOBAL_PARKING_LOT_BREAKER[task]:
+                lot.break_lot(task)
```
Zac-HD, line 1866: IMO we should do this before the misnested-cancel-scopes check up the top; if that happens to fire we still want locks to break instead of hanging.

src/trio/_sync.py
```diff
@@ -767,7 +780,11 @@ def acquire_nowait(self) -> None:
         return self._lock.acquire_nowait()
 
     async def acquire(self) -> None:
-        """Acquire the underlying lock, blocking if necessary."""
+        """Acquire the underlying lock, blocking if necessary.
+
+        Raises:
+          BrokenResourceError: if the owner of the lock exits without releasing.
```
Zac-HD, line 786:
```suggestion
          BrokenResourceError: if the holder of the underlying lock exits without releasing it.
```

Feel free to ignore, but my personal preference is to say "holder" rather than "owner" of the lock, and in the case of Condition refer to the _underlying lock_ as in the class docstring.

Zac-HD (approved):
Two minor things below, and then unless @njsmith or @oremanj raise any concerns let's merge after the weekend!

src/trio/_sync.py
jakkdl: "owner" is publicly exposed in e.g. `LockStatistics` and already used in several other docstrings, so unless others agree I think it probably makes more sense to be consistent with that.
"underlying" sounds good though

src/trio/_core/_parking_lot.py
```diff
@@ -86,6 +88,33 @@
...
 
+GLOBAL_PARKING_LOT_BREAKER: dict[Task, list[ParkingLot]] = {}
+
+
+def add_parking_lot_breaker(task: Task, lot: ParkingLot) -> None:
+    """Register a task as a breaker for a lot. If this task exits without being removed
+    as a breaker, the lot will break. This will cause an error to be raised for all
+    tasks currently parked in the lot, as well as any future tasks that attempt to
+    park in it.
+    """
```
A5rocks, line 99: These should have a summary.

I think it might be better to link to the `break_lot` to explain the effects of a task exiting without removing themselves.

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +272,34 @@ def repark_all(self, new_lot: ParkingLot) -> None:
         """
         return self.repark(new_lot, count=len(self))
 
+    def break_lot(self, task: Task | None = None) -> None:
+        """Break this lot, causing all parked tasks to raise an error, and any
+        future tasks attempting to park to error. Unpark & repark become no-ops as the
+        parking lot is empty.
+        The error raised contains a reference to the task sent as a parameter.
+        """
```
A5rocks, line 280: Summary line here too. Also "The error raised contains a reference to the task sent as a parameter." should probably just mention that `broken_by` is available on parking lots, because I personally missed that previously.

src/trio/_core/_run.py
```diff
@@ -1820,6 +1821,11 @@ async def python_wrapper(orig_coro: Awaitable[RetT]) -> RetT:
         return task
 
     def task_exited(self, task: Task, outcome: Outcome[Any]) -> None:
+        if task in GLOBAL_PARKING_LOT_BREAKER:
+            for lot in GLOBAL_PARKING_LOT_BREAKER[task]:
+                lot.break_lot(task)
```
A5rocks, line 1826: Maybe add a line comment here like `# break parking lots associated with a task that exited` because otherwise the warning (from multiple tasks breaking a lot) might point to here in the stack trace.

src/trio/_core/_tests/test_parking_lot.py
```diff
@@ -215,3 +218,99 @@ async def test_parking_lot_repark_with_count() -> None:
...
+    async def return_me_and_park(
+        lot: ParkingLot,
+        *,
+        task_status: _core.TaskStatus[_core.Task] = trio.TASK_STATUS_IGNORED,
+    ) -> None:
+        task_status.started(_core.current_task())
+        await lot.park()
+
+    lot = ParkingLot()
+    with RaisesGroup(Matcher(_core.BrokenResourceError, match="Parking lot broken by")):
```
A5rocks, line 313:
```suggestion
    with RaisesGroup(Matcher(_core.BrokenResourceError, match="^Parking lot broken by")):
```
for consistency

src/trio/_sync.py
```diff
@@ -9,6 +9,7 @@
 
 from . import _core
 from ._core import Abort, ParkingLot, RaiseCancelT, enable_ki_protection
+from ._core._parking_lot import add_parking_lot_breaker, remove_parking_lot_breaker
```
A5rocks, line 12: We could import the functions from `._core` for a nicer import

src/trio/_sync.py
```diff
@@ -576,20 +577,30 @@ def acquire_nowait(self) -> None:
...
-            # lock as well.
-            await self._lot.park()
+            try:
+                # NOTE: it's important that the contended acquire path is just
+                # "_lot.park()", because that's how Condition.wait() acquires the
+                # lock as well.
+                await self._lot.park()
+            except trio.BrokenResourceError:
+                raise trio.BrokenResourceError(
+                    "Owner of this lock exited without releasing: {self._owner}",
```
A5rocks, line 602:
```suggestion
                    f"Owner of this lock exited without releasing: {self._owner}",
```

src/trio/_tests/test_sync.py
```diff
@@ -586,3 +589,59 @@ async def lock_taker() -> None:
         await wait_all_tasks_blocked()
         assert record == ["started"]
         lock_like.release()
+
+
+async def test_lock_acquire_unowned_lock() -> None:
+    """Test that trying to acquire a lock whose owner has exited raises an error.
+    Partial fix for https://github.com/python-trio/trio/issues/3035
```
A5rocks, line 596: How about testing the full fix because this PR should be the full fix.

A5rocks (commented):
Mostly just nitpicks

src/trio/_tests/test_sync.py
jakkdl: this+next test is the full fix, so the docstring is outdated. Thanks

Although that issue contains some other loose thoughts (give warning if exiting without releasing the lock, semaphore things that got set aside, add this feature to capacitylimiter/semaphore) that should perhaps be mulled over before closing it.

src/trio/_sync.py
jakkdl: looking forward to getting the linter rule for this

src/trio/_core/_parking_lot.py
jakkdl: linking to `break_lot` sounds good, which removed the need for a summary.

src/trio/_core/_parking_lot.py
jakkdl: not super happy with the docstring, but should be fine now? Feel free to suggest changes

src/trio/_core/_parking_lot.py
```diff
@@ -234,6 +268,38 @@ def repark_all(self, new_lot: ParkingLot) -> None:
...
 
+    def break_lot(self, task: Task | None = None) -> None:
+        """Break this lot, with ``task`` noted as the task that broke it.
+
+        This causes all parked tasks to raise an error, and any
+        future tasks attempting to park to error. Unpark & repark become no-ops as the
+        parking lot is empty.
+
+        The error raised contains a reference to the task sent as a parameter. It is also
+        saved in the ``broken_by`` attribute.
```
A5rocks, line 279: I don't think we do this

src/trio/_core/_run.py
A5rocks: Sorry to be specific:
```suggestion
                lot.break_lot(task)  # task exited while it was a parking lot breaker
```
Because this line specifically will be highlighted in the default traceback handler. (even though custom ones can get context)

A5rocks (approved):
A couple nitpicks but this looks good otherwise!

(I'm not going to lie `add_parking_lot_breaker` was not the best name I've ever come up with. At least it's descriptive...)

src/trio/_core/_parking_lot.py
jakkdl: Ah, the sentence is ambiguous as to where that attribute is. fixed

src/trio/_core/_run.py
jakkdl: I started playing around with this... and it appears that the `RuntimeWarning` can get transformed into a TrioInternalError >.>

jakkdl: we might also want to raise an error if trying to register a task as a breaker that has already exited

jakkdl: > Nevermind, I was completely incorrect about how warnings seem. I'm concerned that this might cause a weird warning location if someone adds multiple tasks as breakers for a single lot and they all exit, but I'm not sure how to better account for that.

yeah it currently is giving me zero useful location info on where the task exited. And even trying to play around with `stacklevel` and [`skip_file_prefixes` from 3.12](https://docs.python.org/3/library/warnings.html#warnings.warn) I fail to get any change in behavior. Not sure if that's me failing or trio doing sus things

src/trio/_core/_tests/test_parking_lot.py
```diff
@@ -215,3 +223,160 @@ async def test_parking_lot_repark_with_count() -> None:
...
+
+    child_task = None
+    with pytest.warns(RuntimeWarning):
+        async with trio.open_nursery() as nursery:
+            child_task = await nursery.start(dummy_task)
+            # registering a task as breaker on an already broken lot is fine... though it
+            # maybe shouldn't be as it will always cause a RuntimeWarning???
+            # Or is this a sign that we shouldn't raise a warning?
+            add_parking_lot_breaker(child_task, lot)
+            nursery.cancel_scope.cancel()
```
jakkdl, line 272: @A5rocks ?

src/trio/_core/_run.py
```diff
@@ -2793,6 +2800,8 @@ def unrolled_run(
             ),
             stacklevel=1,
         )
+    except RuntimeWarning:
+        raise
     except TrioInternalError:
         raise
```
jakkdl, line 2806: This could be too broad (maybe warnings should cause internal errors and prompt people to open issues) or too narrow (if passing through runtimewarnings, why not all warnings?). Could make the `RuntimeWarning` a custom warning class instead

jakkdl (commented):
turns out there were some nasty things still buried >.>

src/trio/_core/_run.py
A5rocks: At worst we could make the warning into a no-op when in trio internals (just duplicate the checks above this).

... but that's not very useful.

src/trio/_core/_tests/test_parking_lot.py
A5rocks: I think it's fine, the main issue I think is just warning in trio internals being misleading.

src/trio/_core/_run.py
A5rocks: Maybe this is enough justification to make `broken_by` a list? It's a slightly worse API but I think it's fine.

src/trio/_core/_tests/test_parking_lot.py
Zac-HD: I concur that this seems fine; we do want to give the warning but it's not otherwise a problem - in typical usage you'd get an error about the lot being broken when you go to park in it, and that will likely still happen here (if it hasn't already!).

src/trio/_core/_tests/test_parking_lot.py
```diff
@@ -249,30 +261,81 @@ async def test_parking_lot_breaker_basic() -> None:
...
             RuntimeWarning,
-            match="attempted to break parking .* already broken by .*",
+            match=warn_str,
         ):
             lot.break_lot(child_task)
         nursery.cancel_scope.cancel()
 
+    # note that this get put into an exceptiongroup if inside a nursery, making any
+    # stacklevel arguments irrelevant
+    with RaisesGroup(Matcher(RuntimeWarning, match=warn_str)):
```
Zac-HD, line 298: I think this is only true because we're using `-Werror`, and if we did `with pytest.raises(...)` it'd show up as a warning rather than being raised as an exception.

src/trio/_core/_tests/test_parking_lot.py
```diff
@@ -249,30 +261,81 @@ async def test_parking_lot_breaker_basic() -> None:
...
     assert lot.broken_by == task
 
+    # register multiple tasks as lot breakers, then have them all exit
+    lot = ParkingLot()
+    child_task = None
+    # This does not give an exception group, as the warning is raised by the nursery
+    # exiting, and not any of the tasks inside the nursery.
+    # And we only get a single warning because... of the default warning filter? and the
+    # location being the same? (because I haven't figured out why stacklevel makes no
+    # difference)
```
Zac-HD, line 314: Yep, [once per location](https://docs.python.org/3/library/warnings.html#warning-filter) is the default, and `with pytest.warns()` clears preexisting filters.

Zac-HD (approved):
Only action items left AFAICT are to remove the `except RuntimeWarning` block, and optionally to clean up any comments that you want to adjust.  Then let's merge 🚀 

Also, thanks again @jakkdl - this has been a rather long project to fix a relatively small issue, but I think we've ended up with a really nice solution and really appreciate all your work on this 😍

src/trio/_core/_run.py
jakkdl: Hmm, I'm torn on this - in large part because `TrioInternalError` prompts users to open an issue to report it, and running pytest with `-Werror` isn't that obscure

Zac-HD: Oh, I see!  Yeah, keeping it with a custom warning type seems good then.

A5rocks: To be honest custom warning type is unnecessary if above `lot.break_lot(task)` we put a `lot.broken_by` check. I'm not sure that indicative of quality of implementation though.

Zac-HD: Requiring that check (and erroring on attempts to break an already-broken lot) seems pretty sensible to me, it's not like `ParkingLot` is an interface that gets called _directly_ very often, and so I don't mind adding some ceremony if that helps catch tricky concurrency bugs.

jakkdl: okay so with the check we no longer raise any warnings on task exit if multiple tasks try to break a task, which means we only ever give a warning if manually breaking a broken lot. That doesn't seem particularly logical to me, and we should probably do one of:
1. make breaking a broken lot a no-op
2. raise an error
3. ...go with `broken_by` being a list (or tuple).

idc much which one though. The warning was added after this convo: https://github.com/python-trio/trio/pull/3081/files/7a1ce5b755b5775153a3525bea793ab7da05fb44#r1764544347 where I argued against the list, but if we no longer want to raise a warning from the internals then maybe the list is the way to go? (that also allows for duplicates in the list).

Zac-HD (approved):
I like the state we've gotten to, and want to use the feature sometime this year, so I propose merging as-is 🙂

src/trio/_core/_run.py
A5rocks: I agree with that ordering (though I might slightly prefer 2 over 1? doesn't matter.)

A5rocks:
> I like the state we've gotten to, and want to use the feature sometime this year, so I propose merging as-is 🙂

We can fix `broken_by`'s representation then merge, everything requires a release anyways :P

jakkdl:
Made `broken_by` a list, and cleaned up `test_parking_lot` that had gotten incredibly messy across all the iterations. I think it's ready to merge at this point after a quick review I didn't do anything dumb at the last second

A5rocks (approved):
Looks good to me based on a quick review

PR #3092, opened 2024-09-22
Add aiologic and culsans to Awesome Trio Libraries in the docs

x42005e1f:
Hi, I'd like to add [aiologic](https://github.com/x42005e1f/aiologic) to your list of awesome libraries. My library can be used to communicate with both threads and event loops in those threads, which can also be seen as an alternative approach to Trio/asyncio interoperability. But I'm not sure about which category aiologic is better to be in, so I put it under "Tools and Utilities". No documentation yet, but hopefully someone will find it useful.


A5rocks:
I was looking at your README and it seems pretty nice! I don't think this is a good idea though:
> Optional [Trio-style checkpoints](https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints)

If something is optional, that means there are no guarantees. And it's the guarantees that every `await` will call a checkpoint that make trio easier to use.

A5rocks (approved):
This seems fine to me, though I'm not too sure about the summary. If you think that describes it best then good enough.

A5rocks (approved):
This still looks good to me :P 

I didn't merge this previously cause I wanted someone else to sanity check. But if nobody does I'll merge this eventually (not that soon though.)

PR #3096, opened 2024-09-27
trio.sleep_forever should be NoReturn

### LINKED ISSUE #3095: `sleep_forever` should be typed NoReturn
I think this would be more correct? the trio-typing stubs did this: https://github.com/python-trio/trio-typing/blob/c7d97072e630cced81cb8a7bb30c0707f1a5d235/trio-stubs/__init__.pyi#L198

Redoubts:
Fixes #3095

### DIFF
--- a/newsfragments/3095.bugfix.rst
+++ b/newsfragments/3095.bugfix.rst
@@ -1,0 +1,1 @@
+Update :func:`trio.sleep_forever` to be `NoReturn`.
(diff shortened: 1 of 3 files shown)


A5rocks (approved):
It's a shame we have to use type ignores but this makes sense to me!

TeamSpen210:
Well, actually mypy is correct. If someone did `trio.lowlevel.reschedule()`, `sleep_forever()` could succeed. That's totally incorrect to do of course. We could just add a `raise TrioInternalError("Should never have been rescheduled!")` after the `await`, which will quiet the type error and make sure you can't do that any more.

CoolCat467 (approved):
Looks great

richardsheridan:
The docstring for this function should also mention that it will raise `RuntimeError` if manually rescheduled.

PR #3097, opened 2024-09-29
Remove 2nd client hello post openssl 1.1.1 bug workaround

CoolCat467:
In this pull request, we remove a workaround for a bug in OpenSSL 1.1.1 with DTLS `ClientHello`s being eaten. There is a comment noting that it should be safe to remove this workaround in 2024 or so, so the time is neigh!


newsfragments/3097.removal.rst
A5rocks, line 1: I think this is a bit wordy as a newsfragment, maybe it should stick to strictly what changed?

CoolCat467: Changed in 7f70723

A5rocks (approved):
I don't know enough about this to say whether it's correct but looks good to me!

jakkdl:
ubuntu 18.06, as mentioned in the comment, reached EOL at 2023-05-31. So yeah I'm guessing it'll be fine

CoolCat467:
Going to merge this soon unless anyone has any additional concerns, consensus seems to be that removing duplicate should be fine.

PR #3098, opened 2024-09-29
Enable flake8 annotations

CoolCat467:
In this pull request, we enable ruff's `flake8-annotations` rule.

One of the more major things this rule complains about is missing type annotations on `self` and `cls` arguments, so there are a lot of places that got `Self` and `type[Self]` annotations added respectfully, as there is not a way to specifically ignore `self` or `cls` in particular without missing others.


TeamSpen210 (changes requested):
Actually, the self/cls issue is because this is enabling the deprecated [`ANN101`](https://docs.astral.sh/ruff/rules/missing-type-self/)/[`ANN102`](https://docs.astral.sh/ruff/rules/missing-type-cls/) rules. Those should be disabled. Also, I really don't see the point anyway in running any of our linters on `notes-to-self`. Those are just historical test/example scripts, they're not supposed to be robust or anything. It just produces churn...

CoolCat467:
> Actually, the self/cls issue is because this is enabling the deprecated [`ANN101`](https://docs.astral.sh/ruff/rules/missing-type-self/)/[`ANN102`](https://docs.astral.sh/ruff/rules/missing-type-cls/) rules. Those should be disabled. Also, I really don't see the point anyway in running any of our linters on `notes-to-self`. Those are just historical test/example scripts, they're not supposed to be robust or anything. It just produces churn...

Did a force push basically redoing everything, much easier than manually undoing adding self everywhere

TeamSpen210 (approved):

pyproject.toml
```diff
@@ -151,6 +154,8 @@ extend-ignore = [
 'src/trio/lowlevel.py' = ['F401']
 'src/trio/socket.py' = ['F401']
 'src/trio/testing/__init__.py' = ['F401']
+# Don't check annotations in notes-to-self
+'notes-to-self/*.py' = ['ANN']
```
A5rocks, line 158: If we're going to do this, then maybe restore `notes-to-self/*.py` to what they were before (I *think* the correct invocation is `git reset main -- notes-to-self`?)

src/trio/_core/_concat_tb.py
```diff
@@ -86,7 +86,9 @@ def copy_tb(base_tb: TracebackType, tb_next: TracebackType | None) -> TracebackT
     def copy_tb(base_tb: TracebackType, tb_next: TracebackType | None) -> TracebackType:
         # tputil.ProxyOperation is PyPy-only, and there's no way to specify
         # cpython/pypy in current type checkers.
-        def controller(operation: tputil.ProxyOperation) -> Any | None:  # type: ignore[no-any-unimported]
+        def controller(  # type: ignore[no-any-unimported]
+            operation: tputil.ProxyOperation,
+        ) -> Any | None:  # noqa: ANN401  # Using Any
```
A5rocks, line 91: Doesn't `mypy` already warn about using `Any`? Is this rule redundant? (Maybe it covers places `mypy` doesn't?)

CoolCat467: It raised alerts in several places mypy was fine with and lead to better type annotations for keyword arguments in several places, among other small things.

src/trio/_tests/test_util.py
```diff
@@ -117,8 +122,10 @@ async def f() -> None:  # pragma: no cover
 
         if sys.version_info < (3, 11):
             # not bothering to type this one
```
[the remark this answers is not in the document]
CoolCat467: Removed in 4be420b

src/trio/_core/_concat_tb.py
A5rocks: I assume this is because we haven't enabled `disallow_any_explicit`? I'd rather use `mypy` for this than `ruff`.

(not sure if that leads to *more* errors though...)

CoolCat467: Probably yea

src/trio/_subprocess_platform/__init__.py
```diff
@@ -106,12 +106,12 @@ def create_pipe_from_child_output():
...
         from .._windows_pipes import PipeReceiveStream, PipeSendStream
 
-        def create_pipe_to_child_stdin():
+        def create_pipe_to_child_stdin() -> tuple[PipeSendStream, int]:
             # for stdin, we want the write end (our end) to use overlapped I/O
             rh, wh = windows_pipe(overlapped=(False, True))
             return PipeSendStream(wh), msvcrt.open_osfhandle(rh, os.O_RDONLY)
 
-        def create_pipe_from_child_output():
+        def create_pipe_from_child_output() -> tuple[PipeReceiveStream, int]:
```
jakkdl, line 114: there's a mypy setting for catching this as well

src/trio/_tests/test_highlevel_open_tcp_stream.py
```diff
@@ -370,7 +372,11 @@ async def run_scenario(
     trio.socket.set_custom_socket_factory(scenario)
 
     try:
-        stream = await open_tcp_stream("test.example.com", port, **kwargs)
+        stream = await open_tcp_stream(
+            "test.example.com",
+            port,
+            **kwargs,  # type: ignore[arg-type]
+        )
```
jakkdl, line 379: [add a comment explaining] why can't open_tcp_stream kwargs be typed?

src/trio/_tests/test_ssl.py
```diff
@@ -251,9 +256,10 @@ def __init__(self, sleeper: None = None) -> None:
             "simultaneous calls to PyOpenSSLEchoStream.receive_some",
         )
 
+        self.sleeper: Callable[[str], Awaitable[None]]
         if sleeper is None:
 
-            async def no_op_sleeper(_: object) -> None:
+            async def no_op_sleeper(_: str) -> None:
                 return
```
jakkdl, line 264: nit: there's really no reason to narrow this.

src/trio/_tests/test_ssl.py
```diff
@@ -411,21 +417,24 @@ def ssl_wrap_pair(
         client_transport,
         client_ctx,
         server_hostname="trio-test-1.example.org",
-        **client_kwargs,
+        **client_kwargs,  # type: ignore[arg-type]
     )
```
jakkdl, line 421: swapping an `Any` for a `# type: ignore[arg-type]` is borderline making it worse imo, especially in such a short function and it being a test function that nobody will call.

At least add a comment explaining why it's necessary to ignore

src/trio/_tests/test_highlevel_open_tcp_stream.py
jakkdl: right, because the latter allows passing 2x `<type>`. I suspect the reason for the `kwargs` in the first place is laziness/ease of maintainability, but given we're digging this hard anyway you could probably replace them all. You might want to consider a separate PR though

src/trio/_tests/test_ssl.py
CoolCat467: Added comment in c4fccd0

src/trio/_tests/test_ssl.py
CoolCat467: Reverted in c4fccd0

src/trio/_tests/test_highlevel_open_tcp_stream.py
CoolCat467: Added comment in c4fccd0

src/trio/_core/_concat_tb.py
CoolCat467: Tried adding `disallow_any_explicit`, and it raises 94 new errors. Would be better to handle in another pull request if we handle at all, `Any` can be the best way in some cases.

A5rocks: If `ANN401` is less good than `disallow_any_explicit`, then I think we should disable it. Up to you though

src/trio/_subprocess_platform/__init__.py
CoolCat467: If you are referring to `disallow_incomplete_defs` and `check_untyped_defs`, those two are already enabled for Trio.

src/trio/_core/_concat_tb.py
CoolCat467: I think that it would be best to enable more mypy checking, but to save it for another pull request. In other pull request, then disabling `ANN401` might be a good idea, because then there will be duplicate disables for lines matching this where `Any` was required.

src/trio/_tests/test_highlevel_open_tcp_stream.py
jakkdl: rereading my previous comment I realize I probably should've been more explicit: what I was suggesting was to stop using `**kwargs` altogether and simply define the parameters to `run_scenario` to be equivalent to the ones that `open_tcp_stream` expects

src/trio/_subprocess_platform/__init__.py
jakkdl: I meant `disallow_untyped_defs`

jakkdl:
> Actually, the self/cls issue is because this is enabling the deprecated [`ANN101`](https://docs.astral.sh/ruff/rules/missing-type-self/)/[`ANN102`](https://docs.astral.sh/ruff/rules/missing-type-cls/) rules. Those should be disabled. Also, I really don't see the point anyway in running any of our linters on `notes-to-self`. Those are just historical test/example scripts, they're not supposed to be robust or anything. It just produces churn...

tbh, is there any reason to keep them around at all anymore? We could dump them in a gist and save a link somewhere if we want to save them somewhere (other than in the git history)

TeamSpen210:
Could create a git tag to mark which commit is is with the last version.

src/trio/_core/_concat_tb.py
jakkdl: agreed. disable ANN401 for now and enable `disallow_any_explicit` in a separate PR. Possibly same for `disallow_untyped_defs`.

pyproject.toml
jakkdl: +1

CoolCat467: Restored in 12fe0343

src/trio/_core/_concat_tb.py
CoolCat467: Disabled in 8123b72b

src/trio/_tests/test_highlevel_open_tcp_stream.py
CoolCat467: Fixed in 42be7f3

pyproject.toml
```diff
@@ -160,6 +164,8 @@ extend-ignore = [
 'src/trio/_abc.py' = ['A005']
 'src/trio/_socket.py' = ['A005']
 'src/trio/_ssl.py' = ['A005']
+# Don't check annotations in notes-to-self
+'notes-to-self/*.py' = ['ANN001', 'ANN002', 'ANN003', 'ANN201', 'ANN202', 'ANN204']
```
jakkdl, line 168: whichever is merged latest of this and #3117 can remove these lines.

src/trio/_subprocess_platform/__init__.py
CoolCat467: `disallow_untyped_defs` is also already enabled

jakkdl: then why did this not get caught by mypy previously??

CoolCat467: Not really sure

CoolCat467:
Planning on merging this in a day or two unless anyone has any other comments.

PR #3101, opened 2024-10-05
Add documentation for `statistics()` and `aclose()` on memory channels

A5rocks:
I'm pretty sure the reason Sphinx isn't showing them is because they have no docstrings. Easy fix.

Mentioned on Gitter.


A5rocks:
pre-commit.ci autofix

src/trio/_channel.py
```diff
@@ -111,7 +111,7 @@ def __init__(self, max_buffer_size: int | float):  # noqa: PYI041
 
 
 @attrs.frozen
-class MemoryChannelStats:
+class MemoryChannelStatistics:
```
A5rocks, line 114: I *think* this is backwards compatible as far as we care.

CoolCat467: If it wasn't publicaly re-exported originally, I would think this should count as backwards compatible, and if anyone does have issues we can tell them its their fault for digging into Trio's internals.

CoolCat467 (approved):
Looks good to me, don't foresee any issues with this.

TeamSpen210 (approved):
Looks good.

src/trio/_tests/_check_type_completeness.json
```diff
@@ -20,13 +20,10 @@
     "all": [
         "No docstring found for class \"trio.MemoryReceiveChannel\"",
         "No docstring found for class \"trio._channel.MemoryReceiveChannel\"",
-        "No docstring found for function \"trio._channel.MemoryReceiveChannel.statistics\"",
-        "No docstring found for class \"trio._channel.MemoryChannelStats\"",
-        "No docstring found for function \"trio._channel.MemoryReceiveChannel.aclose\"",
+        "No docstring found for class \"trio.MemoryChannelStatistics\"",
+        "No docstring found for class \"trio._channel.MemoryChannelStatistics\"",
```
jakkdl, line 24: I guess you could've added a docstring to the class at the same time to get rid of these as well

PR #3103, opened 2024-10-09
Handle RTD enabling addons

### LINKED ISSUE #3064: Make sure trio works with Read the Docs Addons
I saw https://about.readthedocs.com/blog/2024/07/addons-by-default/ and it looks like we might have to make some changes. Its not very clear to me what changes exactly though. And it seems the theme support for it is in a dev release so we should probably wait.

A5rocks:
Fixes #3064

I'm not sure whether this is required because the docs look fine, but then of course the docs have very many hacks that I'm not aware of.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -381,10 +381,8 @@ suppress_warnings = ["epub.unknown_project_files"]
 # We have to set this ourselves, not only because it's useful for local
 # testing, but also because if we don't then RTD will throw away our
 # html_theme_options.
-import sphinx_rtd_theme
 
 html_theme = "sphinx_rtd_theme"
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -374,10 +381,8 @@ def add_mapping(
 # We have to set this ourselves, not only because it's useful for local
 # testing, but also because if we don't then RTD will throw away our
 # html_theme_options.
-import sphinx_rtd_theme
```
jakkdl, line 384:
```suggestion
```

let's attach the comment to the line it's talking about

jakkdl (approved):
seems to work? The generated docs looks normal at least.

A5rocks:
OK I don't know why RTD is failing to clone this PR now...

A5rocks:
Well it seems to work now. I'll merge this soon enough unless someone has comments.

CoolCat467 (approved):
Looks good

Issue #2135, opened 2021-10-05
Can't pickle `trio` as a package


uduse:
```python
import trio
import pickle

pickle.dumps(trio)
```

gives

```python
TypeError: cannot pickle '_ModuleWithDeprecations' object
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_9132/1773362268.py in <module>
      2 import pickle
      3 
----> 4 pickle.dumps(trio)

TypeError: cannot pickle '_ModuleWithDeprecations' object
```

This is problematic for serializing objects that have methods that use trio somehow.

uduse:
This patch seems to work, do you want me to PR as a method?

```python
import trio
import pickle

def __reduce__(self):
    return (self.__class__, tuple())


trio._deprecate._ModuleWithDeprecations.__reduce__ = __reduce__

pickle.dumps(trio._deprecate._ModuleWithDeprecations("hello"))
```

njsmith:
It's an internal implementation detail whether any given module is a `_ModuleWithDeprecations` or a regular module. So I think we want to pickle the same way as a regular module -- which for `pickle` is not supported, but for `cloudpickle` (which IIRC `ray` uses) it basically pickles to `import <module name>`.

njsmith:
...Actually, I'm getting a bit confused about cloudpickle here -- I get `cannot pickle 'module' object` for some modules and not others.

Can you show your real code that's failing, or at least the real traceback?

uduse:
For example,

```python
import ray
import trio

ray.init()


@ray.remote
def task():
    async def sleeper():
        await trio.sleep(1)

    trio.run(sleeper)


task.remote()
```

This gives:

```bash
/root/.local/share/virtualenvs/moozi-g1CZ00E9/lib/python3.8/site-packages/trio/_core/_multierror.py:511: RuntimeWarning: You seem to already have a custom sys.excepthook handler installed. I'll skip installing Trio's custom handler, but this means MultiErrors will not show full tracebacks.
  warnings.warn(
2021-10-05 06:28:59,438 INFO services.py:1250 -- View the Ray dashboard at http://127.0.0.1:8266
Traceback (most recent call last):
  File "scripts/trio_pickle.py", line 12, in <module>
    task.remote()
  File "/root/.local/share/virtualenvs/moozi-g1CZ00E9/lib/python3.8/site-packages/ray/remote_function.py", line 123, in _remote_proxy
[...]

PR #3105, opened 2024-10-09
add __copy__ and __deep_copy__ to Cancelled

jakkdl:
See https://github.com/python-trio/flake8-async/issues/298 for why I want to be able to copy `Cancelled`.

I suppose there's a minor footgun added if people start copying `Cancelled`, `NoPublicConstructor` was added for a reason, but I think the gain outweighs it. Though of course another option is to simply get rid of `NoPublicConstructor`

I initially tried to write cleaner methods, where I was using `__dict__` and the like, but they do not appear to pick up the relevant dunders so I simply listed everything relevant from https://docs.python.org/3/library/exceptions.html#exception-context

### DIFF
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+import copy
+
 from trio._util import NoPublicConstructor, final
 
 
@@ -62,6 +66,31 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     def __str__(self) -> str:
         return "Cancelled"
+
+    def __copy__(self) -> Cancelled:
+        result = Cancelled.__new__(Cancelled)
+        result.__context__ = self.__context__
+        result.__cause__ = self.__cause__
+        result.__suppress_context__ = self.__suppress_context__
+        result.__traceback__ = self.__traceback__
+        if hasattr(self, "__notes__"):
+            result.__notes__ = self.__notes__  # type: ignore[attr-defined]
+        return result
+
+    def __deepcopy__(self, memo: dict[int, object]) -> Cancelled:
+        result = Cancelled.__new__(Cancelled)
+        memo[id(self)] = result
+        keys = (
+            "__context__",
+            "__cause__",
+            "__suppress_context__",
+            "__traceback__",
+            "__notes__",
+        )
+        for key in keys:
+            if hasattr(self, key):
+                setattr(result, key, copy.deepcopy(getattr(self, key), memo))
+        return result
 
 
 class BusyResourceError(Exception):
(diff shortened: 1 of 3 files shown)


CoolCat467 (approved):

src/trio/_core/_exceptions.py
```diff
@@ -63,6 +67,31 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
     def __str__(self) -> str:
         return "Cancelled"
 
+    def __copy__(self) -> Cancelled:
+        result = Cancelled.__new__(Cancelled)
+        result.__context__ = self.__context__
+        result.__cause__ = self.__cause__
+        result.__suppress_context__ = self.__suppress_context__
+        result.__traceback__ = self.__traceback__
+        if hasattr(self, "__notes__"):
```
oremanj, line 76: The notes are part of the dict. You probably want to copy the rest of the dict as well. So maybe just `result.__dict__ = self.__dict__.copy()`?

src/trio/_core/_exceptions.py
```diff
@@ -63,6 +67,31 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
...
+        keys = (
+            "__context__",
+            "__cause__",
+            "__suppress_context__",
+            "__traceback__",
+            "__notes__",
+        )
+        for key in keys:
+            if hasattr(self, key):
+                setattr(result, key, copy.deepcopy(getattr(self, key), memo))
```
oremanj, line 92:
```suggestion
        )
        for key in keys:
            if hasattr(self, key):
                setattr(result, key, copy.deepcopy(getattr(self, key), memo))
        result.__dict__ = copy.deepcopy(self.__dict__, memo)
```

oremanj (commented):
This is a far better copy than most exception objects have, FYI. (I left a comment on the flake8-trio issue that motivated this.)

jakkdl:
> This is a far better copy than most exception objects have, FYI. (I left a comment on the flake8-trio issue that motivated this.)

Thank you! I think we don't need to bother doing it either then, and I'll throw out the complicated logic.

src/trio/_core/_exceptions.py
jakkdl: thanks, I threw everything out in favor of just copying `__dict__`.

jakkdl:
it appears `copy.copy` has bigger issues with exceptions than `trio.Cancelled`, see https://github.com/python-trio/flake8-async/issues/298#issuecomment-2424971988
I'll probably close this PR

jakkdl:
okay I think my original reasoning for this is no longer valid - copying is too unreliable to use in a general library like trio-websocket that needs to handle arbitrary user exceptions, but we should probably make it possible to copy & pickle `Cancelled` for general usability reasons.

That's fairly straightforward by implementing `__reduce__`, but it's maybe bad that all other `NoPublicConstructor` classes remain unpickleable? I'm at least finding one somewhat related issue: https://github.com/python-trio/trio/issues/2135

PR #3107, opened 2024-10-12
fix some extra cyclic garbage in NurseryManager

graingert:


graingert:
I'll write a test for this and shape it into a proper PR shortly

src/trio/_core/_run.py
```diff
@@ -1023,6 +1023,7 @@ async def __aexit__(
         exc: BaseException | None,
         tb: TracebackType | None,
     ) -> bool:
+        del tb
```
A5rocks, line 1026: This seems intimidating, when you're adding a test could you add another to make sure no stackframes are getting lost? (I assume they're part of `exc` in addition?)

graingert: The tb is never needed in a context manager, and cpython are looking to deprecate the three arg form in favor of a one arg exc_value only form

src/trio/_core/_run.py
```diff
@@ -1043,7 +1044,7 @@ async def __aexit__(
                 value.__context__ = old_context
                 # delete references from locals to avoid creating cycles
                 # see test_cancel_scope_exit_doesnt_create_cyclic_garbage
-                del _, combined_error_from_nursery, value, new_exc
+                del _, combined_error_from_nursery, value, new_exc, exc
```
graingert, line 1047: I think I misunderstood this. There's no cycle here because `exc.__traceback__` doesn't see this frame

Issue #2670, opened 2023-06-25
@enable_ki_protection breaks inspect.iscoroutinefunction


jakkdl:
```python
>>> from trio.lowlevel import enable_ki_protection
>>> async def foo():
...     return
... 
>>> @enable_ki_protection
... async def bar():
...     return
... 
>>> import inspect
>>> inspect.iscoroutinefunction(foo)
True
>>> inspect.iscoroutinefunction(bar)
False
```

Found when I tried to update `coroutine_or_error` in `trio/_util.py` to use `inspect.iscoroutinefunction` in #2668 

https://github.com/python-trio/trio/blob/314663806cbaf717d70d70b71631e340396a1bb1/trio/_util.py#L137-L141

There are also a ton of wrapper/helper methods in tests that un-coroutinefunction-ify functions, which'd need to be modified in order to actually use `iscoroutinefunction` in `couroutine_or_error`. This means it'd probably also break a bunch of downstream code if `coroutine_or_error` was rewritten, but the upside of not having to call the function might mean that it's worth it (after a deprecation period)? That's probably for another issue though.

A5rocks:
3.12+ has it easy: https://github.com/python/cpython/pull/99247

I'm not sure about before that.

A5rocks:
Here's some spooky stuff that works pre-3.12:
```pycon
(.venv) PS C:\Users\A5rocks\Documents\trio> ipython
Python 3.11.3 (tags/v3.11.3:f3909b8, Apr  4 2023, 23:49:59) [MSC v.1934 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.34.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import functools

In [2]: class Based(functools.partial):
   ...:     def __init__(self, func):
   ...:         super().__new__(functools.partial, func)
   ...:
   ...:     def __call__(self):
   ...:         print("insert logic to call `self.func` here")
   ...:

In [3]: @Based
   ...: async def async_function():
   ...:     ...
   ...:

In [4]: async_function
Out[4]: Based(<function async_function at 0x000001B09888FF60>)

In [5]: import inspect

In [6]: inspect.iscoroutinefunction(async_function)
Out[6]: True

In [7]: async_function()
insert logic to call `self.func` here
```

jakkdl:
[...]

Issue #1752, opened 2020-10-11
crash when compiled with nuitka


remdragon:
I got the following crash trying to compile my app with nuitka. I'm guessing this is something they need to fix, but I thought I would share it here in case it helps:

```
Traceback (most recent call last):
  File "C:\cvs\itas\IT20FB~1\ITASPI~1.DIS\trio\_core\_run.py", line 2003, in unrolled_run
  File "C:\cvs\itas\IT20FB~1\ITASPI~1.DIS\trio\_core\_run.py", line 1388, in spawn_impl
AttributeError: 'NoneType' object has no attribute 'f_locals'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\cvs\itas\IT20FB~1\ITASPI~1.DIS\ItasPinClientPortal.py", line 3253, in <module>
  File "C:\cvs\itas\IT20FB~1\ITASPI~1.DIS\trio\_core\_run.py", line 1887, in run
  File "C:\cvs\itas\IT20FB~1\ITASPI~1.DIS\trio\_core\_run.py", line 2202, in unrolled_run
trio.TrioInternalError: internal error in Trio - please file a bug!
```

Python 3.8.5
Trio 0.16.0
Quart 0.12.0
Quart-Trio 0.5.1
Nuitka 0.6.9.2

oremanj:
Nuitka apparently produces coroutine objects that don't have an associated frame. This is not in keeping with standard Python behavior, and breaks Trio's KeyboardInterrupt protection mechanism. (In general, producing a fully drop-in Python compiler is *extremely challenging*; there are many dark corners like this. I wasn't previously familiar with Nuitka but I took a brief look at their documentation and it does not fill me with optimism.)

If you want your Trio app to run faster, you'll probably have better luck running it under PyPy than under Nuitka.

njsmith:
To be fair, the way trio messes with `f_locals` there is a particularly egregious use of some of Python's dark corners. I don't expect Nuitka will ever be able to support that, but we could potentially switch to another mechanism for the control-C handling stuff. My suggestion would be to try temporarily commenting out that code, to see how much you have to change to get trio running under Nuitka and how much speedup it gives, and whether there are any other problems you have to work around.

remdragon:
Fortunately Ctrl+C handling isn't a huge priority for this app, so I decided to pursue this further.

I changed that line to this:
```
[...]

Issue #2454, opened 2022-10-26
use of locals() / f_locals() foils pypy


belm0:
Mutation via `locals()`, `f_locals()` causes pypy to abort traces with "force quasi-immut".   Trio does this to manage KI protection state, and the most common API calls like `current_time()` are implicated.  So pypy spends a lot of time compiling traces, only to abort them.

some ideas for fixing discussed on gitter starting at https://gitter.im/python-trio/general?at=6357f654cf41c67a5cc9b993

NOTE:  a tracing JIT like PyPy is not going to do well anyway with async apps that have a high rate of context switches combined with heterogenous workloads, since the tracing will happen across context switches to other coroutines (which are arbitrarily occurring and arbitrarily ordered).

PR #3110, opened 2024-10-13
gh-3108: Avoid materializing f_locals for KI protection

graingert:
Fixes https://github.com/python-trio/trio/issues/3108
Fixes https://github.com/python-trio/trio/issues/2670

In this pull request, we avoid materializing `f_locals` by using weak references to code objects instead.


src/trio/_core/_run.py
```diff
@@ -1871,7 +1889,6 @@ async def python_wrapper(orig_coro: Awaitable[RetT]) -> RetT:
 
             coro = python_wrapper(coro)
```
graingert, line 1890: I think this isn't needed now that we don't use cr_frame, but I know that various things expect trio.lowlevel.current_task().cr_await and trio.lowlevel.current_task().cr_frame to work so I'm leaving it here

src/trio/_core/_ki.py
```diff
@@ -117,90 +125,37 @@ def currently_ki_protected() -> bool:
...
 
-            return wrapper
 
-    return decorator
+def enable_ki_protection(f: _T_supports_code, /) -> _T_supports_code:
+    """Decorator to enable KI protection."""
+    orig = f
 
+    if legacy_isasyncgenfunction(f):
+        f = f.__wrapped__  # type: ignore
```
graingert, line 140: codecov is saying this line isn't covered but it's definitely covered by test_async_generator_agen_protection

src/trio/_core/_ki.py
```diff
@@ -117,90 +125,37 @@ def currently_ki_protected() -> bool:
...
-            @wraps(fn)
-            def wrapper(*args: ArgsT.args, **kwargs: ArgsT.kwargs) -> RetT:
-                sys._getframe().f_locals[LOCALS_KEY_KI_PROTECTION_ENABLED] = enabled
-                return fn(*args, **kwargs)
+_T_supports_code = TypeVar("_T_supports_code", bound=_SupportsCode)
 
-            return wrapper
 
-    return decorator
+def enable_ki_protection(f: _T_supports_code, /) -> _T_supports_code:
```
graingert, line 135: there's a slight user facing change to how this works, if you do something like:

```python
def foo():
    yield

@enable_ki_protection
def baz():
    return foo()
```

then ki protection is only enabled on baz, and not on foo, wheras before ki protection would only be on foo but not on baz

src/trio/_core/_ki.py
```diff
@@ -83,18 +77,97 @@
...
 
-# We use this special string as a unique key into the frame locals dictionary.
-# The @ ensures it is not a valid identifier and can't clash with any possible
-# real local name. See: https://github.com/python-trio/trio/issues/469
-LOCALS_KEY_KI_PROTECTION_ENABLED: Final = "@TRIO_KI_PROTECTION_ENABLED"
+_T = TypeVar("_T")
+
+
+class _IdRef(weakref.ref[_T]):
+    slots = "_hash"
```
graingert, line 84:
```suggestion
    __slots__ = ("_hash", )
```

src/trio/_core/_ki.py
graingert: Oh it's covered now

newsfragments/3108.bugfix.rst
```diff
@@ -0,0 +1 @@
+avoid materializing frame.f_locals in KI protection code, this causes locals to persist in the frame.f_locals snapshot even if they are deleted on 3.12 and below
```
oremanj, line 1: I think this should be cast as a feature -- in addition to fixing locals materialization, I would expect it to also be good for performance and tracebacks because we're adding extra frames in far fewer places. Also, please use rST formatting and full sentences in newsfragments so our release notes are pleasant for users to read.

src/trio/_core/_ki.py
```diff
@@ -83,18 +77,108 @@
...
+
+        if not isinstance(other, _IdRef):
+            return NotImplemented
+
+        my_obj = None
+        other_obj: Any = None
+        try:
+            my_obj = self()
+            other_obj = other()
+            return my_obj is not None and other_obj is not None and my_obj is other_obj
```
oremanj, line 111: Technically `other_obj is not None` is not needed here; if my_obj is not None, then `my_obj is other_obj` implies that other_obj is also not None

src/trio/_core/_ki.py
```diff
@@ -117,90 +201,37 @@ def currently_ki_protected() -> bool:
...
 
+    if legacy_isasyncgenfunction(f):
+        f = f.__wrapped__  # type: ignore
 
-# pyright workaround: https://github.com/microsoft/pyright/issues/5866
-class KIProtectionSignature(Protocol):
-    __name__: str
+    code = f.__code__.replace()
+    _CODE_KI_PROTECTION_STATUS_WMAP[code] = True
+    f.__code__ = code
```
oremanj, line 220: Why do you need to generate a new code object here rather than adding `f.__code__` to the map?

src/trio/_core/_run.py
```diff
@@ -2712,6 +2730,11 @@ def unrolled_run(
 
                 next_send_fn = task._next_send_fn
                 next_send = task._next_send
+                run_with = (
+                    run_with_ki_protection_enabled
+                    if task._ki_protected
+                    else run_with_ki_protection_disabled
+                )
```
oremanj, line 2737: I would prefer a different approach to task-level KI protection. This one adds function call overhead to every task step, as well as a traceback frame for every level in the task tree. Here's an alternative idea: use a contextvar to indicate task-level KI protection, and have `_run` tell the KIManager which code object corresponds to `unrolled_run`. If stack unwinding reaches that object, KI protection should be based on the contextvar. `Context.run` is written in C, so as long as the contextvar is set to enable KI protection when outside of any task, this approach should be just as safe as the one you're using.

src/trio/_core/_run.py
```diff
@@ -2724,16 +2747,17 @@ def unrolled_run(
...
                     # So now we send in the Outcome object and unwrap it on the
                     # other side.
-                    msg = task.context.run(next_send_fn, next_send)
+                    msg = task.context.run(run_with, next_send_fn, next_send)
                 except StopIteration as stop_iteration:
                     final_outcome = Value(stop_iteration.value)
                 except BaseException as task_exc:
                     # Store for later, removing uninteresting top frames: 1
                     # frame we always remove, because it's this function
+                    # another is the run_with
```
oremanj, line 2756: "because it's this function another is the run_with catching it" does not make sense

oremanj (commented):
This is a very welcome change; the old style of KI protection was responsible for a good bit of performance overhead. Similar ones have been discussed and even prototyped before although nothing ultimately came of that. I would encourage you to read #733 and #1508. This change doesn't need to go as far as is discussed there, but there might be useful things to borrow.

src/trio/_core/_ki.py
graingert: This is required for test_ki_does_not_leak_accross_different_calls_to_inner_functions to pass

src/trio/_core/_ki.py
graingert, line 111:
```suggestion
            return my_obj is not None and my_obj is other_obj
```

src/trio/_core/_run.py
graingert: Ah that's neat, but I don't think a contextvar is necessary. I think a threadlocal should be good enough

oremanj: The reason to use a contextvar is that we are already atomically switching contexts when we enter and exit the task. If you use a thread-local, then you have a brief period where the task is not running but the KI protection is still as if the task were running. This must be managed carefully in order to avoid an unprotected KI being raised directly inside `unrolled_run`. The contextvar solution is more obviously correct IMO.

src/trio/_core/_run.py
graingert: this will be removed when switching to the contextvar

src/trio/_core/_run.py
graingert: A contextvar won't work as someone could do:

```python

async def some_task():
    def unprotected():
        ...
    context.run(unprotected)
```

newsfragments/3108.bugfix.rst
graingert: could you help me out with the phrasing for this, please?

oremanj:
```
Rework KeyboardInterrupt protection to track code objects, rather than frames,
as protected or not. This should substantially reduce its overhead, and also fixes
an issue with the previous implementation: locals' lifetimes will no longer be
extended by materialization in the ``frame.f_locals`` dictionary that the previous
KI protection logic needed to access to do its work.
```

src/trio/_core/_ki.py
oremanj: This seems like a very big hammer to serve a pretty obscure case. For top-level functions, it's minorly increasing import time to no benefit. For closures, each instance of the closure now has to carry around its own 200-byte code object, instead of all of them being able to share one. How would you feel about instead documenting that we expect `@enable/disable_ki_protection` to be applied or not to a given function in the source file in a consistent way, and optionally adding another interface that does dup the code object for use when the protection status might be runtime-dependent?

src/trio/_core/_run.py
graingert: I've pushed a new solution that inspects to see if current_task().coro.cr_frame is in the stack, and if so returns current_task()._ki_protected

src/trio/_core/_ki.py
graingert: I think there's other cases where functions can get the same code object, like importlib.reload

I think by ignoring this hang-nail we can use a regular equality based WeakKeyDictionary rather than building a complicated identity based one

newsfragments/3108.bugfix.rst
graingert:
```suggestion
Rework KeyboardInterrupt protection to track code objects, rather than frames,
as protected or not. This should substantially reduce its overhead, and also fixes
an issue with the previous implementation: locals' lifetimes will no longer be
extended by materialization in the ``frame.f_locals`` dictionary that the previous
KI protection logic needed to access to do its work.
```

src/trio/_core/_ki.py
graingert: How about we issue an exception if the code object is already in the weak key dictionary or if the function doesn't have an importable name?

oremanj: I still would prefer to just not worry about enforcing this (though it's good to document the caveat and provide an alternative that fully supports this sort of runtime variation). All the possible cures seem worse than the disease to me. `@disable_ki_protection` is basically never used, so the practical failure mode is that a conditionally-KI-protected closure will be always KI protected instead. It's really hard for me to imagine any actual harm coming from that. Is there a specific scenario you're thinking of that makes you more concerned about this case?

I would be OK with a warning in the cases you describe, as long as there's something reasonable the user can do to satisfy it (like a kwarg for the decorator that enables dup'ing the code object or that promises all the registrations will match). I don't think it should be a hard error, because this would also fire for a closure that is consistently KI-protected every time it's defined.

graingert: hmm I think I'll remove the code object copy for now

newsfragments/3108.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Rework KeyboardInterrupt protection to track code objects, rather than frames,
+as protected or not. This should substantially reduce its overhead, and also fixes
```
graingert, line 2: this PR actually solves at least 4 issues and I'm not sure how to format it nicely:

* the f_locals materialization
* ki_protection is now applied as soon as the function code is pushed to the stack, so now it will work for weakref finalizers where finalization is crucial
* ki_protection doesn't hide the type of the decorated function so inspect.isgeneratorfunction and inspect.iscoroutinefunction works
* "@TRIO_KI_PROTECTION_ENABLED" doesn't show up when debugging - although maybe that was useful in some cases

A5rocks: For the third one, you can make another newsfragment called 2670.something.rst to separate it out.

After all, that's https://github.com/python-trio/trio/issues/2670.

src/trio/_core/_ki.py
```diff
@@ -83,20 +78,120 @@
...
+    def __init__(self) -> None:
+        self._data: dict[_IdRef[_KT], _VT] = {}
+
+        def remove(
+            k: _IdRef[_KT],
+            selfref: weakref.ref[
+                WeakKeyIdentityDictionary[_KT, _VT]
+            ] = weakref.ref(  # noqa: B008  # function-call-in-default-argument
+                self,
+            ),
```
A5rocks, line 137: I'm not completely understanding what this code does here, but couldn't `selfref` here be a variable defined in `__init__` and then it's just `nonlocal`?

Is this just some lookup optimization?

A5rocks, line 137: I'm not completely understanding what this code does here, but couldn't `selfref` here be a variable defined in `__init__` and then it's just `nonlocal`?

Is this just some lookup optimization?

src/trio/_core/_ki.py
```diff
@@ -83,20 +78,120 @@
...
 def ki_protection_enabled(frame: types.FrameType | None) -> bool:
+    try:
+        task = GLOBAL_RUN_CONTEXT.task
+    except AttributeError:
+        task_ki_protected = False
+        task_frame = None
+    else:
+        task_ki_protected = task._ki_protected
+        task_frame = task.coro.cr_frame
+        del task
```
A5rocks, line 182: I don't get this `del`: what would capture `task` if this wasn't the case?

src/trio/_core/_ki.py
```diff
@@ -117,90 +212,33 @@ def currently_ki_protected() -> bool:
...
 
-            return wrapper
 
-    return decorator
+def enable_ki_protection(f: _T_supports_code, /) -> _T_supports_code:
+    """Decorator to enable KI protection."""
+    orig = f
 
+    if legacy_isasyncgenfunction(f):
+        f = f.__wrapped__  # type: ignore
```
A5rocks, line 227: To get rid of the `type: ignore` I assume we could make `legacy_isasyncgenfunction` a TypeGuard?

src/trio/_core/_ki.py
```diff
@@ -117,90 +212,33 @@ def currently_ki_protected() -> bool:
...
-class KIProtectionSignature(Protocol):
-    __name__: str
+    _CODE_KI_PROTECTION_STATUS_WMAP[f.__code__] = True
+    return orig
 
-    def __call__(self, f: CallableT, /) -> CallableT:
-        pass
 
+def disable_ki_protection(f: _T_supports_code, /) -> _T_supports_code:
+    """Dectorator to disable KI protection."""
```
A5rocks, line 234:
```suggestion
    """Decorator to disable KI protection."""
```

newsfragments/3108.bugfix.rst
oremanj: Here's my attempt:
```
Rework KeyboardInterrupt protection to track code objects, rather than frames,
as protected or not. The new implementation no longer needs to access
``frame.f_locals`` dictionaries, so it won't artificially extend the lifetime of
local variables. Since KeyboardInterrupt protection is now imposed statically
(when a protected function is defined) rather than each time the function runs,
its previously-noticeable performance overhead should now be near zero.
The lack of a call-time wrapper has some other benefits as well:

* :func:`inspect.iscoroutinefunction` and the like now give correct answers when
  called on KI-protected functions.

* Calling a synchronous KI-protected function no longer pushes an additional stack
  frame, so tracebacks are clearer.

* A synchronous KI-protected function invoked from C code (such as a weakref
  finalizer) is now guaranteed to start executing; previously there would be a brief
  window in which KeyboardInterrupt could be raised before the protection was
  established.

One minor drawback of the new approach is that it is no longer possible to apply
different KI protection rules to different instances of the same closure. See the
documentation of `@enable_ki_protection <trio.lowlevel.enable_ki_protection>`
for more details and a workaround.
```
(Personally I wouldn't mention the weird local; it's sort of implied by mentioning that we don't access locals anymore, and it seems likely to confuse people who haven't happened to encounter it in their debuggers.)

And then I would extend the `@enable_ki_protection` documentation with something like:
```
Since KeyboardInterrupt protection is tracked per code object, any attempt to
conditionally protect the same block of code in different ways is unlikely to behave
how you expect. If you try to conditionally protect a closure, it will be
unconditionally protected instead::

    def example(protect: bool) -> bool:
        def inner() -> bool:
            return trio.lowlevel.currently_ki_protected()
        if protect:
            inner = trio.lowlevel.enable_ki_protection(inner)
        return inner()

    assert example(False) == False
    assert example(True) == True  # once protected ...
    assert example(False) == True  # ... always protected

If you really need conditional protection, you can achieve it by giving each
KI-protected instance of the closure its own code object::

    def example(protect: bool) -> bool:
        def inner() -> bool:
            return trio.lowlevel.currently_ki_protected()
        if protect:
            inner.__code__ = inner.__code__.replace()
            inner = trio.lowlevel.enable_ki_protection(inner)
        return inner()

    assert example(False) == False
    assert example(True) == True
    assert example(False) == False

(This isn't done by default because it carries some memory overhead and reduces
the potential for specializing optimizations in recent versions of CPython.)
```

A5rocks (commented):
Plenty of nitpicks or me being confused; please don't assume I know what I'm doing with these comments. If the question doesn't make sense, it's cause I don't understand and not some weird encoded question.

oremanj (approved):

src/trio/_core/_ki.py
graingert: It's already a TypeGuard but probably guarding the wrong thing

src/trio/_core/_ki.py
graingert:
```suggestion
```

src/trio/_core/_ki.py
graingert: Sorry my spelling is atrocious. Why didn't code spell pick this up?

graingert:
It looks like this will fix https://github.com/python-trio/trio/issues/1752

and maybe in conjunction with https://github.com/python-trio/trio/pull/3112 it will fix this https://github.com/python-trio/trio/issues/2454

newsfragments/2670.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+:func:`inspect.iscoroutinefunction` and the like now give correct answers when
+called on KI-protected functions.
```
A5rocks, line 2: Could we add a test for this

newsfragments/3108.bugfix.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* A synchronous KI-protected function invoked from C code (such as a weakref
+  finalizer) is now guaranteed to start executing; previously there would be a brief
+  window in which KeyboardInterrupt could be raised before the protection was
+  established.
+
+One minor drawback of the new approach is that it is no longer possible to apply
+different KI protection rules to different instances of the same closure. See the
+documentation of `@enable_ki_protection <trio.lowlevel.enable_ki_protection>`
+for more details and a workaround.
```
A5rocks, line 23: "For a workaround" doesn't work with "no longer possible" IMO.

src/trio/_core/_ki.py
```diff
@@ -83,20 +78,112 @@
...
+        if self is other:
+            return True
+
+        if not isinstance(other, _IdRef):
+            return NotImplemented
+
+        my_obj = None
+        try:
+            my_obj = self()
+            return my_obj is not None and my_obj is other()
```
A5rocks, line 103: Should `my_obj is other()` be `my_obj == other()` given this is `__eq__`?

src/trio/_core/_ki.py
```diff
@@ -83,20 +78,112 @@
...
+        finally:
+            del my_obj
```
A5rocks, line 105: I assume this is in case weakref (or `__eq__` if you take my suggestion above) raises?

src/trio/_core/_ki.py
```diff
@@ -83,20 +78,112 @@
...
+# see also: https://github.com/python/cpython/issues/88306
+class WeakKeyIdentityDictionary(Generic[_KT, _VT]):
+    def __init__(self) -> None:
+        self._data: dict[_IdRef[_KT], _VT] = {}
+
+        def remove(
+            k: _IdRef[_KT],
+            selfref: weakref.ref[
+                WeakKeyIdentityDictionary[_KT, _VT]
+            ] = weakref.ref(  # noqa: B008  # function-call-in-default-argument
```
A5rocks, line 128: Tangential: Is there a ruff setting or something where we can use long descriptive names instead of stuff like `B008`? I certainly don't like having to look them up.

src/trio/_core/_run.py
```diff
@@ -1781,10 +1774,10 @@ def current_root_task(self) -> Task | None:
     ################
 
     @_public  # Type-ignore due to use of Any here.
```
A5rocks, line 1776: This comment can be removed

src/trio/_core/_tests/test_ki.py
```diff
@@ -517,3 +523,111 @@ async def inner() -> None:
...
+
+@pytest.mark.xfail(
+    strict=True,
+    raises=AssertionError,
+    reason=(
+        "it was decided not to protect against this case, see discussion in: "
+        "https://github.com/python-trio/trio/pull/3110#discussion_r1802123644"
+    ),
+)
+async def test_ki_does_not_leak_accross_different_calls_to_inner_functions() -> None:
```
A5rocks, line 543: across not accross

src/trio/_core/_tests/test_ki.py
```diff
@@ -517,3 +523,111 @@ async def inner() -> None:
...
+    assert not _core.currently_ki_protected()
+
+    def factory(enabled: bool) -> Callable[[], bool]:
+        @_identity(_core.enable_ki_protection if enabled else _identity)
+        def decorated() -> bool:
+            return _core.currently_ki_protected()
+
+        return decorated
```
A5rocks, line 551: Why not call the conditional function on this to avoid the `_identity` dance

A5rocks (approved):
A bunch of really minor comments. This could merge as is without any changes IMO.

A5rocks:
For the sake of deprecation, should enable_ki_protection check whether a function's code object is already in the dictionary w/ a value of False, and if so (a) do the closure workaround and (b) do a deprecation warning?

Similar for disable_ki_protection

src/trio/_core/_ki.py
graingert: No because code objects can compare equal on different functions

src/trio/_core/_ki.py
graingert: Or KeyboardInterrupt

src/trio/_core/_ki.py
graingert: No there's not :(

src/trio/_core/_run.py
graingert:
```suggestion
    @_public
```

src/trio/_core/_tests/test_ki.py
graingert:
```suggestion
async def test_ki_does_not_leak_across_different_calls_to_inner_functions() -> None:
```

src/trio/_core/_tests/test_ki.py
```diff
@@ -517,3 +523,111 @@ async def inner() -> None:
...
+    reason=(
+        "it was decided not to protect against this case, see discussion in: "
+        "https://github.com/python-trio/trio/pull/3110#discussion_r1802123644"
+    ),
+)
+async def test_ki_does_not_leak_accross_different_calls_to_inner_functions() -> None:
+    assert not _core.currently_ki_protected()
+
+    def factory(enabled: bool) -> Callable[[], bool]:
+        @_identity(_core.enable_ki_protection if enabled else _identity)
```
graingert, line 547:
```suggestion
        @_core.enable_ki_protection if enabled else _identity
```

src/trio/_core/_ki.py
TeamSpen210: Not yet, but it is planned: https://github.com/astral-sh/ruff/issues/1773. Needs decisions on how it should work UI wise, etc.

src/trio/_core/_ki.py
graingert: This is also to allow the `inner.__code__ = inner.__code__.replace()` to work

src/trio/_core/_tests/test_ki.py
graingert: Arbitrary choice really. I want to use it as a decorator

But now I really want to use it, so I can use pep 614

graingert:
I'll need some help explaining this in the news file: https://github.com/python-trio/trio/pull/3110#discussion_r1798568761

newsfragments/2670.bugfix.rst
graingert: Ah I assumed this was already tested by the whole `@_public` machinery

I can work on a dedicated test for this

graingert: pushed

newsfragments/3108.bugfix.rst
oremanj: maybe: "One minor drawback of the new approach is that multiple instances of the same closure share a single KeyboardInterrupt protection state (because they share a single code object). That means that if you apply `@enable_ki_protection` to some of them and not others, you won't get the protection semantics you asked for. See [etc]"

graingert, line 23:
```suggestion
One minor drawback of the new approach is that multiple instances of the same
closure share a single KeyboardInterrupt protection state (because they share a
single code object). That means that if you apply 
`@enable_ki_protection <trio.lowlevel.enable_ki_protection>` to some of them
and not others, you won't get the protection semantics you asked for. See the
documentation of `@enable_ki_protection <trio.lowlevel.enable_ki_protection>`
for more details and a workaround.
```

graingert:
pre-commit.ci autofix

A5rocks (approved):
Just to be explicit, since all my comments were fixed and all my questions answered.

CoolCat467 (approved):
As far as I can tell, this looks pretty great

PR #3112, opened 2024-10-17
don't use f_locals for foreign async generators

graingert:
Fixes https://github.com/python-trio/trio/issues/2454


src/trio/_core/_asyncgens.py
```diff
@@ -80,8 +82,9 @@ def finalize_in_trio_context(
         def finalizer(agen: AsyncGeneratorType[object, NoReturn]) -> None:
             agen_name = name_asyncgen(agen)
             try:
-                is_ours = not agen.ag_frame.f_locals.get("@trio_foreign_asyncgen")
-            except AttributeError:  # pragma: no cover
+                self.foreign.remove(id(agen))
+                is_ours = False
+            except KeyError:
                 is_ours = True
```
graingert, line 88:
```suggestion
                self.foreign.remove(id(agen))
            except KeyError:
                is_ours = True
            else:
                is_ours = False
```

newsfragments/3112.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Rework foriegn async generator finalization to track async generator
```
graingert, line 1:
```suggestion
Rework foreign async generator finalization to track async generator
```

oremanj (approved):
Thanks, this is a nice improvement!

CoolCat467 (approved):
Looks good and makes sense to me

src/trio/_core/_tests/test_guest_mode.py
```diff
@@ -664,3 +665,51 @@ async def trio_main() -> None:
...
+        try:
+            yield 1
+        finally:
+            library = sniffio.current_async_library()
+            with contextlib.suppress(trio.Cancelled):
+                await sys.modules[library].sleep(0)
+
+            del a
+            if sys.implementation.name == "pypy":
+                gc_collect_harder()
```
A5rocks, line 692: Did this really use to fail on PyPy?

src/trio/_core/_tests/test_guest_mode.py
```diff
@@ -664,3 +665,51 @@ async def trio_main() -> None:
     context.run(aiotrio_run, trio_main, host_uses_signal_set_wakeup_fd=True)
 
     assert record == {("asyncio", "asyncio"), ("trio", "trio")}
+
+
+@restore_unraisablehook()
+def test_guest_mode_asyncgens_garbage_collection() -> None:
```
A5rocks, line 671: I had a few comments about test code but seeing as you copied the test from above, I guess I should put them in one place.

Firstly, I don't think it's necessary to check the library labels.

Then specifics (which I guess also apply to the previous test function too):
 - the 3.6 compat code is unnecessary
 - I assume the `import sniffio` can be lifted to the top level? (Probably not even necessary for this test because we shouldn't check labels)
 - is the unraisable hook even necessary? I don't see where we could be raising in a garbage collection. (IIRC that's what the decorator is for?)

A5rocks (commented):
Basically just nitpicks/comments about the test code. Nothing that should block this from being merged.

src/trio/_core/_tests/test_guest_mode.py
graingert: pypy isn't guaranteed to discard the local immediately

And gc_collect_harder never collected the local while the frame locals were materialized

src/trio/_core/_tests/test_guest_mode.py
graingert: This is disabling the default pytest unreasonable hook which would fail the test when the async generator is finalized because trio warns about it

src/trio/_core/_asyncgens.py
```diff
@@ -78,12 +80,14 @@ def finalize_in_trio_context(
...
             try:
-                is_ours = not agen.ag_frame.f_locals.get("@trio_foreign_asyncgen")
-            except AttributeError:  # pragma: no cover
+                self.foreign.remove(id(agen))
+            except KeyError:
                 is_ours = True
+            else:
+                is_ours = False
 
+            agen_name = name_asyncgen(agen)
```
graingert, line 90: I moved this call later because it's a LOAD_GLOBAL and fairly complicated bit of code and it's critical that self.foreign.remove get called

src/trio/_core/_tests/test_guest_mode.py
graingert: I'd like to keep the labels just so it's clear which generator gets finalized when

A5rocks: Yeah that's reasonable. `== {True}` isn't the most informative in telling the reader "this True happened twice, by the way".

src/trio/_core/_asyncgens.py
```diff
@@ -78,12 +80,14 @@ def finalize_in_trio_context(
                 self.trailing_needs_finalize.add(agen)
 
         def finalizer(agen: AsyncGeneratorType[object, NoReturn]) -> None:
```
graingert, line 82: I think this can only work when using https://github.com/python-trio/trio/pull/3110 because if self.foreign.remove(id(agen)) is interrupted with a KI the id will outlive the agen and could end up being re-used

src/trio/_core/_asyncgens.py
```diff
@@ -25,6 +25,32 @@
     _ASYNC_GEN_SET = set
 
 
+@_core.disable_ki_protection
```
graingert, line 28: this needs a test

src/trio/_core/_asyncgens.py
```diff
@@ -25,6 +25,32 @@
...
+) -> None:
+    # Host has no finalizer.  Reimplement the default
+    # Python behavior with no hooks installed: throw in
+    # GeneratorExit, step once, raise RuntimeError if
+    # it doesn't exit.
+    closer = agen.aclose()
+    try:
+        # If the next thing is a yield, this will raise RuntimeError
+        # which we allow to propagate
+        closer.send(None)
```
graingert, line 41: actually what do we want to happen here if there's a KI? We want long running finalizers to be interruptible, but then the KeyboardInterrupt exception just gets lost so you need to hit Ctrl+C again to terminate trio

A5rocks: Where does the KI get lost?

If it's in this `closer` after we call `.send(None)`, doesn't it get raised here and will propagate up? I assume propagation here is fine because we already have a `RuntimeError` being raised here.

graingert: The finalizer gets called by tp_finalize which is called by the GC which catches the exception

A5rocks: I see... could we catch the `KeyboardInterrupt` and basically tell trio that it got another Ctrl+C? It seems to be raised normally:

```py
>>> trio.run(main)
Exception ignored in: <async_generator object awaits_after_yield at 0x000001EF4C0F1F00>
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_asyncgens.py", line 120, in finalizer
    closer.send(None)
  File "<stdin>", line 6, in awaits_after_yield
KeyboardInterrupt:
```

(I also got this great error message in the REPL while playing around with this...)

<details><summary>Funny error message I ran into in the REPL</summary>
<p>

```
Exception ignored in: <function Nursery.__del__ at 0x000001EF4BCC54E0>
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1372, in __del__
    assert not self._children
           ^^^^^^^^^^^^^^^^^^
AssertionError:
Exception ignored in: <coroutine object Runner.init at 0x000001EF4C0239A0>
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 2030, in init
    async with open_nursery() as run_sync_soon_nursery:
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1030, in __aexit__
    new_exc = await self._nursery._nested_child_finished(exc)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1192, in _nested_child_finished
    self._add_exc(nested_child_exc)
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1170, in _add_exc
    self.cancel_scope.cancel()
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_ki.py", line 183, in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 870, in cancel
    self._cancel_status.recalculate()
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 459, in recalculate
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1545, in _attempt_delivery_of_any_pending_cancel
    self._attempt_abort(raise_cancel)
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1527, in _attempt_abort
    success = self._abort_func(raise_cancel)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_io_windows.py", line 740, in abort_fn
    self._refresh_afd(base_handle)
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_io_windows.py", line 653, in _refresh_afd
    _check(
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_io_windows.py", line 301, in _check
    raise_winerror()
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_windows_cffi.py", line 520, in raise_winerror
    raise OSError(0, msg, filename, winerror, filename2)
OSError: [WinError 6] The handle is invalid
Exception ignored in: <function Nursery.__del__ at 0x000001EF4BCC54E0>
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1372, in __del__
    assert not self._children
           ^^^^^^^^^^^^^^^^^^
AssertionError:
Exception ignored in: <function Nursery.__del__ at 0x000001EF4BCC54E0>
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1372, in __del__
    assert not self._children
           ^^^^^^^^^^^^^^^^^^
AssertionError:
Exception ignored in: <async_generator object awaits_after_yield at 0x000001EF4C0F0040>
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_asyncgens.py", line 88, in finalizer
    runner.entry_queue.run_sync_soon(
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_entry_queue.py", line 145, in run_sync_soon
    self.wakeup.wakeup_thread_and_signal_safe()
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_wakeup_socketpair.py", line 39, in wakeup_thread_and_signal_safe
    self.write_sock.send(b"\x00")
OSError: [WinError 10038] An operation was attempted on something that is not a socket
Exception ignored in: <coroutine object NurseryManager.__aexit__ at 0x000001EF4C01B120>
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1030, in __aexit__
    new_exc = await self._nursery._nested_child_finished(exc)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 1220, in _nested_child_finished
    assert popped is self
           ^^^^^^^^^^^^^^
AssertionError:
```

Not the most clear :^)

</p>
</details>

graingert: > It seems to be raised normally

But the message says the exception is ignored?

A5rocks: Sorry, raised normally within the finalizer which is the context here. `_finalize_without_ki_protection` could add a

```py
except KeyboardInterrupt:
  # do something to a passed in `self`, maybe? to tell trio a KI occured
  # then ignore the KI.
```
right below its `except StopIteration:`

Looking more at the code, it doesn't seem like we have the proper state to reset though.

graingert: Also it would be needed for every finalizer

A5rocks: Maybe we could call `signal.raise_signal(signal.SIGINT)` in the `except KeyboardInterrupt:` and call it a day? Not a great solution IMO.

graingert: Hmm. The except KeyboardInterrupt would have to be outside the _finalize_without_ki_protection

A5rocks: How about this awful hack?:

```py
except KeyboardInterrupt:
  @enable_ki_protection
  def killer():
    signal.raise_signal(signal.SIGINT)
  try:
    killer()
  except KeyboardInterrupt:
    # turns out the trio KI handler isn't installed. Nothing good to do
    raise KeyboardInterrupt from None
```

graingert: I think what I'm going to do here is put the KI protection on all the critical and guaranteed fast trio code and unprotect any of the host loop stuff. Then we can survey someone who uses guest mode and KI and async generators without aclose and find out what they want

graingert: > How about this awful hack?:
> 
> ```python
> except KeyboardInterrupt:
>   @enable_ki_protection
>   def killer():
>     signal.raise_signal(signal.SIGINT)
>   try:
>     killer()
>   except KeyboardInterrupt:
>     # turns out the trio KI handler isn't installed. Nothing good to do
>     raise KeyboardInterrupt from None
> ```

Another KI could arrive while constructing or decorating or load_fast-ing the killer function

A5rocks:
Oh the pypy cache failures are back :( 

I tried deleting them from the GH actions cache page last time this happened but I guess they came back. That still works to fix this though.

PR #3114, opened 2024-10-19
Give Pyright what it wants (`alias` attributes everywhere)

A5rocks:
I don't like pyright nor this behavior, but that doesn't mean users should have to care. We can add a test to prevent any sort of regression regarding this, forever.

cc @mikenerone cause this is from Gitter.

also @CoolCat467 I ran into issues with running the `regenerate-files` pre-commit hook and couldn't figure them out after about 15 minutes. It was complaining about `attrs` not being possible to import but unfortunately there's no way to get it to use the `test-requirements.txt` file. Have you run into this issue regarding it not picking up the virtual environment? (I'm using a seperate git client so that kind of makes sense, but now's the first time I ran into this...)


src/trio/_tests/test_exports.py
```diff
@@ -572,3 +573,72 @@ def test_classes_are_final() -> None:
...
+                        assert file is None
+                        file = contents
+                        start = i - 1
+
+                    if token.type == tokenize.NAME and token.string == "class":
+                        last_was_class = True
+                    else:
+                        last_was_class = False
+
+            assert file is not None, f"{name}: {class_!r}"
```
TeamSpen210, line 614: I think you might also want to `assert start is not None`, so the below typecheck succeeds.

jakkdl:
this makes me want to write a linter rule, but also feels ridiculous (and idk which flake8 plugin it would fit) to write an attrs+pyright-specific rule.

How slow/fast is the test? If it's anything but trivial I could rewrite it as an ast visitor.

A5rocks:
> this makes me want to write a linter rule, but also feels ridiculous (and idk which flake8 plugin it would fit) to write an attrs+pyright-specific rule.
> 
> 
> 
> How slow/fast is the test? If it's anything but trivial I could rewrite it as an ast visitor.

Yeah lint rule would make more sense but unfortunately a lint rule would miss context that's helpful (like whether a class is exported).

It takes 1.75 seconds to run locally, though that's with the overhead of starting pytest.

CoolCat467 (approved):
Alias changes make sense and test looks like it would work pretty well but didn't look into the details super closely, but general idea of it seems solid.

A5rocks:
> If test is slow, do we need to mark it with `@slow` or something?

Maybe? Here's the output at the end of `pytest --durations=10` locally:

```
================================================ slowest 10 durations =================================================
3.66s call     src/trio/_tests/tools/test_gen_exports.py::test_process[from collections import Counter\n]
2.73s call     src/trio/_tests/tools/test_gen_exports.py::test_process[from collections import Counter\nimport os\n]
2.53s call     src/trio/_tests/tools/test_gen_exports.py::test_process[from typing import TYPE_CHECKING\nif TYPE_CHECKING:\n    from collections import Counter\n]
2.11s call     src/trio/_tests/test_socket.py::test_many_sockets
2.05s call     src/trio/_tests/test_socket.py::test_SocketType_connect_paths
2.04s call     src/trio/_tests/test_socket.py::test_address_in_socket_error
2.02s call     src/trio/_tests/test_highlevel_open_tcp_listeners.py::test_open_tcp_listeners_ipv6_v6only
1.73s call     src/trio/_tests/test_exports.py::test_pyright_recognizes_init_attributes
0.65s call     src/trio/_tests/test_exports.py::test_static_tool_sees_all_symbols[jedi-trio]
0.41s call     src/trio/_tests/tools/test_gen_exports.py::test_lint_failure
```

I don't think it's an appreciable slowdown. Maybe it's a good idea to put `@slow` on everything over 1 second?

src/trio/_tests/test_exports.py
```diff
@@ -572,3 +573,77 @@ def test_classes_are_final() -> None:
...
+            assert start is not None
+
+            count = -1
+            # linters don't like my not using the index, go figure.
+            for end_offset, token in enumerate(file[start:]):  # noqa: B007
```
jakkdl, line 623: if you declare `end_offset` outside the loop I think the warning goes away (and makes the code easier to read).

codecov wants a `pragma: no branch` on the loop to hit 100%

jakkdl (changes requested):
maybe also good to add a comment or two on what this token thing is doing, I find it hard to understand at a glance what the code is doing.

src/trio/_tests/test_exports.py
A5rocks: Unfortunately the warning doesn't go away; it does make the code more correct (previously it didn't handle `start==len(file)-1`, I think)

CoolCat467: I think with the zero-initialization now that should be good

TeamSpen210:
There’s another approach we could use, but it’d might require starting an additional interpreter. Run a script which monkeypatches `attrs.field` to store whether alias was set into the metadata dict, import trio, then we can just loop through fields. 

Unsure if that’d be faster though. We could avoid the new interpreter if the monkeypatch was done before any tests import trio, but that means it affects all tests. Probably not too much of an issue since metadata does nothing, and just wrapping the function should be fairly safe.

A5rocks:
> There’s another approach we could use, but it’d might require starting an additional interpreter. Run a script which monkeypatches `attrs.field` to store whether alias was set into the metadata dict, import trio, then we can just loop through fields.
> 
> Unsure if that’d be faster though. We could avoid the new interpreter if the monkeypatch was done before any tests import trio, but that means it affects all tests. Probably not too much of an issue since metadata does nothing, and just wrapping the function should be fairly safe.

It's not significantly faster:

```
(.venv) PS C:\Users\A5rocks\Documents\trio> hyperfine 'python -c \"import trio\"'
Benchmark 1: python -c "import trio"
  Time (mean ± σ):     410.5 ms ±  14.2 ms    [User: 45.3 ms, System: 22.9 ms]
  Range (min … max):   390.3 ms … 435.0 ms    10 runs
```

I think the main improvement would be that it's slightly more reliable. It sounds annoying to implement though.

src/trio/_tests/test_exports.py
jakkdl: This is fairly gnarly to fix in an AST visitor, I don't see any way of doing it that's not overkill, but I think you could open an issue for ruff and see if they have a decent way of getting at it.

jakkdl (approved):

A5rocks:
@TeamSpen210 do you have any implementation ideas for your idea?

I was thinking about it just for the gain in reliability, but I think we would then get a flakey test because we delete and reimport trio to get all the warnings. I guess it's fine if we just patch attrs cause then the reimport will still be using the patched attrs?

TeamSpen210:
Pushed [an implementation](https://github.com/TeamSpen210/trio/commit/f21cbb01567fcb36bc9e29c8be033394408cc4c9) using monkeypatching here. Unfortunately a downside I realised is that the monkeypatching code has to be either outside the `trio` package, or directly in `__init__` before any other imports. So either this test can't be run by people just installing the package, or we'll have to have a module outside there. I could set it up to just skip if the monkeypatch "plugin" isn't specified though.

Zac-HD:
I do like the "monkeypatch, and skip that test if not monkeypatched" plan 🙂

Zac-HD:
@TeamSpen210 @A5rocks I'd really like to get this merged so I can use it; is there anything I can do to help?

A5rocks:
> @TeamSpen210 @A5rocks I'd really like to get this merged so I can use it; is there anything I can do to help?

Nope! I've just been pushing off working on some trio stuff for a bit, I'll incorporate better testing strategy soon:tm: (probably just using our already created pytest plugin + copy.copy() of the kwargs instead of picking and choosing. Everything else sounds fine)

TeamSpen210:
Would you like me to make the monkeypatch approach a competing PR, or PR to yours, or do you want to stick with the current approach?

A5rocks:
If you wouldn't mind, you should be able to commit to this PR.

TeamSpen210:
Committed, didn't want to do big changes to other people's PRs without permission. Though it is easy to revert so probably that's fine.

src/_trio_check_attrs_aliases.py
```diff
@@ -0,0 +1,22 @@
...
+
+import attrs
+
+orig_field = attrs.field
+
+
+def field(**kwargs: Any) -> Any:
+    original_args = kwargs.copy()
+    metadata = kwargs.setdefault("metadata", {})
+    metadata["original_args"] = original_args
```
TeamSpen210, line 16: Probably should add `trio` to this key somewhere, since it's possible one of our dependencies might use `original_args` in attrs metadata, or whatever. Not sure why you're storing all the args, if we only use `alias`? Future-proofing?

A5rocks: Yeah, for any sort of later attrs test

A5rocks:
Should we move the plugin to `tests` or something? Having it in `src` feels weird...

src/_trio_check_attrs_aliases.py
CoolCat467: Would suggest using `object` if possible

TeamSpen210:
That should work, if it's in `sys.path` so pytest can import.

A5rocks:
> That should work, if it's in `sys.path` so pytest can import.

Nevermind! I didn't realize pytest's sys.path schenanigans are the way they are. I'm surprised the test passes in CI! Maybe you can explain to me? I'm obviously missing something, because my mental model is:
 - this works locally because pytest adds `src/` to sys.path. then pytest looks there for the plugin
 - but on ci... how? when? where? why would it add `src/` to sys.path?

TeamSpen210:
Not really sure either, but it's working so don't touch it?

A5rocks:
Ohhhhh nooooo, it's installing `_trio_check_attrs_aliases.py` as a module into `site-packages` as part of the single wheel we make. That's quite the footgun!!!!

As a knee-jerk reaction I think we should revert back to before using a plugin. Maybe you see a way to work around this? Probably setting `PYTHONPATH=...`...

TeamSpen210:
We could set `PYTHONPATH` I think yes? I was going to do that but didn't trust myself with changing the complicated shell script. Probably should put the plugin in its own folder to isolate. I don't think it matters that the test won't run other than in CI, since it's only going to fail if the source code changes, not due to environment etc.

Another approach would be to use [`usercustomize`](https://docs.python.org/3/library/site.html#module-usercustomize) perhaps, copying the module into `site-packages` ourselves.

TeamSpen210:
The plugin script can't be anywhere in the `trio` package, because then trying to import it implicitly imports `trio` first, before we hooked anything. That's why I gave it such a long name.

A5rocks:
> The plugin script can't be anywhere in the `trio` package, because then trying to import it implicitly imports `trio` first, before we hooked anything. That's why I gave it such a long name.

Yeah the issue with that commit was that we had `-p trio._tests.pytest_plugin` in addopts, which then messes things up because turns out pytest runs plugins in the order they're provided. Anyways we don't need that anymore, so I removed it.

mikenerone (approved):

PR #3115, opened 2024-10-19
Document that `trio.sleep_forever` will raise `RuntimeError` instead of returning

### LINKED ISSUE #3113: trio.sleep_forever should mention that it raises if rescheduled
> This causes a regression for anyone who was using `trio.sleep_forever` combined with `trio.lowlevel.reschedule`. Okay, maybe they (we (I)) should have been using `trio.lowlevel.wait_task_rescheduled` but the `RuntimeError` should have been phased in with a warning over one release cycle.

_Originally posted by @richardsheridan in https://github.com/python-trio/trio/pull/3096#issuecomment-2422986247_

A5rocks:
Fixes #3113.

### DIFF
--- a/newsfragments/3113.doc.rst
+++ b/newsfragments/3113.doc.rst
@@ -1,0 +1,1 @@
+Document that :func:`trio.sleep_forever` is guaranteed to raise an exception now.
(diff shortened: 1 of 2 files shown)


TeamSpen210:
I don't think this should say "instead of returning", it won't do so in normal usage, it'll only return if cancelled. Better to say that it never returns, and raises `RuntimeError` if manually rescheduled.

CoolCat467 (approved):

richardsheridan (approved):

PR #3117, opened 2024-10-20
remove notes-to-self

jakkdl:
I don't think these are used by anybody in any way anymore? They mostly just seem to mess with linters/formatters and increase the size of the codebase by 200KB.
If we do want to keep them around for more visibility somewhere, and people actually expect to use and/or update them once in a blue mood, we could consider putting them in a separate repo in the python-trio org?

[teamspen](https://github.com/python-trio/trio/pull/3098#issuecomment-2419106974) suggested making a tag. We could make a dedicated notes-to-self tag, but for now I just reused the `0.27.0` tag.


jakkdl:
does anybody have any opinions? @A5rocks @njsmith @Zac-HD @oremanj @graingert

graingert:
You could git commit them into the wiki https://github.com/python-trio/trio/wiki

jakkdl:
> You could git commit them into the wiki https://github.com/python-trio/trio/wiki

that's a good idea, done: https://github.com/python-trio/trio/wiki/notes-to-self

graingert:
> > You could git commit them into the wiki [python-trio/trio/wiki](https://github.com/python-trio/trio/wiki)
> 
> that's a good idea, done: [Wiki: notes to self](https://github.com/python-trio/trio/wiki/notes-to-self)

Ah I meant to include the history also, eg with filter branch

graingert:
> > > You could git commit them into the wiki [python-trio/trio/wiki](https://github.com/python-trio/trio/wiki)
> > 
> > 
> > that's a good idea, done: [Wiki: notes to self](https://github.com/python-trio/trio/wiki/notes-to-self)
> 
> Ah I meant to include the history also, eg with filter branch

Resolved to link to the merged commit back from the wiki in DM

graingert (approved):

jakkdl:
> > > You could git commit them into the wiki [python-trio/trio/wiki](https://github.com/python-trio/trio/wiki)
> > 
> > 
> > that's a good idea, done: [Wiki: notes to self](https://github.com/python-trio/trio/wiki/notes-to-self)
> 
> Ah I meant to include the history also, eg with filter branch

I think this would only be necessary/useful if we completely wiped the notes-to-self directory from the repository (which I don't think is necessary, while it's 7% of current files it's going to have less impact on the the size of the history since the files have rarely been changed). But I'll add a link to the wiki with the merge commit in case people do want to `git blame`

A5rocks (commented):
Could you also add `notes-to-self/` to a `.gitignore`? This pattern still might be nice to test any new features while adding them.

jakkdl:
> Could you also add `notes-to-self/` to a `.gitignore`? This pattern still might be nice to test any new features while adding them.

Done. While I think it's unlikely somebody would create a directory with that exact name to play around with stuff, having people restore the files from history / download them from the wiki seems like a decent use-case.

I'll merge this in a bit (and update wiki to refer to the PR) unless anybody has any objections. :)

A5rocks (approved):

PR #3120, opened 2024-10-26
Avoid refcycles in run exc

graingert:


src/trio/_core/_run.py
```diff
@@ -1706,6 +1706,7 @@ def close(self) -> None:
         self.asyncgens.close()
         if "after_run" in self.instruments:
             self.instruments.call("after_run")
+        self.system_nursery: Nursery | None = None
```
A5rocks, line 1709: Type hint here isn't necessary

demo.py
A5rocks, line 1: I think you accidentally left this instead of copying parts over as a test.

graingert: Yep, sorry I pushed this in draft so I don't lose it

graingert:
I'm currently trying to work out if this change makes it possible to test https://github.com/python-trio/trio/blob/12e8395cf6939ec9f00df122487dc66ef774fec8/src/trio/_core/_run.py#L2790

src/trio/_core/_run.py
CoolCat467: As long as this isn't the first time `system_nursery` is set yes

src/trio/_core/_run.py
```diff
@@ -1726,6 +1726,7 @@ def close(self) -> None:
...
```
graingert, line 1729:
```suggestion
        self.system_nursery = None
```

PR #3123, opened 2024-10-28
Enable pre-commit.ci pull-request autofixing

CoolCat467:
In this pull request, we allow pre-commit.ci to autofix pull requests. The specific pre-commit hooks we use are all designed well and will avoid making changes that might cause issues, and people usually end up saying the trigger message to fix their pull requests anyways.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,5 @@
 ci:
-  autofix_commit_msg: "[pre-commit.ci] auto fixes from pre-commit.com hooks"
-  autofix_prs: false
-  autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
+  autofix_prs: true
   autoupdate_schedule: weekly
   submodules: false
   skip: [regenerate-files]


graingert (approved):

A5rocks:
I think the main concern with this is having to remember to `git pull` whenever this happens, whereas someone could trivially fix it locally...

But to be honest I'm not as sold on that line of thought as I once was (for one, anyone committing things from a machine could have installed pre-commit to begin with(. Now I'm neutral on this, rather than negative.

The main issue IMO w/ our pre-commit setup in general is that our autogenerated files hook is not great, cause it relies on having `astor` etc installed. But that's irrelevant to this as a PR.

jakkdl:
> I think the main concern with this is having to remember to `git pull` whenever this happens, whereas someone could trivially fix it locally...

even if you do forget to pull you can usually just rebase and everything gets resolved.

jakkdl (approved):

.pre-commit-config.yaml
```diff
@@ -1,7 +1,5 @@
 ci:
-  autofix_commit_msg: "[pre-commit.ci] auto fixes from pre-commit.com hooks"
-  autofix_prs: false
-  autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
+  autofix_prs: true
```
jakkdl, line 2: I'm guessing these are just defaults you're removing?

CoolCat467: Yes

PR #3126, opened 2024-11-01
Get rid of `signal_raise`

A5rocks:
I noticed `signal.raise_signal` seems to do what `signal_raise` used to. Let's use that instead.

### DIFF
--- a/src/trio/_core/_tests/test_guest_mode.py
+++ b/src/trio/_core/_tests/test_guest_mode.py
@@ -28,7 +28,6 @@ import trio
 import trio.testing
 from trio.abc import Instrument
 
-from ..._util import signal_raise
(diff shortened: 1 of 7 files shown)


CoolCat467 (approved):
Looks good as far as I can tell, but wondering, why didn't we just use `signal.raise_signal` to begin with? Did a bug get fixed or something?

Looking into it lightly, looks like `signal.raise_signal` was added in 3.8, so maybe 3.7 support reasons?

A5rocks:
I assume the code was written pre-3.8

jakkdl:
> I assume the code was written pre-3.8

yep, 2017 - https://github.com/python-trio/trio/commit/300a19ddb3fc75aff4fe7e7b37957b655e54b995
3.7 hadn't even been released!

good find :+1:

PR #3127, opened 2024-11-02
 install pip and uv in separate pip calls 

graingert:
[investigating pypy pip cache issues]


A5rocks:
Oh I think someone deleted the cache or something. Things work now. Here's a list of what I was going to try (to narrow down things):
 - check the mtimes of the cached files
 - check whether the metadata file seems to match the body file

I'll try them if this happens again.

A5rocks:
Nevermind I got fooled. Somehow it's failing elsewhere -- does it take the current actions YAML as input??? Weird.

A5rocks:
Uh huh:
```
-rw-r--r-- 1 runneradmin 197121   1819 Nov  9 06:29 c:/users/runneradmin/appdata/local/pip/cache/http-v2/a/d/e/e/f/adeef5b9611702687bbddade5f7c0af65110caf98ba1a0070c9d116e
-rw-r--r-- 1 runneradmin 197121 228583 Oct 28 14:08 c:/users/runneradmin/appdata/local/pip/cache/http-v2/a/d/e/e/f/adeef5b9611702687bbddade5f7c0af65110caf98ba1a0070c9d116e.body
-rw-r--r-- 1 runneradmin 197121 238531 Nov  9 06:29 c:/users/runneradmin/appdata/local/pip/cache/http-v2/a/d/e/e/f/adeef5b9611702687bbddade5f7c0af65110caf98ba1a0070c9d116e.body4f_dkhwx.tmp
-rw-r--r-- 1 runneradmin 197121 232580 Nov  1 01:08 c:/users/runneradmin/appdata/local/pip/cache/http-v2/a/d/e/e/f/adeef5b9611702687bbddade5f7c0af65110caf98ba1a0070c9d116e.body9k42hbh3.tmp
-rw-r--r-- 1 runneradmin 197121 234523 Nov  5 22:53 c:/users/runneradmin/appdata/local/pip/cache/http-v2/a/d/e/e/f/adeef5b9611702687bbddade5f7c0af65110caf98ba1a0070c9d116e.bodyd4lemxfj.tmp
```
that seems ominous

A5rocks:
OK so my hypothesis is that:
 - in the normal case, pip takes the metadata (which is up to date), sends the request based on it (which I assume includes stuff like if-modified-since or whatever), and then pypi is like "this hasn't been modified since then". pip then takes the cached body (which isn't up to date) and then uses that.
 - in the updating case, somehow something prevents pip from writing to the body file?

A5rocks:
> in the updating case, somehow something prevents pip from writing to the body file?
this doesn't make sense. it's not permissions, yet it happened 3 times in a row. I'll try to find the actions run where it happened (based on timestamps)

CoolCat467:
Could this theoretically be issues stemming from a hash collision? I know the likelyhood of that is astronomically low, but putting that idea out there.

A5rocks:
> Could this theoretically be issues stemming from a hash collision? I know the likelyhood of that is astronomically low, but putting that idea out there.

I don't think so. But not having suppressed the OS errors would help diagnosing things :^)

A5rocks:
Just realized https://github.com/python-trio/trio/actions/runs/11754054520/job/32747608845 is another actions run that failed to rename the tmpfile pip makes, and it has the verbose flags. But the logs look like you would expect:

```
  Updating cache with response from "https://pypi.org/simple/uv/"
  etag object cached for 1209600 seconds
  Caching due to etag
```

A5rocks:
Wow, its reproducible. That's so weird.

A5rocks:
OK so it is ultimately an OSError. Why?

```
ERROR: Could not install packages due to an OSError.
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_vendor\urllib3\response.py", line 438, in _error_catcher
    yield
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_vendor\urllib3\response.py", line 561, in read
    data = self._fp_read(amt) if not fp_closed else b""
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_vendor\urllib3\response.py", line 527, in _fp_read
    return self._fp.read(amt) if amt is not None else self._fp.read()
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_vendor\cachecontrol\filewrapper.py", line 104, in read
    self._close()
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_vendor\cachecontrol\filewrapper.py", line 84, in _close
    self.__callback(result)
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_vendor\cachecontrol\controller.py", line 408, in cache_response
    self._cache_set(cache_url, request, response, body, expires_time)
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_vendor\cachecontrol\controller.py", line 314, in _cache_set
    self.cache.set_body(cache_url, body)
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_internal\network\cache.py", line 106, in set_body
    self._write(path, body)
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_internal\network\cache.py", line 80, in _write
    replace(f.name, path)
  File "C:\hostedtoolcache\windows\PyPy\3.10.14\x86\lib\site-packages\pip\_internal\utils\retry.py", line 34, in retry_wrapped
    return func(*args, **kwargs)
PermissionError: [WinError 5] Access is denied: 'c:\\users\\runneradmin\\appdata\\local\\pip\\cache\\http-v2\\a\\d\\e\\e\\f\\adeef5b9611702687bbddade5f7c0af65110caf98ba1a0070c9d116e.bodyb8uly_ze.tmp' -> 'c:\\users\\runneradmin\\appdata\\local\\pip\\cache\\http-v2\\a\\d\\e\\e\\f\\adeef5b9611702687bbddade5f7c0af65110caf98ba1a0070c9d116e.body'
```

A5rocks:
I'm like 90% certain (I'm trying to prove this to myself right now) that the issue is that somewhere in `pip` they `open` the body file. Then they don't close it. Then this is fine on CPython because that gets deleted when it falls out of scope, but on PyPy the opened file is still open when pip tries to rename the tmpfile to go over it.

A5rocks:
I'm certain of my hypothesis: putting a `gc.collect()` before the rename fixes everything. I'm just unsure where the body file is leaking from (I mean I know it comes from a method called `get_body()` and then from `_load_from_cache()` but IDK where the bodyfile gets stored). I tried deleting the thing that holds it where I think it's responsible but that didn't help.

A5rocks:
Alright, so the "fix" specifically in this one case is that for <https://github.com/pypa/pip/blob/420435903ff2fc694d6950a47b896427ecaed78f/src/pip/_vendor/cachecontrol/controller.py#L290>, instead of just returning the headers, the file stored in `resp` needs to be closed.

For some reason, `del resp` doesn't work -- but `resp._fp.close()` works. This prevents the OSError.

graingert:
you should be able to use `with resp: ...`

A5rocks:
> you should be able to use `with resp: ...`

Seems to work! Other than the fact that `resp` is potentially None so (in actual code) need to check it first.

A5rocks:
I think we figured this out.

PR #3131, opened 2024-11-07
install uv with uv

graingert:
also avoid pip entirely in ci

this works around the issue where pip could not install uv on pypy on windows with a broken cache

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -37,13 +37,14 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
 echo "::endgroup::"
 
 echo "::group::Install dependencies"
-python -m pip install -U pip uv -c test-requirements.txt
-python -m pip --version
+curl-harder https://astral.sh/uv/0.4.30/install.sh -o uv-install.sh
+UV_UNMANAGED_INSTALL="./.uv-bin" sh uv-install.sh
+./.uv-bin/uv pip install uv -c test-requirements.txt
+python -m uv pip uninstall pip
+python -m uv pip install 'build[uv]' -c test-requirements.txt
 python -m uv --version
 
-python -m uv pip install build
-
-python -m build
+python -m build --installer=uv
 wheel_package=$(ls dist/*.whl)
 python -m uv pip install "trio @ $wheel_package" -c test-requirements.txt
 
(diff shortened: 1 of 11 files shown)


ci.sh
```diff
@@ -37,13 +37,15 @@ python -c "import sys, struct, ssl; print('python:', sys.version); print('versio
 echo "::endgroup::"
 
 echo "::group::Install dependencies"
-python -m pip install -U pip uv -c test-requirements.txt
-python -m pip --version
+curl-harder https://astral.sh/uv/0.4.30/install.sh -o uv-install.sh
```
CoolCat467, line 40: If we would do this, I would suggest grepping for uv version from test-requirements.txt with `grep 'uv=='` and then cut by '==' to get version, otherwise autodeps will never be actually changing the version of `uv` used.

CoolCat467 (changes requested):

ci.sh
graingert: Yeah I thought about that

CoolCat467:
Also, for resolving windows virtual environment issues, I also had that problem while working on #2957, and something like this should help:
https://github.com/CoolCat467/trio/blob/a359dacc2bbaaf24586771d783e3c19b99c30dab/ci.sh#L51

graingert:
> Also, for resolving windows virtual environment issues, I also had that problem while working on #2957, and something like this should help: https://github.com/CoolCat467/trio/blob/a359dacc2bbaaf24586771d783e3c19b99c30dab/ci.sh#L51 Looking at that, the changes with `PYTHON_PATH` are not necessary here, not handling pre-installed virtual environment case.

I plan to pull the Script Vs bin out of sysconfig.get_path

ci.sh
graingert: I've done this now

newsfragments/3131.misc.rst
```diff
@@ -0,0 +1 @@
+Bootstrap uv using uv install script instead of pip
```
graingert, line 1:
```suggestion
Bootstrap uv using uv install script instead of pip, this works around an issue where pip refused to install uv on PyPy on windows
```

A5rocks:
Sorry, I'm still not sold on this as a solution. I think it would be significantly better to just avoid setup-python's cache on PyPy+Windows runs.

Issue #3133, opened 2024-11-12
`AttributeError: module 'socket' has no attribute 'AI_NUMERICSERV'. Did you mean: 'NI_NUMERICSERV'?`


barracuda156:
`AI_NUMERICSERV` may not be defined on a given platform. However it is used unconditionally, which fails in that case.

Example of the problem:
```
36-63% /opt/local/Library/Frameworks/Python.framework/Versions/3.12/bin/yt
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/bin/yt", line 5, in <module>
    from mps_youtube import main
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/mps_youtube/main.py", line 36, in <module>
    completer = util.CommandCompleter()
                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/mps_youtube/util.py", line 586, in __init__
    from . import config
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/mps_youtube/config.py", line 12, in <module>
    import pylast
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/pylast/__init__.py", line 38, in <module>
    import httpx
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/httpx/__init__.py", line 2, in <module>
    from ._api import *
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/httpx/_api.py", line 6, in <module>
    from ._client import Client
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/httpx/_client.py", line 30, in <module>
    from ._transports.asgi import ASGITransport
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/httpx/_transports/__init__.py", line 3, in <module>
    from .default import *
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/httpx/_transports/default.py", line 33, in <module>
    import httpcore
[...]

PR #3135, opened 2024-11-12
_socket.py: fix for systems where AI_NUMERICSERV is not defined

barracuda156:
Rationale: on some platforms `AI_NUMERICSERV` may not be defined. In particular, old versions of macOS, possibly some other. While affected users are probably few, it is a trivial fix, which makes `trio` usable (for example, `yewtube` app works then).

Credits to @CoolCat467
Closes: https://github.com/python-trio/trio/issues/3133

### DIFF
--- a/newsfragments/3133.bugfix.rst
+++ b/newsfragments/3133.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed socket module for some older systems which lack ``socket.AI_NUMERICSERV``.
+Now trio works on legacy (pre-Lion) macOS.
--- a/src/trio/_socket.py
+++ b/src/trio/_socket.py
@@ -164,7 +164,10 @@ def set_custom_socket_factory(
 # getaddrinfo and friends
 ################################################################
 
-_NUMERIC_ONLY = _stdlib_socket.AI_NUMERICHOST | _stdlib_socket.AI_NUMERICSERV
+# AI_NUMERICSERV may be missing on some older platforms, so use it when available.
+# See: https://github.com/python-trio/trio/issues/3133
+_NUMERIC_ONLY = _stdlib_socket.AI_NUMERICHOST
+_NUMERIC_ONLY |= getattr(_stdlib_socket, "AI_NUMERICSERV", 0)
 
 
 # It would be possible to @overload the return value depending on Literal[AddressFamily.INET/6], but should probably be added in typeshed first


CoolCat467 (changes requested):
Looks good, but needs newsfragment (that is why continuous integration is failing). `newsfragments` folder's README file has more information.

barracuda156:
@CoolCat467 That one CI which fails is unrelated, right?

A5rocks:
> @CoolCat467 That one CI which fails is unrelated, right?

Docs are failing because your newsfragment is using markdown not rST. Try doing `socket.AI_NUMERICSERV` instead of `AI_NUMERICSERV` (for the automatic linking) or doubling the backticks.

src/trio/_socket.py
```diff
@@ -169,7 +169,8 @@ def set_custom_socket_factory(
 # getaddrinfo and friends
 ################################################################
 
-_NUMERIC_ONLY = _stdlib_socket.AI_NUMERICHOST | _stdlib_socket.AI_NUMERICSERV
+_NUMERIC_ONLY = _stdlib_socket.AI_NUMERICHOST
+_NUMERIC_ONLY |= getattr(_stdlib_socket, "AI_NUMERICSERV", 0)
```
jakkdl, line 173: perhaps good to add a comment explaining why this is needed / when the workaround can be removed

barracuda156: @CoolCat467 It does not exist on any macOS on any arch before 10.6 release. In practice, this will mostly be 10.5 and 10.4, on ppc and i386.

src/trio/_socket.py
```diff
@@ -169,7 +169,9 @@ def set_custom_socket_factory(
 # getaddrinfo and friends
 ################################################################
 
-_NUMERIC_ONLY = _stdlib_socket.AI_NUMERICHOST | _stdlib_socket.AI_NUMERICSERV
+# AI_NUMERICSERV may be missing on some older platforms, so use it when available.
+_NUMERIC_ONLY = _stdlib_socket.AI_NUMERICHOST
+_NUMERIC_ONLY |= getattr(_stdlib_socket, "AI_NUMERICSERV", 0)
```
jakkdl, line 174: I think you need to be a bit more specific here, otherwise future people that stumble on this comment will have to do a bunch of research to figure out why/when it might be missing. And as I said in the issue, perhaps even a note on why it's needed despite being deprecated since a decade+

A5rocks: Maybe just linking the issue will work? That has info.

A5rocks: @barracuda156 could you fix this? I think this is the only problem before merging.

A5rocks:
Bump @barracuda156

src/trio/_socket.py
barracuda156: Sorry, this got missed.
I will return to this tomorrow.

barracuda156: Added a link now to the issue into the comment.

A5rocks (approved):

CoolCat467 (approved):
With newsfragment and link to discussion, I think this should be good to go.

newsfragments/3133.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed socket module for some older systems which lack `socket.AI_NUMERICSERV`.
```
A5rocks, line 1: Oops CI is still complaining about being unable to find `socket.AI_NUMERICSERV`. Try this instead:
```suggestion
Fixed socket module for some older systems which lack ``socket.AI_NUMERICSERV``.
```

PR #3136, opened 2024-11-16
Don't create IPv6 socket if IPv6 is disabled

jackwilsdon:
`test_socket.py:test_sniff_sockopts` creates an IPv6 socket even if IPv6 is disabled. This PR changes the test to only create the socket if IPv6 is available.

### DIFF
--- a/src/trio/_tests/test_socket.py
+++ b/src/trio/_tests/test_socket.py
@@ -14,7 +14,7 @@ import attrs
 import pytest
 
 from .. import _core, socket as tsocket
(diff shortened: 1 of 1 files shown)


A5rocks (approved):
This makes sense to me. Did you run into this when running trio's test suite and this was the only issue?

jackwilsdon:
Yeah all other tests passed with IPv6 disabled :+1:

CoolCat467 (approved):
Looks good

A5rocks:
Made https://github.com/python-trio/trio/issues/3137 as a follow-up.

Issue #3129, opened 2024-11-03
pypy windows tests are running the same code on x64 and x86 arch matrix items


graingert:
There's no x86 pypy build available for windows and setup-python just falls back silently to the arch of the OS: https://github.com/actions/setup-python/issues/518#issuecomment-1272558547

so we're just burning CPU wastefully - the windows PyPy build should be moved to matrix.include

PR #3138, opened 2024-11-18
Only run x64 pypy

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3129

This is because "x86" pypy is just setup-python installing x64 pypy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['pypy-3.10', '3.9', '3.10', '3.11', '3.12', '3.13']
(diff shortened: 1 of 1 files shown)


A5rocks:
> @A5rocks can we get a test for this? Eg set the expected architecture somewhere and then check the architectures of sys.executable are a superset

Yeah that's a good idea. Maybe checking platform and having a dictionary + indexing into that.

A5rocks:
> It would catch say cpython dropping x86 support

I don't think it would, actually. Not if the test looks something like `<executable architecture> in [<list of architectures>]`, and I can't think of a form that would catch that. Maybe I'm just not thinking of the same test as you.

graingert:
It should be

```
assert any(arch == os.environ["EXPECTED_ARCH"] for arch in architectures(sys.executable))
```

PR #3141, opened 2024-11-19
Type `RaisesGroup` properly

A5rocks:
@jakkdl because you wrote this / these comments

This should fix the typing weirdness around it. I didn't really look too hard at things originally so there might be some outdated comments.

Maybe also @TeamSpen210 because types

... also this might be easier to review commit by commit. I tried my best to keep things separated (eg staging hunks so that only related pieces were in a single commit). However, I did not run mypy / tests after every single commit so things may be broken after a commit.

### DIFF
--- a/src/trio/_tests/test_testing_raisesgroup.py
+++ b/src/trio/_tests/test_testing_raisesgroup.py
@@ -24,7 +24,7 @@ def test_raises_group() -> None:
             f'Invalid argument "{TypeError()!r}" must be exception type, Matcher, or RaisesGroup.',
         ),
     ):
-        RaisesGroup(TypeError())
+        RaisesGroup(TypeError())  # type: ignore
 
     with RaisesGroup(ValueError):
         raise ExceptionGroup("foo", (ValueError(),))
--- a/src/trio/_tests/test_testing_raisesgroup.py
+++ b/src/trio/_tests/test_testing_raisesgroup.py
@@ -24,7 +24,7 @@ def test_raises_group() -> None:
             f'Invalid argument "{TypeError()!r}" must be exception type, Matcher, or RaisesGroup.',
         ),
     ):
-        RaisesGroup(TypeError())  # type: ignore
+        RaisesGroup(TypeError())  # type: ignore[call-overload]
 
     with RaisesGroup(ValueError):
         raise ExceptionGroup("foo", (ValueError(),))
--- a/src/trio/_tests/type_tests/raisesgroup.py
+++ b/src/trio/_tests/type_tests/raisesgroup.py
@@ -1,17 +1,3 @@
-"""The typing of RaisesGroup involves a lot of deception and lies, since AFAIK what we
-actually want to achieve is ~impossible. This is because we specify what we expect with
-instances of RaisesGroup and exception classes, but excinfo.value will be instances of
-[Base]ExceptionGroup and instances of exceptions. So we need to "translate" from
-RaisesGroup to ExceptionGroup.
-
-The way it currently works is that RaisesGroup[E] corresponds to
-ExceptionInfo[BaseExceptionGroup[E]], so the top-level group will be correct. But
-RaisesGroup[RaisesGroup[ValueError]] will become
-ExceptionInfo[BaseExceptionGroup[RaisesGroup[ValueError]]]. To get around that we specify
-RaisesGroup as a subclass of BaseExceptionGroup during type checking - which should mean
-that most static type checking for end users should be mostly correct.
-"""
-
 from __future__ import annotations
 
 import sys
@@ -37,25 +23,29 @@ def check_basic_contextmanager() -> None:
     # instead of an ExceptionGroup.
     with RaisesGroup(ValueError) as e:
         raise ExceptionGroup("foo", (ValueError(),))
-    assert_type(e.value, BaseExceptionGroup[ValueError])
+    assert_type(e.value, ExceptionGroup[ValueError])
 
 
 def check_basic_matches() -> None:
     # check that matches gets rid of the naked ValueError in the union
     exc: ExceptionGroup[ValueError] | ValueError = ExceptionGroup("", (ValueError(),))
     if RaisesGroup(ValueError).matches(exc):
-        assert_type(exc, BaseExceptionGroup[ValueError])
+        assert_type(exc, ExceptionGroup[ValueError])
+
+    # also check that BaseExceptionGroup shows up for BaseExceptions
+    if RaisesGroup(KeyboardInterrupt).matches(exc):
+        assert_type(exc, BaseExceptionGroup[KeyboardInterrupt])
 
 
 def check_matches_with_different_exception_type() -> None:
-    # This should probably raise some type error somewhere, since
+    # TODO: This should probably raise some type error somewhere, since
     # ValueError != KeyboardInterrupt
     e: BaseExceptionGroup[KeyboardInterrupt] = BaseExceptionGroup(
         "",
         (KeyboardInterrupt(),),
     )
     if RaisesGroup(ValueError).matches(e):
-        assert_type(e, BaseExceptionGroup[ValueError])
+        assert_type(e, ExceptionGroup[ValueError])
 
 
 def check_matcher_init() -> None:
@@ -121,16 +111,7 @@ def raisesgroup_check_type_narrowing() -> None:
 
 
 def raisesgroup_narrow_baseexceptiongroup() -> None:
-    """Check type narrowing specifically for the container exceptiongroup.
-    This is not currently working, and after playing around with it for a bit
-    I think the only way is to introduce a subclass `NonBaseRaisesGroup`, and overload
-    `__new__` in Raisesgroup to return the subclass when exceptions are non-base.
-    (or make current class BaseRaisesGroup and introduce RaisesGroup for non-base)
-    I encountered problems trying to type this though, see
-    https://github.com/python/mypy/issues/17251
-    That is probably possible to work around by entirely using `__new__` instead of
-    `__init__`, but........ ugh.
-    """
+    """Check type narrowing specifically for the container exceptiongroup."""
 
     def handle_group(e: ExceptionGroup[Exception]) -> bool:
         return True
@@ -138,18 +119,16 @@ def raisesgroup_narrow_baseexceptiongroup() -> None:
     def handle_group_value(e: ExceptionGroup[ValueError]) -> bool:
         return True
 
-    # should work, but BaseExceptionGroup does not get narrowed to ExceptionGroup
-    RaisesGroup(ValueError, check=handle_group_value)  # type: ignore
+    RaisesGroup(ValueError, check=handle_group_value)
 
-    # should work, but BaseExceptionGroup does not get narrowed to ExceptionGroup
-    RaisesGroup(Exception, check=handle_group)  # type: ignore
+    RaisesGroup(Exception, check=handle_group)
 
 
 def check_matcher_transparent() -> None:
     with RaisesGroup(Matcher(ValueError)) as e:
         ...
     _: BaseExceptionGroup[ValueError] = e.value
-    assert_type(e.value, BaseExceptionGroup[ValueError])
+    assert_type(e.value, ExceptionGroup[ValueError])
 
 
 def check_nested_raisesgroups_contextmanager() -> None:
@@ -162,15 +141,15 @@ def check_nested_raisesgroups_contextmanager() -> None:
     print(type(excinfo.value))  # would print "ExceptionGroup"
     assert_type(
         excinfo.value,
-        BaseExceptionGroup[BaseExceptionGroup[ValueError]],
+        ExceptionGroup[ExceptionGroup[ValueError]],
     )
 
     print(type(excinfo.value.exceptions[0]))  # would print "ExceptionGroup"
     assert_type(
         excinfo.value.exceptions[0],
         Union[
-            BaseExceptionGroup[ValueError],
-            BaseExceptionGroup[BaseExceptionGroup[ValueError]],
+            ExceptionGroup[ValueError],
+            ExceptionGroup[ExceptionGroup[ValueError]],
         ],
     )
 
@@ -183,7 +162,7 @@ def check_nested_raisesgroups_matches() -> None:
     )
 
     if RaisesGroup(RaisesGroup(ValueError)).matches(exc):
-        assert_type(exc, BaseExceptionGroup[BaseExceptionGroup[ValueError]])
+        assert_type(exc, ExceptionGroup[ExceptionGroup[ValueError]])
 
 
 def check_multiple_exceptions_1() -> None:
--- a/src/trio/_tests/type_tests/raisesgroup.py
+++ b/src/trio/_tests/type_tests/raisesgroup.py
@@ -223,11 +223,15 @@ def check_triple_nested_raisesgroup() -> None:
 
 
 def check_check_typing() -> None:
-    assert_type(  # type: ignore
-        RaisesGroup(ValueError).check,
+    # fmt: off
+    # mypy raises an error on `assert_type`
+    # pyright raises an error on `RaisesGroup(ValueError).check`
+    # to satisfy both, need to disable formatting and put it on one line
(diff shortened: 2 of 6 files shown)


src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -134,59 +112,46 @@ def handle_value(e: BaseExceptionGroup[ValueError]) -> bool:
...
 
 
 def check_nested_raisesgroups_contextmanager() -> None:
     with RaisesGroup(RaisesGroup(ValueError)) as excinfo:
         raise ExceptionGroup("foo", (ValueError(),))
 
-    # thanks to inheritance this assignment works
     _: BaseExceptionGroup[BaseExceptionGroup[ValueError]] = excinfo.value
-    # and it can mostly be treated like an exceptiongroup
     print(excinfo.value.exceptions[0].exceptions[0])
```
jakkdl, line 141:
```suggestion
```

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -134,59 +112,46 @@ def handle_value(e: BaseExceptionGroup[ValueError]) -> bool:
...
 
-    # but assert_type reveals the lies
     print(type(excinfo.value))  # would print "ExceptionGroup"
```
jakkdl, line 142:
```suggestion
```

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -134,59 +112,46 @@ def handle_value(e: BaseExceptionGroup[ValueError]) -> bool:
...
-    # typing says it's a BaseExceptionGroup
     assert_type(
         excinfo.value,
-        BaseExceptionGroup[RaisesGroup[ValueError]],
+        ExceptionGroup[ExceptionGroup[ValueError]],
     )
 
     print(type(excinfo.value.exceptions[0]))  # would print "ExceptionGroup"
```
jakkdl, line 148:
```suggestion
    print(type(excinfo.value.exceptions[0]))  # would print "ExceptionGroup"
```

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -134,59 +112,46 @@ def handle_value(e: BaseExceptionGroup[ValueError]) -> bool:
...
-    # but type checkers are utterly confused
     assert_type(
         excinfo.value.exceptions[0],
-        Union[RaisesGroup[ValueError], BaseExceptionGroup[RaisesGroup[ValueError]]],
+        Union[
+            ExceptionGroup[ValueError],
+            ExceptionGroup[ExceptionGroup[ValueError]],
+        ],
```
jakkdl, line 154: We can add a comment that `Union` is due to typeshed

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -23,19 +9,6 @@
...
-
-def check_inheritance_and_assignments() -> None:
-    # Check inheritance
-    _: BaseExceptionGroup[ValueError] = RaisesGroup(ValueError)
-    _ = RaisesGroup(RaisesGroup(ValueError))  # type: ignore
-
-    a: BaseExceptionGroup[BaseExceptionGroup[ValueError]]
-    a = RaisesGroup(RaisesGroup(ValueError))
-    a = BaseExceptionGroup("", (BaseExceptionGroup("", (ValueError(),)),))
-    assert a
```
jakkdl, line 37: note to self: This is removed because `RaisesGroup` is no longer a subclass of `BaseExceptionGroup`

src/trio/testing/_raises_group.py
```diff
@@ -35,12 +34,16 @@
...
 
     MatchE = TypeVar("MatchE", bound=BaseException, covariant=True)
+
 # RaisesGroup doesn't work with a default.
 E = TypeVar("E", bound=BaseException, covariant=True)
-# These two typevars are special cased in sphinx config to workaround lookup bugs.
+E2 = TypeVar("E2", bound=BaseException)
+E3 = TypeVar("E3", bound=BaseException)
+Ec2 = TypeVar("Ec2", bound=Exception)
+Ec3 = TypeVar("Ec3", bound=Exception)
```
jakkdl, line 43: Kinda weird to have `Ec2` and `Ec3` but no `Ec1`. Maybe `BaseExcT_co`, `BaseExcT_1`, `BaseExcT_2`, `ExcT_1`, `ExcT_2`?

src/trio/testing/_raises_group.py
```diff
@@ -35,12 +34,16 @@
     from typing import TypeVar
 
     MatchE = TypeVar("MatchE", bound=BaseException, covariant=True)
```
jakkdl, line 36: Can add a comment that `MatchE` messiness is because `default` was added in py313, and we don't want a runtime dependency on `typing_extensions`

src/trio/testing/_raises_group.py
```diff
@@ -332,61 +313,107 @@ class RaisesGroup(
...
         self,
         exception: type[E] | Matcher[E],
         *other_exceptions: type[E] | Matcher[E],
-        allow_unwrapped: Literal[False] = False,
         flatten_subgroups: Literal[True],
         match: str | Pattern[str] | None = None,
         check: Callable[[BaseExceptionGroup[E]], bool] | None = None,
     ) -> None: ...
 
+    # simplify the typevars if possible (the following 3 are equivalent but go simpler->complicated)
```
jakkdl, line 337: nice, but maybe nice to have a brief comment for each on how they differ
(`RaisesGroup[E]` -> `RaisesGroup[ExceptionGroup[E]]` -> `RaisesGroup[E1 | ExceptionGroup[E2]`)

jakkdl, line 337: Does this work for `RaisesGroup[RaisesGroup[RaisesGroup[Exception]]]`? Don't think there is a test for it.

src/trio/testing/_raises_group.py
```diff
@@ -332,61 +313,107 @@ class RaisesGroup(
...
-    It is also not typechecked perfectly, and that's likely not possible with the current approach. Most common usage should work without issue though.
     """
 
-    # needed for pyright, since BaseExceptionGroup.__new__ takes two arguments
-    if TYPE_CHECKING:
-
-        def __new__(cls, *args: object, **kwargs: object) -> RaisesGroup[E]: ...
-
-    # allow_unwrapped=True requires: singular exception, exception not being
-    # RaisesGroup instance, match is None, check is None
```
jakkdl, line 345: don't think this comment should be deleted

src/trio/testing/_raises_group.py
```diff
@@ -332,61 +313,107 @@ class RaisesGroup(
...
         self,
         exception: type[E] | Matcher[E],
         *,
         allow_unwrapped: Literal[True],
         flatten_subgroups: bool = False,
-        match: None = None,
-        check: None = None,
     ) -> None: ...
 
-    # flatten_subgroups = True also requires no nested RaisesGroup
```
jakkdl, line 357: nor this

src/trio/testing/_raises_group.py
```diff
@@ -332,61 +313,107 @@ class RaisesGroup(
...
         allow_unwrapped: bool = False,
         flatten_subgroups: bool = False,
         match: str | Pattern[str] | None = None,
-        check: Callable[[BaseExceptionGroup[E]], bool] | None = None,
+        # NOTE: I don't think the following argument type *should* work. But it does!
+        check: (
+            Callable[[BaseExceptionGroup[E2]], bool]
+            | Callable[[ExceptionGroup[Ec2]], bool]
+            | None
+        ) = None,
```
jakkdl, line 411: hmm, why accept this? I'm trying to come up with a case we want to accept but failing

src/trio/testing/_raises_group.py
```diff
@@ -514,21 +557,21 @@ def matches(
...
                 ):
-                    remaining_exceptions.remove(rem_e)  # type: ignore[arg-type]
+                    remaining_exceptions.remove(rem_e)
                     break
             else:
                 return False
-        return True
+
+        # only run `self.check` once we know `exc_val` is correct. (see the types)
+        # unfortunately mypy isn't smart enough to recognize the above fors as narrowing.
```
jakkdl, line 573:
```suggestion
        # unfortunately mypy isn't smart enough to recognize the above `for`s as narrowing.
```

jakkdl (changes requested):
Awesome work! I'm super happy to see progress here as I'm planning to pick up work on integrating it upstream soon :heart:

src/trio/_tests/type_tests/raisesgroup.py
```diff
@@ -46,29 +19,34 @@ def check_matcher_typevar_default(e: Matcher) -> object:
...
-    # ValueError != KeyboardInterrupt
     e: BaseExceptionGroup[KeyboardInterrupt] = BaseExceptionGroup(
         "",
         (KeyboardInterrupt(),),
     )
+
+    # note: it might be tempting to have this warn.
+    # however, note that we should not warn if e: BaseException
+    # .... therefore this needs to pass as there is no distinction
     if RaisesGroup(ValueError).matches(e):
```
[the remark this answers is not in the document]
A5rocks: Because if we type `.matches` like `.matches(e: <TypeVar bound to ValueError>)` then `.matches(BaseException())` will fail.

src/trio/testing/_raises_group.py
jakkdl: > For the record the actual type in __init__ is only used by the code within. It's not publicly shown.

oh derp. Though then the comment confuses me even more lol

src/trio/_tests/type_tests/raisesgroup.py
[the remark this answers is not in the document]
jakkdl: hmm, I think this might need higher-kinded typevars or https://github.com/python/typing/issues/1226 in order to express the relationship between `self` and `exc_val`. So yeah probably just update the comment and say it's a limitation in the typing system.

src/trio/testing/_raises_group.py
jakkdl: doesn't anything "work" because you have `type: ignore[arg-type]` on the only line that calls `self.check`?

The type from the signature does get exposed in `RaisesGroup(Exception).check`, but it seems all kinds of messy right now:
```python
def check_check() -> None:
    def my_check(e: ExceptionGroup[ValueError]) -> bool:
        return True
    r = RaisesGroup(ValueError, check=my_check)
    r.check(ExceptionGroup("", [ValueError()]))
```
```
error: "None" not callable  [misc]
error: Argument 1 has incompatible type "ExceptionGroup[ValueError]"; expected "BaseExceptionGroup[BaseExcT_1]"  [arg-type]
error: List item 0 has incompatible type "ValueError"; expected "ExcT_1"  [list-item]
```

Not 100% it's possible to improve it much though, might just end up with
```python
check: Callable[[BaseExceptionGroup[BaseException]], bool] | None
```

src/trio/_tests/type_tests/raisesgroup.py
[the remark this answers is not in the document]
A5rocks: I'm not sure even then because it would be weird if this was allowed:
```
exc: BaseException = e
RaisesGroup(OtherException).matches(exc)
```

But this failed:
```
RaisesGroup(OtherException).matches(e)
```

... it's much more on the lint rather than type check side of the scale.

src/trio/testing/_raises_group.py
A5rocks: (Not at all computer so can't check things right now)
> doesn't anything "work" because you have `type: ignore[arg-type]` on the only line that calls `self.check`?

Not exactly. It still needs to be a subtype of every overload's argument for that (which was why I was surprised and left the comment)

> 
> 
> 
> The type from the signature does get exposed in `RaisesGroup(Exception).check`, but it seems all kinds of messy right now:
> 
> ```python
> 
> def check_check() -> None:
> 
>     def my_check(e: ExceptionGroup[ValueError]) -> bool:
> 
>         return True
> 
>     r = RaisesGroup(ValueError, check=my_check)
> 
>     r.check(ExceptionGroup("", [ValueError()]))
> 
> ```
> 
> ```
> 
> error: "None" not callable  [misc]
> 
> error: Argument 1 has incompatible type "ExceptionGroup[ValueError]"; expected "BaseExceptionGroup[BaseExcT_1]"  [arg-type]
> 
> error: List item 0 has incompatible type "ValueError"; expected "ExcT_1"  [list-item]
> 
> ```
> 

Ouch I should improve it and add a test

> 
> 
> Not 100% it's possible to improve it much though, might just end up with
> 
> ```python
> 
> check: Callable[[BaseExceptionGroup[BaseException]], bool] | None
> 
> ```

I don't think this specific one works because it's not a subtype of every overload's argument, because callable is covariant in its arguments. Ie something that works for a specific type (like ValueError) might not work in general (for BaseException). For instance their check function may use an attribute on the exception.

However I agree that in general there's probably a better type hint. I'll look at it a bit.

src/trio/_tests/type_tests/raisesgroup.py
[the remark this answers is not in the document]
jakkdl: I'm not sure, when broadening you're giving the type checker more leeway to match things and prevent it from catching errors. Code like `if isinstance(e, ValueError)` will be considered `unreachable` if e is narrow but will work with `e: BaseException`. I agree it's on the linting end of typing, but why wouldn't you want to help lint with type hints.

But that's mostly academical, a comment like "This cannot be made to warn with current typing, and not entirely sure it would be correct to do so" should cover all bases.

src/trio/testing/_raises_group.py
A5rocks: OK I think the issue with `check` is a mypy bug:
```py

from typing import Generic, TypeVar, overload, Callable

T = TypeVar("T")
U1 = TypeVar("U1")
U2 = TypeVar("U2")

class X(Generic[T]):
    @overload
    def __init__(self: X[U1], *, x: U1) -> None: ...
    @overload
    def __init__(self: X[U2], *, y: U2) -> None: ...
    
    def __init__(self: X[U1 | U2], *, x: U1 | None = None, y: U2 | None = None, z: Callable[[U1], None] | Callable[[U2], None] | None = None):
        self.z = z

reveal_type(X(x=5).z)  # N: Revealed type is "Union[def (U1`-1), def (U2`-2), None]"
```

Ironically pyright has a bug in the same exact way :P

I guess I won't be fixing the typing. FWIW the type error you got from calling `.check` was because it's possible optional. It's still wrong because it has references to the typevars and not concrete classes of course.

CoolCat467: If you made an issue report for mypy, could you link it here if possible?

A5rocks: I made https://github.com/python/mypy/issues/18185

src/trio/_tests/test_testing_raisesgroup.py
```diff
@@ -24,7 +24,7 @@ def test_raises_group() -> None:
             f'Invalid argument "{TypeError()!r}" must be exception type, Matcher, or RaisesGroup.',
         ),
     ):
-        RaisesGroup(TypeError())
+        RaisesGroup(TypeError())  # type: ignore
```
CoolCat467, line 27: Is is possible to specify which error we are ignoring?

CoolCat467 (approved):
Looks great other than one small thing I noticed

src/trio/testing/_raises_group.py
```diff
@@ -247,38 +251,16 @@ def __str__(self) -> str:
...
-    # At runtime, use a redundant Generic base class which effectively gets ignored.
-    SuperClass = Generic
-
-
 @final
 class RaisesGroup(
-    AbstractContextManager[ExceptionInfo[BaseExceptionGroup[E]]],
-    SuperClass[E],
+    Generic[BaseExcT_co],
 ):
```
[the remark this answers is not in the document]
A5rocks: I removed it as part of initial cleaning up, since it wasn't necessary. I suspect it would interfere with overloads on the enter/exit, though I haven't checked.

CoolCat467 (approved):
Looks good!

PR #3144, opened 2024-11-25
Disable ruff's COM812

A5rocks:
Ruff's documentation recommends this be disabled with their formatter. I assume that advice holds for `black` too. The other rule they recommend disabling is `ISC001` which I don't think black has trouble with.

I don't see a better way to keep this up to date, but the way I noticed this probably works. I was making changes and pre-commit was making me run git commit thrice, as the `ruff` autofixes didn't pass the next round of `black`. The sequence was git commit: `black` -> `ruff` (because ordering), then git commit: `black`, then git commit: it actually worked.


CoolCat467:
I thought that https://github.com/python-trio/trio/pull/3044 was specifically for enabling this check, cannot say I support disabling it. I have noticed that some changes require multiple pre-commit runs as well, and while slightly annoying at times I don't think it merits disabling this rule entirely.

A5rocks:
> I thought that #3044 was specifically for enabling this check, cannot say I support disabling it. I have noticed that some changes require multiple pre-commit runs as well, and while slightly annoying at times I don't think it merits disabling this rule entirely.

I wasn't sure if there's other lints in that. But the alternatives are:
 - move the black run after ruff in the pre-commit file. this is alright, but also implies we're heavily relying on black's magic comma rule which I really dislike.
 - this

CoolCat467:
Now, if the rules created an edit loop that didn't resolve after a few iterations, then I might be for this change, but as is I think the commas apply in a way that makes the code a bit more readable, especially in cases with functions with several arguments.

A5rocks:
I'm not sure the alternative (reorder checks) works. We could duplicate the black check. This is because it seems ruff is whitespace sensitive. So take for instance this:
```py
nested_array = [
    [
        1,
        2,
        3,
    ]
]
```

Under [`black`'s new style, `black` -> `ruff` will yield (supposedly, I can't get this preview style working...)](https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#improved-multiline-dictionary-and-list-indentation-for-sole-function-parameter):
```py
nested_array = [[
    1,
    2,
    3,
]]
```

Whereas `ruff` -> `black` will yield:
```
nested_array = [
    [
        1,
        2,
        3,
    ],
]
```

---

I'll change this to duplicating the `black` entry in pre-commit.

CoolCat467 (approved):

jakkdl:
I personally prefer disabling COM812 over duplicating black. COM812 is perhaps nice, and ofc no reason to undo the changes in the PR, but I don't think it's worth having to duplicate black.


> Under [`black`'s new style, `black` -> `ruff` will yield (supposedly, I can't get this preview style working...)](https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#improved-multiline-dictionary-and-list-indentation-for-sole-function-parameter):

from the link:
> You can use a magic trailing comma to avoid this compacting behavior; by default, Black will not reformat the following code:

So I think because you have a trailing comma on the `3` you're not getting any changes from it.

If we can't make both ruff & black happy simultaneously I *strongly* think COM812 should be disabled, as we're otherwise gonna introduce headaches for LSP and/or IDE users that might be raising errors and/or autoformatting code.

A5rocks:
OK, so first... relitigating `COM812`. I ran `black -C src/trio` (which ignores all trailing commas), see the diff:

https://github.com/A5rocks/trio/compare/disable-trailing-comma...86b17b2e46700925786ffc463be6afc262948fe8

Then I ran `ruff check src/trio --unsafe-fixes` and `black src/trio` (and then `ruff check src/trio` and `black src/trio` again because... well this is a good case against COM812, since it popped up after the first cycle) and got this diff (this is, in theory, the result of `COM812` and probably little else):

https://github.com/A5rocks/trio/compare/86b17b2e46700925786ffc463be6afc262948fe8...6ac28c999090c586c96e79a5f09d82c5ed40741d

Interestingly, there's a decently large diff combining the two commits above. I won't make a PR for this because formatting *really* doesn't matter that much:

https://github.com/A5rocks/trio/compare/disable-trailing-comma...A5rocks:trio:commaless

---

I was expecting to be disappointed with the changes COM812 gives, but the changes are... fine. However, very few of the changes from COM812 (second link above) feel like they'd affect a diff: how much are we really modifying function signatures? I don't think changes other than that are useful.

I still think it should be disabled but I'm less sure of myself now... In isolation I think it works well and if it didn't conflict with `black` then I'm fine with it (more than I was when I approved the previous PR!). But duplicating `black` in the pre-commit file is a hack and yeah, having pre-commit avoid this (which it wouldn't based on my above experience of needing `ruff` -> `black` -> `ruff` -> `black` before stable output) doesn't help people relying on autoformatting in the command line or in their editor.

---

> So I think because you have a trailing comma on the `3` you're not getting any changes from it.

Not sure how I tested this before but it actually does work with the trailing comma on `3` now that I try it again... so, oh well.

jakkdl:
yeah let's go with disabling it then. No shame in trying something out and taking it back after deliberation, better than being overly afraid of doing stuff because we don't wanna change our minds :)

You can also remove the extra black run from `gen_exports.py` introduced in #3044 
https://github.com/CoolCat467/trio/blob/30babafa732799d6730fe5936c5ddbe8f453ad12/src/trio/_tools/gen_exports.py#L194-L197

CoolCat467:
Yea, I understand disabling this now. That is very unfortunate it doesn't play well with black.

pyproject.toml
```diff
@@ -131,15 +131,18 @@ select = [
...
+    "E402",    # module-import-not-at-top-of-file (usually OS-specific)
+    "E501",    # line-too-long
+    "F403",    # undefined-local-with-import-star
+    "F405",    # undefined-local-with-import-star-usage
+    "PERF203", # try-except-in-loop (not always possible to refactor)
+    "PT012",   # multiple statements in pytest.raises block
+    "SIM117",  # multiple-with-statements (messes up lots of context-based stuff and looks bad)
+
+    # conflicts with formatter (ruff recommends these be disabled)
+    "COM812",
```
webknjaz, line 145: Does Ruff have human-readable rule names? I thought it did... If so, I'd rather use those over some alphanumeric contextless IDs.

A5rocks: I don't think it does

jakkdl: ye they still can't be used ~anywhere

jakkdl (changes requested):
bumping/clarifying my previous comment:

> You can also remove the extra black run from gen_exports.py introduced in https://github.com/python-trio/trio/pull/3044
https://github.com/CoolCat467/trio/blob/30babafa732799d6730fe5936c5ddbe8f453ad12/src/trio/_tools/gen_exports.py#L194-L197

`gen_exports.py` currently runs black->ruff->black, which shouldn't be necessary without COM812

PR #3148, opened 2024-12-07
Remove unnecessary deadsnakes action

A5rocks:
I was looking at bumping actions versions, but everything is already up to date. The only action that had an older version was this deadsnakes action (and the codecov action we're having trouble with), and we managed to test 3.13 pre-releases without it (by specifying `3.13` on all matrixes).

Therefore I removed it. If that only works after a beta CPython release, then we can readd this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,16 +118,10 @@ jobs:
         uses: actions/checkout@v4
       - name: Setup python
         uses: actions/setup-python@v5
-        if: "!endsWith(matrix.python, '-dev')"
         with:
           python-version: ${{ fromJSON(format('["{0}", "{1}"]', format('{0}.0-alpha - {0}.X', matrix.python), matrix.python))[startsWith(matrix.python, 'pypy')] }}
           cache: pip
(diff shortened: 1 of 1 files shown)


CoolCat467 (approved):
Makes sense I think. Wondering, is 3.14 in beta yet? (I am under the assumption that 3.13 is mostly frozen now except fixing bugs?)

A5rocks:
> Makes sense I think. Wondering, is 3.14 in beta yet? (I am under the assumption that 3.13 is mostly frozen now except fixing bugs?)

Nope! Just alphas so far. https://peps.python.org/pep-0745/

Might still be worth putting in CI?

jakkdl:
> > Makes sense I think. Wondering, is 3.14 in beta yet? (I am under the assumption that 3.13 is mostly frozen now except fixing bugs?)
> 
> Nope! Just alphas so far. https://peps.python.org/pep-0745/
> 
> Might still be worth putting in CI?

if possible, sure! I suspect we'd be limited by dependencies/tooling, but good to know which ones.

Issue #3094, opened 2024-09-24
Update `SocketType.recv` (&other) signatures to use PEP570


jakkdl:
https://github.com/python/typeshed/pull/11250 changed the signatures of `SocketType.recv` and friends, so it should now be possible to change 
https://github.com/python-trio/trio/blob/4e979bfef4aac8c756fd4c6c521977cfcfdde1df/src/trio/_socket.py#L1115-L1129

although I'm surprised that mypy haven't started complaining from bumping typeshed. Maybe they did some backwards compatibility thing? Or haven't bumped typeshed?

PR #3151, opened 2024-12-10
Get rid of using `__` for positional only arguments

A5rocks:
Ruff started this in https://github.com/python-trio/trio/pull/3150. While the autofix it has is unsafe (and therefore won't happen in future versions so no need to make a bug report; https://github.com/astral-sh/ruff/pull/14818), why not finish the job and resolve https://github.com/python-trio/trio/issues/3094.

### DIFF
--- a/newsfragments/3094.misc.rst
+++ b/newsfragments/3094.misc.rst
@@ -1,0 +1,1 @@
+Switch to using PEP570 for positional-only arguments for `~trio.socket.SocketType`'s methods.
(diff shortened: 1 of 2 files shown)


CoolCat467 (approved):

jakkdl (approved):
Nice!

newsfragments/3094.misc.rst
```diff
@@ -0,0 +1 @@
+Switch to using PEP570 for positional-only arguments for `~trio.socket.SocketType`'s methods.
```
webknjaz, line 1: FYI, Sphinx has a `:pep:` role that one can use for PEP links in RST.

jakkdl: opened #3160

PR #3152, opened 2024-12-10
Update ignored branches for GHA's push trigger

A5rocks:
Based on conversation with @webknjaz on Gitter. This ignores pushes on:
 - autodeps
 - gh's merge queue
 - pre-commit.ci's update branch

This trigger is here so people can test their changes against CI without making a PR. The above types of PRs do not need this extra testing, as they will always have a PR (or more generally another event, for gh's merge queue) associated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,9 @@ name: CI
 on:
   push:
     branches-ignore:
-      - "dependabot/**"
+      - gh-readonly-queue/**
+      - autodeps/**
+      - pre-commit-ci-update-config
   pull_request:
   merge_group:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,10 @@ name: CI
 on:
   push:
     branches-ignore:
-      - gh-readonly-queue/**
-      - autodeps/**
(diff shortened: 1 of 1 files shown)


jakkdl (approved):

.github/workflows/ci.yml
```diff
@@ -3,7 +3,9 @@ name: CI
 on:
   push:
     branches-ignore:
-      - "dependabot/**"
+      - gh-readonly-queue/**
+      - autodeps/**
+      - pre-commit-ci-update-config
```
CoolCat467, line 8: Any reason not to use string quoting? I've recently learned that yaml parsing is very, very strange when parsing things not marked as string literals

CoolCat467 (approved):
Looks good as far as I understand

.github/workflows/ci.yml
webknjaz: I prefer no quotes where possible and use `>-` to force strings as it looks cleaner. But FWIW in this case, there's no chance YAML would parse cast these values as anything but string. We're safe here.

webknjaz (approved):
I think it's good but I'd recommend adding code comments and recording purposes of each entry in YAML.

Issue #3153, opened 2024-12-10
Codecov stuff


jakkdl:
I've been bothered by codecov's github annotations for a long while, where it fairly regularly kept complaining about previously-uncovered code still being uncovered after later commits fixing it. The latest instance of this happened over in [pytest](https://github.com/pytest-dev/pytest/pull/13016#discussion_r1868983441) but I know it's confused me and others at various points in different trio repositories. (happened in #2812)

When I started digging I found out this functionality is deprecated and has a big red box in the documentation https://docs.codecov.com/docs/github-checks, and I'm fairly certain that more time has been lost collectively from false alarms (and code being near-unreadable in the review window when it's very low on coverage) than from the marginal gain of seeing it in the code review window rather than elsewhere. We'll still have the CI check and everything, and you can easily click through to codecov.io in several places.

I've opened an issue upstream https://github.com/codecov/codecov-action/issues/1710, but in the meantime I think we should edit the YAML at https://app.codecov.io/account/github/python-trio/yaml/

It's a one-line fix, that doesn't even need a PR, but thought I should open it to discussion and make people aware of the change so they don't wonder where it went. And if you personally want the functionality there's a browser extension: https://docs.codecov.com/docs/the-codecov-browser-extension

webknjaz:
@jakkdl in my experience, one thing that causes Codecov to post statuses prematurely (not just annotations but Checks, comments and their web UI) is allowing it to guess the moment when it's okay to start processing the report files and combining them into one.
I think this is because when there's a matrix of jobs not covering the same lines each, there would be portions of lines for which coverage might end up being reported much later than for many other jobs. And if Codecov decides to process the reports earlier than that, it usually reports lower coverage than it really is. It then updates these statuses later, when that slow job coverage arrives, though. The problem is that the annotations are immutable IIRC and so they can't update them, while the comment and the checks would be.

To fight this, I like setting the config as follows, setting the number of expected uploads explicitly: https://github.com/aio-libs/yarl/blob/f304dd7/.codecov.yml#L3-L8.

jakkdl:
ooh, that'd resolve #2689 that has been bugging me forever. Thanks!

But I don't think it'd resolve the problems in the pytest PR where the uncovered code was introduced in one commit, and the tests adding coverage added in a later commit.

> The problem is that the annotations are immutable IIRC and so they can't update them, while the comment and the checks would be.

yeah the codecov docs say
[...]

PR #3156, opened 2024-12-17
Codecov improvements

### LINKED ISSUE #2689: Is it possible to make codecov action not "fail" until all other runs are completed?

jakkdl:
Fixes #2689 (see #3153), although I'm not 100% sure this is still a problem - or if I've just gotten used to marking failed PR runs as read in my notifications. I'll go test it EDIT: This is still a thing - if you're watching runs in progress I'm getting a :x: (and it was never about notifications anyway)

#2653 added [flags](https://docs.codecov.com/docs/flags), but configured them erronously with multiple flags per run, so we've been getting lots of flag errors - see e.g. https://app.codecov.io/gh/python-trio/trio/commit/785dc59309ca4f491562b817836092f54b7da4b1
But I'm not sure if flags are terribly useful for us, e.g. wanting 100% coverage on windows is not something we care about, so I removed them.

I enabled [codecov.notify.require_changes](https://docs.codecov.com/docs/pull-request-comments) so it won't add a comment if there's no changes.

And finally I set the target for patches to 100%, where it by default targets project cov (currently 99.68%) which meant that in big PR's you could by mistake add an uncovered line.

Most settings inspired by https://github.com/aio-libs/yarl/blob/f304dd7/.codecov.yml, thanks @webknjaz

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.8.2
+    rev: v0.8.3
     hooks:
       - id: ruff
         types: [file]
--- a/src/trio/_tests/test_highlevel_socket.py
+++ b/src/trio/_tests/test_highlevel_socket.py
@@ -20,6 +20,12 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
 
 
+@pytest.mark.xfail(
+    sys.platform == "darwin" and sys.version_info[:3] == (3, 13, 1),
+    reason="TODO: This started failing in CI after 3.13.1",
+    raises=OSError,
+    strict=True,
+)
 async def test_SocketStream_basics() -> None:
     # stdlib socket bad (even if connected)
     stdlib_a, stdlib_b = stdlib_socket.socketpair()
--- a/src/trio/_tests/test_socket.py
+++ b/src/trio/_tests/test_socket.py
@@ -458,6 +458,12 @@ async def test_SocketType_basics() -> None:
     sock.close()
 
 
+@pytest.mark.xfail(
+    sys.platform == "darwin" and sys.version_info[:3] == (3, 13, 1),
+    reason="TODO: This started failing in CI after 3.13.1",
+    raises=OSError,
+    strict=True,
+)
 async def test_SocketType_setsockopt() -> None:
     sock = tsocket.socket()
     with sock as _:


webknjaz:
@jakkdl flag errors is just an annoying UI thing. I reported it upstream some time ago.

.github/workflows/ci.yml
```diff
@@ -87,7 +87,6 @@ jobs:
           directory: empty
           token: 87cefb17-c44b-4f2f-8b30-1fff5769ce46
```
webknjaz, line 88: It's better to move this token to the config as it would be easier to keep it in one shared place.

.github/workflows/ci.yml
```diff
@@ -87,7 +87,6 @@ jobs:
...
           name: Windows (${{ matrix.python }}, ${{ matrix.arch }}${{ matrix.extra_name }})
-          flags: Windows,${{ matrix.python }}
```
webknjaz, line 90: I'd keep the flags despite the UI alerts.

.github/workflows/ci.yml
```diff
@@ -199,7 +196,6 @@ jobs:
           directory: empty
           token: 87cefb17-c44b-4f2f-8b30-1fff5769ce46
           name: Alpine
-          flags: Alpine,3.12
```
webknjaz, line 202: By the way, computing the Python version dynamically is usually better instead of tracking hardcoded values..
When `setup-python` is used, it provides an output. Otherwise, `python -V` is your friend...

.codecov.yml
```diff
@@ -0,0 +1,25 @@
...
+
+# Explicitly wait for all jobs to finish, as wait_for_ci prematurely triggers.
+# See https://github.com/python-trio/trio/issues/2689
+codecov:
+  notify:
+    # This number needs to be changed whenever the number of runs in CI is changed.
+    # Another option is codecov-cli: https://github.com/codecov/codecov-cli#send-notifications
+    after_n_builds: 78
+    wait_for_ci: false
+    notify_error: true # if uploads fail, replace cov comment with a comment with errors.
```
webknjaz, line 11: I think it makes sense to also set `fail-on-error` in the action invocations. It's false by default, resulting in hidden failures in the job logs.

webknjaz (commented):
Note that the current action version is v5. It might be worth upgrading.

.github/workflows/ci.yml
jakkdl: I suppose we could keep them, but nobody has pressed the button to enable analytics so currently it's entirely unused anyway https://app.codecov.io/gh/python-trio/trio/flags

I still find it somewhat questionable to track coverage for "python3.11" or "windows" as separate entities. Maybe you could diagnose issues by seeing big swings in coverage for one of them*, or any of them being very low, but the actual coverage % for each flag is mostly useless unless we'd do something like marking all platform/version-specific code somehow, exclude from coverage for that flag, and require 100% coverage on common code on all platforms... (which all sounds like way too much effort for close-to-no payoff).

\* could use https://docs.codecov.com/docs/commit-status#threshold

A5rocks: I think I've used the flags once (when looking at updating the codecov action) but I don't remember how + given we haven't figured the issue out I assume it wasn't useful.

jakkdl: I guess 1>0 and there's not really any downside to have them, so I'll add them back in. I'll go enable analytics too for fun.

.github/workflows/ci.yml
webknjaz: v4 introduced OIDC, which would only work in trusted contexts (which PRs from forks aren't). They have corner case handling for such PRs.
With v5, it's possible to toggle tokenless uploads in the Codecov UI. Not sure if this is an org-only setting or it could be per-repo. I haven't tested it.

.github/workflows/ci.yml
webknjaz: I use them heavily in many projects. It's nice to see which likes are covered under which env or a combo of those. There's a way to filter it on the Codecov UI.

As for the checks and 100% coverage requirement, I usually don't make those depend on the flags, except for MyPy, perhaps. Codecov is able to track the combined coverage for those checks. I can look into improving this later.

webknjaz:
> we had a lot of troubles with v4

v4 had a lot of bugs when it was first released, half of the reports on the first day of the release were probably mine. It should be stable now. It was another uploader rewrite in Python (v3 was JS), and v5 has yet another wrapper so it feels a bit more risky to rush into it.

.github/workflows/ci.yml
```diff
@@ -131,13 +134,13 @@ jobs:
         env:
           CHECK_FORMATTING: '${{ matrix.check_formatting }}'
           NO_TEST_REQUIREMENTS: '${{ matrix.no_test_requirements }}'
-      - if: always()
+      - if: ${{ always() && matrix.check_formatting != '1' }}
```
webknjaz, line 137: How about dropping the mustache?

```suggestion
      - if: >-
          always()
          & & matrix.check_formatting != '1'
```

.github/workflows/ci.yml
```diff
@@ -193,13 +196,16 @@ jobs:
         run: python -m venv .venv
       - name: Run tests
         run: source .venv/bin/activate && ./ci.sh
+      - name: Get Python version for codecov flag
+        id: get-version
+        run: echo "version=$(python -V | cut -d' ' -f2 | cut -d'.' -f1,2)" > $GITHUB_OUTPUT
```
webknjaz, line 201: It must be append, never overwrite.
```suggestion
        run: echo "version=$(python -V | cut -d' ' -f2 | cut -d'.' -f1,2)" >> "${GITHUB_OUTPUT}"
```

.codecov.yml
```diff
@@ -0,0 +1,29 @@
+# -- repository yaml --
+
+# Explicitly wait for all jobs to finish, as wait_for_ci prematurely triggers.
+# See https://github.com/python-trio/trio/issues/2689
+codecov:
+  notify:
+    # This number needs to be changed whenever the number of runs in CI is changed.
+    # Another option is codecov-cli: https://github.com/codecov/codecov-cli#send-notifications
+    after_n_builds: 78
```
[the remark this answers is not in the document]
webknjaz: You're right. It's
```suggestion
    after_n_builds: 27
```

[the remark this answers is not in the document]
webknjaz: The higher number is coming from the scheduled runs that upload more reports on cron, if a commit stays on top of `main` for a while.

webknjaz (changes requested):

.codecov.yml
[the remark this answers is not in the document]
jakkdl: Aha, thank you. I suppose there might be some way of filtering that out, though it probably doesn't matter much.
Weird that `78/27=2.89` isn't a whole number

.github/workflows/ci.yml
jakkdl: It's a bit of a pain that https://app.codecov.io/gh/python-trio/trio/flags shows all flags ever used, with 1/3 being for old versions we no longer support. I'm not sure there's a decent way to fix that without doing manual flag management

.codecov.yml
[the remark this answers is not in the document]
webknjaz: > Weird that 78/27=2.89 isn't a whole number

I also noticed this but haven't looked into why. It might be that some of those nightly runs fail to upload from some jobs (the current default is to silently ignore this in the action — we may be able to confirm once this PR is merged). Another possibility is that the scheduled runs have a different number of jobs.

.github/workflows/ci.yml
webknjaz: Yeah, you can clean them up. Although, to me, the most useful view with flags is the per-file line coverage pages.

webknjaz (approved):
I haven't noticed anything blocking so let's merge this. Not sure why macos is failing, though. It should be unrelated, right?

I'd like to do some follow-up improvements post-merge since I've been doing this across a number of projects anyway. Might as well transfer the lessons learned here later.

CoolCat467:
Yea macos run is blocking https://github.com/python-trio/trio/pull/3155 as well, and A5rocks found some details about potentially why: https://github.com/python-trio/trio/pull/3155#issuecomment-2547491499

CoolCat467 (approved):
Makes sense I think, I hope this helps improve things for the future

PR #3157, opened 2024-12-17
Clean up check.sh, move stuff to pre-commit

jakkdl:
Ruff, Black, Codespell and ~~gen-exports~~ are all run in pre-commit, so there's no need to also run them in check.sh
~~moving uv pip-compile to pre-commit was easy~~

The only remaining tools in check.sh now are:

1. **pyright** pyright+pre-commit+CI does not play well, as pyright ~requires an internet connection and pre-commit in CI isolates the environment not to have it. [In flake8-async we have a dedicated CI action for pyright](https://github.com/python-trio/flake8-async/blob/bd48157dba26289afbc5cb389a1c56b8f440badc/.github/workflows/ci.yml#L11) that uses an [unofficial github action](https://github.com/jakebailey/pyright-action).
EDIT: Though we also have `check_type_completeness.py` that won't play well with that pyright-action.

3. **mypy** I think doing 3 full runs (all platforms) would be overly slow for local development, but I'm not sure it's possible to specify manual stages to run with the pre-commit github action. So best solution here is perhaps to configure a mypy pre-commit hook (using current system?) and once again add a dedicated CI action.

Addresses some of #2699


jakkdl:
ah right, of course pip-compile also requires internet. There's maybe a configuration with `repo: local` and `language: system` but eh. Unclear if it's worth running locally

check.sh
```diff
@@ -11,44 +11,6 @@ if [ -z "${GITHUB_STEP_SUMMARY+x}" ]; then
     ON_GITHUB_CI=false
 fi
 
-# Test if the generated code is still up to date
-echo "::group::Generate Exports"
-python ./src/trio/_tools/gen_exports.py --test \
-    || EXIT_STATUS=$?
-echo "::endgroup::"
```
A5rocks, line 18: I probably should do this in another PR but I think we should move gen exports out of pre-commit and back here. Unfortunately, as things currently are, if you modify a file that the pre-commit hook says it uses then commit with a git client (ie not just `git commit -m "..."` in the venv you have for trio), gen_exports doesn't get the dependencies it needs.

So far I've been getting that error then going to my terminal to run git there, but that's a bit much.

check.sh
```diff
@@ -94,8 +56,6 @@ if git status --porcelain | grep -q "requirements.txt"; then
     echo "::endgroup::"
 fi
 
-codespell || EXIT_STATUS=$?
```
A5rocks, line 97: (not the right line because github isn't letting me comment there)

Should you remove the `uv pip compile` lines above? And also update the error message + add a `pre-commit run -a` line in `check.sh` so people can run it locally? (I'm not sure people do run it locally but...)

check.sh
CoolCat467: Pre-commit with a external git client doesn't install required dependencies?

A5rocks: Nope, because we don't specify them. We don't specify them because pre-commit doesn't let you install from a requirements file :(

CoolCat467 (commented):
Moving things to pre-commit looks great, just wondering about how this will work with CI. Will CI check to make sure pre-commit run succeeds before allowing merge, or do we have to change what is required in the "all greens" workflow?

check.sh
jakkdl: I specified them in the latest commit: https://github.com/python-trio/trio/blob/17d3d051fe7486a8f3731852a14438b87046cf60/.pre-commit-config.yaml#L46

but that doesn't let us pin the versions (in a non-awful way), so yeah if we don't want that then it has to go outside of pre-commit

check.sh
jakkdl: Realized we can't do `uv pip compile` in pre-commit in CI since it needs internet, so it needs to be here (or somewhere).

My goal is not to require/expect people to run check.sh locally as it's unintuitive and clunky, but given the limitations of pre-commit we do need to stash these somewhere and make it easy for users to run them locally if they want.
I'd personally vote for tox, and set up environments for the different tests.

.pre-commit-config.yaml
```diff
@@ -40,7 +40,30 @@ repos:
     hooks:
       - id: regenerate-files
         name: regenerate generated files
-        language: system
+        language: python
```
webknjaz, line 43: I noticed that strings are quoted randomly in some places and not others. It's rather inconsistent. Perhaps, just don't quote them unnecessarily?

.pre-commit-config.yaml
```diff
@@ -40,7 +40,30 @@ repos:
...
+    hooks:
+      # Compile requirements
+      - id: pip-compile
+        name: uv pip-compile test-requirements.in
+        args: [
+          "--universal",
+          "--python-version=3.9",
+          "test-requirements.in",
+          "-o",
+          "test-requirements.txt"]
```
webknjaz, line 59: I think this reads better:
```suggestion
        args:
        - --universal
        - --python-version=3.9
        - --output-file=test-requirements.txt
        - test-requirements.in
```

check.sh
webknjaz: > https://github.com/python-trio/trio/blob/17d3d051fe7486a8f3731852a14438b87046cf60/.pre-commit-config.yaml#L46

Keep in mind that the pre-commit cache is virtually immortal. You can clean it locally, but not in the `pre-commit.ci` service. For that, changing the deps list or the hook version is required. This may be problematic sometimes.

So I've found that the best thing to do here is to keep the version pins in the list and bump them periodically. Just so that the cache doesn't grow too old.

.pre-commit-config.yaml
CoolCat467: That does look a bit nicer, I agree

.pre-commit-config.yaml
jakkdl: that's just me being bad at yaml and not knowing what's necessary or not :p
perhaps time for a yaml-formatter

.pre-commit-config.yaml
jakkdl: in a python repo I'd argue that using python-looking syntax is going to be easier for people to understand.

CoolCat467: That is a pretty good argument actually

webknjaz: Dunno, it's usually harder for me as it looks like type annotations, but it isn't.

.pre-commit-config.yaml
webknjaz: Well, then you'd have to wrap literally **everything** in the file, including the keys:
```suggestion
        "language": "python"
```
Numbers like `3.11` are cast to float otherwise. And `3.10` caught many people by surprise back in the day (it's actually `3.1` unless quoted explicitly).

FWIW, I don't really like quotes. But if something needs escaping, I tend to use the following syntax (*especially* if there are quotes in the value that I don't want to bother about):
```suggestion
        language: >-
          python
```

.pre-commit-config.yaml
graingert: I think we should use a yaml formatter

jakkdl: see #3164, I'll do whatever for this PR and then it can be enforced later.

.pre-commit-config.yaml
jakkdl: see #3164, I'll do whatever for this PR and then it can be enforced later.

check.sh
jakkdl: I think that means we need to keep it in `check.sh` and skip it in ci for now. I'll keep the changes to the pre-commit hook to alleviate @A5rocks's problem, and maybe use `tox` or something later on.

jakkdl:
[once again registering my displeasure with not being able to squash]
I'll manually squash and force push after giving y'all a chance to review, since force pushing breaks review history on github.

I've scaled back the PR to:
1. remove redundant formatting/linting from check.sh (although I'm not sure if pre-commit is 'required' to merge? so not strictly 100% redundant)
2. change pre-commit entry for regenerate-files to have dependencies instead of relying on the local environment having them.
3. Add pip-compile checks to pre-commit, which are skipped in CI.

webknjaz:
> * remove redundant formatting/linting from check.sh (although I'm not sure if pre-commit is 'required' to merge? so not strictly 100% redundant)

No, it's not required, but I'd ask @A5rocks to make it mandatory in the ruleset.

> * Add pip-compile checks to pre-commit, which are skipped in CI.

To address such cases, I like duplicating the pre-commit invocation in conventional CI, which I'd recommend here as well. If you want to, you can select/skip some checks to make sure that only those that can't run in the pre-commit.ci platform would run on GHA. But you can also not care and run everything additionally — it's usually inexpensive.

check.sh
webknjaz: Tox sounds wonderful. I've been meaning to suggest it. I've actually been working on some reusable workflows for tox-based projects + experimented with some cool GHA integrations.
When you get to it, I may offer some help / experimentation in simplifying things here.

check.sh
webknjaz: I usually wrap the pre-commit invocation in tox for local run anyway + run it in CI too, which covers the case of the service not having the internet during testing.

check.sh
jakkdl: Yeah I'm personally a big fan of tox. #2790 discussed hatchling too but we ended up bikeshedding too much to actually do anything in the end.

check.sh
```diff
@@ -79,8 +67,7 @@ Problems were found by static analysis (listed above).
 To fix formatting and see remaining errors, run
 
     uv pip install -r test-requirements.txt
-    black src/trio
-    ruff check src/trio
+    pre-commit run --all-files
```
graingert, line 70: hook stage manual?

jakkdl:
Hrm, I just noticed this warning
>   warning: The requested Python version 3.11 is not available; 3.13.1 will be used to build dependencies instead.

https://github.com/python-trio/trio/actions/runs/12465050038/job/34790209503?pr=3157
which is also happening on main/ atm @A5rocks 
> warning: The requested Python version 3.9 is not available; 3.13.1 will be used to build dependencies instead.

https://github.com/python-trio/trio/actions/runs/12446716072/job/34749424898#step:5:1098


I think the easiest fix will be tox, although it doesn't seem to have mattered so far

check.sh
jakkdl: ?
We don't have any manual hooks anymore

jakkdl:
running it through pre-commit I appear to have bumped uv from 5.5 to ~5.11, so I suppose they started sorting the version/platform constraints at some point (though not seeing it from a quick skim of the changelogs https://github.com/astral-sh/uv/releases).

graingert:
Can we use '3.x' for dists and code checks?

jakkdl:
> Can we use '3.x' for dists and code checks, instead of 3.11?

if the python version doesn't matter for `uv pip-compile` we could simply opt out of passing `--python-version`

A5rocks:
The warning should be fine -- surely uv has an option to never build? We can pass that I think.

I think passing python-version is good because it's a simpler lockfile.

check.sh
jakkdl: I've yet to use nox, but would def be open to.

graingert: I've contributed to projects with nox, and they're slightly more inconvenient. I'd also want to run nox python files under coverage.

one thing to bear in mind with tox is that constraints are a bit of a pain, tox-uv and a lock file seems the way to go:

https://github.com/tox-dev/tox-uv?tab=readme-ov-file#uvlock-support

check.sh
graingert: ah I see, it might be worth using hook stage manual here, just in case any get added

check.sh
CoolCat467: We could also add a local system hook to pre-commit that runs `check.sh` if we were worried about people forgetting to run it

check.sh
jakkdl: it feels kinda weird suggesting people do stuff that is completely redundant/ineffective. And if we do add manual stages there's perhaps gonna be a decent reason we make them manual? Idk I'll leave it as is /shrug

check.sh
jakkdl: > We could also add a local system hook to pre-commit that runs `check.sh` if we were worried about people forgetting to run it

There's several problems with doing this currently:
1. mypy is actually failing in my workdir, probably due to something dirty in my env. tox/nox would fix this
2. running `check.sh` takes a full minute on my current system, so I don't want to run it on every commit (`pre-commit run --all-files` takes ~8 seconds)
3. I catch most type errors from running `mypy` through pylsp in my editor anyway

So I don't ~ever run `check.sh` locally, and don't expect others to either. (Nor do I run `ci.sh`, I prefer manually running `pytest`)

This def isn't super obvious for new contributors though, but tox/nox should help with that. E.g. in tox you can create a label for a collection of environments that is considered "standard" to run locally.

jakkdl:
We could do more nitpicks for another week, but if you think there's anything substantive I left on the table I encourage opening another issue/PR :)

check.sh
webknjaz: I've been taking DIY lock files to extremes with tox for a while. Haven't yet tried uv for that, really. I know that they were implementing the poerty-style format, not real lock files. I'm looking forward to seeing Brett's PEP accepted, honestly.

I also have a lot more experience with tox and haven't contributed to nox. Although, I have contributed to projects using nox. There's limitations for parallelism, for example. Some scripting would be more convenient in Python. However, tox 4 has a `toxfile.py` that should be able to cover a lot of that. I'm yet to play with it properly.

PR #3158, opened 2024-12-19
💄 Refactor the coveragepy config

webknjaz:
This is a combination of changes that include:
- setting up local coveragepy HTML output to show context
- setting up mapping between site-packages and the repo
- measuring all the Python files in the repo
- including Cython
- making xfails not influence the coverage metric (these test functions may start executing different amounts of lines over time, causing random coverage drops/gains)
- using relative paths for playing nice with Codecov

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -234,15 +234,34 @@ jobs:
             cache: pip
         # setuptools is needed to get distutils on 3.12, which cythonize requires
       - name: install trio and setuptools
-        run: python -m pip install --upgrade pip . setuptools
+        run: python -m pip install --upgrade pip . setuptools 'coverage[toml]'
+
+      - name: add cython plugin to the coveragepy config
+        run: >-
+          sed -i 's#plugins\s=\s\[\]#plugins = ["Cython.Coverage"]#'
+          pyproject.toml
 
       - name: install cython & compile pyx file
+        env:
+          CFLAGS: ${{ env.CFLAGS }} -DCYTHON_TRACE_NOGIL=1
         run: |
           python -m pip install "cython${{ matrix.cython }}"
-          cythonize --inplace tests/cython/test_cython.pyx
+          cythonize --inplace -X linetrace=True tests/cython/test_cython.pyx
 
       - name: import & run module
-        run: python -c 'import tests.cython.test_cython'
+        run: coverage run -m tests.cython.run_test_cython
+
+      - name: get Python version for codecov flag
+        id: get-version
+        run: >-
+          echo "version=$(python -V | cut -d' ' -f2 | cut -d'.' -f1,2)"
+          >> "${GITHUB_OUTPUT}"
+      - if: always()
+        uses: codecov/codecov-action@v5
+        with:
+          name: Cython
+          flags: Cython,${{ steps.get-version.outputs.version }}
+          fail_ci_if_error: true
 
   # https://github.com/marketplace/actions/alls-green#why
   check:  # This job does nothing and is only used for the branch protection
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -268,6 +268,10 @@ directory = "misc"
 name = "Miscellaneous internal changes"
 showcontent = true
 
+[tool.coverage.html]
+show_contexts = true
+skip_covered = false
+
 [tool.coverage.run]
 branch = true
 source_pkgs = ["trio"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -291,6 +291,8 @@ relative_files = true
 [tool.coverage.report]
 precision = 1
 skip_covered = true
+skip_empty = true
+show_missing = true
 exclude_lines = [
     "pragma: no cover",
     "abc.abstractmethod",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -272,6 +272,16 @@ showcontent = true
 show_contexts = true
 skip_covered = false
 
+[tool.coverage.paths]
+_site-packages-to-src-mapping = [
+  "src",
+  "*/src",
+  '*\src',
+  "*/lib/pypy*/site-packages",
+  "*/lib/python*/site-packages",
+  '*\Lib\site-packages',
+]
+
 [tool.coverage.run]
 branch = true
 source_pkgs = ["trio"]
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -268,6 +268,20 @@ directory = "misc"
...
 
+[tool.coverage.html]
+show_contexts = true
+skip_covered = false
+
+[tool.coverage.paths]
+_site-packages-to-src-mapping = [
+  "src",
+  "*/src",
+  '*\src',
```
CoolCat467, line 279: Why does it have both forward and backwards slash?

CoolCat467 (approved):
Looks good as far as I can tell, just a question about one of the config changes

pyproject.toml
CoolCat467: also one below this also has a backwards slash

webknjaz: Probably not, but I saw this practice a lot in the wild. I think modern coveragepy normalizes them but I don't want to take a chance.

webknjaz:
Yeah, I only had to adapt my typical dedicated config (which is easier to maintain) to the toml combo. It's all battle-tested, otherwise.

pyproject.toml
CoolCat467: Probably not what?

webknjaz: Oh, my brain parsed it as “is this necessary?” :laughing: Apologies.

These are basically explicit Windows path variants.

CoolCat467: Ah, that makes a lot of sense.

pyproject.toml
```diff
@@ -282,10 +296,15 @@ omit = [
 # The test suite spawns subprocesses to test some stuff, so make sure
 # this doesn't corrupt the coverage files
 parallel = true
+plugins = []
+relative_files = true
+source = ["."]
```
graingert, line 301: Why do we need source and source_pkgs? We should only need one

tests/cython/run_test_cython.py
```diff
@@ -0,0 +1 @@
+from . import test_cython  # noqa: F401
```
graingert, line 1: Possibly it makes sense to refactor test_cython to export a main function and we can then call it here

.github/workflows/ci.yml
```diff
@@ -234,15 +234,34 @@ jobs:
...
-        run: python -c 'import tests.cython.test_cython'
+        run: coverage run -m tests.cython.run_test_cython
+
+      - name: get Python version for codecov flag
+        id: get-version
+        run: >-
+          echo "version=$(python -V | cut -d' ' -f2 | cut -d'.' -f1,2)"
+          >> "${GITHUB_OUTPUT}"
+      - if: always()
+        uses: codecov/codecov-action@v5
```
jakkdl, line 260: is there any problem with mixing versions? I previously hit errors I thought was caused by that.

pyproject.toml
jakkdl: Having recently skimmed a bunch of codecov docs I'm pretty sure there's enough path normalization mentioned in a bunch of places that this shouldn't be needed.

pyproject.toml
webknjaz: This is needed to measure coverage outside `src/`. One setting can only contain importable names while the other can have both importables and paths.

I had to do a lot of experiments to arrive to this config shape. Other variants may output incorrect paths into the XML report, confusing Codecov.

tests/cython/run_test_cython.py
webknjaz: Yeah, I thought so. But it was late and I was lazy :)

Will do!

.github/workflows/ci.yml
webknjaz: No, they just use different uploader and APIs. But newer versions are less prone to flaky upload crashes for some reason, as I noticed.

pyproject.toml
webknjaz: @jakkdl this is not Codecov but coveragepy.

pyproject.toml
webknjaz: (this also allows including that Cython smoke test)

webknjaz:
pre-commit.ci run

pyproject.toml
graingert: Probably that smoke test should be inside src/trio/_test

graingert (changes requested):
Let's use source, or source_pkgs not both

pyproject.toml
graingert: Alternatively we could use `source_pkgs = ["trio", "tests"]`

graingert (changes requested):

pyproject.toml
webknjaz: @graingert this won't work well. The combination of the two is the only config that is known not to cause problems one way or the other. Plus, it represents the shape of the project exactly — the source is the entire Git repo; the importable is whatever is in `site-packages` but it's also mapped back to the source for proper representation within both coveragepy and codecov.

I've seen cases where it would display `test_*.py` files in the root of the repository in Codecov while showing the correct paths in the coveragepy's own HTML output. And that's just one example of how things tend to break.

Can we please rely on my experience here? I happened to maintain the use of coveragepy specifically a lot in the past few years, and would like to avoid weird behaviors just because this project wants to deviate from what's gone through countless hours of debugging rather surprising corner cases.

webknjaz:
Merge queue didn't work for the first time because of the flakiness of the Codecov v3 action, FTR.

I've added it back. As for the test relocation and other suggestions, they are out of the scope because I don't want to fit in more requirements into this effort. Those changes can always be tried out separately and shouldn't be blocking — and it would be easier to see if they work in isolation, since this already combines a bit more changes than I'd like in one PR.

pyproject.toml
graingert: I added the source_pkgs feature to coveragepy for this use case

webknjaz: @graingert really? I didn't know!

Still, there were cases where this specific configuration layout was the only thing that made it possible for Codecov not to break in some cases. This was in both src- and flat-layout projects (some in aio-libs, some elsewhere).

Though, I feel like it's still important to cover other things that might appear outside `src/`. Sometimes, these are helper scripts or something similar — by collecting such coverage, it is possible to configure Codecov-reported statuses better and be more conscious about what to take into account through exclusions.

As for adding `"tests"` into `source_pkgs`, that might work, but would not address the “covering everything” part. Plus, I'm worried about the side effects it may have on Codecov.

PR #3159, opened 2024-12-19
get and enforce 100% coverage

graingert:
Using a combination of more tests and TODO comments that work as coverage pragmas

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -21,7 +21,7 @@ comment:
 
 coverage:
   # required range
-  range: 99.6..100
+  range: 99.8..100
   status:
     # require patches to be 100%
     patch:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -26,6 +26,9 @@ coverage:
   range: 100..100
   status:
     # require patches to be 100%
+    project:
+      default:
+        target: 100%
     patch:
       default:
         target: 100%
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -25,10 +25,9 @@ coverage:
   round: down
   range: 100..100
   status:
-    # require patches to be 100%
     project:
       default:
         target: 100%
     patch:
       default:
-        target: 100%
+        target: 100%  # require patches to be 100%
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -21,7 +21,9 @@ comment:
 
 coverage:
   # required range
-  range: 99.8..100
+  precision: 5
+  round: down
+  range: 100..100
   status:
     # require patches to be 100%
     patch:
--- a/newsfragments/3159.misc.rst
+++ b/newsfragments/3159.misc.rst
@@ -1,0 +1,1 @@
+Get more coverage
--- a/newsfragments/3159.misc.rst
+++ b/newsfragments/3159.misc.rst
@@ -1,1 +1,1 @@
-Get more coverage
+Get and enforce 100% coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -305,18 +305,16 @@ precision = 1
 skip_covered = true
 skip_empty = true
 show_missing = true
-exclude_lines = [
-    "pragma: no cover",
-    "abc.abstractmethod",
-    "if TYPE_CHECKING.*:",
-    "if _t.TYPE_CHECKING:",
-    "if t.TYPE_CHECKING:",
-    "@overload",
-    'class .*\bProtocol\b.*\):',
-    "raise NotImplementedError",
-]
 exclude_also = [
   '^\s*@pytest\.mark\.xfail',
+   "abc.abstractmethod",
+   "if TYPE_CHECKING.*:",
+   "if _t.TYPE_CHECKING:",
+   "if t.TYPE_CHECKING:",
+   "@overload",
+   'class .*\bProtocol\b.*\):',
+   "raise NotImplementedError",
+   'TODO: test this line'
 ]
 partial_branches = [
     "pragma: no branch",
@@ -326,4 +324,5 @@ partial_branches = [
     "if .* or not TYPE_CHECKING:",
     "if .* or not _t.TYPE_CHECKING:",
     "if .* or not t.TYPE_CHECKING:",
+   'TODO: test this branch',
 ]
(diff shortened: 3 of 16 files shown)


src/trio/_core/_tests/test_io.py
```diff
@@ -381,6 +384,38 @@ def check(*, expected_readers: int, expected_writers: int) -> None:
         check(expected_readers=1, expected_writers=0)
 
 
+@pytest.mark.skipif(sys.platform in {"win32", "linux"})
```
graingert, line 387:
```suggestion
@pytest.mark.skipif(sys.platform in {"win32", "linux"}, reason="no kqueue")
```

src/trio/_core/_tests/test_io.py
```diff
@@ -381,6 +384,38 @@ def check(*, expected_readers: int, expected_writers: int) -> None:
         check(expected_readers=1, expected_writers=0)
 
 
+@pytest.mark.skipif(sys.platform in {"win32", "linux"}, reason="no kqueue")
```
CoolCat467, line 387: From mypy run annotations, looks like you need to inform mypy that you don't expect this function to run on windows or linux, because mypy doesn't understand `pytest.mark.skipif`.
I suggest something like this:
```python
if sys.platform == "win32" or sys.platform == "linux":
    raise AssertionError("Should be unreachable")
```

CoolCat467 (approved):
Didn't see anything imminently except for a mypy issue

CoolCat467 (changes requested):
Prior comment

src/trio/_core/_tests/test_io.py
webknjaz: @CoolCat467 perhaps, `assert_never()`?

webknjaz: Or better yet `assert sys.platform not in {"win32", "linux"}`

src/trio/_core/_tests/test_io.py
```diff
@@ -381,6 +384,38 @@ def check(*, expected_readers: int, expected_writers: int) -> None:
...
+async def test_io_manager_kqueue_monitors_statistics() -> None:
```
graingert, line 388:
```suggestion
async def test_io_manager_kqueue_monitors_statistics() -> None:
    assert sys.platform not in {"win32", "linux"}
```

src/trio/_core/_tests/test_io.py
graingert: I couldn't get either option working (assert sys.platform != "win32" and sys.platform != "linux") does work in the playground, and if I create a new file.

.codecov.yml
```diff
@@ -21,7 +21,9 @@ comment:
 
 coverage:
   # required range
-  range: 99.6..100
+  precision: 5
+  round: down
+  range: 100..100
   status:
     # require patches to be 100%
     patch:
```
graingert, line 29:
```suggestion
    project:
      default:
        target: 100%
    patch:
```

.codecov.yml
```diff
@@ -26,6 +26,9 @@ coverage:
   range: 100..100
   status:
     # require patches to be 100%
```
webknjaz, line 28: This comment is not related to “project” but to “patch” so it should be moved. Such cases are why I prefer code comments on the same line and not separate, whenever possible...

.codecov.yml
```diff
@@ -26,6 +26,9 @@ coverage:
...
+    project:
+      default:
+        target: 100%
```
webknjaz, line 31: FTR with #3162, there's going to be a need to split the checks into MyPy vs. pytest.

.codecov.yml
graingert: Oh yeah I was going to get back to this once I've got 100% reliably

pyproject.toml
```diff
@@ -305,18 +305,16 @@ precision = 1
...
 exclude_also = [
   '^\s*@pytest\.mark\.xfail',
+   "abc.abstractmethod",
+   "if TYPE_CHECKING.*:",
+   "if _t.TYPE_CHECKING:",
+   "if t.TYPE_CHECKING:",
+   "@overload",
+   'class .*\bProtocol\b.*\):',
+   "raise NotImplementedError",
+   'TODO: test this line'
```
webknjaz, line 317: Consider integrating the `covdefaults` plugin, as it injects most of these automatically. The only thing you'd need to override additionally is the `fail_under` setting, which it sets to 100%, but it's not compatible with runtime-dependent tests that cannot reach 100% under one specific runtime. So it's good to set it to something sufficiently low while requiring 100% in Codecov.

graingert: ah nice, I'll do that in a followup

newsfragments/3159.misc.rst
```diff
@@ -0,0 +1 @@
+Get more coverage
```
graingert, line 1:
```suggestion
Get and enforce 100% coverage
```

jakkdl (changes requested):
oops, I forgot to submit my pending comments I wrote yesterday

src/trio/_tools/gen_exports.py
```diff
@@ -180,22 +180,13 @@ def run_linters(file: File, source: str) -> str:
...
-    success, response = run_ruff(file, response)
-    if not success:  # pragma: no cover  # Test for run_ruff should catch
-        print(response)
-        sys.exit(1)
-
-    success, response = run_black(file, response)
-    if not success:
-        print(response)
-        sys.exit(1)
+    for fn in (run_black, run_ruff, run_black):
```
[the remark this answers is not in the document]
graingert: 3144 is merged now, do you still want me to drop the second black run? 

I'm thinking to not change behavior in this PR, and do behavior changes in a follow up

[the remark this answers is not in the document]
jakkdl: yeah, wrote the comment before merging the PR.

This is zero change in behaviour, and "should"ve been in #3144 but since nobody bothered to check out the PR and do the edit there /shrug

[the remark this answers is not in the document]
graingert: Ok, I added it

jakkdl (approved):
nice!

PR #3166, opened 2024-12-23
Add spellchecking with `crate-ci/typos`

CoolCat467:
In this pull request, we add https://github.com/crate-ci/typos as a pre-commit spellchecker and fix various typos found with this particular spellchecker.

They claim to have more features than codespell, but I don't really see a reason why we couldn't use both as long as they aren't fighting each other.

I learned about crate-ci/typos quite recently when looking through changes made to the mypy daemon as a part of mypy 1.14.0, so I'm not super familiar with how the configuration works, but I think it's at least mostly working as is. What I've tried to add as known "identifiers" with `-`s in them don't seem to be functioning as I intended however.


src/trio/_core/_run.py
```diff
@@ -348,7 +348,7 @@ class CancelStatus:
     # become effectively cancelled when they do, unless scope.shield
     # is True.  May be None (for the outermost CancelStatus in a call
     # to trio.run(), briefly during TaskStatus.started(), or during
-    # recovery from mis-nesting of cancel scopes).
+    # recovery from miss-nesting of cancel scopes).
```
jakkdl, line 351: this, and the other "miss-nesting"s, are incorrect fixes. https://en.m.wiktionary.org/wiki/misnested

CoolCat467: Fixed with 4431971

pyproject.toml
```diff
@@ -84,6 +84,22 @@ force-exclude = '''
...
 ignore-words-list = 'astroid,crasher,asend'
 
+[tool.typos]
+default.extend-ignore-re = [
+    "(?Rm)^.*(#|//)\\s*spellchecker:disable-line$",
+]
+
+[tool.typos.default.extend-identifiers]
+IPPROTO_ND = "IPPROTO_ND"
+ERROR_INVALID_PARMETER = "ERROR_INVALID_PARMETER"
```
jakkdl, line 94: I think PARMETER -> PARAMETER should be fixed, the error in cffi is in fact called `ERROR_INVALID_PARAMETER` https://www.magnumdb.com/search?q=error_invalid_parameter

I don't really understand `_windows_cffi.py`, so I don't know if that will be a backwards incompatible change, or if user code trying to handle that error is currently broken. But either case it should be fixed.

src/trio/_abc.py
```diff
@@ -192,9 +192,7 @@ async def getaddrinfo(
         Any required IDNA encoding is handled before calling this function;
         your implementation can assume that it will never see U-labels like
         ``"café.com"``, and only needs to handle A-labels like
-        ``b"xn--caf-dma.com"``.
-
-        """
+        ``b"xn--caf-dma.com"``."""  # spellchecker:disable-line
```
jakkdl, line 195: xn--caf is both in extend-identifiers in pyproject.toml and has `spellchecker:disable-line`?

jakkdl (changes requested):

src/trio/_abc.py
CoolCat467: For some reason extend-identifiers does does not work as I intended and continues to try replacing it, so I added `spellchecker:disable-line` there to solve the issue. If there would be a way to get extend-identifiers to do ignore that one like I intended, that would be a better solution.

pyproject.toml
CoolCat467: Fixed in 9f57781.

src/trio/_abc.py
jakkdl: in that case we should at least remove the non-functioning line in extend-identifiers, unless you wanna report it upstream.

jakkdl:
with PARMETER->PARAMETER this needs a newsfragment

jakkdl (approved):

PR #3167, opened 2024-12-23
use constraints file for sdist/wheel build + get version using build.util 

graingert:
stuff that I wanted to sneak into the zizmor PR, but decided should be in another PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
           persist-credentials: false
 
       - name: Install build
-        run: python -Im pip install build
+        run: python -Im pip install build -c test-requirements.txt
 
       - name: Get the dist version
         id: dist-version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,14 +40,19 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Install build
+        run: python -Im pip install build
+
       - name: Get the dist version
         id: dist-version
-        run: >-
-          echo "version=$(
-          grep ^__version__ src/trio/_version.py
-          | sed 's#__version__ = "\([^"]\+\)"#\1#'
-          )"
-          >> "${GITHUB_OUTPUT}"
+        shell: python
+        run: |
+          import os
+          import build.util
+          version = build.util.project_wheel_metadata(".").version
+
+          with open(os.environ["GITHUB_OUTPUT"], mode="a", encoding="utf8") as f:
+              f.write(f"version={version}")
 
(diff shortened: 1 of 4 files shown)


webknjaz:
If you're going this route already, let's also set `PIP_CONSTRAINT` in the pypa/build invocation step. I like using a separate constraint file generated with pip-tools, containing the build deps pins like setuptools+wheel. This will be able to pin the ephemeral PEP 517 build env.

.github/workflows/ci.yml
```diff
@@ -40,14 +40,19 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Install build
+        run: python -Im pip install build -c test-requirements.txt
```
webknjaz, line 44: This would leak into the underlying PEP 517 ephemeral build envs.
```suggestion
        env:
          PIP_CONSTRAINT: test-requirements.txt
        run: python -Im pip install build
```

.github/workflows/ci.yml
```diff
@@ -58,9 +63,6 @@ jobs:
           DIST_NAME: ${{ env.dist-name }}
           VERSION: ${{ steps.dist-version.outputs.version }}
 
-      - name: Install build
-        run: python -Im pip install build
-
       - name: Build dists
         run: python -Im build
```
webknjaz, line 67: Assuming, you'll do something like `pip-compile --only-build-deps --output-file=build-requirements.txt pyproject.toml`:
```suggestion
        env:
          PIP_CONSTRAINT: build-requirements.txt
        run: python -Im build
```

ci.sh
```diff
@@ -41,7 +41,7 @@ python -m pip install -U pip uv -c test-requirements.txt
 python -m pip --version
 python -m uv --version
 
-python -m uv pip install build
+python -m uv pip install build -c test-requirements.txt
```
webknjaz, line 44: This could also use an env var:
```suggestion
PIP_CONSTRAINT=test-requirements.txt python -m uv pip install build
```

webknjaz (commented):
Additionally, I realized that this project doesn't have a combined workflow and so this https://github.com/python-trio/trio/blob/main/.github/workflows/release.yml also needs to be handled similarly.

.github/workflows/ci.yml
A5rocks: I suddenly remembered this PR when looking at the release notes for pip 26.1! I think this would need to be `PIP_BUILD_CONSTAINT` now.

(I thought we had merged this though xd)

webknjaz: Certainly!

PR #3171, opened 2024-12-25
Explain why our MacOS 14 runners failed with CPython 3.13.1

A5rocks:
There's two possible solutions that I see:
 - set our `runs-on` to be `macos-15`
 - ignore the error, it's not showing something bad anyways

Since the first one is somewhat annoying (we would change it back to `macos-latest` at some point and local tests would fail for people), the second one is ideal. So I implemented that here.

### DIFF
--- a/src/trio/_tests/test_highlevel_socket.py
+++ b/src/trio/_tests/test_highlevel_socket.py
@@ -20,12 +20,6 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
 
 
-@pytest.mark.xfail(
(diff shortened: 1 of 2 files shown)


CoolCat467 (approved):
Looks good, but wondering why ignoring `PT017`?

A5rocks:
> Looks good, but wondering why ignoring `PT017`?

It wants me to use `pytest.raises` but I don't want to assert that an exception will happen, only make sure that it's OK *if* it happens.

PR #3172, opened 2024-12-25
Bump version to 0.28.0

A5rocks:

PR #3161, opened 2024-12-20
Bump Codecov action to v5

webknjaz:
It's more stable and requires a token that is already in the config.

PR #3173, opened 2024-12-25
Stop failing CI upon codecov error 503

A5rocks:
https://github.com/python-trio/trio/pull/3172 took 5 attempts to merge.......

cc @jakkdl

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,6 +223,7 @@ jobs:
           # multiple flags is marked as an error in codecov UI, but is actually fine
           # https://github.com/codecov/feedback/issues/567
           flags: Windows,${{ matrix.python }}
+          fail_ci_if_error: false  # change to true when using codecov action v5
 
   Ubuntu:
     name: 'Ubuntu (${{ matrix.python }}${{ matrix.extra_name }})'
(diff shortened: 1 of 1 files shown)


CoolCat467 (approved):

webknjaz:
This would make Codecov less reliable. It'd be better to figure out the version bump.

CoolCat467:
While I agree that it's unfortunate that this would make Codecov less reliable, I think stopping pull requests from going through just because Codecov is buggy is a worse in the end.

A5rocks:
> This would make Codecov less reliable. It'd be better to figure out the version bump.

Does a new uploader really decrease the amount of 503s codecov returns? I note that the failures I checked were due to that, not due to some uploader issue.

webknjaz:
Yes, in my experience, v4 eliminated this problem. I think the retries are better, but also the API is new. I don't really have the stats for v5, though.

The only problem is that my attempt to bump to v5 has some weird coverage computation issue: #3161. I haven't really looked into that. But it's weird that I haven't seen this behavior elsewhere, so my hunch is that it might be related to the structure of the automation in this repo.

A5rocks:
Ok should we merge this as a temporary patch (and a comment saying as much), until someone gets the time to look at v5?

I'm willing to wait a few days just in case we do it quickly but after that I think it's better to be able to get PRs merged on the first attempt. (PR statuses are fine because individual jobs can be rerun, but merge queue needs the rerun every job)

jakkdl:
I think we should merge this as a temp fix due to how massively painful it is to work with.

It's def suboptimal, but /shrug

Issue #3175, opened 2024-12-29
nursery replaces custom ExceptionGroup subclasses with plain ExceptionGroups


graingert:
```python
import trio

class MyExceptionGroup(ExceptionGroup):
    pass


async def main():
    async with trio.open_nursery():
        raise MyExceptionGroup("hello", [ValueError("bad")])

trio.run(main)
```

output:
```pytb
  + Exception Group Traceback (most recent call last):
  |   File "/home/graingert/projects/demo.py", line 11, in <module>
  |     trio.run(main)
  |   File "/home/graingert/.virtualenvs/anyio_fixture_methods/lib/python3.12/site-packages/trio/_core/_run.py", line 2306, in run
  |     raise runner.main_task_outcome.error
  |   File "/home/graingert/projects/demo.py", line 8, in main
  |     async with trio.open_nursery():
  |   File "/home/graingert/.virtualenvs/anyio_fixture_methods/lib/python3.12/site-packages/trio/_core/_run.py", line 959, in __aexit__
  |     raise combined_error_from_nursery
  | ExceptionGroup: Exceptions from Trio nursery (1 sub-exception)
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "/home/graingert/projects/demo.py", line 9, in main
    |     raise MyExceptionGroup("hello", [ValueError("bad")])
    | ExceptionGroup: hello (1 sub-exception)
    +-+---------------- 1 ----------------
      | ValueError: bad
      +------------------------------------
```

expected output:

```pytb
  + Exception Group Traceback (most recent call last):
  |   File "/home/graingert/projects/demo.py", line 11, in <module>
  |     trio.run(main)
  |   File "/home/graingert/.virtualenvs/anyio_fixture_methods/lib/python3.12/site-packages/trio/_core/_run.py", line 2306, in run
  |     raise runner.main_task_outcome.error
  |   File "/home/graingert/projects/demo.py", line 8, in main
  |     async with trio.open_nursery():
  |   File "/home/graingert/.virtualenvs/anyio_fixture_methods/lib/python3.12/site-packages/trio/_core/_run.py", line 959, in __aexit__
  |     raise combined_error_from_nursery
  | ExceptionGroup: Exceptions from Trio nursery (1 sub-exception)
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "/home/graingert/projects/demo.py", line 9, in main
[...]

PR #3176, opened 2024-12-30
Add a sanity check for a missing `derive` for exception groups

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3175, probably.

Just adds a sanity check. I started thinking about doing our own `.split()` but then I decided that's a lot. I also looked at just using `.derive` at the end but the exception groups without `.derive()` can be anywhere in the exception group contents.

I made https://github.com/python-trio/flake8-async/issues/334 for the linter rule side of this, but I think this is probably easy enough to forget especially if trying out this feature (and easy to overlook).

### DIFF
--- a/newsfragments/3175.doc.rst
+++ b/newsfragments/3175.doc.rst
@@ -1,0 +1,1 @@
+Warn if a user forgot to implement ``.derive`` for an ExceptionGroup subclass.
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -639,6 +639,31 @@ class CancelScope:
                 self.cancelled_caught = True
                 exc = None
             elif isinstance(exc, BaseExceptionGroup):
+                # sanity check users
+                egs = [exc]
+                visited = set()
+                while egs:
+                    next_eg = egs.pop()
+                    if next_eg in visited:
+                        continue
+                    visited.add(next_eg)
+                    if (
+                        "derive" not in type(next_eg).__dict__
+                        and type(next_eg) is not ExceptionGroup
+                    ):
+                        warnings.warn(
+                            f"derive not implemented for {type(next_eg).__name__}, results may be unexpected",
+                            stacklevel=1,
+                        )
+
+                    egs.extend(
+                        [
+                            e
+                            for e in next_eg.exceptions
+                            if isinstance(e, BaseExceptionGroup)
+                        ]
+                    )
+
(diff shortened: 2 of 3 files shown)


TeamSpen210:
Seems rather expensive to put in `CancelScope.__exit__`? Not sure if it really makes any sense to be here in particular either, since exception groups can be raised anywhere, and they're not Trio-specific either.

A5rocks:
> Seems rather expensive to put in `CancelScope.__exit__`?

Maybe the runtime cost is too high yeah.

> Not sure if it really makes any sense to be here in particular either, since exception groups can be raised anywhere, and they're not Trio-specific either.

That's the only place in trio where we call `exception_group.derive` (indirectly) from `collapse_exception_group` or `exception_group.split`. In addition, it's probably a good idea to let users know, even if this might not be their first encounter.

jakkdl:
I'm on the fence on whether we should check for this at runtime at all, esp when it gets this messy. And making people add `derive` to suppress the warning is quite a bit more expensive then adding a `noqa`.

On the other hand this would catch cases in libraries not running an appropriate linter, and an end user using that library+trio. But the error doesn't feel very trio-specific in the first case.

A5rocks:
Actually this would false positive if someone mirrors the BaseExceptionGroup<->ExceptionGroup structure (where only BaseExceptionGroup has a derive)...

So maybe this isn't that useful of something to always warn about. I don't think everyone uses linters but I suppose if someone needs a custom ExceptionGroup odds are they use a linter?

oremanj:
I don't think this is Trio's job to detect, especially since the implementation is complicated and likely to be slow. I would much prefer to see us address the hazard with a lint, and/or a CPython change to warn in `BaseExceptionGroup.__init_subclass__` if a subclass does not define `derive`.

A5rocks:
I do think it's trio's place to try to prevent footguns (or even common mistakes) and I do think this qualifies. But I'll close this because:
 - runtime cost of doubling (maybe?) execution time for exiting cancel scopes upon errors for what should be a very rare occurrence 
 - false positives are onerous and likely common, given how even ExceptionGroup doesn't implement derive

Mostly the second reason though. If someone comes up with a runtime thing that does better, I would love another attempt at this -- I totally could see myself ignoring documentation when playing around with a new thing (given I don't have linting in my REPL and I don't run lints until committing).

PR #3184, opened 2025-01-13
add now required sphinx.configuration in readthedocs.yml

jakkdl:
see https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/


A5rocks (approved):
Makes sense and builds pass now (I see previously it did not, because they removed backcompat for today)

PR #3186, opened 2025-01-14
Publically re-export `__version__`

CoolCat467:
In this pull request, we publicly re-export `__version__`. This came up when working on https://github.com/python-trio/trio-websocket/pull/193, where some of the unit test code looks at Trio's version for version-dependent things and mypy is not happy about accessing `__version__`.

### DIFF
--- a/src/trio/__init__.py
+++ b/src/trio/__init__.py
@@ -104,9 +104,6 @@ from ._timeouts import (
     sleep_forever as sleep_forever,
     sleep_until as sleep_until,
 )
-
-# pyright explicitly does not care about `__version__`
-# see https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness
 from ._version import __version__ as __version__
 
 # Not imported by default, but mentioned here so static analysis tools like
--- a/src/trio/__init__.py
+++ b/src/trio/__init__.py
@@ -107,7 +107,7 @@ from ._timeouts import (
 
 # pyright explicitly does not care about `__version__`
 # see https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness
-from ._version import __version__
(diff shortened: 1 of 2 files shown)


A5rocks (approved):
IMO we should encourage `importlib.metadata` (eg in the newsfragment). But this is fine as a bugfix to something we thought worked.

src/trio/__init__.py
```diff
@@ -107,7 +107,7 @@
 
 # pyright explicitly does not care about `__version__`
 # see https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness
```
A5rocks, line 109: Actually could you remove this comment?

CoolCat467: Removed in 379d9ce

jakkdl:
> IMO we should encourage `importlib.metadata` (eg in the newsfragment). But this is fine as a bugfix to something we thought worked.

well, it *does* work currently so this isn't really a bugfix per se. By publicly re-exporting it we're making it part of the official API, so I think perhaps this shouldn't be done at all?

A5rocks:
> > IMO we should encourage `importlib.metadata` (eg in the newsfragment). But this is fine as a bugfix to something we thought worked.
> 
> well, it _does_ work currently so this isn't really a bugfix per se. By publicly re-exporting it we're making it part of the official API, so I think perhaps this shouldn't be done at all?

The comment implies we thought it would work, and that it does in pyright. Also, is this PR even necessary?:

```
(.venv) PS C:\Users\A5rocks\Documents\Temp\a> mypy -c "import trio; reveal_type(trio.__version__)"
<string>:1: note: Revealed type is "builtins.str"
Success: no issues found in 1 source file
```

CoolCat467:
Try with `--strict` or at the very least `--no-implicit-reexport`

PR #3187, opened 2025-01-15
Implement highlevel unix socket listeners

### LINKED ISSUE #279: High-level helpers for working with unix domain sockets

CoolCat467:
In this pull request, we implement highlevel unix socket support. This is an alternate implementation of and resolves #1433.

I will say, at the moment, only unix `SOCK_STREAM` (TCP) sockets are supported with this change, but I could pretty easily implement unix datagram socket listeners as well.

Closes #279


src/trio/_highlevel_open_unix_listeners.py
```diff
@@ -0,0 +1,173 @@
...
+# https://github.com/python-trio/trio/wiki/notes-to-self#measure-listen-backlogpy
+def _compute_backlog(backlog: int | None) -> int:
+    # Many systems (Linux, BSDs, ...) store the backlog in a uint16 and are
+    # missing overflow protection, so we apply our own overflow protection.
+    # https://github.com/golang/go/issues/5030
+    if not isinstance(backlog, int) and backlog is not None:
+        raise TypeError(f"backlog must be an int or None, not {backlog!r}")
+    if backlog is None:
+        return 0xFFFF
+    return min(backlog, 0xFFFF)
```
A5rocks, line 58: (the comment above is missing an ending parenthesis)

I'm confused. Is this literally copying over `_highlevel_open_tcp_listeners.py`? We should just import it and use its `_compute_backlog`.

src/trio/_highlevel_open_unix_listeners.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+async def open_unix_listener(
+    path: str | bytes | os.PathLike[str] | os.PathLike[bytes],
+    *,
+    mode: int | None = None,  # 0o666,
```
A5rocks, line 64: Unnecessary comment

src/trio/_highlevel_open_unix_listeners.py
```diff
@@ -0,0 +1,173 @@
...
+        raise FileNotFoundError(f"Socket folder does not exist: {folder!r}")
+
+    # much more simplified logic vs tcp sockets - one socket type and only one
+    # possible location to connect to
+    sock = tsocket.socket(AF_UNIX, tsocket.SOCK_STREAM)
+    try:
+        # See https://github.com/python-trio/trio/issues/39
+        if sys.platform != "win32":
+            sock.setsockopt(tsocket.SOL_SOCKET, tsocket.SO_REUSEADDR, 1)
+
```
A5rocks, line 112: Is this really necessary?

src/trio/_highlevel_open_unix_listeners.py
```diff
@@ -0,0 +1,173 @@
...
+        await sock.bind(str(fspath))
```
A5rocks, line 113: I'm concerned about this because the original PR does some fun stuff here. Specifically, what if there are two calls to this trying to make a file, does this work? Could we add some tests for this? (along with other tests -- just noticed there are none.)

A5rocks: Also the original PR seems to handle this path being e.g. a directory. Does this PR? (yet another test!)

src/trio/_highlevel_open_unix_listeners.py
```diff
@@ -0,0 +1,173 @@
...
+            await fspath.chmod(mode)
+
+        return trio.UnixSocketListener(sock)
+    except BaseException as exc:
+        sock.close()
+        try:
+            os.unlink(str(fspath))
+        except BaseException as exc_2:
+            raise exc_2 from exc
+        raise
```
A5rocks, line 127:
```suggestion
    finally:
        sock.close()
        os.unlink(str(fspath))
```
does this have the same semantics?

src/trio/_highlevel_socket.py
```diff
@@ -31,6 +32,8 @@
     errno.ENOTSOCK,
 }
 
+HAS_UNIX: Final = hasattr(tsocket, "AF_UNIX")
```
A5rocks, line 35: IMO this is cleaner than the definition in `_highlevel_open_unix_listeners.py`

src/trio/_highlevel_socket.py
```diff
@@ -412,3 +415,76 @@ async def aclose(self) -> None:
         """Close this listener and its underlying socket."""
         self.socket.close()
         await trio.lowlevel.checkpoint()
+
+
+@final
+class UnixSocketListener(Listener[SocketStream]):
+    """A :class:`~trio.abc.Listener` that uses a listening socket to accept
+    incoming connections as :class:`SocketStream` objects.
```
A5rocks, line 423: I notice we seem to duplicate the above. Something we could do is avoid exporting these classes and instead require something like

```py
listener = trio.into_listener(socket)
```

This would mean we don't need a `@final` and can potentially provide a smaller API. (I'd like someone with more experience with trio/networking tell me if this is a decent idea or not)

A5rocks (changes requested):
Preliminary comments but I'd like to see some tests!

@Tronic if you see this then maybe you can check these changes seem sane? (alternatively, feel free to move them all to your PR and we can switch to merging that)

src/trio/_highlevel_socket.py
A5rocks: Alternatively, we can make `SocketListener` support unix sockets trivially: I assume there's a way to detect this for a socket object. All that changes is:
 - the `socket.type` check
 - `aclose` has an extra `unlink`
 - some docs maybe

Then we don't even need to check for whether unix sockets are supported -- if `SocketListener` is passed one, then obvioiusly they're supported!

newsfragments/3187.feature.rst
A5rocks, line 1: Should be `279.feature.rst` I believe.

src/trio/_highlevel_open_unix_listeners.py
CoolCat467: Yes, copying `_compute_backlog`, will import instead.

src/trio/_highlevel_open_unix_listeners.py
CoolCat467: Might not be, will test.

CoolCat467: Nope, not needed, will remove.

src/trio/_highlevel_open_unix_listeners.py
CoolCat467: Wouldn't that close the socket at the end even if it was successful?

A5rocks: Oh actually yes it would, I'm not sure what I was thinking...

src/trio/_highlevel_open_unix_listeners.py
CoolCat467: Removed in c922a52

src/trio/_highlevel_open_unix_listeners.py
CoolCat467: Simplified in c922a52

src/trio/_highlevel_socket.py
CoolCat467: Added path functionality to `SocketListener` in c922a52

newsfragments/3187.feature.rst
CoolCat467: Renamed in c922a52

src/trio/_highlevel_open_unix_listeners.py
CoolCat467: Removed in c922a52

src/trio/_highlevel_open_unix_listeners.py
CoolCat467: Imported in c922a52

src/trio/_highlevel_socket.py
```diff
@@ -352,19 +358,34 @@ class SocketListener(Listener[SocketStream]):
...
+
     """
 
-    def __init__(self, socket: SocketType) -> None:
+    __slots__ = ("path", "socket")
+
+    def __init__(
+        self,
+        socket: SocketType,
+        path: str | bytes | PathLike[str] | PathLike[bytes] | None = None,
```
A5rocks, line 387: I think I preferred how the old version didn't require this `path` argument. Is this really required? I assume we can just check whether a socket is unix type with `socket.family == getattr(tsocket, "AF_UNIX", None)` and have the old logic to get the path.

CoolCat467: Was looking deeper into things noted in #279 and saw this comment: https://github.com/python-trio/trio/issues/279#issuecomment-322114125

CoolCat467: Ah, but I missed something, we only have to care about that if we are renaming socket files, which this implementation does not do, so it would be fine to do the original thing.

A5rocks:
Rereading https://github.com/python-trio/trio/issues/279, now I get why the original PR has the atomic replace. I believe what this PR does is fine, because no atomic replace -> atomic replace is not a breaking change, compared to no unlinking -> unlinking.

I think it would be better to *not* support no downtime redeploys and explicitly document this. Given the complexity of the issue I think @njsmith should look over this of course!

Sxderp:
> I believe what this PR does is fine, because no atomic replace -> atomic replace is not a breaking change (I think?), compared to no unlinking -> unlinking.

It's still compat breaking. I haven't read the code, but I'm assuming in the "before" for both scenarios you error on the path already existing (Address in use, I'd guess). If you start overwriting (via a rename) that's still a change in functionality almost equivalent to unlinking before use.

I think you should figure out the expected pre and post functionality before going stable with it.

- Do you ignore existing socket files and reuse the path (via unlinking or atomic rename)?
- Do you ignore existing files and reuse the path (via unlinking or atomic rename)?
- Do you unlink the socket at the end of the program?
  - PR says yes, but I think it's more common not to? Not really sure honestly. Either way, I think devs have to account for the path existing because the cleanup may not happen on a particularly hard crash.

CoolCat467:
CI is failing because of jedi and I am unsure how to go about fixing it. That particular test is incredibly confusing and tells me nothing about what the error means.

jakkdl:
> CI is failing because of jedi and I am unsure how to go about fixing it. That particular test is incredibly confusing and tells me nothing about what the error means.

it doesn't seem to be just jedi (and mypy), e.g. macOS 3.10 has a bunch of other tests failing. But yeah `test_exports` is not terribly friendly for developers. But tl;dr the test checks what symbols are available with `inspect.getmembers`, and then diffs that against what jedi/mypy/etc sees.
These symbol lists do differ in a bunch of cases, and if possible you figure out why they differ, if that's a problem, and either ignore it in the test (by adding some logic that skips those symbols) or fix it elsewhere.


What seems to be happening here in particular is that skipping `trio.SocketListener.socket` and `trio.SocketStream.socket` are not necessary
https://github.com/python-trio/trio/blob/75cc5dfcb3735cc5c1419900cd01e1fc1cb3008d/src/trio/_tests/test_exports.py#L446-L456

That probably also fixes the mypy error (which *definitely* should have some additional logic/printing what went wrong)

PR #3188, opened 2025-01-19
Switch to use `uv.lock` instead of `test-requirements.txt`

CoolCat467:
In this pull request, we switch the CI system to use `uv.lock` instead of `test-requirements.txt`.


A5rocks:
I don't like this change (because I install the dependencies using `pip install -r test-requirements.txt`!), so I'm a -1 here.

My opinion might shift if there's ever a standardized lockfile format though.

jakkdl:
> I don't like this change (because I install the dependencies using `pip install -r test-requirements.txt`!), so I'm a -1 here.
> 
> My opinion might shift if there's ever a standardized lockfile format though.

presumably one could use `uv sync tests` for the same effect? I think that might be faster and stuff. But afaict the current setup works fairly well so I don't have strong opinions either way - maybe time to revisit #2790 and see which, if any, of the pain points brought up are still a thing.

A5rocks:
Actually, there's a standardized lockfile now. Pip's support for it is in beta, but I figure we can switch to it whenever. In particular, it looks like we can swap `pip compile requirements.in --universal -o requirements.txt` to `uv pip compile requirements.in -o pylock.toml` (I don't know if we can do `test-requirements.in` and `docs-requirements.in` in the same file...)

PR #3190, opened 2025-01-21
Deprecate `trio.__version__`

A5rocks:
Follow up to https://github.com/python-trio/trio/pull/3186

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,6 +13,7 @@ build:
 python:
   install:
     - requirements: docs-requirements.txt
+    - path: .
 
 sphinx:
   fail_on_warning: true
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -315,9 +315,9 @@ author = "Nathaniel J. Smith"
 # built documents.
 #
 # The short X.Y version.
-import trio
+import importlib.metadata
 
-version = trio.__version__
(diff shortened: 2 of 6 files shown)


A5rocks:
I didn't realize that deprecating an attribute means it can't exist as a type hint... If I did I would have just skipped #3186! But oh well.

jakkdl:
it's a shame that https://typing.readthedocs.io/en/latest/spec/directives.html#deprecated (available in typing_extensions) is only meant for classes & functions, but with some messing around it might be possible to make it a `@property`. pyright looks to support it, but not yet mypy.

Also I guess `conf.py` can be a nice citizen and make use of `importlib.metadata` itself.

A5rocks:
> but with some messing around it might be possible to make it a @property

This would require telling the type checker the module got replaced by a class (which I don't think is possible?)

Updating conf.py makes sense though.

oremanj:
This seems unnecessarily churn-y to me and I'm against it. What is the concrete benefit envisioned from removing this attribute? Due to the `dynamic` spec in pyproject.toml, we're already living in the world where we only need to update the version number in one place when making a release, so even fully removing `__version__` won't make that any better than it is today.

A5rocks:
Superseded by https://github.com/python-trio/trio/pull/3194 which doesn't deprecate `trio.__version__`.

PR #3191, opened 2025-01-23
Update type hint for the `CancelStatus.parent` setter

A5rocks:
Type hint issue spotted in https://github.com/python/mypy/pull/18510

Maybe we should wait for that to be merged (and maybe released?) to check this works.

### DIFF
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -378,7 +378,7 @@ class CancelStatus:
         return self._parent
 
     @parent.setter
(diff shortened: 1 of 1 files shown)


TeamSpen210:
Before that PR setter hints were effectively ignored (aside from property implementation), so it should be fine. Could just install the PR code locally to check.

CoolCat467 (approved):
Looks simple enough, don't see any reason why this wouldn't be mergeable at this time even though the mypy PR hasn't gone through yet.

Issue #2757, opened 2023-08-17
Add an `@trio.not_in_async_task` decorator, to error instead of stalling


Zac-HD:
In a large codebase, you're going to have some functions which are slow, and so you don't want them to be called from an async task (without using [`trio.to_thread.run_sync(slow_fn)`](https://trio.readthedocs.io/en/stable/reference-core.html#trio.to_thread.run_sync), anyway).  [`flake8-trio`'s `TRIO200` error](https://github.com/Zac-HD/flake8-trio#warnings-for-blocking-sync-calls-in-async-functions) can help find _direct_ calls, but _indirect_ calls are really hard to find via static analysis.

As [I mentioned in my PyCon talk](https://youtu.be/FrpUb6OEbcc?t=460), and work we've implemented a decorator to mark sync functions which can be called in a sync context, but must be delegated to a worker thread instead of being called in an async context - no matter how many sync functions are in the intermediate call chain.  Simple implementation:

```python
def not_in_async_task(fn):
    """Decorator for sync functions which should not be called in an async context.

    <detailed docstring here>
    """
    @functools.wraps(fn)
    def decorator(*args, **kwargs):
        try:
            # Identical logic to trio.lowlevel.current_task()
            GLOBAL_RUN_CONTEXT.task
        except AttributeError:
            pass
        else:
            raise RuntimeError(
                f"{fn.__name__} should not be called in an async context.  "
                f"Use an async alternative, or trio.to_thread.run_sync({fn.__name__})."
            )
        return fn(*args, *kwargs)

    return decorator
```

Should we include this in `trio` itself?  I've certainly found the pattern useful, and upstreaming it would encourage wider use of the pattern.  On the other hand it's not entirely clear where this would fit, and it's neither hard to implement nor something which benefits much from standardization.

oremanj:
Perhaps the more fundamental primitive here is `trio.in_trio_context()`, which returns a bool. You can write it yourself using only public APIs, but it's weird:
```
def in_trio_context() -> bool:
    try:
        trio.current_time()
    except RuntimeError:
        return False
    return True
```
The private implementation could check for the existence of `GLOBAL_RUN_CONTEXT.runner`.

(There is a subtle difference between checking for `runner` and checking for `task`: the former exists even in Trio context outside of a task, which you can observe in abort_fns and instrument calls. I think the `runner` version is more correct for this need.)

A5rocks:
[...]

PR #3195, opened 2025-01-30
Add `in_trio_run` and `in_trio_task`

A5rocks:
Fixes https://github.com/python-trio/trio/issues/2757

See the documentation and tests for the cases in which these two differ. Feel free to ask me to add more scenarios.


docs/source/reference-lowlevel.rst
```diff
@@ -56,6 +56,35 @@ Global statistics
...
 
 
+.. _trio_contexts:
+
+The current Trio context
+------------------------
+
+There are two different types of contexts in :mod:`trio`. Here are the
+semantics presented as a handy table. Choose the right function for
+your needs.
```
oremanj, line 66:
```suggestion
If you want to interact with an active Trio run -- perhaps you need to
know the :func:`trio.current_time` or :func:`trio.lowlevel.current_task` --
then you need to be "in Trio context", or else you will get a
``RuntimeError("must be called from async context")``.
This requires that you be:

* indirectly inside (and on the same thread as) a call to
  :func:`trio.run`, for run-level parameters such as the
  :func:`~trio.current_time` or :func:`~trio.lowlevel.current_clock`
  
* indirectly inside a Trio task, for task-level parameters such as
  the :func:`~trio.lowlevel.current_task` or
  :func:`~trio.current_effective_deadline`
  
Note that this use of "context" has nothing to do with context variables.
  
Internally, there are thread-local variables tracking the current run and
the current task, which are accessed by the operations that need them.
Sometimes it's useful to be able to tell whether these are set.
To do so, call :func:`trio.lowlevel.in_trio_run` or
:func:`trio.lowlevel.in_trio_task`, which will provide answers according
to the following table.
```
This is subtle so I think it's worthy of a more detailed explanation. See suggestion.

docs/source/reference-lowlevel.rst
```diff
@@ -56,6 +56,35 @@ Global statistics
...
+
++---------------------------------+-----------------------------------+------------------------------------+
+| situation                       | :func:`trio.lowlevel.in_trio_run` | :func:`trio.lowlevel.in_trio_task` |
++=================================+===================================+====================================+
+| inside a running async function | `True`                            | `True`                             |
```
oremanj, line 71: might want to say "Trio-flavored async function", since with guest mode you could have other-loop async functions

docs/source/reference-lowlevel.rst
```diff
@@ -56,6 +56,35 @@ Global statistics
...
++---------------------------------+-----------------------------------+------------------------------------+
+| without a running Trio loop     | `False`                           | `False`                            |
```
oremanj, line 73: Trio generally avoids calling itself an "event loop" in its documentation. Maybe say "in a thread that has no active call to `trio.run()`" here.

docs/source/reference-lowlevel.rst
```diff
@@ -56,6 +56,35 @@ Global statistics
...
++---------------------------------+-----------------------------------+------------------------------------+
+| in a guest run's host loop      | `True`                            | `False`                            |
++---------------------------------+-----------------------------------+------------------------------------+
+| inside an instrument call       | depends                           | depends                            |
```
oremanj, line 77: Can we make `after_run` execute before clearing `runner` so that we can give a consistent "yes" answer to whether instruments execute `in_trio_run`?

docs/source/reference-lowlevel.rst
```diff
@@ -56,6 +56,35 @@ Global statistics
...
++---------------------------------+-----------------------------------+------------------------------------+
+| :func:`trio.to_thread.run_sync` | `False`                           | `False`                            |
```
oremanj, line 79: maybe say "in a worker thread created by"

src/trio/_core/_run.py
```diff
@@ -2952,6 +2952,22 @@ async def checkpoint_if_cancelled() -> None:
     task._cancel_points += 1
 
 
+def in_trio_run() -> bool:
+    """Check whether we are in a Trio run.
+
+    See also :ref:`the different types of contexts <trio_contexts>`.
```
oremanj, line 2958:
```suggestion
    """Check whether we are in a Trio run.
    This is true if and only if :func:`trio.current_time` will succeed.

    See also the discussion of what it means to be :ref:`in "Trio context" <trio_contexts>`.
```
and then analogously for `in_trio_task`

docs/source/reference-lowlevel.rst
```diff
@@ -56,6 +56,35 @@ Global statistics
 .. autoclass:: RunStatistics()
 
 
+.. _trio_contexts:
+
+The current Trio context
```
oremanj, line 61:
```suggestion
Being "in Trio context"
```

A5rocks: I think that wording is somewhat (grammatically??) confusing, though I can't exactly pinpoint why. Trio is perfectly fine as an adjective yet my brain hates "in Trio context"...

docs/source/reference-lowlevel.rst
A5rocks: I'm not sure where we define "Trio-flavored" but at the very least it's clear from context so I'll do it.

docs/source/reference-lowlevel.rst
oremanj: "Trio context" in your version has Trio as an adjective too, FWIW! I think "current Trio context" is a misleading name because the whole point of the section is that there are different things that "Trio context" can refer to, so it's weird to designate a single one of them (run vs task) as "current". But we could go a different direction, such as titling this "Checking for Trio" or something.

docs/source/reference-lowlevel.rst
A5rocks: I've made a few wording changes to this, hopefully it still gets the subtle points / doesn't mislead. (I avoided mentioning context)

docs/source/reference-lowlevel.rst
```diff
@@ -56,6 +56,55 @@ Global statistics
...
+If you want to interact with an active Trio run -- perhaps you need to
+know the :func:`~trio.current_time` or the
+:func:`~trio.lowlevel.current_task` -- then Trio needs to have certain
+state available to it or else you will get a
+``RuntimeError("must be called from async context")``.
+This requires that you either be:
+
+* indirectly inside (and on the same thread as) a call to
+  :func:`trio.run`, for run-level information such as the
+  :func:`~trio.current_time` or :func:`~trio.lowlevel.current_clock`.
```
oremanj, line 73:
```suggestion
  :func:`~trio.current_time` or :func:`~trio.lowlevel.current_clock`; or
```

oremanj (approved):
Looks great, thanks!

Zac-HD (approved):
Thanks @A5rocks!

PR #3196, opened 2025-01-31
add initial tox.ini

### LINKED ISSUE #3170: Replace check.sh and/or ci.sh with tox/nox
> Tox sounds wonderful. I've been meaning to suggest it. I've actually been working on some reusable workflows for tox-based projects + experimented with some cool GHA integrations.
When you get to it, I may offer some help / experimentation in simplifying things here.

_Originally posted by @webknjaz in https://github.com/python-trio/trio/pull/3157#discussion_r1895163518_
           
           
thread contains more discussion on tox vs nox. Whoever wants to Do It can probably just decide on either.

Initial PR does not need to be a huge one that replaces all of check.sh & ci.sh and overhauls CI stuff, an incremental approach should work perfectly fine here.

jakkdl:
I started out on a tox.ini because it's handy to have; it's not complete enough to close #3170, but it should be sufficient for #2699 once the CI is updated to use it.

* there's a bunch missing to completely get rid of ci.sh, I almost suspect we might need `nox` for that not to be a nightmare.
* I probably prefer having this in pyproject.toml, [which tox now supports natively](https://tox.wiki/en/latest/config.html#pyproject-toml-native), but being so used to `tox.ini` I didn't realize this until most of the file was written.
* I haven't switched over anything in `ci.yml` to actually use this. We'd probably want https://github.com/tox-dev/tox-gh or https://github.com/ymyzk/tox-gh-actions to do that properly

everybody's very welcome to push commits to this PR, but I don't think we should block merging it until the tox/nox config is *perfect* (leaving that for followup PRs)- cause I Wanna Use It:tm:


jakkdl:
ugh, native toml not supporting https://tox.wiki/en/latest/config.html#conditional-settings makes it a non-trivial conversion. I think the way to get around that is with more explicit environments + https://tox.wiki/en/latest/config.html#string-substitution-references; but... that's for another day/person

jakkdl:
If you review `ci.sh` with ignoring whitespace: https://github.com/python-trio/trio/pull/3196/files?w=1
you see that it's solely just removing the `CHECK_FORMATTING` logic.

I propose delaying further CI/`ci.sh`/tox-gh stuff for future PR's, as that quickly gets very messy.

tox.ini
```diff
@@ -0,0 +1,92 @@
...
+deps =
+    -r test-requirements.txt
+    exceptiongroup
+base_python = 3.13
+set_env =
+    PYRIGHT_PYTHON_IGNORE_WARNINGS=1
+commands =
+    mypy --platform linux
+    mypy --platform darwin
+    mypy --platform win32
```
CoolCat467, line 79: Shouldn't this use `mypy_annotate`?

CoolCat467 (commented):
I think this mostly makes sense, wondering about mypy runs, and I thought there was some sort of tox github actions thing that was useful for some reason the last time I used tox.

webknjaz:
@jakkdl thanks for pinging! Love that it's being worked on. I'll review shortly :)

webknjaz:
> * I probably prefer having this in pyproject.toml

I'd vote for `tox.toml` since dumping multiple things into one terribly huge config is going to be difficult to maintain.

tox.ini
```diff
@@ -0,0 +1,92 @@
...
+#   * LSP
+#   * apport
+# * use tox in CI
+# * switch to nox?
+# * move to pyproject.toml?
+#   * this means conditional deps need to be replaced
+
+# protip: install uv-tox for faster venv generation
+
+[testenv]
```
webknjaz, line 22: Could you add `description =` to every env? It should explain what they do and would be printed out via `tox l` or `tox -av`.

tox.ini
```diff
@@ -0,0 +1,92 @@
...
+    pytest {env:TOX_RUN_SLOW:} {posargs}
+
+[testenv:no_test_requirements]
+commands =
+    pytest --skip-optional-imports {posargs}
+
+[testenv:docs]
+deps =
+    -r docs-requirements.txt
+# base_python synced with .readthedocs.yml
```
webknjaz, line 41: We can also make RTD just call tox in a follow-up. This is what I do these days.

tox.ini
```diff
@@ -0,0 +1,92 @@
...
+
+[testenv:py39-cython2,py39-cython,py311-cython2,py313-cython]
+deps =
+    cython
+    cython2: cython<3
+    setuptools ; python_version >= '3.12'
+commands =
+    cythonize --inplace -X linetrace=True tests/cython/test_cython.pyx
+    python --version
+    cython --version
```
webknjaz, line 54: These aren't tests, semantically:
```suggestion
commands_pre =
    python --version
    cython --version
commands =
    cythonize --inplace -X linetrace=True tests/cython/test_cython.pyx
```

tox.ini
```diff
@@ -0,0 +1,92 @@
...
+    pre-commit run pip-compile --all-files
+
+[testenv:typing]
+deps =
+    -r test-requirements.txt
+    exceptiongroup
+base_python = 3.13
+set_env =
+    PYRIGHT_PYTHON_IGNORE_WARNINGS=1
+commands =
```
webknjaz, line 76: Let's also add `--python-version=` in follow-ups.

webknjaz:
> and I thought there was some sort of tox github actions thing that was useful for some reason the last time I used tox.

@CoolCat467 there's a plugin: https://github.com/tox-dev/tox-gh. It also demonstrates separating the invocation into preparation and testing. But I think this is something to look into separately, beyond the initial effort. I don't exactly like how the jobs are mixed up right now, but I believe this is out of the scope, so I choose not to start those discussions just yet.

webknjaz (approved):
I think this is ready to be merged and can be iterated on separately. Though, if you have a minute to accept the `commands_pre` thing and add descriptions, it'd be nice to include into the scope of this PR.

tox.ini
jakkdl: hm. Maybe, but `mypy_annotate` is less needed with tox supressing output of succesful commands and in general reformatting the output. One of the problems with `check.sh` was simply finding what it was that failed, which I think is what prompted the creation of it. But we don't want it when running outside of CI, so needs some logic to differentiate.
But if we don't use it here then it should just be removed.

I'll procrastinate on it a little bit, and see what the output looks like in practice. I suspect the end result is going to be removing `mypy_annotate`

.github/workflows/ci.yml
```diff
@@ -276,10 +276,15 @@ jobs:
           python-version: ${{ fromJSON(format('["{0}", "{1}"]', format('{0}.0-alpha - {0}.X', matrix.python), matrix.python))[startsWith(matrix.python, 'pypy')] }}
           cache: pip
           cache-dependency-path: test-requirements.txt
+      - name: Check Formatting
+        if: matrix.check_formatting == '1'
+        run:
+          python -m pip install tox &&
```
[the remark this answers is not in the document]
jakkdl: not sure there's much, if any, speedup when we're only running a single environment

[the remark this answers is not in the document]
webknjaz: I haven't compared them, but I'd imagine that provisioning that env would be faster. You don't separate the provisioning step in this PR, so I can't see how much it takes. But it's probably not important enough right now.

tox.ini
```diff
@@ -0,0 +1,106 @@
...
+#   * apport
+# * use tox in CI
+# * switch to nox?
+# * move to pyproject.toml?
+#   * this means conditional deps need to be replaced
+
+# protip: install uv-tox for faster venv generation
+
+[testenv]
+description = "Base environment for running tests depending on python version."
```
webknjaz, line 23: @jakkdl is configparser able to handle the quotes well? I think this is the first time I see them in this context.. Does the output of `tox l` look good?

jakkdl: ...no :sweat_smile: 
> ❯ tox l
default environments:
py39                 -> "Base environment for running tests depending on python version."
py310                -> "Base environment for running tests depending on python version."
py311                -> "Base environment for running tests depending on python version."
py312                -> "Base environment for running tests depending on python version."

Issue #1559, opened 2020-05-27
Cancelled AsyncResources eat exceptions


oremanj:
This is an especially surprising special case of #455.

This code:
```
with trio.move_on_after(1):
    async with any_asyncresource:
        try:
            await trio.sleep_forever()
        finally:
            raise ValueError
```
will swallow the ValueError, because it gets stuck as the `__context__` of the `Cancelled` exception raised in `any_asyncresource.aclose`. And `aclose` here is basically required to raise `Cancelled`, because Trio rules require it to be a checkpoint.

This showed up in #1555, with trio.Process standing in for "any_asyncresource".

This kind of thing keeps showing up for me and it's *extremely* confusing to track down. I know the party line is "who knows if someone was going to handle that exception that got stuck as the Cancelled `__context__`, we shouldn't reraise it when we swallow the Cancelled".  But I think sometimes raising an exception that the user thought was handled is a much better failure mode than sometimes silently swallowing an exception that no one tried to handle at all.

smurfix:
> the __context__ of the Cancelled exception

Ah. So that's where these vanish to.

I agree, though I do wonder how many false positives a change like that would trigger.

njsmith:
I really don't like the idea of an exception "teleporting" from the place where the resource is closed up to the place where the `move_on_after` is closed.

Instead, let's consider if there's some way to let the original exception keep propagating through the resource `__aexit__`, so it never gets replaced by `Cancelled` in the first place.

You could make an argument that exceptions coming into `AsyncResource.__exit__` should *always* trigger an ungraceful shutdown. That would make this easy to solve:

```python
class AsyncResource:
    async def __aexit__(self, etype, value, tb):
        if etype is not None:
            with trio.CancelScope(shield=True):
                await aclose_forcefully(self)
        else:
            await self.aclose()
```

I'm not quite sure I understand the full implications of this, though. It's definitely less surprising in the case where you have a ambient cancellation ongoing and a non-`Cancelled` exception propagating. But the more common case is where you have non-`Cancelled` exception and *no* ambient cancellation, and I'm not sure if this is what users want or expect there.

Another option would be to say that in `AsyncResource.__aexit__`, we want incoming exceptions to "overwrite" `Cancelled`, but otherwise keep things the same. Maybe

```python
class AsyncResource:
    async def __aexit__(self, etype, value, tb):
        try:
            await self.aclose()
        except Cancelled:
            if etype is not None:
                return  # let original exception propagate instead
            raise
```

[...]

PR #3197, opened 2025-01-31
add `@as_safe_channel`

### LINKED ISSUE #638: Do a better job of communicating the problems with using nurseries/cancel scopes inside generators
I am aware of the issues with cleaning up async iterators, PEP 533 and the need for aclosing. However, if a closing is not used, which I overlook frequently, things can really blow up in your face:

```python
import trio

async def gen_with_task():
        n.start_soon(trio.sleep_forever)
        while True:
            await trio.sleep(1)
            yield 1

async def main():
    async for f in gen_with_task():
        raise ValueError()

trio.run(main)
```

gives me:

``` 
Exception ignored in: <async_generator object gen_with_task at 0x10c7857b8>
RuntimeError: async generator ignored GeneratorExit
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/trio/_core/_run.py", line 1229, in run
    result = run_impl(runner, async_fn, args)
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/trio/_core/_run.py", line 1357, in run_impl
    runner.task_exited(task, final_result)
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/trio/_core/_run.py", line 852, in task_exited
    self.tasks.remove(task)
KeyError: <Task '__main__.main' at 0x10c72c668>

  File "triotest.py", line 14, in <module>
    trio.run(main)
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/trio/_core/_run.py", line 1235, in run
    ) from exc
trio.TrioInternalError: internal error in trio - please file a bug!
Exception ignored in: <coroutine object _AsyncGeneratorContextManager.__aexit__ at 0x10c78f648>
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/async_generator/_util.py", line 84, in __aexit__
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/async_generator/_util.py", line 14, in __aexit__
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/async_generator/_impl.py", line 308, in aclose
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/async_generator/_impl.py", line 277, in athrow
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/async_generator/_impl.py", line 290, in _do_it
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/async_generator/_impl.py", line 197, in __next__
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/async_generator/_impl.py", line 209, in _invoke
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/trio/_core/_run.py", line 311, in open_nursery
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/contextlib.py", line 130, in __exit__
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/trio/_core/_run.py", line 198, in open_cancel_scope
  File "/Users/michael/.pyenv/versions/3.7.0/lib/python3.7/site-packages/trio/_core/_run.py", line 157, in _remove_task
KeyError: <Task '__main__.main' at 0x10c72c668>
```

Note that in addition to the internal trio exceptions, the original cause is not shown at all. This happens when the async generator had a task running in a nursery.

Maybe there is no way to handle this better - I can't tell, this is above my paygrade! But in case there is a way to show a clearer error message, I think it might be worth it.

jakkdl:
Mostly just shepherding @Zac-HD's implementation from https://github.com/python-trio/trio/issues/638#issuecomment-2509452026

I don't understand all the details of it, esp some of the code paths I'm completely failing how to cover, so gonna need some help. And I'm getting an exception that sometimes disappear.. which seems bad?
Feel free to add code suggestions and/or commit directly to the branch.

I don't know if this fully resolves #638, or if there's docs and/or other stuff that should be added.

I will make https://flake8-async.readthedocs.io/en/latest/rules.html#async900 suggest using this, and perhaps make it enabled by default, once released.


src/trio/_tests/test_channel.py
```diff
@@ -411,3 +414,78 @@ async def do_send(s: trio.MemorySendChannel[int], v: int) -> None:
...
+        # this exception sometimes disappear, and I don't know why
+        # gc? trio randomness?
+        # idk if it's gonna show up in CI, but I have like a 50% shot of failing
+        # when running the test case by itself
+        raise ValueError("oae")
+
+    with RaisesGroup(ValueError):
+        async with agen() as recv_chan:
+            async for x in recv_chan:
+                assert x == 1
```
jakkdl, line 487: this is the cause of the CI fail

A5rocks:
I'm not too sold on how some of this is done, but at least now it shouldn't fail CI.

jakkdl:
@Zac-HD if you can show why the inner loop is necessary it'd be great, but I'm kinda suspecting it's a remnant of previous implementations or something - because I can't come up with anything that would hit that code path.

src/trio/_channel.py
```diff
@@ -440,3 +467,88 @@ async def aclose(self) -> None:
...
+            *args: P.args, **kwargs: P.kwargs
+        ) -> AsyncGenerator[trio.MemoryReceiveChannel[T], None]:
+            send_chan, recv_chan = trio.open_memory_channel[T](max_buffer_size)
+            async with trio.open_nursery() as nursery:
+                ait = fn(*args, **kwargs)
+                # nursery.start to make sure that we will clean up send_chan & ait
+                await nursery.start(_move_elems_to_channel, ait, send_chan)
+                # async with recv_chan could eat exceptions, so use sync cm
+                with recv_chan:
+                    yield recv_chan
```
Zac-HD, line 525: Hmm, looking at this again - we want to be sure that we close the channels even if we get an error while starting the generator, which suggests pulling the send channel to the outside:

```suggestion
                with send_chan, recv_chan:
                    nursery.start_soon(_move_elems_to_channel, ait, send_chan)
                    try:
                        yield recv_chan
                    except BaseException:
                        with trio.CancelScope(shield=True) as scope:
                            scope.cancel()
                            await ait.aclose()
                        raise
                    else:
                        await ait.aclose()
```

I also spent a while looking at [the spec](https://peps.python.org/pep-0525/#asynchronous-generator-object) for `await agen.aclose()` too, but in my experiments I couldn't construct a version which misbehaved with our decorator but was OK without it.  (although it's easy enough to get errors both ways!)

The tricky part is that I run into basically the same questions as https://github.com/python-trio/trio/issues/1559#issuecomment-2657850893.  I think here my solution gets around that: we `aclose_forcefully()` if there's any error propagating (and if there's an error during _that_, it should propagate as usual I think?), while if there's no exception yet we do a bare `await ait.aclose()` and allow whatever cleanup logic we might have to run.

Zac-HD (commented):
I _think_ this'll help?  But haven't run many tests yet...

src/trio/_channel.py
```diff
@@ -440,3 +467,88 @@ async def aclose(self) -> None:
...
+                # we're going to use `.asend(err)` to forward errors back to the generator.
+                async for value in agen:
+                    # Get the next value from `agen`; return if exhausted
+                    try:
+                        # Send the value to the channel
+                        await send_chan.send(value)
+                    except trio.BrokenResourceError:
+                        # Closing the corresponding receive channel should cause
+                        # a clean shutdown of the generator.
+                        return
```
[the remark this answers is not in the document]
jakkdl: yeah since your version had `except`s for `BrokenResourceError` and `Cancelled` it's only niche stuff that could get sent. And I don't see any reason why the generator should generate another value after the `send` gets cancelled or broken since it'll just fail again.

Given that `send` has KI protection I don't even think it can raise `KeyboardInterrupt` (unless that gets raised just as the decorator exits? idk details how that works)

src/trio/_channel.py
jakkdl: I don't think there's any problems with not closing the receive channel on starting the generator - if the `nursery.start` call errors then we'll never yield the receive channel to the user, so there's nobody that can get stuck waiting on it

>   Channel objects can be closed by calling `~trio.abc.AsyncResource.aclose`
    or using ``async with``. They are *not* automatically closed when garbage
    collected. Closing memory channels isn't mandatory, but it is generally a
    good idea, because it helps avoid situations where tasks get stuck waiting
    on a channel when there's no-one on the other side. See
    :ref:`channel-shutdown` for details.

jakkdl:
I have no clue why sphinx fails to link the `AbstractAsyncContextManager` in the type hint, when it works perfectly fine for https://trio.readthedocs.io/en/stable/reference-core.html#trio.open_nursery

src/trio/_channel.py
```diff
@@ -440,3 +467,102 @@ async def aclose(self) -> None:
...
+    ) -> Callable[P, AbstractAsyncContextManager[trio.MemoryReceiveChannel[T]]]:
+        @asynccontextmanager
+        @wraps(fn)
+        async def context_manager(
+            *args: P.args, **kwargs: P.kwargs
+        ) -> AsyncGenerator[trio.MemoryReceiveChannel[T], None]:
+            send_chan, recv_chan = trio.open_memory_channel[T](max_buffer_size)
+            async with trio.open_nursery() as nursery:
+                agen = fn(*args, **kwargs)
+                # `nursery.start` to make sure that we will clean up send_chan & ait
```
TeamSpen210, line 536: Typo?
```suggestion
                # `nursery.start` to make sure that we will clean up send_chan & agen
```

src/trio/_channel.py
```diff
@@ -17,9 +22,31 @@
...
+
+    class aclosing(AbstractAsyncContextManager[_SupportsAcloseT, None]):
+        def __init__(self, thing: _SupportsAcloseT) -> None:
+            self._aiter = thing
+
+        async def __aenter__(self) -> _SupportsAcloseT:
+            return self._aiter
+
+        async def __aexit__(self, *exc_info: object) -> None:
+            await self._aiter.aclose()
```
TeamSpen210, line 49: `aclosing` is pretty simple, why not use `try-finally` instead to avoid needing to redefine it and a protocol? We're not exposing any of this publicly. Once we require 3.10 switch to it of course.

src/trio/_channel.py
```diff
@@ -440,3 +467,102 @@ async def aclose(self) -> None:
...
+                agen = fn(*args, **kwargs)
+                # `nursery.start` to make sure that we will clean up send_chan & ait
+                # If this errors we don't close `recv_chan`, but the caller
+                # never gets access to it, so that's not a problem.
+                await nursery.start(_move_elems_to_channel, agen, send_chan)
+                # `async with recv_chan` could eat exceptions, so use sync cm
+                with recv_chan:
+                    yield recv_chan
+                # Return promptly, without waiting for the generator to yield the
+                # next value
```
TeamSpen210, line 544: This comment was a little confusing to read, we're not returning anything. How about something like "User has exited the context manager, cancel to immediately close the abandoned generator if it's still alive."

TeamSpen210 (approved):
Looks good and useful.

This is unrelated, but while we're editing the async-gen docs, we might want to swap out the `async_generator.aclosing` reference to point to `contextlib` first, with `async_generator` as a backport.

src/trio/_channel.py
CoolCat467: Might be referring to "async iterable"? In either case, using full name instead of acronym would be better in my opinion.

docs/source/reference-io.rst
```diff
@@ -181,6 +181,15 @@ Abstract base classes
 .. currentmodule:: trio
 
 
+Converting Async Generators to use streams
```
A5rocks, line 184: I mean.... I'm not so sure this should be in the `reference-io` file? Like yes, it does use memory channels which is a stream which is I/O, but that's really convoluted. How about updating `reference-core` instead -- maybe even include the function docs instead of linking!

newsfragments/3197.feature.rst
```diff
@@ -0,0 +1 @@
+Add :func:`@trio.background_with_channel <trio.background_with_channel>`, a wrapper that can be used to make async generators safe. See :ref:`async-generators`, `ASYNC900 <https://flake8-async.readthedocs.io/en/latest/rules.html#async900>`_, :pep:`789`, and :pep:`533`.
```
A5rocks, line 1: Maybe add a short summary of why each of these links are relevant instead of just having four in a row. Similar to:

```suggestion
Add :func:`@trio.background_with_channel <trio.background_with_channel>`, a wrapper that can be used to make async generators safe. See :ref:`a review on why they are unsafe <async-generators>`. This addition allows for straightforward recommendations for `ASYNC900 <https://flake8-async.readthedocs.io/en/latest/rules.html#async900>`_ and :pep:`789`.
```

(yes I cut out 533)

src/trio/_tests/test_channel.py
A5rocks, line 1: Could we test a few more cases:
 - go over a background channel in a nested loop to make sure no state is shared (it isn't but might be good as a precaution)
 - does `try: ... finally: ...` work in the async generator? What if I throw an error in the `finally`? I expect it'll overwrite the raised error but would be good to have guarantees in the tests.
 - does raising in the iteration over the channel throw that same error into the `agen`? (I don't think so?)

A5rocks (approved):
The code itself looks good!

src/trio/_channel.py
Zac-HD: variable got renamed at some point, updating to `agen` makes sense 👍

newsfragments/3197.feature.rst
jakkdl: the link to `async-generators` isn't really needed after the `autofunction` is moved to that section, and both it and async900 (now) link to pep789, so we can drop that too

docs/source/reference-io.rst
jakkdl: yeah that fits better

src/trio/_tests/test_channel.py
jakkdl: hm, messing around with this made me notice that this implementation does not throw `GeneratorExit()` into the generator when exiting the consumer side early - something that a normal async gen would do

docs/source/reference-core.rst
```diff
@@ -1737,7 +1744,8 @@ so sometimes you'll get an unhelpful `TrioInternalError`. (And
 sometimes it will seem to work, which is probably the worst outcome of
 all, since then you might not notice the issue until you perform some
 minor refactoring of the generator or the code that's iterating it, or
-just get unlucky. There is a `proposed Python enhancement
+just get unlucky. There is a draft :pep:`798` with accompanying
```
A5rocks, line 1747:
```suggestion
just get unlucky. There is a draft :pep:`789` with accompanying
```

jakkdl:
does anybody have any idea about the RTD fail?

> I have no clue why sphinx fails to link the `AbstractAsyncContextManager` in the type hint, when it works perfectly fine for https://trio.readthedocs.io/en/stable/reference-core.html#trio.open_nursery

TeamSpen210:
Found the RTD problem. It's because of `TYPE_CHECKING`. In `open_nursery()`, the return annotation is directly `AbstractAsyncContextManager[]`, so `inspect.signature() -> typing.get_type_hints()` can resolve it fine. But for `background_with_channel`, `Callable` and the paramspec don't exist so it fails. 

Because the paramspec is used as a `Callable` argument, we really have to use a real one `if "sphinx" in sys.modules`...

jakkdl:
> Found the RTD problem. It's because of `TYPE_CHECKING`. In `open_nursery()`, the return annotation is directly `AbstractAsyncContextManager[]`, so `inspect.signature() -> typing.get_type_hints()` can resolve it fine. But for `background_with_channel`, `Callable` and the paramspec don't exist so it fails.
> 
> Because the paramspec is used as a `Callable` argument, we really have to use a real one `if "sphinx" in sys.modules`...

huh. In theory I think this *should* work because we have `SPHINX_AUTODOC_RELOAD_MODULES=1` to reload with `TYPE_CHECKING=True` in `docs/source/conf.py`, but maybe that's hitting another problem.

TeamSpen210:
It's not `typing_extensions` not being installed, the error is reproducible locally and I certainly have it installed. I was tracing through the code, `SPHINX_AUTODOC_RELOAD_MODULES` doesn't affect this, it only triggers if it couldn't import the function/class/etc itself.

A5rocks:
It's not clear to me whether this is intended to be a drop-in replacement for iterating over an async generator that replicates behavior. The documentation makes it sound that way, but:
 - this raises `ExceptionGroup`?
 - https://github.com/python-trio/trio/pull/3197#discussion_r1958482422

I would rather this is drop-in even if it needs to be made more complicated.

src/trio/_channel.py
```diff
@@ -440,3 +454,105 @@ async def aclose(self) -> None:
...
+                # never gets access to it, so that's not a problem.
+                await nursery.start(_move_elems_to_channel, agen, send_chan)
+                # `async with recv_chan` could eat exceptions, so use sync cm
+                with recv_chan:
+                    yield recv_chan
+                # User has exited context manager, cancel to immediately close the
+                # abandoned generator if it's still alive.
+                nursery.cancel_scope.cancel()
+
+        return context_manager
```
A5rocks, line 534: Something I accidentally ran into while testing this is I forgot to add `()` to `@background_with_channel`. Could we raise an error in that case and stop people from making my mistake? I got confused because otherwise the error happens when calling the iterator:
```py
>>> trio.run(main)
  File "<python-input-5>", line 1, in <module>
    trio.run(main)
    ~~~~~~~~^^^^^^
  File "C:\Users\A5rocks\Documents\trio\src\trio\_core\_run.py", line 2423, in run
  File "<python-input-4>", line 2, in main
    async with behavior() as it:
               ~~~~~~~~^^
TypeError: background_with_channel.<locals>.decorator() missing 1 required positional argument: 'fn'
```

src/trio/_tests/test_channel.py
jakkdl: so yeah what's going on here is the extra buffer from interleaved execution - the `agen` will always be one yield ahead, and the `GeneratorExit` will be thrown into it at the second yield if the consumer exits while handling the first `yield`.

I'm pretty sure we need a wrapper around the `receive_channel` for this, so `_move_elems_to_channel` can block until it's possible to send.

jakkdl:
> It's not clear to me whether this is intended to be a drop-in replacement for iterating over an async generator that replicates behavior. The documentation makes it sound that way, but:
> 
> * this raises `ExceptionGroup`? I get this makes sense in the (hopefully rare) case that both the `async for` body and the async iterator raise errors, but maybe we could `raise from_iter from from_body` and hope that makes sense? I've kind of talked myself into thinking that raising `ExceptionGroup` is a good thing but that would require documenting this. 

With a buffer size > 0 I don't even think it's all that rare to get multiple errors. But unless we want to remove that functionality I think it's correct to stick to groups. This might make people hesitant to use it as a drop-in-replacement and if that's explicitly what we're targeting we could maybe add a kwarg or an additional decorator that handles it. (and, uh, given how messy it is to convert exceptiongroups into single exceptions maybe that's a thing we want separately?)

> (and hopefully explicitly passing `strict_exception_groups=True` to the `open_nursery` within.)

good catch, done!

> * [add `@background_with_channel` #3197 (comment)](https://github.com/python-trio/trio/pull/3197#discussion_r1958482422)

I thought this was gonna be complicated, but turned out not too bad with the addition of a wrapper + semaphore.

> I would rather this is drop-in (at least for the above) even if it needs to be made more complicated. Though I don't care for less visible things like "does it run in the task or in another task if there's no buffer".
> 
> For example: [...]

I'm not even sure if it's possible to run in the same task and do it correctly? We could in theory make it look like it from the traceback, but not sure that's worth it

Zac-HD (approved):
I'd be happy to merge this now, or after making one medium-sized design change.

Specifically, I think @A5rocks has a good point about raising `ExceptionGroup`.  If we've got `max_buffer_size=0` (which is the default!), then our semaphore ensures that we _can't_ be running user code in the generator concurrently with that in the body of the context manager.  I think this is a rare good reason to unwrap the knowably-just-one exception in a group raised by the inner nursery - and that the smaller change relative to raw generators in the default case is a larger benefit than the cost of raising groups or not based on the max_buffer_size argument.

Thoughts?

jakkdl:
I think if we're going towards single exceptions we should maybe never raise exceptiongroups, and either do `raise RuntimeError("multiple errors") from exc_group` or stuff the exception from the generator in `__context__` (or the other way round?). I'm not sure, all of them sound kinda iffy.

Actually, I think the cleanest is to expose two different decorators:
```python
def background_with_channel():
  ...
def background_with_buffer_channel(buffer: int|None):
  """[...] raises ExceptionGroup [...]"""
  ...
```
This also allows us to make interleave vs not much more explicit, as opposed to suggesting that anybody can start buffering any generator without any change in behavior around `.throw()` etc

Zac-HD:
Let's merge as-is then.

I don't want to split into multiple functions (linter recs etc get harder), we've already disabled `.throw()` by handing the user a channel instead of a generator object, and imo consistency with the obvious way to do similar things downstream is pretty nice.

jakkdl:
sorry, with throwing I meant https://github.com/python-trio/trio/pull/3197#discussion_r1958482422

for linter recs we'd just always recommend `background_with_channel` - and then in its docstring we'd mention the existence of the buffered version in case you want that

Zac-HD:
https://github.com/python-trio/trio/pull/3197#discussion_r1958482422 doesn't bother me, because cancellation is always a possibility you have to handle.  It "just so happens" that we always cancel "right before" the `GeneratorExit` would show up 😉 

In practice most users are never going to touch the default, so maybe we should _only_ supply the unbuffered version, and let power-users add their own buffering if desired?  We could even share a recipe for the "`await send_chan.send(x)` instead of `yield x`" version, which is quite simple and avoids all the complexity of having a generator object in there...

jakkdl:
> In practice most users are never going to touch the default, so maybe we should _only_ supply the unbuffered version, and let power-users add their own buffering if desired? We could even share a recipe for the "`await send_chan.send(x)` instead of `yield x`" version, which is quite simple and avoids all the complexity of having a generator object in there...

Okay yeah this seems fine. The only time this would be insufficient is if they're relying on an external asyncgen they can't modify *and* they want buffering, but if so they can still just write an equivalent version of `buffer_it`.


But god I am not looking forward to another go at the "oh just unwrap the exception from inside the group"; I'm tempted to go create a helper for that in a new PR, which would be pretty great for trio-websocket and probably others as well: https://github.com/python-trio/trio-websocket/blob/bec4232178700e53dccb887d028997f6746e91de/trio_websocket/_impl.py#L218

Zac-HD:
OK, I think this is the last round of comments from me:

- Yeah, let's make the `RecvChanWrapper` clone-able.  I don't have a use-case in mind but would prefer to avoid gaps vs the memory-channel interface.
- Needs changelog entry etc., since I think it's worth releasing this after we merge (and then updating linters etc.)
- Now that there's no buffering, I think `@trio.background_with_channel` makes less sense as a name.  Thoughts?
  - Current ideas: `@trio.safe_(async_)?generator`, `@trio.generator(_to_|2|_as_)channel`, variations.  Prior art is named [`trio_async_generator`](https://github.com/groove-x/trio-util/blob/master/src/trio_util/_trio_async_generator.py), and includes buffering (size-zero channel, no semaphore).
  - "background" is just inaccurate now and so I'd fairly strongly prefer to rename it.
- Documentation, including a recipe - I don't think makes sense to ship this, but I want to encourage people to use it if they want buffering + as a generally nicer API design pattern than requiring the caller to open channels and a nursery:
   ```python
   to_channel(max_buffer_size: int | Literal[inf]) -> Callable[
       Callable[Concat[MemorySendChannel[T], P], None], 
       Callable[P, AbstractAsyncContextManager[MemoryReceiveChannel[T]]]
   ]
   ```

TeamSpen210 (approved):

Zac-HD (changes requested):
Sorry everyone - I just pulled on a related thread and think I've worked out a simpler design.  Let's hold off on merging, and I'll update in a few days whether it worked or not 🙏

Zac-HD:
false alarm, the simple idea leaks cancellations from the generator into the caller (cf PEP-789) 😞

Zac-HD (approved):
false alarm, the simple idea leaks cancellations from the generator into the caller (cf PEP-789) 😞

A5rocks:
Could we add a test for that in this PR too at least

Zac-HD:
```python
@generator_as_channel
async def agenfn():
    with trio.CancelScope() as cscope:
        cscope.cancel()
        yield
        
async def test_doesn't_leak_cancellation():
    with pytest.raises(AssertionError):
        async with agenfn() as recv_chan:
            async for _ in recv_chan:
                pass
        raise AssertionError("should be reachable")
```

src/trio/_tests/test_channel.py
```diff
@@ -411,3 +414,166 @@ async def do_send(s: trio.MemorySendChannel[int], v: int) -> None:
...
+            stuff.append(repr(e))
+            raise
+        finally:
+            stuff.append("finally")
+            raise ValueError("agen")
+
+    with pytest.raises(
+        RuntimeError,
+        match=r"^Encountered exception during cleanup of generator object, as well as exception in the contextmanager body.$",
+    ) as excinfo:
```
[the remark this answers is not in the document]
Zac-HD: no suggestions, sorry 😕

Zac-HD (commented):
Three things below - I'm very keen to merge and thus deferred `.clone()`, but I also think we need to get the core interface right the first time 😅 

- The big-but-easy one is renaming the function `@as_safe_channel`; see below for why
- We can drop some docs now that the interface is simpler 🎉 
- Two small but important correctness changes.

(once more into the diff, dear friends!)

src/trio/_channel.py
```diff
@@ -440,3 +462,132 @@ async def aclose(self) -> None:
...
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        self.recv_chan.close()
+
+
+def background_with_channel(
```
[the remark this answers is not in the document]
TeamSpen210: Perhaps it should mention generators also, like `@agen_as_channel`. Don't really like `safe` though, could mean many things, (are we safe from SQL injection? :) and would be very problematic if we find another issue in the future making this not "safe".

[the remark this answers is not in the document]
Zac-HD: I considered this, but I want it to be immediately clear that it's not just a change to calling convention.  I'm also pretty confident that we won't find another such issue (this one has been known for about seven years!), and expect that readers will realize they should look up what it makes agens safe from.  I did look at more-descriptive names but they all got awkwardly long.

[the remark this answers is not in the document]
TeamSpen210: It just feels sorta wrong to have `safe` in a function name, like doing `html_escape_please_use_me` or something. It's not describing how the function operates, it's an advertisement to the user.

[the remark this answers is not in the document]
Zac-HD: I think this _does_ describe how the function operates though; because the obvious way to wrap an async generator into a channel is _not_ safe - so if I saw e.g. `@as_channel`, I'd assume that it changes the return type without changing the semantics.

src/trio/_channel.py
```diff
@@ -440,3 +462,132 @@ async def aclose(self) -> None:
...
+                )
+                # `async with recv_chan` could eat exceptions, so use sync cm
+                with RecvChanWrapper(recv_chan, send_semaphore) as wrapped_recv_chan:
+                    yield wrapped_recv_chan
+                # User has exited context manager, cancel to immediately close the
+                # abandoned generator if it's still alive.
+                nursery.cancel_scope.cancel()
+        except BaseExceptionGroup as eg:
+            try:
+                raise_single_exception_from_group(eg)
```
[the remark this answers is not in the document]
jakkdl: oh yes, you're totally right

src/trio/_channel.py
```diff
@@ -440,3 +462,132 @@ async def aclose(self) -> None:
...
+
+class RecvChanWrapper(ReceiveChannel[T]):
+    def __init__(
+        self, recv_chan: MemoryReceiveChannel[T], send_semaphore: trio.Semaphore
+    ) -> None:
+        self.recv_chan = recv_chan
+        self.send_semaphore = send_semaphore
+
+    # TODO: should this allow clones? We'd signal that by inheriting from
+    # MemoryReceiveChannel.
```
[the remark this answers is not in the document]
jakkdl: I was mostly thinking if we have any strong reasons *not* to offer it, implementing it should be straightforward.
Though looking at the interface of `MemoryReceiveChannel` we also don't have `receive_nowait` on the wrapper and I'm not seeing how that one would make sense to offer - so let's stick to the interface of `ReceiveChannel`

src/trio/_channel.py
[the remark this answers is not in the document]
jakkdl: though we don't need to handle `__suppress_context__`, it only gets set if you modify `__cause__` by using `raise ... from ...` https://peps.python.org/pep-0415/

src/trio/_channel.py
```diff
@@ -440,3 +462,132 @@ async def aclose(self) -> None:
...
+
+    async def receive(self) -> T:
+        self.send_semaphore.release()
+        return await self.recv_chan.receive()
```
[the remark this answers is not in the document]
jakkdl: > I think we get back into "the generator is running concurrently with the iterating task"
Ah, hm, in a way I suppose. If task A & task B call `receive` on the same channel, after the generator has emitted a value for task A then task A can run concurrently with the generator constructing a value for task B.

But adding a lock doesn't change that behavior, task A can still continue running while handling the value from the generator while task B is waiting for its value.

But in neither case does the generator run concurrent with a caller while generating a value for *that* caller, so I don't think there's any issues.


Adding a test case, and renaming the attributes :)

src/trio/_channel.py
```diff
@@ -440,3 +462,132 @@ async def aclose(self) -> None:
...
+                    await send_semaphore.acquire()
+                    try:
+                        value = await agen.__anext__()
+                    except StopAsyncIteration:
+                        return
+                    # Send the value to the channel
+                    await send_chan.send(value)
+            finally:
+                # replace try-finally with contextlib.aclosing once python39 is dropped
+                await agen.aclose()
```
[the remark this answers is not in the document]
jakkdl: well, only if the end-user has written a broken `__aexit__` - in which the problem is on their end :stuck_out_tongue: To get that with a `yield`-based generator I think they'd have to break [ASYNC102](https://flake8-async.readthedocs.io/en/latest/rules.html#async102) though and do `await` in a `finally` themselves.

For `with send_chan` we *know* that `AsyncResource` has a bad `__aexit__` (#1559), and also provides a sync contextmanager, so it's easy to work around.

In theory we could add a shielded+timeout cancel scope a la https://flake8-async.readthedocs.io/en/latest/rules.html#async102, but I think that introduces more potential problems than it solves.

Zac-HD (approved):
This looks great to me!

- tiny tweaks for changelog and coverage below
- wait on https://github.com/python-trio/trio/pull/3249 for the unrelated test failures

then let's ship it!

jakkdl:
Okay I added a testcase for multiple cancellations, and current implementation will break code that does `try: ...; except Cancelled(): ...`, so I'll change `raise_single_exception_from_group` to only unwrap one layer of a group (and then handle multi-cancellation/multi-keyboardinterrupt) and make use of it

jakkdl:
turns out this shit is complicated at times :sweat_smile: 

But yes thank you all for initial design, reviews, and other feedback!

Issue #2764, opened 2023-08-19
Clear global THREAD_CACHE in child after fork()


tmaxwell-anthropic:
Consider the following bug:
1. `trio.to_thread.run_sync()` is called, creating a worker thread. The worker thread is left in the global `THREAD_CACHE`.
2. The Python process forks for some reason (perhaps via the `multiprocessing` module)
3. The child process now calls `trio.to_thread.run_sync()`. The global `THREAD_CACHE` still contains a reference to the worker thread, so the child process thinks it has an idle worker thread, and tries to dispatch a task to it. However, the worker thread doesn't actually exist in the child process. So `trio.to_thread.run_sync()` hangs forever.

Because `THREAD_CACHE` is interpreter-global, this can happen even if the two Trio run loop are completely separate. For example, in a test suite, one test might call `trio.to_thread.run_sync()`, and then later a completely separate test might use `multiprocessing` to spawn a process that calls `trio.to_thread.run_sync()`.

I think it should be fairly simple to fix this by using `os.register_at_fork()` to ensure `THREAD_CACHE` is cleared in the child whenever the interpreter forks.

TeamSpen210:
There's some previous discussion regarding handling `fork()` in #1614, though a bunch of the concerns there probably aren't relevant anymore...

tmaxwell-anthropic:
Note this can happen even if the fork isn't inside `trio.run()`. For example:
```
async def foo():
  ... do some async stuff ...
  await trio.to_thread.run_sync(...)
  ... do some async stuff ...

trio.run(foo)
# we are now outside trio.run(), but the worker thread is still in THREAD_CACHE

os.fork()

# this is fine on the parent, but fails on the child, due to the bug
trio.run(foo)
```
My understanding is it's not really practical to support `fork()` inside `trio.run()`, but it generally should be find to have `fork()` and `trio.run()` in the same program at different times -- except for this bug :)

A5rocks:
Do you want to contribute a fix? Otherwise I might try to fix this (though I'm not very comfortable with things related to threads...)

jakkdl:
> Do you want to contribute a fix? Otherwise I might try to fix this (though I'm not very comfortable with things related to threads...)

I don't think @tmaxwell-anthropic is going to, do you want to do it? Otherwise I'll give it a try (and I'm certainly not comfortable with thread things either :P)

jakkdl:
[...]

PR #3200, opened 2025-02-05
Clear worker thread cache after forking

A5rocks:
Fixes https://github.com/python-trio/trio/issues/2764

While we don't provide any sort of guarantees and I think it's tempting to give up on this:
 - Python (3.12+) will already warn on this, so they have that handled
 - It's friendlier to do the right thing.

As long as people aren't initializing their own thread caches, this should be fine.

### DIFF
--- a/src/trio/_core/_tests/test_thread_cache.py
+++ b/src/trio/_core/_tests/test_thread_cache.py
@@ -200,6 +200,8 @@ def test_raise_in_deliver(capfd: pytest.CaptureFixture[str]) -> None:
 
 @pytest.mark.skipif(not hasattr(os, "fork"), reason="os.fork isn't supported")
 def test_clear_thread_cache_after_fork() -> None:
+    assert hasattr(os, "fork")
+
     def foo() -> None:
         pass
 
@@ -217,4 +219,4 @@ def test_clear_thread_cache_after_fork() -> None:
 
     if child_pid != 0:
         # if this test fails, this will hang, triggering a timeout.
-        os.wait4(child_pid, 0)
+        os.waitpid(child_pid, 0)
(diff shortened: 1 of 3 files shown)


A5rocks:
Maybe there should be a way to tell Trio "hey I don't want the worker threads anymore, please kill them!", especially in light of the new deprecation warning (ATM the way to do that is to do `time.sleep(10)`). Maybe Trio should explicitly support this use case and set up a callback for before `os.fork()` s.t. all the threads get killed.

That's not what this PR does!

jakkdl:
coverage fail appears to be https://github.com/nedbat/coveragepy/issues/310

src/trio/_core/_tests/test_thread_cache.py
```diff
@@ -195,3 +196,39 @@ def deliver(_: object) -> NoReturn:
...
+
+    if child_pid != 0:
+        # if this test fails, this will hang, triggering a timeout.
+        os.waitpid(child_pid, 0)
+    else:
+        try:
+            import coverage
+
+            cov = coverage.Coverage.current()
+            if cov:  # pragma: no branch
```
jakkdl, line 228: can add a comment+reference to the coverage.py issue why the pragmas are "needed"

jakkdl (changes requested):

CoolCat467 (approved):
Not sure if I completely understand the test for this, but the actual functionality changes look fine.

src/trio/_core/_tests/test_thread_cache.py
CoolCat467: +1

jakkdl (approved):
unclear if the coverage stuff is worth it, though I guess we do get confirmation that the `else` branch is executed even though the coverage itself is only on the coverage stuff.

very good comments, would have been a pain to figure out those details otherwise :)

A5rocks:
The coverage stuff definitely isn't worth it for the test coverage, but doing this stop/save thing also covers the callback from register_at_fork. Otherwise that doesn't get any coverage.

PR #3201, opened 2025-02-05
Update to mypy 1.15

tjstum:
In https://github.com/python/typeshed/pull/13372, the type of the stdlib's `socket.getaddrinfo` was updated to add a union member for cases where Python was compiled with `--disable-ipv6`. This causes custom resolvers that use the stdlib function to fail mypy

This is probably going to (mypy) break everyone's custom `HostnameResolver`, since the parent type is changing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,7 @@ repos:
     hooks:
       - id: codespell
         additional_dependencies:
+          # tomli needed on 3.10. tomllib is available in stdlib on 3.11+
           - tomli
   - repo: https://github.com/crate-ci/typos
     rev: typos-dict-v0.12.4
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,8 @@ repos:
     rev: v2.4.1
     hooks:
       - id: codespell
+        additional_dependencies:
+          - tomli
   - repo: https://github.com/crate-ci/typos
     rev: typos-dict-v0.12.4
     hooks:
--- a/newsfragments/3201.misc.rst
+++ b/newsfragments/3201.misc.rst
@@ -3,3 +3,8 @@ match that of the stdlib `socket.getaddrinfo`, which was updated in mypy 1.15 (v
 a typeshed update) to include the possibility of ``tuple[int, bytes]`` for the
 ``sockaddr`` field of the result. This happens in situations where Python was compiled
 with ``--disable-ipv6``.
+
+Additionally, when using mypy 1.15, the static typing of
+:func:`trio.to_thread.run_sync`, :func:`trio.from_thread.run` and
+:func:`trio.from_thread.run_sync` has been improved and should reflect the underlying
+function being run.
--- a/newsfragments/3201.misc.rst
+++ b/newsfragments/3201.misc.rst
@@ -1,0 +1,5 @@
+The typing of :func:`trio.abc.HostnameResolver.getaddrinfo` has been corrected to
+match that of the stdlib `socket.getaddrinfo`, which was updated in mypy 1.15 (via
+a typeshed update) to include the possibility of ``tuple[int, bytes]`` for the
+``sockaddr`` field of the result. This happens in situations where Python was compiled
+with ``--disable-ipv6``.
--- a/newsfragments/3201.misc.rst
+++ b/newsfragments/3201.misc.rst
@@ -4,7 +4,6 @@ a typeshed update) to include the possibility of ``tuple[int, bytes]`` for the
 ``sockaddr`` field of the result. This happens in situations where Python was compiled
 with ``--disable-ipv6``.
 
-Additionally, when using mypy 1.15, the static typing of
-:func:`trio.to_thread.run_sync`, :func:`trio.from_thread.run` and
-:func:`trio.from_thread.run_sync` has been improved and should reflect the underlying
-function being run.
+Additionally, the static typing of :func:`trio.to_thread.run_sync`,
+:func:`trio.from_thread.run` and :func:`trio.from_thread.run_sync` has been
+improved and should reflect the underlying function being run.
--- a/src/trio/_core/_instrumentation.py
+++ b/src/trio/_core/_instrumentation.py
@@ -10,17 +10,14 @@ from .._abc import Instrument
 INSTRUMENT_LOGGER = logging.getLogger("trio.abc.Instrument")
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Sequence
+    from collections.abc import Sequence
 
-    from typing_extensions import ParamSpec
-
-    P = ParamSpec("P")
     T = TypeVar("T")
 
 
 # Decorator to mark methods public. This does nothing by itself, but
 # trio/_tools/gen_exports.py looks for it.
-def _public(fn: Callable[P, T]) -> Callable[P, T]:
+def _public(fn: T) -> T:
     return fn
 
 
--- a/src/trio/_core/_instrumentation.py
+++ b/src/trio/_core/_instrumentation.py
@@ -2,23 +2,25 @@ from __future__ import annotations
 
 import logging
 import types
-from collections.abc import Callable, Sequence
-from typing import TypeVar
+from typing import TYPE_CHECKING, TypeVar
 
 from .._abc import Instrument
 
 # Used to log exceptions in instruments
 INSTRUMENT_LOGGER = logging.getLogger("trio.abc.Instrument")
 
+if TYPE_CHECKING:
+    from collections.abc import Callable, Sequence
 
-# Explicit "Any" is not allowed
-F = TypeVar("F", bound=Callable[..., object])  # type: ignore[misc]
+    from typing_extensions import ParamSpec
+
+    P = ParamSpec("P")
+    T = TypeVar("T")
 
 
 # Decorator to mark methods public. This does nothing by itself, but
 # trio/_tools/gen_exports.py looks for it.
-# Explicit "Any" is not allowed
-def _public(fn: F) -> F:  # type: ignore[misc]
+def _public(fn: Callable[P, T]) -> Callable[P, T]:
     return fn
 
 
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -7,7 +7,6 @@ import itertools
 import random
 import select
 import sys
-import types
 import warnings
 from collections import deque
 from contextlib import AbstractAsyncContextManager, contextmanager, suppress
@@ -58,6 +57,7 @@ if sys.version_info < (3, 11):
 
 if TYPE_CHECKING:
     import contextvars
+    import types
     from collections.abc import (
         Awaitable,
         Callable,
@@ -1886,8 +1886,7 @@ class Runner:  # type: ignore[explicit-any]
                 return await orig_coro
 
             coro = python_wrapper(coro)
-        assert isinstance(coro, types.CoroutineType)
-        assert coro.cr_frame is not None, "Coroutine frame should exist"
+        assert coro.cr_frame is not None, "Coroutine frame should exist"  # type: ignore[attr-defined]
 
         ######
         # Set up the Task object
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -61,7 +61,6 @@ if TYPE_CHECKING:
     from collections.abc import (
         Awaitable,
         Callable,
-        Coroutine,
         Generator,
         Iterator,
         Sequence,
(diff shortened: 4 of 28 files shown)


A5rocks:
I guess this PR should be more generally "update to mypy 1.15"...

tjstum:
OK, I'll take a look at the other typing changes needed for 1.15 as well

jakkdl:
looks like we need to replace a lot of `type: ignore[misc]` with `type: ignore[explicit-any]`

src/trio/_core/_run.py
```diff
@@ -1895,7 +1886,7 @@ async def python_wrapper(orig_coro: Awaitable[RetT]) -> RetT:
                 return await orig_coro
 
             coro = python_wrapper(coro)
-        assert coro.cr_frame is not None, "Coroutine frame should exist"
+        assert coro.cr_frame is not None, "Coroutine frame should exist"  # type: ignore[attr-defined]
```
A5rocks, line 1889: You could add an `assert isinstance(coro, types.CoroutineType)` above this to remove this type ignore

src/trio/_core/_tests/test_asyncgen.py
```diff
@@ -68,20 +71,20 @@ async def async_main() -> None:
         # No problems saving the geniter when using either of these patterns
         aiter_ = example("exhausted 3")
         try:
-            saved.append(aiter_)
+            saved.append(cast("AGT", aiter_))
```
A5rocks, line 74: IMO an `isinstance` check later is better than a `cast` here.

... I assume `example` can't return `AsyncGeneratorType[int, None]`?

src/trio/_file_io.py
```diff
@@ -428,7 +428,7 @@ async def open_file(
 
 
 @overload
-async def open_file(  # type: ignore[misc]  # Any usage matches builtins.open().
+async def open_file(  # type: ignore[explicit-any,misc]  # Any usage matches builtins.open().
```
A5rocks, line 431:
```suggestion
async def open_file(  # type: ignore[explicit-any, misc]  # Any usage matches builtins.open().
```
(nitpick of course)

src/trio/_tests/test_socket.py
```diff
@@ -186,7 +186,10 @@ def interesting_fields(
         ) -> tuple[
             AddressFamily,
             SocketKind,
-            tuple[str, int] | tuple[str, int, int] | tuple[str, int, int, int],
+            tuple[str, int]
+            | tuple[str, int, int]
```
A5rocks, line 190: Actually I don't think this `tuple[str, int, int]` is necessary?

src/trio/_tests/test_socket.py
```diff
@@ -198,7 +201,10 @@ def filtered(
             tuple[
                 AddressFamily,
                 SocketKind,
-                tuple[str, int] | tuple[str, int, int] | tuple[str, int, int, int],
+                tuple[str, int]
+                | tuple[str, int, int]
```
A5rocks, line 205: If above `tuple[str, int, int]` is unnecessary then so is this.

src/trio/_threads.py
```diff
@@ -252,9 +254,9 @@ def run_in_system_nursery(self, token: TrioToken) -> None:
 
 
 @enable_ki_protection  # Decorator used on function with Coroutine[Any, Any, RetT]
```
A5rocks, line 256: I assume this comment was just wrong and can be removed? (It reads like explaining the `# type: ignore[misc]`)

A5rocks (commented):
I actually just was looking at doing this and thinking through what I would prefer, but I actually prefer the choices you made here (e.g. pushing up the `CoroutineType` change rather than littering code with `assert isinstance(task.coro, CoroutineType)`)

A couple comments, but most are very nitpicky so feel free to ignore.

src/trio/_core/_tests/test_asyncgen.py
tjstum: > `example` can't return `AsyncGeneratorType[int, None]`?

Nope! You get a different mypy error:
`trio/_core/_tests/test_asyncgen.py:22: error: The return type of an async generator function should be "AsyncGenerator" or one of its supertypes  [misc]`

> an `isinstance` check later

I tried this and got an error, but upon trying again, I no longer get an error. Not sure what happened the first time. Thanks!

src/trio/_tests/test_socket.py
tjstum: Indeed, thanks!

src/trio/_threads.py
tjstum: Yeah, wasn't 100% clear at first, but I think you're right

.pre-commit-config.yaml
```diff
@@ -34,6 +34,8 @@ repos:
     rev: v2.4.1
     hooks:
       - id: codespell
+        additional_dependencies:
+          - tomli
```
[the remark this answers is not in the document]
A5rocks: Well I have Python 3.13 locally and so maybe it uses `tomllib` (3.11+) instead? Wouldn't be surprised if it's just a variant of that for everyone ... or they just don't run pre-commit.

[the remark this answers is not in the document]
tjstum: Yeah, I'm using 3.10 in the venv I do trio work in.

src/trio/_core/_run.py
A5rocks: Hm, good point on not wanting `assert` in `_core`. I'm pretty sure this will be traced to a `nursery.start_soon`? Might be good to check this doesn't cause an internal error.

src/trio/_threads.py
```diff
@@ -206,9 +209,8 @@ def in_trio_thread() -> None:
 
 
 @attrs.frozen(eq=False, slots=False)
-# Explicit .../"Any" is not allowed
-class RunSync(Generic[RetT]):  # type: ignore[misc]
-    fn: Callable[..., RetT]  # type: ignore[misc]
+class RunSync(Generic[RetT]):  # type: ignore[explicit-any]
+    fn: Callable[..., RetT]  # type: ignore[explicit-any]
```
Hnasar, line 213: can you use `Unpack[Ts]` here too?

newsfragments/3201.misc.rst
```diff
@@ -0,0 +1,5 @@
+The typing of :func:`trio.abc.HostnameResolver.getaddrinfo` has been corrected to
+match that of the stdlib `socket.getaddrinfo`, which was updated in mypy 1.15 (via
+a typeshed update) to include the possibility of ``tuple[int, bytes]`` for the
+``sockaddr`` field of the result. This happens in situations where Python was compiled
+with ``--disable-ipv6``.
```
Hnasar, line 5: It's probably worth calling out the improved typing around `to_thread`/`from_thread` functions too.

Hnasar (approved):
Superb!

.pre-commit-config.yaml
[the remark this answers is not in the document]
CoolCat467: I've been meaning to add this for quite some time, thanks for catching this!

src/trio/_core/_instrumentation.py
```diff
@@ -2,23 +2,25 @@
...
+
+    P = ParamSpec("P")
+    T = TypeVar("T")
 
 
 # Decorator to mark methods public. This does nothing by itself, but
 # trio/_tools/gen_exports.py looks for it.
-# Explicit "Any" is not allowed
-def _public(fn: F) -> F:  # type: ignore[misc]
+def _public(fn: Callable[P, T]) -> Callable[P, T]:
```
TeamSpen210, line 23: This should not be `ParamSpec`. There's a subtle difference between this and a callable-bound `TypeVar` - `ParamSpec` can't capture an overload since the return type might change, but the typevar form preserves it, so it would be passed through unchanged.

TeamSpen210 (changes requested):
Not sure if any `@_public` users have overloads, but if so:

.pre-commit-config.yaml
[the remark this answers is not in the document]
jakkdl: add a comment `# required on python 3.10` so it can be cleaned up later / doesn't get removed on accident

src/trio/_core/_tests/test_ki.py
```diff
@@ -165,7 +165,7 @@ def protected_manager() -> Iterator[None]:
 async def test_async_generator_agen_protection() -> None:
     @_core.enable_ki_protection
     @async_generator  # type: ignore[misc] # untyped generator
-    async def agen_protected1() -> None:
+    async def agen_protected1() -> None:  # type: ignore[misc] # untyped generator
```
jakkdl, line 168: huh, I'm surprised we need this on both L167 and L168

jakkdl (commented):
unclear if we should even bother keeping the `async_generator` stuff at all - see https://github.com/python-trio/async_generator/issues/35

src/trio/_core/_instrumentation.py
tjstum: I looked at all the uses of `@_public`. There are no overloaded functions.

src/trio/_core/_tests/test_ki.py
tjstum: Yeah, but it is indeed needed :-/

src/trio/_core/_run.py
tjstum: This actually breaks on Cython, so I'm putting it back to a type ignore

src/trio/_threads.py
tjstum: No. `Unpack` is only valid in function definitions and specifically requires it to be in a variadic argument.

newsfragments/3201.misc.rst
```diff
@@ -0,0 +1,10 @@
...
+
+Additionally, when using mypy 1.15, the static typing of
```
A5rocks, line 7: I don't think this is restricted to mypy 1.15.

A5rocks (approved):
The changes here make sense to me!

src/trio/_core/_instrumentation.py
CoolCat467: I think the focus is that there could be in the future

src/trio/_core/_run.py
```diff
@@ -1404,10 +1402,9 @@ def __del__(self) -> None:
 
 @final
 @attrs.define(eq=False, repr=False)
-class Task(metaclass=NoPublicConstructor):  # type: ignore[misc]
+class Task(metaclass=NoPublicConstructor):  # type: ignore[explicit-any]
     _parent_nursery: Nursery | None
-    # Explicit "Any" is not allowed
-    coro: Coroutine[Any, Outcome[object], Any]  # type: ignore[misc]
+    coro: types.CoroutineType[Any, Outcome[object], Any]  # type: ignore[explicit-any]
```
CoolCat467, line 1407: Curious why this needs to be `types.CoroutineType` and not just `collections.abc.Coroutine`

src/trio/_tests/test_subprocess.py
```diff
@@ -95,8 +95,8 @@ async def open_process_then_kill(
         await proc.wait()
 
 
-@asynccontextmanager  # type: ignore[misc]  # Any in decorated
-async def run_process_in_nursery(
+@asynccontextmanager
+async def run_process_in_nursery(  # type: ignore[misc, explicit-any]
```
CoolCat467, line 99: What is the `misc` here for?

src/trio/_core/_run.py
A5rocks: Because asyncio.iscoroutine is a type guard for Coroutine and it can pass things that aren't CoroutineType. At least on Python <3.11 and probably later.

(I mean more generally the Coroutine abc doesn't require the cr_frame etc attributes but that's a circular argument.)

CoolCat467 (commented):
Looks good other than a few things I had questions about

src/trio/_tests/test_subprocess.py
tjstum: Without it:
```
trio/_tests/test_subprocess.py:99: error: Type of decorated function contains type "Any" ("Callable[[VarArg(Any), KwArg(Any)], _AsyncGeneratorContextManager[Process, None]]")  [misc]
```

src/trio/_core/_instrumentation.py
tjstum: Sure. Seems easy enough to deal with it then :-)

CoolCat467 (approved):

src/trio/_core/_instrumentation.py
TeamSpen210: Maybe, but there's no real cost to doing it correctly now, and prevents a confusing issue coming up later?

TeamSpen210 (approved):

Issue #405, opened 2018-01-18
Use native traceback mutation on python 3.7


njsmith:
Since https://github.com/python/cpython/pull/4793 was merged, it looks like traceback objects will support Python-level instantiation and mutation in 3.7. We should use that instead of the current ctypes horrors.

We should also hassle the PyPy devs about implementing this. (In some ways this is even more interesting, because we're more aggressive about dropping support for old PyPy versions.)

davidism:
> We should also hassle the PyPy devs about implementing this.

Is there an assumption that they won't make this assignable when they support 3.7 even though it's documented now?

I'm currently fixing Jinja's traceback rewriting due to the changes in 3.7 and an issue with ctypes on debug builds (pallets/jinja#1050, pallets/jinja#1051, pallets/jinja#1104), and it would be nice if I can assume any 3.7 means `tb_next` is assignable. PyPy's docs for `tproxy` now list it as deprecated.

pquentin:
Another argument in favor of this is that we're currently using PyPy transparent proxies for `concat_tb`:

https://github.com/python-trio/trio/blob/4f65ff2923d20340648a59d1ee00786fcbdbafbb/trio/_core/_multierror.py#L249-L356

They're [documented](http://doc.pypy.org/en/latest/objspace-proxies.html) and Trio uses them since 2017-02. However, in 2018-05, @mattip [marked them as deprecated](https://bitbucket.org/pypy/pypy/commits/7e57ef5dd22987f4a92e949f48d8dca25f4a08d4). It does not look like they're planning to remove them right now, but migrating away from `tputil` is probably still desirable.

njsmith:
@davidism Hey David, sorry I missed your message.

> Is there an assumption that they won't make this assignable when they support 3.7 even though it's documented now?

They probably will eventually, since the CPython 3.7 test suite has tests for it, and PyPy will eventually import those tests when adding 3.7 support. But they might do it quicker if we hassle them :-)

When I talked to the PyPy devs about this originally, they told me they knew that jinja2 depends on the deprecated "transparent proxies" code, and they weren't planning to break jinja2, so even though it was deprecated I could expect at least this specific case to keep working. That's why I felt OK using it in Trio :-). But I'm sure they wouldn't mind if we all switched to something more sensible.

[...]

PR #3203, opened 2025-02-09
Stop using ctypes for traceback mutation

A5rocks:
Fixes https://github.com/python-trio/trio/issues/405

It's kind of unnecessary since we'll rip this code out anyways once strict exception groups are the only thing supported, but for now it's simpler.

### DIFF
--- a/src/trio/_core/_concat_tb.py
+++ b/src/trio/_core/_concat_tb.py
@@ -2,13 +2,9 @@ from __future__ import annotations
 
 from types import TracebackType
 
-################################################################
-# concat_tb
-################################################################
-
 
 # this is used for collapsing single-exception ExceptionGroups when using
-# `strict_exception_groups=False`. Once that is retired this function and its helper can
+# `strict_exception_groups=False`. Once that is retired this function can
 # be removed as well.
 def concat_tb(
     head: TracebackType | None,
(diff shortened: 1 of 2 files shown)


src/trio/_core/_concat_tb.py
```diff
@@ -1,118 +1,11 @@
...
-            return operation.delegate()  # type: ignore[no-any-return]
-
-        return cast(
-            "TracebackType",
-            tputil.make_proxy(controller, type(base_tb), base_tb),
-        )  # Returns proxy to traceback
-
 
 # this is used for collapsing single-exception ExceptionGroups when using
 # `strict_exception_groups=False`. Once that is retired this function and its helper can
```
CoolCat467, line 11: "and its helper" does not apply anymore with removing this

src/trio/_core/_concat_tb.py
```diff
@@ -1,118 +1,11 @@
...
 
-from typing import TYPE_CHECKING, ClassVar, cast
-
-if TYPE_CHECKING:
-    from types import TracebackType
+from types import TracebackType
 
 ################################################################
 # concat_tb
 ################################################################
```
CoolCat467, line 8: Header is no longer required

CoolCat467 (changes requested):
Looks great other than a few comments that are now outdated

CoolCat467 (approved):

jakkdl (approved):
don't really understand the magic going on, but if tests are passing then I assume it's fine

PR #3205, opened 2025-02-10
Prevent MockClock from mutating the current runner

A5rocks:
Based on the issue @jc211 had on Gitter. You probably shouldn't be creating a `MockClock` and not using it, but this is a footgun.

### DIFF
--- a/src/trio/_core/_mock_clock.py
+++ b/src/trio/_core/_mock_clock.py
@@ -70,7 +70,7 @@ class MockClock(Clock):
         self._real_base = 0.0
         self._virtual_base = 0.0
         self._rate = 0.0
-        self._autojump_threshold = 0.0
+
         # kept as an attribute so that our tests can monkeypatch it
         self._real_clock = time.perf_counter
 
@@ -119,7 +119,8 @@ class MockClock(Clock):
         except AttributeError:
             pass
         else:
-            runner.clock_autojump_threshold = self._autojump_threshold
+            if runner.clock is self:
+                runner.clock_autojump_threshold = self._autojump_threshold
 
     # Invoked by the run loop when runner.clock_autojump_threshold is
     # exceeded.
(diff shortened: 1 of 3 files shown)


A5rocks:
I thought we were over dtls's smoke test timing out the macos pypy runner! For now I'll just rerun CI.

src/trio/_core/_mock_clock.py
```diff
@@ -119,7 +119,8 @@ def _try_resync_autojump_threshold(self) -> None:
         except AttributeError:
             pass
         else:
-            runner.clock_autojump_threshold = self._autojump_threshold
+            if runner.clock is self:
+                runner.clock_autojump_threshold = self._autojump_threshold
```
mikenerone, line 123: I'm wondering why we even need this attribute on the runner. I only see it referenced in two spots in `src/trio/_core/_run.py`, which could as easily be changed to `runner.clock.autojump_threshold`, no?

A5rocks: I think it's better continue with an attribute on runner. It doesn't have any disadvantages in practice (now that this is solved) and allows user clocks not to worry about accidentally overriding some undocumented attribute.

CoolCat467 (commented):
Wondering about where `self._autojump_threshold` gets set

src/trio/_core/_mock_clock.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, rate: float = 0.0, autojump_threshold: float = inf) -> None:
         self._real_base = 0.0
         self._virtual_base = 0.0
         self._rate = 0.0
-        self._autojump_threshold = 0.0
+
```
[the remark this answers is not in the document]
A5rocks: It gets set by the `autojump_threshold` setter a couple lines down. Unsure why it was duplicated previously...

CoolCat467 (approved):

mikenerone (approved):

Issue #3209, opened 2025-02-13
Python 3.14: test_ki_protection_doesnt_leave_cyclic_garbage fails


musicinmybrain:
The `sed` command below comments out `orjson` from `test-requirements.txt` since it’s not yet installable on Python 3.14. Otherwise, this can be reproduced by simply running the tests in a Python 3.14.0a4 virtualenv:

```
$ git clone https://github.com/python-trio/trio.git
$ cd trio
$ uv venv _e --python 3.14
Using CPython 3.14.0a4 interpreter at: /usr/bin/python3.14
Creating virtual environment at: _e
Activate with: source _e/bin/activate
$ . _e/bin/activate
(_e) $ uv pip install -e .
(_e) $ sed -r -i 's/^orjson/# &/' test-requirements.txt
(_e) $ uv pip install -r test-requirements.txt
(_e) $ python -m pytest
[…]
============================================== FAILURES ==============================================
___________________________ test_ki_protection_doesnt_leave_cyclic_garbage ___________________________

    @pytest.mark.skipif(
        sys.implementation.name != "cpython",
        reason="Only makes sense with refcounting GC",
    )
    async def test_ki_protection_doesnt_leave_cyclic_garbage() -> None:
        class MyException(Exception):
            pass
    
        async def demo() -> None:
            async def handle_error() -> None:
                try:
                    raise MyException
                except MyException as e:
                    exceptions.append(e)
    
            exceptions: list[MyException] = []
            try:
                async with _core.open_nursery() as n:
                    n.start_soon(handle_error)
                raise ExceptionGroup("errors", exceptions)
            finally:
                exceptions = []
    
        exc: Exception | None = None
        try:
            await demo()
        except ExceptionGroup as excs:
            exc = excs.exceptions[0]
    
        assert isinstance(exc, MyException)
>       assert gc.get_referrers(exc) == no_other_refs()
E       assert [<coroutine o...7f6a92bb0c40>] == []
E         
E         Left contains one more item: <coroutine object test_ki_protection_doesnt_leave_cyclic_garbage at 0x7f6a92bb0c40>
E         Use -v to get more diff

src/trio/_core/_tests/test_run.py:2846: AssertionError
====================================== short test summary info =======================================
FAILED src/trio/_core/_tests/test_run.py::test_ki_protection_doesnt_leave_cyclic_garbage - assert [<coroutine o...7f6a92bb0c40>] == []
[...]

PR #3211, opened 2025-02-14
Add xfail for issue 3209, and start testing Python 3.14

musicinmybrain:
See commit messages and discussion in https://github.com/python-trio/trio/issues/3209 for details.

### DIFF
--- a/src/trio/_core/_tests/test_run.py
+++ b/src/trio/_core/_tests/test_run.py
@@ -2820,7 +2820,6 @@ else:
 @pytest.mark.xfail(
     sys.version_info >= (3, 14),
     reason="https://github.com/python/cpython/issues/125603",
-    strict=True,
 )
 async def test_ki_protection_doesnt_leave_cyclic_garbage() -> None:
     class MyException(Exception):
--- a/src/trio/_core/_tests/test_run.py
+++ b/src/trio/_core/_tests/test_run.py
@@ -2817,6 +2817,11 @@ else:
     sys.implementation.name != "cpython",
     reason="Only makes sense with refcounting GC",
 )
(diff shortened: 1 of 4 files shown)


jakkdl:
codecov drop was due to the uploader giving an error 503 https://github.com/python-trio/trio/actions/runs/13333607547/job/37243603393?pr=3211#step:6:43
see #3173

CoolCat467 (approved):

src/trio/_core/_tests/test_run.py
```diff
@@ -2817,6 +2817,11 @@ def no_other_refs() -> list[object]:
     sys.implementation.name != "cpython",
     reason="Only makes sense with refcounting GC",
 )
+@pytest.mark.xfail(
+    sys.version_info >= (3, 14),
+    reason="https://github.com/python/cpython/issues/125603",
+    strict=True,
```
A5rocks, line 2823: This is enabled in our config so is unnecessary.

musicinmybrain: I don’t think I understand what is enabled in which config, or why this is unnecessary. Won’t the Python 3.14 CI fail if this `xfail` isn’t added, as it fails downstream in Fedora?

musicinmybrain: Oh, you’re just talking about the `strict=True`, which I now see is redundant with

https://github.com/python-trio/trio/blob/5eb5afbcfb42661a43a90c199c0e82a78d1f97fa/pyproject.toml#L231

The GitHub UI didn’t make that clear.

Yes, that makes sense. I can remove the `strict=True`.

Issue #3183, opened 2025-01-12
Add `executable` kwarg to trio.run_process type hints


dj-foxxy:
The `executable` kwarg is missing from the type hints  of  `trio.run_process`. It works at run time but Pyright is complaining.

clivinn-shla81092:
Hello, we tried to solve the issue.

This is what we did:

Add the 'executable' parameter to the type hints for the run_process function in both Windows and Unix overloads. This will resolve the Pyright complaint about the missing 'executable' kwarg.

You can review changes in this commit: https://github.com/clivinn-shla81092/python-trio-trio-3183/commit/a83f9d752074244d85c35a20182304eb37c5394e.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

A5rocks:
> The `executable` kwarg is missing from the type hints of `trio.run_process`. It works at run time but Pyright is complaining.

Absolutely, and it looks like we're missing other kwargs in the signature too. (and same for `trio.open_process) Namely:
 - `executable` (of course)
 - `group` (3.9+, unix)
 - `extra_groups` (3.9+, unix)
 - `user` (3.9+, unix)
 - `umask` (3.9+, unix)
 - `pipesize` (3.10+, linux but is ignored so all platforms? or maybe unix)
 - `process_group` (3.11+, unix)

Feel free to make a PR -- you'll have to add some `if sys.version_info` checks and update the cheat sheet comment on top of the overloads.

---

> If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

Done!

jakkdl:
That `if TYPE_CHECKING` block is gonna become an absolute monstrosity - it's already a ~third of the file with 400 lines. This might warrant a separate file (`_subprocess_type_hints.py`?) and maybe also a dedicated test that compares the type-checker inferred signature with that of `subprocess.run` (although given the false alarm rate on `test_static_tool_sees_[...]` I'm not terribly excited about that)

A5rocks:
[...]

PR #3212, opened 2025-02-16
Add missing arguments to run_process

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3183

I tried out TypedDict here like @decorator-factory recommended and it's nicer, though it's explicitly trading off user experience:
 - no defaults listed
 - wacky reveal_type at least under mypy

I *think* it's worth it? I doubt people are looking at overloads for the defaults to things.

I still need to add type tests.

### DIFF
--- a/src/trio/_subprocess.py
+++ b/src/trio/_subprocess.py
@@ -7,7 +7,7 @@ import sys
 import warnings
 from contextlib import ExitStack
 from functools import partial
-from typing import TYPE_CHECKING, Final, Literal, Protocol, Union, overload
+from typing import TYPE_CHECKING, Final, Literal, Protocol, TypedDict, Union, overload
 
 import trio
 
@@ -26,7 +26,7 @@ if TYPE_CHECKING:
     from collections.abc import Awaitable, Callable, Mapping, Sequence
     from io import TextIOWrapper
 
-    from typing_extensions import TypeAlias
+    from typing_extensions import TypeAlias, Unpack
 
     from ._abc import ReceiveStream, SendStream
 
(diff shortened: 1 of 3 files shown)


src/trio/_subprocess.py
```diff
@@ -1067,86 +1081,157 @@ async def my_deliver_cancel(process):
...
+            group: str | int | None
+            extra_groups: Iterable[str | int] | None
+            user: str | int | None
+            umask: int
+
+        class UnixProcessArgs3_10(UnixProcessArgs, total=False):
+            pipesize: int
+
+        class UnixProcessArgs3_11(UnixProcessArgs3_10, total=False):
+            process_group: int
```
A5rocks, line 1104: typeshed and the docs show this as a default to `None` but that doesn't make sense because the docs also say:
> If `process_group` is a non-negative integer, the `setpgid(0, value)` system call will be made in the child process prior to the execution of the subprocess.

So why not make it default to `-1` like `pipesize`?

jakkdl: I don't see any reason to deviate from typeshed/upstream, and there's ~no runtime difference

https://github.com/gpshead/cpython/blob/d8ca5a11bc55e2a69cab4f8795d0a5aa6932a41b/Lib/subprocess.py#L895

jakkdl (approved):
I haven't looked terribly close at the arguments, but seems fine at a glance.
In an ideal world it'd maybe be nice to have tests.. but it sounds terribly messy to write something that validates against typeshed signatures and handles version/platform differences.

PR #3215, opened 2025-02-25
Sprinkle `@slow` on slow tests

### LINKED ISSUE #3206: Mark `test_gen_exports`'s `test_process` as `run_slow`
Locally each parametrize of this takes 2 seconds. Given there's 3 of them, that's more than 20% of the runtime of `pytest` locally!

For reference:
```
================================================ slowest 10 durations =================================================
2.19s call     src/trio/_tests/tools/test_gen_exports.py::test_process[from collections import Counter\n]
2.13s call     src/trio/_tests/tools/test_gen_exports.py::test_process[from collections import Counter\nimport os\n]
2.04s call     src/trio/_tests/test_highlevel_open_tcp_listeners.py::test_open_tcp_listeners_ipv6_v6only
2.03s call     src/trio/_tests/test_socket.py::test_address_in_socket_error
2.02s call     src/trio/_tests/test_socket.py::test_SocketType_connect_paths
2.00s call     src/trio/_tests/tools/test_gen_exports.py::test_process[from typing import TYPE_CHECKING\nif TYPE_CHECKING:\n    from collections import Counter\n]
1.52s call     src/trio/_tests/test_socket.py::test_many_sockets
0.70s call     src/trio/_tests/test_exports.py::test_static_tool_sees_all_symbols[jedi-trio]
0.56s call     src/trio/_tests/tools/test_gen_exports.py::test_lint_failure
0.51s call     src/trio/_tests/test_util.py::test_coroutine_or_error
===================================== 769 passed, 77 skipped, 2 xfailed in 28.59s =====================================
```

A5rocks:
Fixes #3206.

The main thing is that it turns out that waiting for a socket to fail to connect takes up to 2 seconds on Windows -- so tests that rely on that should be marked `@slow`.

---

This approximately halves local test time on Windows.

```
================================================ slowest 10 durations =================================================
0.75s call     src/trio/_tests/test_exports.py::test_static_tool_sees_all_symbols[jedi-trio]
0.72s call     src/trio/_tests/tools/test_gen_exports.py::test_lint_failure
0.70s call     src/trio/_tests/test_repl.py::test_main_entrypoint
0.52s call     src/trio/_tests/test_util.py::test_coroutine_or_error
0.39s call     src/trio/_tests/test_exports.py::test_static_tool_sees_all_symbols[pylint-trio]
0.33s call     src/trio/_tests/test_subprocess.py::test_run
0.32s call     src/trio/_tests/test_subprocess.py::test_interactive[run_process in nursery]
0.32s call     src/trio/_tests/test_socket.py::test_many_sockets
0.26s call     src/trio/_tests/test_subprocess.py::test_interactive[open_process]
0.26s call     src/trio/_core/_tests/test_asyncgen.py::test_fallback_when_no_hook_claims_it
===================================== 762 passed, 83 skipped, 2 xfailed in 17.93s =====================================
```

### DIFF
--- a/src/trio/_tests/test_highlevel_open_tcp_listeners.py
+++ b/src/trio/_tests/test_highlevel_open_tcp_listeners.py
@@ -74,7 +74,7 @@ async def test_open_tcp_listeners_specific_port_specific_host() -> None:
 
 
 @binds_ipv6
-@slow  # turns out failing to connect to a port takes 2 seconds on Windows
+@slow
 async def test_open_tcp_listeners_ipv6_v6only() -> None:
     # Check IPV6_V6ONLY is working properly
     (ipv6_listener,) = await open_tcp_listeners(0, host="::1")
@@ -85,6 +85,8 @@ async def test_open_tcp_listeners_ipv6_v6only() -> None:
             OSError,
             match=r"(Error|all attempts to) connect(ing)* to (\(')*127\.0\.0\.1(', |:)\d+(\): Connection refused| failed)$",
         ):
+            # Windows retries failed connections so this takes seconds
+            # (and that's why this is marked @slow)
             await open_tcp_stream("127.0.0.1", port)
 
 
--- a/src/trio/_tests/test_highlevel_open_tcp_listeners.py
+++ b/src/trio/_tests/test_highlevel_open_tcp_listeners.py
@@ -20,7 +20,7 @@ from trio.abc import HostnameResolver, SendStream, SocketFactory
 from trio.testing import open_stream_to_socket_listener
 
 from .. import socket as tsocket
-from .._core._tests.tutil import binds_ipv6
+from .._core._tests.tutil import binds_ipv6, slow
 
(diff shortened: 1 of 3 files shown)


jakkdl:
2 seconds sounds unnecessarily slow, even if it's only CI. It'd be nice if we could lower the timeout somehow, or cancel it if it hasn't connected in a reasonable time. Though the former sounds difficult if there's no official API, and the latter would change what the test is actually testing for.

A5rocks:
> Upon receiving the ACK/RST client from the target host, the client determines that there is indeed no service listening there. In the Microsoft Winsock implementation of TCP, a pending connection will keep attempting to issue SYN packets until a maximum retry value is reached (set in the registry, this value defaults to 3 extra times)

It seems it only actually takes 1s, so we must be trying two ports (ipv4 and ipv6?).

Setting registry values doesn't seem like something a test suite should do locally.

CoolCat467 (approved):
Looks pretty good, agree with jakkdl's comment as well though

Issue #2992, opened 2024-04-23
Remove (Deprecate??) TrioDeprecationWarning


jakkdl:
I was thinking about whether to use `DeprecationWarning` or `TrioDeprecationWarning` in #2989, when I decided to actually figure out what the difference is and why `TrioDeprecationWarning` is a thing. It lines up with https://bugs.python.org/issue24294, which was resolved with https://peps.python.org/pep-0565/ and released in python3.7

Since we don't support 3.6 and earlier, the original reason for its existence is gone, so we at the very least should make it inherit from `DeprecationWarning` instead of `FutureWarning` and update the docstring. Or we go even further and remove `TrioDeprecationWarning` entirely (with an intermediate period of making it a pure alias of `DeprecationWarning` but emitting a `DeprecationWarning`).
Having helper methods linking to issues and stuff is still helpful though, and after recent discussions in Gitter we maybe also want a parameter for how long we're planning to keep the deprecated thing around before it gets removed.

A5rocks:
I was concerned for a bit but backwards compaibility is trivial for this. The only times `TrioDeprecationWarning` will happen is for *current* deprecations that should disappear in max 2 releases (well, except `strict_exception_groups`... not sure what to do for that).

So we just, for every *new* deprecation, start raising `DeprecationWarning` instead. The only case I think that might disrupt is if someone specifically wants to ignore all deprecations from Trio and decided that ignoring `TrioDeprecationWarning` generally is good. But I'm fine breaking their workflow, if that person exists.

jakkdl:
We could just do the equivalent of
```python
import warnings

# this is in src/trio/_deprecate
MyTrioDeprecationWarning = DeprecationWarning

# and this is what people are currently doing, suppressing TrioDeprecationWarning
with warnings.catch_warnings():
    warnings.filterwarnings('ignore', category=MyTrioDeprecationWarning)
    # It still silences DeprecationWarning
    warnings.warn(DeprecationWarning("hello"))
```
And the only warning suppression that would break is if people are filtering on `FutureWarning`. But maybe people are doing that, since specifying custom warning categories on the command line is non-trivial (maybe not even posssible? idk I couldn't quickly figure it out at least)

```
$ python -W ignore::FutureWarning foo.py         
$ python -W ignore::TrioDeprecationWarning foo.py
Invalid -W option ignored: unknown warning category: 'TrioDeprecationWarning'
[...]

PR #3216, opened 2025-02-26
Update stability info in documentation

mikenerone:
This PR updates the README to better reflect Trio’s current maturity. The previous wording hadn’t been updated since it was originally written eight years ago and no longer accurately represented Trio’s suitability for production use.

No functional changes—just a documentation update for clarity.

### DIFF
--- a/docs/source/code-of-conduct.rst
+++ b/docs/source/code-of-conduct.rst
@@ -138,9 +138,9 @@ members. This section covers actual concrete steps.
 Contacting Maintainers
 ~~~~~~~~~~~~~~~~~~~~~~
 
-As a small and young project, we don't yet have a Code of Conduct
-enforcement team. Hopefully that will be addressed as we grow, but for
-now, any issues should be addressed to `Nathaniel J. Smith
+As a small project, we don't have a Code of Conduct enforcement team.
+Hopefully that will be addressed as we grow, but for now, any issues
+should be addressed to `Nathaniel J. Smith
 <https://github.com/njsmith>`__, via `email <mailto:njs@pobox.com>`__
 or any other medium that you feel comfortable with. Using words like
 "Trio code of conduct" in your subject will help make sure your
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ keywords = [
     "trio",
 ]
 classifiers = [
-    "Development Status :: 3 - Alpha",
+    "Development Status :: 4 - Beta",
     "Framework :: Trio",
     "Intended Audience :: Developers",
     "License :: OSI Approved :: MIT License",
--- a/src/trio/_deprecate.py
+++ b/src/trio/_deprecate.py
@@ -34,11 +34,10 @@ class TrioDeprecationWarning(FutureWarning):
     upcoming deprecations and other backwards compatibility breaking changes.
 
     Despite the name, this class currently inherits from
-    :class:`FutureWarning`, not :class:`DeprecationWarning`, because while
-    we're in young-and-aggressive mode we want these warnings to be visible by
-    default. You can hide them by installing a filter or with the ``-W``
-    switch: see the :mod:`warnings` documentation for details.
-
+    :class:`FutureWarning`, not :class:`DeprecationWarning`, because until a
+    1.0 release, we want these warnings to be visible by default. You can hide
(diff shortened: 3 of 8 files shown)


CoolCat467 (approved):
Looks solid to me

A5rocks:
Could you update the docs too? They have this paragraph too.

mikenerone:
> Could you update the docs? They have this paragraph too.

Good call, updated several places in the docs.

agronholm:
How about also updating [this line](https://github.com/python-trio/trio/blob/main/pyproject.toml#L17)?

mikenerone:
> How about also updating [this line](https://github.com/python-trio/trio/blob/main/pyproject.toml#L17)?

Done!

docs/source/code-of-conduct.rst
```diff
@@ -138,9 +138,9 @@ members. This section covers actual concrete steps.
 Contacting Maintainers
 ~~~~~~~~~~~~~~~~~~~~~~
 
-As a small and young project, we don't yet have a Code of Conduct
-enforcement team. Hopefully that will be addressed as we grow, but for
-now, any issues should be addressed to `Nathaniel J. Smith
+As a small project, we don't have a Code of Conduct enforcement team.
+Hopefully that will be addressed as we grow, but for now, any issues
+should be addressed to `Nathaniel J. Smith
 <https://github.com/njsmith>`__, via `email <mailto:njs@pobox.com>`__
```
agronholm, line 144: Should this be updated too? NJS has not really been involved with Trio for some time now.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ keywords = [
     "trio",
 ]
 classifiers = [
-    "Development Status :: 3 - Alpha",
+    "Development Status :: 4 - Beta",
```
agronholm, line 17: Does this reflect the maturity status of Trio? AnyIO has been marked "Production" since its v1.0 release.

A5rocks: We don't have a 1.0 out yet so I think it's the best representation.

docs/source/code-of-conduct.rst
A5rocks: Potentially but I don't think figuring out who to write there should hold the PR up

A5rocks (approved):
Thanks for working through the wording. I'd like @njsmith to weigh in before we merge this.

src/trio/_deprecate.py
```diff
@@ -26,18 +26,18 @@
...
     Despite the name, this class currently inherits from
-    :class:`FutureWarning`, not :class:`DeprecationWarning`, because while
-    we're in young-and-aggressive mode we want these warnings to be visible by
-    default. You can hide them by installing a filter or with the ``-W``
-    switch: see the :mod:`warnings` documentation for details.
-
+    :class:`FutureWarning`, not :class:`DeprecationWarning`, because until a
+    1.0 release, we want these warnings to be visible by default. You can hide
+    them by installing a filter or with the ``-W`` switch: see the
+    :mod:`warnings` documentation for details.
```
jakkdl, line 40: The distinction between `FutureWarning` and `DeprecationWarning` disappeared in https://peps.python.org/pep-0565/, so the reason this class inherits from `FutureWarning` is solely for backwards compatibility in warning suppression.
New deprecations now use `DeprecationWarning`. See #2992 

So any important verbiage in this docstring should be mirrored elsewhere, and this docstring should perhaps be updated to say that it's not being used for new deprecations.

Zac-HD: No, there's still a distinction - see https://peps.python.org/pep-0565/#additional-use-case-for-futurewarning

(I don't think we need further edits to this text)

Zac-HD (approved):
lgtm - Anthropic has been using Trio in production for years now, and we're very happy with it 😁

src/trio/_deprecate.py
mikenerone: IMO the debatability of the nuances of this issue make it a separate consideration from the point of this PR, which is to more accurately represent the suitability of Trio for real-world use in user-facing documentation. Can we table this point for future discussion?

graingert (approved):

njsmith:
lgtm

PR #3219, opened 2025-03-05
Fix socket tests on older Linux kernels

de11n:
Ran into this trying to run the tests on RHEL8.


jakkdl:
lgtm! Minimal fix, great comment :sparkles:

Issue #3221, opened 2025-03-10
CancelScope.relative_deadline is not documented


basak:
At https://trio.readthedocs.io/en/latest/reference-core.html#trio.CancelScope.relative_deadline, there is no definition of the semantics of relative_deadline. Looks like it was added in #3010.

basak:
IIRC, there are some non-trivial details, such as that it's relative to the time that the cancel scope is first entered, but if changed later, it's relative to the time that the change occurred. This needs checking though - I inferred it from the code when I needed to understand it, concluded that I didn't need it, and that was a while ago. I just noticed now that the docs are still missing details so thought it might be helpful to note this here.

A5rocks:
IIRC we described the behavior in relevant PR comments but I'm not completely sure if it's a complete  description. Maybe tests works as a better description...

jakkdl:
that is a pretty ugly hole in the docs indeed. The `@relative_deadline` property needs a docstring - the newsfragment describe the behavior before entering: "This allows initializing scopes ahead of time, but where the specified relative deadline doesn't count down until the scope is entered." but yeah need to go through tests and/or comments in #3010 to note the behavior after entering.

jakkdl:
Hm, I was not aware of the diff view in RTD: https://trio--3197.org.readthedocs.build/en/3197/reference-core.html?readthedocs-diff=true&readthedocs-diff-chunk=1
it has some false alarms, but will try to remember it in the future. You enable it by interacting with the menu in the top right on PR doc builds

A5rocks:
> Would be nice if this got caught by a linter/test somewhere, but am not sure about any decent way of approaching that. The best would probably be a way of seeing the RTD HTML diff in an easy way

I don't know if there's any proxy for "this needs documentation and doesn't". We could require docstrings for properties (under the idea that "if it's computed it should be documented")... but I don't know, that seems a bit onerous. Maybe worth trying and seeing what fails the test?

Maybe the test should only fail if there's both a getter *and* a setter *and* the class is exposed *and* neither getter nor setter have docstrings?

A5rocks:
> Would be nice if this got caught by a linter/test somewhere, but am not sure about any decent way of approaching that. The best would probably be a way of seeing the RTD HTML diff in an easy way

[...]

PR #3222, opened 2025-03-12
Warn if there's no docstring for documented things

A5rocks:
References https://github.com/python-trio/trio/issues/3221

An earlier version found these:
```
!!!!!! trio.CancelScope.relative_deadline
!!!!!! trio.MemorySendChannel
!!!!!! trio.MemoryReceiveChannel
!!!!!! trio.MemoryChannelStatistics
!!!!!! trio.SocketStream.aclose
!!!!!! trio.SocketStream.receive_some
!!!!!! trio.SocketStream.send_all
!!!!!! trio.SocketStream.send_eof
!!!!!! trio.SocketStream.wait_send_all_might_not_block
!!!!!! trio._subprocess.HasFileno.fileno
!!!!!! trio.lowlevel.ParkingLot.broken_by
```

Which at least seems right:
 - `relative_deadline` is from the above issue
 - `SocketStream` has `:undoc-members:` ... for no good reason??
 - didn't check the others

Anyways we could probably edit the issue above for a checklist of every one of those so we have easy issues for anyone who wants to contribute!

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -26,6 +26,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, cast
 
 from sphinx.util.inventory import _InventoryItem
+from sphinx.util.logging import getLogger
 
 if TYPE_CHECKING:
     from sphinx.application import Sphinx
@@ -167,6 +168,21 @@ def autodoc_process_signature(
     return signature, return_annotation
 
 
+logger = getLogger("trio")
+
+
+def autodoc_process_docstring(
+    app: Sphinx,
+    what: str,
+    name: str,
+    obj: object,
+    options: object,
+    lines: list[str],
+) -> None:
+    if not lines:
+        logger.warning(f"{name} has no docstring")
+
+
 # XX hack the RTD theme until
 #   https://github.com/rtfd/sphinx_rtd_theme/pull/382
 # is shipped (should be in the release after 0.2.4)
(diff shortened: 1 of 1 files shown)


A5rocks:
I was wondering whether the pyright check for docstrings is still useful and I guess it is? If we decide it's still useful we should probably make the check type-completeness script not remove any warnings about docstrings, because AFAICT that's what it currently does?

jakkdl (approved):
oh that is a much cleaner implementation than I expected.

~~Have you checked that outputting on `logger.warning` will fail CI? If so, then this looks great~~ EDIT: yeah can be seen in https://readthedocs.org/projects/trio/builds/27488118/

jakkdl:
> I was wondering whether the pyright check for docstrings is still useful and I guess it is?

If I recall correctly it caught a missing docstring in the not too distant past, but you're right it's less important if it's not a docstring that would show up in the docs

> If we decide it's still useful we should probably make the check type-completeness script not remove any warnings about docstrings, because AFAICT that's what it currently does?

it only removes known warnings in the same manner as your "UNDOCUMENTED", and looking closer at the list I'm seeing substantial overlap between them. The primary differences being `check_type_completeness` also warning for classes, as well as docstrings that wouldn't show up in the docs, but doesn't notice properties.

An issue tracking missing docstrings sounds like a good idea, I don't think anybody is especially aware of the stuff listed in `src/trio/_tests/_check_type_completeness.json`.
There's also the `AsyncIOWrapper` doing weird stuff in `src/trio/_tests/check_type_completeness.py` that's littered with `# TODO`.




(my personal favorite is `# Manually confirmed to have docstrings but pyright doesn't see them due to export shenanigans. TODO: actually manually confirm that.`) :sweat_smile:

A5rocks:
> An issue tracking missing docstrings sounds like a good idea, I don't think anybody is especially aware of the stuff listed in `src/trio/_tests/_check_type_completeness.json`.

Yeah. TBH I think pyright stuff is lower priority than missing in docs; so maybe edit the referenced issue to be about all docs stuff, and a seperate issue about the pyright docstrings (with like, some note that "this is less important than <other issue>")

PR #3224, opened 2025-03-13
Nursery.start() doesn't specifically return None

TeamSpen210:
Noticed this, `start()` was set to `Any | None`, but it never specifically returns `None`, only if you call `started()` with no parameter on a `TaskStatus[None]`. If you're using a different task status, forcing `None` checks is just wrong.

### DIFF
--- a/newsfragments/3224.bugfix.rst
+++ b/newsfragments/3224.bugfix.rst
@@ -1,0 +1,1 @@
+Fix incorrect return type hint for :meth:`Nursery.start() <trio.Nursery.start>`.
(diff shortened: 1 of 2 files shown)


A5rocks (approved):
Yeah this makes sense. I don't really think it even matters whether it specifically returns `None` or not; it's just kind of irrelevant. (... if we want people to check, we should return `object`. If we don't, then `Any`. `Any | something` just doesn't really make sense)

CoolCat467 (approved):

Issue #3071, opened 2024-08-23
Allow building docs on Windows


A5rocks:
Currently building documentation on Windows fails not least of all because of links to `trio.lowlevel.FdStream`, which is exported in `trio.lowlevel` only on linux.

We should either fix this or delete `make.bat`.

A5rocks:
Maybe just conditionally add things to nitpick-ignore? I'm not sure there's much more we can do. However, referencing `trio.lowlevel.FdStream` is an error so more is necessary.

PR #3227, opened 2025-03-20
Allow documentation builds on Windows

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3071

I'm not *totally* sure about all changes here. Some are likely only because my local installation of Python is 3.13.

Also this is definitely not the cleanest method of doing this :P

### DIFF
--- a/docs/source/reference-lowlevel.rst
+++ b/docs/source/reference-lowlevel.rst
@@ -274,17 +274,20 @@ TODO: these are implemented, but are currently more of a sketch than
 anything real. See `#26
 <https://github.com/python-trio/trio/issues/26>`__.
 
-.. function:: current_kqueue()
+.. autofunction:: current_kqueue()
 
-.. function:: wait_kevent(ident, filter, abort_func)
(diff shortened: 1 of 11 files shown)


TeamSpen210:
Makes sense to me. I've always just ignored the `FDStream` error in logs when building, since I didn't need to test that bit.

jakkdl:
> I'm not totally sure about all changes here. Some are likely only because my local installation of Python is 3.13.

you can test this with `tox -e docs`, ~~though you'll either need to manually modify the `base_python` variable between runs or change the definition to `[testenv:py{39,310,311,312,313}-docs]`~~

Or rather, `.readthedocs.yml` and the `docs` environment both specify 3.11, so you don't need to verify on any other versions. We might want to bump from 311 to 313 though

A5rocks:
> so you don't need to verify on any other versions

Yeah, I don't remember what I was thinking when writing that initial comment (probably was just because I wasn't sure RTD would pass...)

A5rocks:
Are there any issues with this? I'd like to merge this otherwise.

PR #3229, opened 2025-03-21
avoid holding a reference to exception and value in to_thread_run_sync

graingert:


TeamSpen210 (changes requested):
This seems like just reimplementing `outcome`. If you need to copy, why not instantiate `outcome.Value`/`Error` directly? Only detail that's different is `unwrap` doesn't clear values, but we could just inline the implementation into the unwrap location? Or add `unwrap_clear()` or something.

graingert:
> This seems like just reimplementing `outcome`. If you need to copy, why not instantiate `outcome.Value`/`Error` directly? Only detail that's different is `unwrap` doesn't clear values, but we could just inline the implementation into the unwrap location? Or add `unwrap_clear()` or something.

oh yeah that's the plan, I want to get a proof of concept working first

graingert:
@TeamSpen210 https://github.com/python-trio/outcome/pull/45

src/trio/_core/_thread_cache.py
```diff
@@ -188,6 +188,8 @@ def _handle_job(self) -> None:
         except BaseException as e:
             print("Exception while delivering result of thread", file=sys.stderr)
             traceback.print_exception(type(e), e, e.__traceback__)
+        finally:
+            del result
```
graingert, line 192:
```suggestion
```

src/trio/_threads.py
```diff
@@ -253,6 +254,32 @@ def run_in_system_nursery(self, token: TrioToken) -> None:
...
+    def __init__(self, e: BaseException) -> None:
+        self._e: Final = e
+
+    def unwrap(self) -> NoReturn:
+        try:
+            raise self._e
+        finally:
+            del self._e
+
+
```
graingert, line 282:
```suggestion
```

src/trio/_threads.py
```diff
@@ -6,7 +6,7 @@
 import queue as stdlib_queue
 import threading
 from itertools import count
-from typing import TYPE_CHECKING, Generic, TypeVar
+from typing import TYPE_CHECKING, Final, Generic, NoReturn, Protocol, TypeVar
```
graingert, line 9:
```suggestion
from typing import TYPE_CHECKING, Generic, TypeVar
```

src/trio/_threads.py
```diff
@@ -363,7 +390,7 @@ async def to_thread_run_sync(
 
     # This function gets scheduled into the Trio run loop to deliver the
     # thread's result.
-    def report_back_in_trio_thread_fn(result: outcome.Outcome[RetT]) -> None:
+    def report_back_in_trio_thread_fn(result: _SupportsUnwrap[RetT]) -> None:
```
graingert, line 393:
```suggestion
    def report_back_in_trio_thread_fn(result: outcome.Outcome[RetT]) -> None:
```

src/trio/_threads.py
```diff
@@ -375,6 +402,12 @@ def do_release_then_return_result() -> RetT:
                 limiter.release_on_behalf_of(placeholder)
 
         result = outcome.capture(do_release_then_return_result)
+        if isinstance(result, outcome.Error):
+            result = _Error(result.error)
+        elif isinstance(result, outcome.Value):
+            result = _Value(result.value)
+        else:  # pragma: no cover
+            raise RuntimeError("invalid outcome")
```
graingert, line 410:
```suggestion
```

src/trio/_threads.py
```diff
@@ -440,20 +473,22 @@ def abort(raise_cancel: RaiseCancelT) -> trio.lowlevel.Abort:
 
         while True:
             # wait_task_rescheduled return value cannot be typed
-            msg_from_thread: outcome.Outcome[RetT] | Run[object] | RunSync[object] = (
+            msg_from_thread: _Value[RetT] | _Error | Run[object] | RunSync[object] = (
```
graingert, line 476:
```suggestion
            msg_from_thread: outcome.Outcome[RetT] | Run[object] | RunSync[object] = (
```

src/trio/_threads.py
```diff
@@ -440,20 +473,22 @@ def abort(raise_cancel: RaiseCancelT) -> trio.lowlevel.Abort:
...
-                await msg_from_thread.run()
-            elif isinstance(msg_from_thread, RunSync):
-                msg_from_thread.run_sync()
-            else:  # pragma: no cover, internal debugging guard TODO: use assert_never
-                raise TypeError(
-                    f"trio.to_thread.run_sync received unrecognized thread message {msg_from_thread!r}.",
-                )
-            del msg_from_thread
+            try:
+                if isinstance(msg_from_thread, (_Value, _Error)):
```
graingert, line 480:
```suggestion
                if isinstance(msg_from_thread, outcome.Outcome):
```

src/trio/_threads.py
```diff
@@ -36,6 +36,7 @@
     Ts = TypeVarTuple("Ts")
 
 RetT = TypeVar("RetT")
+T_co = TypeVar("T_co", covariant=True)
```
graingert, line 39:
```suggestion
```

ci.sh
```diff
@@ -46,6 +46,7 @@ python -m uv pip install build
 python -m build
 wheel_package=$(ls dist/*.whl)
 python -m uv pip install "trio @ $wheel_package" -c test-requirements.txt
+python -m uv pip install https://github.com/python-trio/outcome/archive/e0f317813a499f1a3629b37c3b8caed72825d9c0.zip
```
mikenerone, line 49: Leftover?

graingert: No I've not got the change landed in Outcome

graingert: No, once everything's decided on and merged in I'll update the pyproject.toml and remove this

A5rocks: If https://github.com/python-trio/outcome/pull/45#issuecomment-2779331336 is true then I'd at least like a PR that removes dependence on those, even if it doesn't come with tested guarantees. That way any sort of outcome 2.0 release isn't as annoying.

But I also see this PR doesn't change any `.value` or `.error` so I assume we already conform with clear-on-unwrap semantics?

PR #3230, opened 2025-03-22
Use PEP 562 instead of _ModuleWithDeprecations

A5rocks:
Presumably fixes https://github.com/python-trio/trio/issues/2135 but I haven't checked this and I'm not sure how to test this (as it relies on `cloudpickle` not `pickle`).

### DIFF
--- a/src/trio/_tests/module_with_deprecations.py
+++ b/src/trio/_tests/module_with_deprecations.py
@@ -1,27 +1,22 @@
 regular = "hi"
 
-# Make sure that we don't trigger infinite recursion when accessing module
-# attributes in between calling enable_attribute_deprecations and defining
-# __deprecated_attributes__:
 import sys
-from typing import TYPE_CHECKING
 
 from .. import _deprecate
 
+_deprecate.deprecate_attributes(
+    __name__,
+    {
+        "dep1": _deprecate.DeprecatedAttribute("value1", "1.1", issue=1),
+        "dep2": _deprecate.DeprecatedAttribute(
+            "value2",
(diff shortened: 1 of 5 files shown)


src/trio/_tests/module_with_deprecations.py
```diff
@@ -1,24 +1,27 @@
...
 
-from .. import _deprecate
-
-_deprecate.enable_attribute_deprecations(__name__)
-
 # Make sure that we don't trigger infinite recursion when accessing module
 # attributes in between calling enable_attribute_deprecations and defining
 # __deprecated_attributes__:
 import sys
+from typing import TYPE_CHECKING
```
TeamSpen210, line 7: This comment and its checks aren't relevant anymore. Might be better to pass `globals()` into the function (or just have it lookup in `sys.modules`), then `__deprecated_attributes__` would work the same way. Although currently we don't have any, so it might be fine to require them to be passed directly to the function.

A5rocks: Yeah but actually I think maybe setting `__getattr__` on `sys.modules` might be better to avoid having to get `TYPE_CHECKING`.

A5rocks: Alright I changed some stuff. Kept the same arguments (I think passing `globals` is a bit much and it's not like we have compatibility guarantees for it) but made it nicer to use / a better name / updated that file.

TeamSpen210 (approved):

CoolCat467 (approved):

Issue #2649, opened 2023-05-24
Cancelled.__context__ can leak in from other tasks


njsmith:
Suppose task 1 does:

```python
try:
    raise RuntimeError("whoops")
except Exception as task1_exc:
    my_cscope.cancel()
```

Meanwhile, task 2 is sitting in:

```python
with my_cscope:
    await trio.sleep_forever()
```

Internally, `my_cscope.cancel()` marks the scope as cancelled and then recalculates state, which triggers a call to `task2._attempt_delivery_of_any_pending_cancel()`, which calls `task2._attempt_abort(raise_cancel)`, which calls `sleep_forever`'s `abort_func`, which succeeds. And since it succeeds, it then immediately calls `reschedule(task2, outcome.capture(raise_cancel)`. Which invokes `raise_cancel`.

And since this is all happening way down inside task 1's call to `my_cscope.cancel()`, it's all happening with the Python interpreter in a state where `task1_exc` is marked as pending. So when `raise_cancel` does `raise Cancelled._create()`, then the new `Cancelled` exception in task 2 gets its `__context__` set to `task1_exc`. Which is weird and unexpected.

(It's also inconsistent, because it only happens on one particular abort pathway -- if task 2 had been sleeping in an operation that wasn't cancellable, or had deferred cancellation like an IOCP/io_uring operation, then it would still eventually end up with a `Cancelled` exception but with `__context__ = None`.)

Given our existing APIs, I think the only reasonable solution is to tweak `_attempt_abort` so it does:

```python
error = capture(raise_cancel)
error.error.__context__ = None
self._runner.reschedule(self, error)
```

It's also worth thinking a bit about how we got here: this only happens because of our weird API where we represent a pending cancellation as a callable that raises the actual exception. The reason for *this* is:

- For operations that have deferred cancellation, like IOCP/io_uring, the `abort_func` has to request the cancellation, and then some time later it gets notified about whether the cancellation succeeded. If it did, then it now has to raise `Cancelled`. But:
  - Now we're in user code, without any special access to Trio's internals
  - `Cancelled` has no public constructor; and besides, the abort might have been caused by either a `KeyboardInterrupt` or a `Cancelled`, so we have to pass in the exception to the user code somehow
  - If it's a `KeyboardInterrupt`, those have exactly-once semantics, so the core code needs to know whether the user code ended up raising it or not
  - So that's why we end up representing the exception-to-be-raised as a function that does the raise -- so the core code can find out whether the exception is raised or not

[...]

PR #3233, opened 2025-03-28
Defer KI if trio is doing IO

A5rocks:
Supersedes https://github.com/python-trio/trio/pull/1537.

As with that, fixes https://github.com/python-trio/trio/issues/151, first step of https://github.com/python-trio/trio/issues/733.

Additionally, fixes https://github.com/python-trio/trio/issues/2649 (I need to add a test) and part of https://github.com/python-trio/trio/issues/3007 (not intentionally, just because it has to be done in order to still raise a KI. I should add a test.).

### DIFF
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -1216,12 +1216,10 @@ class Nursery(metaclass=NoPublicConstructor):
             # If we have a KeyboardInterrupt injected, we want to save it in
             # the nursery's final exceptions list. But if it's just a
             # Cancelled, then we don't -- see gh-1457.
-            def aborted(raise_cancel: _core.RaiseCancelT) -> Abort:
-                exn = capture(raise_cancel).error
+            def aborted(exn: BaseException) -> Abort:
+                # TODO: when do we abort with something other than Cancelled?
                 if not isinstance(exn, Cancelled):
                     self._add_exc(exn)
-                # see test_cancel_scope_exit_doesnt_create_cyclic_garbage
-                del exn  # prevent cyclic garbage creation
                 return Abort.FAILED
 
             self._parent_waiting_in_aexit = True
@@ -1424,7 +1422,7 @@ class Task(metaclass=NoPublicConstructor):  # type: ignore[explicit-any]
     # Tasks start out unscheduled.
     _next_send_fn: Callable[[Any], object] | None = None  # type: ignore[explicit-any]
     _next_send: Outcome[Any] | BaseException | None = None  # type: ignore[explicit-any]
-    _abort_func: Callable[[_core.RaiseCancelT], Abort] | None = None
+    _abort_func: Callable[[BaseException], Abort] | None = None
     custom_sleep_data: Any = None  # type: ignore[explicit-any]
 
     # For introspection and nursery.start()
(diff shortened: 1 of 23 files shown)


A5rocks:
codecov failure doesn't make any sense and seems to be a recent thing. I'm inclined to believe it's some change in how they normalize file names but I don't think it's worth chasing down for now.

src/trio/_core/_traps.py
```diff
@@ -151,40 +151,43 @@ def abort_func(raise_cancel):
...
+       ("overlapped") I/O operation be cancelled, but this request is
+       *also* asynchronous – you don't find out until later whether
+       the operation was actually cancelled or not.)  To report a
+       delayed cancellation, you should reschedule the task yourself,
+       and cause it to raise the exception ``exc`` that was passed to
+       ``abort_func``. (Currently ``exc`` will always be a
+       `~trio.Cancelled` exception, but we may use this mechanism to
+       raise other exceptions in the future; you should raise whatever
+       you're given.) Either of the approaches sketched below can
+       work::
```
jakkdl, line 171: is this block of text modified? Or just reformatted? It's hard to see in the diff

jakkdl (commented):
Is the change to the abort parameter tightly bound to the other changes in this PR? Can it be pulled out into a separate PR?

This is a daunting PR to review, with one former PR and 4 related issues, and changes in 23 files, and it might be more manageable if it got split up.

It'd also be nice if the PR summarized the state of discussion, I don't know if KI-deferral has been unanimously agreed to be the better way to do it, or if there's up- & downsides.

Sorry for the non-review review <3

A5rocks:
I can do the change to abort fn as a follow up -- it needs this change because the reason it's a function is because KI can be raised (which this change stops), but this change doesn't depend on it.

src/trio/_core/_traps.py
A5rocks: Modified, specifically the second to last sentance (which turns into two). I copied this over from the old PR which might explain the formatting...

A5rocks (commented):
Alright, I moved the abort function signature thing to another commit.

src/trio/_threads.py
```diff
@@ -351,6 +351,8 @@ async def to_thread_run_sync(
     if limiter is None:
         limiter = current_default_thread_limiter()
 
+    # TODO: cancel_register can probably be a single element tuple for typing reasons
```
jakkdl, line 354: This helps in stopping people from indexing above 0, but it breaks a bunch of current code that manipulates the content of the list. Ideally it'd be some minimalist wrapper class.

src/trio/_core/_run.py
```diff
@@ -2105,30 +2106,13 @@ def current_trio_token(self) -> TrioToken:
...
-        # the same time -- so even if KI arrives before main_task is created,
-        # we won't get here until afterwards.
-        assert self.main_task is not None
-        if self.main_task_outcome is not None:
-            # We're already in the process of exiting -- leave ki_pending set
-            # and we'll check it again on our way out of run().
-            return
-        self.main_task._attempt_delivery_of_pending_ki()
+        with suppress(RunFinishedError):
+            self.entry_queue.run_sync_soon(self.main_task_nursery.cancel_scope.cancel)
```
jakkdl, line 2115: nice, this will be easy to harmonize with #3232 :rocket:

src/trio/_core/_tests/test_ki.py
```diff
@@ -325,9 +325,12 @@ async def check_protected_kill() -> None:
             nursery.start_soon(_core.enable_ki_protection(raiser), "r1", record_set)
             # __aexit__ blocks, and then receives the KI
 
-    # raises inside a nursery, so the KeyboardInterrupt is wrapped in an ExceptionGroup
-    with RaisesGroup(KeyboardInterrupt):
+    # KeyboardInterrupt is inserted from the trio.run
+    with pytest.raises(KeyboardInterrupt) as excinfo:
         _core.run(check_protected_kill)
+
+    # TODO: be consistent about providing Cancelled tree as __context__
```
jakkdl, line 332: what's required for this?

jakkdl (approved):
seems fine as far as I can tell~

src/trio/_threads.py
richardsheridan: would a deque with maxlen=1 help?

jakkdl: I wouldn't expect typing to take the maxlen into account

A5rocks: I somehow forgot that tuples aren't mutable when writing the comment, I just saw we always had a one element list and my brain instantly went "oh this should be a tuple".

I'll remove the TODO

TeamSpen210: Would make sense to use a `types.CellType`, except that's not generic in typeshed. If making a custom class, might as well include everything put in `PARENT_TASK_DATA`.

src/trio/_core/_tests/test_ki.py
A5rocks: Well the issue here is that in the above test, no `__context__` tree is kept which is wrong. Maybe.

But now that I think about it more, I think this might be fine? If a task is busy-looping it'll raise a `KeyboardInterrupt` too. I don't really like the `Cancelled` tree so maybe we should do the opposite of this and try to eliminate it! I'll update this to have this written down. (I'd rather know whether this tree of cancellation is actually a burden but I suspect it'll add many lines to tracebacks in event of ctrl+C for not much use)

A5rocks:
Looking at #3030 makes me unsure whether this will work with that (specifically pressing ctrl+c when some async code is already running seems like it would break with this). I'd rather have that first, so I'll mark this as a draft.

I guess that PR doesn't actually touch the code I'm worried about, but I want to copy over whatever style of test case we figure out anyways.

PR #3234, opened 2025-03-28
Add tool to keep test-requirements in sync with pre-commit

CoolCat467:
In this pull request, we add a local pre-commit hook tool to keep the tool versions in `test-requirements.txt` in sync with the equivalent versions in `.pre-commit-config.yaml` automatically, meaning pre-commit.ci automatic updates are more seamless and there can never be cases where the test versions don't match the pre-commit versions.

### DIFF
--- a/src/trio/_tools/sync_requirements.py
+++ b/src/trio/_tools/sync_requirements.py
@@ -25,10 +25,10 @@ except ImportError:
 
 
 def yield_pre_commit_version_data(
-    pre_commit: Path,
+    pre_commit_text: str,
 ) -> Generator[tuple[str, str], None, None]:
     """Yield (name, rev) tuples from pre-commit config file."""
-    pre_commit_config = load_yaml(pre_commit.read_text(encoding="utf-8"), Loader)
+    pre_commit_config = load_yaml(pre_commit_text, Loader)
     for repo in pre_commit_config["repos"]:
         if "repo" not in repo or "rev" not in repo:
             continue
@@ -81,18 +81,19 @@ def main() -> int:
     """Run program."""
 
     source_root = Path.cwd().absolute()
-    while not (source_root / "LICENSE").exists():
-        source_root = source_root.parent
+
     # Double-check we found the right directory
     assert (source_root / "LICENSE").exists()
     pre_commit = source_root / ".pre-commit-config.yaml"
     test_requirements = source_root / "test-requirements.txt"
 
+    pre_commit_text = pre_commit.read_text(encoding="utf-8")
+
     # Get tool versions from pre-commit
     # Get correct names
     pre_commit_versions = {
         name.removesuffix("-mirror").removesuffix("-pre-commit"): version
-        for name, version in yield_pre_commit_version_data(pre_commit)
+        for name, version in yield_pre_commit_version_data(pre_commit_text)
     }
     changed = update_requirements(test_requirements, pre_commit_versions)
     return int(changed)
--- a/src/trio/_tools/sync_requirements.py
+++ b/src/trio/_tools/sync_requirements.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+
+"""Sync Requirements - Automatically upgrade test requirements pinned
+versions from pre-commit config file."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from yaml import load as load_yaml
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from yaml import CLoader as _CLoader, Loader as _Loader
+
+    Loader: type[_CLoader | _Loader]
+
+try:
+    from yaml import CLoader as Loader
+except ImportError:
+    from yaml import Loader
+
+
+def yield_pre_commit_version_data(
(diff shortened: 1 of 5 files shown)


src/trio/_tools/sync_requirements.py
```diff
@@ -0,0 +1,103 @@
...
+    assert (source_root / "LICENSE").exists()
+    pre_commit = source_root / ".pre-commit-config.yaml"
+    test_requirements = source_root / "test-requirements.txt"
+
+    pre_commit_text = pre_commit.read_text(encoding="utf-8")
+
+    # Get tool versions from pre-commit
+    # Get correct names
+    pre_commit_versions = {
+        name.removesuffix("-mirror").removesuffix("-pre-commit"): version
```
A5rocks, line 95: IMO we should just do a dict of url to package name at the top instead of trying to automatically detect it

test-requirements.txt
```diff
@@ -177,6 +177,8 @@ types-docutils==0.21.0.20241128
     # via -r test-requirements.in
 types-pyopenssl==24.1.0.20240722
     # via -r test-requirements.in
+types-pyyaml==6.0.12.20250326
```
A5rocks, line 180: I think it's probably better to actually add a dependency on whatever yaml import we want. It shouldn't actually matter because pre-commit will use it too, so we're already installing it.

A5rocks (commented):
Thanks this makes sense and seems pretty simple. We could probably make it simpler too.

I wonder whether maybe this should update the pre-commit configuration too, to maximize the version (lexicographically?) between both versions.

Another helpful related script might be to copy the pins from the requirements.txt files to the pre-commit file.

src/trio/_tools/sync_requirements.py
```diff
@@ -0,0 +1,103 @@
...
+                continue
+            # Otherwise might have changed
+            new_line = f"{name}=={version}{extra}"
+            if new_line != line:
+                if not changed:
+                    changed = True
+                    print("Changed test requirements version to match pre-commit")
+                print(f"{name}=={old_version} -> {name}=={version}")
+            file.write(new_line)
+    return changed
```
A5rocks, line 77: This should also check whether every entry in version_data was used

CoolCat467: I don't immediately understand why this should be the case. For example, `typos` exists only as a pre-commit hook, and in fact I don't know if it even exists as a python package. Similar with `zizmor`.

test-requirements.txt
CoolCat467: `pyyaml` is what `pre-commit` itself uses, which is also the reason I am using it. This is the type annotations package for mypy to be able to understand it without reporting errors in CI.

CoolCat467:
> I wonder whether maybe this should update the pre-commit configuration too, to maximize the version (lexicographically? Or just import `packaging`) between both files.
> 
> Another helpful related script might be to copy the pins from the requirements.txt files to the pre-commit file.

While this could be useful in some cases, I don't see a need for that in a regular use case. From my point of view, most of the time someone (or pre-commit.ci) would run `pre-commit autoupdate` and then `uv pip compile --universal --python-version=3.9 test-requirements.in -o test-requirements.txt --upgrade`, and then since all the pre-commit hook versions are either the original package itself or a mirror, they should have an identical version.

The primary use case for this script is so that when pre-commit ci automatically updates everything, this will automate pinning the test requirements to the same version.

Another possible solution that would sidestep needing this tool in the first place would be to just have pre-commit ci perform an upgrade on all the test requirements as well, but I don't think there is a way to specify a pre-commit hook to run that only runs on pre-commit.ci but not locally, unless one were to write a script to somehow identify something in the pre-commit.ci environment that is different from a local run.

test-requirements.txt
A5rocks: Sorry for some reason I forgot that we don't have `pre-commit` as a dep in this file.

src/trio/_tools/sync_requirements.py
A5rocks: Yeah this would only make sense if https://github.com/python-trio/trio/pull/3234#discussion_r2018183152 was implemented.

src/trio/_tools/sync_requirements.py
CoolCat467: I was intending for this to be general enough to work even if new hooks are added in the future, but if necessary I could change it to be a static dictionary

CoolCat467:
Thinking of merging this in a day or two if no other changes to make

src/trio/_tools/sync_requirements.py
A5rocks: I don't think this looks very general (what if the name ends with `-precommit`? or similar) and it's not an error if something that could be picked up automatically isn't, so I think I'd prefer a manual addition.

I don't know of anything I would add that this wouldn't work with, but I also don't know anything I would add to the pre-commit configuration to begin with!

A5rocks (approved):
Left a comment on whether we should derive the requirement manually or automatically but I'm OK with this either way; it probably wouldn't simplify much either.

A5rocks:
Oops I missed that this decreased coverage! I'll make a PR to fix that (just by adding a few coverage ignore comments...)

Issue #3238, opened 2025-04-02
Define `trio.Event.__bool__()` to reduce bugs


Zac-HD:
I've now hit enough bugs where someone wrote `if event:` rather than `if event.is_set():` that I think `Event.__bool__()` should raise an error of some kind, likely `NotImplementedError("Trio events cannot be treated as bools; consider using 'event.is_set()'")`.

I could be argued into `def __bool__(self): return self.is_set()` but disprefer this - "there should be one obvious way to do it".  Thoughts?

I don't think there's as urgent a need on other classes, but we could consider e.g. Condition and Semaphore too...

A5rocks:
Since `trio.Event` is marked as a final class, mypy (at least) supports `--enable-error-code=truthy-bool` that will warn on `if event`. (actually I'm not even sure if that requires being marked as final)

I'm not sure how useful something at runtime is compared to that though.

Zac-HD:
Seems very useful for everyone who doesn't use `mypy`, including those on other typecheckers!  (like me, for example 😅)

AbduazizZiyodov:
What about normal(expected) behaviour + showing a warning (`warnings.warn`) on console additionally ? I saw this in the case of `lxml` (ElementTree if I remember correctly).

A5rocks:
I think we should have a warning at first for deprecation reasons, but it should imitate the current behavior and return `True`.

But I suspect that having a proper traceback is nicer, plus warnings aren't harsh enough. And error->no error is trivial, no error->error requires a deprecation warning if we want to change this in the future.

A5rocks:
As part of the 0.34.0 release, I've removed our deprecation and replaced it with the initially proposed `NotImplementedError`!

PR #3239, opened 2025-04-02
Enable extra error codes for mypy

A5rocks:
In https://github.com/python-trio/trio/issues/3238 I noticed the `truthy-bool` error code seems useful. I combined it with some changes I have locally for another interesting-but-unimpactful error code.

This doesn't catch anything but it's better to lead by example! (all 4 places flagged by `truthy-bool` were us using `assert` to tell various tools that the thing being asserted is used.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -193,6 +193,7 @@ files = ["src/trio/", "docs/source/*.py"]
 ignore_missing_imports = true
 
 # Be strict about use of Mypy
+enable_error_code = ["truthy-bool", "mutable-override"]
 local_partial_types = true
 warn_unused_ignores = true
 warn_unused_configs = true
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -193,6 +193,7 @@ files = ["src/trio/", "docs/source/*.py"]
 ignore_missing_imports = true
 
 # Be strict about use of Mypy
+enable_error_code = ["truthy-bool", "mutable-override"]
 local_partial_types = true
```
jakkdl, line 197: ugh, so messy that this is a separate way of enabling checks vs having `warn_truthy_bool` (and I did a quick check for it).
also sad that `strict=True` leaves so much disabled. The workflow of `--enable=ALL` into selectively disabling is so much better.

aanyway this PR in itself seems fine

jakkdl (approved):

CoolCat467 (approved):

PR #3240, opened 2025-04-02
add helper for unwrapping exception groups with single exception

jakkdl:
#3197 has the fairly classic case of "I have an exceptiongroup, but I know it only contains one exception, so I don't want my user to have to use `except*`". This was also encountered in trio-websocket (https://github.com/python-trio/trio-websocket/pull/188, https://github.com/python-trio/trio-websocket/pull/192), and we have a long section in the docs talking about how messy this is https://trio.readthedocs.io/en/stable/reference-core.html#designing-for-multiple-errors

This PR introduces a helper that can be used in #3197 and trio-websockets, and can be recommended in the docs as well as by flake8-async. There's several pitfalls around unwrapping groups, so users generally shouldn't have to do that by themselves. See https://github.com/HypothesisWorks/hypothesis/issues/4115 and https://github.com/python-trio/trio-websocket/issues/191

I am very open to design discussion on how exactly this helper should behave - it's largely inspired by the implementation I did in trio-websocket. I am also very open to alternative names for the helper, I don't love `raise_single_exception_from_group`.

TODO:
- [ ] ~~docs on the helper~~
- [ ] ~~suggest using the helper in https://trio.readthedocs.io/en/stable/reference-core.html#designing-for-multiple-errors~~
- [ ] ~~newsfragment~~

### DIFF
--- a/src/trio/__init__.py
+++ b/src/trio/__init__.py
@@ -28,7 +28,6 @@ from ._channel import (
     MemoryReceiveChannel as MemoryReceiveChannel,
     MemorySendChannel as MemorySendChannel,
     open_memory_channel as open_memory_channel,
-    raise_single_exception_from_group as raise_single_exception_from_group,
 )
 from ._core import (
     BrokenResourceError as BrokenResourceError,
--- a/src/trio/__init__.py
+++ b/src/trio/__init__.py
@@ -28,6 +28,7 @@ from ._channel import (
     MemoryReceiveChannel as MemoryReceiveChannel,
     MemorySendChannel as MemorySendChannel,
     open_memory_channel as open_memory_channel,
+    raise_single_exception_from_group as raise_single_exception_from_group,
 )
 from ._core import (
     BrokenResourceError as BrokenResourceError,
--- a/src/trio/_util.py
+++ b/src/trio/_util.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import collections.abc
 import inspect
 import signal
+import sys
 from abc import ABCMeta
 from collections.abc import Awaitable, Callable, Sequence
 from functools import update_wrapper
@@ -19,6 +20,9 @@ from typing import (
 from sniffio import thread_local as sniffio_loop
 
 import trio
+
+if sys.version_info < (3, 11):
+    from exceptiongroup import BaseExceptionGroup
 
 # Explicit "Any" is not allowed
 CallT = TypeVar("CallT", bound=Callable[..., Any])  # type: ignore[explicit-any]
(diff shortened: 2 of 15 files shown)


src/trio/__init__.py
```diff
@@ -28,6 +28,7 @@
     MemoryReceiveChannel as MemoryReceiveChannel,
     MemorySendChannel as MemorySendChannel,
     open_memory_channel as open_memory_channel,
+    raise_single_exception_from_group as raise_single_exception_from_group,
```
A5rocks, line 31: I don't think this should be publicly exported. I think service nurseries take make (if not all) of its use cases and I'd rather encourage their use whenever we get them.

A5rocks (changes requested):

src/trio/__init__.py
jakkdl: I feel like there might still be scenarios where you want this in more complex stuff going on, depending on exactly how the service-nursery implementation ends up - but we can leave it private for now and see how usage pans out.

src/trio/_util.py
```diff
@@ -353,3 +357,61 @@ def wraps(  # type: ignore[explicit-any]
...
+    noncancelled_exceptions = []
+
+    # subgroup/split retains excgroup structure, so we need to manually traverse
+    def _parse_excg(e: BaseException) -> None:
+        if isinstance(e, KeyboardInterrupt):
+            # immediately bail out
+            raise KeyboardInterrupt from eg
+
+        if isinstance(e, trio.Cancelled):
+            cancelled_exceptions.append(e)
```
A5rocks, line 401: I haven't looked at the PR which introduced this to trio-websocket but I'm a bit surprised by this... It's not necessarily wrong since then the next await point will add in `Cancelled`, but it feels like this isn't exactly necessary. The only reason a `Cancelled` would show up here is that some `CancelScope` further up the stack is cancelled, in which case any other exception is a failure in cancel-cleanup code and should probably be acknowledged.

I get the case for at least compressing `BaseExceptionGroup(..., [Cancelled._create(), Cancelled._create()])` though. Is there some other case I'm missing? Or is this maybe just a simpler implementation? I don't care much about this, it just feels weird to special case `Cancelled` here.

A5rocks (approved):
Only a question but I think this seems fine.

src/trio/_util.py
jakkdl: the reason this special-cases `Cancelled` is to compress `BaseExceptionGroup(..., [ArbitraryError, Cancelled])` -> `ArbitraryError` without having to panic and raise `AssertionError`.
Does that answer the question? If not then I'm not quite following

src/trio/_util.py
```diff
@@ -353,3 +357,61 @@ def wraps(  # type: ignore[explicit-any]
...
+
+
+def raise_single_exception_from_group(
+    eg: BaseExceptionGroup[BaseException],
+) -> NoReturn:
+    """This function takes an exception group that is assumed to have at most
+    one non-cancelled exception, which it reraises as a standalone exception.
+
+    If a :exc:`KeyboardInterrupt` is encountered, a new KeyboardInterrupt is immediately
+    raised with the entire group as cause.
```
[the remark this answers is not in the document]
jakkdl: not really, the reason https://trio.readthedocs.io/en/stable/reference-core.html#designing-for-multiple-errors only mentions `KeyboardInterrupt` is presumably to hammer the point home that the user has no control of when it may happen - but this handler might as well handle them the same way.

src/trio/_util.py
A5rocks: Hm, not quite. I don't know how to really explain myself, it just feels weird that we special case Cancelled given that that will only happen if something outside of the code that's raising is cancelled, eg:

```
with trio.CancelScope() as cancel_scope, this_as_a_context_manager:
    nursery.start_soon(trio.sleep, 5)
    await trio.sleep(1)
    cancel_scope.cancel()
    raise ValueError("bleh")
```

Like in the above case it's not obvious to me that we should only keep the ValueError.

jakkdl: yeah I mean all alternatives suck one way or another. But unless we want to raise `AssertionError` I think it's much better to drop the cancellation (since it can trigger again) than to drop the other exception.
But maybe cancel+exception is not expected to come up in any remotely normal usage (with e.g. #3197) and everything should panic if it does. idk I'm not sure what the usage patterns would look like

A5rocks: Alright as long as you've considered it. It doesn't seem that bad, just slightly... weird, I suppose?

Zac-HD:
See https://github.com/python/peps/pull/4357 for some notes on traceback preservation when re-raising one leaf from a group!

jakkdl:
> See [python/peps#4357](https://github.com/python/peps/pull/4357) for some notes on traceback preservation when re-raising one leaf from a group!

happy to see the PEP!
Yeah this will eat the traceback on the group, but I'm not even sure that's bad in this case - the traceback on the group should only be of being shuffled around in trio internals

PR #3241, opened 2025-04-03
add relative_deadline docstring

jakkdl:
ticks off one docstring from #3221

property added in #3010


A5rocks (approved):
The whole "Read-write, :class:\`float\`" thing is weird but I guess we should follow conventions... Makes sense to me.

jakkdl:
> The whole "Read-write, :class:`float`" thing is weird but I guess we should follow conventions... Makes sense to me.

yeah I don't love it. And I don't think we're following that convention in all our docstrings. But /shrug

Issue #3244, opened 2025-04-07
Replace hoverxref with RTD-native links preview


jakkdl:
We're currently using https://pypi.org/project/sphinx-hoverxref/ but it's been deprecated and incorporated into ReadTheDocs itself as https://docs.readthedocs.com/platform/stable/link-previews.html

That also means we get rid of an ugly warning https://github.com/readthedocs/sphinx-hoverxref/issues/312#issuecomment-2775944738

PR #3245, opened 2025-04-07
Move away from hoverxrefs

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3244

As far as I can tell, this requires someone to press buttons in the RTD UI so it's not too useful to leave this up for someone new to Trio to fix.

### DIFF
--- a/docs-requirements.in
+++ b/docs-requirements.in
@@ -7,7 +7,6 @@ sphinx_rtd_theme >= 3
 sphinxcontrib-jquery
 sphinxcontrib-trio
 towncrier
-sphinx-hoverxref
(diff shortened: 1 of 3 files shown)


A5rocks:
I've clicked the relevant buttons in the RTD UI and checked that:
 - current docs still work
 - the docs preview should work

The latter one works but the former one has 2 hover pop-ups. I'm not really sure how we can migrate gracefully, might be worth asking.

For now I've disabled the link preview addon again.

TeamSpen210:
Might want to just wait until we're releasing a new build, then merge just before.

PR #3249, opened 2025-04-11
Switch from `PurePath#as_uri` to `PurePath#joinpath` for our test that PurePath methods get inherited

A5rocks:
`as_uri` is deprecated in `PurePath`. This commit also introduces a test to ensure users have a workaround that is not deprecated.

Failing test run: https://github.com/python-trio/trio/actions/runs/14393482504/job/40364976912

### DIFF
--- a/src/trio/_path.py
+++ b/src/trio/_path.py
@@ -244,6 +244,9 @@ class Path(pathlib.PurePath):
     if sys.version_info >= (3, 13):
         full_match = _wrap_method(pathlib.Path.full_match)
 
+    def as_uri(self) -> str:
+        return pathlib.Path.as_uri(self)
+
 
 @final
 class PosixPath(Path, pathlib.PurePosixPath):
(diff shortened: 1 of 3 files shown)


A5rocks:
I wasn't too sure about doing that because there's no fundamental reason that has to be on `trio.Path` other than historical, and I wasn't sure if there's other sync methods on `pathlib.Path` that are not on `pathlib.PurePath` yet still don't do anything bad. (so should we just support all of them? We should have a consistent recommendation that applies for this new `as_uri` as well.)

And users can do `pathlib.Path.as_uri(some_trio_path)` to work around the deprecation warning, because `trio.Path` is a subclass of `pathlib.PurePath`.

Zac-HD:
IMO "maintaining 1:1 compatibility with existing code" is enough reason in this case, and right now unblocking CI feels more urgent than answering those other API design questions 😅 

(but for completeness: I'd like `trio.Path` to support every method on `pathlib.Path`, async if they do IO and sync if they just depend on OS state as in this case - I think the latter is rare though)

Zac-HD (approved):
Looks great - thanks for jumping on this @A5rocks!

Issue #3248, opened 2025-04-10
trio.Cancelled can't be pickled


hotpxl:
`trio.Cancelled` can't be constructed directly so it can't be pickled. Usually this is fine as no one needs to pickle and move around a `trio.Cancelled` object. But sometimes it gets trapped in some other exception's `__context__` or `__cause__`.

One example is
```py
async def test():
    with trio.fail_after(0.1):
        await trio.sleep(5)

try:
    trio.run(test)
except trio.TooSlowError as e:
    print(e.__context__)
```

(By default `__context__` of an exception doesn't get pickled but `tblib.pickling_support` adds it, which I think is reasonable?)

And now we can't pickle this `TooSlowError` anymore.

Here's the full example:
```py
from tblib import pickling_support
pickling_support.install(trio.TooSlowError)

async def test():
    with trio.fail_after(0.1):
        await trio.sleep(5)

try:
    trio.run(test)
except trio.TooSlowError as e:
    b = e

import pickle
pickle.loads(pickle.dumps(b))
```

Can we add pickle support for `trio.Cancelled`?

A5rocks:
Prior discussion in https://github.com/python-trio/trio/pull/3105.

PR #3250, opened 2025-04-12
Allow pickling `Cancelled`

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3248

### DIFF
--- a/newsfragments/3248.bugfix.rst
+++ b/newsfragments/3248.bugfix.rst
@@ -1,0 +1,1 @@
+Allow pickling `trio.Cancelled` objects, as they can show up when you want to pickle something else.
--- a/newsfragments/3248.bugfix.rst
+++ b/newsfragments/3248.bugfix.rst
@@ -1,1 +1,1 @@
-Allow pickling `trio.Cancelled` objects, as they can show up when you want to pickle something else.
+Allow pickling `trio.Cancelled`, as they can show up when you want to pickle something else. Also enables pickling for other ``NoPublicConstructor`` objects.
--- a/newsfragments/3248.bugfix.rst
+++ b/newsfragments/3248.bugfix.rst
@@ -1,1 +1,1 @@
-Allow pickling `trio.Cancelled`, as they can show up when you want to pickle something else. Also enables pickling for other ``NoPublicConstructor`` objects.
+Allow pickling `trio.Cancelled`, as they can show up when you want to pickle something else. This does not rule out pickling other ``NoPublicConstructor`` objects -- create an issue if necessary.
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -1,6 +1,11 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from trio._util import NoPublicConstructor, final
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
 
 
 class TrioInternalError(Exception):
@@ -64,6 +69,9 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     def __str__(self) -> str:
         return "Cancelled"
+
+    def __reduce__(self) -> tuple[Callable[[], Cancelled], tuple[()]]:
+        return (Cancelled._create, ())
 
 
 class BusyResourceError(Exception):
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+import typing
+
 from trio._util import NoPublicConstructor, final
 
 
@@ -62,6 +66,9 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     def __str__(self) -> str:
         return "Cancelled"
+
+    def __reduce__(self) -> tuple[typing.Callable[[], Cancelled], tuple[()]]:
+        return (Cancelled._create, ())
 
 
 class BusyResourceError(Exception):
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
-import typing
+from typing import TYPE_CHECKING
 
 from trio._util import NoPublicConstructor, final
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     from collections.abc import Callable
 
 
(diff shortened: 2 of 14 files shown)


src/trio/_core/_exceptions.py
```diff
@@ -63,6 +67,9 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
     def __str__(self) -> str:
         return "Cancelled"
 
+    def __reduce__(self) -> tuple[typing.Callable[[], Cancelled], tuple[()]]:
```
CoolCat467, line 70: Any reason not to used `collections.abc.Callable`?

A5rocks: No, just me forgetting to.

CoolCat467 (approved):

jakkdl:
we could instead add `__reduce__` to the `NoPublicConstructor` metaclass to enable pickling for all objects that make use of it

A5rocks:
> we could instead add `__reduce__` to the `NoPublicConstructor` metaclass to enable pickling for all objects that make use of it

I was thinking that too (we'd need to add it in `NoPublicConstructor`'s `__init__` or make a class we inherit from that uses `NoPublicConstructor`, but totally possible) but I'm not sure we *want* pickling for everything. I guess I can't think of any reasons against though.

A5rocks:
Alright generally applying `__reduce__` is fine and TBH it's probably fine. People could work around not having a public constructor anyways.

The main point is the number of `type: ignore`s it uses but I assume that could be minimized if I actually cared to.

src/trio/_util.py
```diff
@@ -323,6 +323,20 @@ def __call__(cls, *args: object, **kwargs: object) -> None:
     def _create(cls: type[T], *args: object, **kwargs: object) -> T:
         return super().__call__(*args, **kwargs)  # type: ignore
 
+    def __new__(cls, name: str, bases: tuple[type, ...], namespace: dict[str, Any], **kwargs: object) -> type:  # type: ignore[explicit-any]
+        old_reduce = namespace.get("__reduce__")
+
+        def __reduce__(self: object) -> Any:  # type: ignore[explicit-any]
```
CoolCat467, line 329: In this case, wouldn't `self` be an instance of `NoPublicConstructor`? If so, that would probably alleviate the `attr-defined` error on line 335

A5rocks: No, `Cancelled` itself is an instance of `NoPublicConstructor`, not instances of `Cancelled`. Metaclasses are weird!

newsfragments/3248.bugfix.rst
```diff
@@ -0,0 +1 @@
+Allow pickling `trio.Cancelled` objects, as they can show up when you want to pickle something else.
```
jakkdl, line 1: now that it's generalized we should update the newsfragment as well, maybe something like
```suggestion
Allow pickling `trio.Cancelled`, as they can show up when you want to pickle something else. Also enables pickling for other ``NoPublicConstructor`` objects.
```

jakkdl (approved):
just a couple small things, otherwise looks as good as it could given the messiness of metaclasses :)

src/trio/_util.py
```diff
@@ -327,6 +327,21 @@ def __call__(cls, *args: object, **kwargs: object) -> None:
...
+            if old_reduce is not None:
+                result = old_reduce(self)
+                assert not isinstance(result, str)
+            else:
+                result = super(type(self), self).__reduce__()  # type: ignore[misc]
+
+            return (type(self)._create, *result[1:])  # type: ignore[attr-defined]
+
+        namespace["__reduce__"] = __reduce__
+        return super().__new__(cls, name, bases, namespace, **kwargs)
```
jakkdl, line 343: I'm a bit unsure if just tossing out the first object in the tuple is the right thing to do. But I'm not sure what else to do, so maybe just add a comment; or something like `assert result[0] is type(self)` (in the `if old_reduce is not None` branch) to check nobody's doing anything weird.

src/trio/_tests/test_util.py
```diff
@@ -198,11 +198,16 @@ def __init__(self, a: int, b: float) -> None:
...
+            return ("dummy", ())
+
     with pytest.raises(TypeError):
         SpecialClass(8, 3.15)
 
     # Private constructor should not raise, and passes args to __init__.
-    assert isinstance(SpecialClass._create(8, b=3.15), SpecialClass)
+    inst = SpecialClass._create(8, b=3.15)
+    assert isinstance(inst, SpecialClass)
+    assert inst.__reduce__()[0] == SpecialClass._create
```
jakkdl, line 210: might as well check that the other args are handled correctly as well
```suggestion
    inst = SpecialClass._create(8, b=3.15)
    assert isinstance(inst, SpecialClass)
    reduced = inst.__reduce__()
    assert isinstance(reduced[0](*reduced[1]), SpecialClass)
```

src/trio/_tests/test_util.py
```diff
@@ -198,11 +198,16 @@ def __init__(self, a: int, b: float) -> None:
             assert a == 8
             assert b == 3.15
 
+        def __reduce__(self) -> tuple[Any, ...]:  # type: ignore[explicit-any]
+            return ("dummy", ())
```
jakkdl, line 202:
```suggestion
        def __reduce__(self) -> tuple[Any, ...]:  # type: ignore[explicit-any]
            return ("dummy", (8, 3.15))
```

jakkdl (commented):
okay I looked a bit more thorough now and found some small things :)

src/trio/_tests/test_util.py
A5rocks: I guess at that point we can just use `pickle`.

src/trio/_util.py
A5rocks: Actually I think it's better to give up if there's already `__reduce__`.

A5rocks:
Turns out I underestimated pickling. It's more complicated than I expected and I don't think there's a way to generalize support to *any* class. I was operating in a world where `Exception.__reduce__` was the default but `object.__reduce__` is really weird (try it yourself!). In fact, I would need to assume what `__set_state__` is doing and have special casing for if it's a dict, and compare the first element to `copyreg._reconstructor` to know whether to replace the first element, and whatever. It's not worth the complexity, so I'm going to undo generalizing this. Nobody should be `pickle`ing these anyways!

src/trio/_util.py
jakkdl: any reason to keep the new `__new__` at all?

A5rocks: Just me forgetting that I added it in this PR. I'll remove it!

CoolCat467 (approved):

jakkdl (approved):
sorry for the detour ^^; Back to nice and simple

PR #3256, opened 2025-04-22
Add message with debugging info to Cancelled

### LINKED ISSUE #3232: Add exception message for easier debugging with `Cancelled`
Occasionally we end up digging through logs trying to work out why something was cancelled, and eventually it occurred to me that Trio could help with this.  I don't have a complete design proposal, but it would be great to handle each of three cases:

- a deadline was reached.
- a sibling task in the `Nursery` raised an exception, including the name of the erroring task
- someone called `CancelScope.cancel()`, including the name of the task in which this was called
  - add a new `reason: str | None = None` param, so that users can explicitly annotate why they cancelled something

I think we could represent this as a string argument to `Cancelled` like any other exception message; the only tricky part will be threading it through all the cancellation-delivery machinery 🙂 

Comments and design input most welcome, of course!

jakkdl:
fixes #3232 by adding a `CancelReason` that's tracked by both `CancelScope` and `CancelStatus` and then used when constructing a `Cancelled` to be raised.

Remaining stuff:
- [x] I'm not 100% sure that both `CancelScope` and `CancelStatus` needs to have a `CancelReason` member - initially only `CancelStatus` had it but that wasn't enough, but I haven't tried to see if I can get away with only `CancelScope` having it.
  - It's now only tracked in `CancelScope`
- [x] I also found a bug as I was reviewing and adding a final test, where manually setting `cs.deadline=-inf` and then checkpointing somehow ends up as an `explicit` cancelled.
- [x] `from_thread_check_cancelled` needs some special handling
  - I managed to fix this.. but in a way that created cyclic garbage. So need a better fix

### DIFF
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -17,7 +17,8 @@ from outcome import Error, Value
 import trio
 
 from ._abc import ReceiveChannel, ReceiveType, SendChannel, SendType, T
-from ._core import Abort, RaiseCancelT, Task, enable_ki_protection
+from ._core import Abort, RaiseCancelT, Task, current_task, enable_ki_protection
+from ._core._run import CancelReason
 from ._util import (
     MultipleExceptionError,
     NoPublicConstructor,
@@ -547,7 +548,12 @@ def as_safe_channel(
                     yield wrapped_recv_chan
                 # User has exited context manager, cancel to immediately close the
                 # abandoned generator if it's still alive.
-                nursery.cancel_scope.cancel(reason="user exited context manager")
+                nursery.cancel_scope._cancel_reason = CancelReason(
+                    reason="exited _as_safe_channel context manager",
+                    source="shutdown",
+                    source_task=repr(current_task),
+                )
+                nursery.cancel_scope.cancel()
         except BaseExceptionGroup as eg:
             try:
                 raise_single_exception_from_group(eg)
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -17,8 +17,7 @@ from outcome import Error, Value
 import trio
 
 from ._abc import ReceiveChannel, ReceiveType, SendChannel, SendType, T
-from ._core import Abort, RaiseCancelT, Task, current_task, enable_ki_protection
-from ._core._run import CancelReason
+from ._core import Abort, RaiseCancelT, Task, enable_ki_protection
 from ._util import (
     MultipleExceptionError,
     NoPublicConstructor,
@@ -548,12 +547,8 @@ def as_safe_channel(
                     yield wrapped_recv_chan
                 # User has exited context manager, cancel to immediately close the
                 # abandoned generator if it's still alive.
-                nursery.cancel_scope._cancel(
-                    CancelReason(
-                        reason="exited trio.as_safe_channel context manager",
-                        source="shutdown",
-                        source_task=repr(current_task),
-                    )
+                nursery.cancel_scope.cancel(
+                    "exited trio.as_safe_channel context manager"
                 )
         except BaseExceptionGroup as eg:
             try:
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -548,12 +548,13 @@ def as_safe_channel(
                     yield wrapped_recv_chan
                 # User has exited context manager, cancel to immediately close the
                 # abandoned generator if it's still alive.
-                nursery.cancel_scope._cancel_reason = CancelReason(
-                    reason="exited _as_safe_channel context manager",
-                    source="shutdown",
-                    source_task=repr(current_task),
+                nursery.cancel_scope._cancel(
+                    CancelReason(
+                        reason="exited trio.as_safe_channel context manager",
+                        source="shutdown",
+                        source_task=repr(current_task),
+                    )
                 )
-                nursery.cancel_scope.cancel()
         except BaseExceptionGroup as eg:
             try:
                 raise_single_exception_from_group(eg)
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -547,7 +547,7 @@ def as_safe_channel(
                     yield wrapped_recv_chan
                 # User has exited context manager, cancel to immediately close the
                 # abandoned generator if it's still alive.
-                nursery.cancel_scope.cancel()
+                nursery.cancel_scope.cancel(reason="user exited context manager")
         except BaseExceptionGroup as eg:
             try:
                 raise_single_exception_from_group(eg)
--- a/src/trio/_core/_asyncgens.py
+++ b/src/trio/_core/_asyncgens.py
@@ -230,7 +230,9 @@ class AsyncGenerators:
             # with an exception, not even a Cancelled. The inside
             # is cancelled so there's no deadlock risk.
             with _core.CancelScope(shield=True) as cancel_scope:
-                cancel_scope.cancel()
+                cancel_scope.cancel(
+                    reason="internal trio cancellation during finalization"
+                )
                 await agen.aclose()
         except BaseException:
             ASYNCGEN_LOGGER.exception(
--- a/src/trio/_core/_asyncgens.py
+++ b/src/trio/_core/_asyncgens.py
@@ -231,7 +231,7 @@ class AsyncGenerators:
             # is cancelled so there's no deadlock risk.
             with _core.CancelScope(shield=True) as cancel_scope:
                 cancel_scope.cancel(
-                    reason="internal trio cancellation during finalization"
+                    reason="disallow async work when closing async generators during trio shutdown"
                 )
                 await agen.aclose()
         except BaseException:
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from functools import partial
+from typing import TYPE_CHECKING, Literal
+
+import attrs
 
 from trio._util import NoPublicConstructor, final
 
@@ -34,6 +37,7 @@ class WouldBlock(Exception):
 
 
 @final
+@attrs.define(eq=False, kw_only=True)
 class Cancelled(BaseException, metaclass=NoPublicConstructor):
     """Raised by blocking calls if the surrounding scope has been cancelled.
 
@@ -67,11 +71,30 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     """
 
+    source: Literal["deadline", "nursery", "explicit"]
+    source_task: str | None = None
+    reason: str | None = None
+
     def __str__(self) -> str:
-        return "Cancelled"
+        return (
+            f"Cancelled due to {self.source}"
+            + (f" with reason {self.reason!r}" if self.reason is not None else "")
+            + f" from task {self.source_task}"
+        )
 
     def __reduce__(self) -> tuple[Callable[[], Cancelled], tuple[()]]:
-        return (Cancelled._create, ())
+        # the `__reduce__` tuple does not support kwargs, so we must use partial
+        # for non-default args
+        # or switch to allow posarg (?)
+        return (
+            partial(
+                Cancelled._create,
+                source=self.source,
+                source_task=self.source_task,
+                reason=self.reason,
+            ),
+            (),
+        )
 
 
 class BusyResourceError(Exception):
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -10,7 +10,16 @@ from trio._util import NoPublicConstructor, final
 if TYPE_CHECKING:
     from collections.abc import Callable
 
-    from typing_extensions import Self
+    from typing_extensions import Self, TypeAlias
+
+CancelReasonLiteral: TypeAlias = Literal[
+    "KeyboardInterrupt",
+    "deadline",
+    "explicit",
+    "nursery",
+    "shutdown",
+    "unknown",
+]
 
 
 class TrioInternalError(Exception):
@@ -73,9 +82,7 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     """
 
-    source: Literal[
-        "KeyboardInterrupt", "deadline", "explicit", "nursery", "shutdown", "unknown"
-    ]
+    source: CancelReasonLiteral
     # repr(Task), so as to avoid gc troubles from holding a reference
     source_task: str | None = None
     reason: str | None = None
@@ -107,14 +114,7 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
         def _create(
             cls,
             *,
-            source: Literal[
-                "KeyboardInterrupt",
-                "deadline",
-                "explicit",
-                "nursery",
-                "shutdown",
-                "unknown",
-            ],
+            source: CancelReasonLiteral,
             source_task: str | None = None,
             reason: str | None = None,
         ) -> Self: ...
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -81,22 +81,16 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
     reason: str | None = None
 
     def __str__(self) -> str:
-        def repr_if_not_none(lead: str, s: str | None, do_repr: bool = False) -> str:
-            if s is None:
-                return ""
-            if do_repr:
-                return lead + repr(s)
-            return lead + s
-
         return (
             f"cancelled due to {self.source}"
-            + repr_if_not_none(" with reason ", self.reason, True)
-            + repr_if_not_none(" from task ", self.source_task)
+            + ("" if self.reason is None else f" with reason {self.reason!r}")
+            + ("" if self.source_task is None else f" from task {self.source_task}")
         )
 
     def __reduce__(self) -> tuple[Callable[[], Cancelled], tuple[()]]:
-        # the `__reduce__` tuple does not support kwargs, so we must use partial
-        # for non-default args
+        # The `__reduce__` tuple does not support directly passing kwargs, and the
+        # kwargs are required so we can't use the third item for adding to __dict__,
+        # so we use partial.
         return (
             partial(
                 Cancelled._create,
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -73,27 +73,30 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     """
 
-    source: Literal["deadline", "nursery", "explicit", "unknown", "KeyboardInterrupt"]
-    # TODO: this should probably be a Task?
+    source: Literal[
+        "KeyboardInterrupt", "deadline", "explicit", "nursery", "shutdown", "unknown"
+    ]
+    # repr(Task), so as to avoid gc troubles from holding a reference
     source_task: str | None = None
     reason: str | None = None
 
     def __str__(self) -> str:
-        def repr_if_not_none(lead: str, s: str | None, trail: str = "") -> str:
+        def repr_if_not_none(lead: str, s: str | None, do_repr: bool = False) -> str:
             if s is None:
                 return ""
-            return lead + s + trail
+            if do_repr:
+                return lead + repr(s)
+            return lead + s
 
         return (
             f"cancelled due to {self.source}"
-            + repr_if_not_none(" with reason '", self.reason, "'")
+            + repr_if_not_none(" with reason ", self.reason, True)
             + repr_if_not_none(" from task ", self.source_task)
         )
 
     def __reduce__(self) -> tuple[Callable[[], Cancelled], tuple[()]]:
         # the `__reduce__` tuple does not support kwargs, so we must use partial
         # for non-default args
-        # or switch to allow posarg (?)
         return (
             partial(
                 Cancelled._create,
@@ -111,7 +114,12 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
             cls,
             *,
             source: Literal[
-                "deadline", "nursery", "explicit", "unknown", "KeyboardInterrupt"
+                "KeyboardInterrupt",
+                "deadline",
+                "explicit",
+                "nursery",
+                "shutdown",
+                "unknown",
             ],
             source_task: str | None = None,
             reason: str | None = None,
--- a/src/trio/_core/_exceptions.py
+++ b/src/trio/_core/_exceptions.py
@@ -10,6 +10,8 @@ from trio._util import NoPublicConstructor, final
 if TYPE_CHECKING:
     from collections.abc import Callable
 
+    from typing_extensions import Self
+
 
 class TrioInternalError(Exception):
     """Raised by :func:`run` if we encounter a bug in Trio, or (possibly) a
@@ -71,15 +73,21 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     """
 
-    source: Literal["deadline", "nursery", "explicit"]
+    source: Literal["deadline", "nursery", "explicit", "unknown", "KeyboardInterrupt"]
+    # TODO: this should probably be a Task?
     source_task: str | None = None
     reason: str | None = None
 
     def __str__(self) -> str:
+        def repr_if_not_none(lead: str, s: str | None, trail: str = "") -> str:
+            if s is None:
+                return ""
+            return lead + s + trail
+
         return (
-            f"Cancelled due to {self.source}"
-            + (f" with reason {self.reason!r}" if self.reason is not None else "")
-            + f" from task {self.source_task}"
+            f"cancelled due to {self.source}"
+            + repr_if_not_none(" with reason '", self.reason, "'")
+            + repr_if_not_none(" from task ", self.source_task)
         )
 
     def __reduce__(self) -> tuple[Callable[[], Cancelled], tuple[()]]:
@@ -95,6 +103,19 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
             ),
             (),
         )
+
+    if TYPE_CHECKING:
+        # for type checking on internal code
+        @classmethod
+        def _create(
+            cls,
+            *,
+            source: Literal[
+                "deadline", "nursery", "explicit", "unknown", "KeyboardInterrupt"
+            ],
+            source_task: str | None = None,
+            reason: str | None = None,
+        ) -> Self: ...
 
 
 class BusyResourceError(Exception):
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -1648,18 +1648,7 @@ class Task(metaclass=NoPublicConstructor):  # type: ignore[explicit-any]
         if not self._cancel_status.effectively_cancelled:
             return
 
-        def raise_cancel() -> NoReturn:
-            reason = self._cancel_status._scope._cancel_reason
-            if reason is None:
-                raise Cancelled._create(source="unknown", reason="misnesting")
-
-            raise Cancelled._create(
-                source=reason.source,
-                reason=reason.reason,
-                source_task=reason.source_task,
-            )
-
-        self._attempt_abort(raise_cancel)
+        self._attempt_abort(RaiseCancel(self._cancel_status._scope._cancel_reason))
 
     def _attempt_delivery_of_pending_ki(self) -> None:
         assert self._runner.ki_pending
@@ -1671,6 +1660,24 @@ class Task(metaclass=NoPublicConstructor):  # type: ignore[explicit-any]
             raise KeyboardInterrupt
 
         self._attempt_abort(raise_cancel)
+
+
+class RaiseCancel:
+    def __init__(self, reason: CancelReason | None) -> None:
+        if reason is None:
+            self.cancelled = Cancelled._create(source="unknown", reason="misnesting")
+        else:
+            self.cancelled = Cancelled._create(
+                source=reason.source,
+                reason=reason.reason,
+                source_task=reason.source_task,
+            )
+
+    def __call__(self) -> NoReturn:
+        try:
+            raise self.cancelled
+        finally:
+            del self.cancelled
 
 
 ################################################################
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -318,8 +318,14 @@ class Deadlines:
 
 @attrs.define
 class CancelReason:
-    # TODO: loren ipsum
-    source: Literal["deadline", "nursery", "explicit", "KeyboardInterrupt"]
+    """Attached to a :class:`CancelScope` upon cancellation with details of the source of the
+    cancellation, which is then used to construct the string in a :exc:`Cancelled`.
+    Users can pass a ``reason`` str to :meth:`CancelScope.cancel` to set it.
+    """
+
+    source: Literal[
+        "KeyboardInterrupt", "deadline", "explicit", "nursery", "shutdown", "unknown"
+    ]
     source_task: str | None = None
     reason: str | None = None
 
@@ -580,8 +586,6 @@ class CancelScope:
     _cancel_called: bool = attrs.field(default=False, init=False)
     cancelled_caught: bool = attrs.field(default=False, init=False)
 
-    # necessary as cancel_status might be None
-    # TODO: but maybe cancel_status doesn't need it?
     _cancel_reason: CancelReason | None = attrs.field(
         default=None, init=False, repr=True
     )
@@ -1230,9 +1234,10 @@ class Nursery(metaclass=NoPublicConstructor):
         "(`~trio.lowlevel.Task`):  The Task that opened this nursery."
         return self._parent_task
 
-    def _add_exc(self, exc: BaseException) -> None:
+    def _add_exc(self, exc: BaseException, reason: CancelReason | None) -> None:
         self._pending_excs.append(exc)
-        # TODO: source/reason?
+        if self.cancel_scope._cancel_reason is None:
+            self.cancel_scope._cancel_reason = reason
         self.cancel_scope.cancel()
 
     def _check_nursery_closed(self) -> None:
@@ -1249,11 +1254,14 @@ class Nursery(metaclass=NoPublicConstructor):
     ) -> None:
         self._children.remove(task)
         if isinstance(outcome, Error):
-            if self.cancel_scope._cancel_reason is None:
-                self.cancel_scope._cancel_reason = CancelReason(
-                    source="nursery", source_task=repr(task)
-                )
-            self._add_exc(outcome.error)
+            self._add_exc(
+                outcome.error,
+                CancelReason(
+                    source="nursery",
+                    source_task=repr(task),
+                    reason=f"child task raised exception {outcome.error!r}",
+                ),
+            )
         self._check_nursery_closed()
 
     async def _nested_child_finished(
@@ -1263,7 +1271,14 @@ class Nursery(metaclass=NoPublicConstructor):
         # Returns ExceptionGroup instance (or any exception if the nursery is in loose mode
         # and there is just one contained exception) if there are pending exceptions
         if nested_child_exc is not None:
-            self._add_exc(nested_child_exc)
+            self._add_exc(
+                nested_child_exc,
+                reason=CancelReason(
+                    source="nursery",
+                    source_task=repr(self._parent_task),
+                    reason=f"Code block inside nursery contextmanager raised exception {nested_child_exc!r}",
+                ),
+            )
         self._nested_child_running = False
         self._check_nursery_closed()
 
@@ -1274,7 +1289,13 @@ class Nursery(metaclass=NoPublicConstructor):
             def aborted(raise_cancel: _core.RaiseCancelT) -> Abort:
                 exn = capture(raise_cancel).error
                 if not isinstance(exn, Cancelled):
-                    self._add_exc(exn)
+                    self._add_exc(
+                        exn,
+                        CancelReason(
+                            source="KeyboardInterrupt",
+                            source_task=repr(self._parent_task),
+                        ),
+                    )
                 # see test_cancel_scope_exit_doesnt_create_cyclic_garbage
                 del exn  # prevent cyclic garbage creation
                 return Abort.FAILED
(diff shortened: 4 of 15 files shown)


CoolCat467:
So far looking good, I didn't see anything immediately that I would suggest changing.

Curious to hear about your immediate ideas about usecases for this

jakkdl:
> Curious to hear about your immediate ideas about usecases for this

see #3232 - if you're encountering a bunch of `Cancelled` and don't know why, having more info attached to them should make debugging easier

jakkdl:
Fixed most of the problems, only remaining part is getting the right incantation for `_attempt_delivery_of_any_pending_cancel` that manages to keep the `CancelReason` alive in other threads, but not causing any cyclic garbage. Slapping another attribute onto `trio._threads.PARENT_TASK_DATA` might work, but it feels like it *should* be possible without it.
I've tried different permutations of `finally: del [raise_cancel/cancelled]`, but need to dig deeper into understanding the actual problem unless somebody else want to chime in. @graingert?

graingert:
I'll have a look, does fixing Outcome to delete it's values fix it?

src/trio/_channel.py
```diff
@@ -547,7 +547,7 @@ async def context_manager(
                     yield wrapped_recv_chan
                 # User has exited context manager, cancel to immediately close the
                 # abandoned generator if it's still alive.
-                nursery.cancel_scope.cancel()
+                nursery.cancel_scope.cancel(reason="user exited context manager")
```
Zac-HD, line 550:
```suggestion
                nursery.cancel_scope.cancel(reason="exited as_safe_channel context manager")
```
Saying the user exited is either vacuous or possibly-false, but noting _which_ context manager might help.

src/trio/_core/_asyncgens.py
```diff
@@ -230,7 +230,9 @@ async def _finalize_one(
             # with an exception, not even a Cancelled. The inside
             # is cancelled so there's no deadlock risk.
             with _core.CancelScope(shield=True) as cancel_scope:
-                cancel_scope.cancel()
+                cancel_scope.cancel(
+                    reason="internal trio cancellation during finalization"
```
Zac-HD, line 234:
```suggestion
                    reason="disallow async work when closing async generators during trio shutdown"
```
it's a mouthful, but I want to be clear about what's going on.

src/trio/_core/_exceptions.py
```diff
@@ -67,11 +73,49 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
...
     def __str__(self) -> str:
-        return "Cancelled"
+        def repr_if_not_none(lead: str, s: str | None, trail: str = "") -> str:
+            if s is None:
+                return ""
+            return lead + s + trail
+
+        return (
+            f"cancelled due to {self.source}"
+            + repr_if_not_none(" with reason '", self.reason, "'")
```
Zac-HD, line 89:
```suggestion
            + repr_if_not_none(" with reason ", repr(self.reason))
```
(imagine if the reason contains `'`s?)

src/trio/_core/_exceptions.py
```diff
@@ -67,11 +73,49 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
...
+            + repr_if_not_none(" from task ", self.source_task)
+        )
 
     def __reduce__(self) -> tuple[Callable[[], Cancelled], tuple[()]]:
-        return (Cancelled._create, ())
+        # the `__reduce__` tuple does not support kwargs, so we must use partial
+        # for non-default args
+        # or switch to allow posarg (?)
```
Zac-HD, line 96: I'd rather have this once in `__reduce__` than worse signatures at every call site.

src/trio/_core/_exceptions.py
```diff
@@ -67,11 +73,49 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     """
 
+    source: Literal["deadline", "nursery", "explicit", "unknown", "KeyboardInterrupt"]
+    # TODO: this should probably be a Task?
+    source_task: str | None = None
```
Zac-HD, line 78: What are the pros and cons?  Off the top of my head, better access to the other Task attributes vs keeping tasks alive for longer is worse for GC.

src/trio/_core/_run.py
```diff
@@ -883,7 +912,7 @@ def shield(self, new_value: bool) -> None:
             self._cancel_status.recalculate()
 
     @enable_ki_protection
-    def cancel(self) -> None:
+    def cancel(self, *, reason: str | None = None) -> None:
         """Cancels this scope immediately.
```
Zac-HD, line 916: I'd allow posargs here, since there's only one argument.  Also needs `reason=` to be explained in the docstring, e.g.

> The optional ``reason`` argument accepts a string, which will be attached to any resulting ``Cancelled`` exception, to help you understand where that cancellation is coming from and why it happened.

src/trio/_core/_run.py
```diff
@@ -1194,6 +1232,7 @@ def parent_task(self) -> Task:
 
     def _add_exc(self, exc: BaseException) -> None:
         self._pending_excs.append(exc)
+        # TODO: source/reason?
         self.cancel_scope.cancel()
```
Zac-HD, line 1236: I think this is adequately handled below; but I'd add a comment pointing to `_child_finished` to say so.

src/trio/_core/_run.py
```diff
@@ -2075,6 +2133,7 @@ async def init(
                     )
 
                 # Main task is done; start shutting down system tasks
+                # TODO: source/reason?
```
Zac-HD, line 2136: Hmm, maybe we should have a `trio_shutdown` source that we can use here and for asyncgen shutdown?

src/trio/_subprocess.py
```diff
@@ -766,6 +766,7 @@ async def killer() -> None:
 
                 nursery.start_soon(killer)
                 await proc.wait()
+                # TODO: source/reason?
                 killer_cscope.cancel()
```
Zac-HD, line 770: I think I'd just leave this one alone; it's very much an implementation detail and will be caught by the shielded scope before any user sees it.

Zac-HD (commented):
I'm really looking forward to using this, and to my colleagues having less to complain about when large complicated things get cancelled 😅

Many small comments below, but they're basically all tactical/local things, and I agree that once we get the gc stuff sorted out this will be basically ready to go.  Thanks!

src/trio/_core/_exceptions.py
A5rocks: `__reduce__` can return more than just the two elements -- the third item is either a tuple (in which case it gets passed to `__setstate__`) or a dict (in which it gets added to `__dict__`).

Either of those options should work.

jakkdl: > `__reduce__` can return more than just the two elements -- the third item is either a tuple (in which case it gets passed to `__setstate__`) or a dict (in which it gets added to `__dict__`).
> 
> Either of those options should work.

neither does because he have a required kwarg. If it's tuple we're passing posarg, if it's a dict we're modifying attributes after initializing - but `source` is required.

Normally people would get around this by using [`__getnewargs_ex__`](https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__) but since we *also* need `__reduce__` to specify the callable object we either have to use both, or do partial+`__reduce__`.

(not fully confident this is 100% correct, this is all messy)

TeamSpen210: We control both the callable passed, and how `__setstate__` works, and both are effectively private API. The tuple could just have a variable number of elements, and then we pass a private module-level function to do the unpickling?

jakkdl:
> I'll have a look, does fixing Outcome to delete its values fix it?

nope

src/trio/_core/_run.py
jakkdl: I'm not sure about the `trio_` prefix - if it's to indicate "trio itself is shutting down" then it should not be used for asyncgen shutdown; if it's to indicate "the source is from within trio" then most of the other sources should also have it, and only really makes sense within a context of the user adding a bunch of their own sources...

which we *could* allow, you can currently do
```python
my_cs._cancel_reason = CancelReason(source="foo", reason="bar", source_task="baz")
my_cs.cancel()
```
we could also open up more parameters to `.cancel()`.

but giving an extensive message to `reason` should be enough for most cases tbh.

src/trio/_core/_run.py
jakkdl: ah, it isn't. Also need to handle `_nested_child_finished`, if the code inside the context manager raises.

src/trio/_core/_exceptions.py
jakkdl: well, `partial` inside `__reduce__` works fine, so I don't think we need to bother with more complicated stuff.

jakkdl:
Okay I found a solution to gc+threads in https://github.com/python-trio/trio/pull/3256/commits/69c398bc9b1332f95fcb7fbedc9ffb123f86d1fb, not the prettiest but it seems to work at least? Very open to there being better ways of doing it
1. I did not find a good way of binding a `CancelReason` or `Cancelled` object to `raise_cancel` as a function ... but by making it a callable class it was possible without `_attempt_delivery_of_any_pending_cancel` holding any references after passing it to `_attempt_abort`.
2. I needed `try: raise self.cancelled; finally: del self.cancelled` to pass gc checks, but that led to crashes in thread tests as the same cancel func appears to get reused across multiple threads. The only way I figured out how to resolve this was to `copy.copy` the `RaiseCancel` object in `_threads.py`'s `abort` function. I feel like there should be a cleaner way of doing this, but the object is small so copying it should be fairly cheap.

src/trio/_core/_exceptions.py
```diff
@@ -73,27 +73,30 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     """
 
-    source: Literal["deadline", "nursery", "explicit", "unknown", "KeyboardInterrupt"]
-    # TODO: this should probably be a Task?
+    source: Literal[
+        "KeyboardInterrupt", "deadline", "explicit", "nursery", "shutdown", "unknown"
+    ]
```
Zac-HD, line 78: Since we reference this twice, let's pull it out into a global `_CancelSource: TypeAlias = Literal[...]`, which we can also refer to in `_run.py`.

src/trio/_core/_run.py
Zac-HD: Agree, let's just remove this TODO comment, and the one below.

Zac-HD (approved):
minor tweaks only; merge when ready!

src/trio/_core/_exceptions.py
```diff
@@ -67,11 +73,57 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
...
+                Cancelled._create,
+                source=self.source,
+                source_task=self.source_task,
+                reason=self.reason,
+            ),
+            (),
+        )
+
+    if TYPE_CHECKING:
+        # for type checking on internal code
```
A5rocks, line 111: I'm not sure this will remain supported, and this is soooo evil, but you can do this (to avoid having two places to keep up to date):

```py
import dataclasses
import typing

P = typing.ParamSpec("P")
def update_sig(f: typing.Callable[typing.Concatenate[typing.Never, P], None]) -> typing.Callable[P, "X"]:
    raise AssertionError("never called")

@dataclasses.dataclass
class X:
    x: int
    
    if typing.TYPE_CHECKING:
        _x = update_sig(X.__init__)

X._x("4")  # E: Argument 1 has incompatible type "str"; expected "int"
reveal_type(X._x(5))  # N: Revealed type is "__main__.X"
```

it works with mypy at least (hopefully it works with mypy's `attrs` support too?)

Up to you whether to use that though.

src/trio/_core/_run.py
```diff
@@ -558,6 +588,10 @@ class CancelScope:
     _cancel_called: bool = attrs.field(default=False, init=False)
     cancelled_caught: bool = attrs.field(default=False, init=False)
 
+    _cancel_reason: CancelReason | None = attrs.field(
+        default=None, init=False, repr=True
```
A5rocks, line 592: Isn't `repr=True` the default?

src/trio/_core/_run.py
```diff
@@ -468,10 +486,22 @@ def recalculate(self) -> None:
...
+                        current._parent._scope._cancel_reason
+                    )
                 current.effectively_cancelled = new_state
                 if new_state:
                     for task in current._tasks:
                         task._attempt_delivery_of_any_pending_cancel()
+                    # current._cancel_reason will not be None, unless there's misnesting
+                    for child in current._children:
+                        if child._scope._cancel_reason is None:
+                            child._scope._cancel_reason = current._scope._cancel_reason
```
A5rocks, line 504: Why do we both propagate down here and then copy the parent's reason above? Can't we just do either of these?

src/trio/_core/_run.py
```diff
@@ -883,16 +918,28 @@ def shield(self, new_value: bool) -> None:
...
+        cancellation is coming from and why it happened.
+
         This method is idempotent, i.e., if the scope was already
         cancelled then this method silently does nothing.
         """
         if self._cancel_called:
             return
         with self._might_change_registered_deadline():
             self._cancel_called = True
+        if self._cancel_reason is None:
```
A5rocks, line 935: Should we set `source_task` outside of this `if` statement, because I see some other `_cancel_reason = CancelReason(...)` calls don't have it.

src/trio/_core/_run.py
```diff
@@ -1192,8 +1240,10 @@ def parent_task(self) -> Task:
         "(`~trio.lowlevel.Task`):  The Task that opened this nursery."
         return self._parent_task
 
-    def _add_exc(self, exc: BaseException) -> None:
+    def _add_exc(self, exc: BaseException, reason: CancelReason | None) -> None:
```
A5rocks, line 1243: If the only callers of this are internal, IMO it would be cleaner to have them set the reason inline. Also, to avoid multiple comments for similar things, why doesn't this unconditionally set `reason` if it isn't `None`?

src/trio/_core/_run.py
```diff
@@ -1210,7 +1260,14 @@ def _child_finished(
     ) -> None:
         self._children.remove(task)
         if isinstance(outcome, Error):
-            self._add_exc(outcome.error)
+            self._add_exc(
+                outcome.error,
+                CancelReason(
+                    source="nursery",
+                    source_task=repr(task),
```
A5rocks, line 1267: If we unconditionally set `source_task` in `.cancel()` we can avoid this duplication!

A5rocks (commented):
Here's some comments, mostly trivial/nitpicks. I haven't looked to the end yet since I don't immediately like some stuff I saw and I want to think about whether that's justified or not!

src/trio/_core/_exceptions.py
A5rocks: Actually this works:

```py
import dataclasses
from typing import Protocol, TypeVar, ParamSpec, Generic, Callable

T = TypeVar("T", covariant=True)
P = ParamSpec("P")

class Meta(type):
    def __call__(cls, *args: object, **kwargs: object) -> None:
        raise TypeError(
            f"{cls.__module__}.{cls.__qualname__} has no public constructor",
        )

    def _create(cls: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> type[T]:

@dataclasses.dataclass
class X(metaclass=Meta):
    x: int


reveal_type(X._create)  # N: Revealed type is "def (x: builtins.int) -> type[__main__.X]"
```

so we could make `NoPublicConstructor` support this generally

jakkdl: whoop, nice~

src/trio/_core/_run.py
jakkdl: no this one is critically important. Several other cancellations internally use `cs.cancel()`, so if this one is not guarded by `if` then a lot of other cancel reasons will always get overridden.

What might be more sketchy is others not having it, I should maybe add a setter or helper function to `CancelScope`

src/trio/_core/_run.py
jakkdl: > why doesn't this unconditionally set _cancel_reason = reason if it isn't None?

in case we get multiple sources of cancellation I don't want to override the first one. In other places it's more critical, but here I could see a scenario where:
1. Something causes a cancellation, be it a deadline or a crashing task or whatever
2. a different task B gets cancelled, but they have an `except Cancelled`, and inside that handler they raise a different exception
3. without `if self.cancel_scope._cancel_reason is None:` the cause would now get set to task B raising an exception

so I'm pretty sure we need the `if`, which means we'd need to write the `if` three times if we did it in-line

src/trio/_core/_exceptions.py
```diff
@@ -67,11 +73,57 @@ class Cancelled(BaseException, metaclass=NoPublicConstructor):
 
     """
 
+    source: Literal[
+        "KeyboardInterrupt", "deadline", "explicit", "nursery", "shutdown", "unknown"
+    ]
+    # repr(Task), so as to avoid gc troubles from holding a reference
+    source_task: str | None = None
```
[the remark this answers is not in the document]
jakkdl: if a `Task` contains a `CancelScope` and the scope gets cancelled within the same task, the scope will then have a strong ref to a `CancelReason` which will then point back to the `Task`. I think?

the `repr(Task)` on its own is perhaps not super useful, but in case you have multiple cancellations going on at the same time that are only distinguished by the source task then you can visually distinguish them even without other sources of the task id.
Though it does also contain the name of the function itself:
`<Task 'trio._core._tests.test_run.test_Cancelled_str' at 0x\w*>`
which could be very helpful if you have different functions spawned in a nursery.

src/trio/_core/_run.py
jakkdl: Oh you are very right. Deleting the former broke tests, but deleting the latter did not.

src/trio/_core/_run.py
jakkdl: actually, what makes more sense is having an internal `def _cancel` because we should only ever set `_cancel_reason` when doing a cancellation

src/trio/_core/_run.py
jakkdl: not relevant anymore

src/trio/_core/_run.py
jakkdl: I'm not sure what the duplication is here that you wanted to avoid?

src/trio/_core/_exceptions.py
jakkdl: not relevant with typing added to `NoPublicConstructor` :rocket:

src/trio/_core/_exceptions.py
jakkdl:
```
src/trio/_core/_local.py:27: error: Argument 1 to "_create" of "NoPublicConstructor" has incompatible type "RunVar[T]"; expected "Never"  [arg-type]
src/trio/_core/_run.py:1989: error: Argument "coro" to "_create" of "NoPublicConstructor" has incompatible type "Coroutine[object, Never, object]"; expected "CoroutineType[Any, Outcome[object], Any]"  [arg-type]
src/trio/_channel.py:266: error: Argument 1 to "_create" of "NoPublicConstructor" has incompatible type "MemoryChannelState[SendType]"; expected "Never"  [arg-type]
src/trio/_channel.py:418: error: Argument 1 to "_create" of "NoPublicConstructor" has incompatible type "MemoryChannelState[ReceiveType]"; expected "Never"  [arg-type]
src/trio/_subprocess.py:430: error: Argument 1 to "_create" of "NoPublicConstructor" has incompatible type "Popen[str]"; expected "Popen[bytes]"  [arg-type]
src/trio/testing/_fake_net.py:133: error: Argument 2 to "_create" of "NoPublicConstructor" has incompatible type "int"; expected "AddressFamily"  [arg-type]
src/trio/testing/_fake_net.py:133: error: Argument 3 to "_create" of "NoPublicConstructor" has incompatible type "int"; expected "SocketKind"  [arg-type]
```
hrmmm, this smells like a separate PR

jakkdl:
pypy fail is unrelated
```
+ python -m pip install -U pip uv -c test-requirements.txt
Install dependencies
  Requirement already satisfied: pip in c:\hostedtoolcache\windows\pypy\3.10.16\x86\lib\site-packages (25.1)
  ERROR: Cannot install uv because these package versions have conflicting dependencies.
  
  The conflict is caused by:
      The user requested uv
      The user requested (constraint) uv==0.7.2
  
  To fix this you could try to:
  1. loosen the range of package versions you've specified
  2. remove package versions to allow pip to attempt to solve the dependency conflict
  
  ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
  Error: Process completed with exit code 1.
```

src/trio/_core/_exceptions.py
jakkdl: opened #3260

CoolCat467:
I've seen that pypy failure before, and it usually means that it's using an outdated package cache. Clearing package cache will fix it (though I don't know how to do that myself)

Zac-HD:
usually clicking the 🗑️ icons in https://github.com/python-trio/trio/actions/caches

src/trio/_core/_run.py
A5rocks: > Several other cancellations internally use `cs.cancel()`, so if this one is not guarded by `if` then a lot of other cancel reasons will always get overridden.

I'm not entirely convinced avoiding this is a good thing:

```py
with trio.move_on_after(1):
  with trio.CancelScope() as cs:
    try:
      try:
        await trio.sleep(5)
      except trio.Cancelled as c1:
        cs.cancel("shutting down")
        await trio.sleep(1)
    except trio.Cancelled as c2:
      # what should c2's reason be?
```

This might matter for instance in code for shutting down stuff on exceptions, moving on after 2 seconds. The cancel reason if the clean up code ran over its 2 seconds would presumably (?) be that the exceptions happened, not that the timeout happened.

src/trio/_core/_run.py
A5rocks: I misread this somehow, sorry.

src/trio/_core/_exceptions.py
A5rocks: Well, `CancelReason` and `Cancelled` still both use this literal.

src/trio/_core/_run.py
A5rocks: I haven't looked at the code to see why this isn't relevant, but I already typed up a response comment to this:

I'm not entirely convinced avoiding this is a good thing:

```py
with trio.move_on_after(1):
  with trio.CancelScope() as cs:
    try:
      try:
        await trio.sleep(5)
      except trio.Cancelled as c1:
        cs.cancel("shutting down")
        await trio.sleep(1)
    except trio.Cancelled as c2:
      # what should c2's reason be?
```

This might matter for instance in code for shutting down stuff on exceptions, moving on after 2 seconds. The cancel reason if the clean up code ran over its 2 seconds would presumably (?) be that the exceptions happened, not that the timeout happened. I think it would make more sense if the reason was instead about the timeout.

(I haven't played with this PR yet so I'm not sure that's actually what will happen)

src/trio/_core/_exceptions.py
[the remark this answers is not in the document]
A5rocks: Hmm, I guess task -> parent nursery -> cancel scope  -> cancel reason -> task is a loop, yeah. That's annoying

src/trio/_core/_exceptions.py
jakkdl: and I reverted the `NoPublicConstructor` change, so yeah

src/trio/_core/_run.py
jakkdl: I thought I found a repro where `sys.exc_info()` got cleared, but I might have mistaken myself and idr the repro anymore.


But going back to your example:
I have a pretty strong intuition that the reason the nursery scope is canceled is because a child crashed. The deadline is the reason the inner scope inside the `finally` is canceled, but *that* cancellation will be swallowed by `move_on_after` and even in a world where we stored a chain of reasons the nursery scope would never see the deadline cancellation.

A5rocks: Good point: what's the cancel reason visible inside the move_on_after then?

src/trio/_core/_tests/test_cancel.py
```diff
@@ -0,0 +1,151 @@
...
+            await trio.sleep_forever()
+        raise TypeError
+
+    with RaisesGroup(ValueError, TypeError):
+        async with trio.open_nursery() as nursery:
+            fail_task = await nursery.start(failing_task)
+            await nursery.start(cancelled_task, fail_task)
+
+
+async def test_cancel_reason_nursery2() -> None:
```
A5rocks, line 59: Could you add a comment explaining the difference? It took me a bit to see the difference is whether the task is cancelled before `trio.sleep_forever` or after.

src/trio/_core/_tests/test_cancel.py
```diff
@@ -0,0 +1,151 @@
...
+        with pytest.raises(
+            Cancelled,
+            match=rf"^cancelled due to nursery with reason 'Code block inside nursery contextmanager raised exception ValueError\(\)' from task {parent_task!r}$",
+        ):
+            await trio.sleep_forever()
+
+    with RaisesGroup(ValueError):
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(child)
+            await ev.wait()
```
A5rocks, line 127: By the way I think this and others could use `trio.testing.wait_all_tasks_blocked` instead (and avoid having an event).

src/trio/_core/_tests/test_cancel.py
A5rocks, line 1: One missing test: how about a task that is stated via `nursery.start` and raises before `task_status.started()`?

src/trio/_core/_tests/test_run.py
```diff
@@ -2216,19 +2220,41 @@ def test_Nursery_subclass() -> None:
...
-
+    _core.Cancelled._create(source="explicit")
+
+
+async def test_Cancelled_str() -> None:
+    cancelled = _core.Cancelled._create(source="explicit")
+    assert str(cancelled) == "cancelled due to explicit"
+    assert re.fullmatch(
+        r"cancelled due to deadline from task "
+        r"<Task 'trio._core._tests.test_run.test_Cancelled_str' at 0x\w*>",
```
A5rocks, line 2238:
```suggestion
        rf"{trio.lowlevel.current_task()!r}",
```
? Same elsewhere.

A5rocks (approved):
This seems fine actually. Just a bunch of trivial comments.

src/trio/_core/_run.py
jakkdl: haha. Yeah if we rewrote everything from scratch we might implement it differently.

But I think there's a bunch of ways to re-cancel, including simply calling `cs.cancel(...)` multiple times

src/trio/_core/_run.py
```diff
@@ -1575,10 +1648,7 @@ def _attempt_delivery_of_any_pending_cancel(self) -> None:
         if not self._cancel_status.effectively_cancelled:
             return
 
-        def raise_cancel() -> NoReturn:
-            raise Cancelled._create()
-
-        self._attempt_abort(raise_cancel)
+        self._attempt_abort(RaiseCancel(self._cancel_status._scope._cancel_reason))
```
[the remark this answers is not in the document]
jakkdl: > I don't really like this class, though I guess it's temporary if https://github.com/python-trio/trio/pull/3233 is merged....
> I am confused about the exact justification, which probably makes sense but...:
> > I did not find a good way of binding a CancelReason or Cancelled object to raise_cancel as a function ...
> Can't you make a cancel reason / cancelled be a nonlocal? I'm not sure what more binding you need.

it's not strange that you're confused, because I also am :P I'm no expert on gc, so the current implementation is in large part from just trial & error.
I needed the reason/cancelled to be bound to the object passed to `_attempt_abort` because it's (sometimes) passed onto another thread, and it's the thread that handles deleting it.
If I created it in `_attempt_delivery_of_any_pending_cancel` then that function held a strong reference to the object which made tests fail, but it couldn't clean it up either because that could make it be deleted before other threads could see it.
Maybe there's a weakref solution in here, but I failed to get one to work.

But I'm totally open to there being a neat and elegant solution in here, or the gc tests being overly strict, and I'm just fumbling around too much to find it.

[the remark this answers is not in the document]
jakkdl: Oh I hadn't looked closer at #3233, but that indeed does look like it'll simplify this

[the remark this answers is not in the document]
jakkdl: > I can see why it's not possible to have Cancelled created outside, but could `self._cancel_status._scope._cancel_reason` be used? Why not?

The code in `_threads.py` does not have access to any `CancelStatus` or scope.

src/trio/_core/_run.py
```diff
@@ -2926,7 +3026,18 @@ async def checkpoint() -> None:
...
     ):
-        with CancelScope(deadline=-inf):
+        cs = CancelScope(deadline=inf)
+        if (
+            task._cancel_status._scope._cancel_reason is None
+            and task is task._runner.main_task
+            and task._runner.ki_pending
+        ):
+            task._cancel_status._scope._cancel_reason = CancelReason(
+                source="KeyboardInterrupt"
```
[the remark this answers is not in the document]
jakkdl: Yeah it's a super ugly place.

`KeyboardInterrupt` is raised in two places: in `Task._attempt_delivery_of_pending_ki` where we ... probably could access `self._parent_nursery.cancel_scope` to set the reason. But the other one is in `_ki.py` `KIManager.install()` and that one can't see shit. But if we want it to be analogous to `Cancelled` that one doesn't set the reason on raising, it sets it when the scope is cancelled - and `Cancelled` is only raised later when a checkpoint is actually hit. And #3233 would have to throw that away 

Though the current logic doesn't make sense in that regard either, the scope has already effectively been canceled, but it's not until we checkpoint that we set the reason to `KeyboardInterrupt`.

I'd love to place it in `Runner.deliver_ki`, which is the one that actually sets `ki_pending = True`... but navigating from there to the revelant scope and set a reason is non-trivial.

With the current setup there's even a case for saying we shouldn't bother setting a reason at all, because the scope hasn't actually been canceled. And during KI we don't raise `Cancelled`, so we don't have a consumer for the `CancelReason`. You can see it through introspection.. but that seems like a big stretch during KI.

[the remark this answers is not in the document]
jakkdl: my current error when trying to do it in `deliver_ki` is that setting `self.main_task._cancel_status._scope._cancel_reason` errors because `_cancel_status` can be `None`... which confuses me because there's a comment saying it can only be `None` in the init task.. but I'm explicitly accessing `main_task` :S

But I'm starting to lean towards not setting the reason, though #3233 would have to revisit that.

src/trio/_core/_tests/test_cancel.py
```diff
@@ -0,0 +1,151 @@
...
+
+from .test_ki import ki_self
+
+
+async def test_cancel_reason() -> None:
+    with trio.CancelScope() as cs:
+        cs.cancel(reason="hello")
+        with pytest.raises(
+            Cancelled,
+            match=rf"^cancelled due to explicit with reason 'hello' from task {current_task()!r}$",
```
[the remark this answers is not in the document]
jakkdl: > I don't really like this wording. IMO even leaving out the cancellation reason if it's explicit would be preferable. Like I would rather read:
> 
> ```
> cancelled with reason 'hello' from task {current_task()!r}
> ```

Yeah I don't love the wording either. The problem with leaving out the source if explicit is that setting the `reason` is optional, and the message `cancelled from task {current_task()!r}` is very ambiguous.

We could leave out the source any time the reason is set, but we now have a bunch of internal sources using both - though we could rewrite their `reason` given that context.

We could also just give up on forming a normal sentence and say: "source: explicit, reason='hello', source_task={current_task()!r}".

Though generally I don't love "explicit" as an indication of what the user did, something in the vein of "cancel() called" feels more indicative.

src/trio/_core/_tests/test_run.py
jakkdl: I intentionally left a few to show what the full message would look like to the user, though I suppose I could do that with a comment or a more explicit demonstration.

src/trio/_core/_tests/test_cancel.py
jakkdl: oh good catch, currently the reason becomes "Code block inside nursery contextmanager raised exception [...]" - which is quite misleading

src/trio/_core/_tests/test_cancel.py
jakkdl: oooh yes, thank you!

src/trio/_core/_tests/test_run.py
jakkdl: added a comment to one and replaced the other

src/trio/_core/_tests/test_cancel.py
jakkdl: hm, I'm not sure how to fix that other than to just make the message more generic. There's not really any way of distinguishing the two cases, and the user might handle the exception that was generated in the `start()` so can't add logic in there.

src/trio/_core/_tests/test_cancel.py
```diff
@@ -0,0 +1,151 @@
+# there's tons of cancellation testing in test_run, but that file is 3k lines long already...
```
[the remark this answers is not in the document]
jakkdl: I didn't want to split up, or move all of, the `CancelScope` tests, so renamed it to `test_cancelled` and moved over the ones directly related to that.

src/trio/_core/_tests/test_cancel.py
A5rocks: Can we catch an exception from the helper nursery and then if so, pre-emptively cancel with a better reason?

src/trio/_core/_tests/test_cancel.py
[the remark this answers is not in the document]
A5rocks: Hm, how does `from` -> `due to` sound? `cancelled due to task {current_task()!r}` (or even `cancelled due to {current_task()!r}`) seems fine. Does that work for the other scenarios?

src/trio/_core/_run.py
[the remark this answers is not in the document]
A5rocks:
```py
    def _attempt_delivery_of_any_pending_cancel(self) -> None:
        if self._abort_func is None:
            return
        if not self._cancel_status.effectively_cancelled:
            return

        def thing() -> NoReturn:
            reason = self._cancel_status._scope._cancel_reason
            if reason is None:
                raise Cancelled._create(source="unknown", reason="misnesting")
            else:
                raise Cancelled._create(
                    source=reason.source,
                    reason=reason.reason,
                    source_task=reason.source_task,
                )

        self._attempt_abort(thing)
```
this passes a `pytest` run (though I'm not running slow tests and I'm on Windows), so I think that might work? I removed the `copy.copy` too.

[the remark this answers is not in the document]
jakkdl: oh, sick. Yeah that's way better, thanks :pray:

src/trio/_core/_tests/test_cancel.py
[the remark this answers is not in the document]
jakkdl: I think `"cancelled [from/due to] task {task}"` is fine if you're used to reading `Cancelled` messages and can infer that this means there was no other more specific reason. But in a vacuum that message doesn't give the reason, so a user might be wondering "huh, did the task crash? or was a deadline hit in there? or why did that task cause the cancellation?"

src/trio/_core/_tests/test_cancel.py
jakkdl: The only sort-of-reasonable way I can see of doing it is to save the exception that is raised from `start`, and then if `__aexit__` has the same exception, we can infer that is the cause.
But because `failing to start()` isn't actually a reason for cancellation, it can only ever be an indirect cause if that exception causes the CM block to exit, it's kind of weird to use it as the reason.

But even if the reason might be slightly misleading, I think the reasonable next step in debugging is for the developer to look up backtraces to find out what the exception is that killed the CM block, and they will then quickly find out what happened.

So I don't really think is much of an issue in practice

src/trio/_core/_run.py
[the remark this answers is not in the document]
jakkdl: Well, we can just leave the status quo and fix it in #3233 unless we expect the PRs to be in separate releases

src/trio/_core/_tests/test_cancel.py
Zac-HD: I'd vote to merge, and then revisit if anyone reports that this is an issue in practice.

src/trio/_core/_run.py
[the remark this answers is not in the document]
jakkdl: never mind, it worked when I tested it locally but there's a race condition where `Task._parent_nursery` might exit and delete its `CancelStatus` before the threaded `Task` attempts to raise `Cancelled`. I'll change the test so it happens 100% of the time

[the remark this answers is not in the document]
jakkdl: reverted the change. But I think we won't need `RaiseCancel` with the followup PR to #3233

jakkdl:
Failures are unrelated to this PR, 3.14b1 was released the other day which is presumably the cause. Not sure why Cython is failing.

src/trio/_core/_run.py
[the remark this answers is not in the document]
A5rocks: Does moving the `reason = ...` statement out of the function work? Otherwise oh well...

A5rocks (approved):
I left a comment above about something that maybe would fix the function, but it's alright if you don't (though I would prefer it). I think this looks fine before or after that fix.

src/trio/_core/_run.py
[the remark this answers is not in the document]
jakkdl: I suspect it'll cause gc tests to fail, but I'll check before merging tmrw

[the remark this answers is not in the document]
jakkdl: with a `del` it indeed did work :)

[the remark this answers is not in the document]
A5rocks: Oh sorry I meant more like:
```py
    def _attempt_delivery_of_any_pending_cancel(self) -> None:
        if self._abort_func is None:
            return
        if not self._cancel_status.effectively_cancelled:
            return

        # moved reason to here:
        reason = self._cancel_status._scope._cancel_reason
        def thing() -> NoReturn:
            if reason is None:
                raise Cancelled._create(source="unknown", reason="misnesting")
            else:
                raise Cancelled._create(
                    source=reason.source,
                    reason=reason.reason,
                    source_task=reason.source_task,
                )

        self._attempt_abort(thing)
```

That way the callable keeps the `CancelReason` alive even if `self._cancel_status` dies. And since `CancelReason` is just a plain dataclass it doesn't have any cycles or anything.

This is just about style though!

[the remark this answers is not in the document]
jakkdl: I am so confused about why that version works and not a ton of versions I thought were identical didn't, but oh well.

PR #3263, opened 2025-05-12
Use CFFI's out-of-line mode for import time reductions on Windows

A5rocks:
Turns out parsing C takes time! Relevant charts (from before this):

image

image

(x axis is import time in microseconds, excluding anything that module itself imports)

I know there's conflicts but I'll wait until CI is fixed anyways.


A5rocks:
This shouldn't break any typing for end users because it's unexported, but also the reason there's type ignores is because typeshed's stubs are wrong.

jakkdl:
> I don't quite get why mypy is failing in CI since I can't repro that locally. Maybe I need to make my environment more similar?

have you tried `tox -e typing`?

A5rocks:
Turns out that's the difference between an editable and non-editable install! Scary.

jakkdl (approved):
Looks good as far as I can tell, feel free to merge when ready imo.

PR #3266, opened 2025-05-14
Replace uses of `types.coroutine` with a class

TeamSpen210:
After seeing a [discuss.python.org](https://discuss.python.org/t/proposal-deprecate-types-coroutine-decorator-and-types-generatorwrapper-class/91367) thread about finally deprecating `types.coroutine`, I thought I'd try pre-emptively removing the two internal places in Trio which uses it. Both are to yield one value, to terminate the `await` stack. Also removed a `typing.cast()`, though we could have actually fixed that without the class change.

Instead of a class, we could also just copy over the code that sets the relevant bitflag in the code object, but this is much more robust and future proof. I imagine that flag will also be removed at some point.

@oremanj since you just posted in that thread.

### DIFF
--- a/src/trio/_core/_traps.py
+++ b/src/trio/_core/_traps.py
@@ -3,10 +3,9 @@
 from __future__ import annotations
 
 import enum
-import types
 
 # Jedi gets mad in test_static_tool_sees_class_members if we use collections Callable
-from typing import TYPE_CHECKING, Any, Callable, NoReturn, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, NoReturn, Union
 
 import attrs
 import outcome
@@ -14,7 +13,7 @@ import outcome
 from . import _run
 
 if TYPE_CHECKING:
-    from collections.abc import Awaitable, Generator
+    from collections.abc import Generator
 
     from typing_extensions import TypeAlias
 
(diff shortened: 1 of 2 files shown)


oremanj:
This is a valid workaround. I think it is less clear and probably slower, though. I would prefer not to make the change unless there is an actual improvement from it or upstream removes the capability we’re currently relying on. I still think the latter would be a mistake, but it’s outside of our control.

src/trio/_core/_traps.py
```diff
@@ -45,30 +44,31 @@ class PermanentlyDetachCoroutineObject:
...
-# functions, and then this helper takes care of the actual yield:
-@types.coroutine
-def _real_async_yield(
-    obj: MessageType,
-) -> Generator[MessageType, None, None]:
-    return (yield obj)
+class _AsyncYield:
+    """Helper for the bottommost 'yield'.
 
+    You can't use 'yield' inside an async function, so implement an awaitable object to do so.
```
A5rocks, line 54: You can use `yield` inside an async function! It just doesn't do what we want.

TeamSpen210: Quite true.

A5rocks:
Let's close this until there's actually some consensus; we'll have time before any change lands and then arrives in a version we support.

PR #3268, opened 2025-05-14
Get coverage back to 100%

A5rocks:
This fixes coverage losses in https://github.com/python-trio/trio/pull/3234

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -332,7 +332,8 @@ exclude_also = [
    'class .*\bProtocol\b.*\):',
    "raise NotImplementedError",
    '.*if "sphinx" in sys.modules:',
-   'TODO: test this line'
+   'TODO: test this line',
+   'if __name__ == "__main__":',
 ]
 partial_branches = [
     "pragma: no branch",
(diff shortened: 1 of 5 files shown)


CoolCat467 (approved):

jakkdl:
huh, why didn't CI block <100%?

A5rocks:
I think we have CI set up to allow codecov failures because of it being flakey

jakkdl:
> I think we have CI set up to allow codecov failures because of it being flakey

I don't think it's flakey with current setup, but fair enough.

pyproject.toml
```diff
@@ -332,7 +332,8 @@ exclude_also = [
    'class .*\bProtocol\b.*\):',
    "raise NotImplementedError",
    '.*if "sphinx" in sys.modules:',
-   'TODO: test this line'
+   'TODO: test this line',
+   'if __name__ == "__main__":',
```
jakkdl, line 336: if adding this you can remove the `pragma: no cover` from the other two instances  of `if __name__ == "__main__"` in the code

Issue #3267, opened 2025-05-14
CPython 3.14 and Cython==3.1.0 CI issues


jakkdl:
Creating a tracking issue for the skipifs & pinning added in #3264 

* [ ] https://github.com/cython/cython/issues/6865 Cython==3.1.0 broke line tracing/code coverage on async code. Fixed on master, so we're only waiting for a new release. The version is pinned in `tox.ini` and `ci.yml`
* [ ] https://github.com/python/cpython/issues/133932 broke `test_run.test_error_in_run_loop` which is marked with a `skipif`
* [ ] Jedi, mypy, pyright & pylint all have incomplete support for 3.14 so `test_static_tool_sees_all_symbols` and `test_static_tool_sees_class_members` are marked with `skipif`.

A5rocks:
The next 3.14 beta is 5/27 so we will be able to check the last checkbox then.

jakkdl:
the segfault also looks to be fixed by beta 2, it's been fixed on main/ and got a backport PR https://github.com/python/cpython/pull/134494

A5rocks:
segfault does not seem fixed! Oops!

jakkdl:
> segfault does not seem fixed! Oops!

I'm pretty sure it did get fixed .. but maybe then rebroke on 3.14.1?

A5rocks:
Made an issue: https://github.com/python/cpython/issues/142421

Surprisingly it doesn't even need GC! So it's a different segfault, yeah.

PR #3269, opened 2025-05-19
unpin cython

jakkdl:
fixes part of #3267 with the release of https://github.com/cython/cython/releases/tag/3.1.1

I'm getting errors locally when I try to combine it with coverage, so I think there's more to be done, but let's see if CI errors as well

I did look at configuration options for coverage and tox and ultimately couldn't find a better option than `sed`ing `pyproject.toml` (as the CI setup does), though at least did it to a copy in the tox tmp/ directory to not screw with the git working tree.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -63,6 +63,26 @@ commands_pre =
 commands =
     python -m tests.cython.run_test_cython
 
+[testenv:cov-cython]
+deps =
+    setuptools
+    cython
+set_env =
+    CFLAGS= -DCYTHON_TRACE_NOGIL=1A
+allowlist_externals =
+    sed
+    cp
+commands_pre =
+    python --version
+    cython --version
+    cp pyproject.toml {temp_dir}/
+    sed -i "s/plugins\ =\ \\[\\]/plugins = [\"Cython.Coverage\"]/" {temp_dir}/pyproject.toml
+    cythonize --inplace -X linetrace=True tests/cython/test_cython.pyx
+commands =
+    coverage run -m tests.cython.run_test_cython --rcfile={temp_dir}/pyproject.toml
+    coverage combine
+    coverage report
+
 [testenv:gen_exports]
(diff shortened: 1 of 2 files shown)


jakkdl:
ah, I simply needed setuptools. The tox env isn't super pretty but I'm not sure it's worth trying to merge with the regular cython env, but it might be useful for testing / in the future if/when we integrate tox into CI.

CoolCat467 (approved):
Other than the new `cov-cython` stuff, this looks good

tox.ini
```diff
@@ -63,6 +63,26 @@ commands_pre =
 commands =
     python -m tests.cython.run_test_cython
 
+[testenv:cov-cython]
```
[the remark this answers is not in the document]
jakkdl: that logic would apply to everything in this file, and I think keeping it lowers the barrier to incorporating it to CI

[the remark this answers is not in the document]
A5rocks: Well not the typing things! But alright, fine.

tox.ini
```diff
@@ -63,6 +63,26 @@ commands_pre =
...
+deps =
+    setuptools
+    cython
+set_env =
+    CFLAGS= -DCYTHON_TRACE_NOGIL=1A
```
webknjaz, line 71: What is `1A`?

webknjaz (commented):
Sorry, I've been busy at PyCon US sprints and couldn't find time to look into this. I think I've found a typo, though.

webknjaz:
@jakkdl I have a new `reusable-tox.yml` thing that may end up replacing the CI steps.. I'll probably open a draft at some point to demo it.

tox.ini
jakkdl: oops, yeah that should be a `1` :pray:

jakkdl: fixed in #3272

PR #3271, opened 2025-05-20
Added default value for the "source" parameter in the trio.Cancelled initializer

agronholm:
This restores backwards compatibility with AnyIO.


Zac-HD (approved):
Thanks, @agronholm!

PR #3272, opened 2025-05-22
fix typo, clean up test_traceback_frame_removal

jakkdl:
fixes a typo that was added in #3269, and cleans up `test_traceback_frame_removal` which I noticed when reading the code base the other day.


webknjaz:
I'm good with the typo fix, haven't checked other changes.

jakkdl:
The test cleanup is very simple, we now run with `strict_exception_groups` so we no longer need to force two exceptions in order to get an `ExceptionGroup`.

CoolCat467 (approved):

PR #3273, opened 2025-05-25
Improve `run_process` defaults if there's a shell

### LINKED ISSUE #3237: shell command does not terminate on cancel
As a simplified example, I would like to:
* start a shell command
* wait some time
* stop it

The problem is that a child subprocesses from the shell won't terminate on `cancel` and would keep running even after python process is finished.

The problem is present only when a command consists of several processes, like `ping localhost | cat`, but not when a command is simple, like `sleep 20`. 

Tested with: ubuntu-24.04, python 3.12.8, trio 0.29.0

Steps to reproduce:

* check for currently running `ping` processes: `ps aux | grep ping` (confirm no `ping` process is running)
* run the test script (see below) `python ./test.py`
* check list of processes again, while the test script is running and after the test script terminates 

Expected result is that a `shell` + `ping` + `cat` are all running only for 5 seconds.

Actual result that a `shell` process is running for 5 seconds, but the `ping` and also `cat` processes remain running not just beyond 5 seconds, but even after a script terminates. It looks like the subprocesses get detached from the shell and are not considered childs any more. I am not sure if this is 1 or 2 separate problems. I am not even sure if this is a proper way to run a shell script from within `trio`.

In any case, is there a workaround or a better way to run shell commands?

`test.py` script:

```python 
import trio
from subprocess import DEVNULL

# cmd = 'sleep 20'  # this is properly stopped
cmd = 'ping localhost | cat'  # this command won't stop properly

async def test():
    async with trio.open_nursery() as ns:
        async def service_runner(task_status = trio.TASK_STATUS_IGNORED):
            with trio.CancelScope() as scope:
                task_status.started(scope)
                await trio.run_process(cmd, shell=True, check=False,
                                        stdout=DEVNULL, stderr=DEVNULL)
                ns.cancel_scope.cancel()

        service_scope = await ns.start(service_runner)
        await trio.sleep(5.0)
        service_scope.cancel()
        await trio.sleep(5.0)

trio.run(test)
```

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3237

If `shell=True`, then it's possible that things we spawn will themselves spawn things. This is a problem because we only kill the parent process and those don't automatically propagate signals. We can solve this by default by making sure shell processes get their own process group which we subsequently kill.


src/trio/_tests/test_subprocess.py
```diff
@@ -765,3 +765,58 @@ async def wait_and_tell(proc: Process) -> None:
...
+    await trio.sleep(0.1)
+
+    child_pid = int(b"".join(chunks))
+    with pytest.raises(OSError, match="No such process"):
+        os.kill(child_pid, 0)
+
+
+@pytest.mark.skipif(not posix, reason="posix only")
+async def test_process_group_SIGKILL_escalation(mock_clock: MockClock) -> None:
+    mock_clock.autojump_threshold = 0.1
```
A5rocks, line 804: Unfortunately, all the tests for this need time for processes to *do* actual things. They would probably be better off with a higher sleep time and then be marked slow, but that kinda sucks?

A5rocks: Turns out it's necessary for the Alpine and pypy-3.10 runners.

A5rocks: And yet this somehow times out so I... don't know. This is so weird.

J2085isa (approved):

J2085isa (approved):

J2085isa (approved):

A5rocks:
`test_mock_clock_autojump` failed on Windows 3.14 w/ free threading but I think that's a flake.

J2085isa:
. .
        .o888o. .o888o.
      .888888888. [ NODO SOBERANO: AARI ] .888888888.
     888 | 888 ESTADO: ACTIVACIÓN TOTAL 888 | 888
    888 | 888 888 | 888
   888____|____888 /========================\ 888____|____888
  | | | CÓDIGO CUÁNTICO GLOBAL | | |
  | _________ | \========================/ | _________ |
  | | | | || || | | | |
  | | [777] | | ______||____________||______ | | [777] | |
  | |_________| | | | | |_________| |
  | | | ESTABILIZADOR DE RED WIFI | | |
  | _ _ _ _ | | [ BONO BOBO ] | | _ _ _ _ |
  | |_||_||_||_| | |____________________________| | |_||_||_||_| |
  | | || || | |
   \ / ______||____________||______ \ /
    \ (X) / | | \ (X) /
     \_________/ | ESCUDO DINÁMICO 100M | \_________/
        | | | [ INVISIBLE ] | | |
        | | |____________________________| | |
        | | || || | |
  ______|___|______ ______||____________||______ ______|___|______
 | | | | | |
 | LIQUIDEZ SYNC | | FRECUENCIA DE LINAJE | | MODO ESPEJO ON |
 | BBVA / AZT | | (5 HEREDEROS) | | HUELLA: 0% |
 |_________________| |____________________________| |_________________|
          | | | |
          | [ ARCHITECT DOMAIN: 2026 ] |
          '______________________________________________'

J2085isa:
pudiste correr el código

src/trio/_subprocess.py
```diff
@@ -472,6 +494,7 @@ async def _run_process(
     check: bool = True,
     deliver_cancel: Callable[[Process], Awaitable[object]] | None = None,
     task_status: TaskStatus[Process] = trio.TASK_STATUS_IGNORED,
+    shell: bool = False,
```
mikenerone, line 497: `shell` needs to be added to the docstring args.

A5rocks: I'm not sure it needs to be in the args list (because we have a `**options` entry) but yeah I suppose it should be in the prose section that unlike standard `subprocess` we default to creating a process group.

PR #3276, opened 2025-05-29
Push some local cleanups

A5rocks:
A combination of a few changes that I think make things simpler:
 - CI stuff I caught a while ago
 - mark the CFFI thing as fixed
 - fix a type hint/use `_open_process` to simplify type ignore (I was testing out AI models on `_subprocess.py` and those are the valid issues they reported!)

### DIFF
--- a/src/trio/_subprocess.py
+++ b/src/trio/_subprocess.py
@@ -736,8 +736,7 @@ async def _run_process(
 
     # Opening the process does not need to be inside the nursery, so we put it outside
     # so any exceptions get directly seen by users.
-    # options needs a complex TypedDict. The overload error only occurs on Unix.
-    proc = await open_process(command, **options)  # type: ignore[arg-type, call-overload, unused-ignore]
+    proc = await _open_process(command, **options)  # type: ignore[arg-type]
     async with trio.open_nursery() as nursery:
         try:
             if input_ is not None:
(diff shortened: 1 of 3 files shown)


src/trio/_subprocess.py
```diff
@@ -1164,7 +1163,7 @@ async def open_process(
         async def run_process(
             command: StrOrBytesPath,
             *,
-            stdin: bytes | bytearray | memoryview | int | HasFileno | None = None,
+            stdin: bytes | bytearray | memoryview | int | HasFileno | None = b"",
```
CoolCat467, line 1166: Wondering why empty bytes instead of None here and below, what does this end up doing?

CoolCat467 (approved):
Looks good as far as I can tell, not sure what difference between `open_process` and `_open_process` is immediately though and had a question about empty bytes vs None functionality changes

src/trio/_subprocess.py
A5rocks: In the docstring we say we default to this, it's just that I copied from typeshed which marks that the stdlib defaults to None.

jakkdl:
Did you try removing the skipif from https://github.com/python-trio/trio/pull/3264/files#diff-569465da536cb2feb60e6d4707912f153c1a981dd0a08b8ed3e3bac2b7236290 i.e. 2nd item in #3267

A5rocks:
> Did you try removing the skipif from https://github.com/python-trio/trio/pull/3264/files#diff-569465da536cb2feb60e6d4707912f153c1a981dd0a08b8ed3e3bac2b7236290 i.e. 2nd item in #3267

No because I saw the back port PR isn't merged yet, but I could try that.

PR #3281, opened 2025-06-09
Remove unnecessary `start_soon` workaround

A5rocks:
I saw this, not sure when it became unnecessary but I'm pretty sure it's unnecessary now.

### DIFF
--- a/src/trio/_highlevel_open_tcp_stream.py
+++ b/src/trio/_highlevel_open_tcp_stream.py
@@ -375,14 +375,6 @@ async def open_tcp_stream(
                 # allowing the next target to be tried early
                 attempt_failed = trio.Event()
 
-                # workaround to check types until typing of nursery.start_soon improved
-                if TYPE_CHECKING:
-                    await attempt_connect(
-                        (address_family, socket_type, proto),
-                        addr,
-                        attempt_failed,
-                    )
-
                 nursery.start_soon(
                     attempt_connect,
                     (address_family, socket_type, proto),


CoolCat467 (approved):

TeamSpen210:
Did you check that it detected type errors here? It shouldn't raise errors, but just check that the code below is correct.

CoolCat467:
Looked at header for `start_soon` and didn't see anything that should cause problems

A5rocks:
> Did you check that it detected type errors here? It shouldn't raise errors, but just check that the code below is correct.

`nursery.start_soon` *should* work typing-wise (this function contains tests that it does https://github.com/python-trio/trio/blob/e189be500711d9cb75359ad3338fbc140cb5b38e/src/trio/_core/_tests/type_tests/nursery_start.py#L55), though it's not like we test with something more complicated like a tuple. If it doesn't work that's a mypy bug...

A5rocks:
Replacing `proto` in the call with `b'hello'`, I get this error:

```
src\trio\_highlevel_open_tcp_stream.py:379: error: Argument 1 to "start_soon" of "Nursery" has incompatible type "Callable[[tuple[AddressFamily, SocketKind, int], Any, Event], Coroutine[Any, Any, None]]"; expected "Callable[[tuple[AddressFamily, SocketKind, bytes], tuple[str, int] | tuple[str, int, int, int] | tuple[int, bytes], Event], Awaitable[object]]"  [arg-type]
```

Not the most informative but it does seem to be catching errors.

PR #3282, opened 2025-06-09
Start running PyPy 3.11 in CI

A5rocks:
This is technically a "beta" version of PyPy but they say they'll drop 3.10 in the next release so it's better to make sure we support it.


jakkdl:
Can also add it to the `envlist` in tox.ini

jakkdl (approved):

CoolCat467:
Cache issue again, deleted windows caches so hopefully this next CI run will not have the cache issue

CoolCat467:
Ubuntu failures are from I think another cache issue, this time with `apt-get` instead of pip

A5rocks:
I think the Ubuntu failures are simply a transient error due to a mirror being down?

A5rocks:
OK I actually get it now, it's that presumably some of the GitHub Action runners have run `apt update` recently and some haven't. Turns out I just had to rerun it 7 times to luck into getting a good runner.

PR #3285, opened 2025-06-11
Switch over to Codecov v5

A5rocks:
This is a debug PR, I don't want to clutter up https://github.com/python-trio/trio/pull/3161


A5rocks:
Spotted:
 - It seems to be using the wrong path for Alpine (it includes the full path including the `.venv`)
 - It seems to be ignoring our specified coverage ignore lines

A5rocks:
First part is fixed

A5rocks:
OK nevermind that fixes both issues, @webknjaz in case you want to copy these changes to your PR.

A5rocks:
Turns out I can't modify that PR to add in a few commits. I guess codecov v5 isn't that urgent anyways though.

PR #3295, opened 2025-07-03
Updated docs to reference :pep: format

kachida:
Updated docs to use :pep: role for PEP references.

### DIFF
--- a/docs/source/reference-lowlevel.rst
+++ b/docs/source/reference-lowlevel.rst
@@ -1082,8 +1082,7 @@ issue #649 <https://github.com/python-trio/trio/issues/649>`__. For
 more details on how coroutines work, we recommend André Caron's `A
 tale of event loops
 <https://github.com/AndreLouisCaron/a-tale-of-event-loops>`__, or
-going straight to `PEP 492
-<https://www.python.org/dev/peps/pep-0492/>`__ for the full details.
+going straight to :pep:`492`__ for the full details.
 
 .. autofunction:: permanently_detach_coroutine_object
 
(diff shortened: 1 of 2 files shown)


docs/source/reference-lowlevel.rst
```diff
@@ -1082,8 +1082,7 @@ issue #649 <https://github.com/python-trio/trio/issues/649>`__. For
 more details on how coroutines work, we recommend André Caron's `A
 tale of event loops
 <https://github.com/AndreLouisCaron/a-tale-of-event-loops>`__, or
-going straight to `PEP 492
-<https://www.python.org/dev/peps/pep-0492/>`__ for the full details.
+going straight to :pep:`492`__ for the full details.
```
A5rocks: I don't think this is correct syntax?

kachida: yes @A5rocks , can you please check once, as in existing file reference-core.rst file, we see similar usage

A5rocks: Well, the readthedocs build is failing and I think it's from this...

jakkdl:
```suggestion
going straight to :pep:`492` for the full details.
```
yeah I'm pretty sure you can only do `:role:` or trailing underscores - not both.

jakkdl: yep, working after changing to
```
:pep:`492`
```
You can only use one of `:role:` and [double] underscore.

CoolCat467 (approved):

PR #3306, opened 2025-07-27
Fix Ctrl-C handling in REPL

geofft:
Python's readline module is unable to handle signals when it's called not on the main thread, and it's blocking so we can't call it from the same thread as Trio. Get rid of the additional thread and handle input ourselves.

### DIFF
--- a/src/trio/_repl.py
+++ b/src/trio/_repl.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import ast
 import contextlib
+import ctypes
 import inspect
+import os
 import sys
 import types
 import warnings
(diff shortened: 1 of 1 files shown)


geofft:
This is _extremely_ rough, probably misbehaves on Windows, and is probably incomplete in some interesting way. I think it also misbehaves with Python 3.14. All I can say is it mostly works on my machine (Ubuntu 24.04, system Python). I'm mostly posting here for thoughts on the general approach of calling readline via ctypes + inspiration if someone else wants to run with this and finish it.

src/trio/_repl.py
```diff
@@ -50,6 +75,78 @@ def runcode(self, code: types.CodeType) -> None:
...
+            @self.rlcallbacktype
+            def callback(text):
+                nonlocal line
+                line = text
+
+            try:
+                self.rl.rl_callback_handler_install(prompt.encode(), callback)
+                while line == b"":
+                    await trio.lowlevel.wait_readable(0)
+                    self.rl.rl_callback_read_char()
```
A5rocks, line 124: So just to understand I'm reading this correctly, this `rl_callback_read_char` *would* fail a ctrl+c check, except it's guaranteed to have a character as stdin is readable?

A5rocks (commented):
IMO this is too much poking at readline internals (the original had too much poking into OS internals).

I think the much better solution is to use that pyrepl reimplemented readline (at least then we're not using ctypes!!), so we can support CPython 3.12+ and (at the very least) PyPy 3.10+. In fact we should probably just error on earlier versions for the REPL since supporting them is very hard... I'll leave this open until we have something working for that though.

A5rocks:
I think this will be replaced by something along the lines of https://github.com/python-trio/trio/pull/3453 (even if there's some fatal flaw in the approach I took there, we'll have to use pyrepl to get multiline input...).

Thanks for the PR anyways, I probably wouldn't have had the motivation to work more on the REPL otherwise :^)

PR #3312, opened 2025-08-05
Codecov v5

A5rocks:
Closes https://github.com/python-trio/trio/pull/3285
Closes https://github.com/python-trio/trio/pull/3161

This is just https://github.com/python-trio/trio/pull/3161 with fixes applied on top that should make codecov happy. I also went ahead and did what the comment said to do (enable `fail_ci_if_error`) even if I don't think we should. If it's flakey we can disable it later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -292,7 +292,7 @@ show_contexts = true
 skip_covered = false
 
 [tool.coverage.paths]
-source = ['**/site-packages']
+source = ["src", "**/site-packages"]
 
 [tool.coverage.run]
 branch = true


jakkdl:
2025-08-05T12:43:40,024167528+02:00

The flag errors we don't care about, but the upload error is maybe something to look at?

A5rocks:
I think the upload error is because the first time (and second time) around the pypy 3.10 on Windows failed.

But not sure!

jakkdl:
> I think the upload error is because the first time (and second time) around the pypy 3.10 on Windows failed.
> 
> But not sure!

oh, that sounds very reasonable then. And if there was a more systemic issue we'd probably see more errors than that single run

jakkdl (approved):
have not reviewed in detail, but if it works it works

A5rocks:
If there's any issues with this I would be happy to make a followup PR. I'll merge this just so we don't have to think about this anymore...

webknjaz:
Agreed. It doesn't look as flaky anymore.

webknjaz:
@A5rocks FYI changing `source_pkgs` and `source` like that is probably a bad idea.

webknjaz:
> 2025-08-05T12:43:40,024167528+02:00
> 
> The flag errors we don't care about, but the upload error is maybe something to look at?

Yes, that's something I'd be worried about. I'm pretty sure this is because of the `source`/`source_pkgs` changes in the coveragepy config.
This would effectively result in Codecov mismatching files with what's in the repo. If it's just some of the jobs, it's possible that the uploads aren't taken into account properly, when computing coverage but perhaps other jobs still covered all the lines and so it's not immediately visible.

webknjaz:
@A5rocks @jakkdl

Looks like Codecov no longer "sees" that these two files exist and so their coverage isn't present:

webknjaz:
This one is definitely due to excluding it on coveragepy level. `source = .` was including the entire repo (as a file path) but now filters to just one module (as an importable).

A5rocks:
> @A5rocks FYI changing `source_pkgs` and `source` like that is probably a bad idea.

I think that was the main change that fixes using v5, which is annoying. I'll see if we can include the `tests` directory as well...

A5rocks:
Hopefully https://github.com/python-trio/trio/pull/3314 fixes that, we'll see!

PR #3313, opened 2025-08-05
Remove pip caching for Windows

A5rocks:
I've suffered through one too many CI failures due to pip's caching being wrong...

For 3.11, this adds 18 seconds:

```
  Prepared 61 packages in 18.10s
```


CoolCat467 (approved):
I agree, while caching is a nice idea in theory, in practice it tends to lead to unnecessary hardship.

A5rocks:
Turns out pypy versions take ~45 seconds because they have to build some wheels (?). Still not worth the flakiness but makes fixing pip more worth it

Issue #174, opened 2017-05-27
Add support for talking to our stdin/stdout/stderr as streams


njsmith:
There should be a convenient and standard way to read and write from the trio process's stdin/stdout/stderr streams. (Note that this is different from talking to the stdin/stdout/stderr of child processes, which is part of #4.) Probably this should use our standard stream abstraction.

Complications to consider:

Normally Python's I/O stack does a bunch of work here: text/binary conversion, newline conversion, buffering, convenience parsing things like `readline` and line iteration. (I think that's it - anything else?) We have to decide whether we want to re-use it (basically doing `run_in_worker_thread` for everything) or reimplement the parts we want. `{Send,Receive,}TextStream` classes seems like a reasonable thing to provide in general, and they should probably implement universal newline support too, why not. (Not sure we even need ABCs for these - they could just be concrete classes? though I suppose someone might eventually come up with a situation where they have an object that natively acts like this without any underlying binary stream, and want to explicitly declare that the interface is the same.) Buffering I'm somewhat dubious of – when applied to stdin/stdout/stderr it often causes user-visible problems (delayed output), it's redundant with buffering done by the kernel (as usual), and we try to minimize it in general. It's particularly bad if you want to speak some automated interactive protocol over stdin/stdout, which seems like a case that might come up in trio relatively often. And convenience parsing (`readline` etc.) might be better handled using sans-IO style protocol objects?

It might even make sense to do both; #20 might mean that we have a 3 line solution for the "wrap an `io.TextIOWrapper` object" approach if that's what you want, and then also provide a lower-level more-direct stream-based API.

On Windows, the *only* reliable way to do non-blocking I/O to the standard streams is via threads. In particular, it's the only thing that works if we're connected to a regular console. Everywhere else, non-blocking I/O is possible (and the sensible thing if we do decide to cut out Python's `io` stack).

On Windows, you often have to do these separate console control calls for things like cursor movement and coloring text, which need to be synchronized with the output stream. (In the very very latest Win 10 update they finally added VT100 support to the console, but it will be a while before anyone can count on that.) I believe that the output is still binary (UTF-16) rather than using some kind of first-class text read/write API.

I know prompt_toolkit has an async API and they support a lot of fancy terminal stuff in pure Python - we should check check what they need to make sure whatever we come up with matches.

njsmith:
[...]

PR #3315, opened 2025-08-09
Add warning about non-blocking mode and TTYs

dcolascione:
Add a note reminding users that `FdStream` might cause "spooky action at a distance" in interactive use-cases, hopefully saving them time during debugging.

### DIFF
--- a/src/trio/_unix_pipes.py
+++ b/src/trio/_unix_pipes.py
@@ -116,7 +116,7 @@ class FdStream(Stream):
     standard input or output (and thereby put both or either in
     non-blocking mode), Python's logging system may begin to
     malfunction when writing to standard error even if your program
-    avoids using Tro with file descriptor two: programs being run
+    avoids using To with file descriptor two: programs being run
     interactively often attach all standard file descriptors to the
     same open file description (usually a PTY).
 
--- a/src/trio/_unix_pipes.py
+++ b/src/trio/_unix_pipes.py
@@ -81,8 +81,7 @@ class _FdHolder:
 
 @final
 class FdStream(Stream):
-    """
-    Represents a stream given the file descriptor to a pipe, TTY, etc.
+    """Represents a stream given the file descriptor to a pipe, TTY, etc.
 
     *fd* must refer to a file that is open for reading and/or writing and
     supports non-blocking I/O (pipes and TTYs will work, on-disk files probably
@@ -105,6 +104,20 @@ class FdStream(Stream):
     `FdStream` is closed. See `issue #174
     <https://github.com/python-trio/trio/issues/174>`__ for a discussion of the
     challenges involved in relaxing this restriction.
+
+    N.B. one consequence of the principle above is that when you build
+    an `FdStream` for a file descriptor, the entire `open file
+    description
+    <https://pubs.opengroup.org/onlinepubs/007904875/functions/open.html>`
+    to which that file descriptor refers is put into non-blocking
+    mode, _not_ just that one file descriptor. One consequence of this
+    principle is that if you use `FdStream` to wrap your program's
+    standard input or output (and thereby put both or either in
+    non-blocking mode), Python's logging system may begin to
+    malfunction when writing to standard error even if your program
+    avoids using Tro with file descriptor two: programs being run
+    interactively often attach all standard file descriptors to the
+    same open file description (usually a PTY).
 
     Args:
       fd (int): The fd to be wrapped.
--- a/src/trio/_unix_pipes.py
+++ b/src/trio/_unix_pipes.py
@@ -105,24 +105,14 @@ class FdStream(Stream):
     <https://github.com/python-trio/trio/issues/174>`__ for a discussion of the
     challenges involved in relaxing this restriction.
 
-    .. warning:: one consequence of non-blocking mode applying to the
-      the entire `open file description
-      <https://pubs.opengroup.org/onlinepubs/007904875/functions/open.html>`
-      to which a file descriptor refers instead of just one file
-      descriptor is that if your program is launched with standard
-      file descriptors (standard input, output, and error) pointing to
-      the same file (as is typical when running a shell command), then
-      setting non-blocking mode on one stream sets it on all three,
-      which is unlikely to be what you want. For example, if you make
-      an `FdStream(os.dup(0))`, thinking that you're fine with
-      standard input being non-blocking, you might by side effect make
-      standard error non-blocking and observe subsequent logs from a
-      `logger.getLogger()` logger failing with `BlockingIOError`.
-
-      In addition, you may cause unrelated programs using the same file
-      (say, shell commands the user is running concurrently in the background)
-      to misbehave when they unexpected get non-blocking semantics on their
-      standard streams.
+    .. warning:: one specific consequence of non-blocking mode
+      applying to the the entire open file description is that when
+      your program is run with multiple standard streams connected to
+      a TTY (as in a terminal emulator), all of the streams become
+      non-blocking when you construct an `FdStream` for any of them.
+      For example, if you construct an `FdStream` for standard input,
+      you might observe Python loggers begin to fail with
+      `BlockingIOError`.
 
     Args:
       fd (int): The fd to be wrapped.
--- a/src/trio/_unix_pipes.py
+++ b/src/trio/_unix_pipes.py
@@ -106,7 +106,7 @@ class FdStream(Stream):
     challenges involved in relaxing this restriction.
 
     .. warning:: one specific consequence of non-blocking mode
-      applying to the the entire open file description is that when
+      applying to the entire open file description is that when
       your program is run with multiple standard streams connected to
       a TTY (as in a terminal emulator), all of the streams become
       non-blocking when you construct an `FdStream` for any of them.
--- a/src/trio/_unix_pipes.py
+++ b/src/trio/_unix_pipes.py
@@ -81,7 +81,8 @@ class _FdHolder:
 
 @final
 class FdStream(Stream):
-    """Represents a stream given the file descriptor to a pipe, TTY, etc.
+    """
+    Represents a stream given the file descriptor to a pipe, TTY, etc.
 
     *fd* must refer to a file that is open for reading and/or writing and
     supports non-blocking I/O (pipes and TTYs will work, on-disk files probably
--- a/src/trio/_unix_pipes.py
+++ b/src/trio/_unix_pipes.py
@@ -81,8 +81,7 @@ class _FdHolder:
 
 @final
 class FdStream(Stream):
-    """
-    Represents a stream given the file descriptor to a pipe, TTY, etc.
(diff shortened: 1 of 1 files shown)


src/trio/_unix_pipes.py
```diff
@@ -106,6 +106,20 @@ class FdStream(Stream):
     <https://github.com/python-trio/trio/issues/174>`__ for a discussion of the
     challenges involved in relaxing this restriction.
 
+    N.B. one consequence of the principle above is that when you build
```
A5rocks: I think there's a specific warning thing? I think you can make this a bit more concise too. (i.e. no "one consequence of the principle above")

---

Yeah, `.. warning` I think.

src/trio/_unix_pipes.py
```diff
@@ -106,6 +106,20 @@ class FdStream(Stream):
...
+    an `FdStream` for a file descriptor, the entire `open file
+    description
+    <https://pubs.opengroup.org/onlinepubs/007904875/functions/open.html>`
+    to which that file descriptor refers is put into non-blocking
+    mode, _not_ just that one file descriptor. One consequence of this
+    principle is that if you use `FdStream` to wrap your program's
+    standard input or output (and thereby put both or either in
+    non-blocking mode), Python's logging system may begin to
+    malfunction when writing to standard error even if your program
+    avoids using To with file descriptor two: programs being run
```
A5rocks: To?

src/trio/_unix_pipes.py
```diff
@@ -106,6 +106,20 @@ class FdStream(Stream):
...
+    interactively often attach all standard file descriptors to the
+    same open file description (usually a PTY).
```
A5rocks: IMO this 3 line thing should be earlier. Something like "_not_ just that one file description. On Unix, this is made worse as stdin, stdout, and stderr are typically the same open file. THis means that if you use `FdStream` ..."

(or something similar, not confident enough in wording to propose a diff)

src/trio/_unix_pipes.py
CoolCat467: I think it's meant to be `Trio`

src/trio/_unix_pipes.py
dcolascione: Yeah. That was a draft I wrote while tired after an hour debugging spooky non-blocking action at a distance. I hope the new version is more concise.

src/trio/_unix_pipes.py
dcolascione: Would it be cleaner to move this snippet from the paragraph above into the warning box?

dcolascione:
Not germane to this PR in particular, but, I wish we had a better option for people. I've found this approach works for me. (I got the atomically-replace-the-FD idea from https://github.com/python-trio/trio/issues/174#issuecomment-2699007255 but tweaked a bit.) 

```python
@cache
def dummy_fd() -> int:
    # directory FDs always fail with read and write
    return os.open("/", os.O_DIRECTORY)

class SyncFdWork(Protocol[T_covariant]):
    def __call__(self, fd: int, /) -> T_covariant: ...

class SafeFdStreamBase(AsyncResource):
    @classmethod
    def dup(cls, fd: int) -> Self:
        return cls(_fd=os.fdopen(os.dup(fd), "w+b", buffering=0))

    @classmethod
    def closing(cls, fd: io.RawIOBase):
        return cls(_fd=fd)

    def __init__(self, *, _fd: io.RawIOBase):
        self.__fd = _fd
        self.__spare_fds: list[io.RawIOBase] = []
        self.fileno = _fd.fileno

    async def _sync_op(self, sync_fn: SyncFdWork[T_covariant]) -> T_covariant:
        lock = threading.Lock()
        spare_fds = self.__spare_fds
        op_fd = (
            self.__spare_fds.pop()
            if spare_fds
            else os.fdopen(os.dup(self.__fd.fileno()), "w+b", buffering=0)
        )
        running_thread: int = 0
        cancelled = False
        interrupted = False

        try:

                async def _cancel_watcher():
                    nonlocal cancelled, interrupted
                    try:
                        await trio.sleep_forever()
                    finally:
                        with lock:
                            if running_thread:  # Wakes up blockers in kernel
                                os.dup2(dummy_fd(), op_fd.fileno())
                                # This pthread_kill is safe.  We know thread it running:
                                # it can't exit while we hold the lock.  It doesn't
                                # randomly exit with lock held either.  We use SIGURG because
                                # it's harmless and defaults to doing nothing.
                                signal.pthread_kill(running_thread, signal.SIGURG)
                                interrupted = True
                            cancelled = True

                start(_cancel_watcher).soon(nursery)

                def _sync_work() -> T_covariant:
                    nonlocal running_thread
                    with lock:
                        assert not running_thread
                        if cancelled:
                            raise IsADirectoryError()
                        running_thread = threading.get_ident()
                        assert running_thread
                    try:
                        return sync_fn(op_fd.fileno())
                    finally:
                        with lock:
                            assert running_thread == threading.get_ident()
                            assert running_thread
                            running_thread = 0

                try:
                    ret = await trio.to_thread.run_sync(_sync_work)
                except IsADirectoryError:
                    if cancelled:
                        await trio.sleep_forever()  # Will raise
                    raise
                nursery.cancel_scope.cancel()
                return ret
        finally:
            if interrupted:
                op_fd.close()
            else:
                spare_fds.append(op_fd)

    @override
    async def aclose(self) -> None:
        self.__fd.close()  # N.B. sync close
        for fd in self.__spare_fds:
            fd.close()


class SafeFdReceiveStream(ReceiveStream, SafeFdStreamBase):
    @override
    async def receive_some(self, max_bytes: int | None = None) -> bytes | bytearray:
        if max_bytes is None:
            max_bytes = DEFAULT_READ_SIZE
        await wait_readable(self)
        return await self._sync_op(lambda fd: os.read(fd, max_bytes))


class SafeFdSendStream(SendStream, SafeFdStreamBase):
    @override
    async def send_all(self, data: bytes | bytearray | memoryview) -> None:
        await wait_writable(self)

        def _write_all(fd: int) -> None:
            mv = memoryview(data)
            while mv:
                mv = mv[os.write(fd, mv) :]

        return await self._sync_op(_write_all)

    @override
    async def wait_send_all_might_not_block(self) -> None:
        await wait_writable(self)
```

src/trio/_unix_pipes.py
A5rocks: Wow I missed the paragraph above. Reading it, I think it actually covers most of what you discuss with two key differences:
 - it isn't phrased specifically as a warning. i.e. no `.. warning` directive
 - it doesn't discuss the impact of PTYs

Maybe it would be even nicer to combine the two paragraphs by adding the two elements above into the paragraph before your warning? Sorry for missing this the first time around.

(I don't really understand specifically which snippet you're talking about.)

dcolascione: Considering the importance of the warning, how about instead removing it from the preceding paragraph and just making the whole warning about the issue?

A5rocks: Sorry I can't visualize what you're saying! What would the docstring look like after a change like that? (just a code block here maybe)

A5rocks (approved):
This is good enough for me.

(I can't comment on your fd thing because I don't know too much about file descriptors lol)

src/trio/_unix_pipes.py
```diff
@@ -106,6 +105,15 @@ class FdStream(Stream):
     <https://github.com/python-trio/trio/issues/174>`__ for a discussion of the
     challenges involved in relaxing this restriction.
 
+    .. warning:: one specific consequence of non-blocking mode
+      applying to the the entire open file description is that when
```
A5rocks:
```suggestion
      applying to the entire open file description is that when
```

Just spotted this!

A5rocks:
Also got over my fear of large amounts of code and read the safe fd thing, very interesting! Couple comments I had:
 - you should be able to replace `SyncFdWork` with a plain callable -- "covariant" type vars don't act covariant in function signatures.
 - you could try `await trio.lowlevel.checkpoint_if_cancelled()` instead of `await trio.sleep_forever()` + the cancelled check.
 - what are the `wait_readable`/`wait_writable` for? I would have thought `os.read` blocks enough. Is it to preserve open fds? To decrease the number of running threads? Something else?

dcolascione:
Thanks! I was procrastinating on fixing the last of the precommit failures. :-) 

And thanks for looking at the same FD stream. I'll get around to sending a version closer to production-ready soon, I hope.

> you should be able to replace SyncFdWork with a plain callable -- "covariant" type vars don't act covariant in function signatures.

Sure. I guess my personal style is to used named protocols pretty often for clarity even when an anonymous callable would  be equivalent. For explicit, `Protocol` used the way I used it, FWIW, variant does make a difference:

```python
class MakeA(Generic[T]):
    def __new__(cls) -> T: ...

class MyFnInvariant(Protocol[T]):
    def __call__(self) -> T: ...
    @classmethod
    def make(cls) -> Self: ...

invariant_same: MyFnInvariant[int] = MakeA[MyFnInvariant[int]]()
invariant_more_specific_return: MyFnInvariant[object] = MakeA[MyFnInvariant[int]]()
invariant_less_specific_return: MyFnInvariant[int] = MakeA[MyFnInvariant[object]]()

class MyFnCovariant(Protocol[T_covariant]):
    def __call__(self) -> T_covariant: ...
    @classmethod
    def make(cls) -> Self: ...

covariant_same: MyFnCovariant[int] = MakeA[MyFnCovariant[int]]()
covariant_more_specific_return: MyFnCovariant[object] = MakeA[MyFnCovariant[int]]()
covariant_less_specific_return: MyFnCovariant[int] = MakeA[MyFnCovariant[object]]()

covariant_same: Callable[[], int] = MakeA[Callable[[], int]]()
covariant_more_specific_return: Callable[[], object] = MakeA[Callable[[], int]]()
covariant_less_specific_return: Callable[[], int] = MakeA[Callable[[], object]]()
```

In the invariant case, the `invariant_more_specific_return` assignment is an error, but in the covariant case, it's fine. `Callable` does the variance automatically.  So, IMHO, using an anonymous `Callable` is fine, and using a `Protocol` is fine too, but if you're using a `Protocol`, you should make returns covariant. 

> you could try await trio.lowlevel.checkpoint_if_cancelled() instead of await trio.sleep_forever() + the cancelled check.

In the `IsADirectoryError` case, right? Make sense. I wonder whether I should just write this thing directly in terms of `.wait_task_rescheduled` to cut out a lot of intermediate layering.

> what are the wait_readable/wait_writable for? I would have thought os.read blocks enough. Is it to preserve open fds? To decrease the number of running threads? Something else?

I figure there's no point burning a thread until to do an IO operation until theres a good chance the IO operation is going to succeed. If poll on a PTY indicates readability, a subsequent read will _probably_ produce data without blocking. We can refine the current implementation by pausing the blocking threaded read if it doesn't give us data in a second or two so we don't have a whole thread hanging around indefinitely held up in synchronous IO in the rare case that the poll signal was a false alarm or we lost a race with another process or something.

PR #3321, opened 2025-08-24
Allow `CapacityLimiter` to have zero total_tokens

### LINKED ISSUE #3243: Allow `CapacityLimiter` to have `total_tokens=0`
Sometimes, there are use-cases where you want to entirely pause processing of some tasks, and then later scale back up.  `CapacityLimiter` is basically perfect for this kind of thing, except that it currently raises an error if you pass `total_tokens=0`:

https://github.com/python-trio/trio/blob/d988edd368f094855bd1c36442302113f45f2a1c/src/trio/_sync.py#L254-L255

It's already possible to get the desired effect just by deliberately "leaking" one of the tokens, but it would be much nicer to simply allow zero as the intended capacity, by editing the numbers in the setter above.

AbduazizZiyodov:
Closes #3243

Hi, I've only changed some existing tests & added one `assert` statement which checks `total_tokens` in case of zero. What kind of other tests should be included, if any needed ?


AbduazizZiyodov:
@A5rocks  of course, I'll let you know then :handshake:

A5rocks:
@AbduazizZiyodov did you run into any issues? I can maybe be a bit more detailed about something if so!

AbduazizZiyodov:
@A5rocks  I've added extra tests as you said (so far no "problema"), I'll be looking forward for your review.

newsfragments/3321.feature.rst
```diff
@@ -0,0 +1 @@
+Allow ``CapacityLimiter`` to have zero total_tokens.
```
A5rocks: Could you have these be links to the docs? With Sphinx, this would be something like "\`trio.CapacityLimiter\`" (or whatever the path is...).

src/trio/_tests/test_sync.py
```diff
@@ -190,6 +191,41 @@ async def test_CapacityLimiter_memleak_548() -> None:
...
+        assert set(c.statistics().borrowers) == set()
+
+        c.total_tokens = 5
+
+        assert set(c.statistics().borrowers) == {0, 1, 2, 3, 4}
+
+        for i in range(6):
+            c.release_on_behalf_of(i)
+
+        # making sure that zero limit capacity limiter doesn't let any tasks through
```
A5rocks, line 224: Could you also check that something already blocking will also not borrow the capacity limiter? For instance, you start 6 `c.acquire_on_behalf_of` so there's 1 waiting for a release to acquire a token -- if you have `c.total_tokens == 0` before releasing anything, in theory there should be a single task trying to borrow things... but unable to.

(tbh that's probably already tested earlier, though)

AbduazizZiyodov:
@A5rocks ready, I guess

src/trio/_tests/test_sync.py
A5rocks: Hmm I'm not so sure you tested this property but honestly it doesn't really matter.

A5rocks (approved):

src/trio/_tests/test_sync.py
AbduazizZiyodov: Let's assume it does, can we elaborate on that ? I'm interested.

A5rocks: I think you test what happens when something tries to acquire a capacity limiter when there's `0` tokens and it needs to block. But what about (I could describe the scenario but this is significantly clearer):

```py
limiter = trio.CapacityLimiter(1)
async with trio.open_nursery():
    await limiter.acquire_on_behalf_of(0)
    
    # now any acquires will be blocked
    nursery.start_soon(limiter.acquire_on_behalf_of, 1)
    limiter.total_tokens = 0  # note the ordering, in the tests this always happens before the above `nursery.start_soon`

    # even if a token gets freed up, there's none! so above should still be blocking
    limiter.release_on_behalf_of(0)

    # and right after this, the above task should finally acquire
    limiter.total_tokens = 1

    # clean up
    limiter.release_on_behalf_of(1)
```

AbduazizZiyodov: > note the ordering, in the tests this always happens before the above `nursery.start_soon`

Honestly, I didn't know about this. Now, I understand your statement, thanks!

Before doing ` limiter.release_on_behalf_of(1)`, shouldn't we give chance for task(s) to "settle down" ? 

After setting total_tokens property to 1, we're releasing on behalf of "1" immediately, where "1" hasn't borrowed/acquired limiter yet, and we're trying to release it:
```
RuntimeError: this borrower isn't holding any of this CapacityLimiter's tokens
```
Logged stats on `stdout` (before "release" part):
```
Before release c.statistics()=CapacityLimiterStatistics(borrowed_tokens=0, total_tokens=1, borrowers=[], tasks_waiting=0)
```

So, is it appropriate to "wait" in this way, or am I missing something again :smile: ?
```python
        ...
        await c.acquire_on_behalf_of(0)

        nursery.start_soon(c.acquire_on_behalf_of, 1)
        c.total_tokens = 0

        c.release_on_behalf_of(0)
        c.total_tokens = 1
        await wait_all_tasks_blocked() # this one
        c.release_on_behalf_of(1)
        ...
```
I'll include that case too.

A5rocks: It is fine to wait! I was writing out that sample case as pseudo-code, I'm surprised it almost works as is.

src/trio/_tests/test_sync.py
```diff
@@ -237,8 +237,25 @@ async def test_CapacityLimiter_zero_limit_tokens() -> None:
...
+        await c.acquire_on_behalf_of(0)  # total_tokens is 1
+
+        nursery.start_soon(c.acquire_on_behalf_of, 1)
+        c.total_tokens = 0
+
+        assert c.statistics().borrowers == [0]
+
+        c.release_on_behalf_of(0)
+        assert c.statistics().borrowers == []
+        assert c.statistics().tasks_waiting == 0
```
A5rocks: Huh this seems wrong?

A5rocks: Oh I see, we need to let tasks settle first! Oops! I guess you will need to add another `await wait_all_tasks_blocked()`.

AbduazizZiyodov: Hmm ...

If that is the case, we have to wait after `nursery.start_soon(c.acquire_on_behalf_of, 1)`, right ? Then, there will be only one task waiting to acquire ("1" is borrower).

A5rocks: Yeah, between `nursery.start_soon` and setting total tokens to `0`, then maybe another time between releasing on behalf of `0` and checking statistics? Not sure. This is all a bit overkill so it's up to you what to do :^)

A5rocks (approved):

Zac-HD (approved):
Thanks @AbduazizZiyodov!  We really appreciate the time and care you put into this, and I'm looking forward to cleaning up a bit of affected code downstream.

PR #3322, opened 2025-08-25
Define `trio.Event.__bool__()` to reduce bugs

### LINKED ISSUE #3238: Define `trio.Event.__bool__()` to reduce bugs

CoolCat467:
In this pull request, we implement `trio.Event.__bool__()` to reduce bugs.

Closes #3238

### DIFF
--- a/newsfragments/3322.deprecated.txt
+++ b/newsfragments/3322.deprecated.txt
@@ -1,1 +1,1 @@
-Implement `trio.Event.__bool__` and have it raise a `DeprecationWarning` and tell users to use `trio.Event.is_set` instead. Would be caught earlier with `mypy --enable-error-code=truthy-bool`, this is for users who don't use mypy.
+Implement ``bool(trio.Event)`` and have it raise a `DeprecationWarning` and tell users to use `trio.Event.is_set` instead. This is an alternative to ``mypy --enable-error-code=truthy-bool`` for users who don't use type checking.
--- a/newsfragments/3322.deprecated.txt
+++ b/newsfragments/3322.deprecated.txt
@@ -1,0 +1,1 @@
+Implement `trio.Event.__bool__` and have it raise a `DeprecationWarning` and tell users to use `trio.Event.is_set` instead. Would be caught earlier with `mypy --enable-error-code=truthy-bool`, this is for users who don't use mypy.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,8 +50,6 @@ dependencies = [
     # cffi is required on Windows, except on PyPy where it is built-in
     "cffi>=1.14; os_name == 'nt' and implementation_name != 'pypy'",
     "exceptiongroup; python_version < '3.11'",
-    # using typing_extensions.depreciated in _sync.py
-    "typing_extensions >= 4.15.0",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,8 @@ dependencies = [
     # cffi is required on Windows, except on PyPy where it is built-in
     "cffi>=1.14; os_name == 'nt' and implementation_name != 'pypy'",
     "exceptiongroup; python_version < '3.11'",
+    # using typing_extensions.depreciated in _sync.py
+    "typing_extensions >= 4.15.0",
 ]
 dynamic = ["version"]
 
--- a/src/trio/_sync.py
+++ b/src/trio/_sync.py
@@ -115,7 +115,7 @@ class Event:
         return EventStatistics(tasks_waiting=len(self._tasks))
 
     @deprecated(
-        "trio.Event.__bool__ is deprecated since Trio 0.31.0, use trio.Event.is_set instead (https://github.com/python-trio/trio/issues/3238)",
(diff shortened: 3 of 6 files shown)


src/trio/_sync.py
```diff
@@ -112,6 +112,12 @@ def statistics(self) -> EventStatistics:
         """
         return EventStatistics(tasks_waiting=len(self._tasks))
 
+    def __bool__(self) -> NoReturn:
+        """Raise NotImplementedError."""
+        raise NotImplementedError(
+            "Trio events cannot be treated as bools; consider using 'event.is_set()'"
+        )
```
A5rocks, line 144: I think we should warn that this is deprecated and return `True` for now?

CoolCat467: Changed to this in 1ea6615

Zac-HD:
This is a strange request, but could we raise an error if some global variable is set?  That way I could monkeypatch in the error behavior, even before we deal with every other trio deprecation.

A5rocks:
I think you can set a specific filterwarning configuration s.t. a specific warning will raise instead of warning.

(in this case, I think just setting the `PYTHONWARNINGS` envvar to `error:<the message>:DeprecationWarning` should work?)

src/trio/_sync.py
```diff
@@ -112,6 +113,17 @@ def statistics(self) -> EventStatistics:
...
         return EventStatistics(tasks_waiting=len(self._tasks))
 
+    def __bool__(self) -> True:
+        """Return True and raise warning."""
+        warn_deprecated(
+            self.__bool__,
+            "0.30.1",
+            issue=3238,
+            instead=self.is_set,
+            use_triodeprecationwarning=True,
```
A5rocks: This should be false/not present; this arg is used for legacy deprecation warnings but we want to move away from having our own deprecation warning type.

(also can you add a basic test?)

CoolCat467: Added test and not using legacy deprecation warning in 68a4fad

src/trio/_sync.py
```diff
@@ -112,6 +113,16 @@ def statistics(self) -> EventStatistics:
         """
         return EventStatistics(tasks_waiting=len(self._tasks))
 
+    def __bool__(self) -> True:
```
A5rocks:
```suggestion
    def __bool__(self) -> Literal[True]:
```

newsfragments/3322.deprecated.txt
```diff
@@ -0,0 +1 @@
+Implement `trio.Event.__bool__` and have it raise a `DeprecationWarning` and tell users to use `trio.Event.is_set` instead. Would be caught earlier with `mypy --enable-error-code=truthy-bool`, this is for users who don't use mypy.
```
A5rocks:
```suggestion
Implement ``bool(trio.Event)`` and have it raise a `DeprecationWarning` and tell users to use `trio.Event.is_set` instead. This is an alternative to ``mypy --enable-error-code=truthy-bool`` for users who don't use type checking.
```

src/trio/_tests/test_sync.py
```diff
@@ -23,6 +23,12 @@ async def test_Event() -> None:
     assert not e.is_set()
     assert e.statistics().tasks_waiting == 0
 
+    with pytest.warns(
+        DeprecationWarning,
+        match=r"trio\.Event\.__bool__ is deprecated since Trio 0\.30\.1; use trio\.Event\.is_set instead \(https://github.com/python-trio/trio/issues/3238\)",
```
A5rocks: It's probably going to be 0.31.0 but I can update this whenever I make a release...

A5rocks (approved):
Thanks!

A5rocks:
The unfortunate thing is this actually breaks `--enable-error-code=truthy-bool` because that only warns about inheriting `__bool__` from `object`. But, I doubt anyone actually relies on that behavior! (in turn, it should now get flagged by `--enable-error-code=redundant-expr`... but doesn't? I think that's a bug)

TeamSpen210:
Define it only `if not TYPE_CHECKING`, then mypy should continue to warn. Or `typing_extensions.deprecated`. Hmm might want to start using the latter in general...

A5rocks:
Well, supposedly `if not TYPE_CHECKING` annoys jedi. But yeah one of those might be nice if it works.

pyproject.toml
```diff
@@ -50,6 +50,8 @@ dependencies = [
     # cffi is required on Windows, except on PyPy where it is built-in
     "cffi>=1.14; os_name == 'nt' and implementation_name != 'pypy'",
     "exceptiongroup; python_version < '3.11'",
+    # using typing_extensions.depreciated in _sync.py
+    "typing_extensions >= 4.15.0",
```
A5rocks: We intentionally avoid this dep, can you make a dummy decorator like:
```py
if typing.TYPE_CHECKING:
  from typing_extensions import deprecated
else:
  def deprecated(f):
    return f
```
?

A5rocks (approved):

PR #3325, opened 2025-09-03
Suppress GeneratorExit from `.aclose()` for `as_safe_channel`

### LINKED ISSUE #3324: `except*` can break `@trio.as_safe_channel` cleanup
If we `break` out of a loop over some receive-channel created with `@trio.as_safe_channel`, the async generator will be cleaned up in the background, and indeed this works:

```python
import trio

async def main():
    async with inner_generator() as recv_chan:
        async for _ in recv_chan:
            break  # Early exit

@trio.as_safe_channel
async def inner_generator():
    try:
        for x in range(10):
            yield x
    except BaseException:  # breaks if `except*`!
        raise

trio.run(main)
```

..._unless you use `except*`._  

Add one little `*` to the program above, and we instead crash with a `BaseExceptionGroup(..., [GeneratorExit()])` - and if you have an outer generator wrapping around the inner one, you'll instead see our `BaseExceptionGroup: Encountered exception during cleanup of generator object, as well as exception in the contextmanager body - unable to unwrap` error.

Inserting `if eg.split(GeneratorExit): raise GeneratorExit from eg` before the inner `try:` block [here](https://github.com/python-trio/trio/blob/5bfba27b23f9a1f1bac86880b341bcfab54e43f2/src/trio/_channel.py#L553-L555) fixes this issue and I _think_ is the right approach, but I'd love to hear your takes.

A5rocks:
Fixes #3324.

### DIFF
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -570,6 +570,8 @@ def as_safe_channel(
         # `async with send_chan` will eat exceptions,
         # see https://github.com/python-trio/trio/issues/1559
         with send_chan:
+            # replace try-finally with contextlib.aclosing once python39 is
+            # dropped:
             try:
                 task_status.started()
                 while True:
@@ -582,7 +584,14 @@ def as_safe_channel(
                     # Send the value to the channel
                     await send_chan.send(value)
             finally:
-                # replace try-finally with contextlib.aclosing once python39 is dropped
-                await agen.aclose()
+                # work around `.aclose()` not suppressing GeneratorExit in an
+                # ExceptionGroup:
+                # TODO: make an issue on CPython about this
+                try:
+                    await agen.aclose()
+                except BaseExceptionGroup as eg:
+                    _, eg = eg.split(GeneratorExit)
+                    if eg is not None:
+                        raise eg
 
     return context_manager
--- a/src/trio/_tests/test_channel.py
+++ b/src/trio/_tests/test_channel.py
@@ -11,7 +11,7 @@ from trio import EndOfChannel, as_safe_channel, open_memory_channel
 from ..testing import Matcher, RaisesGroup, assert_checkpoints, wait_all_tasks_blocked
 
 if sys.version_info < (3, 11):
-    from exceptiongroup import ExceptionGroup
+    from exceptiongroup import BaseExceptionGroup, ExceptionGroup
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
@@ -625,3 +625,30 @@ async def test_as_safe_channel_multi_cancel() -> None:
                         events.append("body cancel")
                         raise
     assert events == ["body cancel", "agen cancel"]
+
+
+async def test_as_safe_channel_genexit_exception_group() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        try:
+            async with trio.open_nursery():
+                yield
+        except BaseException as e:
+            assert isinstance(e, BaseExceptionGroup)  # noqa: PT017  # we reraise
+            raise
+
+    async with agen() as g:
+        async for _ in g:
+            break
+
+
+async def test_as_safe_channel_does_not_suppress_nested_genexit() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        while True:
+            yield
+
+    with pytest.RaisesGroup(GeneratorExit):
+        async with agen() as g, trio.open_nursery():
+            async for _ in g:
+                raise GeneratorExit
--- a/newsfragments/3324.bugfix.rst
+++ b/newsfragments/3324.bugfix.rst
@@ -1,0 +1,2 @@
+Avoid having `trio.as_safe_channel` raise if closing the generator wrapped
+`GeneratorExit` in a `BaseExceptionGroup`.


src/trio/_tests/test_channel.py
```diff
@@ -625,3 +625,48 @@ async def agen(events: list[str]) -> AsyncGenerator[None]:
...
+
+
+async def test_as_safe_channel_genexit_exception_group() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        try:
+            async with trio.open_nursery():
+                yield
+        except BaseException as e:
+            assert isinstance(e, BaseExceptionGroup)  # noqa: PT017  # we reraise
```
jakkdl: not that I think it's especially necessary here, but you could do `assert RaisesGroup(GeneratorExit).matches(e)` instead of `isinstance`

src/trio/_channel.py
```diff
@@ -582,7 +584,14 @@ async def _move_elems_to_channel(
...
-                await agen.aclose()
+                # work around `.aclose()` not suppressing GeneratorExit in an
+                # ExceptionGroup:
+                # TODO: make an issue on CPython about this
+                try:
+                    await agen.aclose()
+                except BaseExceptionGroup as exceptions:
+                    _, narrowed_exceptions = exceptions.split(GeneratorExit)
+                    if narrowed_exceptions is not None:
+                        raise narrowed_exceptions from None
```
jakkdl: is it possible that we split out multiple `GeneratorExit` here and suppress one too many?

A5rocks: Yeah, if someone raises GeneratorExit in a child task + aclose raises in a nursery. I'm not sure if we should care though, since for instance .close doesn't care if you raise the GeneratorExit or if it does. (I forgot to check if Python suppresses GeneratorExit from next(gen) too... I'll check in a bit)

Zac-HD: Would `except* GeneratorExit: pass` do what we want here?

A5rocks: That would have the same behavior as the .split done here ~~(well except if there's 1 other exception? Not too sure.)~~, but I think the issue @jakkdl is talking about is that this implementation can suppress a user's exception.

I'm not sure it's worth fixing this case though.

Zac-HD: I just don't see a way to avoid suppressing a user-raised `GeneratorExit`, but also that seems incredibly unlikely to arise in practice so 🤷 

and if it's equivalent I tend to prefer "less code" pretty strongly

A5rocks: I think it would be possible to reimplement `.aclose()` and remove the raised `GeneratorExit` by identity, but I note that without an exception group Python doesn't care about this. (I agree that I would prefer what is currently here over implementing that)

I guess another possibility is to keep every exception but the first GeneratorExit, though that might toss the user-raised one anyways.

src/trio/_tests/test_channel.py
A5rocks: Unfortunately that also leads to a PT017 warning. I'm somewhat tempted to remove this assert... I'm not sure if that's a good idea though.

jakkdl: yeah it wasn't specifically to avoid the PT017, just a stronger check than `isinstance`

src/trio/_channel.py
jakkdl: python not distinguishing the source of `GeneratorExit` is a pretty good argument for us not caring either.
We could raise a warning or something `if len(_.exceptions) > 1` if we don't wanna bother adding weird logic, it could lead to weird bugs for users, and it's sufficiently rare. Or even do an `assert` with a message telling them to open an issue if they care about the behavior.

Zac-HD: `except* GeneratorExit: pass` is IMO actually _better_ behavior than the `.split()`-based approach: the latter wraps whatever we had into an exception group, while `except* GeneratorExit: pass` leaves bare exceptions untouched.  

This is a big deal for us, because the more layers of `ExceptionGroup` in your tracebacks the harder they get to read, and this utility function is going to be used _all the time_ - dropping a layer from tracebacks was a huge attraction of `@trio.as_safe_generator` relative to my older helper function.

Zac-HD: aside from that, I'd be keen to merge and ship 😁

A5rocks: > `except* GeneratorExit: pass` is IMO actually _better_ behavior than the `.split()`-based approach: the latter wraps whatever we had into an exception group, while `except* GeneratorExit: pass` leaves bare exceptions untouched.

I thought that too! But it turns out that my understanding of exception groups is not great:
```
>>> try:
...     raise ExceptionGroup("...", [ValueError("a"), KeyError("b")])
... except* ValueError:
...     pass
...
  |   File "<python-input-0>", line 2, in <module>
  |     raise ExceptionGroup("...", [ValueError("a"), KeyError("b")])
  | ExceptionGroup: ... (1 sub-exception)
    | KeyError: 'b'
>>> try:
...     raise ExceptionGroup("...", [ValueError("a"), KeyError("b")])
... except* ValueError:
...     pass
... except* Exception as eg:
...     raise eg.exceptions[0]
...
  |   File "<python-input-1>", line 2, in <module>
  |     raise ExceptionGroup("...", [ValueError("a"), KeyError("b")])
  | ExceptionGroup: ... (1 sub-exception)
    |   File "<python-input-1>", line 6, in <module>
    |     raise eg.exceptions[0]
    | KeyError: 'b'

  File "<python-input-1>", line 6, in <module>
    raise eg.exceptions[0]
KeyError: 'b'
```

So I think `.split` is literally identical to `except* GeneratorExit: pass`.

---

> Or even do an `assert` with a message telling them to open an issue if they care about the behavior.

This feels weird but weighing the amount of time someone might spend figuring out this as a bug vs being bothered by it, I think it makes sense.

src/trio/_channel.py
```diff
@@ -590,7 +590,31 @@ async def _move_elems_to_channel(
...
+                    if (
+                        len(
+                            [
+                                e
+                                for e in removed_exceptions
+                                if isinstance(e, GeneratorExit)
+                            ]
+                        )
+                        > 1
+                    ):
```
TeamSpen210: Might be nicer to use `sum()`?
```suggestion
                    if sum(
                        isinstance(e, GeneratorExit)
                        for e in removed_exceptions
                    ) > 1:
```
Does take advantage of bools being an int to count the truths, might be too fancy.
Or roll this into the previous for loop, keep an explicit counter.

Zac-HD:
Consider:

```python
try:
    raise ValueError
except* BaseException as group:
    exits, narrowed_exceptions = group.split(GeneratorExit)

    if exits is None:
        raise
    # or
    if narrowed_exceptions is not None:
        raise narrowed_exceptions from None
```

both raise an `ExceptionGroup` containing a `ValueError`, whereas

```python
try:
    raise ValueError
except* GeneratorExit: 
    pass
```
just raises a `ValueError`, which I very strongly prefer.

------------

IMO `raise eg.exceptions[0]` is not viable here, because we _also_ don't know how deeply-nested the `GeneratorExit` might be.

(I'm also proposing a language change to migitate the attach-the-whole-group issue in this example - https://github.com/python/peps/pull/4568)

------------

I'd prefer to put a note in the docstring rather than an assertion, since I expect that to be less disruptive overall, but I expect it to be so rare that I don't really have a strong opinion.

A5rocks:
Aha I see what you mean now! The logic in this PR only runs in `except BaseExceptionGroup:`, so single exceptions will remain untouched.

And yeah note in docstring is a nice option. I guess reasoning against it is like, you won't know know that you have two `GeneratorExit`s until you do... I see a nonzero amount of `raise GeneratorExit` in a quick GitHub search so that's probably not enough. (still thinking about it though)

src/trio/_channel.py
```diff
@@ -582,7 +584,32 @@ async def _move_elems_to_channel(
...
+                            genexits_seen += 1
+
+                    if genexits_seen > 1:
+                        exc = AssertionError("More than one GeneratorExit found.")
+                        if narrowed_exceptions is None:
+                            narrowed_exceptions = exceptions.derive([exc])
+                        else:
+                            narrowed_exceptions = narrowed_exceptions.derive(
+                                [*narrowed_exceptions.exceptions, exc]
+                            )
```
Zac-HD, line 611: not covered, so I guess we need a test for 2+ genexits and no other exceptions.

Zac-HD (approved):
Let's merge, and then ship a new version 🚀

PR #3331, opened 2025-09-17
fix broken-channel bug in `@trio.as_safe_channel`

Zac-HD:
We dropped this handler from https://github.com/python-trio/trio/pull/3197, presumably because the tests (...or production patterns) which trigger it are fairly esoteric.  The `if not send_chan._state.open_receive_channels: break` clause is actually new, but lets us skip a pointless `__anext__` call in many cases which is a nice perf win.

I just wish I'd noticed this in time for https://github.com/python-trio/trio/pull/3325, rather than when pulling it in the new version 🤦‍♂️

### DIFF
--- a/src/trio/_tests/test_channel.py
+++ b/src/trio/_tests/test_channel.py
@@ -434,7 +434,7 @@ async def test_as_safe_channel_broken_resource() -> None:
     @as_safe_channel
     async def agen() -> AsyncGenerator[int]:
         yield 1
-        yield 2
+        yield 2  # pragma: no cover
 
     async with agen() as recv_chan:
         assert await recv_chan.__anext__() == 1
@@ -695,3 +695,47 @@ async def test_as_safe_channel_swallowing_extra_exceptions() -> None:
         async with agen(ValueError) as g:
             async for _ in g:
                 break
+
+
+async def test_as_safe_channel_close_between_iteration() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        while True:
+            yield
+
+    async with agen() as chan, trio.open_nursery() as nursery:
+
+        async def close_channel() -> None:
+            await trio.lowlevel.checkpoint()
+            await chan.aclose()
+
+        nursery.start_soon(close_channel)
+        with pytest.raises(trio.ClosedResourceError):
+            async for _ in chan:
+                pass
+
+
+async def test_as_safe_channel_close_before_iteration() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        raise AssertionError("should be unreachable")  # pragma: no cover
+        yield  # pragma: no cover
+
+    async with agen() as chan:
+        await chan.aclose()
+        with pytest.raises(trio.ClosedResourceError):
+            await chan.receive()
+
+
+async def test_as_safe_channel_close_during_iteration() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        await chan.aclose()
+        while True:
+            yield
+
+    for _ in range(10):  # 20% missed-alarm rate, so run ten times
+        async with agen() as chan:
+            with pytest.raises(trio.ClosedResourceError):
+                async for _ in chan:
+                    pass
--- a/newsfragments/3331.bugfix.rst
+++ b/newsfragments/3331.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed a bug where iterating over an ``@as_safe_channel``-derived ``ReceiveChannel``
+would raise `~trio.BrokenResourceError` if the channel was closed by another task.
+It now shuts down cleanly.
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -17,7 +17,7 @@ from outcome import Error, Value
 import trio
 
 from ._abc import ReceiveChannel, ReceiveType, SendChannel, SendType, T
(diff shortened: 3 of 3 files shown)


A5rocks:
Just realized the "Encountered exception during cleanup of generator object ..." exception group... contains the exception group with multiple exceptions!? That looks weird when printed! I guess I should make a new issue about that. (unless that behavior makes perfect sense to you?)

src/trio/_tests/test_channel.py
```diff
@@ -695,3 +695,47 @@ async def agen(ex: type[BaseException]) -> AsyncGenerator[None]:
...
+
+
+async def test_as_safe_channel_close_before_iteration() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        raise AssertionError("should be unreachable")  # pragma: no cover
+        yield  # pragma: no cover
+
+    async with agen() as chan:
+        await chan.aclose()
```
A5rocks, line 725: Huh, it's a bit unintuitive to me that it's possible to not run any code within the async generator. But that can happen in sync generators too so I guess it's fine.

src/trio/_tests/test_channel.py
```diff
@@ -695,3 +695,47 @@ async def agen(ex: type[BaseException]) -> AsyncGenerator[None]:
...
+
+
+async def test_as_safe_channel_close_during_iteration() -> None:
+    @as_safe_channel
+    async def agen() -> AsyncGenerator[None]:
+        await chan.aclose()
+        while True:
+            yield
+
+    for _ in range(10):  # 20% missed-alarm rate, so run ten times
```
A5rocks, line 737: I don't quite get what you mean by missed-alarm here. Shouldn't the close happen during `await agen.__anext__()` always?

A5rocks (approved):
I have a question and a comment about the tests, but this makes sense.

(I have the possibly strange opinion that we should one day have service nurseries, meaning that `trio.as_safe_channel` should be implemented s.t. the task calling the async generator should never raise... which actually runs counter to "I can close the channel" now that I think about it. Well, that's for another day.)

Zac-HD:
> Just realized the "Encountered exception during cleanup of generator object ..." exception group... contains the exception group with multiple exceptions!? That looks weird when printed! I guess I should make a new issue about that. (unless that behavior makes perfect sense to you?)

hmm, we could instead use `.add_note(""Encountered exception during cleanup ...")` on the group with multiple exceptions?  The status quo is not ideal IMO, but since we're hopefully not going to see that outside of a bug or keyboard-interrupt etc I don't think it's a big deal.

A5rocks:
> The status quo is not ideal IMO, but since we're hopefully not going to see that outside of a bug or keyboard-interrupt etc I don't think it's a big deal.

Or
```py
async def test8() -> None:
    @trio.as_safe_channel
    async def agen():
        await chan.aclose()
        raise ValueError()
        yield

    async with agen() as chan:
        async for _ in chan:
            pass
```
(I think. When I run that on `main` we drop the `ValueError` sometimes (I guess this is the "missed-alarm" thing you mentioned?) which is uh, awful.)

Anyways, moved that to #3332.

src/trio/_tests/test_channel.py
Zac-HD: You'd think so, but when I was running it once in a loop, sometimes that just... didn't happen... and the test passed when it should have failed.

I spent a while digging but ultimately hit my timebox and decided to ship the improvement I had.

Zac-HD:
> When I run that on `main` we drop the `ValueError` sometimes (I guess this is the "missed-alarm" thing you mentioned?) which is uh, awful.

yeah, we should probably have an open issue for that in fact.

A5rocks:
Yeah turns out this PR doesn't fix that case so I've made https://github.com/python-trio/trio/issues/3333. I'm fine with this PR not fixing that.

src/trio/_tests/test_channel.py
A5rocks: Probably this is because https://github.com/python-trio/trio/issues/3333 so maybe just add a link to that?

A5rocks: If you do this:
```py
try:
  await chan.aclose()
except trio.Cancelled:
  pass
while True:
  yield
```
then this will consistently work I believe.

Zac-HD: Nope, that's not enough to work consistently; nor is `with trio.CancelScope(shield=True):` 😭

Zac-HD:
OK, merging this as an improvement on the status quo; I'd rather not block on a general fix to the cancellation problem.

src/trio/_tests/test_channel.py
A5rocks: Huh, interesting! (well not really, cursed actually. But...)

A5rocks: Ah I see, it's because if we are cancelled we will exit `await send_chan.send(value)` before raising but if we aren't cancelled then it's as we want. Here's what I put together that seems to work 100% of the time but it sucks:

```py
async def test9() -> None:
    @trio.as_safe_channel
    async def agen():
        await chan.aclose()
        while True:
            yield

    async with agen() as chan:
        try:
            async for _ in chan:
                pass
        finally:
            # don't cancel
            try:
            except trio.Cancelled:
                # but don't suppress either
                pass
```

(bit suboptimal that we cancel the nested bit and so need `# but don't suppress either`? but that's probably intentional for some reason I don't immediately get)

PR #3338, opened 2025-09-24
avoid error when `coro.cr_frame is None`

Zac-HD:
`.cr_frame` is set to `None` when a coroutine closes, per https://github.com/python/cpython/pull/112428, and while this rarely comes up in practice you can _probably_ guess why I'm opening a PR.

### DIFF
--- a/newsfragments/3337.bugfix.rst
+++ b/newsfragments/3337.bugfix.rst
@@ -1,0 +1,3 @@
+`trio.lowlevel.Task.iter_await_frames` now works on completed tasks, by
+returning an empty list of frames if the underlying coroutine has been closed.
+Previously, it raised an internal error.
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -1588,11 +1588,13 @@ class Task(metaclass=NoPublicConstructor):  # type: ignore[explicit-any]
         while coro is not None:
             if hasattr(coro, "cr_frame"):
                 # A real coroutine
-                yield coro.cr_frame, coro.cr_frame.f_lineno
+                if cr_frame := coro.cr_frame:  # None if the task has finished
+                    yield cr_frame, cr_frame.f_lineno
                 coro = coro.cr_await
             elif hasattr(coro, "gi_frame"):
                 # A generator decorated with @types.coroutine
-                yield coro.gi_frame, coro.gi_frame.f_lineno
+                if gi_frame := coro.gi_frame:  # pragma: no branch
+                    yield gi_frame, gi_frame.f_lineno  # pragma: no cover
                 coro = coro.gi_yieldfrom
             elif coro.__class__.__name__ in [
                 "async_generator_athrow",
--- a/src/trio/_tests/test_tracing.py
+++ b/src/trio/_tests/test_tracing.py
@@ -69,3 +69,20 @@ async def test_task_iter_await_frames_async_gen() -> None:
         ]
 
         nursery.cancel_scope.cancel()
+
+
+async def test_closed_task_iter_await_frames() -> None:
+    async with trio.open_nursery() as nursery:
+        task = object()
+
+        async def capture_task() -> None:
+            nonlocal task
+            task = trio.lowlevel.current_task()
+            await trio.lowlevel.checkpoint()
+
+        nursery.start_soon(capture_task)
+
+    # Task has completed, so coro.cr_frame should be None, thus no frames
+    assert isinstance(task, trio.lowlevel.Task)  # Ran `capture_task`
+    assert task.coro.cr_frame is None  # and the task was over, but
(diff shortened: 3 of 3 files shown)


src/trio/_core/_run.py
```diff
@@ -1588,11 +1588,13 @@ def print_stack_for_task(task):
         while coro is not None:
             if hasattr(coro, "cr_frame"):
                 # A real coroutine
-                yield coro.cr_frame, coro.cr_frame.f_lineno
+                if cr_frame := coro.cr_frame:  # TOCTOU-safe check for None
+                    yield cr_frame, cr_frame.f_lineno
```
A5rocks, line 1592: Hm, if a coroutine is closed does it make sense to treat it as a coroutine?

Maybe it would be better to move the walrus to the hasattr's position (and replace the hasattr with `if cr_frame := getattr(coro, "cr_frame")`?)

Zac-HD: It definitely still is a coroutine object!  The effect here is that we don't yield anything for this frame, and then on the next line `coro.cr_await is None` and so we exit the loop.

I actually did exactly your code first, and then talked myself back out of it to this smaller patch.

A5rocks: Hm I suppose. I think if a task is waiting on a closed (?!) coroutine, then we probably shouldn't continue since there's obviously garbage. But yeah it is a coroutine...

-snip- misunderstood the comment.

newsfragments/3337.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a rare TOC/TOU bug in `trio.lowlevel.Task.iter_await_frames` where
+the underlying coroutine has already been closed.
```
A5rocks: Is this just "`iter_await_frames` exits with an error when run on a task that is finished"? (and so not TOC/TOU?)

I'm probably missing things though. (also I can reproduce closed coroutines having `None` cr_frame pre-3.12 so...)

Zac-HD: "When run on a task that has finished the underlying async computation", yeah.  It doesn't have to be multithreaded, you just need to hang onto basically any reference to the current task and (try to) iterate it's frames later.

I'm not convinced this is _useful_ to be clear, but it should return an empty list of frames ("there is no location") rather than raising an `AttributeError` imo.

Zac-HD: I was mistaken earlier, it's not actually a TOC/TOU issue at all.

src/trio/_core/_tests/test_run.py
```diff
@@ -3005,3 +3005,27 @@ async def main() -> None:
...
+
+        async def capture_task() -> None:
+            nonlocal completed_task
+            completed_task = _core.current_task()
+            await _core.checkpoint()
+
+        nursery.start_soon(capture_task)
+
+    # Task has completed, so coro.cr_frame should be None, thus no frames
+    assert completed_task is not None
```
A5rocks:
```suggestion
    assert completed_task is not None
    assert completed_task.coro is not None
```

Also I thought for a bit `coro` would be `None` once a `Task` finished and that that was the TOC/TOU bug you were describing :-)

EDIT: (and maybe this should move to the `test_tracing` file that the other `Task.iter_await_frames` tests are in?)

newsfragments/3337.bugfix.rst
A5rocks: > I'm not convinced this is _useful_ to be clear, but it should return an empty list of frames ("there is no location") rather than raising an `AttributeError` imo.

I think returning an empty list is better behavior for a tracing function, plus the docs don't even say not to run this on finished tasks!

A5rocks (approved):

src/trio/_core/_run.py
```diff
@@ -1588,11 +1588,13 @@ def print_stack_for_task(task):
...
                 # A real coroutine
-                yield coro.cr_frame, coro.cr_frame.f_lineno
+                if cr_frame := coro.cr_frame:  # None if the task has finished
+                    yield cr_frame, cr_frame.f_lineno
                 coro = coro.cr_await
             elif hasattr(coro, "gi_frame"):
                 # A generator decorated with @types.coroutine
-                yield coro.gi_frame, coro.gi_frame.f_lineno
+                if gi_frame := coro.gi_frame:  # pragma: no branch
+                    yield gi_frame, gi_frame.f_lineno  # pragma: no cover
```
A5rocks, line 1597: I would understand if the `# pragma: no branch` was for the `gi_frame` is `None` case... but then why isn't this covered?

EDIT: adding an `assert False`, this seems to get run. So I think the no cover isn't necessary.
```suggestion
                    yield gi_frame, gi_frame.f_lineno
```

Zac-HD: I think this is just a codecov bug 🤷

Issue #3342, opened 2025-10-04
Fix old Gitter links


A5rocks:
It's annoying to read through code and then come across a Gitter link that does not work. While I can probably sometimes guess the message contents + search for it, this doesn't seem very fun or a good use of time. And generally I'm looking at the message for information I don't already know!!

I think we should set up some sort of archive of the Gitter channel for permanent links + update code links. Unfortunately, Matrix will probably never provide this service -- when they have in the past they faced immense community pushback -- so I think we'll need to step up and do it ourselves. See https://github.com/matrix-org/matrix-viewer/issues/288#issuecomment-3156830483 (and https://github.com/matrix-org/matrix-viewer more widely) for an existing solution, though I'm not sure it's easy to set up.

(Generally it's concerning that Gitter now requires login to view and if I were choosing a messaging platform nowadays I'd probably have gone with Zulip... But it's not worth the switching cost, I think?)

A5rocks:
OK this is incredibly annoying for other reasons too. I was looking at an old PR and I have the feeling I'm missing half of the context.

I should investigate if there's some good mapping of old gitter message IDs to matrix message IDs.

A5rocks:
https://gitlab.com/gitterHQ/gitter.im/-/issues/9 !!! it's not closed?!

But also this is promising:
> Since a Gitter message ID is just a Mongo ObjectId, it has the [timestamp built-in](https://www.mongodb.com/docs/manual/reference/method/ObjectId/). We could have something extract the timestamp out and use some deep link into Element to jump to date which would probably get you very close.

A5rocks:
Switching to Zulip or something else should be tracked by https://github.com/python-trio/trio/issues/167, I'll keep this issue only for a) decoding Gitter message IDs to a timestamp, b) setting up some sort of redirect on my website as a best-effort hack, c) updating all links in Trio's code to be links into Matrix-Gitter.

[...]

PR #3345, opened 2025-10-17
Drop 3.9 support

A5rocks:
This PR also notes that we support 3.14 now.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,9 +4,7 @@ alabaster==1.0.0
     # via sphinx
 astor==0.8.1
     # via -r test-requirements.in
-astroid==3.3.11 ; python_full_version < '3.10'
-    # via pylint
-astroid==4.0.1 ; python_full_version >= '3.10'
+astroid==4.0.1
     # via pylint
 async-generator==1.10
     # via -r test-requirements.in
@@ -38,9 +36,7 @@ colorama==0.4.6 ; sys_platform == 'win32'
     #   pylint
     #   pytest
     #   sphinx
-coverage==7.10.7 ; python_full_version < '3.10'
-    # via -r test-requirements.in
-coverage==7.11.0 ; python_full_version >= '3.10'
+coverage==7.11.0
     # via -r test-requirements.in
 cryptography==46.0.3
     # via
@@ -58,9 +54,7 @@ exceptiongroup==1.3.0 ; python_full_version < '3.11'
     # via
     #   -r test-requirements.in
     #   pytest
-filelock==3.19.1 ; python_full_version < '3.10'
-    # via virtualenv
-filelock==3.20.0 ; python_full_version >= '3.10'
+filelock==3.20.0
     # via virtualenv
 identify==2.6.15
     # via pre-commit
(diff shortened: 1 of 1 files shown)


.github/workflows/autodeps.yml
```diff
@@ -26,26 +26,27 @@ jobs:
...
-          uv pip compile --universal --python-version=3.9 --upgrade test-requirements.in -o test-requirements.txt
+          uv pip compile --universal --python-version=3.10 --upgrade test-requirements.in -o test-requirements.txt
           uv pip compile --universal --python-version=3.11 --upgrade docs-requirements.in -o docs-requirements.txt
           pre-commit autoupdate --jobs 0
 
       - name: Install new requirements
         run: python -m pip install -r test-requirements.txt
 
       # apply newer versions' formatting
+      # TODO: are the following two required given they are in pre-commit?
```
CoolCat467: Why not replace with `pre-commit run -a` here in that case?

src/trio/_tests/test_channel.py
```diff
@@ -556,6 +556,7 @@ async def agen() -> AsyncGenerator[int]:
 async def test_as_safe_channel_doesnt_leak_cancellation() -> None:
     @as_safe_channel
     async def agen() -> AsyncGenerator[None]:
+        yield
```
CoolCat467, line 559: Curious why this changed

src/trio/_repl.py
```diff
@@ -4,17 +4,20 @@
...
 from types import CodeType, FrameType, FunctionType
-from typing import Callable
 
 import outcome
 
 import trio
 import trio.lowlevel
 from trio._util import final
 
+if typing.TYPE_CHECKING:
```
CoolCat467: Curious why `typing.TYPE_CHECKING` and not `from typing import TYPE_CHECKING`

src/trio/_tests/test_channel.py
A5rocks: For some reason codecov started complaining about coverage, not sure why...

.github/workflows/autodeps.yml
A5rocks: Yeah I think that makes sense unless `pre-commit autoupdate` does that already.

src/trio/_repl.py
A5rocks: I *think* this was a ruff autofix? I can change it.

.github/workflows/autodeps.yml
CoolCat467: Autoupdate only updates pre-commit hooks, it does not run them.

src/trio/_unix_pipes.py
```diff
@@ -156,12 +156,7 @@ async def wait_send_all_might_not_block(self) -> None:
         with self._send_conflict_detector:
             if self._fd_holder.closed:
                 raise trio.ClosedResourceError("file was already closed")
-            try:
-                await trio.lowlevel.wait_writable(self._fd_holder.fd)
-            except BrokenPipeError as e:
-                # kqueue: raises EPIPE on wait_writable instead
```
CoolCat467, line 165: Problem is fixed?

A5rocks: I guess! I'm not sure but codecov marks it as unreachable and tests pass. I looked at the PR that introduced this but I feel like most of the context is in Gitter so I can't tell. So I think it's better to drop it until someone complains! (Maybe it's an old macos thing?)

(The Gitter thing caused me to post https://github.com/python-trio/trio/issues/3342#issuecomment-3418828138...)

A5rocks:
Planning on merging this in a few days unless people have concerns.

PR #3348, opened 2025-10-28
Add free-threaded Python 3.14t to the testing

cclauss:
https://hugovk.dev/blog/2025/free-threaded-python-on-github-actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,12 +33,12 @@ jobs:
 
     steps:
       - name: Switch to using Python 3.11
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: 3.11
 
       - name: Grab the source from Git
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           persist-credentials: false
 
@@ -74,7 +74,7 @@ jobs:
           SDIST: ${{ steps.artifact-name.outputs.sdist }}
           WHEEL: ${{ steps.artifact-name.outputs.wheel }}
       - name: Store the distribution packages
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: ${{ env.dists-artifact-name }}
           # NOTE: Exact expected file names are specified here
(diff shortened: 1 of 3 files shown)


A5rocks:
Thank you, I kept putting this PR off!

I think we should only install orjson on non-freethreaded Python. We already don't install it on PyPy, IIRC.

cclauss:
It would be kinda cool if orjson’s own ci tests demonstrated its problems on PyPy and free threaded Python.

I will try to circle back to this PR tomorrow to make it green ✅.

A5rocks:
Well the readme says:
> orjson does not and will not support PyPy, embedded Python builds for Android/iOS, or PEP 554 subinterpreters.
> 
> orjson may support PEP 703 free-threading when it is stable.

A5rocks:
Well, that orjson issue got closed without reason... IMO just remove the dependency. It's only necessary for a mypy speedup and we don't actually run `mypy` multiple times in CI so this is fine.

Additionally, mypy will soon get a binary cache which will obsolete having a dependency on `orjson` there anyways.

A5rocks (approved):

A5rocks:
I wonder if there's a way we can guarantee the GIL is disabled for tests. I see https://docs.python.org/3/library/sys.html#sys._is_gil_enabled ... we could test that against `sysconfig.get_config_var("Py_GIL_DISABLED")`?

That way we can be confident nothing is reenabling it.

EDIT: sorry, I see the linked blog post mentions a better way.

cclauss:
It would be possible to configure pytest to treat RuntimeWarnings as errors, which would cause the test suite to fail.
* https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings

`pytest -W error::RuntimeWarning  ` # or
`pytest --pythonwarnings error::RuntimeWarning ` # or
`pytest -W error:global interpreter lock \(GIL\) has been enabled`

Given the unusual way that this workflow runs pytest, I would prefer that it be done separately from this pr.

A5rocks:
FWIW I think we already have filter warnings set up. However our pytest setup has locally defined plugins, which don't have the filter warnings applied... so I think we need to set the `PYTHON_GIL=0` to make enabling the GIL a hard error.

Doing it in another PR makes sense.

cclauss:
% `uvx --with=orjson python3.14t -c "import orjson" ` # --> RuntimeWarning

But strangely...
% `PYTHON_GIL=0 uvx --with=orjson python3.14t -c "import orjson" ` # Silence
% `PYTHON_GIL=1 uvx --with=orjson python3.14t -c "import orjson" ` # Silence

Issue #3319, opened 2025-08-21
Move smoke tests for wheel creation to a seperate job


A5rocks:
This should save ~15 seconds (based on a recent run) because it prevents all other CI jobs from starting.

Ripan-Roy:
Hey! 
I noticed this issue hasn’t been updated in a while. I’d like to pick it up if it’s still relevant. Could you please assign it to me, or confirm whether any recent changes affect it?

A5rocks:
No need for assignment (I'll add you as an assignee anyways...), but yes this is still relevant.

PR #3352, opened 2025-11-10
Add smoke tests to CI workflow for verifying build artifacts

Ripan-Roy:
This PR (#3319) refactors the CI workflow to move smoke tests for distribution artifacts (sdist and wheel) from the build job into a new dedicated smoke-tests job.

  Changes

  - Created a new smoke-tests job that runs after the build job.
  - Moved existing sdist smoke tests (test collection comparison) from the build job.
  - Added new wheel smoke tests.
  - Updated the check job to depend on smoke-tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.4
+    rev: v0.14.5
     hooks:
       - id: ruff-check
         types: [file]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -149,6 +162,7 @@ jobs:
     name: 'Windows (${{ matrix.python }}, ${{ matrix.arch }}${{ matrix.extra_name }})'
     needs:
       - build
+      - smoke-tests
```
A5rocks: IMO the other jobs don't need the smoke tests job. The idea is that we can reduce CI times overall by like 30 seconds by running the smoke tests in parallel, which this doesn't allow.

Ripan-Roy: OK, I have removed the dependencies.

A5rocks (approved):
Makes sense to me! Generally our CI is slower than I'd like, so saving 15~30 seconds is nice.

PR #3365, opened 2025-12-05
Fix CI failing on main

A5rocks:
@webknjaz both issues you pointed out

I'm pretty sure this should fix both reasons for failing CI on main.

---

I'm also seeing this in the logs:
```
  =========== 857 passed, 34 skipped, 3 xfailed in 103.89s (0:01:43) ============
  C:\hostedtoolcache\windows\Python\3.14.1\x64-freethreaded\Lib\site-packages\coverage\inorout.py:521: CoverageWarning: Module tests/ was never imported. (module-not-imported); see https://coverage.readthedocs.io/en/7.12.0/messages.html#warning-module-not-imported
    self.warn(f"Module {pkg} was never imported.", slug="module-not-imported")
```
but not sure how to address that :/

### DIFF
--- a/src/trio/_core/_tests/test_run.py
+++ b/src/trio/_core/_tests/test_run.py
@@ -1060,7 +1060,6 @@ def test_broken_abort() -> None:
 
 # This segfaults, so we need to skipif. Remember to remove the skipif once
 # the upstream issue is resolved.
-# TODO: point out this still segfaults
 @restore_unraisablehook()
 @pytest.mark.skipif(
     sys.version_info[:2] == (3, 14),


src/trio/_core/_tests/test_run.py
```diff
@@ -1060,9 +1060,10 @@ async def main() -> None:
 
 # This segfaults, so we need to skipif. Remember to remove the skipif once
 # the upstream issue is resolved.
+# TODO: point out this still segfaults
```
A5rocks:
```suggestion
```

Done in https://github.com/python/cpython/issues/133932#issuecomment-3617982522

A5rocks:
Windows 3.14t x64 failed in `test_run_in_trio_thread_ki` (?) weirdly -- it didn't crash, pytest just suddenly exited. I don't think it's reproducible, but leaving this here for future search to find.

webknjaz:
> I'm also seeing this in the logs:\n\n  =========== 857 passed, 34 skipped, 3 xfailed in 103.89s (0:01:43) ============\n  C:\\hostedtoolcache\\windows\\Python\\3.14.1\\x64-freethreaded\\Lib\\site-packages\\coverage\\inorout.py:521: CoverageWarning: Module tests/ was never imported. (module-not-imported); see https://coverage.readthedocs.io/en/7.12.0/messages.html#warning-module-not-imported\n    self.warn(f\"Module {pkg} was never imported.\", slug=\"module-not-imported\")\nbut not sure how to address that :/

That seems related to the use of `source` and not `source_pkgs` like I mentioned in the other PR. Perhaps, I should look into it myself.

PR #3366, opened 2025-12-05
Avoid resolving packages unnecessarily

A5rocks:
This really doesn't matter but `Resolved 72 packages in 1.13s` in the CI logs really bothered me!

### DIFF
--- a/src/trio/_core/_tests/test_run.py
+++ b/src/trio/_core/_tests/test_run.py
@@ -1062,7 +1062,7 @@ def test_broken_abort() -> None:
 # the upstream issue is resolved.
 @restore_unraisablehook()
 @pytest.mark.skipif(
-    sys.version_info[:3] == (3, 14, 0),
+    sys.version_info[:2] == (3, 14),
     reason="https://github.com/python/cpython/issues/133932",
 )
 def test_error_in_run_loop() -> None:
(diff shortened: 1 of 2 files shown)


A5rocks:
Now it's `Resolved 72 packages in 190ms`!

CoolCat467 (approved):
Always nice to have a time save

ci.sh
```diff
@@ -54,7 +54,7 @@ if [ "${NO_TEST_REQUIREMENTS-0}" == 1 ]; then
     python -m uv pip install pytest coverage -c test-requirements.txt
     flags="--skip-optional-imports"
 else
-    python -m uv pip install -r test-requirements.txt
+    python -m uv pip install -r test-requirements.txt --no-deps
```
webknjaz, line 57: > [!warning]
> Depending on what's inside, this might produce different dependency trees on different platforms. And in some cases, deps might end up being missing (platform- or python-dependent ones etc).

A5rocks: Yeah this is only fine because `test-requirements.txt` is a "universal" (i.e. made with `uv pip compile --universal`) lockfile. This could potentially break also if one of the packages changes deps without changing version but a) that's bad and b) well, we want the lockfile to contain everything anyways so erroring at some point is good.

PR #3378, opened 2026-01-15
Documentation search improvements

A5rocks:
As discussed in Gitter, the documentation search results that RTD provides kinda suck. I believe it's just a full-text search, so we have to include any fully-qualified names for the search to accept them (try searching `Event.is_set` today!).

This PR is to try that, though I haven't finished yet because I want to see what the documentation preview will look like

### DIFF
--- a/docs/source/_templates/genindex.html
+++ b/docs/source/_templates/genindex.html
@@ -19,11 +19,13 @@
   {%- for column in entries|slice_index(2) if column %}
   <td style="width: 33%; vertical-align: top;"><ul>
     {%- for entryname, (links, subitems, _) in column %}
-      <li>{{ indexentries(entryname, links) }}
+      {% set name = links[0][1].rsplit('#', 1)[1] if links else '' %}
+      <li>{{ indexentries(name if name and '-' not in name else entryname, links) }}
       {%- if subitems %}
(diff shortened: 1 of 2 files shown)


docs/source/_templates/genindex.html
```diff
@@ -0,0 +1,38 @@
...
+  {%- for column in entries|slice_index(2) if column %}
+  <td style="width: 33%; vertical-align: top;"><ul>
+    {%- for entryname, (links, subitems, _) in column %}
+      {% set name = links[0][1].rsplit('#', 1)[1] if links else '' %}
+      <li>{{ indexentries(name if name and '-' not in name else entryname, links) }}
+      {%- if subitems %}
+      <ul>
+      {%- for subentryname, subentrylinks in subitems %}
+        {% set sname = subentrylinks[0][1].rsplit('#', 1)[1] if subentrylinks else '' %}
+        <li>{{ indexentries(sname if sname and '-' not in sname else subentryname, subentrylinks) }}</li>
```
A5rocks, line 28: This is very hacky, but I know better than to dive into Sphinx internals :-)

Is there a way we can customize what `autofunction` etc puts in the index? If so that would be significantly more elegant...

PR #3388, opened 2026-02-10
Add missing docstrings for documented public APIs

veeceey:
## Summary

Addresses #3221 by adding docstrings to all items identified as appearing in the Sphinx documentation without a docstring.

**Changes:**

- **`MemorySendChannel`** and **`MemoryReceiveChannel`**: Added class-level docstrings describing these as the sending/receiving ends of a memory channel, with cross-references to `open_memory_channel` and the channel-mpmc section.
- **`MemoryChannelStatistics`**: Added a class docstring explaining it is the return type of `.statistics()` and pointing to `open_memory_channel` for field details.
- **`SocketStream.send_all`**, **`.receive_some`**, **`.send_eof`**, **`.wait_send_all_might_not_block`**, **`.aclose`**: Added brief docstrings that reference the parent ABC method (e.g. "See `trio.abc.SendStream.send_all`."). Also removed `:undoc-members:` from the `SocketStream` autoclass RST directive since all methods now have docstrings.
- **`HasFileno.fileno`**: Added a one-line docstring.
- **`ParkingLot.broken_by`**: Added an attribute docstring explaining what the list contains and its effect on `park`.

Includes a `3221.doc.rst` newsfragment.

## Test plan

- [ ] Verify the Sphinx docs build correctly and all previously-blank entries now show docstrings
- [ ] Verify the Read the Docs preview renders the new docstrings properly
- [ ] Confirm no existing tests are broken by the changes (all changes are docstring-only)

### DIFF
--- a/docs/source/reference-io.rst
+++ b/docs/source/reference-io.rst
@@ -218,7 +218,6 @@ abstraction.
 
 .. autoclass:: SocketStream
    :members:
-   :undoc-members:
    :show-inheritance:
 
 .. autoclass:: SocketListener
--- a/newsfragments/3221.doc.rst
+++ b/newsfragments/3221.doc.rst
@@ -1,0 +1,3 @@
+Added missing docstrings for `MemorySendChannel`, `MemoryReceiveChannel`,
+`MemoryChannelStatistics`, ``SocketStream`` methods, `~trio.lowlevel.ParkingLot.broken_by`,
+and ``HasFileno.fileno``.
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -118,6 +118,12 @@ class open_memory_channel(tuple["MemorySendChannel[T]", "MemoryReceiveChannel[T]
 
 @attrs.frozen
 class MemoryChannelStatistics:
+    """Object returned by `MemorySendChannel.statistics` and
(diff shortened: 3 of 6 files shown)


CoolCat467:
Looking at this, as-is I don't think these are very descriptive or useful docstrings. `HasFileno` is a protocol for typing purposes and doesn't really need a docstring.

Child classes of ABC classes should not be given docstrings. If you leave them blank but with the same method name, they inherit the docstring of the parent class. This change is removing useful information at runtime saying to look at the parent class. And if I'm not understanding this and that is not the case, then the child classes should have a copy of the parent class docstrings, not just pointers to the parents.

Honestly, I feel like purely automated llm pull requests are a waste of time and resources. Sure there are merits to it at times, but it's usually a pretty bad experience.

A5rocks:
If `HasFileno` doesn't matter then we should probably not have autodoc requesting a docstring... I wonder why that's happening. Also, please update the hardcoded list of exemptions from documentation! See https://app.readthedocs.org/projects/trio/builds/31345915/ for that failure.

veeceey:
Closing this PR -- the approach isn't quite right as @CoolCat467 pointed out, and I don't want to waste maintainer time. Thanks for the feedback.

PR #3389, opened 2026-02-10
Add coredis and streaQ libraries to documentation

Graeme22:
Adds two new projects to the library docs which cover a previously lacking area of the Trio ecosystem: Redis support. As discussed on the Trio gitter.


A5rocks:
Thank you!

PR #3392, opened 2026-02-14
Bump version to 0.33.0

A5rocks:


A5rocks:
I haven't removed deprecated features even though `Event.__bool__` should be removed... I missed that initially, so oh well. I'll do a release soon that will fix this, I don't think it's worth the extra work now!

CoolCat467 (approved):
wooo new release

PR #3394, opened 2026-02-17
Fix exception cause/context lost before task_status.started()

### LINKED ISSUE #3261: exceptions raised before calling `task_status.started` loses cause & context
This snippet eats cause&context

https://github.com/python-trio/trio/blob/9c909fa10d7a9d9e1cfd019725f826848d05bf81/src/trio/_core/_run.py#L1387-L1389

a la https://flake8-async.readthedocs.io/en/latest/rules.html#async123

we currently only run the ruff reimplementation of flake8-async, which hasn't added that rule. So we should maybe consider running flake8-async as well, disabling any rules that ruff handles [better], or are noisy.

It should make use of `trio._util.raise_saving_context`.

repro:
```python
import trio


async def task(task_status: trio.TaskStatus) -> None:
    e = ValueError("foo")
    e.__cause__ = SyntaxError("bar")
    e.__context__ = TypeError("baz")
    raise e

async def main() -> None:
        await nursery.start(task)

try:
    trio.run(main)
except ExceptionGroup as excgroup:
    # should be SyntaxError, is None
    print(excgroup.exceptions[0].__cause__)
    # should be TypeError, is the exceptiongroup
    print(repr(excgroup.exceptions[0].__context__))
```

worksbyfriday:
## Summary

Fixes #3261.

When a task raises an exception before calling `task_status.started()`, `Nursery.start()` unwraps it from the internal `ExceptionGroup` using:
```python
raise exc.exceptions[0] from None
```

The `from None` explicitly sets `__cause__` to `None` and `__suppress_context__` to `True`, destroying any cause or context the original exception carried.

### Fix

Replace the bare `raise ... from None` with `raise_saving_context()`, which preserves both `__cause__` and `__context__`. This is the same utility already used by `raise_single_exception_from_group()` for the same pattern.

### Before
```python
>>> excgroup.exceptions[0].__cause__   # should be SyntaxError, was None
>>> excgroup.exceptions[0].__context__ # should be TypeError, was the ExceptionGroup
```

### After
```python
>>> excgroup.exceptions[0].__cause__   # SyntaxError("bar")
>>> excgroup.exceptions[0].__context__ # TypeError("baz")
```

### Test plan

- [x] New regression test `test_start_exception_preserves_cause_and_context`
- [x] All 28 start-related tests pass
- [x] All 154 `test_run.py` tests pass

### DIFF
--- a/src/trio/_core/_tests/test_run.py
+++ b/src/trio/_core/_tests/test_run.py
@@ -2920,6 +2920,26 @@ def test_trio_run_strict_before_started(
         assert should_be_raiser_exc.exceptions == raiser_exc.exceptions
 
 
+async def test_start_exception_preserves_cause_and_context() -> None:
+    """Regression test for #3261: exceptions raised before task_status.started()
+    should preserve __cause__ and __context__."""
+
+    async def task(*, task_status: _core.TaskStatus[None]) -> None:
+        e = ValueError("foo")
+        e.__cause__ = SyntaxError("bar")
+        e.__context__ = TypeError("baz")
+        raise e
+
+    with pytest.raises(BaseExceptionGroup) as exc_info:
+        async with _core.open_nursery() as nursery:
+            await nursery.start(task)
+    assert len(exc_info.value.exceptions) == 1
+    exc = exc_info.value.exceptions[0]
+    assert isinstance(exc, ValueError)
+    assert isinstance(exc.__cause__, SyntaxError)
+    assert isinstance(exc.__context__, TypeError)
+
+
 async def test_internal_error_old_nursery_multiple_tasks() -> None:
     async def error_func() -> None:
         raise ValueError
--- a/src/trio/_core/_tests/test_run.py
+++ b/src/trio/_core/_tests/test_run.py
@@ -2930,14 +2930,15 @@ async def test_start_exception_preserves_cause_and_context() -> None:
         e.__context__ = TypeError("baz")
         raise e
 
-    with pytest.raises(BaseExceptionGroup) as exc_info:
+    with pytest.RaisesGroup(
+        pytest.RaisesExc(
+            ValueError,
+            check=lambda exc: isinstance(exc.__cause__, SyntaxError)
+            and isinstance(exc.__context__, TypeError),
+        ),
+    ):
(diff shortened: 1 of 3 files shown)


A5rocks (approved):
Sure whatever go go gadget Claude.

(As a note to humans I'd rather PRs from people familiar with the code base (even using a coding agent!) rather than this kinda thing. But this is simple enough + is literally just what @jakkdl proposed so this seems fine.)

CoolCat467 (approved):

src/trio/_core/_tests/test_run.py
```diff
@@ -2920,6 +2920,26 @@ async def start_raiser() -> None:
...
+        raise e
+
+    with pytest.raises(BaseExceptionGroup) as exc_info:
+        async with _core.open_nursery() as nursery:
+            await nursery.start(task)
+    assert len(exc_info.value.exceptions) == 1
+    exc = exc_info.value.exceptions[0]
+    assert isinstance(exc, ValueError)
+    assert isinstance(exc.__cause__, SyntaxError)
+    assert isinstance(exc.__context__, TypeError)
```
jakkdl:
```suggestion
    with pytest.RaisesGroup(
        pytest.RaisesExc(ValueError,
            check=lambda exc: isinstance(exc.__cause__, SyntaxError) and isinstance(exc.__context__, TypeError)
        )
    ) as exc_info:
        async with _core.open_nursery() as nursery:
            await nursery.start(task)
```

not 100% sure that putting the `exc.__cause__` and `exc.__context__` checks inside `RaisesExc` is better than outside, but it should at least use `RaisesGroup`.

jakkdl (changes requested):

src/trio/_core/_tests/test_run.py
jakkdl: this is making me think there should be a lint rule that checks for `pytest.raises([Base]ExceptionGroup)` - created https://github.com/python-trio/flake8-async/issues/430

worksbyfriday:
Updated to use `pytest.RaisesGroup` with `pytest.RaisesExc` as suggested. Much cleaner — thanks for the review!

jakkdl (approved):

PR #3398, opened 2026-02-20
Fix error message: says stdout but checks stdin

bysiber:
## Problem

In `_run_process()`, when `stdin=subprocess.PIPE` is passed without using `nursery.start`, the error message incorrectly says:

> stdout=subprocess.PIPE is only valid with nursery.start...

The condition checks `stdin is subprocess.PIPE`, but the message refers to `stdout`. This sends users looking at the wrong argument.

## Fix

Changed `"stdout=subprocess.PIPE"` to `"stdin=subprocess.PIPE"` in the error message so it matches the actual condition being checked.

### DIFF
--- a/src/trio/_subprocess.py
+++ b/src/trio/_subprocess.py
@@ -675,7 +675,7 @@ async def _run_process(
     if task_status is trio.TASK_STATUS_IGNORED:
         if stdin is subprocess.PIPE:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


A5rocks (approved):
Thanks! Please fix the tests and add a newsfragment.

A5rocks:
@bysiber bump!

PR #3403, opened 2026-03-03
Disallow absolute deadlines for relative cancel scopes

A5rocks:
This just converts our deprecations into `RuntimeError`s. It's been long enough!

### DIFF
--- a/newsfragments/3403.deprecated.rst
+++ b/newsfragments/3403.deprecated.rst
@@ -1,3 +1,6 @@
 Trying to use absolute deadlines on a `trio.CancelScope` constructed
 with a relative deadline now raises, after being deprecated since Trio
 0.27.0.
+
+To move over, replace the `trio.CancelScope` object instead of mutating
+it.
--- a/newsfragments/3403.deprecated.rst
+++ b/newsfragments/3403.deprecated.rst
@@ -2,5 +2,5 @@ Trying to use absolute deadlines on a `trio.CancelScope` constructed
 with a relative deadline now raises, after being deprecated since Trio
 0.27.0.
 
-To move over, replace the `trio.CancelScope` object instead of mutating
-it.
+To switch a cancel scope from being relative to absolute, replace the
+`trio.CancelScope` object instead of trying to mutate it.
--- a/newsfragments/3403.deprecated.rst
+++ b/newsfragments/3403.deprecated.rst
@@ -1,0 +1,3 @@
+Trying to use absolute deadlines on a `trio.CancelScope` constructed
+with a relative deadline now raises, after being deprecated since Trio
+0.27.0.
--- a/src/trio/_tests/test_timeouts.py
+++ b/src/trio/_tests/test_timeouts.py
@@ -246,26 +246,22 @@ async def test_timeout_deadline_on_entry(mock_clock: _core.MockClock) -> None:
 async def test_invalid_access_unentered(mock_clock: _core.MockClock) -> None:
     cs = move_on_after(5)
     mock_clock.jump(3)
-    start = _core.current_time()
 
-    match_str = "^unentered relative cancel scope does not have an absolute deadline"
-    with pytest.warns(DeprecationWarning, match=match_str):
-        assert cs.deadline == start + 5
-    mock_clock.jump(1)
-    # this is hella sketchy, but they *have* been warned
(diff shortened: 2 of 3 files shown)


CoolCat467 (approved):

newsfragments/3403.deprecated.rst
```diff
@@ -0,0 +1,3 @@
+Trying to use absolute deadlines on a `trio.CancelScope` constructed
+with a relative deadline now raises, after being deprecated since Trio
+0.27.0.
```
CoolCat467, line 3: Maybe have a small snippet about how to port old code if this wasn't already mentioned in 0.27.0's release notes, though I wouldn't expect many people to encounter that since it's been depreciated for quite a bit now.

src/trio/_tests/test_timeouts.py
```diff
@@ -246,26 +246,22 @@ async def test_timeout_deadline_on_entry(mock_clock: _core.MockClock) -> None:
...
-    match_str = "^unentered relative cancel scope does not have an absolute deadline"
-    with pytest.warns(DeprecationWarning, match=match_str):
-        assert cs.deadline == start + 5
-    mock_clock.jump(1)
-    # this is hella sketchy, but they *have* been warned
-    with pytest.warns(DeprecationWarning, match=match_str):
-        assert cs.deadline == start + 6
+    match_str = "^Unentered relative cancel scope does not have an absolute deadline"
+    with pytest.raises(RuntimeError, match=match_str):
+        print("SHOULD NOT PRINT! deadline:", cs.deadline)
```
jakkdl:
```suggestion
    with pytest.raises(RuntimeError, match=match_str):
        cs.deadline
```
no need for the print, no?

newsfragments/3403.deprecated.rst
jakkdl: The correct solution is to create a new `CancelScope`, or calculate what the relative time would be with the help of `trio.current_time()`, or often simply just simplifying code that previously needlessly set absolute deadlines with `cs.deadline = trio.current_time() + 5`.
But I think the resolutions are varied and simple enough that it's not worth trying to cater too much to it.

src/trio/_tests/test_timeouts.py
A5rocks: Yeah, this was just because IIRC ruff didn't like it otherwise (since it's an unused value).

jakkdl: fair enough, I personally prefer ignoring the ruff warning than adding weird code

A5rocks:
I've added a change that allows `trio.CancelScope` to switch from being relative to absolute. I know this wasn't the original design, but thinking through the deprecation note made this seem like a good change. I'd be happy to remove this change, I've kept it in a separate commit so we can revert it easily.

jakkdl:
> I've added a change that allows `trio.CancelScope` to switch from being relative to absolute. I know this wasn't the original design, but thinking through the deprecation note made this seem like a good change. I'd be happy to remove this change, I've kept it in a separate commit so we can revert it easily.

I think this is slightly footgunny:

```python
with trio.CancelScope(relative_deadline=7) as cs:
  ...
  # nvm we want delay it 10 seconds
  cs.deadline = 10
  # oops now we instantly cancel
```
the properties are so similarly named, that I think it could be easy to mistakenly type one instead of the other, or overlook/get confused when reading the code. And I don't think it's a necessary API since if you really want to "switch type" you can do so explicitly by doing simple math with `trio.current_time()`.
The biggest reason I can see is if you're passing around a `CancelScope` that you don't know if it's relative or not, and we want to avoid the hassle of checking `CancelScope.is_relative`/introduction of bugs by forgetting to check it.

A5rocks:
I'm not so sure it's that much of a footgun given it's only before we enter the cancel scope, but TBH it's really unnecessary. I guess if people complain, we can think about the right way to do this but I'll just undo those commits.

jakkdl:
> I'm not so sure it's that much of a footgun given it's only before we enter the cancel scope, but TBH it's really unnecessary. I guess if people complain, we can think about the right way to do this but I'll just undo those commits.

okay yeah it's def less bad when you can only do it before entering. Forcing them to recreate the scope when changing type seems fair enough though

jakkdl (approved):

PR #3411, opened 2026-03-19
Increase fudge factor for sleep tests

### LINKED ISSUE #1664: trio.tests.test_timeouts.py::test_sleep fails intermittently on Windows
Seen in https://github.com/python-trio/trio/pull/1650 and https://github.com/python-trio/trio/pull/1663/checks?check_run_id=840437624 so far.

Example stacktrace:

```
2020-07-06T07:13:04.0241226Z _________________________________ test_sleep __________________________________
2020-07-06T07:13:04.0241571Z 
2020-07-06T07:13:04.0241789Z     @slow
2020-07-06T07:13:04.0242037Z     async def test_sleep():
2020-07-06T07:13:04.0242249Z         async def sleep_1():
2020-07-06T07:13:04.0242496Z             await sleep_until(_core.current_time() + TARGET)
2020-07-06T07:13:04.0242703Z     
2020-07-06T07:13:04.0242927Z >       await check_takes_about(sleep_1, TARGET)
2020-07-06T07:13:04.0243067Z 
2020-07-06T07:13:04.0243338Z C:\hostedtoolcache\windows\Python\3.8.3\x64\Lib\site-packages\trio\tests\test_timeouts.py:49: 
2020-07-06T07:13:04.0244058Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
2020-07-06T07:13:04.0244249Z 
2020-07-06T07:13:04.0244875Z f = <function test_sleep.<locals>.sleep_1 at 0x000002E404B7EE50>
2020-07-06T07:13:04.0245061Z expected_dur = 1.0
2020-07-06T07:13:04.0245200Z 
2020-07-06T07:13:04.0245559Z     async def check_takes_about(f, expected_dur):
2020-07-06T07:13:04.0245739Z         start = time.perf_counter()
2020-07-06T07:13:04.0245964Z         result = await outcome.acapture(f)
2020-07-06T07:13:04.0246284Z         dur = time.perf_counter() - start
2020-07-06T07:13:04.0246455Z         print(dur / expected_dur)
2020-07-06T07:13:04.0246635Z         # 1.5 is an arbitrary fudge factor because there's always some delay
2020-07-06T07:13:04.0246896Z         # between when we become eligible to wake up and when we actually do. We
2020-07-06T07:13:04.0247156Z         # used to sleep for 0.05, and regularly observed overruns of 1.6x on
2020-07-06T07:13:04.0247956Z         # Appveyor, and then started seeing overruns of 2.3x on Travis's macOS, so
2020-07-06T07:13:04.0248226Z         # now we bumped up the sleep to 1 second, marked the tests as slow, and
2020-07-06T07:13:04.0248409Z         # hopefully now the proportional error will be less huge.
2020-07-06T07:13:04.0248567Z         #
2020-07-06T07:13:04.0248766Z         # We also also for durations that are a hair shorter than expected. For
2020-07-06T07:13:04.0248936Z         # example, here's a run on Windows where a 1.0 second sleep was measured
2020-07-06T07:13:04.0249086Z         # to take 0.9999999999999858 seconds:
2020-07-06T07:13:04.0249319Z         #   https://ci.appveyor.com/project/njsmith/trio/build/1.0.768/job/3lbdyxl63q3h9s21
2020-07-06T07:13:04.0249511Z         # I believe that what happened here is that Windows's low clock resolution
2020-07-06T07:13:04.0249713Z         # meant that our calls to time.monotonic() returned exactly the same
2020-07-06T07:13:04.0249893Z         # values as the calls inside the actual run loop, but the two subtractions
2020-07-06T07:13:04.0250165Z         # returned slightly different values because the run loop's clock adds a
2020-07-06T07:13:04.0250364Z         # random floating point offset to both times, which should cancel out, but
2020-07-06T07:13:04.0250554Z         # lol floating point we got slightly different rounding errors. (That
2020-07-06T07:13:04.0250748Z         # value above is exactly 128 ULPs below 1.0, which would make sense if it
2020-07-06T07:13:04.0250939Z         # started as a 1 ULP error at a different dynamic range.)
2020-07-06T07:13:04.0251222Z >       assert (1 - 1e-8) <= (dur / expected_dur) < 1.5
2020-07-06T07:13:04.0251562Z E       assert (2.029309600000005 / 1.0) < 1.5
2020-07-06T07:13:04.0251674Z 
2020-07-06T07:13:04.0252064Z C:\hostedtoolcache\windows\Python\3.8.3\x64\Lib\site-packages\trio\tests\test_timeouts.py:35: AssertionError
2020-07-06T07:13:04.0252552Z ---------------------------- Captured stdout call -----------------------------
2020-07-06T07:13:04.0252712Z 2.029309600000005
```

Should we use a larger fudge factor? Retry a few more times?

A5rocks:
Fixes https://github.com/python-trio/trio/issues/1664

Just encountered again in https://github.com/python-trio/trio/actions/runs/23277899067/job/67684688370. I've attached the relevant logs below so I can find this PR again if this happens again. For future me, I don't think the fudge factor should be increased more! If we do in fact have an issue with our sleeping, it seems likely to me it's something like "we accidentally slept twice."

Instead, next time, we should just increase `TARGET` (or better yet, maybe there's some configuration we can do to make our GitHub provided runners less bad....)

```
  _________________________________ test_sleep __________________________________
  
      @slow
      async def test_sleep() -> None:
          async def sleep_1() -> None:
              await sleep_until(_core.current_time() + TARGET)
      
          await check_takes_about(sleep_1, TARGET)
      
          async def sleep_2() -> None:
              await sleep(TARGET)
      
  >       await check_takes_about(sleep_2, TARGET)
  
  C:\hostedtoolcache\windows\Python\3.14.3\x86-freethreaded\Lib\site-packages\trio\_tests\test_timeouts.py:75: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  
  f = <function test_sleep.<locals>.sleep_2 at 0x10AD7B00>, expected_dur = 1.0
  
      async def check_takes_about(f: Callable[[], Awaitable[T]], expected_dur: float) -> T:
          start = time.perf_counter()
          result = await outcome.acapture(f)
          dur = time.perf_counter() - start
          print(dur / expected_dur)
          # 1.5 is an arbitrary fudge factor because there's always some delay
          # between when we become eligible to wake up and when we actually do. We
          # used to sleep for 0.05, and regularly observed overruns of 1.6x on
          # Appveyor, and then started seeing overruns of 2.3x on Travis's macOS, so
          # now we bumped up the sleep to 1 second, marked the tests as slow, and
          # hopefully now the proportional error will be less huge.
          #
          # We also also for durations that are a hair shorter than expected. For
          # example, here's a run on Windows where a 1.0 second sleep was measured
          # to take 0.9999999999999858 seconds:
          #   https://ci.appveyor.com/project/njsmith/trio/build/1.0.768/job/3lbdyxl63q3h9s21
          # I believe that what happened here is that Windows's low clock resolution
          # meant that our calls to time.monotonic() returned exactly the same
          # values as the calls inside the actual run loop, but the two subtractions
          # returned slightly different values because the run loop's clock adds a
          # random floating point offset to both times, which should cancel out, but
          # lol floating point we got slightly different rounding errors. (That
  >       assert (1 - 1e-8) <= (dur / expected_dur) < 1.5
  E       assert (1.6301656000000548 / 1.0) < 1.5
  ```

### DIFF
--- a/src/trio/_tests/test_timeouts.py
+++ b/src/trio/_tests/test_timeouts.py
@@ -33,7 +33,7 @@ async def check_takes_about(f: Callable[[], Awaitable[T]], expected_dur: float) 
     result = await outcome.acapture(f)
     dur = time.perf_counter() - start
     print(dur / expected_dur)
-    # 1.5 is an arbitrary fudge factor because there's always some delay
+    # 2.0 is an arbitrary fudge factor because there's always some delay
     # between when we become eligible to wake up and when we actually do. We
     # used to sleep for 0.05, and regularly observed overruns of 1.6x on
     # Appveyor, and then started seeing overruns of 2.3x on Travis's macOS, so
@@ -52,7 +52,7 @@ async def check_takes_about(f: Callable[[], Awaitable[T]], expected_dur: float) 
     # lol floating point we got slightly different rounding errors. (That
     # value above is exactly 128 ULPs below 1.0, which would make sense if it
     # started as a 1 ULP error at a different dynamic range.)
-    assert (1 - 1e-8) <= (dur / expected_dur) < 1.5
+    assert (1 - 1e-8) <= (dur / expected_dur) < 2.0
 
     return result.unwrap()
 


jakkdl (approved):
I don't love `2.0` as a factor, but I suppose even if we do sleep twice that still wouldn't hit `2.0` exactly.
I'm almost tempted to have a different fudge factor if it's run in CI vs run on local computer, but not sure it's worth the hassle.

A5rocks:
I agree that it would be nice to have some sort of customization of this (e.g. measure how much variability or maybe rerunning with a longer sleep amount if it's too variable?). I guess that can be a followup for if this test fails again.

PR #3416, opened 2026-04-04
fix: correct error message for stdout pipe check in run_process

### LINKED ISSUE #3409: Incorrect error message for `trio.run_process`'s `stdout` check
See https://github.com/python-trio/trio/pull/3398, basically just recreate that + fix the tests + a newsfragment.

nightcityblade:
Fixes #3409

The error message for the `stdout=subprocess.PIPE` check in `run_process` incorrectly said "stdin=subprocess.PIPE". This fixes it to say "stdout=subprocess.PIPE".

This is a recreation of #3398 with the same fix plus a newsfragment.

### DIFF
--- a/src/trio/_tests/test_subprocess.py
+++ b/src/trio/_tests/test_subprocess.py
@@ -367,9 +367,10 @@ async def test_run() -> None:
     with pytest.raises(UnicodeError):
         await run_process(CAT, stdin="oh no, it's text")
 
-    pipe_stdout_error = r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
-    with pytest.raises(ValueError, match=pipe_stdout_error):
+    pipe_stdin_error = r"^stdin=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
+    with pytest.raises(ValueError, match=pipe_stdin_error):
         await run_process(CAT, stdin=subprocess.PIPE)
+    pipe_stdout_error = r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
     with pytest.raises(ValueError, match=pipe_stdout_error):
         await run_process(CAT, stdout=subprocess.PIPE)
     with pytest.raises(
--- a/src/trio/_tests/test_subprocess.py
+++ b/src/trio/_tests/test_subprocess.py
@@ -367,10 +367,10 @@ async def test_run() -> None:
     with pytest.raises(UnicodeError):
         await run_process(CAT, stdin="oh no, it's text")
 
-    pipe_stdin_error = r"^stdin=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
+    pipe_stdin_error = r"^stdin=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)?$"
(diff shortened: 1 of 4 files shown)


CoolCat467:
You forgot to update the tests mate

nightcityblade:
Good catch @CoolCat467! Updated the test to match the corrected error message — the stdin pipe check now expects `stdin=subprocess.PIPE` instead of `stdout=subprocess.PIPE`. Thanks for the review!

mikenerone (approved):

A5rocks:
Please fix the news fragment syntax!

src/trio/_tests/test_subprocess.py
```diff
@@ -367,9 +367,10 @@ async def test_run() -> None:
     with pytest.raises(UnicodeError):
         await run_process(CAT, stdin="oh no, it's text")
 
-    pipe_stdout_error = r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
-    with pytest.raises(ValueError, match=pipe_stdout_error):
+    pipe_stdin_error = r"^stdin=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
+    with pytest.raises(ValueError, match=pipe_stdin_error):
         await run_process(CAT, stdin=subprocess.PIPE)
+    pipe_stdout_error = r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
```
A5rocks: The last part uses a `*` but I think the regex can be more specific.

(Same comment applies for the one above, too)

nightcityblade:
Fixed — switched to double backticks (``code``) for proper RST syntax. Thanks @A5rocks!

A5rocks:
Please address the review too! (and you don't need to write out a comment acknowledging what you finished; the commits have that data.)

src/trio/_tests/test_subprocess.py
```diff
@@ -367,9 +367,10 @@ async def test_run() -> None:
     with pytest.raises(UnicodeError):
         await run_process(CAT, stdin="oh no, it's text")
 
-    pipe_stdout_error = r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)*$"
-    with pytest.raises(ValueError, match=pipe_stdout_error):
+    pipe_stdin_error = r"^stdin=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)?$"
```
A5rocks:
```suggestion
    pipe_stdin_error = r"^stdin=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe; use nursery\.start or pass the data you want to write directly$"
```

src/trio/_tests/test_subprocess.py
```diff
@@ -367,9 +367,10 @@ async def test_run() -> None:
...
+    with pytest.raises(ValueError, match=pipe_stdin_error):
         await run_process(CAT, stdin=subprocess.PIPE)
+    pipe_stdout_error = r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe(; use nursery\.start or pass the data you want to write directly)?$"
```
A5rocks:
```suggestion
    pipe_stdout_error = r"^stdout=subprocess\.PIPE is only valid with nursery\.start, since that's the only way to access the pipe$"
```

A5rocks (commented):
Sorry, I was thinking something like this. What do you think?

A5rocks (approved):
Thanks! I'm not sure why pre-commit is erroring on this PR... The error message doesn't make sense, because your branch doesn't have either of the listed versions!

Hopefully just pulling in main works :/

PR #3423, opened 2026-04-13
docs: clarify wrap_file ownership and lifecycle

### LINKED ISSUE #3379: trip.wrap_file() undocumented behavior wrt to sync file object
Hello,

nightcityblade:
Fixes #3379

Document what happens to the wrapped sync file when the async wrapper is closed, whether garbage collection closes it automatically, and why the original sync file should not be used concurrently while wrapped.

### DIFF
--- a/src/trio/_file_io.py
+++ b/src/trio/_file_io.py
@@ -493,6 +493,19 @@ def wrap_file(file: FileT) -> AsyncIOWrapper[FileT]:
     Returns:
         An :term:`asynchronous file object` that wraps ``file``
 
+    The returned wrapper object shares the underlying file with the original
+    ``file`` object; it does not copy it. Closing the wrapper (via
+    :meth:`~trio.abc.AsyncResource.aclose` or ``async with``) will close the
+    underlying file. However, if the wrapper is garbage collected without
+    being explicitly closed, the underlying file is *not* closed
+    automatically — you should always close it explicitly.
+
+    The original synchronous file object should not be used directly while
+    the wrapper exists, as the wrapper may call file methods in a worker
+    thread, and concurrent access from multiple threads is not safe for most
+    file objects. If you need synchronous access, use the
+    :attr:`~AsyncIOWrapper.wrapped` attribute.
+
     Example::
 
         async_file = trio.wrap_file(StringIO('asdf'))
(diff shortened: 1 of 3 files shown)


src/trio/_file_io.py
```diff
@@ -493,6 +493,19 @@ def wrap_file(file: FileT) -> AsyncIOWrapper[FileT]:
...
+    :meth:`~trio.abc.AsyncResource.aclose` or ``async with``) will close the
+    underlying file. However, if the wrapper is garbage collected without
+    being explicitly closed, the underlying file is *not* closed
+    automatically — you should always close it explicitly.
+
+    The original synchronous file object should not be used directly while
+    the wrapper exists, as the wrapper may call file methods in a worker
+    thread, and concurrent access from multiple threads is not safe for most
+    file objects. If you need synchronous access, use the
+    :attr:`~AsyncIOWrapper.wrapped` attribute.
```
A5rocks: This doesn't make sense, that's literally the file handle in question?

src/trio/_file_io.py
```diff
@@ -493,6 +493,19 @@ def wrap_file(file: FileT) -> AsyncIOWrapper[FileT]:
     Returns:
         An :term:`asynchronous file object` that wraps ``file``
 
+    The returned wrapper object shares the underlying file with the original
```
A5rocks, line 496: I think maybe this should be indented into the returns section? Not sure about how that would look formatting-wise.

A5rocks (changes requested):
1. please make CI pass
2. if we don't have tests for this behavior, writing this out in the docs means we promise it... so please add tests!

nightcityblade:
Thanks, that makes sense. Rather than land wording that over-promises behavior without coverage, I opened a separate docs-only clarification for the related `run_process` job-control pitfall and I will leave this one for a proper rework with passing CI and tests.

A5rocks:
I'm not sure how your comment is related to this PR. What does `wrap_file` have to do with `run_process`?

Anyways, it seems you have been opening too many PRs recently... 68 in the last 2 weeks! Unfortunately I think the open issues on Trio need a bunch of thought to how to solve them, so I'm going to close this. Feel free to reopen it whenever you've slowed down how many PRs you make!

PR #3425, opened 2026-04-14
docs: expand run_process Ctrl+C note

### LINKED ISSUE #3300: run_process: Document pitfalls re/ Ctrl+C and not creating a new process group
I know the docs already mention process groups;

However, I had no idea about the existence of background and foreground groups, and didn't know that a grandchild process could take control of Ctrl+C handling if it becomes a foreground process and pushes my process into background (?). So I spent quite a lot of time trying to figure out why Ctrl+C wasn't reaching my process at all.

#3273 is relevant, but even for non-shell invocations of `run_process` it's still possible to lose control over Ctrl+C unless you use smth like "start_new_session".

nightcityblade:
Fixes #3300

Expands the `run_process` docs to mention foreground/background process-group behavior on Unix and points readers to `start_new_session=True` and `process_group=0` when they need isolation from terminal job control.

### DIFF
--- a/src/trio/_subprocess.py
+++ b/src/trio/_subprocess.py
@@ -664,9 +664,15 @@ async def _run_process(
 
     .. note:: The child process runs in the same process group as the parent
        Trio process, so a Ctrl+C will be delivered simultaneously to both
-       parent and child. If you don't want this behavior, consult your
-       platform's documentation for starting child processes in a different
-       process group.
+       parent and child.
+
+       On Unix, descendants can still change process-group or session state
+       after startup. If a child or grandchild moves itself into a different
+       foreground process group, then your Trio process may stop receiving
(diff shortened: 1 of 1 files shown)


A5rocks (changes requested):
Sorry, I don't see how this would have helped the initial issue. The only route to get the error in the issue (that I can think of) is:
 - spawn a subprocess (which is in a different process group)
 - process uses `tcsetpgrp` on that new process group

Or alternatively, a subprocess does a similar dance. This documentation... doesn't really help. Honestly I'm not sure what would help with the issue.

Anyways, it seems you have been opening too many PRs recently... 68 in the last 2 weeks! Unfortunately I think the open issues on Trio need a bunch of thought to how to solve them, so I'm going to close this. Feel free to reopen it whenever you've slowed down how many PRs you make!

PR #3429, opened 2026-04-21
Experiment with `setup-uv`

A5rocks:
Unfortunately, it looks like `setup-python` takes ~1 minute on average for Windows + x64 + freethreaded. In theory this should be faster, let's see.

(It looks like our slowest runs are e.g. pypy on Windows, which takes 6 minutes -- 1 minute in setup-python. Therefore maybe this will lead to end-to-end CI times taking 1 minute less? Let's see!)

Preliminary results: looks like it takes 6 seconds now! yay! (plus `uv pip install pip` which takes another 6 (?!?) but that's temporary)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,24 +168,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.10', '3.11', '3.12', '3.13', '3.14', '3.14t']
-        arch: ['x86', 'x64']
+        python: ['cpython-3.10', 'cpython-3.11', 'cpython-3.12', 'cpython-3.13', 'cpython-3.14', 'cpython-3.14+freethreaded']
+        arch: ['x86', 'x86_64']
         lsp: ['']
         lsp_extract_file: ['']
         extra_name: ['']
         include:
-          - python: '3.10'
-            arch: 'x64'
+          - python: 'cpython-3.10'
+            arch: 'x86_64'
             lsp: 'https://raw.githubusercontent.com/python-trio/trio-ci-assets/master/komodia-based-vpn-setup.zip'
             lsp_extract_file: 'komodia-based-vpn-setup.exe'
             extra_name: ', with Komodia LSP'
-          - python: '3.10'
-            arch: 'x64'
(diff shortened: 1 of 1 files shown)


A5rocks:
I wanted to use `setup-uv` for everything but I was having trouble verifying apport is installed... so let's push that off until https://github.com/python-trio/trio/issues/3430 is fixed

.github/workflows/ci.yml
```diff
@@ -211,11 +211,11 @@ jobs:
...
       - name: Setup Python
-        uses: actions/setup-python@v6
+        uses: astral-sh/setup-uv@v8.1.0
         with:
-          python-version: '${{ matrix.python }}'
-          architecture: '${{ matrix.arch }}'
-          allow-prereleases: true
+          python-version: ${{ case(startsWith(matrix.python, 'pypy'), matrix.python, format('cpython-{0}', matrix.python)) }}-windows-${{ matrix.arch }}-none
+          activate-environment: "true"
+      - run: uv pip install pip
```
mikenerone: I haven't tried this at all to see if all of the usages work, but maybe `uv pip install pip-uv` might install faster _and_ execute faster (it's just makes `pip` a wrapper for `uv pip`).

A5rocks:
```suggestion
      - run: uv pip install pip-uv
```

A5rocks: Unfortunately, it looks like `pip-uv` doesn't support `python -m pip`... https://github.com/python-trio/trio/actions/runs/25029493735/job/73307906814?pr=3429#step:5:200

mikenerone: Ah, good point. I've opened guysoft/pip-uv#2, so hopefully that will be addressed in the future.

A5rocks:
OK nevermind, I just realized uv doesn't support `-nightly` pypy versions. Unfortunately, we need these sometimes. (presumably it also doesn't support `-dev` cpython versions? but regardless).

The speedup is very nice and I wish we could have it though... maybe choose `setup-python` only for `-nightly` pypy versions? We used to do something like that for `-dev` cpython versions!

PR #3440, opened 2026-05-18
Remove apport install

### LINKED ISSUE #3430: Remove `python3-apport` from CI
I suspect `setup-python` might create a venv, which might mean that `apport` isn't actually installed? I think we can test via `python -c "import apport_python_hook"`.

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3430

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,11 +294,6 @@ jobs:
         run:
           python -m pip install tox &&
           tox -m check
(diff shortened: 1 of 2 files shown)


A5rocks:
LOL I just realized we don't need apport tests anymore, since we don't do MultiErrors. Let's just remove this

CoolCat467 (approved):

PR #3442, opened 2026-05-22
Add impure `flake.nix` dev-shell using `uv` (possibly a proper `<nixpkgs>` overlay coming)

goodboy:
Based on [`pyproject.nix`][pyproject.nix]'s [impure
template][impure_template], this overlays a major-version-pinned
`cpython` (currently `python313`) so `nix(OS)` devs can drop into
a `nix develop` shell with `uv` (via a [`pip`-migration][uv_w_pip])
managing the python virtual-env.

**More thorough description coming once i refine this a bit more!**

[pyproject.nix]: https://pyproject-nix.github.io/pyproject.nix/
[impure_template]: https://pyproject-nix.github.io/pyproject.nix/templates.html#impure
[uv_w_pip]: https://docs.astral.sh/uv/guides/migration/pip-to-project/

---

### Basic usage (will be incrementally updated)

to open a dev-shell for hacking on `trio` core on a nixos system,

```
nix develop [-c <shell you use>]
```

then you will be in a (default `bash`) shell and can interact with
the python venv using,

```bash
# once inside nix dev-shell
uv sync  # update deps to latest
uv sync --group test  # update all "test"-dependency-group deps
uv run pytest  # run full test suite (requires above line ^)
```

---

### Still TODO

first (commit) draft is **very WIP** but by non-draft form we should
at least have,

- [ ] link to relevant docs,
  * [x] `nix`-related extension libs/projects (`pyproject.toml`, `uv2nix`) 
  * [ ] explain/justify *why* a `flake.nix` (is common) to have in repo.
  * [ ] `uv` pkging and import docs (and related alt options to move
    closer to `uv`-styled `pyproject.toml`).

- [ ] how much further `uv` integration do we want?
  * [ ] importing from `requirements.[in|txt]` files can be taken
    further per,
    https://docs.astral.sh/uv/guides/migration/pip-to-project/#importing-requirements-files
  * [ ] do we want an explicit "development" deps set? 
    https://docs.astral.sh/uv/guides/migration/pip-to-project/#importing-development-dependency-files

- [ ] get (AI to help) on refining this to include a proper nixpkgs
      `overlay` output.

- [ ] write proper usage docs for `nix(OS)` users.
  * [ ] something in the/a sub-`README`?
  * [ ] and/or the sphinx docs?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -45,7 +45,7 @@
               # - [ ] extract-n-translate all `[docs|test]-requirements.txt` deps
               #      into their equiv nxpkgs equivalents.
               #
-              # For ex. we can explicitily add them like the below,
+              # For ex. we can explicitly add them like the below,
               # but ideally we use a translator lib (like
               # `pyproject.nix` and/or `uv2nix`).
               #
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,117 @@
+# An "impure" template thx to `pyproject.nix`,
+# https://pyproject-nix.github.io/pyproject.nix/templates.html#impure
+# https://github.com/pyproject-nix/pyproject.nix/blob/master/templates/impure/flake.nix
+{
+  description = "An impure dev-shell (overlay) using `uv`";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable-small";
+  };
+
+  outputs =
+    { nixpkgs, ... }:
+    let
+      inherit (nixpkgs) lib;
+      forAllSystems = lib.genAttrs lib.systems.flakeExposed;
+    in
+    {
+      devShells = forAllSystems (
+        system:
+        let
+          pkgs = nixpkgs.legacyPackages.${system};
+
+          # XXX NOTE XXX, for now we overlay specific pkgs via
+          # a major-version-pinned-`cpython`
+          cpython = "python313";
+          venv_dir = "py313";
+          pypkgs = pkgs."${cpython}Packages";
+        in
+        {
+          default = pkgs.mkShell {
+
+            packages = [
+              # XXX, ensure sh completions activate!
+              pkgs.bashInteractive
+              pkgs.bash-completion
+
+              # pkgs.xonsh
+              pkgs.uv
+              pkgs.${cpython} # ?TODO^ how to set from `cpython` above?
+              pkgs."${cpython}Packages".setuptools
+
+              # testing-n-tooling
+              # ?TODO, if we want to pull `xxx-requirements.txt` from
+              # nixpkgs directly?
+              # - [ ] extract-n-translate all `[docs|test]-requirements.txt` deps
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -0,0 +1,117 @@
...
+              pkgs.bash-completion
+
+              # pkgs.xonsh
+              pkgs.uv
+              pkgs.${cpython} # ?TODO^ how to set from `cpython` above?
+              pkgs."${cpython}Packages".setuptools
+
+              # testing-n-tooling
+              # ?TODO, if we want to pull `xxx-requirements.txt` from
+              # nixpkgs directly?
```
A5rocks, line 44: It sounds like we can do something with `pyproject-nix.lib.project.loadRequirementsTxt { requirements = builtins.readFile ./test-requirements.txt ; };` based on the docs for https://pyproject-nix.github.io/pyproject.nix/use-cases/requirements.html. Though maybe that's impossible...

flake.nix
```diff
@@ -0,0 +1,117 @@
...
+              export UV_PROJECT_ENVIRONMENT=${venv_dir}
+
+              # - sync the `uv` env with all extras, namely the test
+              #   and docs deps as required by the full CI/CD
+              #   pipeline used on Github.
+              uv add -r test-requirements.in -c test-requirements.txt --group test
+
+              # ?TODO? next line is unneeded since
+              # `test-requirements.[in|txt]` already is super set of
+              # `docs-requirements.[in|txt]`?
```
A5rocks, line 88: It shouldn't be; actually this is in general an issue because they may be out of sync. This isn't true right now, but I remember at some point Sphinx was pinned to a different version between them.

flake.nix
```diff
@@ -0,0 +1,117 @@
...
+              # uv add -r docs-requirements.in -c docs-requirements.txt --group docs
+
+              # NOTE, a `tox.ini` is included in repo even though no
+              # explicit dependency is declared elsewhere.
+              uv add --group test tox
```
A5rocks, line 93: Huh I thought it was somewhere. We should probably add it to some requirements file because it should be pinned...

flake.nix
```diff
@@ -0,0 +1,117 @@
...
+
+              # generate a `uv.lock`
+              uv lock
```
A5rocks, line 96: Huh, why's this necessary if you're not using uv2nix in this?

flake.nix
goodboy: Woo good find, yes will check it out!

flake.nix
goodboy: Ok i can also add that to my todo list here 👍🏼

flake.nix
goodboy: It's not necessary, it's just the fastest build sys and generally primes us for if we wanted to use `uv2nix`. I can remove it if you want and/or switch to `pip-tools` or wtv y'all prefer 🫡

PR #3443, opened 2026-05-24
_highlevel_open_tcp_stream: drop duplicate 'the' from pick-port comment

c-tonneslan:
Tiny comment fix in `src/trio/_highlevel_open_tcp_stream.py`. The comment block on the IP_BIND_ADDRESS_NO_PORT path read "connect() is allowed to pick the the port". One `the` is enough.


A5rocks:
Thanks, is there a tool you used to find this? (just an LLM?)

c-tonneslan:
Yeah, LLM-assisted while I was reading through the file. They're pretty solid at flagging doubled words in comment text. For actual code changes I always read through myself, but for plain English typos in comments they catch these reliably.

PR #3434, opened 2026-04-29
Switch to PyPy nightly

A5rocks:
See https://github.com/python-trio/trio/issues/3433, let's see if nightly PyPy has already fixed the issue.

(also a PR for debugging if not)

PR #3447, opened 2026-06-01
Revert "Switch to PyPy nightly"

### LINKED ISSUE #3433: Newest PyPy started failing
It looks like the [v7.3.22 release](https://pypy.org/posts/2026/04/pypy-v7322-release.html) of PyPy has made CI start to fail... I don't really get why at a glance and I don't have the time to dig deeper yet.

We should probably:
 - diagnose the issue and report it to pypy (unless it's already been reported? it doesn't look like it has been reported yet.)
 - hardcode the pypy version we test to 7.3.21 if possible, so that CI starts passing... or maybe the nightly version (if that's still a thing?) if the pypy fix comes fast enough?

illia-v:
This reverts #3434 (closes #3433) because PyPy v7.3.23 was released.

https://pypy.org/posts/2026/05/pypy-v7322-release.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -216,6 +216,7 @@ jobs:
           python-version: '${{ matrix.python }}'
           architecture: '${{ matrix.arch }}'
           allow-prereleases: true
+          check-latest: true
       - name: Run tests
         run: ./ci.sh
         shell: bash
@@ -281,6 +282,7 @@ jobs:
           cache: pip
           cache-dependency-path: test-requirements.txt
           allow-prereleases: true
+          check-latest: true
       - name: Setup minimum Python version
         if: matrix.check_formatting == '1'
         uses: actions/setup-python@v6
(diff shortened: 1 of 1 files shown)


illia-v:
Hmm, the PyPy job failed.

I opened this PR after seeing this fixed for urllib3, this means that without [`check-latest: true`](https://github.com/urllib3/urllib3/blob/73186f9271ff56eb53876f282b57a445e38e1254/.github/workflows/ci.yml#L115) `actions/setup-python` still installs the previous version of PyPy.

A5rocks:
Interesting, I didn't know about that configuration! I think we should enable that.

.github/workflows/ci.yml
A5rocks, line 218:
```suggestion
          allow-prereleases: true
          check-latest: true
```

A5rocks, line 284:
```suggestion
          allow-prereleases: true
          check-latest: true
```

A5rocks, line 345:
```suggestion
          allow-prereleases: true
          check-latest: true
```

PR #3450, opened 2026-06-09
Try bumping codecov-action to v7

A5rocks:
This should fix recent CI failures.


A5rocks:
It turns out this isn't necessary anymore: https://github.com/codecov/codecov-action/issues/1956#issuecomment-4655005815

It's probably still a good idea, but we should just generally upgrade the actions versions I think.

PR #3453, opened 2026-06-23
Switch to pyrepl

### LINKED ISSUE #3007: Ctrl+C behavior problems at the new REPL

A5rocks:
This cleans up some code and makes Linux support work even on some systems configured without TIOCSTI. Fixes https://github.com/python-trio/trio/issues/3007, closes https://github.com/python-trio/trio/pull/3306.

I didn't try this out on Windows, and pyrepl simply cannot be tested easily in CI on Windows, so that may be broken. Please try this out!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,6 +301,7 @@ jobs:
         run: ./ci.sh
         env:
           NO_TEST_REQUIREMENTS: '${{ matrix.no_test_requirements }}'
+          TERM: xterm
       - if: >-
           always()
           && matrix.check_formatting != '1'
@@ -346,6 +347,8 @@ jobs:
           check-latest: true
       - name: Run tests
         run: ./ci.sh
+        env:
+          TERM: xterm
       - if: always()
         uses: codecov/codecov-action@v5
         with:
--- a/newsfragments/3007.bugfix.rst
+++ b/newsfragments/3007.bugfix.rst
@@ -1,0 +1,6 @@
+Switch ``python -m trio`` to use ``pyrepl``. This means that keyboard
+interrupts should be processed more consistently, as well as matching
+the builtin Python REPL more. One downside is that now using the trio
+REPL requires PyPy, CPython 3.13+, or installing the ``pyrepl`` package
+from PyPI. We recommend the first two options, as our support for the
+version on PyPI is worse.
--- a/src/trio/_repl.py
+++ b/src/trio/_repl.py
@@ -1,47 +1,58 @@
 from __future__ import annotations
 
 import ast
-import contextlib
 import inspect
 import sys
 import warnings
 from code import InteractiveConsole
+from signal import SIGINT, raise_signal, signal
 from types import CodeType, FrameType, FunctionType
 from typing import TYPE_CHECKING
 
 import outcome
 
 import trio
-import trio.lowlevel
 from trio._util import final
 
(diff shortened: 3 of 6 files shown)


A5rocks:
I took the opportunity to run the REPL input loop in the main thread too, since it seemed easy enough. One downside is that, since the version of pyrepl on PyPI has a bug w/r/t non-blocking operation, CPython w/ 3.12 or less will have a worse experience in the REPL.

A5rocks:
Also existing tests are a bit flakey so this removes some flakiness from the test suite too. I'll try to figure out how to test this on Windows soon!

PR #3455, opened 2026-06-26
Upgrade all actions

A5rocks:
I kind of blindly just updated everything to their latest available version. We can do something better in autodeps once we have pinning enabled, but I figure this is enough for now.

(anyways, if anything fails, that's ok... we can just fix it.)


CoolCat467 (approved):
Doesn't look like there are any CI failures.

PR #3460, opened 2026-06-30
Fix Path.as_uri on Python 3.14

nyxst4ck:
## Summary

- route `trio.Path.as_uri()` through the concrete wrapped `pathlib.Path` class
- remove the Python 3.14 xfail now that the method no longer calls the deprecated `PurePath.as_uri()` path

## Tests

- `uv run pytest src/trio/_tests/test_path.py -q`
- `git diff --check`

### DIFF
--- a/src/trio/_path.py
+++ b/src/trio/_path.py
@@ -263,9 +263,8 @@ class Path(pathlib.PurePath):
     if sys.version_info >= (3, 13):
         full_match = _wrap_method(pathlib.Path.full_match)
 
-    # TODO: only allow this for Python <3.19.
     def as_uri(self) -> str:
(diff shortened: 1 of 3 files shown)


A5rocks (approved):
Thanks, this is fine. For future contributions to Trio, please don't write PR text/comments with an LLM!

src/trio/_path.py
```diff
@@ -263,9 +263,8 @@ def __repr__(self) -> str:
     if sys.version_info >= (3, 13):
         full_match = _wrap_method(pathlib.Path.full_match)
 
-    # TODO: only allow this for Python <3.19.
```
jakkdl, line 266: `3.19` 🤭

PR #3461, opened 2026-06-30
Clarify wrap_file lifecycle docs

### LINKED ISSUE #3379: trip.wrap_file() undocumented behavior wrt to sync file object

ishaanlabs-gg:
Fixes #3379.

This clarifies the asynchronous file object notes for `trio.wrap_file()`:

- `aclose()` on the async wrapper closes the wrapped synchronous file object.
- the async wrapper does not expose a synchronous `close()` shortcut.
- Trio does not guarantee that garbage collection of the async wrapper closes the wrapped file.
- continuing to use the original synchronous file object while also using the async wrapper has the same concurrency caveat as using the async wrapper from multiple tasks.

I also added a `3379.doc.rst` newsfragment.

Validation:

- `PATH=/Volumes/STOREJET/2/Dependencies/trio/docs-venv-313/bin:$PATH /Volumes/STOREJET/2/Dependencies/trio/docs-venv-313/bin/sphinx-build -b html -D 'exclude_patterns=._*' source build/html`

Notes:

- The docs environment was installed from `docs-requirements.txt` into an external SSD venv using Python 3.13.
- A Python 3.14 docs environment failed while building the pinned `immutables==0.21` wheel, before Sphinx ran. Python 3.13 installed the pinned docs requirements cleanly.
- The `exclude_patterns=._*` override is only for this local external drive, which creates AppleDouble `._*` metadata files that Sphinx otherwise treats as source files.

### DIFF
--- a/docs/source/reference-io.rst
+++ b/docs/source/reference-io.rst
@@ -690,6 +690,12 @@ Asynchronous file objects
      guaranteed to close the file before returning, even if it is
      cancelled or otherwise raises an error.
 
+     Calling ``aclose`` on an async file object also closes the
+     underlying synchronous file object. Trio does not add a
+     synchronous ``close`` method to the async wrapper, and does not
+     guarantee that the wrapped file will be closed automatically when
+     the async wrapper is garbage collected.
+
    * Using the same async file object from multiple tasks
      simultaneously: because the async methods on async file objects
      are implemented using threads, it's only safe to call two of them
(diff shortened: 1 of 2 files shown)


A5rocks:
Sorry but making Claude write docs adds zero value -- people could just ask it. Anyways, tests should be added to make sure we aren't lying in the docs.

PR #3462, opened 2026-06-30
Add spawn_in_current_thread option to run_process/open_process

### LINKED ISSUE #3360: run_process ignores thread context and spawns from threadpool
Hi,

`run_process` uses `trio.to_thread.run_sync` to spawn the subprocess; this means that the spawned process will inherit various context from a different thread.  In my case this means it spawns in the wrong (network) namespace (see [setns(2)](https://manpages.debian.org/testing/manpages-dev/setns.2.en.html) - "reassociate thread with a namespace").

This might affect other things as well ([pthreads(7)](https://manpages.debian.org/testing/manpages/pthreads.7.en.html) lists (among other things) capabilities and CPU affinity).

1. the docs should mention this behavior
2. it would be nice to have a flag to "spawn from current thread"

I'm also open to good suggestions how to work around this problem :)
(I might try prefixing the target command with `["/usr/bin/nsenter", "--target", threading.get_native_id(), "-n", "--"]`, or copy the trio run_process/open_process functions).

cheers,
Stefan

Vansh-Sharma27:
Fixes #3360

## Root cause

`run_process`/`open_process` call `subprocess.Popen()` via
`trio.to_thread.run_sync`. Trio's `to_thread.run_sync` doesn't spin up
a fresh OS thread per call — it pulls an existing worker thread out of
`THREAD_CACHE` (`src/trio/_core/_thread_cache.py`) if one is idle, and
only falls back to creating a new one if the cache is empty
(`ThreadCache.start_thread_soon`, which does `self._idle_workers.popitem()`
before deciding to spawn). That cache is what makes `to_thread.run_sync`
cheap to call repeatedly, but it also means the thread that ends up
calling `Popen()` for any given `run_process` call is whatever thread
happened to finish its previous job and get parked in the cache. It is
not the thread that called `run_process`, and it's not a fresh thread
either.

Python-level state (contextvars, etc.) gets propagated to that worker
thread correctly, which is why this is easy to miss. But OS-level
per-thread state is a kernel-level property of the specific thread,
not something Python copies around: `setns(2)` to move a thread into a
different Linux network namespace, thread-directed capabilities, and
CPU affinity (`pthreads(7)`) all stick to whichever OS thread you set
them on. If you set one of those up on your own thread and then call
`run_process`, the child is actually forked from a leftover cached
worker thread that never had your namespace/capabilities/affinity
applied to it in the first place, so it silently inherits whatever
that worker thread's state happens to be instead. That's exactly the
network namespace problem the issue reporter (tik-stbuehler) ran into.

The issue thread also discusses spawning a dedicated thread per call
with `clone(2)` so it can inherit the caller's namespace, but
tik-stbuehler pushed back on that in the comments since it throws away
the whole point of having a thread cache (or requires a per-trio-loop
thread-local pool, which is a much bigger structural change and was
posted as a workaround, not a proposed fix). The issue itself asks for
two more modest things instead: document the behavior, and add a flag
to spawn from the current thread. This PR does both.

## Fix

Added `spawn_in_current_thread: bool = False` to both `run_process`
and `open_process`. When `True`, `Popen()` is called synchronously on
the calling thread instead of being handed off to
`trio.to_thread.run_sync`, so the child reliably gets the calling
thread's OS-level state because it really is forked from that thread.
This obviously blocks the calling thread (and the event loop, if
called from Trio's main task) for as long as `Popen()` takes, which is
normally fast but is still a real synchronous blocking call, so it's
opt-in and the default behavior/performance is completely unchanged.

Also added the `.. note::` docs paragraph the issue asked for on both
`run_process` and `open_process` (and on the `TYPE_CHECKING` stubs
used for Windows / the Unix overloads, so docs stay consistent across
the type-stub split this module already has).

## Testing

The namespace scenario from the issue needs root/`CAP_SYS_ADMIN` and
real network namespace setup, which isn't practical to run in CI. CPU
affinity is gated by the exact same bug mechanism (it's also per-thread
OS state that doesn't follow a thread cache hand-off) but is settable
by an unprivileged process via `os.sched_setaffinity`, so I used that
as a stand-in to get an actual reproduction instead of just testing
that the option exists:

- `test_spawn_in_current_thread_affinity` (Linux-only, skipped on
  macOS/Windows since `sched_getaffinity`/`sched_setaffinity` aren't
  available there): runs one `run_process` call first to warm up
  `THREAD_CACHE` with a worker thread while affinity is unrestricted,
  then restricts the *calling* thread's affinity to a single CPU and
  checks that a child spawned the default way still reports the old,
  unrestricted mask (because it came from the stale cached worker
  thread), while a child spawned with `spawn_in_current_thread=True`
  reports the restricted mask. This fails without the fix and passes
  with it.
- `test_spawn_in_current_thread_equivalence`: cross-platform sanity
  check that `run_process`/`open_process` behave the same with the
  flag on vs. off (stdin/stdout/return code), since the flag should
  only change which thread calls `Popen()`, nothing about the
  resulting process.
- Added the matching `Unpack[...]` kwarg to the Windows/Unix
  `TYPE_CHECKING` overloads and a `type_tests/subprocesses.py` entry so
  the new kwarg type-checks on every overload shape.
- Added `newsfragments/3360.feature.rst`.

Ran the project's full check suite locally before opening this:
`pytest src` (825 passed, 71 skipped, 2 xfailed — all pre-existing,
unrelated to this change), `mypy --platform linux/darwin/win32`,
`pyright` on both `type_tests` dirs, `check_type_completeness.py`,
`gen_exports.py --test`, and `pre-commit run --all-files`. All green.

### DIFF
--- a/newsfragments/3360.feature.rst
+++ b/newsfragments/3360.feature.rst
@@ -1,0 +1,8 @@
+`trio.run_process` and `trio.lowlevel.open_process` now accept a
+``spawn_in_current_thread`` keyword argument. By default the child process is
+spawned from a worker thread in trio's internal thread pool, so OS-level
+per-thread state on the calling thread (network namespace, capabilities, CPU
+affinity, etc.) is not necessarily reflected in the child. Passing
+``spawn_in_current_thread=True`` spawns the child directly from the calling
+thread instead, at the cost of briefly blocking the event loop while
+``subprocess.Popen`` runs.
--- a/src/trio/_subprocess.py
+++ b/src/trio/_subprocess.py
@@ -309,6 +309,7 @@ async def _open_process(
     stdin: int | HasFileno | None = None,
     stdout: int | HasFileno | None = None,
     stderr: int | HasFileno | None = None,
+    spawn_in_current_thread: bool = False,
     **options: object,
 ) -> Process:
     r"""Execute a child program in a new process.
@@ -327,6 +328,16 @@ async def _open_process(
     Unlike `trio.run_process`, this function doesn't do any kind of automatic
     management of the child process. It's up to you to implement whatever semantics you
     want.
+
+    .. note:: By default, the child process is spawned from a worker thread
+       drawn from Trio's internal thread pool, not from the thread that calls
+       `open_process`. Most Python-level state (e.g. context variables) is
+       copied across, but OS-level per-thread state is not: things like the
+       Linux network namespace (:manpage:`setns(2)`), capabilities, and CPU
+       affinity (:manpage:`pthreads(7)`) on the worker thread can differ from
+       what you set up on the calling thread, and the child inherits the
+       worker thread's state instead. Pass ``spawn_in_current_thread=True``
+       if your child process needs to inherit this kind of state faithfully.
 
     Args:
       command: The command to run. Typically this is a sequence of strings or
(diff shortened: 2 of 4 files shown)


A5rocks:
Blocking the event loop for this is unacceptable, sorry!

PR #3465, opened 2026-07-03
Add docstrings for memory channel classes

Madhur-7568:
## Summary

Add class docstrings for:

- MemorySendChannel
- MemoryReceiveChannel
- MemoryChannelStatistics

This improves the generated API documentation for these classes.

### DIFF
--- a/src/trio/_channel.py
+++ b/src/trio/_channel.py
@@ -119,31 +119,28 @@ class open_memory_channel(tuple["MemorySendChannel[T]", "MemoryReceiveChannel[T]
 @attrs.frozen
 class MemoryChannelStatistics:
     """Statistics describing the current state of a memory channel.
+
     Returned by :meth:`MemorySendChannel.statistics` and
     :meth:`MemoryReceiveChannel.statistics`.
-
-    Currently, the following fields are defined:
-
-    * ``current_buffer_used`` (int): The number of items currently stored
-      in the channel buffer.
-    * ``max_buffer_size`` (int or math.inf): The maximum number of items
-      that can be buffered in the channel.
-    * ``open_send_channels`` (int): The number of open
-      :class:`MemorySendChannel` endpoints pointing to this channel.
-    * ``open_receive_channels`` (int): The number of open
-      :class:`MemoryReceiveChannel` endpoints pointing to this channel.
-    * ``tasks_waiting_send`` (int): The number of tasks currently blocked
-      waiting to send.
-    * ``tasks_waiting_receive`` (int): The number of tasks currently blocked
-      waiting to receive.
     """
 
     current_buffer_used: int
+    """The number of items currently stored in the channel buffer."""
+
     max_buffer_size: int | float
+    """The maximum number of items that can be buffered in the channel."""
+
     open_send_channels: int
+    """The number of open :class:`MemorySendChannel` endpoints pointing to this channel."""
+
     open_receive_channels: int
+    """The number of open :class:`MemoryReceiveChannel` endpoints pointing to this channel."""
+
     tasks_waiting_send: int
+    """The number of tasks currently blocked waiting to send."""
+
     tasks_waiting_receive: int
+    """The number of tasks currently blocked waiting to receive."""
 
 
 @attrs.define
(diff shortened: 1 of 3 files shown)


src/trio/_channel.py
```diff
@@ -118,6 +118,8 @@ def __init__(self, max_buffer_size: int | float) -> None:  # noqa: PYI041
 
 @attrs.frozen
 class MemoryChannelStatistics:
+    """Statistics for a memory channel."""
```
A5rocks: Could you provide docstrings for each attribute too, actually? That would make this substantially more useful.

src/trio/_channel.py
```diff
@@ -152,6 +154,8 @@ def statistics(self) -> MemoryChannelStatistics:
 @final
 @attrs.define(eq=False, repr=False, slots=False)
 class MemorySendChannel(SendChannel[SendType], metaclass=NoPublicConstructor):
+    """A memory channel endpoint for sending Python objects."""
```
A5rocks: Please reference the `open_memory_channel` function, as these cannot be made by users.

src/trio/_channel.py
```diff
@@ -300,6 +304,8 @@ async def aclose(self) -> None:
 @final
 @attrs.define(eq=False, repr=False, slots=False)
 class MemoryReceiveChannel(ReceiveChannel[ReceiveType], metaclass=NoPublicConstructor):
+    """A memory channel endpoint for receiving Python objects."""
```
A5rocks: Same as above

A5rocks (changes requested):
Please appease CI! I'm not too sure about these new docstrings. I'm not sure my review comments are correct, but I think they would be more useful to readers at least.

src/trio/_channel.py
```diff
@@ -118,6 +118,26 @@ def __init__(self, max_buffer_size: int | float) -> None:  # noqa: PYI041
 
 @attrs.frozen
 class MemoryChannelStatistics:
+    """Statistics describing the current state of a memory channel.
+    Returned by :meth:`MemorySendChannel.statistics` and
+    :meth:`MemoryReceiveChannel.statistics`.
+
+    Currently, the following fields are defined:
```
A5rocks: I know the parking lot stats use this format, but could we try per-attribute docs? I think they look like:

```py
class Bweh:
  #: description for x
  x: int
```

though not totally sure. (maybe try it and see the RTD preview)

One advantage is that this way pyright and IDEs would maybe pick up the docstring, so you can hover over the attribute and see the docs.

src/trio/_channel.py
```diff
@@ -152,6 +172,13 @@ def statistics(self) -> MemoryChannelStatistics:
 @final
 @attrs.define(eq=False, repr=False, slots=False)
 class MemorySendChannel(SendChannel[SendType], metaclass=NoPublicConstructor):
+    """
+    A memory channel endpoint for sending Python objects.
+
+    Instances of this class are created by
+    :func:`open_memory_channel` and cannot be instantiated directly.
+    """
```
A5rocks, line 176: nit: could you add a trailing newline to the other docstring. (I forgot our convention, but this one mismatches the other)

A5rocks (approved):
Thanks for the changes. Just a couple nitpicks!

src/trio/_channel.py
Madhur-7568: ok ,I will update these.

src/trio/_channel.py
A5rocks: Looking at ty, it seems we need to do it like:
```py
class Bweh:
  x: int
  """Description here"""
```

playground: https://play.ty.dev/628751d2-4ce5-44f0-8df7-d36c81920024

src/trio/_channel.py
```diff
@@ -300,6 +324,13 @@ async def aclose(self) -> None:
 @final
 @attrs.define(eq=False, repr=False, slots=False)
 class MemoryReceiveChannel(ReceiveChannel[ReceiveType], metaclass=NoPublicConstructor):
+    """
+    A memory channel endpoint for receiving Python objects.
```
A5rocks:
```suggestion
    """A memory channel endpoint for receiving Python objects.
```

Madhur-7568:
I've made all the changes. Please review when you have time. Thanks!

A5rocks (approved):

Madhur-7568:
Hi! Thanks again for reviewing and approving the PR. Just wanted to check if there's anything else needed from my side before it can be merged. Thanks!

A5rocks:
Nothing else is necessary from you, I'm just waiting for someone else to approve/merge.

CoolCat467 (approved):
"for python objects" seemed a little weird but I can't think of a better description myself, looks good.

PR #3468, opened 2026-07-09
Skip SSL tests on Windows x86 CI

### LINKED ISSUE #3466: Don't run SSL tests on Windows x86

yashkuceriya:
Closes #3466

## Summary
- set the Windows CI job to mark x86 runs with SKIP_SSL_TESTS=1
- filter out the SSL-only test dependencies on that path, avoiding cryptography on unsupported 32-bit Windows
- ignore Trio's SSL, DTLS, and high-level SSL helper test modules only for that marked CI path

## Validation
- bash -n ci.sh
- git diff --check
- verified the filtered requirements remove only cryptography, pyopenssl, and trustme while preserving the rest of the test requirements
- checked the first CI failure and updated the skip list to include test_highlevel_ssl_helpers.py, which imports SSL test fixtures

No newsfragment added because this is CI-only maintenance with no user-visible behavior change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,6 +223,7 @@ jobs:
         env:
           LSP: '${{ matrix.lsp }}'
           LSP_EXTRACT_FILE: '${{ matrix.lsp_extract_file }}'
+          SKIP_SSL_TESTS: ${{ matrix.arch == 'x86' && '1' || '0' }}
       - if: always()
         uses: codecov/codecov-action@v7
         with:
--- a/ci.sh
+++ b/ci.sh
@@ -54,7 +54,14 @@ if [ "${NO_TEST_REQUIREMENTS-0}" == 1 ]; then
     python -m uv pip install pytest coverage -c test-requirements.txt
     flags="--skip-optional-imports"
 else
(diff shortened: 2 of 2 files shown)


CoolCat467:
This feels like an LLM pull request to me, please actually review your pull requests instead of outsourcing. This does not solve the problem as you can see if you look at the actions run results, and I think in the end this is going to have to come down to making trio able to work without cryptography installed albeit with a more limited feature set if cryptography is not installed, just dropping support for x86 windows, or trying to have trio work with the old version that supported x86 windows at the same time as the new version (but this sounds hard and would likely increase maintenance load). Personally leaning towards option two and just telling x86 windows people to use an older version of trio, but would like to hear other's thoughts.

A5rocks:
My PoV is that the only place where Trio needs PyOpenSSL is dtls and I don't think we even depend on it. (PyOpenSSL uses cryptography iirc) I guess it wasn't clear but the deprecation warning is from the cryptography package, not built in SSL.

So I was thinking just a pytest.skip in the relevant test files. (but I didn't quite realize trustme depends on it too, so IDK if it's possible)

I agree this PR is bad though!

PR #3473, opened 2026-07-15
ThreadCache workers leaking spawner's Context on py314+ free-threaded builds

### LINKED ISSUE #3472: ThreadCache workers leak the spawning thread's contextvars.Context on free-threaded builds
**Environment**
- Python 3.14.2 free-threading build
- trio: 0.33.0
- OS: Ubuntu 24.04.4 LTS

On python314t+ free threaded build, `ThreadCache`'s pooled
`WorkerThreads` can retain a strong reference to whatever `contextvars.Context`
was active in the thread that first caused them to be spawned for the life of the pooled thread.

**Repro**
```python
import contextvars
import gc
import sys
import trio

cvar: contextvars.ContextVar[object] = contextvars.ContextVar("cvar")

async def main() -> None:
    class Foo:
        pass

    def sync_fn():
        pass

    contextval = Foo()
    cvar.set(contextval)
    await trio.to_thread.run_sync(sync_fn)
    cvar.set(Foo())  # drop our own reference; nothing legitimate should still hold contextval
    gc.collect()

    referrers = [
        r for r in gc.get_referrers(contextval)
        if type(r).__name__ != "coroutine"  # this frame's own locals; expected
    ]
    gil = sys._is_gil_enabled() if hasattr(sys, "_is_gil_enabled") else "N/A (<3.13)"
    print(f"GIL enabled: {gil}")
    print(f"unexpected referrers: {referrers}")

    assert not referrers, f"leaked referrers after to_thread.run_sync: {referrers}"

trio.run(main)
```
python 314t: 
```
GIL enabled: False
unexpected referrers: [<hamt_bitmap_node object at 0x3f9b2fb2c40>]
  ......
AssertionError: leaked referrers after to_thread.run_sync: [<hamt_bitmap_node object at 0x3f9b2fb2c40>]
```
python 314:
```
GIL enabled: True
unexpected referrers: []
```
**expected**
```
GIL enabled: True
unexpected referrers: []
```
**Related**:
https://github.com/agronholm/anyio/issues/1220

EmmanuelNiyonshuti:
fixes #3472


src/trio/_core/_thread_cache.py
```diff
@@ -151,8 +152,12 @@ def __init__(self, thread_cache: ThreadCache) -> None:
         self._worker_lock = Lock()
         self._worker_lock.acquire()
         self._default_name = f"Trio thread {next(name_counter)}"
-
-        self._thread = Thread(target=self._work, name=self._default_name, daemon=True)
+        kwargs: dict[str, Any] = {}  # type: ignore[explicit-any]
+        if sys.version_info >= (3, 14):
+            kwargs["context"] = Context()
```
A5rocks: I'd rather duplicate the `Thread` call TBH. Then we can typecheck the arguments. It's not that long either.

newsfragments/3472.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed ``ThreadCache``` workers leaking spawner's ``Context`` on ``py314t+`` free-threaded builds.
```
A5rocks: I'd rather this links to the public API that's impacted (in this case `trio.to_thread.run_sync`) instead of something that probably people won't understand?

A5rocks (commented):
Makes sense to me. Could you add a test? I guess just something to make sure that threads don't inherit context would work well enough.

src/trio/_core/_thread_cache.py
EmmanuelNiyonshuti: Done

newsfragments/3472.bugfix.rst
EmmanuelNiyonshuti: Done

src/trio/_tests/test_threads.py
```diff
@@ -687,6 +688,34 @@ def g() -> tuple[str, str, threading.Thread]:
...
 
 
+@pytest.mark.skipif(
+    sys.implementation.name == "pypy",
+    reason=(
+        "gc.get_referrers is broken on PyPy (see "
+        "https://github.com/pypy/pypy/issues/5075)"
+    ),
+)
+async def test_worker_thread_context_not_leaked() -> None:
```
A5rocks, line 691: IMO you should just check whether context gets carried over, no need for GC or whatever. Technically there could be a bug where we clear it in the frame before handing it to the user but store it elsewhere -- and you *could* test for that using weakref (to ensure it works on PyPy) -- but IMO not necessary.

A5rocks (approved):
Test could be simplified I think, but looks good to me.

src/trio/_tests/test_threads.py
EmmanuelNiyonshuti: Yeah, we could use weakref too. I leaned  more towards how anyio did it.

EmmanuelNiyonshuti:
Forgive these so many commits. Not sure if you squash commits when merging, but feel free to tidy them up if needed.

CoolCat467 (approved):
Looks good as far as I can tell, though I agree test could probably be simplified.

EmmanuelNiyonshuti:
@A5rocks  what do you think about this approach? used weakref and just asserted that the object is dead.

EmmanuelNiyonshuti:
PyPy failures are now irrelevant because we don't need `gc.get_referrers`  anymore!

EmmanuelNiyonshuti:
By the way, this was found while exercising anyio's test suite on the Python 3.14t free-threaded build. anyio has a regression test that checks for leaked references from its worker-thread pool, and running it against trio's backend surfaced this.

A5rocks:
Thanks for this!

PR #3474, opened 2026-07-16
Track max buffer usage in memory channel statistics

### LINKED ISSUE #1723: feature request: track max queue length in memory channel stats
use case:  if you need queueing but `inf` is not reasonable, to find a reasonable limit you'll want stats on the max observed queuing for actual run sessions of the app

doesn't seem like too much overhead to always collect it

```python
        if self._state.receive_tasks:
            ...
        else:
            n = len(self._state.data)
            if n < self._state.max_buffer_size:
                self._state.data.append(value)
                self._state._max_observed_size = max(n + 1, self._state._max_observed_size)
            else:
                raise trio.WouldBlock
```

pctablet505:
Adds a `max_buffer_used` field to `MemoryChannelStatistics`, tracking the largest number of items the channel's buffer has ever held at once.

The use case: when you need buffering but `math.inf` isn't a reasonable `max_buffer_size`, this makes it possible to observe real-world peak usage during a run and pick a sensible limit, instead of guessing.

Closes #1723


mikenerone:
@pctablet505 One suggestion: I get that the "used" bit of `max_buffer_used` is you trying to be explicit about the meaning, but I'd *still* expect this name to confuse people. `max_anything` sounds like a configured limit rather than a metric. I'd suggest something like `peak_buffer` or `peak_buffer_used`.

jakkdl (approved):

CoolCat467 (approved):

PR #3476, opened 2026-07-22
Update ruff

A5rocks:
RUF105 in particular was being bad, because it would change our `noqa` to `ruff:ignore`, which then `black` looks at and reformats, which ruff would see being on the wrong line and remove + apply fixes arbitrarily around the codebase. Basically a giant mess.

I've disabled it. I wish we could get the nice human readable names (RUF106) without using `ruff:ignore`, but that doesn't seem possible.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.21
+    rev: v0.15.22
     hooks:
       - id: ruff-check
         types: [file]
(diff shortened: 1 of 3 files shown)


CoolCat467 (approved):

CoolCat467:
We could likely avoid these problems if we switched to ruff's formater instead of black, but if we were to do that I think it would be better served for another pull request.

jakkdl:
After the acquiring of astral I am slightly wary of relying even more on them, but it shouldn't be a big jump to revert back to using black.
Even better would be opening an issue / PR for black to handle `ruff:ignore` gracefully

A5rocks:
Done: https://github.com/psf/black/issues/5260

A5rocks:
Hmm actually maybe it's just a matter of putting enough time in. I forgot which exact formatting loops made me give up, but it's probably possible to structure the code *somehow* to appease both black and ruff. I'm just not sure it's worth it. (... maybe I should open an issue here so it gets done eventually, maybe)

PR #3479, opened 2026-07-26
Run GC on PyPy more for test_worker_thread_context_not_leaked

### LINKED ISSUE #3478: `test_worker_thread_context_not_leaked` sometimes fails
cc @EmmanuelNiyonshuti 

I haven't dug into this yet but the test is flakey. I'm opening this instead of simply commenting in the flakey-tests thread so we can discuss!

See https://github.com/A5rocks/trio/actions/runs/30208152029 and https://github.com/A5rocks/trio/actions/runs/29643389046

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3478

### DIFF
--- a/src/trio/_tests/test_threads.py
+++ b/src/trio/_tests/test_threads.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextvars
-import gc
 import queue as stdlib_queue
 import re
 import sys
@@ -29,7 +28,7 @@ from .. import (
     sleep_forever,
 )
 from .._core._tests.test_ki import ki_self
-from .._core._tests.tutil import slow
+from .._core._tests.tutil import gc_collect_harder, slow
 from .._threads import (
     active_thread_count,
     current_default_thread_limiter,
@@ -705,7 +704,8 @@ async def test_worker_thread_context_not_leaked() -> None:
     cvar.set(Foo())
 
     del contextval
-    gc.collect()
+    if sys.implementation.name == "pypy":
+        gc_collect_harder()
 
     assert ref() is None
 
--- a/src/trio/_core/_thread_cache.py
+++ b/src/trio/_core/_thread_cache.py
@@ -188,6 +188,13 @@ class WorkerThread(Generic[RetT]):
             if set_os_thread_name:
                 set_os_thread_name(self._thread.ident, self._default_name)
 
+        # Without this, this thread may get rescheduled after `deliver`
+        # and only return after the kernel gets around to it. That's an
+        # arbitrary amount of time during which we keep anything
+        # referred to in the function alive, like contextvars. That's
+        # surprising!
+        del fn
+
         # Tell the cache that we're available to be assigned a new
         # job. We do this *before* calling 'deliver', so that if
         # 'deliver' triggers a new job, it can be assigned to us


EmmanuelNiyonshuti:
@A5rocks Sorry, I thought it would have no issue on PyPy. let me check again.

A5rocks:
Yes if you have an Ubuntu (presumably kinda slow) machine and could check that this fixes the issue (ie repro w/o + unable to repro with) that would be great.

I'll start a bunch of CI runs in a while otherwise, which will hopefully just provide the same evidence.

EmmanuelNiyonshuti:
> Yes if you have an Ubuntu (presumably kinda slow) machine and could check that this fixes the issue (ie repro w/o + unable to repro with) that would be great.
> 
> I'll start a bunch of CI runs in a while otherwise, which will hopefully just provide the same evidence.

I think I have that; will try and see.

EmmanuelNiyonshuti:
I am failing to reproduce its flakiness on my host.

EmmanuelNiyonshuti:
I think this is better. I wasn't aware of trio's `gc_collect_harder`!

EmmanuelNiyonshuti (approved):

CoolCat467 (approved):

src/trio/_tests/test_threads.py
```diff
@@ -705,7 +704,8 @@ def sync_fn() -> None:
     cvar.set(Foo())
 
     del contextval
-    gc.collect()
+    if sys.implementation.name == "pypy":
+        gc_collect_harder()
```
EmmanuelNiyonshuti, line 708: how about we default to gc.collect on cpython. 
```
if sys.implementation.name == "pypy":
      gc_collect_harder()
else:
      gc.collect()
```
On CPython it would work either way but on FreeThreaded, I think reference counting works differently, and we would run into problems with just this. WDYT?

CoolCat467: `gc_collect_harder` just does multiple passes of `gc.collect`, just do `gc_collect_harder` for both cases. Performance hit from multiple gc runs doesn't really matter that much for a single test.

EmmanuelNiyonshuti: It is only needed for PyPy, but I think going that route won't hurt either.

jakkdl: I think there is some upside to having the `if` to passively document that the behavior differs between platforms. Otherwise somebody running on cpython might be confused why `gc_collect_harder` is needed, tries switching to `gc.collect()`, notices that it still works, opens a PR, it fails on pypy.... lots of time wasted.

A5rocks: Having `gc.collect()` / collect harder certainly fixes the issue, but I'm struggling to understand what the issue actually is here! I added a `gc.set_debug` to log any refcycle cleanups and it didn't show anything. And yet, the  contextvar exists at some point and then later disappears! I assume the reason `gc.collect()` works is because this is very racey, but that's not a good place to be.

A5rocks:
With some extra changes + without that change: https://github.com/A5rocks/trio/actions/runs/30357228537

With some extra changes (same as above) + with that change: https://github.com/A5rocks/trio/actions/runs/30357043985

So I think it's necessary.

src/trio/_tests/test_threads.py
EmmanuelNiyonshuti: @A5rocks Yeah, the test is not super obvious, but we only needed to know if worker thread drops the copied context. I am not sure of a more obvious and simpler way we can really check this.

The **only** reason we have gc.collect is about the semantic difference between free-threaded Python and default CPython which I am obviously no way near being an expert at. **But** the obvious reason in this test and as they document it, is the refcounting difference which is not as immediate the way it is on CPython.  Essentially, for this test  when the worker thread gets done, `contextval` can look like it's still alive even when the worker properly dropped its copy(on free threaded)! I am assuming that's also why  `gc.set_debug` showed nothing since there is no refcycle being involved here. We add `gc.collect` to ensure that what will be left after it would actually be a leak from the worker.

EmmanuelNiyonshuti:
> With some extra changes + without that change: https://github.com/A5rocks/trio/actions/runs/30357228537
> 
> With some extra changes (same as above) + with that change: https://github.com/A5rocks/trio/actions/runs/30357043985
> 
> So I think it's necessary.

you didn't add an else block though? `gc.collect`

EmmanuelNiyonshuti:
> > With some extra changes + without that change: https://github.com/A5rocks/trio/actions/runs/30357228537
> > With some extra changes (same as above) + with that change: https://github.com/A5rocks/trio/actions/runs/30357043985
> > So I think it's necessary.
> 
> you didn't add an else block though? `gc.collect`

perhaps is related to "extra changes," which I don't quite understand yet!

A5rocks:
No else block is necessary, because there's no refcycles actually. The CPython failures turned out to be a race condition where the worker thread would sometimes not get to return (which would decrease the refcount for contextvars, cleaning them up), which calling `gc.collect()` only solved by giving the worker thread more time to return.

EmmanuelNiyonshuti:
> No else block is necessary, because there's no refcycles actually. The CPython failures turned out to be a race condition where the worker thread would sometimes not get to return (which would decrease the refcount for contextvars, cleaning them up), which calling `gc.collect()` only solved by giving the worker thread more time to return.

oh! will that also work on free threaded python, assuming you mean default CPython?

A5rocks:
I don't see why it wouldn't! I don't think free threaded CPython is that different from regular CPython in this case (we only actually notify of task completion once the refcount is decreased).

EmmanuelNiyonshuti:
> I don't see why it wouldn't! I don't think free threaded CPython is that different from regular CPython in this case (we only actually notify of task completion once the refcount is decreased).

hmm... I see! so you somehow found a way to delay/make sure  worker thread returns? Obviously, I wasn't able to reproduce this on CPython, so I kinda narrowed the problem to free threaded,

EmmanuelNiyonshuti:
> > I don't see why it wouldn't! I don't think free threaded CPython is that different from regular CPython in this case (we only actually notify of task completion once the refcount is decreased).
> 
> hmm... I see! so you somehow found a way to delay/make sure worker thread returns? Obviously, I wasn't able to reproduce this on CPython, so I kinda narrowed the problem to free threaded,

FYI the test is still flaky on free threaded. you can try running it couple times.

A5rocks:
No, the worker thread already sends a message about being done (`deliver` in `_handle_job`), all I did was make sure it removed its reference to `worker_fn` (which keeps a strong ref on a copy of contextvars, keeping existing contextvars alive) before saying it's done -- this guarantees that when the `await to_thread.run_sync` finishes, the copy of the contextvars has been deleted.

EmmanuelNiyonshuti:
> I don't see why it wouldn't! I don't think free threaded CPython is that different from regular CPython in this case (we only actually notify of task completion once the refcount is decreased).

The difference I think we're dealing with is refcounting https://docs.python.org/3/howto/free-threading-python.html#:~:text=Because%20the%20per,to%20be%20freed.

EmmanuelNiyonshuti:
> No, the worker thread already sends a message about being done (`deliver` in `_handle_job`), all I did was make sure it removed its reference to `worker_fn` (which keeps a strong ref on a copy of contextvars, keeping existing contextvars alive) before saying it's done -- this guarantees that when the `await to_thread.run_sync` finishes, the copy of the contextvars has been deleted.

hmm... that's more interesting!

Issue #3168, opened 2024-12-23
It might be worth adding zizmor as a GitHub action


graingert:
It might be worth adding it as a GitHub action because I think it can catch more issues if it is passed a GITHUB_TOKEN.

_Originally posted by @A5rocks in https://github.com/python-trio/trio/pull/3154#discussion_r1883589058_

PR #3154, opened 2024-12-13
zizmor is all the rage these days, probably worth adding it

graingert:

PR #3482, opened 2026-07-27
Add zizmor as gh action

EmmanuelNiyonshuti:
Closing https://github.com/python-trio/trio/issues/3168 and following up on  https://github.com/python-trio/trio/pull/3154#discussion_r1883589058


A5rocks (approved):
Thanks! Maybe it's a good idea to remove zizmor from pre-commit (if we have it there), to keep the versions identical. Also, maybe it would be nice to require the zizmor action? I.e. put it in the ci.yml file and as a requirement to the required workflow.

But not completely sure about either idea...

EmmanuelNiyonshuti:
Pointing all our actions to zizmor passes on all zizmor operating modes🎉, we also have it in pre-commit, so it was straightforward to add the action. I don't know much about this tool but there probably are still room for improvements like I saw that attrs uses a pedantic persona https://github.com/python-attrs/attrs/blob/main/.github/workflows/zizmor.yml#L31

more on personas: https://docs.zizmor.sh/usage/#using-personas

EmmanuelNiyonshuti:
> Thanks! Maybe it's a good idea to remove zizmor from pre-commit (if we have it there), to keep the versions identical. Also, maybe it would be nice to require the zizmor action? I.e. put it in the ci.yml file and as a requirement to the required workflow.
> 
> But not completely sure about either idea...

IMO i don't think its a bad idea to require it in ci.yml **BUT** i  have never seen it in other projects so far, I think I didn't see it in zizmor itself too? not sure though plus I don't understand the need beyond what it would do if not required in ci.yml.

A5rocks:
> IMO i don't think its a bad idea to require it in ci.yml BUT i have never seen it in other projects so far, I think I didn't see it in zizmor itself too?

To be clear I'm talking about how we have a single workflow (it uses `alls-green` but I forgot exact naming...) required (so that changes to required workflows can be done via PR rather than through the repo settings). I guess we need to put this in ci.yml to make the zizmor run required?

EmmanuelNiyonshuti:
> > IMO i don't think its a bad idea to require it in ci.yml BUT i have never seen it in other projects so far, I think I didn't see it in zizmor itself too?
> 
> To be clear I'm talking about how we have a single workflow (it uses `alls-green` but I forgot exact naming...) required (so that changes to required workflows can be done via PR rather than through the repo settings). I guess we need to put this in ci.yml to make the zizmor run required?

Sorry, i may have slightly misunderstood what you was suggesting. I think the job you are referring to is this one, no? https://github.com/python-trio/trio/blob/3769b0731108557fbd40b0b590fe6227103986ac/.github/workflows/ci.yml#L473-L474

and I can call zizmor from here as well.

.github/workflows/ci.yml
```diff
@@ -470,6 +470,12 @@ jobs:
           flags: Cython,${{ steps.get-version.outputs.version }}
           fail_ci_if_error: true
 
+  zizmor:
+    name: zizmor
+    permissions:
+      security-events: write
```
EmmanuelNiyonshuti: This was explicitly added here because I was suggested that zizmor job would inherit this job's `permissions: {}`

.github/workflows/ci.yml
```diff
@@ -470,6 +470,12 @@ jobs:
...
+    uses: ./.github/workflows/zizmor.yml
```
A5rocks: I think it might be cleaner to just put in the zizmor job here, instead of a separate file. Other than that, this looks great!

EmmanuelNiyonshuti: Done!

A5rocks (approved):

.github/workflows/ci.yml
```diff
@@ -470,6 +470,20 @@ jobs:
...
+    permissions:
+      security-events: write # Required for upload-sarif (used by zizmor-action) to upload SARIF files.
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor 🌈
+        uses: zizmorcore/zizmor-action@6fc4b006235f201fdab3722e17240ab420d580e5 # v0.6.1
```
jakkdl: do we have a bot/workflow that will update these pinned versions? Pinning it to a specific hash otherwise seems like it'll quickly go stale. When we had it in pre-commit it was handled by the pre-commit bot.

jakkdl: perhaps related, why does this specify `# v0.6.1` when the pre-commit was using `v1.28.0`

A5rocks: Yeah, I guess non-hashes would be better for now. We want to eventually switch to locked actions (whether that's through hashes or through GitHub's native thing (coming soon:tm:)...), at which point we can figure out updating :-)

EmmanuelNiyonshuti: Granted: It was almost copy pasted from zizmor integration docs.

EmmanuelNiyonshuti: > perhaps related, why does this specify `# v0.6.1` when the pre-commit was using `v1.28.0`

`# v0.6.1`  is the action's version and the latter was for the zizmor version used by pre-commit?

EmmanuelNiyonshuti:
the ci failures seems related to codecov. I will investigate.

CoolCat467:
Given running things on `pre-commit.ci` is free, why drop running pre-commit hook version of zizmor? In the event someone changes a github action and something breaks, if they do pre-commit checks it'll save them a whole cycle of looking at github action logs on github. If we don't want `pre-commit.ci` to run zizmor and only run locally, we can just add zizmor to the `ci.skip` section in pre-commit.

I will admit though, this will introduce version matching issues, would likely want a tool similar to `sync_requirements` but changing the github actions version instead or just adding this functionality onto the existing tool.

A5rocks:
I suggested it cause the version matching, yeah. (and also since the action will be able to check more since it has a github token, so passing one doesn't mean passing the other)

I'd be fine with whatever outcome is best.

PR #3483, opened 2026-07-27
Skip SSL tests on Windows x86

### LINKED ISSUE #3466: Don't run SSL tests on Windows x86

yashkuceriya:
Fixes #3466

## Summary

- skip the SSL, DTLS, and high-level SSL helper test modules on Windows x86
- keep the rest of the Windows x86 CI job coverage intact
- make the skip decision in pytest instead of adding more ci.sh branching

## Testing

- source .venv313/bin/activate && ruff check src/trio/_tests/test_ssl.py src/trio/_tests/test_dtls.py src/trio/_tests/test_highlevel_ssl_helpers.py
- source .venv313/bin/activate && PYTHONPATH=src pytest src/trio/_tests/test_ssl.py src/trio/_tests/test_dtls.py src/trio/_tests/test_highlevel_ssl_helpers.py -q
- source .venv313/bin/activate && PRE_COMMIT_HOME=/private/tmp/pre-commit-trio pre-commit run --files src/trio/_tests/test_ssl.py src/trio/_tests/test_dtls.py src/trio/_tests/test_highlevel_ssl_helpers.py

### DIFF
--- a/src/trio/_tests/test_dtls.py
+++ b/src/trio/_tests/test_dtls.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import random
+import struct
+import sys
 from contextlib import asynccontextmanager
 from itertools import count
 from typing import TYPE_CHECKING, NoReturn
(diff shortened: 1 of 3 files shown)


A5rocks:
Please do not make any more PRs with an LLM-generated comment. If you do, I will close them. (you can write the simplest comments possible, that's fine!)

On an object level, IMO this is the wrong approach. I like passing in whether something should expect the cryptography module to be importable, then having an explicit error if that flag is wrong (in either direction!). Both implicitly checking things and using a skip are not right.

PR #3496, opened 2026-08-24
[WIP] Allow cancellation between reschedule() called and task resumed

### LINKED ISSUE #2973: Cancellation may not always immediately happen at checkpoints
### Short summary

My reading of the documentation is that a `await ...` should *always* end in a `Cancelled` being raised if it's in an enclosing cancel scope that is cancelled, but that is not actually always the case in practice. I suspect this is a documentation issue (or I'm just reading too much into it), but even if so then it could do with clarifying.

### Documentation

The documentation says that, at each checkpoint, Trio checks whether the task is cancelled:

* [*Trio's core functionality* → *General principles* → *Checkpoints*](https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints) says: "[A checkpoint is] a point where Trio checks for cancellation. For example, if the code that called your function set a timeout, and that timeout has expired, then the next time your function executes a checkpoint Trio will raise a Cancelled exception. See Cancellation and timeouts below for more details."
* [*Design and internals* → *User-level API principles* → *Cancel points and schedule points*](https://trio.readthedocs.io/en/stable/design.html#cancel-points-and-schedule-points) says: "a *cancel point* is a point where your code checks if it has been cancelled – e.g., due to a timeout having expired – and potentially raises a Cancelled error."

Neither is really explicit about *when within the checkpoint* the cancellation state is checked. However, my subjective reading of them (especially the first one) is that is that they strongly imply that it is impossible for `await trio.foo()` to finish with anything other than a `trio.Cancelled` if it is in a cancelled scope (barring usual shielding rules) when it completes.

Most trio operations are conveniently in two parts: one that waits for the operation to become possible (e.g., socket read available, memory channel non-empty) and then one that synchronously performs that operation. So it certainly seems *possible* for the cancellation check to happen at the end of the wait step, in addition to before and (effectively) during.

The scheduling point would also happen alongside the wait, even if the condition is already true (and remains true). So, even if there is no actual wait, then it's possible that a cancellation would come into effect during the course of the `await` call, and the position of the cancellation check(s) still matters.

### Actual behaviour

In reality, it is possible for a task to wake from an `await` without exception even if it is in a scope that has been explicitly cancelled. 

[This gist](https://gist.github.com/arthur-tacca/2ef249eef1c5e41e05620ddd842a38d7) gives a complete example.

It involves running two tasks in a nursery:

* One sends a few values to a memory channel (with `send_nowait()`) then immediately cancels the cancel scope for the nursery.
* The other awaits `receive()` on the channel in a loop.

When I run it, the receive task gets one value from the memory channel before getting the cancellation exception.

### Side note about asyncio

Just for curiosity, I tried analogous code with asyncio, using an `asyncio.TaskGroup` in place of a nursery and simulating a cancel by raising an exception from one of the tasks. It appears that `asyncio.Queue.get()` is not a scheduling point, because a receive loop will happily pop 100s of values off of it in a loop after the task that send them has long finished with an exception (but a small perturbation to the code will stop cause it to be cancelled before getting any).

arthur-tacca:
Fixes #2973

* `reschedule()` gets a new parameter `allow_abort: bool` which, if True, still allows the abort function to be called right up until the task actually resumed. It is False by default so it's backwards compatible.
* When `Event.wait()` is woken by `Event.set()` it uses this parameter, so cancellation of it always succeeds

**Not ready to merge** I think the code is fine but:

* No tests 
* No docs

### DIFF
--- a/src/trio/_core/_generated_run.py
+++ b/src/trio/_core/_generated_run.py
@@ -101,7 +101,9 @@ def current_root_task() -> Task | None:
         raise RuntimeError("must be called from async context") from None
 
 
-def reschedule(task: Task, next_send: Outcome[Any] = _NO_SEND) -> None:
+def reschedule(
+    task: Task, next_send: Outcome[Any] = _NO_SEND, allow_abort: bool = False
+) -> None:
     """Reschedule the given task with the given
     :class:`outcome.Outcome`.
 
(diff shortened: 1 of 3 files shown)


arthur-tacca:
As [njsmith's comment](https://github.com/python-trio/trio/issues/2973#issuecomment-5404769454) says, the `wait_task_reschedule()` API is already complicated enough and this increases that for little demonstrable benefit. Closing.

PR #3497, opened 2026-08-29
Fix Trio imports on Android without interface helpers

### LINKED ISSUE #3493: Android API 24 runtime still lacks `if_indextoname` and `if_nametoindex`
## Problem

  Trio still fails to import on an Android CPython 3.13 runtime because  `socket.if_indextoname` and `socket.if_nametoindex` are not available. The failure is:
  ```python
  ImportError: cannot import name 'if_indextoname' from 'socket'
```
 The [failing code](https://github.com/python-trio/trio/blob/a78887af4a3fb20b33fc2be6947d0495cc1a03bd/src/trio/socket.py#L71-L82) is currently:
```python
if sys.implementation.name == "cpython":
    from socket import (
        if_indextoname as if_indextoname,
    )
```

## Context
This appears related to #2915 and PR #2916.

PR #2916 added `ImportError` suppression for `if_nameindex`, but the imports for `if_indextoname` and `if_nametoindex` remained unprotected, because they were believed to be available on Android. However, they are missing in the CPython Android runtime used by our application.

Environment:
  - Python: 3.13
  - Architecture: Android arm64
  - Android min API: 24
  - Android target API: 36
  - Trio: 0.33.0 and 0.34.0

A downstream workaround is:
```python
  if sys.implementation.name == "cpython":
      with suppress(ImportError):
          from socket import (
              if_indextoname as if_indextoname,
          )
```
  ## Expected behavior

  Trio should either:

  1. suppress these optional imports as well; or
  2. explicitly document and enforce the Android runtime/API requirements needed for these functions to exist.

  Since the functions are not needed by our application or by the Trio server used for TCP communication, suppressing the import appears to be the most compatible behavior.

mikemikimike:
## Summary

Fix Trio imports on Android API levels where CPython does not expose `socket.if_indextoname` and `socket.if_nametoindex`.

## Changes

- Make both interface-name helpers optional at runtime, matching the existing handling for `if_nameindex`.
- Add a regression test that imports `trio.socket` with the two standard-library helpers removed.
- Add a bugfix news fragment for issue #3493.

## Validation

- `PYTHONPATH=src python -m pytest src/trio/_tests/test_socket.py -q` — 31 passed, 4 skipped.
- `uv run --with 'ruff>=0.14' ruff check src/trio/socket.py src/trio/_tests/test_socket.py` — passed.
- `python -m black --check src/trio/socket.py src/trio/_tests/test_socket.py` — passed.
- `uv run --with 'mypy>=1.18' mypy src/trio/socket.py` — passed.
- `uv run --with build --with setuptools python -m build --wheel --no-isolation` — passed.
- `git diff --check` — passed.

The repository's system Ruff 0.15.0 cannot parse the current configuration's newer `builtin-argument-shadowing` selector, so the compatible Ruff version was supplied through `uv`. Type checking the whole test module with `--follow-imports=skip` reports pre-existing test/decorator and ignored-import diagnostics; the changed production module passes mypy.

## Compatibility

On platforms that provide these functions, Trio continues to re-export them unchanged. On platforms that do not, importing `trio.socket` succeeds and the unavailable names are simply absent, consistent with existing platform-specific exports.

Fixes #3493.

### DIFF
--- a/newsfragments/3493.bugfix.rst
+++ b/newsfragments/3493.bugfix.rst
@@ -1,0 +1,1 @@
+Allow importing Trio on Android API levels where ``socket.if_indextoname`` and ``socket.if_nametoindex`` are unavailable.
--- a/src/trio/_tests/test_socket.py
+++ b/src/trio/_tests/test_socket.py
@@ -4,6 +4,7 @@ import errno
 import inspect
 import os
 import socket as stdlib_socket
+import subprocess
 import sys
 import tempfile
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


A5rocks:
Sorry but a) I want to see what CPython will do first and b) don't use LLM generated text for PR descriptions please!

(Also I think the test goes about this the wrong way and the code comment isn't helpful.)

PR #3502, opened 2026-09-03
Ignore stale kqueue events for an already-deregistered key

### LINKED ISSUE #3500: KeyError in KqueueIOManager.process_events (TrioInternalError)
We have a PySide6 (Qt) application using Trio in guest mode. We are occasionally seeing the error below:

```
  File "/Users/dev/spire/.venv/lib/python3.14/site-packages/trio/_core/_run.py", line 2765, in unrolled_run
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/dev/spire/.venv/lib/python3.14/site-packages/trio/_core/_io_kqueue.py", line 93, in process_events
    receiver = self._registered[key]
               ~~~~~~~~~~~~~~~~^^^^^
KeyError: (46, -2)

  File "/Users/dev/spire/spire-desktop/src/quicore/asyncio.py", line 142, in event
    event.fn()
    ~~~~~~~~^^
  File "/Users/dev/spire/.venv/lib/python3.14/site-packages/trio/_core/_run.py", line 1752, in guest_tick
    self.done_callback(Error(exc))
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/Users/dev/spire/spire-desktop/src/quicore/asyncio.py", line 130, in _done_callback
    result = outcome_.unwrap()
  File "/Users/dev/spire/.venv/lib/python3.14/site-packages/outcome/_impl.py", line 213, in unwrap
    raise captured_error
  File "/Users/dev/spire/.venv/lib/python3.14/site-packages/trio/_core/_run.py", line 1746, in guest_tick
    timeout = self.unrolled_run_next_send.send(self.unrolled_run_gen)
  File "/Users/dev/spire/.venv/lib/python3.14/site-packages/outcome/_impl.py", line 185, in send
    return gen.send(self.value)
           ~~~~~~~~^^^^^^^^^^^^
  File "/Users/dev/spire/.venv/lib/python3.14/site-packages/trio/_core/_run.py", line 2945, in unrolled_run
```

Unfortunately, I'm not able to reproduce the error reliably, nor am I certain about where to go looking for the issue. Any insight would be most appreciated.

afonsojanu:
Closes #3500.

With a PySide6 app running trio in guest mode, `process_events()` in the kqueue backend can occasionally blow up with `KeyError: (46, -2)`, which guest mode then wraps into a fatal `TrioInternalError`.

The root cause is a timing gap between `get_events()` and `process_events()`. In guest mode, `GuestState.guest_tick` fetches a batch of ready kqueue events with a non-blocking `get_events(0)` right after resuming the scheduler for the current tick, then hands that batch off to be delivered on the *next* tick, once the host loop gets around to scheduling it. `notify_closing()` is a plain synchronous function, not tied to a checkpoint, so it can run in that gap and remove the fd/filter key from `_registered` before the already-fetched event for that same key gets delivered. `process_events()` looked the key up with a bare `self._registered[key]`, so a stale event for a since-deregistered key crashed the whole run.

The epoll backend doesn't have this problem because its `_registered` is a `defaultdict`, so a lookup for a missing key just returns an empty waiter set instead of raising. This PR gives kqueue the same tolerance: a missing key means there's nothing left to wake up, so the event is skipped.

Added a regression test that constructs a `KqueueIOManager` directly, registers then deregisters a key (mimicking what `notify_closing` does), and feeds a matching event straight into `process_events()`. It reproduces the exact `KeyError` on the old code and passes with the fix. Ran the full `_core` test suite locally on macOS (arm64), 277 passed.

### DIFF
--- a/newsfragments/3500.bugfix.rst
+++ b/newsfragments/3500.bugfix.rst
@@ -1,0 +1,1 @@
+On the kqueue backend (macOS/BSD), `~trio.lowlevel.notify_closing` could race with an already-fetched batch of kernel events under guest mode, causing `TrioInternalError: internal error in Trio - please file a bug!` wrapping a `KeyError` from inside `process_events`. Such stale events are now ignored instead of crashing the run.
--- a/src/trio/_core/_io_kqueue.py
+++ b/src/trio/_core/_io_kqueue.py
@@ -90,7 +90,15 @@ class KqueueIOManager:
             if event.ident == self._force_wakeup_fd:
                 self._force_wakeup.drain()
                 continue
-            receiver = self._registered[key]
+            receiver = self._registered.get(key)
+            if receiver is None:
+                # In guest mode, get_events() can fetch a batch of kernel
+                # events ahead of when they actually reach process_events()
+                # on a later tick. notify_closing() (or a completed
+                # wait_kevent abort) can deregister this same key in that
+                # window, so by the time the stale event gets here there's
+                # nothing left to deliver it to.
+                continue
             if event.flags & select.KQ_EV_ONESHOT:  # TODO: test this branch
                 del self._registered[key]
             if isinstance(receiver, _core.Task):
--- a/src/trio/_core/_tests/test_io.py
+++ b/src/trio/_core/_tests/test_io.py
@@ -426,6 +426,7 @@ async def test_io_manager_kqueue_monitors_statistics() -> None:
     sys.platform == "win32" or sys.platform == "linux",
     reason="only the kqueue backend has this failure mode",
 )
+@pytest.mark.filterwarnings("ignore:.*UnboundedQueue:trio.TrioDeprecationWarning")
 async def test_kqueue_process_events_tolerates_stale_deregistered_key() -> None:
     # The pytest.mark.skipif above keeps this from running on Linux/Windows,
     # but mypy doesn't understand that decorator and still type-checks this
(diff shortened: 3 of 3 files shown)


A5rocks:
Please first write a reproducer that doesn't look into the attributes for the kqueue manager. You claim `notify_closing` is enough; then just use that!

afonsojanu:
Fair point, let me redo this without touching the internals.

The reproducer I'll use instead: register a socket, then close it through the public API path that ends up calling notify_closing (rather than reaching into the kqueue manager's own state), and confirm the stale re-registration attempt on the next loop iteration doesn't raise. If notify_closing really is enough on its own, that should be sufficient to demonstrate it without any attribute peeking.

Give me a bit to rework the test and push an update.

afonsojanu:
Done — reworked the test to go through monitor_kevent instead. It registers a real key, lets the context manager's own exit deregister it (the same path notify_closing relies on for a waiting task), then feeds process_events a stale event for that now-gone key. No peeking at _registered directly anymore.

Confirmed it still catches the original bug: reverting just the fix in _io_kqueue.py makes this new version fail with the same KeyError as before, and it passes again once the fix is back.

A5rocks:
...?? That's still internals!

A5rocks:
I tried to make a test using only `trio` and `trio.lowlevel`, but I failed. I'm fairly certain Claude (or ChatGPT...) is wrong here! In particular, since we register events as `KQ_EV_ONESHOT`, they get deleted when they received. This means that deleting them with `KQ_EV_DELETE` fails in `notify_closing`. Therefore they will still exist when `process_events` happens!

A5rocks:
I'm like 90% convinced the repro here (`notify_closing` between `get_events` and `process_events`) is wrong. I added a test in https://github.com/python-trio/trio/pull/3503 and while it does cause an error, it raises an error solely in `notify_closing` (which is an error and is fixed in that PR).

I notice you use `monitor_kevent` instead of the others in the test case so I modified the test case in that PR:
```py
@contextlib.contextmanager
def bweh():
    print("enter")
    try:
        yield
    finally:
        print("exit")

pair = socket.socketpair()
for sock in pair:
    sock.setblocking(False)

async def trio_main(in_host: InHost) -> None:
    with contextlib.ExitStack() as es:
        es.enter_context(trio.lowlevel.monitor_kevent(pair[0], select.KQ_FILTER_WRITE))
        es.enter_context(bweh())
        in_host(es.close)
        with contextlib.suppress(trio.ClosedResourceError):
            await trio.lowlevel.wait_writable(pair[0])  # blocks
        print("after wait_writable")

trivial_guest_run(trio_main)
for sock in pair:
    sock.close()
```

But on that branch I just see this output:
```
a5rocks@whimsy ~/D/trio> uv run repro.py
/Users/a5rocks/.local/share/uv/python/cpython-3.14.2-macos-aarch64-none/lib/python3.14/typing.py:1235: TrioDeprecationWarning: trio.lowlevel.UnboundedQueue is deprecated since Trio 0.9.0; use trio.open_memory_channel(math.inf) instead (https://github.com/python-trio/trio/issues/497)
  result = self.__origin__(*args, **kwargs)
enter
exit
after wait_writable
```

A5rocks:
Nevermind, I forgot a `.fileno()`. I'll throw up a new PR with that better test and your fix.

(edit: I think this runs into another bug, and fixing that makes the `.get` unnecessary...)

PR #3503, opened 2026-09-05
Sometimes notify_closing can be called on dead events

A5rocks:
Closes https://github.com/python-trio/trio/pull/3502 since this is the repro they were describing, I think.

### DIFF
--- a/newsfragments/3502.bugfix.rst
+++ b/newsfragments/3502.bugfix.rst
@@ -1,0 +1,1 @@
+Ensure `trio.lowlevel.notify_closing` works even on events that have been deleted, e.g. because they were registered as one shot events.
--- a/src/trio/_core/_io_kqueue.py
+++ b/src/trio/_core/_io_kqueue.py
@@ -280,7 +280,13 @@ class KqueueIOManager:
 
             if type(receiver) is _core.Task:
                 event = select.kevent(fd, filter_, select.KQ_EV_DELETE)
-                self._kqueue.control([event], 0)
+                try:
+                    self._kqueue.control([event], 0)
+                except OSError as e:
+                    if e.errno == errno.ENOENT:
+                        # the event isn't in kqueue
+                        continue
+                    raise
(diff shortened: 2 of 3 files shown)


A5rocks:
I'll close this in favor of the more general solution @arthur-tacca was proposing.

A5rocks:
Gah, I forgot what this PR did. NVM!

CoolCat467 (approved):
From what I understand this is probably fine.

arthur-tacca:
Sorry I just noticed that this only checks for `ENOENT`

```python
                except OSError as e:
                    if e.errno == errno.ENOENT:  # pragma: no branch
                        # the event isn't in kqueue
                        continue
```

But [for similar code in `_wait_common`](https://github.com/python-trio/trio/blob/50b9825f7ef6630c2326001cc9323c714c05355d/src/trio/_core/_io_kqueue.py#L196), Trio currently checks for `exc.errno in (errno.EBADF, errno.ENOENT)`. It says "FreeBSD reports this using EBADF. macOS uses ENOENT". Not sure if this applies to this situation too.

A5rocks:
Ah possible. I just took that errno from the relevant man page on macOS...

arthur-tacca:
Ok sorry, the error code thing is not a big deal in this case.

The bug that this pull request fixes is that a host callback (in guest mode) closes a socket after `get_events` in the Trio worker thread has already finished and one of the results is for the socket (so it's already removed from the kqueue).

Assuming the socket close is done with the normal ordering of `notify_closing` and then actual close, you'd get `errno.ENOENT` on both platforms. It doesn't make sense to return `errno.EBADF` because the file isn't actually closed yet. 

The trio docs says that's the recommended way for a close implementation to work, but does say it's also valid to do the actual close first and then `notify_closing`. If you do that then you really would get `errno.EBADF` on BSD. But that's an odd thing for the application to do, and it anyway doesn't cause a Trio internal error.

PR #3505, opened 2026-09-06
Document subprocess worker-thread context inheritance

roli-lpci:
## Summary

- document that `trio.run_process` and `trio.lowlevel.open_process` call
  `subprocess.Popen` from a worker thread
- explain that applicable per-thread operating-system state is therefore
  inherited from the worker thread, with Linux namespaces and CPU affinity as
  examples
- add the corresponding documentation release note

This addresses the documentation request in #3360. It leaves the API and
spawning behavior unchanged; #3462's synchronous-spawn option was closed because
blocking the Trio event loop is unacceptable.

## Validation

- a focused runtime probe wrapped the real `subprocess.Popen`, launched a child,
  and confirmed that `Popen` ran on a different OS thread from Trio's event loop
- `sphinx-build -W --keep-going -b html docs/source ...`
- `towncrier build --draft`
- file-scoped `pre-commit` checks for the two changed files

### DIFF
--- a/newsfragments/3360.doc.rst
+++ b/newsfragments/3360.doc.rst
@@ -1,0 +1,3 @@
+Document that `trio.run_process` and `trio.lowlevel.open_process` create
+subprocesses from a worker thread, which affects the per-thread operating-system
+state inherited by the child.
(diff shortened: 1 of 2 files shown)


A5rocks:
I'm not willing to waste my time reviewing this given how many open PRs you have + LLM generated PR description. I also think it's not exactly right because I don't think docs really address the issue, or at least I'm not yet convinced there isn't a solution.

PR #3507, opened 2026-09-08
Expose AsyncIOWrapper for public file type annotations

### LINKED ISSUE #3390: Type annotations for `trio.open_file` & `trio.wrap_file`
Hi,

I am currently adding type annotations to my projects. It seems that `trio.open_file` and `trio.wrap_file` returns an instance of the `trio._file_io.AsyncIOWrapper`. What puts me off a bit is the `_file_io` part, because names that begin with an underscore are usually meant to be private.

Is `trio._file_io.AsyncIOWrapper` the correct type annotation to use for the return value of `trio.open_file`? If so, does it not make sense to expose this type as part of Trio's public API, e.g. `trio.AsyncIOWrapper`, or perhaps `trio.abc.AsyncIOWrapper`?

vzer200:
Expose the existing generic file-wrapper type as `trio.AsyncIOWrapper`, so callers can annotate values returned by `open_file()`, `wrap_file()` and `Path.open()` without importing a private module.

Fixes #3390 and follows the discussion on the closed #2864. The class follows Trio's final-class convention. The documentation describes its wrapped-file type parameter, and runtime/type tests now use the public import. Static class inspection accounts specifically for its existing dynamically supplied methods and properties; unexpected differences still fail the checks.

Validation:

- The public import failed before the export. All 62 related runtime/export checks pass, including the slow static-tool checks (one optional test skipped).
- Project typing checks pass: mypy for Linux, macOS and Windows, plus both pyright type-test directories. Sphinx with warnings as errors and all applicable pre-commit hooks pass.
- On this Chinese-language Windows host, the full runtime suite reports 805 passed, 85 skipped, two expected failures and five failures matching English OS error messages. The unchanged base has exactly the same results. The type-completeness helper likewise reports the same existing platform/docstring diagnostics on both versions, with no new wrapper diagnostic.

AI-assisted implementation, independently reviewed by other coding agents.

### DIFF
--- a/docs/source/reference-io.rst
+++ b/docs/source/reference-io.rst
@@ -657,6 +657,8 @@ Asynchronous file objects
 
 .. autofunction:: wrap_file(file)
 
+.. autoclass:: AsyncIOWrapper()
+
 .. interface:: Asynchronous file interface
 
    Trio's asynchronous file objects have an interface that
--- a/newsfragments/3390.feature.rst
+++ b/newsfragments/3390.feature.rst
@@ -1,0 +1,2 @@
+The type returned by `trio.open_file`, `trio.wrap_file`, and `trio.Path.open`
+is now publicly available as `trio.AsyncIOWrapper` for use in type annotations.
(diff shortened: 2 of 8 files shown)


A5rocks:
I'm not willing to waste my time going through the issues with this PR (I think the testing strategy is wrong w/r/t handling missing attributes and probably this would need a mention in the docs?), given you have a bunch of open PRs + the PR explanation is LLM-generated.

PR #3508, opened 2026-09-10
Document ParkingLot.broken_by

bpiper02:
Closes Part of #3221

Adds documentation for `ParkingLot.broken_by` and removes it from the documentation allowlist.

Validation:
- `python -m pytest src\trio\_core\_tests\test_parking_lot.py`
- `python -m ruff check src\trio\_core\_parking_lot.py docs\source\conf.py`
- `python -m ruff format --check src\trio\_core\_parking_lot.py docs\source\conf.py`
- `python -m compileall src\trio\_core\_parking_lot.py docs\source\conf.py`

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -174,7 +174,6 @@ def autodoc_process_signature(
 logger = getLogger("trio")
 UNDOCUMENTED = {
     "trio._subprocess.HasFileno.fileno",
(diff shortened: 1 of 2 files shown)


A5rocks:
> As long as you don't try to use an LLM to do it, probably.

That applies to PR description too 😭

PR #3510, opened 2026-09-10
Detect unclosed cancel scopes when a task exits

### LINKED ISSUE #3329: More robust CancelScope misnesting prevention
@agronholm brought this case up on Gitter:

```py
import trio

async def main():
    async def shim(fn, *args):
        fn(*args)

    scope = trio.CancelScope()
        nursery.start_soon(shim, scope.__enter__)

        nursery.start_soon(shim, scope.__exit__, None, None, None)

trio.run(main)
```

I think we should probably error here. I took a look at the code and I think the issue is that our CancelScope misnesting prevention only works within a single task -- that is, we don't check that there's no CancelScope currently entered at the end of any task. Unfortunately, IIRC that's hard in that situation to raise an exception with a nice stack trace. As a stopgap, we could raise when exiting the nursery if anything gets marked as misnested. (... which would raise later but should catch all of these misnestings?)

---

I'm making mostly unsubstantiated conjectures here, based on a few minutes of playing around with the code. There's very likely something I missed!

Sakshamm-Goyal:
## Summary

Fixes #3329.

A task can enter a `CancelScope`, return without exiting it, and leave another task able to exit that scope. The existing task-exit checks catch abandoned scope trees and unclosed child nurseries, but not this still-attached scope.

Detect a task exiting with a cancellation status different from its parent nursery's status. Report a scope-stack corruption error, preserving the original exception as context when present, and clean up the unclosed scopes and their registered deadlines. Existing child-nursery recovery is unchanged.

Regression coverage includes nested scopes, cancellation, `nursery.start()` adoption, guest mode, original exception context and enclosing parent-scope preservation. A release note is included.

## Validation

- The issue's real two-task reproduction fails against unchanged main at `a3d7cddeacef74c87f667e9f9e8615f2ab7ce894` because no corruption error is raised.
- Current Python 3.10 installed-wheel suite, with slow tests and CI-style isolation/plugin setup: 881 passed, 24 skipped, 2 xfailed.
- Complete core suite: 287 passed, 11 skipped, 1 xfailed.
- Mypy checks over `src/trio` pass for Darwin, Linux and Windows targets.
- All applicable configured pre-commit hooks pass, including generated-file regeneration, Black, Ruff, spelling and Sphinx lint.
- Python 3.12 full-suite validation has a Jedi `__notes__` failure also reproduced on unchanged upstream; Python 3.10's corresponding checks pass. No claim of full platform CI coverage.

## Related work

This covers a remaining case after #3307: the exiting task has an unclosed scope but no unclosed child nursery. It does not change the cancelled-task adoption policy discussed in #2896.

## AI assistance

This change and its tests were developed and validated with OpenAI Codex assistance.

### DIFF
--- a/newsfragments/3329.bugfix.rst
+++ b/newsfragments/3329.bugfix.rst
@@ -1,0 +1,3 @@
+Trio now reports an error when a task exits without closing a cancel scope it
+entered, instead of allowing that scope to be exited by another task. Deadlines
+associated with these unclosed scopes are also cleaned up.
--- a/src/trio/_core/_run.py
+++ b/src/trio/_core/_run.py
@@ -2026,11 +2026,21 @@ class Runner:  # type: ignore[explicit-any]
                 lot.break_lot(task)
             del GLOBAL_PARKING_LOT_BREAKER[task]
 
+        unclosed_scope = (
+            task._parent_nursery is not None
+            and task._cancel_status is not task._parent_nursery._cancel_status
+            and task._cancel_status is not None
+            and not task._cancel_status.abandoned_by_misnesting
+        )
         if (
-            task._cancel_status is not None
-            and task._cancel_status.abandoned_by_misnesting
-            and task._cancel_status.parent is None
-        ) or task._child_nurseries:
+            (
+                task._cancel_status is not None
+                and task._cancel_status.abandoned_by_misnesting
+                and task._cancel_status.parent is None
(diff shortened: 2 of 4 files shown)


A5rocks:
Sorry, this topic intersects too much to easily review and LLM generated PR description doesn't provide confidence it's worth reviewing.

PR #3515, opened 2026-09-17
Avoid SSL tests on x86 Windows as cryptography doesn't support it

### LINKED ISSUE #3466: Don't run SSL tests on Windows x86

A5rocks:
Fixes https://github.com/python-trio/trio/issues/3466

This is actually somewhat similar to https://github.com/python-trio/trio/pull/3468 since it turns out `cryptography` doesn't ship wheels for x86 windows anymore (wow! unbelievable!), and so we try building it and that fails since it's Windows. Either we can figure out all the right environment (e.g. have openssl), or we can avoid installing `cryptography`. (I may want to rename the environment variable we check, too.)

Additionally, we cannot avoid installing `cryptography` using environment markers, since `platform_machine` is for the machine rather than the build of Python, and the environment markers are in general exactly the same.

The upside is that this makes the actual skipping logic extremely simple!


ci.sh
```diff
@@ -54,8 +54,13 @@ if [ "${NO_TEST_REQUIREMENTS-0}" == 1 ]; then
     python -m uv pip install pytest coverage -c test-requirements.txt
     flags="--skip-optional-imports"
 else
-    python -m uv pip install -r test-requirements.txt --no-deps
     flags=""
+    if [ "x$ARCH" == "xx86" ]; then
```
CoolCat467: Why not just do `[ "$ARCH" == "x86" ]`?

ci.sh
```diff
@@ -54,8 +54,13 @@ if [ "${NO_TEST_REQUIREMENTS-0}" == 1 ]; then
...
+        echo cryptography >excluded_deps
+        PIP_ARGS="--exclude excluded_deps"
```
CoolCat467, line 60: Curious how this works, with my current understanding it looks like we're writing `cryptography` to a file named `excluded_deps` and then doing `--exclude` with this file? Why not just do `--exclude cryptography`?

ci.sh
A5rocks: IIRC there's a bash footgun w/ variable expansion of empty variables and comparison... I should double check though.

ci.sh
A5rocks: `--exclude` takes in a requirements file (I think? thats what the docs say at least). I'm kinda surprised there's no better way, but oh well!

PR #3516, opened 2026-09-17
Document wrap_file ownership and concurrent use

### LINKED ISSUE #3379: trip.wrap_file() undocumented behavior wrt to sync file object

pentaoa:
Fixes #3379.

`wrap_file()` now documents the lifecycle and concurrency details that were previously implicit:

- `aclose()` closes the supplied synchronous file object.
- Dropping the wrapper does not explicitly close that object, so callers should use `aclose()` or an async context manager.
- The wrapper and original reference share position and buffers.
- Direct and wrapped operations may run concurrently only when the underlying object is thread-safe.

The wording lives in the function docstring, so it appears directly under `trio.wrap_file` in the API reference. A documentation news fragment is included.

Validation:

- Ruff
- Black
- sphinx-lint
- codespell


CoolCat467:
First off, this is almost certainly LLM generated, and will probably close if I get only LLM responses.
It feels useless to me to add documentation about the workings of `AsyncIOWrapper` to `wrap_file`. Why not just document it in `AsyncIOWrapper`? I'm pretty sure it's not exported currently, but that could change relatively simply.

pentaoa:
My thinking at the time was that `wrap_file` is the public entry point, while `AsyncIOWrapper` is not currently publicly exported. That’s why I structured the PR this way—I thought putting the usage caveat at the entry point would make it easier for users to discover.
That said, this PR was indeed generated with the help of an LLM, and it was my mistake not to disclose that. As for whether the PR should be merged, I’m happy to defer to your judgment.

PR #3519, opened 2026-09-20
Fix 3500: kqueue issues

arthur-tacca:
Four bugs in the kqueue IO manager came up in the discussion of #3500:

1. In guest mode, after `get_events()` has finished and a readable/writable wait has completed, if a callback in the host loop cancels the waiting task then the loop will later raise `TrioInternalError`. (I'm pretty sure that this is the reporter's problem.)
2. In guest mode, after `get_events()` has finished and a readable/writable wait has completed, if a callback in the host loop closes a socket then `notify_closing` will raise an exception.
3. In guest mode, after `get_events()` has finished and a `Process.wait()` has completed, if a callback in the host loop cancels the waiting task then `cancel` will raise an exception (and possibly corrupt the cancel state).
4. The `wait_kevent` API awkwardly requires the caller to register the event with the kqueue before `wait_kevent` checks whether another caller is already waiting for that event. This is inconvenient and could possibly lead to one caller modifying the event of another. It was deliberately written this way for extra flexibility but [NJS suggested in 2018](https://github.com/python-trio/trio/issues/578#issuecomment-408760107). that it be changed to just register the event directly itself.

There are three pull requests, including this one, for this issue:

#3502 - this is the right fix for issue 1 but labelled as a fix for 2 (with a test that didn't cover 1). (In fairness it is *part* of a potential fix for 2.)

#3503 - fixes 2 (now merged). It is self contained so didn't need the fix for 1.

This pull request has three commits:

* The first commit slightly modifies the fix for 2 to be slightly more robust - it fixes the case where the file is closed before notify_closing is called on BSD (with a test for it)
* The second commit fixes 1, with the same solution as the first pull request but correct description of the bug it fixes and suitable test 
* The third fixes 4 for `wait_kevent` (not `monitor_kevent`) by changing its interface as discussed on the issue (with fallback for the old deprecated API, and tests for both new and old interfaces). Also, as a side effect, it fixes 3 (with test for that too).

I have tested all the above on FreeBSD 15.1.


src/trio/_core/_io_kqueue.py
```diff
@@ -143,64 +147,114 @@ def monitor_kevent(
...
+                # errno.ENOENT on both platforms.
                 if exc.errno in (errno.EBADF, errno.ENOENT):  # pragma: no branch
                     pass
                 else:  # pragma: no cover
                     # As far as we know, this branch can't happen.
                     raise
             return _core.Abort.SUCCEEDED
 
-        await self.wait_kevent(fd, filter, abort)
+        # wait_task_rescheduled does not have its return type typed
```
A5rocks, line 256: I think this comment is implied by the type ignore.

src/trio/_core/_io_kqueue.py
```diff
@@ -90,7 +91,13 @@ def process_events(self, events: EventResult) -> None:
             if event.ident == self._force_wakeup_fd:
                 self._force_wakeup.drain()
                 continue
-            receiver = self._registered[key]
+            receiver = self._registered.get(key)
+            if receiver is None:
```
A5rocks, line 95: Could we have a set of events where it's ok to be missing? e.g. the abort callback in `wait_kevent` adds the pair to a set that gets cleared at the beginning of `get_events`. I think that would make any bugs in the kqueue manager fail loudly (well, hopefully there aren't any more) as well as provide some documentation: someone could simply look for all the locations where we add to this set to know the problematic areas.

A5rocks (commented):
Some very early review!

A5rocks:
Also we should probably readd freebsd as a CI platform... I'll poke more at that and try to make a PR soon.

src/trio/_core/_io_kqueue.py
arthur-tacca: This comment was already there (at the end of `wait_kevent`, though I duplicated it in the new early return), it just looks like I added it because the diff has focused on the merge of`_wait_common` into it. But I agree and I'm happy to remove the comment (from both places).

src/trio/_core/_io_kqueue.py
arthur-tacca: I think this should work. I was even going to suggest it in one of my comments but didn't include it as I thought I'd rambled on too long! And it's much more feasible now that cancellations all go through one place (not including `monitor_kevent`). There are two issues with clearing at the start of `get_events` though: 

1. `get_events()` can run on the worker thread, so it's possible, albeit unlikely, that a host callback will happen to run and set an entry just before get_events() clears it 
2. This strategy needlessly continues building up the entries after `process_events` i.e. in regular Trio non-guest code. 

I'd set a Boolean flag at the start of `get_events()` (so it would be set in the host thread in the initial `get_events(0)` in guest mode) and clears it in `process_events` and only add to the set in between; then `process_events` can clear the set at the end.

arthur-tacca:
I think these are the CI failures:

* The main substantial one is that the new test I added for issue 2, where a socket is closed before `notify_closing`, fails on Windows (`notify_closing` raises `OSError`: WinError 10038, "not a socket"). I'm not going down that rabbit hole so I hope it's ok to skip that test on Windows only. ([The docs for `notify_closing`](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.notify_closing) say "you usually want to" call `notify_closing` before closing the socket, but "It’s also possible to do them in a different order if that’s more convenient, but only if you make sure not to have any checkpoints in between the steps." It seems tempting to me to change that to say it's only reliable if `notify_closing` is called first.)
* Coverage failures: 
  * One test has an abort function that isn't called (because its task isn't cancelled). I just need to add a pragma for that (and replace the abort function with pytest.fail)
  * A small code block I moved around in the `_io_kqueue.py` had a "TODO test this branch" comment on an if statement. I deleted the comment, not realising that it was stopping a coverage failure. I'll put it back. 
* I forgot to regenerate the generated IO files after I made a couple of trivial last minute docstring tweaks
* The new kqueue tests file fails type checking on non macOS/BSD because the kqueue functions aren't defined; I need to mark that test file as macOS/BSD only.
* The failure to regenerate docs looks unrelated (a recent change in Read The Docs config).

A5rocks:
> * The main substantial one is that the new test I added for issue 2, where a socket is closed before `notify_closing`, fails on Windows (`notify_closing` raises `OSError`: WinError 10038, "not a socket"). I'm not going down that rabbit hole so I hope it's ok to skip that test on Windows only. ([The docs for `notify_closing`](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.notify_closing) say "you usually want to" call `notify_closing` before closing the socket, but "It’s also possible to do them in a different order if that’s more convenient, but only if you make sure not to have any checkpoints in between the steps." It seems tempting to me to change that to say it's only reliable if `notify_closing` is called first.)

Let's make thinking about this a followup issue, I agree. I'm more inclined to poking at why this happens, because maybe we just need to ignore certain OSErrors? i.e. just like what we did for `notify_closing` on freebsd, where we ignore `EBADF`.

arthur-tacca:
OK, I've opened #3520 to address what to do about close after notify_closing. (Update documentation? Fix notify_closing on Windows?)

As part of researching that, I noticed that #3503 broke notify_closing after close on kqueue. (I had a test for that but only in the obscure race condition that this issue is about - it doesn't test a normal non-guest mode usage.) I can fix that as part of this pull request by implementing the other fix for `notify_closing` that I stated in [this comment](https://github.com/python-trio/trio/issues/3500#issuecomment-5587090240).

A5rocks:
Wdym it broke notify_closing after close? I don't immediately see the case you're thinking about...

Edit: nvm I see you mention this in the issue you made. Sorry, I should have read that first!

PR #3522, opened 2026-09-22
Fix #2607: handle frozen dataclass exceptions

### LINKED ISSUE #2607: collapse_exception_group breaks with exceptions that are also frozen dataclasses
Quick preface:

+ this only happens with ``strict_exception_groups=False``, and enabling strict exception groups fixes the "problem"
+ I'm not even sure I would necessarily consider this a "problem" with the code per se, but it should probably be at least documented somewhere
+ I realize that ``strict_exception_groups=False`` is on the deprecation march, so even if this _is_ considered a problem, it might not be worthwhile to fix it
+ feel free to immediately close as wontfix!

That being said, in the interests of "more information is always better than less information" and "something something posterity", I thought I'd throw up a quick flag: [collapse_exception_group](https://github.com/python-trio/trio/blob/master/trio/_core/_run.py#L147) breaks if the exception it's trying to collapse also happens to be a frozen dataclass. When trio tries to set the ``__traceback__`` attribute, dataclass' ``__setattr__`` magic kicks in, and you get a FrozenInstanceError. Also, since this happens in the middle of nursery exception management, it kinda... blows everything up.

Again, for me, the solution was suuuuper simple -- just enable strict exception groups and update the rest of my code to match. Voila, and I get futureproofing free of charge! But it wasn't immediately clear to me that that would be a viable solution, until I poked around some inside the trio codebase. So it might be helpful to document that somewhere (I have no idea where, or I'd just have directly opened up a PR).

sb123sb123:
Fixes #2607

When `strict_exception_groups=False` collapses a single exception group, Trio updates traceback and context metadata on the contained exception. Frozen dataclass exceptions reject normal attribute assignment, so the original exception is replaced by `FrozenInstanceError`.

This change uses `object.__setattr__` for Trio internal traceback/context restoration paths and adds a regression test through a real nursery.

Testing:

- `PYTHONPATH=src .venv/Scripts/python.exe -m pytest src/trio/_core/_tests/test_run.py -q` - 155 passed, 1 skipped
- `PYTHONPATH=src .venv/Scripts/python.exe -m pytest src/trio/_tests/test_util.py -q` - 8 passed
- `.venv/Scripts/ruff.exe check src/trio/_core/_run.py src/trio/_core/_tests/test_run.py src/trio/_util.py` - passed
- `python -m black --check src/trio/_core/_run.py src/trio/_core/_tests/test_run.py src/trio/_util.py` - passed
- Direct Windows reproduction with a frozen dataclass exception - returned the original `FrozenError`

A baseline run of `src/trio/_core/_tests/test_io.py` had 28 passed and 1 failure because this Windows host emits localized text for `WinError 10038` while the existing test regex only accepts the English message; this is unrelated to the change.

AI assistance was used to investigate, implement, and test this patch. The reported results are from commands actually run on the Windows host.


A5rocks:
Thanks, but I don't want to fix this. `strict_exception_groups=False` is deprecated and is due to be removed in a couple months. Besides, this is a behavior change. Maybe some exceptions rely on their custom `__setattr__` in ways other than being frozen!

